PR #348, opened 2018-03-30
CREATEing SolrCore, Error  non legacy mode coreNodeName missing {schema=schema.xml, dataDir=data, config=solrconfig.xml}

hadoop835:


ctargett:
The number of commits and files changed here look like it's an accidental PR. Closing.

PR #350, opened 2018-04-03
SOLR match mode change for the rouding off instead of taking floor

MighTguY:
Instead of directly using the floor of  the mm %, we need to take the round of value, 
Ecample, 
When we have 5 boolean clauses and mm=75%, we get calc as 3.75 currently it took 3, so instead of 3 it should have taken 4. as Math.round()


tflobbe:
I didn't look at the change, but you should create a Jira issue in https://issues.apache.org/jira/projects/SOLR first. Then you can change the title of the PR to include the Jira code to get them linked.

PR #353, opened 2018-04-06
SOLR-12194: Deprecate SolrRequest#setBasicAuthCredentials

janhoy:
First step
Still need to modify `BasicAuthIntegrationTest` and document in ref-guide, but that could also be done in separate issues.


janhoy:
Closed since we decided not to do this after all

PR #354, opened 2018-04-17
Change the method identifier from "getShardNames" to "addShardNames".

BruceKuiLiu:
The method seems to add "sliceName" to "shardNames", thus the method name "addShardNames" should be better "getShardNames" since "get" means getting something.


tflobbe:
You should create a Solr Jira, and then rename this PR to include the Jira code, that way it'll get visibility

ShangwenWang:
There is already a List variable named _shardNames_ and the method's function is to add content to this list.
So how about naming these methods **addNames**? Is it more suitable?

PR #355, opened 2018-04-17
SOLR-12233

millerjeff0:
Instead of recreating the classes every time we reload a solr core, just create them once.  The same is done in other classes such as TransformerFactory


ctargett:
https://issues.apache.org/jira/browse/SOLR-12233 was released in Solr 7.4, so this PR can be closed.

PR #356, opened 2018-04-18
SOLR-12232

millerjeff0:
Update to use uninterruptible file channel due to interrupted threads closing connection until solrcore is recreated


ctargett:
https://issues.apache.org/jira/browse/LUCENE-8262 (which replaced SOLR-12232) was closed with Won't Fix, so this PR should be closed too.

PR #357, opened 2018-04-19
[SOLR-12238] Synonym Queries boost

alessandrobenedetti:
This patch involves the synonym management in the query building/parser phase.
It adds the possibility of getting the payloads from a Graph Token Stream and then use the payload to build boost queries for terms and phrase queries.

Tests are included for a better understanding of the use cases.
This is an initial patch, happy to extend it or improve it !


alessandrobenedetti:
I have been doing merge conflicts, a huge cleaning up and re-design.
Most of the logic is now in Lucene Query Parser, this avoids passing back and forth token streams.
It seems to me a much cleaner approach, happy to discuss

TO DO:
This is what I'll do next week
- Add additional tests on Lucene Query Builder (that is green at the moment)
- discuss about Paylod decoding and possible use of org.apache.lucene.analysis.payloads.PayloadHelper
- discuss why "/*confirm autoGeneratePhraseQueries always builds OR queries*/ from org.apache.solr.search.TestSolrQueryParser#testSynonymQueryStyle

lucene/core/src/java/org/apache/lucene/util/QueryBuilder.java
```diff
@@ -349,25 +361,78 @@ protected Query createFieldQuery(TokenStream source, BooleanClause.Occur operato
...
+      } else {
+        return singleTermQuery;
+      }
     } else {
-      return new SpanNearQuery(terms.toArray(new SpanTermQuery[0]), 0, true);
+      SpanNearQuery multiTermQuery = new SpanNearQuery(terms.toArray(new SpanTermQuery[0]), 0, true);
+      if (queryPayloadBoost != 0) {
+        return new SpanBoostQuery(multiTermQuery, queryPayloadBoost);
+      } else {
+        return multiTermQuery;
```
diegoceccarelli: nit: maybe you can have a function:
```
private SpanNearQuery getBoostedSpanQuery(spanQuery, queryPayloadBoost){
    if (queryPayloadBoost != 0) {
        return new SpanBoostQuery(spanQuery, queryPayloadBoost);
      } else {
        return spanQuery;
      }
}
```
and reuse it in both the branches?

lucene/core/src/java/org/apache/lucene/util/QueryBuilder.java
```diff
@@ -349,25 +361,78 @@ protected Query createFieldQuery(TokenStream source, BooleanClause.Occur operato
...
+    for (BytesRef singlePayload : payloadsForQueryTerms) {
+      if (singlePayload != null) {
+        float decodedPayload = decodeFloat(singlePayload.bytes, singlePayload.offset);
+        return decodedPayload;
+      }
     }
+    return 0;
   }
 
+  public static final float decodeFloat(byte [] bytes, int offset){
```
diegoceccarelli: public intended?

lucene/core/src/java/org/apache/lucene/util/QueryBuilder.java
```diff
@@ -349,25 +361,78 @@ protected Query createFieldQuery(TokenStream source, BooleanClause.Occur operato
...
+
+    return Float.intBitsToFloat(decodeInt(bytes, offset));
+  }
+
+  public static final int decodeInt(byte [] bytes, int offset){
```
diegoceccarelli: public intended?

lucene/core/src/java/org/apache/lucene/util/QueryBuilder.java
```diff
@@ -509,38 +591,75 @@ protected Query analyzeGraphBoolean(String field, TokenStream source, BooleanCla
...
+        @Override
+        public BytesRef[] next() {
+          TokenStream sidePath = sidePathsForPayloads.next();
+          return getPayloadsFromStream(sidePath);
+        }
+      };
       if (graph.hasSidePath(start)) {
-        final Iterator<TokenStream> it = graph.getFiniteStrings(start, end);
-        Iterator<Query> queries = new Iterator<Query>() {
+        final Iterator<TokenStream> sidePaths = graph.getFiniteStrings(start, end);
```
diegoceccarelli: nit: maybe keep iterator in the name? `sidePathsIterator`?

alessandrobenedetti:
I followed the refactor comments from both @diegoceccarelli  and @romseygeek .
The PR seems much cleaner right now both Lucene and Solr side.
Copious tests are present and should cover the various situations.

Few questions remain:

- from a test I read a comment from @dsmiley  saying: "confirm autoGeneratePhraseQueries always builds OR queries" from org.apache.solr.search.TestSolrQueryParser#testSynonymQueryStyle

- what can we do for the SpanBoostQuery, I was completely not aware they are going to be deprecated

Let me know

dsmiley:
I noticed the use of {{AttributeSource[]}} (array of AttributeSource), done at the behest of @romseygeek .  That seems fishy... shouldn't it be a TokenStream, which is a more memory efficient iterator over AttributeSource changing state?  I see, for example, the _existing_ {{createSpanQuery(TokenStream in, String field)}} but the PR adds {{newSpanQuery(String field, AttributeSource[] attributes)}} and makes the former call the latter.  Why bother; why not retain createSpanQuery and if Solr wants to override it to do payload boosting then it can?

alessandrobenedetti:
No strong opinion on that, it was actually the first time I used the AttributeSource so I am happy to switch to TokenStream if it is more memory efficient.
The change shouldn't be too heavy.
I will just wait for confirmation, and when we are all aligned I'll proceed with the implementation.

alessandrobenedetti:
I would love to leverage the momentum and the fresh memory on this project to finalise the contribution, I didn't get any feedback on the TokenStream / AttributeSource controversy in the last few days...
If i don't get any additional feedback on that by mid next week, I'll take [~dsmiley] consideration and change the implementation toward the use of the TokenStream directly.
Super happy to receive additional feedback and considerations by [~romseygeek], maybe the AttributeSource approach was misunderstood?

solr/core/src/java/org/apache/solr/parser/SolrQueryParserBase.java
```diff
@@ -600,29 +617,133 @@ protected Query newRegexpQuery(Term regexp) {
...
+    }
+    if (isAcceptableBoost(payloadBoost)) {
+      return new SpanBoostQuery(query, payloadBoost);
+    }
+    return query;
+  }
+
+  private float getDecodedPayload(AttributeSource attribute) {
+    float payloadBoost = 0f;
+    PayloadAttributeImpl payloadAttribute = attribute.getAttributeImpl(PayloadAttributeImpl.class);
```
romseygeek: You can just use `PayloadAttribute` here?  The implementation is, well, an implementation detail :)

lucene/core/src/java/org/apache/lucene/util/AttributeSource.java
```diff
@@ -171,6 +171,23 @@ public boolean hasNext() {
...
+   *         {@link #addAttributeImpl} even in consumers  of TokenStreams, because you cannot 
+   *         know if a specific TokenStream really uses a specific AttributeImpl. 
+   *         {@link #addAttributeImpl} will automatically make the attribute impl available. 
+   *         If you want to only use the attribute , if it is available (to optimize
+   *         consuming), use {@link #hasAttribute}.
+   */
+  public final <T extends AttributeImpl> T getAttributeImpl(Class<T> attClass) {
+    return attClass.cast(attributeImpls.get(attClass));
+  }
+
```
romseygeek: This shouldn't be necessary, we use interfaces everywhere else?

solr/core/src/java/org/apache/solr/parser/SolrQueryParserBase.java
alessandrobenedetti: definitely make sense!

lucene/core/src/java/org/apache/lucene/util/AttributeSource.java
alessandrobenedetti: I am pretty sure , when I was debugging that at the moment that is the only way to get the Payload attributes (they seem to be there just in the attr impl list.
But I can double check this!

lucene/analysis/common/src/java/org/apache/lucene/analysis/boost/DelimitedBoostTokenFilter.java
```diff
@@ -0,0 +1,63 @@
...
+      for (int i = 0; i < length; i++) {
+        if (buffer[i] == delimiter) {
+          float boost = Float.parseFloat(new String(buffer, i + 1, (length - (i + 1))));
+          boostAtt.setBoost(boost);
+          termAtt.setLength(i);
+          return true;
+        }
+      }
+      // we have not seen the delimiter
+      boostAtt.setBoost(1.0f);
```
dsmiley: Shouldn't be needed; leave the boost be -- defaults to 1.0 any way.

lucene/core/src/java/org/apache/lucene/util/QueryBuilder.java
```diff
@@ -509,33 +549,40 @@ protected Query analyzeGraphBoolean(String field, TokenStream source, BooleanCla
...
-            return createFieldQuery(ts, BooleanClause.Occur.MUST, field, getAutoGenerateMultiTermSynonymsPhraseQuery(), 0);
+            TokenStream sidePath = sidePathsIterator.next();
+            return createFieldQuery(sidePath, BooleanClause.Occur.MUST, field, getAutoGenerateMultiTermSynonymsPhraseQuery(), 0);
           }
         };
-        queryPos = newGraphSynonymQuery(queries);
+        positionalQuery = newGraphSynonymQuery(queries);
       } else {
-        Term[] terms = graph.getTerms(field, start);
+        List<AttributeSource> attributes = graph.getTerms(start);
```
dsmiley: I think I mentioned a List of AttributeSource is weird (I've never seen this) and it's heavyweight.  Why not a TokenStream or TermAndBoost[] ?

solr/core/src/java/org/apache/solr/schema/TextField.java
```diff
@@ -43,6 +43,7 @@
 public class TextField extends FieldType {
   protected boolean autoGeneratePhraseQueries;
   protected boolean enableGraphQueries;
+  protected boolean synonymBoostByPayload;
```
dsmiley: I thought we switched the approach from a payload to boost attribute?  Besides; it's not clear we need this toggle at all since the user could arrange for this behavior simply by having the new DelimitedBoost filter thing in the chain.

solr/core/src/test-files/solr/collection1/conf/schema12.xml
```diff
@@ -238,6 +227,18 @@
     </analyzer>
   </fieldType>
 
+  <fieldType name="text_pick_best_boost_by_payload" class="solr.TextField" positionIncrementGap="100" synonymQueryStyle="pick_best" autoGeneratePhraseQueries="true">
```
dsmiley: You can remove "payload" everywhere from this PR now; no?

lucene/core/src/java/org/apache/lucene/util/QueryBuilder.java
```diff
@@ -450,9 +485,13 @@ protected Query analyzePhrase(String field, TokenStream stream, int slop) throws
         position += 1;
       }
       builder.add(new Term(field, termAtt.getBytesRef()), position);
+      phraseBoost = boostAtt.getBoost();
```
romseygeek: I think this isn't quite right, because we need to combine boosts together somehow; currently your phrase boost is just the boost of the last term in the phrase.

lucene/core/src/java/org/apache/lucene/util/QueryBuilder.java
romseygeek: This is what GraphTokenStreamFiniteStrings returns currently, for multiple tokens at the same position.  Maybe `TermAndBoost[]` would make more sense though.

alessandrobenedetti:
hi @romseygeek , @dsmiley ,
first of all, thank you again for your patience and very useful insights.
I have incorporated Alan's changes and cleaned everything up.

My un-resolved questions:
- boostAttribute doesn’t use BytesRef but directly float, is it a concern? We are expected to use it at query time, so we could actually see a query time minimal benefit in not encoding/decoding?
- Alan expressed concerns over SpanBoostQuery, mentioning they are sort of broken, what should we do in that regard? right now the create span query seems to work as expected with boosted synonyms(see the related test), I suspect if SpanBoostQuery are broken , they should get resolved in another ticket?
- from an original comment in the test code org.apache.solr.search.TestSolrQueryParser#testSynonymQueryStyle:
"confirm autoGeneratePhraseQueries always builds OR queries"
I changed that, was there any reason for that behaviour?

solr/core/src/java/org/apache/solr/schema/TextField.java
alessandrobenedetti: agreed and fixed!

lucene/analysis/common/src/java/org/apache/lucene/analysis/boost/DelimitedBoostTokenFilter.java
alessandrobenedetti: Fixed in the next coming commit

solr/core/src/test-files/solr/collection1/conf/schema12.xml
alessandrobenedetti: Fixed in the next coming commit!

lucene/core/src/java/org/apache/lucene/util/QueryBuilder.java
alessandrobenedetti: a tentative change is coming in the next commit, I added also few tests to cover that else coding branch

lucene/core/src/java/org/apache/lucene/util/QueryBuilder.java
alessandrobenedetti: I implemented a simple multiplicative boost.
It's back compatible with the designed use case (multi term synonym -> single concept -> single boost -> e.g. panthera onca => jaguar|0.95, big cat|0.85, black panther|0.65))

But it's also compatible in not synonym cases, if the user needs a boost per token in phrase and span queries.
It's in the upcoming commit, let me know if you believe something different is necessary

alessandrobenedetti:
I have applied the changes to solve the feedback points and consequentially added additional tests to cover some missing scenario.
We should be almost ready to go :)

lucene/analysis/common/src/java/org/apache/lucene/analysis/boost/DelimitedBoostTokenFilter.java
```diff
@@ -0,0 +1,61 @@
...
+      for (int i = 0; i < length; i++) {
+        if (buffer[i] == delimiter) {
+          float boost = Float.parseFloat(new String(buffer, i + 1, (length - (i + 1))));
+          boostAtt.setBoost(boost);
+          termAtt.setLength(i);
+          return true;
+        }
+      }
+      return true;
+    } else return false;
```
dsmiley: I know this is a minor matter of taste but please but brackets on the false side of the else with the code on its own line.  This is for consistency with our defacto code style in the project.

lucene/analysis/common/src/java/org/apache/lucene/analysis/boost/package-info.java
```diff
@@ -0,0 +1,21 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Provides various convenience classes for creating boosts on Tokens.
+ */
+package org.apache.lucene.analysis.boost;
```
dsmiley: While I can see why you chose a new "boost" sub-package because the payload based filter from which you drew inspiration was in a "payload" sub-package, I lean towards the "miscellaneous" package.  Note that DelimitedTermFrequencyTokenFilter is in "miscellaneous" too.  WDYT @romseygeek ?  Or maybe we need a new "delimited" sub-package for all these to go; I dunno.

lucene/core/src/java/org/apache/lucene/util/graph/GraphTokenStreamFiniteStrings.java
```diff
@@ -124,6 +126,15 @@ public boolean hasSidePath(int state) {
         .toArray(Term[]::new);
   }
 
+  /**
+   * Returns the list of terms that start at the provided state
+   */
+  public QueryBuilder.TermAndBoost[] getTermsAndBoosts(String field, int state) {
```
dsmiley: Given that this class, GraphTokenStreamFiniteStrings deals with List<AttributeSource> (something I did not know when I made a previous comment), and also that TermAndBoost is an inner class to QueryBuilder, I think it's better to put this back into QueryBuilder.  I still think `List<AttributeSource>` is weird and heavyweight but you didn't add it.

lucene/analysis/common/src/java/org/apache/lucene/analysis/boost/package-info.java
romseygeek: I like the `boost` package - I'm already thinking about a `TypeToBoostTokenFilter` that would automatically boost tokens marked with a `SYNONYM` type for example, and there are probably other boosting filters we can come up with, so a package to collect them all makes sense to me.  I prefer to group packages by functionality rather than implementation.

lucene/core/src/java/org/apache/lucene/util/graph/GraphTokenStreamFiniteStrings.java
romseygeek: Yes, let's go back to `AttributeSource` - sorry for the back and forth on this @alessandrobenedetti

lucene/core/src/java/org/apache/lucene/util/QueryBuilder.java
```diff
@@ -63,6 +66,25 @@
   protected boolean enableGraphQueries = true;
   protected boolean autoGenerateMultiTermSynonymsPhraseQuery = false;
 
+  /**
+   * Wraps a term and boost
+   */
+  public static class TermAndBoost {
+    private static final float DEFAULT_BOOST = 1.0f;
```
romseygeek: I think this should probably be on `BoostAttribute` rather than here?

lucene/analysis/common/src/java/org/apache/lucene/analysis/boost/package-info.java
alessandrobenedetti: So let's keep boost package then? no strong opinion here my side

lucene/core/src/java/org/apache/lucene/util/graph/GraphTokenStreamFiniteStrings.java
alessandrobenedetti: no worries at all, done in the upcoming commit!

lucene/core/src/java/org/apache/lucene/util/QueryBuilder.java
alessandrobenedetti: I agree, coming in the upcoming commit

lucene/analysis/common/src/java/org/apache/lucene/analysis/boost/DelimitedBoostTokenFilter.java
alessandrobenedetti: coming in the next commit, can you check it? I took it from the delimitedPayload, I guess code style is somewhat inconsistent across the project (I verified that multiple times in the past)

alessandrobenedetti:
Latest comments have been addressed, let me know if there's anything else needed here :)

lucene/core/src/java/org/apache/lucene/util/graph/GraphTokenStreamFiniteStrings.java
dsmiley: Can't you remove this now?

alessandrobenedetti: just forgot, it's done now

alessandrobenedetti:
So the code should be ok now, should we think where and how to properly document it ?
I will definitely write a blog post on that (that we can later link) but I guess we should think to the official documentation part now

dsmiley:
@romseygeek can you approve the PR if your comments are addressed?  Looks good from my end.  I really appreciate your input on use of BoostAttribute -- I didn't know about that one!

RE documentation: Yeah, I'm glad you are thinking of this; so many contributors (and sometimes us committers) don't.  I think there's a new TokenFilter you added here to document.  And then somewhere in synonym docs, some information suggesting how this can be used together.  I'd prefer you keep the ref guide changes minimally sufficient and leave the more wordy walk-through to a blog post and other mediums.  I'm looking forward to your blog post :-)

dsmiley:
Personally I'm fine with backporting this although QueryBuilder is not labelled as experimental so I wonder if we are "allowed" to change the API like we did in a minor release?  It's debatable.  Maybe it should be labelled experimental now.

Please merge @romseygeek as I'll be on a vacation shortly.  Otherwise I could get to it the last week of this month.

alessandrobenedetti:
Documentation has been provided, I had difficulties because of the preCommit (it seems I can't reproduce locally what I get via Github).
Anyway, let me know if you need any additional input my side!
I'll write a blog post to explain in details the contribution and I will share it as usual :)

alessandrobenedetti:
As I was writing the blog I thought that the Lucene changes would automatically bring the same feature to Elasticsearch, but then I kinda ended up in an another analysis common package in Elasticsearch that is a duplicated of the Lucene one (but still it imports the Lucene stuff).

https://github.com/elastic/elasticsearch/blob/24e1858a70bd255ebc210415acaac1bfb40340d3/modules/analysis-common/src/main/java/org/elasticsearch/analysis/common/CommonAnalysisPlugin.java

Am I correct to say that this feature won't appear automatically in Lucene unless we create a DelimitedBoostTokenFilterFactory there as well?

alessandrobenedetti:
Anything else I can do my side to close this pull request?

romseygeek:
Sorry, have been away and/or busy - I'm running tests now and will merge once everything passes.  Thanks for all your work @alessandrobenedetti!

PR #361, opened 2018-04-25
 SOLR-12271: Fix for analytics component reading negative values from double and float fields

HoustonPutman:


ctargett:
https://issues.apache.org/jira/browse/SOLR-12271 was released in Solr 7.4, so this PR can be closed.

PR #364, opened 2018-04-27
Branch 6x  . clzz.cast(object),Lost part of the data in a field of object。

dingziyang:
### Env: 
- solr-6.5.1
- jdk 1.8

### Question : 
- I got the query result from the solr server and tried to assign it to my java bean. After debugging, I found the problem code：```clzz cast(object)```,the object invoked a class named "Article",and has a member variables named “content”，it should got about 2000 bytes but only got about 500 bytes.
- Is a bug with solr, or  where am I doing wrong? ，Please help me,thanks.
### Code:
``` java
System.out.println(object); // debug in here，successfully got solr's return data,everything is fine.
list.add(clzz.cast(object)); 
System.out.println(object); // debug in here，clzz.cast(object),Lost part of the data in a field of object。
```


tflobbe:
This looks like it should be a question to the users list: http://lucene.apache.org/solr/community.html#mailing-lists-irc

PR #365, opened 2018-04-27
[SOLR-12243] span query generalization + query parser tests

alessandrobenedetti:
This is an extension of Elizabeth Haubert patch, the addition is : 

- spanNearQuery generalized to SpanQuery as it may happen to have SpanOrQuery
- added a test that verifies the correct query is (built) parsed .

It is a draft, additional tests and review are welcome, I have not investigated further but it seems to solve the bug


janhoy:
Merged manually

PR #368, opened 2018-05-02
[SOLR-12304] More Like This component interesting term fix +tests

alessandrobenedetti:
This Pull Request cover the bug fix and tests for the Interesting Term parameter of the More Like This component.


ctargett:
It looks like this PR was superseded by a patch in SOLR-12304; that is resolved, so closing this.

PR #372, opened 2018-05-07
LUCENE-8267: removed references to memory codecs.

dweiss:


solr/core/src/test/org/apache/solr/uninverting/TestFieldCacheVsDocValues.java
```diff
@@ -619,6 +619,6 @@ private void assertEquals(long numOrds, TermsEnum expected, TermsEnum actual) th
   
   protected boolean codecAcceptsHugeBinaryValues(String field) {
```
s1monw: do we still need this method?

dweiss: Right, we don't. I actually removed larger conditional blocks of code after this method is gone.

ctargett:
https://issues.apache.org/jira/browse/LUCENE-8267 will be released with Lucene 8.0, so this can be closed. Commit: 85c00e77efdf53f30da6eaffd38c2b016a7805bc

PR #374, opened 2018-05-09
[SOLR-12334] Improve detection of recreated lockfiles

pgerber:
I've been running into issues with the detection of deleted and then
recreated files when using GlusterFS. Since, on most platforms, there
are more reliable ways to detect recreated files, I decided improved
the detection of such files.

Background:

As part of LUCENE-6508 [3] detection of recreated files was added.
In such a case, another instance may have obtained lock on the newly
created file. This isn't something that should happen on a properly
configured Solr instance but there is a good chance for this
happening when an index is shared by multiple instances that use a
different locking mechanism.

Implementation:

On platform that support it, fileKey() [1] is now used. On Unix-like
systems this key consists of the device ID and and inode. For locks
that keep the lock file open, this should ensure we detect recreation
since no two files can have the same device ID and inode even if the
last hard link has been removed already. Other locks, that don't keep
the file open, should still detect recreation at a low error rate.

Platforms without fileKey() support keep using the creation timestamp
or modification timestamp (subject to availability) [2].

[1]: https://docs.oracle.com/javase/8/docs/api/java/nio/file/attribute/BasicFileAttributes.html#fileKey--
[2]: https://docs.oracle.com/javase/8/docs/api/java/nio/file/attribute/BasicFileAttributes.html#creationTime--
[3]: https://issues.apache.org/jira/browse/LUCENE-6508

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/store/FileID.java
+++ b/lucene/core/src/java/org/apache/lucene/store/FileID.java
@@ -1,0 +1,68 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.lucene.store;
+
+import java.io.IOException;
+import java.nio.file.attribute.BasicFileAttributes;
+import java.nio.file.attribute.FileTime;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+/**
+ * <p>A (hopefully) unique ID for a file</p>
+ *
+ * <p>This creates an ID for a file which can be compared to other IDs in order to attempt to
+ * verify that both are the same file. This comparison is done on an best-effort basis and is
+ * useful to ensure that a file has not been removed and recreated.</p>
+ *
+ * <p>Implementation details:</p>
+ *
+ * <ul>
+ *   <li>On plattforms that support {@link BasicFileAttributes#fileKey} this will be used to verify that
+ *   two {@link FileID}s refer to same file. On Unix-like systems, the key consists of the <em>device id</em>
+ *   and <em>inode</em> allowing detection of recreated files at a very low error rate.</li>
+ *   <li>On all other platforms, namely those that don't support {@link BasicFileAttributes#fileKey}, the
+ *   creation time or modification time, if available, is used in an attempt to ensure two {@link FileID}s
+ *   refer to the same file .</li>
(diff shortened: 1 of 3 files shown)


rmuir:
FileKey was avoided intentionally for several reasons here: specifically because of problems on windows operating systems. In this case the root problem is GlusterFS, it has inconsistent metadata. I don't think lucene should work around that.

pgerber:
@rmuir, for Windows nothing changes, timestamps are still used. The "several reason", do they include anything outside Windows?

I've not thoroughly investigated the issue with Gluster since I figured there is a better way to implement the check on non-Windows operating systems. I'm also not sure the bug is in Gluster, that's just were I've observed it. What I've observed is that Solr reports a different timestamp than [`stat`](https://linux.die.net/man/1/stat). Oddly, the timestamp observed at first is identical to the one reported by `stat` but Solr later complain the timestamp changed. Usually by something like 200μs farther into the past. Yet, stat keeps reporting the initial timestamp. As I have said, I haven't looked into the incorrect timestamp closely but if I had to choose, I'd believe stat.

rmuir:
fileKey is implemented on windows, so its untrue that nothing would change.

in general, the problem is with the metadata on gluster. lockfactory gets picked on because its the first thing you will hit (that's its job!), but hacking around the problem for gluster won't really help you. Its problems go much deeper, e.g. rename() isn't atomic. Its not appropriate for lucene indexes.

pgerber:
> fileKey is implemented on windows, so its untrue that nothing would change.

Looking at the [OpenJDK 10 source code](http://hg.openjdk.java.net/jdk10/jdk10/jdk/file/777356696811/src/java.base/windows/classes/sun/nio/fs/WindowsFileAttributes.java#l417), keyFile returns null invariably on Windows. Unless I'm wrong about it returning null on Windows, there should indeed not be any change. (I don't have Windows, so I couldn't actually test it.)

> in general, the problem is with the metadata on gluster. lockfactory gets picked on because its the first thing you will hit (that's its job!), but hacking around the problem for gluster won't really help you. Its problems go much deeper, e.g. rename() isn't atomic. Its not appropriate for lucene indexes.

So I've noticed and we'll probably move away from Gluster before long. However, I'd still like to see this change merged, independent of any issues with Gluster. To me, it would appear to be more sensible to use a natively by Java supported way for uniquely identifying files on platforms that support it as opposed to a hack using file creation/modification timestamps.

rmuir:
right it has FileKey but its not used for keyFile, just locking and stuff :)

I don't think its more useful to use an unreliable method, as opposed to using simple filesystem metadata that is supported everywhere. Your problem is really not the method we use, its that gluster filesystem metadata is completely broken and inconsistent across replicas. file keys (when they work at all) are still imperfect (inode numbers can be recycled) and harder to reason about. It would add additional paths to the code to worry about as we still need to fall back to the current code for windows anyway, I really see it as just trying to shove glusterfs's broken design under the rug.

rmuir:
FWIW i would maybe support a change that calls Objects.html#equals(java.lang.Object, java.lang.Object) on file keys *in addition* to the current ctime check.

More checks are better, but i don't want to weaken or remove any existing checks. File locking is such a sad state that we need all the paranoia we can get. This isn't gonna help your gluster case, but there is no helping that case anyway: its good that lock factory fails because file metadata is indeed changing behind the scenes.

pgerber:
I made the pull request since I figured it would make sense to use fileKey upstream too, independent of filesystem. However, I have to agree that it would make more sense to use both, timestamp and key. Also, I can see how the added complexity and different code path for different platforms isn't desirable. I'll update the pull request to have, both, key and timestamp checked.

pgerber:
Closing, I don't have any intention to still do this.

PR #375, opened 2018-05-16
LUCENE-8287: Ensure that empty regex completion queries always return no results.

jtibshirani:
This ensures consistency with the `PrefixCompletionQuery`, which returns no results for an
empty regex. It also fixes the exception described in the ticket when an empty regex is provided to a context query, as the empty prefix no longer matches any suggestion terms.

### DIFF
--- a/lucene/suggest/src/java/org/apache/lucene/search/suggest/document/ContextQuery.java
+++ b/lucene/suggest/src/java/org/apache/lucene/search/suggest/document/ContextQuery.java
@@ -167,11 +167,19 @@ public class ContextQuery extends CompletionQuery {
   @Override
   public Weight createWeight(IndexSearcher searcher, ScoreMode scoreMode, float boost) throws IOException {
     final CompletionWeight innerWeight = ((CompletionWeight) innerQuery.createWeight(searcher, scoreMode, boost));
+    final Automaton innerAutomaton = innerWeight.getAutomaton();
+
+    // If the inner automaton matches nothing, then we return an empty weight to avoid
+    // traversing all contexts during scoring.
+    if (innerAutomaton.getNumStates() == 0) {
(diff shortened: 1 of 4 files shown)


jtibshirani:
@jimczi I tried a few different approaches (including updating `RegexCompletionQuery#rewrite`), and this one turned out to be cleanest + most consistent. I'm happy to give more details about the alternatives if it'd be useful while reviewing.

jtibshirani:
I'm sorry for the confusion -- I didn't mean complete parity with `RegexpQuery`, as they are different queries with different behaviors, but rather consistency how they treat an empty regex. After thinking more I now see your point, it makes a lot more sense to describe it as consistency with `PrefixCompletionQuery`!

I've made the following changes:
- Replaced references to 'parity with `RegexpQuery`' to 'consistency with `PrefixCompletionQuery`, as this is more accurate.
- Added an optimization to `ContextQuery#createWeight` that returns an empty automaton if the inner weight is empty. Note that this optimization only helps with `RegexCompletionQuery`. When `PrefixCompletionQuery` contains an empty term, its automaton is actually `*` -- maybe we could look into adding an optimization there in a follow-up.

rmuir:
{quote}
When PrefixCompletionQuery has an empty term, its automaton is actually * -- maybe we could look into adding a fix/ optimization there in a follow-up.
{quote}

Regexp query should handle that case fine though. It uses Operations.isEmpty (see the CompiledAutomaton class) to detect that the automation accepts no strings. The logic here is different and a subset of that, and I think thats why there might be such cases.

jtibshirani:
Thank you @rmuir for the pointer, and I'm sorry for the slow reply! I just opened https://github.com/apache/lucene-solr/pull/383 to switch to `Operations.isEmpty` -- @jimczi feel free to take a look if you think this would be a good follow-up.

PR #376, opened 2018-05-16
SOLR-12361: _childDocuments as a map in SolrInputDocument

moshebla:


moshebla:
links to wrong ticket

PR #382, opened 2018-05-23
WIP: SOLR-12361

moshebla:
pass tests but the TestChildDocTransformer.
We have to think of a how we want to change the transformer, which should probably be discussed in another issue.
Currently it does not add the _childDocuments_ field to fl, causing the childDocuments to be ommited from the response.

### DIFF
--- a/solr/core/src/java/org/apache/solr/response/JSONResponseWriter.java
+++ b/solr/core/src/java/org/apache/solr/response/JSONResponseWriter.java
@@ -376,20 +376,6 @@ class JSONWriter extends TextResponseWriter {
         writeVal(fname, val);
       }
     }
-
-    if(doc.hasChildDocuments()) {
-      if(first == false) {
-        writeMapSeparator();
-        indent();
-      }
-      writeKey("_childDocuments_", true);
-      writeArrayOpener(doc.getChildDocumentCount());
-      List<SolrDocument> childDocs = doc.getChildDocuments();
-      for(int i=0; i<childDocs.size(); i++) {
-        writeSolrDocument(null, childDocs.get(i), null, i);
-      }
-      writeArrayCloser();
-    }
     
     decLevel();
     writeMapCloser();
--- a/solr/solrj/src/java/org/apache/solr/common/SolrDocument.java
+++ b/solr/solrj/src/java/org/apache/solr/common/SolrDocument.java
@@ -16,14 +16,17 @@
  */
 package org.apache.solr.common;
 
+import java.util.AbstractMap;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.HashMap;
 import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Stream;
 
 import org.apache.solr.common.util.NamedList;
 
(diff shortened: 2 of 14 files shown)


solr/solrj/src/java/org/apache/solr/common/SolrDocument.java
```diff
@@ -388,20 +386,47 @@ public void addChildDocuments(Collection<SolrDocument> children) {
...
+     return childDocs;
+   }
+
    /** Returns the list of child documents, or null if none. */
    @Override
    public List<SolrDocument> getChildDocuments() {
-     return _childDocuments;
+     List<SolrDocument> childDocs = new ArrayList<>();
+     Stream<AbstractMap.SimpleEntry<String, SolrDocument>> fields = _fields.entrySet().stream()
+         .filter(value -> value.getValue() instanceof SolrInputDocument)
```
dsmiley: Or the value might be a List of SolrInputDocument, so we have to check for that too; right?

solr/solrj/src/java/org/apache/solr/common/SolrDocument.java
```diff
@@ -388,20 +386,47 @@ public void addChildDocuments(Collection<SolrDocument> children) {
...
+         .map(value -> new AbstractMap.SimpleEntry<>(value.getKey(), (SolrDocument) value.getValue()));
+     fields.forEach(e -> childDocs.add(e.getValue()));
+     return childDocs.size() > 0 ? childDocs: null;
    }
    
    @Override
    public boolean hasChildDocuments() {
```
dsmiley: We'd probably deprecate these if we go with this overall approach.  It's too much internal cost to call this method.

solr/solrj/src/java/org/apache/solr/common/SolrDocument.java
moshebla: yeah you're right, I'll add another commit to fix this

michaelbraun: Instead of creating an outer List<SolrDocument> childDocs = new ArrayList<>() up front, can this look like List<...> fields = ......collect(Collectors.toList()), with the size check at the end? 
Also would be nice if this API could be changed to always return non-null, and if it's empty, it's just empty.

solr/core/src/java/org/apache/solr/response/JSONResponseWriter.java
```diff
@@ -376,20 +376,6 @@ public void writeSolrDocument(String name, SolrDocument doc, ReturnFields return
         writeVal(fname, val);
       }
     }
-
```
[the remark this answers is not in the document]
moshebla: Removed in GeoJSONResponseWriter too.

dsmiley:
As discussed in the JIRA issue, perhaps you should update the patch to try and leave the existing anonymous children in place and merely add semantic values via field values?  (or start a new PR if you wish).  Maybe modify AddBlockUpdateTest to add a test that works with SolrInputDocument directly to add child docs as field values, and index it, and test?  You can use EmbeddedSolrServer wrapping h.getCore()) to work with the test harness using SolrJ (SolrInputDocument), thus avoiding the XML/JSON incoming which we have not yet added semantic key support to in formats.

moshebla:
new pull request #385

PR #383, opened 2018-05-25
LUCENE-8287: In ContextQuery, use a more comprehensive check for an empty prefix automaton

jtibshirani:
Follow-up to https://github.com/apache/lucene-solr/pull/375.


madrob:
Going through old patches trying to clean some up before the end of the year. @jtibshirani this patch looks good to me, would you like to commit it?

jtibshirani:
Thanks @madrob for pinging on this, I'll commit in the next few days.

PR #384, opened 2018-05-25
LUCENE-8332 move CompletionTokenStream to ConcatenateGraphFilter

dsmiley:
See JIRA issue for overview comments.

### DIFF
--- a/lucene/analysis/common/src/java/org/apache/lucene/analysis/miscellaneous/ConcatenateGraphFilter.java
+++ b/lucene/analysis/common/src/java/org/apache/lucene/analysis/miscellaneous/ConcatenateGraphFilter.java
@@ -18,11 +18,9 @@ package org.apache.lucene.analysis.miscellaneous;
 
 import java.io.IOException;
 
-import org.apache.lucene.analysis.TokenFilter;
 import org.apache.lucene.analysis.TokenStream;
 import org.apache.lucene.analysis.TokenStreamToAutomaton;
 import org.apache.lucene.analysis.tokenattributes.CharTermAttribute;
-import org.apache.lucene.analysis.tokenattributes.PayloadAttribute;
 import org.apache.lucene.analysis.tokenattributes.PositionIncrementAttribute;
 import org.apache.lucene.analysis.tokenattributes.TermToBytesRefAttribute;
 import org.apache.lucene.util.AttributeImpl;
@@ -47,7 +45,7 @@ import org.apache.lucene.util.fst.Util;
  *
  * @lucene.experimental
  */
-public final class ConcatenateGraphFilter extends TokenFilter {
+public final class ConcatenateGraphFilter extends TokenStream {
 
   /*
    * Token stream which converts a provided token stream to an automaton.
@@ -61,25 +59,22 @@ public final class ConcatenateGraphFilter extends TokenFilter {
    * Represents the separation between tokens, if
    * <code>preserveSep</code> is <code>true</code>.
    */
-  public final static char SEP_CHAR = '\u001F';
+  public final static int SEP_LABEL = TokenStreamToAutomaton.POS_SEP;
   public final static int DEFAULT_MAX_GRAPH_EXPANSIONS = Operations.DEFAULT_MAX_DETERMINIZED_STATES;
   public final static boolean DEFAULT_PRESERVE_SEP = true;
   public final static boolean DEFAULT_PRESERVE_POSITION_INCREMENTS = true;
 
-  private final PayloadAttribute payloadAttr = addAttribute(PayloadAttribute.class);
   private final BytesRefBuilderTermAttribute bytesAtt = addAttribute(BytesRefBuilderTermAttribute.class);
   private final PositionIncrementAttribute posIncrAtt = addAttribute(PositionIncrementAttribute.class);
 
-  //nocommit add getters for these instead?  Accessor: org.apache.lucene.search.suggest.document.ContextSuggestField.wrapTokenStream
-  public final boolean preserveSep;
-  public final boolean preservePositionIncrements;
-  public final int maxGraphExpansions;
-  public TokenStream getInput() { return super.input; } // nocommit remove? see ContextSuggestField.wrapTokenStream
+  private final TokenStream inputTokenStream;
+  private final boolean preserveSep;
+  private final boolean preservePositionIncrements;
+  private final int maxGraphExpansions;
 
   private LimitedFiniteStringsIterator finiteStrings;
-  private BytesRef payload;
   private CharTermAttribute charTermAttribute;
-  private State endState;
+  private boolean wasReset = false;
 
   /**
    * Creates a token stream to convert <code>input</code> to a token stream
@@ -96,8 +91,11 @@ public final class ConcatenateGraphFilter extends TokenFilter {
    * of accepted strings by its token stream graph.
    *
    * @param inputTokenStream The input/incoming TokenStream
-   * @param preserveSep
-   * @param preservePositionIncrements Whether to generate holes in the automaton for missing positions
+   * @param preserveSep Whether {@link #SEP_LABEL} should separate the input tokens in the concatenated token
+   * @param preservePositionIncrements Whether to add an empty token for missing positions.
+   *                                   The effect is a consecutive {@link #SEP_LABEL}.
+   *                                   When false, it's as if there were no missing positions
+   *                                     (we pretend the surrounding tokens were adjacent).
    * @param maxGraphExpansions If the tokenStream graph has more than this many possible paths through, then we'll throw
    *                           {@link TooComplexToDeterminizeException} to preserve the stability and memory of the
    *                           machine.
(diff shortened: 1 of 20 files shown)


lucene/analysis/common/src/java/org/apache/lucene/analysis/miscellaneous/ConcatenateGraphFilter.java
```diff
@@ -31,80 +33,106 @@
...
- * a completion's payload (see {@link CompletionTokenStream#setPayload(org.apache.lucene.util.BytesRef)})
+ * Concatenates/Joins every incoming token with a separator into one output token for every path through the
+ * token stream (which is a graph).  In simple cases this yields one token, but in the presence of any tokens with
+ * a zero positionIncrmeent (e.g. synonyms) it will be more.  This filter uses the token bytes, position increment,
+ * and position length of the incoming stream.  Other attributes are not used or manipulated.
  *
  * @lucene.experimental
  */
-public final class CompletionTokenStream extends TokenStream {
+public final class ConcatenateGraphFilter extends TokenFilter {
```
jimczi: I still don't understand why you need to use a TokenFilter instead of a TokenStream. You need to override all the functions anyway so I don't see the benefit.

lucene/analysis/common/src/java/org/apache/lucene/analysis/miscellaneous/ConcatenateGraphFilter.java
```diff
@@ -31,80 +33,106 @@
...
+   * {@link org.apache.lucene.analysis.tokenattributes.TermToBytesRefAttribute} attribute
+   * The token stream uses a {@link org.apache.lucene.analysis.tokenattributes.PayloadAttribute} to store
+   * a completion's payload (see {@link ConcatenateGraphFilter#setPayload(org.apache.lucene.util.BytesRef)})
+   */
+
+  /**
+   * Represents the separation between tokens, if
+   * <code>preserveSep</code> is <code>true</code>.
+   */
+  public final static char SEP_CHAR = '\u001F';
```
jimczi: I think it's better to keep the SEP_LABEL as an int. It needs to be equal to TokenStreamToAutomaton.POS_SEP anyway so you can maybe used this value directly ? PAYLOAD_SEP has the same requirement so we should also set it to POS_SEP.

lucene/analysis/common/src/java/org/apache/lucene/analysis/miscellaneous/ConcatenateGraphFilter.java
```diff
@@ -119,27 +147,19 @@ public boolean incrementToken() throws IOException {
...
+    restoreState(endState);
   }
 
+  //nocommit move method to before incrementToken
   @Override
   public void reset() throws IOException {
-    super.reset();
-    if (hasAttribute(CharTermAttribute.class)) {
-      // we only create this if we really need it to safe the UTF-8 to UTF-16 conversion
-      charTermAttribute = getAttribute(CharTermAttribute.class);
```
jimczi: The automaton should be created lazily on the first call to incrementToken like described in the comment above ;).

lucene/analysis/common/src/java/org/apache/lucene/analysis/miscellaneous/ConcatenateGraphFilterFactory.java
```diff
@@ -0,0 +1,61 @@
+/*
+ This software was produced for the U. S. Government
+ under Contract No. W15P7T-11-C-F600, and is
+ subject to the Rights in Noncommercial Computer Software
+ and Noncommercial Computer Software Documentation
+ Clause 252.227-7014 (JUN 1995)
+
```
jimczi: Is this really needed ? I understand that you have some constraint for the development of the main feature but this doesn't seem related so maybe use the usual license ?

lucene/suggest/src/java/org/apache/lucene/search/suggest/document/NRTSuggesterBuilder.java
```diff
@@ -43,6 +44,9 @@
    * in the output
    */
   public static final int PAYLOAD_SEP = '\u001F';
```
jimczi: You can set it to TokenStreamToAutomaton.POS_SEP directly and add a test to ensure that PAYLOAD_SEP and SEP_LABEL are equals to `0x001F` for BWC.

lucene/suggest/src/java/org/apache/lucene/search/suggest/document/NRTSuggesterBuilder.java
```diff
@@ -43,6 +44,9 @@
...
+  static {
+    assert PAYLOAD_SEP == ConcatenateGraphFilter.SEP_CHAR;
```
jimczi: This can ensured in a dedicated test ?

lucene/analysis/common/src/java/org/apache/lucene/analysis/miscellaneous/ConcatenateGraphFilterFactory.java
dsmiley: Good catch; no it's not needed.

lucene/suggest/src/java/org/apache/lucene/search/suggest/document/NRTSuggesterBuilder.java
dsmiley: I think it's useful to read the code and both see what's it's value is and that it's equal to something else without having to navigate or use IDE functions.  But this is no big deal to me; I can put back.  Presumably if these values were not the same then some a test would fail but I'm not sure.

lucene/analysis/common/src/java/org/apache/lucene/analysis/miscellaneous/ConcatenateGraphFilter.java
dsmiley: I think this is the difference between an implementation detail and something more meaningful/intuitive to the user.  I'm trying to cater to the user here.  Maybe we could have both?

lucene/analysis/common/src/java/org/apache/lucene/analysis/miscellaneous/ConcatenateGraphFilter.java
dsmiley: I understand what you are asking me to do (to put back) but don't understand _why_.  What comment above do you refer to?  Where your feedback is?  That comment doesn't reference that; or at least it's very non-obvious to me if there's a relationship (and we should add comments explaining).

lucene/analysis/common/src/java/org/apache/lucene/analysis/miscellaneous/ConcatenateGraphFilter.java
dsmiley: Perhaps it _never_ matters; yet TokenFilter exists nonetheless.  TokenFilter is more consistent with public/visible analysis components, and it makes it easier to pass-through the calls to the delegate/input.  

Let me put it another way... shouldn't FingerprintFilter & ConcatenateGraphFilter do the same, whatever that is?  FingerprintFilter is already public/known so I'm trying to introduce this in a way that is consistent.  That CompletionTokenStream, kinda hidden, is currently not a TokenFilter isn't relevant.

lucene/analysis/common/src/java/org/apache/lucene/analysis/miscellaneous/ConcatenateGraphFilter.java
jimczi: I am referring to this comment: 
````
//nocommit move method to before incrementToken
````
but maybe I misunderstood what you meant by move. Let me ask you this differently, why do you need to build the automaton on reset ? Unless there is a good reason to do it we shouldn't change the behavior here which is to build the automaton on the first call to incrementToken.

File: lucene/analysis/common/src/java/org/apache/lucene/analysis/miscellaneous/ConcatenateGraphFilter.java (hunk 2 above)
Comment (jimczi) [reply lucene/analysis/common/src/java/org/apache/lucene/analysis/miscellaneous/ConcatenateGraphFilter.java]: This value should only be used internally to encode a separator in the automaton or in the binary terms that we index. Which user are you referring to ?

File: lucene/analysis/common/src/java/org/apache/lucene/analysis/miscellaneous/ConcatenateGraphFilter.java (hunk 3 above)
Comment (dsmiley) [reply lucene/analysis/common/src/java/org/apache/lucene/analysis/miscellaneous/ConcatenateGraphFilter.java]: Oh *that* comment.  By that I mean move the _method_ to follow a natural sequence.  I like classes to have methods that are organized sensibly.  When one uses a TokenStream, reset() should come before incrementToken() !  -- then end(), then close(), and any utility methods should come after callers.

I moved the initialization to reset simply because it felt natural this way rather than incrementToken where it's lazy.  That's all.  That said... I hinted at a TestRandomChains failure about lifecycle needing to throw an IllegalStateException or something like that (pertaining to missing close()), and I think making this back to lazy will "fix" that.

File: lucene/analysis/common/src/java/org/apache/lucene/analysis/miscellaneous/ConcatenateGraphFilter.java (hunk 2 above)
Comment (dsmiley) [reply lucene/analysis/common/src/java/org/apache/lucene/analysis/miscellaneous/ConcatenateGraphFilter.java]: One user is the SolrTextTagger which needs to know the character separator between concatenated terms.  https://github.com/OpenSextant/SolrTextTagger/blob/master/src/main/java/org/opensextant/solrtexttagger/TermPrefixCursor.java#L45   (being added to Solr in SOLR-12376).  Hmmm, though it is a byte there, which is to your point.  Maybe I should change to a byte and add a little comment about this?

File: lucene/analysis/common/src/java/org/apache/lucene/analysis/miscellaneous/ConcatenateGraphFilter.java (hunk 1 above)
Comment (jimczi) [reply lucene/analysis/common/src/java/org/apache/lucene/analysis/miscellaneous/ConcatenateGraphFilter.java]: Converting this token stream into a token filter is trappy. It requires to rewrite the logic completely. For instance you removed the `close` implementation, this means that it is called twice on the input stream since it is already closed here https://github.com/apache/lucene-solr/pull/384/files#diff-04d4c6352889dbffd0eb4d1e6ecd6097R197. You also added a call to super() which was avoided clearly in the original impl:
````
// Don't call the super(input) ctor - this is a true delegate and has a new attribute source since we consume
    // the input stream entirely in the first call to incrementToken
````
My idea was to move this token stream if only minor changes are required. If it needs a complete rewrite then we should probably reconsider.

Comment (jimczi) [reply lucene/analysis/common/src/java/org/apache/lucene/analysis/miscellaneous/ConcatenateGraphFilter.java]: +1

**Comment by dsmiley** [comment]
There's a test failure in TestFactories.test which calls BaseTokenStreamTestCase.checkResetException.

`ant test -Dtestcase=TestFactories -Dtests.seed=543C55AE2375085C`
```
   [junit4]    > Throwable #1: java.lang.AssertionError: didn't get expected exception when close() not called
   [junit4]    >        at __randomizedtesting.SeedInfo.seed([543C55AE2375085C:DC686A748D8965A4]:0)
   [junit4]    >        at org.apache.lucene.analysis.BaseTokenStreamTestCase.checkResetException(BaseTokenStreamTestCase.java:453)
   [junit4]    >        at org.apache.lucene.analysis.BaseTokenStreamTestCase.checkRandomData(BaseTokenStreamTestCase.java:546)
   [junit4]    >        at org.apache.lucene.analysis.core.TestFactories.doTestTokenFilter(TestFactories.java:118)
   [junit4]    >        at org.apache.lucene.analysis.core.TestFactories.test(TestFactories.java:68)
   [junit4]    >        at java.lang.Thread.run(Thread.java:748)
```
The issue is that if some faulty code forgets to call close(), an exception would be ideal to be thrown but for this filter it won't be because the inputTokenStream is closed properly internally correctly by toAutomaton.  Checks like this are kinda... I dunno, I don't like it.  IMO it's asking too much of a TokenFilter to insist that it's cranky when it's mistreated; maybe some components (like this one) are more resilient?  Maybe BaseTokenStreamTestCase (or somewhere else) should keep track of components that are resilient in this way so that it doesn't exercise these sort of tests?  If we really want to insist that every component be cranky when mistreated, then perhaps we could have toAutomaton _not_ close the inputTokenStream so that we can delegate that from CGF's close?

jimczi:
> If we really want to insist that every component be cranky when mistreated, then perhaps we could have toAutomaton not close the inputTokenStream so that we can delegate that from CGF's close?

I think it's the right thing to do and the change is minimal. Good catch !

PR #385, opened 2018-05-28
WIP: SOLR-12361

moshebla:


solr/core/src/java/org/apache/solr/update/processor/IgnoreLargeDocumentProcessorFactory.java
```diff
@@ -156,7 +160,17 @@ private static long fastEstimate(Map<Object, Object> map) {
       if (map.isEmpty()) return 0;
       long size = 0;
       for (Map.Entry<Object, Object> entry : map.entrySet()) {
-        size += primitiveEstimate(entry.getKey(), 0L) + primitiveEstimate(entry.getValue(), 0L);
+        size += primitiveEstimate(entry.getKey(), 0L);
+        Object value = entry.getValue();
+        if (value instanceof Map) {
+          size += fastEstimate(entry.getValue());
+        } else if (value instanceof Collection) {
+          for(Object item: ((Collection) value)) {
```
dsmiley: couldn't you call the fastEstimate(Collection) version instead of looping here?

solr/core/src/test/org/apache/solr/update/AddBlockUpdateTest.java
```diff
@@ -260,10 +266,124 @@ public void testExceptionThrown() throws Exception {
...
+
+  @Test
+  public void testSolrNestedFieldsList() throws Exception {
+    SolrInputDocument document1 = new SolrInputDocument() {
+      {
+        final String id = id();
+        addField("id", id);
+        addField("parent_s", "X");
+        addField("children",
+            new ArrayList<SolrInputDocument>()
```
dsmiley: Use sdoc(...) from SolrTestCaseJ4 instead of this trick

solr/core/src/test/org/apache/solr/update/AddBlockUpdateTest.java
```diff
@@ -398,7 +523,43 @@ public void testXML() throws IOException, XMLStreamException {
...
+    try (JavaBinCodec jbc = new JavaBinCodec()) {
+      jbc.marshal(topDocument, os);
+    }
+    byte[] buffer = os.toByteArray();
+    //now read the Object back
+    SolrInputDocument result;
+    try (JavaBinCodec jbc = new JavaBinCodec(); InputStream is = new ByteArrayInputStream(buffer)) {
+      result = (SolrInputDocument) jbc.unmarshal(is);
+    }
+    assertEquals(2 + childsNum, result.size());
```
dsmiley: This is a round-trip test.  Can we do assertEquals(topDocument, result)?

solr/core/src/test/org/apache/solr/update/AddBlockUpdateTest.java
```diff
@@ -521,6 +682,23 @@ private void attachSubDocs(Element parent, String relation, int typeValue) {
...
+    SolrQueryRequest coreReq = new LocalSolrQueryRequest(h.getCore(), new ModifiableSolrParams());
+    AddUpdateCommand updateCmd = new AddUpdateCommand(coreReq);
+    for (SolrInputDocument doc: docs) {
+      long version = getNewClock();
+      updateCmd.setVersion(Math.abs(version));
+      updateCmd.solrDoc = doc;
+      h.getCore().getUpdateHandler().addDoc(updateCmd);
+      updateCmd.clear();
+    }
+    assertU(commit());
```
dsmiley: randomly commit, thus produce different segments?

solr/solrj/src/java/org/apache/solr/common/SolrInputDocument.java
```diff
@@ -259,6 +260,17 @@ public void addChildDocuments(Collection<SolrInputDocument> children) {
...
+    for (SolrInputField field: values()) {
+      Object value = field.getFirstValue();
+      if (value instanceof SolrInputDocument) {
+        childDocs.put(field.getName(), field);
+      }
+    }
+    return childDocs.size() > 0 ? childDocs: null;
+  }
+
   /** Returns the list of child documents, or null if none. */
```
dsmiley: This method now needs some clear documentation to say it returns *anonymous* child documents.

solr/solrj/src/java/org/apache/solr/common/SolrDocument.java
```diff
@@ -373,6 +374,23 @@ public int size() {
     return _fields.values();
   }
 
+  public Map<String, Object> getChildDocumentsMap() {
```
dsmiley: This method is tempting; it's unsurprising you added it, but I'd rather we not have it.  It's not particularly cheap; raises questions of mutability, and it's really not necessary; only one caller (for indexing, recunwrap) needs this algorithm; and it need not build up some HashMap as we see here to do its job.

solr/solrj/src/java/org/apache/solr/common/SolrInputDocument.java
```diff
@@ -259,6 +260,17 @@ public void addChildDocuments(Collection<SolrInputDocument> children) {
     }
   }
 
+  public Map<String, SolrInputField> getChildDocumentsMap() {
```
dsmiley: again; we don't need this?

solr/core/src/java/org/apache/solr/update/AddUpdateCommand.java
```diff
@@ -175,69 +176,92 @@ public String getHashableId() {
...
-    return new Iterator<Document>() {
-      Iterator<SolrInputDocument> iter;
-
-      {
-        List<SolrInputDocument> all = flatten(solrDoc);
-
-        String idField = getHashableId();
-
-        boolean isVersion = version != 0;
+  public boolean isBlock() {
```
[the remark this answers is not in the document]
moshebla: perhaps it could be made a static, with the List of documents passed as a parameter, to avoid duplication of code.

solr/core/src/java/org/apache/solr/update/DirectUpdateHandler2.java
```diff
@@ -973,16 +976,43 @@ private void updateDocOrDocValues(AddUpdateCommand cmd, IndexWriter writer, Term
...
+      log.debug("updateDocuments({})", docs);
+      writer.updateDocuments(updateTerm, toDocumentsIter(docs, cmd.req.getSchema()));
     } else {
       Document luceneDocument = cmd.getLuceneDocument(false);
       log.debug("updateDocument({})", cmd);
       writer.updateDocument(updateTerm, luceneDocument);
     }
   }
 
+  private Iterable<Document> toDocumentsIter(Collection<SolrInputDocument> docs, IndexSchema schema) {
```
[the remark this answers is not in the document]
moshebla: Great thinking,
reduced so many lines of unnecessary code

solr/core/src/java/org/apache/solr/update/processor/IgnoreLargeDocumentProcessorFactory.java
moshebla: Oops, my bad

solr/core/src/java/org/apache/solr/update/DirectUpdateHandler2.java
```diff
@@ -417,7 +417,8 @@ private void addAndDelete(AddUpdateCommand cmd, List<UpdateLog.DBQ> deletesAfter
   }
 
   private Term getIdTerm(AddUpdateCommand cmd) {
```
dsmiley: I think it's trappy/dangerous to invoke getDocsList() which is an innocent looking getter that actually flattens the input and caches it rendering any possible further changes to the unflattend docs silently ignored.  Can you change this method to accept the List of flattened documents as its argument?  And change getDocsList to not cache the result and be named something like computeFlattenedDocs()

solr/test-framework/src/java/org/apache/solr/SolrTestCaseJ4.java
```diff
@@ -2181,16 +2182,38 @@ public boolean compareSolrInputDocument(Object expected, Object actual) {
     Iterator<String> iter1 = sdoc1.getFieldNames().iterator();
     Iterator<String> iter2 = sdoc2.getFieldNames().iterator();
 
-    if(iter1.hasNext()) {
+    while (iter1.hasNext()) {
```
dsmiley: ouch; good catch!

solr/core/src/test/org/apache/solr/update/AddBlockUpdateTest.java
```diff
@@ -693,7 +668,6 @@ private void indexSolrInputDocumentsDirectly(List<SolrInputDocument> docs) throw
       h.getCore().getUpdateHandler().addDoc(updateCmd);
       updateCmd.clear();
     }
-    assertU(commit());
```
dsmiley: this part was fine, especially for nested docs since we want there to be a higher likelihood of more segments since if there's a bug pertaining to not adding docs in a block, then more segments will help prove this out.

solr/core/src/java/org/apache/solr/update/AddUpdateCommand.java
```diff
@@ -250,6 +261,7 @@ private void recUnwrapRelations(List<SolrInputDocument> unwrappedDocs, SolrInput
     recUnwrapRelations(unwrappedDocs, currentDoc, false);
   }
 
+  /** Extract all anonymous child documents from parent. */
   private void recUnwrapp(List<SolrInputDocument> unwrappedDocs, SolrInputDocument currentDoc, boolean isRoot) {
```
dsmiley: Oh; I didn't know -- the docs then suggest to me this method ought to have a clearer name ;-)    `e.g. recUnwrapAnonymous`

solr/core/src/java/org/apache/solr/update/AddUpdateCommand.java
```diff
@@ -228,11 +225,25 @@ private void buildDocsList() {
     return unwrappedDocs;
   }
 
+  private static Collection<String> getChildDocumentsKeys(SolrInputDocument doc) {
```
dsmiley: But this is only called in one place; no?  I think it's better if recUnwrapRelations does this logic itself as it avoids the intermediate HashSet.

solr/core/src/java/org/apache/solr/update/AddUpdateCommand.java
```diff
@@ -184,14 +179,7 @@ public String getHashableId() {
...
-      buildDocsList();
-    }
-    return docsList;
-  }
-
-  private void buildDocsList() {
+  public List<SolrInputDocument> computeFlattenedDocs() {
     List<SolrInputDocument> all = flatten(solrDoc);
 
     String idField = getHashableId();
```
dsmiley: rename rootId?  "field" feels very wrong in this variable name.

solr/core/src/java/org/apache/solr/update/AddUpdateCommand.java
```diff
@@ -175,69 +172,83 @@ public String getHashableId() {
...
+          recUnwrapRelations(unwrappedDocs, ((SolrInputDocument) val));
+          continue;
+        }
+        Collection<SolrInputDocument> childrenList = ((Collection) val);
+        for (SolrInputDocument child : childrenList) {
+          recUnwrapRelations(unwrappedDocs, child);
+        }
+      }
+    }
+
```
dsmiley: maybe add a check that these field keyed relations have no anonymous children?  It's quick to check.  The reverse shouldn't be allowed either but I wouldn't bother enforcing that as it's a bit more work.

solr/core/src/java/org/apache/solr/update/DirectUpdateHandler2.java
dsmiley: Latest diff shows this logic calls cmd.computeFlattenedDocs() and we definitely don't want this method computing that!
Now that I look at the code in my IDE, I can appreciate that this is a bit of a tricky issue though, since neither callers of getIdTerm yet have the List<SolrInputDocument> to give to this method.  Hmmmm.   Maybe `updateDocument` should not take an updateTerm as an argument but should instead figure it out and return it?  Note that if isInPlaceUpdate, the updateTerm will always be the unique key field; you needn't check for a list of docs.

solr/core/src/java/org/apache/solr/update/DirectUpdateHandler2.java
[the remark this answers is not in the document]
dsmiley: I think `Iterables.transform` is a little leaner than indirectly getting there via FluentIterable.  I don't think FluentIterable is more concise here.

solr/core/src/test/org/apache/solr/update/AddBlockUpdateTest.java
```diff
@@ -260,10 +266,90 @@ public void testExceptionThrown() throws Exception {
...
+  public void testSolrNestedFieldsSingleVal() throws Exception {
+    SolrInputDocument document1 = sdoc("id", id(), parent, "X",
+        "child1_s", sdoc("id", id(), "child_s", "y"),
+        "child2_s", sdoc("id", id(), "child_s", "z"));
+
+    SolrInputDocument document2 = sdoc("id", id(), parent, "A",
+        "child1_s", sdoc("id", id(), "child_s", "b"),
+        "child2_s", sdoc("id", id(), "child_s", "c"));
+
+    List<SolrInputDocument> docs = new ArrayList<SolrInputDocument>() {
```
dsmiley: remember to use Arrays.asList

solr/core/src/java/org/apache/solr/update/AddUpdateCommand.java
```diff
@@ -175,69 +172,83 @@ public String getHashableId() {
...
       }
-    };
+      if(isVersion) sdoc.setField(CommonParams.VERSION_FIELD, version);
+      // TODO: if possible concurrent modification exception (if SolrInputDocument not cloned and is being forwarded to replicas)
+      // then we could add this field to the generated lucene document instead.
+    }
+    return all;
   }
 
   private List<SolrInputDocument> flatten(SolrInputDocument root) {
```
dsmiley: note that if cmd.isInplaceUpdate(), then there is no flattening to be done -- or at least, it's an error.

solr/core/src/java/org/apache/solr/update/AddUpdateCommand.java
```diff
@@ -146,17 +146,14 @@ public String getPrintableId() {
      return "(null)";
    }
 
-  /**
-   * @return String id to hash
-   */
-  public String getHashableId() {
+  public String getHashableId(SolrInputDocument doc) {
```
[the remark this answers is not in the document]
moshebla: I changed it for something else that was scraped.
I'll revert the changes

solr/core/src/java/org/apache/solr/update/AddUpdateCommand.java
moshebla: Yeah guess an exception should be thrown

solr/core/src/test/org/apache/solr/update/AddBlockUpdateTest.java
moshebla: Oops, my bad

solr/core/src/java/org/apache/solr/update/AddUpdateCommand.java
moshebla: That seems about right, and is exactly what happens in [DirectUpdateHandler2](https://github.com/moshebla/lucene-solr/blob/3b7d33f3cc7363fcb36f7f91ae7ca5f6bd4bee3a/solr/core/src/java/org/apache/solr/update/DirectUpdateHandler2.java#L963).

solr/core/src/java/org/apache/solr/update/AddUpdateCommand.java
moshebla: I just thought this through, and perhaps it would be better to throw the exception in dataloader e.g. JsonLoader, XmlLoader?

dsmiley: but it's not specific to the input type?

dsmiley:
it's a shame we've lost some consistency in AddUpdateCommand returning Lucene docs from both cmd.getLuceneDocument and cmd.iterable()... we've removed the iterable with something computing a List of SolrInputDocument (not a Lucene Document).  It would be good to keep the use of DocumentBuilder to one place, either DirectUpdateHandler2 or AddUpdateCommand but not both.  Hmmm; let me know what you think.

moshebla:
@dsmiley , we could move the [toDocumentsIter](https://github.com/apache/lucene-solr/pull/385/files/4407c73c4b893bbcc64f3f720bec2dd6dbf05cb3#diff-ebdc4ecf6a2398f102ba7fae37648d10R991) to AddUpdateCommand, thus limiting the use of DocumentBuilder to AddUpdateCommand, as per the previous commits.

solr/core/src/java/org/apache/solr/update/AddUpdateCommand.java
moshebla: True, never mind.

solr/core/src/java/org/apache/solr/update/AddUpdateCommand.java
```diff
@@ -175,69 +172,83 @@ public String getHashableId() {
...
-      public Document next() {
-        return DocumentBuilder.toDocument(iter.next(), req.getSchema());
-      }
+    boolean isVersion = version != 0;
 
-      @Override
-      public void remove() {
-        throw new UnsupportedOperationException();
+    for (SolrInputDocument sdoc : all) {
+      if (all.size() > 1) {
```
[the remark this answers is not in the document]
moshebla: @dsmiley, It would be lovely if you could

solr/core/src/java/org/apache/solr/update/DirectUpdateHandler2.java
moshebla: After looking at the code, it seems like updateDocOrDocValues has to return the idTerm to be used by the caller, and the updateTerm can then be figured out by the caller, either the cmd.updateTerm or the idTerm.
I will push a commit soon, hopefully the code can do the talking.

solr/core/src/java/org/apache/solr/update/DirectUpdateHandler2.java
```diff
@@ -952,12 +941,15 @@ public void split(SplitIndexCommand cmd) throws IOException {
...
    * @param cmd - cmd apply to IndexWriter
    * @param writer - IndexWriter to use
-   * @param updateTerm - used if this cmd results in calling {@link IndexWriter#updateDocument}
    */
-  private void updateDocOrDocValues(AddUpdateCommand cmd, IndexWriter writer, Term updateTerm) throws IOException {
+  private Term updateDocOrDocValues(AddUpdateCommand cmd, IndexWriter writer) throws IOException {
     assert null != cmd;
     final SchemaField uniqueKeyField = cmd.req.getSchema().getUniqueKeyField();
     final String uniqueKeyFieldName = null == uniqueKeyField ? null : uniqueKeyField.getName();
+    List<SolrInputDocument> docs = cmd.computeFinalFlattenedSolrDocs();
```
dsmiley: I don't think these lines you've added should be placed here, as it's computing final flattened docs when they may not even be needed -- which is when cmd.isInPlaceUpdate.  You could move this entirely to updateDocument()?

solr/core/src/java/org/apache/solr/update/DirectUpdateHandler2.java
```diff
@@ -972,27 +964,22 @@ private void updateDocOrDocValues(AddUpdateCommand cmd, IndexWriter writer, Term
       log.debug("updateDocValues({})", cmd);
       writer.updateDocValues(updateTerm, fieldsToUpdate.toArray(new Field[fieldsToUpdate.size()]));
     } else {
-      updateDocument(cmd, writer, updateTerm);
+      updateDocument(cmd, docs, writer, updateTerm, isBlock);
```
dsmiley: IMO of updateDocument is a method then the code above underneath cmd.isInPlaceUpdate should be a method as well -- `updateDocValues` to be balanced.  Or inline updateDocument.

solr/core/src/java/org/apache/solr/update/DirectUpdateHandler2.java
```diff
@@ -952,12 +941,15 @@ public void split(SplitIndexCommand cmd) throws IOException {
...
+  private Term updateDocOrDocValues(AddUpdateCommand cmd, IndexWriter writer) throws IOException {
     assert null != cmd;
     final SchemaField uniqueKeyField = cmd.req.getSchema().getUniqueKeyField();
     final String uniqueKeyFieldName = null == uniqueKeyField ? null : uniqueKeyField.getName();
+    List<SolrInputDocument> docs = cmd.computeFinalFlattenedSolrDocs();
+    boolean isBlock = docs.size() > 1;
+    Term idTerm = getIdTerm(cmd, isBlock);
+    Term updateTerm = cmd.hasUpdateTerm() ? cmd.updateTerm: idTerm;
 
     if (cmd.isInPlaceUpdate()) {
```
dsmiley: IMO since updateDocument is a method then the code here underneath cmd.isInPlaceUpdate should be a method as well -- `updateDocValues` to be balanced.  Or inline updateDocument.   (I know you didn't touch this)

solr/core/src/java/org/apache/solr/update/DirectUpdateHandler2.java
moshebla: Yeah, but that means we wouldn't have a way to figure out if the update is a block, which is a parameter which needs to be passed to getIdTerm.
I'm not as familiar with the internals, would it always be false during an inPlace update?

moshebla: If i pass isBlock as false in case of an inPlaceUpdate to getIdTerm, the overwrite test seems to fail, as one of the documents is not overwritten.

dsmiley:
I think I'll take it from here.  I'll post a patch later.  I started tweaking the IgnoreLargeDocuments URP just as you did last night.

PR #386, opened 2018-05-29
SOLR-12416 Add autoDeleteAge to optional router params

saua:
Since autoDeleteAge is not in either REQUIRED_ROUTER_PARAMS or OPTIONAL_ROUTER_PARAMS the CollectionsHandler will not pass it on and it will be ignored in the CREATEALIAS command.


ctargett:
https://issues.apache.org/jira/browse/SOLR-12416 was released in Solr 7.4, so this PR can be closed.

PR #387, opened 2018-05-29
Solr 12421

millerjeff0:
If we allow the passed in properties object to override the name written to originalCoreProperties then the core.properties file gets written with the wrong name in it.


janhoy:
Linking with SOLR-12421

PR #393, opened 2018-06-04
Lucene 8347

alessandrobenedetti:
Introduced multi term management for the BlendedInfix suggester.
The score of each suggestion will be calculated based on : 

- the positional coefficient of each token matched
- the length of the suggestion ( minor)

This patch is supposed to go in after LUCENE-8343


janhoy:
Linking with LUCENE-8347

PR #394, opened 2018-06-04
Synchronized disruption

cahilltr:


sigram:
FWIW, Solr supports a ScheduledTrigger and there's a default configuration called `.scheduled_maintenance` that executes `solr.InactiveShardPlanAction`. The functionality in your patch could be more or less implemented using this trigger and a custom TriggerAction that calls `System.gc()`.

(`ScheduledTriggger` doesn't support cron expressions but this could be fixed).

sigram:
Oh, I see - triggers execute ONLY on the current Overseer leader node - you would need some sort of fan-out call to each node in that action, and a handler to actually execute System.gc() on the node...

PR #395, opened 2018-06-05
WIP SOLR-12362: add tests for working relational docs

moshebla:


solr/core/src/test/org/apache/solr/handler/JsonLoaderTest.java
```diff
@@ -946,5 +946,87 @@ public void testGrandChildDocs() throws Exception {
 
   }
 
+  @Test
+  public void testRelationalChildDocs() throws Exception {
```
dsmiley: A bit on terminology.... if so-called "relational" child docs are going to be the only way to have a child doc in the future, then lets just start calling these simply child docs without the "relational" qualifier.  "Relational" may be kinda confusing since an anonymous relationship is still related.  Perhaps the new way is a "labelled" child doc?  If you disagree and want to keep some qualifier then okay.  If you see an existing test that says "child doc" for anonymous child doc then change its name to be about "anonymous" child docs.

solr/core/src/java/org/apache/solr/handler/loader/JsonLoader.java
```diff
@@ -556,82 +556,105 @@ private void parseFieldValue(SolrInputField sif) throws IOException {
       if (ev == JSONParser.OBJECT_START) {
         parseExtendedFieldValue(sif, ev);
       } else {
-        Object val = parseNormalFieldValue(ev, sif.getName());
+        Object val = parseNormalFieldValue(ev, sif);
         sif.setValue(val);
       }
     }
 
+    private Map<String, Object> generateExtendedValueMap(int ev) throws IOException {
```
dsmiley: Can you please sequence the methods so this follows where it's called from so readers needn't look back up?

solr/core/src/java/org/apache/solr/handler/loader/JsonLoader.java
```diff
@@ -556,82 +556,105 @@ private void parseFieldValue(SolrInputField sif) throws IOException {
...
+      assert ev == JSONParser.OBJECT_START;
+      Map<String, Object> extendedInfo = new HashMap<>();
+
+      for(; ; ) {
```
dsmiley: nitpick: I prefer while(true)

solr/core/src/java/org/apache/solr/handler/loader/JsonLoader.java
```diff
@@ -556,82 +556,105 @@ private void parseFieldValue(SolrInputField sif) throws IOException {
...
+        }
+        String label = parser.getString();
+        SolrInputField sif = new SolrInputField(label);
+        parseFieldValue(sif);
+        extendedInfo.put(label, sif.getValue());
+      }
+    }
+
+    private boolean isChildDoc(Map<String, Object> extendedMap) {
+      if (extendedMap.containsKey("value") && extendedMap.containsKey("boost")) {
```
dsmiley: lets instead simply check for the uniqueKey label

dsmiley:
Looks good!

solr/core/src/java/org/apache/solr/handler/loader/JsonLoader.java
```diff
@@ -661,13 +664,13 @@ private Object parseSingleFieldValue(int ev, SolrInputField sif) throws IOExcept
       }
     }
 
-    private boolean isChildDoc(Map<String, Object> extendedMap) {
+    private boolean isChildDoc(SolrInputDocument extendedMap) {
       return extendedMap.containsKey(req.getSchema().getUniqueKeyField().getName());
     }
 
-    private Map<String, Object> generateExtendedValueMap(int ev) throws IOException {
+    private SolrInputDocument generateExtendedValueMap(int ev) throws IOException {
```
dsmiley: this method name is no longer appropriate

solr/core/src/java/org/apache/solr/handler/loader/JsonLoader.java
```diff
@@ -249,14 +251,27 @@ public void handle(Map<String, Object> record, String path) {
...
           if (e.getValue() instanceof List) {
             List value = (List) e.getValue();
             for (Object o : value) {
-              if (o instanceof Map) result.addChildDocument(buildDoc((Map) o));
+              if (o instanceof Map) {
+                if (anonChildDocFlag) {
+                  result.addChildDocument(buildDoc((Map) o));
+                } else {
+                  if(!result.containsKey(e.getKey())) {
+                    result.setField(e.getKey(), new ArrayList<>(1));
```
dsmiley: This looks unnecessary; SolrInputDocument internally handles ensuring multiple values can be captures.

solr/core/src/java/org/apache/solr/handler/loader/JsonLoader.java
```diff
@@ -249,14 +251,27 @@ public void handle(Map<String, Object> record, String path) {
     private SolrInputDocument buildDoc(Map<String, Object> m) {
       SolrInputDocument result = new SolrInputDocument();
       for (Map.Entry<String, Object> e : m.entrySet()) {
-        if (e.getKey() == null) {// special case. JsonRecordReader emits child docs with null key
+        if (entryIsChildDoc(e.getValue())) {// special case. JsonRecordReader emits child docs with null key
```
dsmiley: the comment is concerning; we don't want JsonRecordReader to do that anymore?

solr/core/src/java/org/apache/solr/handler/loader/JsonLoader.java
```diff
@@ -668,7 +682,40 @@ private boolean isChildDoc(SolrInputDocument extendedMap) {
...
+    private boolean entryIsChildDoc(Object val) {
+      if(val instanceof List) {
+        List listVal = (List) val;
+        if (listVal.size() == 0) return false;
+        return  listVal.get(0) instanceof Map;
+      }
+      return val instanceof Map;
+    }
+
+    private void safeAddValue(SolrInputDocument doc, String fieldName, Object value) {
```
dsmiley: I'm confused why this is necessary.  What goes wrong if doc.addField(name,value) is called?  If something goes wrong and this is necessary, that explanation needs to be in the comments.

solr/core/src/test/org/apache/solr/handler/JsonLoaderTest.java
```diff
@@ -435,61 +435,80 @@ public void testJsonDocFormat() throws Exception{
 
   public void testFewParentsJsonDoc() throws Exception {
     String json = PARENT_TWO_CHILDREN_JSON;
+    assertFewParents(json, true);
+    tearDown();
```
dsmiley: I'm don't think I've ever seen a test do this, and I don't think we should start now.  Simply add two test methods, one that passes true, one that passes false.  test methods need to be named differently of course.  This pattern in testing is pretty common.

solr/solrj/src/java/org/apache/solr/common/util/JsonRecordReader.java
```diff
@@ -438,18 +442,19 @@ void walkObject() throws IOException {
       }
     }
 
-    private void addChildDoc2ParentDoc(Map<String, Object> record, Map<String, Object> values) {
+    private void addChildDoc2ParentDoc(Map<String, Object> record, Map<String, Object> values, String path) {
+      String trimmedPath = trimPath(path);
```
dsmiley: I think "path" should be renamed "key", and the trimPath() should be done by the caller, and not be called trimPath either for that matter -- it takes the "leaf" or "suffix" part of the path, or perhaps you have an idea of a better name.

solr/core/src/java/org/apache/solr/handler/loader/JsonLoader.java
moshebla: I changed the dictionary implementation to insert child documents into the map under their key, instead of null, so their relation to the parent document can be saved and used when SolrInputDocument is constructed.

solr/core/src/java/org/apache/solr/handler/loader/JsonLoader.java
moshebla: I tried the same method, except because SolrInputDocument implements iterable, each document key is added, instead of the SolrInputDocument object. This chain of event occurs because [SolrInputDocument.addField](https://github.com/moshebla/lucene-solr/blob/3c51c1c414c6f7cb359e3ef442ca706e33cf3a7a/solr/solrj/src/java/org/apache/solr/common/SolrInputDocument.java#L84) calls [SolrInputField.addValue](https://github.com/moshebla/lucene-solr/blob/3c51c1c414c6f7cb359e3ef442ca706e33cf3a7a/solr/solrj/src/java/org/apache/solr/common/SolrInputField.java#L91).

solr/core/src/java/org/apache/solr/handler/loader/JsonLoader.java
moshebla: It can handle it internally, but we have to ensure a JSON list is added as a list of one document, and not as a single document, to remain true to the original JSON document.

moshebla:
Hopefully these new commits address the problems with the previous changes

solr/core/src/java/org/apache/solr/handler/loader/JsonLoader.java
dsmiley: Okay I see.  Interesting -- there are clearly pros/cons to having SolrInputDocument implement core JDK abstractions.  It makes working with it easier, though now some things like what you see here happens automatically when we don't want it to.

There is no perfect solution here but I'd rather see SolrInputField.addValue do an instanceof check when it sees that `v` implements Iterable to guard against iterating the child doc.  This is much smaller/simpler than your safeAddValue, and it's something that won't be tied to Json or any other input format.  Even someone who wants to use SolrJ and thinks that they can just call set/addValue would be surprised it doesn't work unless we make the change at SolrInputField.
Also admittedly, this is a gotcha to the path of putting nested child docs in values, but not a big deal.

solr/core/src/java/org/apache/solr/handler/loader/JsonLoader.java
dsmiley: ok so you are trying to retain the semantic information that the input was an array (potentially multi-valued) even if the particular input doc only had one value.  Maybe say more or that, hinting at why we even care.  (why do we care)?  For regular field values, we don't at this level -- the schema holds multiValue info so it's dealth with later.  Again I'm liking my suggestion of putting a virtual child doc field in the schema as it addresses the desire to know this semantic info, plus I think it may add some consistency (avoids special case you're doing here)

solr/solrj/src/java/org/apache/solr/common/util/JsonRecordReader.java
```diff
@@ -442,19 +442,18 @@ void walkObject() throws IOException {
       }
     }
 
-    private void addChildDoc2ParentDoc(Map<String, Object> record, Map<String, Object> values, String path) {
-      String trimmedPath = trimPath(path);
+    private void addChildDoc2ParentDoc(Map<String, Object> record, Map<String, Object> values, String key) {
```
dsmiley: much better; thanks

solr/core/src/java/org/apache/solr/handler/loader/JsonLoader.java
moshebla: So if I understood correctly, I also thought of checking to see if the addField value is of SolrDocumentBase, to prevent going through the iterator. I was afraid of backwards compatibility, but after some thought, until [SOLR-12361](https://issues.apache.org/jira/browse/SOLR-12361) childDocuments inside doc keys was not supported.

solr/core/src/java/org/apache/solr/handler/loader/JsonLoader.java
moshebla: if we wanted to build the original JSON document as it was indexed, we would have to store it as a single valued array if that was the case. I'm afraid I did not quite get what you meant by using virtual fields, what for?

dsmiley: By "virtual field" I'm referring to my earlier idea about having a special field in the schema that designates a child document with this relation.  it would have a name, multiValued option, and probably that's it.  https://issues.apache.org/jira/browse/SOLR-12298?focusedCommentId=16467523&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-16467523

dsmiley: Any way; lets not increase the scope of this particular issue with my idea to add this info to the schema. It seems adequate to force the use of an internal array for a single value in SolrInputField that we know originated from an array.

solr/core/src/java/org/apache/solr/handler/loader/JsonLoader.java
```diff
@@ -556,82 +572,71 @@ private void parseFieldValue(SolrInputField sif) throws IOException {
       if (ev == JSONParser.OBJECT_START) {
         parseExtendedFieldValue(sif, ev);
       } else {
-        Object val = parseNormalFieldValue(ev, sif.getName());
+        Object val = parseNormalFieldValue(ev, sif);
         sif.setValue(val);
       }
     }
 
     private void parseExtendedFieldValue(SolrInputField sif, int ev) throws IOException {
```
dsmiley: Lets put "ev" as first arg to be more consistent

solr/core/src/java/org/apache/solr/handler/loader/JsonLoader.java
```diff
@@ -556,82 +572,71 @@ private void parseFieldValue(SolrInputField sif) throws IOException {
...
-              if (ev != JSONParser.NUMBER &&
-                  ev != JSONParser.LONG &&
-                  ev != JSONParser.BIGNUMBER) {
-                throw new SolrException(SolrException.ErrorCode.BAD_REQUEST, "Boost should have number. "
-                    + "Unexpected " + JSONParser.getEventString(ev) + " at [" + parser.getPosition() + "], field=" + sif.getName());
-              }
+      if (isChildDoc(extendedSolrDocument)) {
+        SolrInputDocument cDoc = new SolrInputDocument();
+        for (Map.Entry<String, SolrInputField> extendedEntry: extendedSolrDocument.entrySet()) {
+          cDoc.setField(extendedEntry.getKey(), extendedEntry.getValue().getValue());
```
dsmiley: Why can't you simply do sif.addValue(extendedSolrDocument)?  Thus avoiding the rebuilding of a SolrInputDocument that seems to have no purpose.

solr/core/src/java/org/apache/solr/handler/loader/JsonLoader.java
```diff
@@ -664,8 +672,38 @@ private Object parseSingleFieldValue(int ev, String fieldName) throws IOExceptio
...
+
+      while(true) {
+        ev = parser.nextEvent();
+        if (ev == JSONParser.OBJECT_END) {
+          return extendedInfo;
+        }
+        String label = parser.getString();
+        SolrInputField sif = new SolrInputField(label);
+        parseFieldValue(sif);
+        extendedInfo.addField(label, sif.getValue());
```
dsmiley: I think you can use extendedInfo.putField(key,SolrInputField) here?  Oh I see comments elsewhere in this file explaining why addField is used; maybe copy-paste those comments?

solr/core/src/java/org/apache/solr/handler/loader/JsonLoader.java
```diff
@@ -664,8 +672,38 @@ private Object parseSingleFieldValue(int ev, String fieldName) throws IOExceptio
...
+    private boolean entryIsChildDoc(Object val) {
+      if(val instanceof List) {
+        List listVal = (List) val;
+        if (listVal.size() == 0) return false;
+        return  listVal.get(0) instanceof Map;
+      }
+      return val instanceof Map;
+    }
+
+    private SolrInputDocument generateExtendedValueDoc(int ev) throws IOException {
```
dsmiley: the name seems off -- 'generate' isn't bad but this method is parsing and plenty of other methods here parse and are named as such so I think this should be `parseExtendedValueAsDoc` or better `parseExtendedFieldValueAsDoc` since it is only called by `parseExtendedFieldValue`.

solr/core/src/java/org/apache/solr/handler/loader/JsonLoader.java
```diff
@@ -664,8 +672,38 @@ private Object parseSingleFieldValue(int ev, String fieldName) throws IOExceptio
...
           return lst;
         }
-        Object val = parseSingleFieldValue(ev, fieldName);
+        Object val = parseSingleFieldValue(ev, sif);
         lst.add(val);
+        sif.setValue(null);
+      }
+    }
+
+    private boolean isChildDoc(SolrInputDocument extendedMap) {
```
dsmiley: rename extendedMap to extendedFieldValue

solr/core/src/java/org/apache/solr/handler/loader/JsonLoader.java
```diff
@@ -556,82 +572,71 @@ private void parseFieldValue(SolrInputField sif) throws IOException {
...
-            if (extendedInfo != null) {
-              if (normalFieldValue != null) {
-                extendedInfo.put("value", normalFieldValue);
-              }
-              sif.setValue(extendedInfo);
-            } else {
-              sif.setValue(normalFieldValue);
-            }
-            return;
+      for (String label: extendedSolrDocument.keySet() ) {
```
dsmiley: Whenever I see looping over keys from a map-like thing, I often see a line of code that then fetches the value, as seen here.  This internally results in a bunch of lookups that aren't necessary.  Instead,  note that SolrInputDocument is Iterable<SolrInputField> (which itself has the key and value), so just loop over that.

moshebla:
Tried tackling the javadoc references, as I'm fairly new this I hope I didn't mess it up too badly.

solr/core/src/java/org/apache/solr/handler/loader/JsonLoader.java
```diff
@@ -570,37 +573,34 @@ private SolrInputDocument parseDoc(int ev) throws IOException {
...
 
       Object normalFieldValue = null;
       Map<String, Object> extendedInfo = null;
 
-      for (String label: extendedSolrDocument.keySet() ) {
-        Object fieldVal = extendedSolrDocument.get(label).getValue();
-        if ("boost".equals(label)) {
+      for (SolrInputField field: extendedSolrDocument) {
+        Object fieldVal = field.getValue();
+        String fieldName = field.getName();
```
dsmiley: I can understand why you chose the names `fieldName` and `fieldVal` because you are getting this from a SolrInputField.  But recognize what's going on here is very unusual ... we are at this point treating this SolrInputDocument as if it were a map and not as a document.  Lets add comments about this very explicitly before we loop because this is so strange.  Maybe the `field` var could be named `entry`.  The _former_ name `label` (or perhaps `extLabel`) var name is much more appropriate as it is _not_ a treated as *field* label/name.  Likewise, just choose `val` or `value` or `extValue`.

Lets also add a bit of docs to parseExtendedFieldValue explain what this is for, since it's very non-obvious, I think.  Essentially, this is a JSON object that is _either_ a child doc, or it's a Map (used for atomic updates), or it's an outdated mechanism to specify the boost which is no longer in Lucene/Solr.

solr/core/src/java/org/apache/solr/handler/loader/JsonLoader.java
```diff
@@ -703,6 +703,10 @@ private SolrInputDocument generateExtendedValueDoc(int ev) throws IOException {
         String label = parser.getString();
         SolrInputField sif = new SolrInputField(label);
         parseFieldValue(sif);
+        // pulling out the pieces may seem weird, but it's because
```
dsmiley: Good.  The duplication of comment now has me wondering about this method overall -- why does it seem to be duplicating existing logic -- buildDoc?  Would things "just work" if we replaced generateExtendedValueDoc with buildDoc()?  I think it would?  Less look-a-like code to maintain.  I suppose the only problem with that is it would permit weird things like anonymous child docs on a labelled child doc.  I just wouldn't worry about that now though; we catch that later in AddUpdateCommand.flatten().

BTW I'm sure you realize now that I don't have everything figured out from the start; it's an iterative journey for you and me/the-reviewer.  :-)

moshebla: It's actually quite refreshing to get another person's opinion.
I don't think it would work flawlessly without changing a lot of the code, since buildDoc expects a Map as an input, so we would have to parse the map beforehand.

solr/core/src/java/org/apache/solr/handler/loader/JsonLoader.java
moshebla: Just tackled this one.

solr/core/src/java/org/apache/solr/handler/loader/JsonLoader.java
```diff
@@ -593,14 +602,14 @@ private void parseExtendedFieldValue(int ev, SolrInputField sif) throws IOExcept
       Object normalFieldValue = null;
       Map<String, Object> extendedInfo = null;
 
-      for (SolrInputField field: extendedSolrDocument) {
-        Object fieldVal = field.getValue();
-        String fieldName = field.getName();
-        if ("boost".equals(fieldName)) {
-          Object boostVal = fieldVal;
+      for (SolrInputField entry: extendedSolrDocument) {
+        Object label = entry.getValue();
```
dsmiley: LOL you have these mixed up :-)  entry.getName should be the label.

solr/core/src/java/org/apache/solr/handler/loader/JsonLoader.java
moshebla: Skimming through the code on master, it looks as if the map parsing was done here before only partially. Currently it seems like we either duplicate the map parsing part, or the document parsing part. If we go with the map part, we would build an intermediate map and then parse it to a document instead of streaming it using the parser. I'm not quite sure if using buildDoc would be as beneficial, since in either case, we have to duplicate some part.

solr/core/src/java/org/apache/solr/handler/loader/JsonLoader.java
moshebla: oops

solr/core/src/java/org/apache/solr/handler/loader/JsonLoader.java
dsmiley: Sorry I meant `parseDoc`

moshebla: True, didn't quite notice it, I'll upload a commit soon. :smile:

moshebla: It worked.

moshebla:
Skimmed through, making some last minute changes.

moshebla:
Is it OK to close this, since SOLR-12362 was resolved?

dsmiley:
Yes. Oddly enough,I don’t have permissions to close PRs unless I remember to mention it inthe commit message.

PR #397, opened 2018-06-06
Matches highlighter

romseygeek:


lucene/core/src/java/org/apache/lucene/index/LeafReaderContext.java
```diff
@@ -35,7 +35,7 @@
   /**
    * Creates a new {@link LeafReaderContext} 
    */    
-  LeafReaderContext(CompositeReaderContext parent, LeafReader reader,
+  public LeafReaderContext(CompositeReaderContext parent, LeafReader reader,
                     int ord, int docBase, int leafOrd, int leafDocBase) {
```
romseygeek: This is required to fake up the OffsetsReader

PR #398, opened 2018-06-08
Lucene 8343 data type migration

alessandrobenedetti:
Different approach, data type migration to fix the bugs  : 

1) Weight for the Document dictionary moved to Long from long
2) Suggestion score moved to double from long


lucene/suggest/src/java/org/apache/lucene/search/suggest/analyzing/BlendedInfixSuggester.java
```diff
@@ -200,7 +201,13 @@ protected FieldType getTextFieldType() {
       textDV.advance(fd.doc);
 
       final String text = textDV.binaryValue().utf8ToString();
-      long weight = (Long) fd.fields[0];
+
+      NumericDocValues weightDV = MultiDocValues.getNumericValues(searcher.getIndexReader(), WEIGHT_FIELD_NAME);
```
nvnmandadhi: Could you please make local variables final to prevent reassignment.

alessandrobenedetti: Thank you for your note, It was not the main scope of the bug fix, but It's a good recommendation so I just contributed the little refactor.
Thanks!

nvnmandadhi: Thanks for incorporating the changes!!!!!

lucene/suggest/src/java/org/apache/lucene/search/suggest/InputIterator.java
```diff
@@ -34,7 +34,7 @@
 public interface InputIterator extends BytesRefIterator {
 
   /** A term's weight, higher numbers mean better suggestions. */
```
mikemccand: Maybe add javadocs explaining what `null` means?  Though, why do we need to allow for `null`?  Shouldn't the iterator not return a suggestion that has no weight?

lucene/suggest/src/java/org/apache/lucene/search/suggest/Lookup.java
```diff
@@ -53,7 +53,7 @@
     public final Object highlightKey;
 
     /** the key's weight */
-    public final long value;
+    public final double value;
```
mikemccand: Maybe improve javadocs here, explaining that this is not just the weight originally supplied during indexing, for some suggesters?

lucene/suggest/src/java/org/apache/lucene/search/suggest/InputIterator.java
alessandrobenedetti: Hi Michael,
The reason to allow for null at the InputIterator level is to distinguish it from an explicit 0 weight.
In the DocumentDictionary this translates in differentiating when  the weight field was missing for the original document ( NULL ) in opposition to when the weight field was present and with 0 value.
At this level we just want to ensure that the same behavior is maintained when we build the auxiliary index : 
i.e. if the weight field was missing for the original document, I want it to be null for the auxiliary index as well.
How the different suggesters implementation will use this to return a suggestion score, I think will depend on a case by case scenario.
Did I misunderstand anything here ?

lucene/suggest/src/java/org/apache/lucene/search/suggest/Lookup.java
alessandrobenedetti: I agree, just adde!

PR #402, opened 2018-06-11
SOLR-12477 - Return server error(500) for AlreadyClosedException instead of client Errors(400)

jefferyyuan:
https://issues.apache.org/jira/browse/SOLR-12477

In some cases(for example: corrupt index), addDoc0 throws AlreadyClosedException, but solr server returns client error 400 to client

This will confuse customers and especially monitoring tool.

### DIFF
--- a/solr/core/src/java/org/apache/solr/handler/RequestHandlerBase.java
+++ b/solr/core/src/java/org/apache/solr/handler/RequestHandlerBase.java
@@ -209,7 +209,16 @@ public abstract class RequestHandlerBase implements SolrRequestHandler, SolrInfo
       }
     } catch (Exception e) {
       if (req.getCore() != null) {
-        req.getCore().getCoreContainer().checkTragicException(req.getCore());
+        boolean isTragic = req.getCore().getCoreContainer().checkTragicException(req.getCore());
+        if (isTragic) {
+          if (e instanceof SolrException) {
+            // Tragic exceptions should always throw a server error
+            assert ((SolrException) e).code() == 500;
+          } else {
+            // wrap it in a solr exception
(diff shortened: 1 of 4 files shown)


solr/core/src/java/org/apache/solr/handler/RequestHandlerBase.java
```diff
@@ -208,14 +208,17 @@ public void handleRequest(SolrQueryRequest req, SolrQueryResponse rsp) {
         }
       }
     } catch (Exception e) {
+      boolean hasTragicException = false;
```
vthacker: Here is what I was thinking of after this statement. What do you think?

        if (isTragic) {
          if (e instanceof SolrException) {
            assert ((SolrException) e).code() == 500; //Tragic exceptions should always throw a server error
          } else {
            //wrap it in a solr exception
            e = new SolrException(SolrException.ErrorCode.SERVER_ERROR, e);
          }
        }

vthacker: Looking at LeaderTragicEventTest.java maybe we should check if the code is a 500 or a 404 also?

jefferyyuan: @vthacker 
Changed the code as you suggested, and like you variable name: isTragic :)
I checked the code is one of 500 or 404 in LeaderTragicEventTest.java for update or commit:
```java
   assertThat(se.code(), anyOf(is(500), is(404)));
```

jefferyyuan:
Close this as @vthacker has submitted this in https://issues.apache.org/jira/browse/SOLR-12477 :)
- (cherry picked from commit 8d28bbc)

PR #405, opened 2018-06-17
don't die when java prints tool options

tjakway:
If you have JAVA_TOOL_OPTIONS set (e.g. in my .profile, `export JAVA_TOOL_OPTIONS="-Dfile.encoding=\"UTF-8\" -Dawt.useSystemAAFontSettings=on"`) you get a confusing message about having the wrong version of java installed because the launch script only looks at the first line of the output of `java -version`.  Simple 1-liner fix.

### DIFF
--- a/solr/bin/solr
+++ b/solr/bin/solr
@@ -124,7 +124,7 @@ if [ -z "$SOLR_STOP_WAIT" ]; then
   SOLR_STOP_WAIT=180
 fi
 # test that Java exists, is executable and correct version
-JAVA_VER=$("$JAVA" -version 2>&1)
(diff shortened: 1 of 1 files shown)


sarowe:
Can you make a JIRA issue to track this please?  https://wiki.apache.org/solr/HowToContribute#JIRA_tips_.28our_issue.2Fbug_tracker.29

Also, it might be better to extract the line containing version info, rather than ignoring specific lines that don't contain it, since there may be other bad output lines in the future.

janhoy:
Thanks for the contribution. Have you sees @sarowe 's comment?

PR #407, opened 2018-06-19
SOLR-12413 - prefer version zero from zookeeper

nsoft:


ctargett:
https://issues.apache.org/jira/browse/SOLR-12413 was released in Solr 7.5, so this PR can be closed.

PR #410, opened 2018-06-25
SOLR-12441: add deeply nested URP for nested documents metadata

moshebla:


solr/core/src/java/org/apache/solr/update/processor/DeeplyNestedUpdateProcessor.java
```diff
@@ -0,0 +1,101 @@
...
+    processDocChildren(doc, null);
+    super.processAdd(cmd);
+  }
+
+  private void processDocChildren(SolrInputDocument doc, String fullPath) {
+    for(SolrInputField field: doc.values()) {
+      if(field.getFirstValue() instanceof SolrInputDocument) {
+        Object val = field.getValue();
+        fullPath = Objects.isNull(fullPath) ? field.getName(): String.format(Locale.ROOT,"%s.%s", fullPath, field.getName());
+        if (val instanceof Collection) {
```
dsmiley: SolrInputField is directly Iterable over its values, so you needn't call getValue and check its type.

solr/core/src/java/org/apache/solr/update/processor/DeeplyNestedUpdateProcessor.java
```diff
@@ -0,0 +1,101 @@
...
+    SolrInputDocument doc = cmd.getSolrInputDocument();
+    processDocChildren(doc, null);
+    super.processAdd(cmd);
+  }
+
+  private void processDocChildren(SolrInputDocument doc, String fullPath) {
+    for(SolrInputField field: doc.values()) {
+      if(field.getFirstValue() instanceof SolrInputDocument) {
+        Object val = field.getValue();
+        fullPath = Objects.isNull(fullPath) ? field.getName(): String.format(Locale.ROOT,"%s.%s", fullPath, field.getName());
```
dsmiley: I can't say I'm a fan of String.format -- it usually is just a more verbose & complex way to do trivial string concatenation.  Granted String.format _sometimes_ has value (e.g. formatting floating point); it isn't exercised here.

Shouldn't we throw a BAD_REQUEST exception if field.getName() contains the character we're using to join ('.')?

There's an O(N^2) string concatenation here where N is the depth of the tree but hopefully the docs won't be *that* nested to worry about this.  We could add a comment to at least acknowledge this.

solr/core/src/java/org/apache/solr/update/processor/DeeplyNestedUpdateProcessor.java
```diff
@@ -0,0 +1,101 @@
...
+          }
+        } else {
+          processChildDoc((SolrInputDocument) val, doc, fullPath);
+        }
+      }
+    }
+  }
+
+  private void processChildDoc(SolrInputDocument sdoc, SolrInputDocument parent, String fullPath) {
+    if(fields == NestedFlag.ALL) {
```
dsmiley: I don't think we need special-case ALL

solr/core/src/java/org/apache/solr/update/processor/DeeplyNestedUpdateProcessor.java
```diff
@@ -0,0 +1,101 @@
...
+      }
+      if(fields.contains(NestedFlag.LEVEL)) {
+        setLevelKey(sdoc, fullPath);
+      }
+    }
+    processDocChildren(sdoc, fullPath);
+  }
+
+  private void setLevelKey(SolrInputDocument sdoc, String fullPath) {
+    sdoc.addField(IndexSchema.LEVEL_FIELD_NAME, fullPath.split("\\.").length);
```
dsmiley: yuck; sorry.  Perhaps instead of pre-concatenating the full path, we instead pass full path around as a List<String> to the various arguments and then we only materialize it in setPathField?  This may also ameliorate the O(N^2) problem although we would be re-concatenating the earlier part of the string but that doesn't seem like a big deal.

solr/core/src/java/org/apache/solr/update/processor/DeeplyNestedUpdateProcessor.java
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  private void setParentKey(SolrInputDocument sdoc, SolrInputDocument parent) {
+    sdoc.addField(IndexSchema.PARENT_FIELD_NAME, parent.getFieldValue(req.getSchema().getUniqueKeyField().getName()));
```
dsmiley: For these "set" methods, I think we want to be calling *setField* and not *addField*; no?  Granted it'd be weird if there was an existing value... and if there was, maybe we shouldn't overwrite it?  Probably shouldn't.

solr/core/src/java/org/apache/solr/update/processor/DeeplyNestedUpdateProcessorFactory.java
```diff
@@ -0,0 +1,74 @@
...
+  private static EnumSet<NestedFlag> generateNestedFlags(List<String> fields) {
+    return EnumSet.copyOf(fields.stream().map(e -> NestedFlag.valueOf(e.toUpperCase(Locale.ROOT))).collect(Collectors.toList()));
+  }
+
+  public enum NestedFlag {
+    PATH,
+    PARENT,
+    LEVEL;
+
+    public static final EnumSet<NestedFlag> ALL = EnumSet.allOf(NestedFlag.class);
```
dsmiley: Nice

solr/core/src/java/org/apache/solr/update/processor/DeeplyNestedUpdateProcessor.java
```diff
@@ -0,0 +1,101 @@
...
+
+import org.apache.solr.common.SolrInputDocument;
+import org.apache.solr.common.SolrInputField;
+import org.apache.solr.request.SolrQueryRequest;
+import org.apache.solr.response.SolrQueryResponse;
+import org.apache.solr.schema.IndexSchema;
+import org.apache.solr.update.AddUpdateCommand;
+import static org.apache.solr.update.processor.DeeplyNestedUpdateProcessorFactory.NestedFlag;
+
+public class DeeplyNestedUpdateProcessor extends UpdateRequestProcessor {
```
dsmiley: I wonder if we want to call this as-such.  Can this be used if the nesting is "shallow"?  Perhaps remove the "Deeply" part.  Perhaps this URP should handle "_root_" as well to thus unify where nested child doc fields are added?  Although that raises other questions.  With 8.0 on the horizon this fall, we can think of the best solution that isn't necessarily fully backwards compatible.  This conversation should probably be taken to the JIRA as it's a bigger question and not a code detail.

solr/core/src/java/org/apache/solr/schema/IndexSchema.java
```diff
@@ -104,9 +104,12 @@
   public static final String FIELD_TYPE = "fieldType";
   public static final String FIELD_TYPES = FIELD_TYPE + "s";
   public static final String INTERNAL_POLY_FIELD_PREFIX = "*" + FieldType.POLY_FIELD_SEPARATOR;
+  public static final String LEVEL_FIELD_NAME = "_nestLevel_";
```
dsmiley: I think these should nest ones should start with "NEST_"

solr/core/src/java/org/apache/solr/update/processor/NestedUpdateProcessor.java
```diff
@@ -0,0 +1,85 @@
...
+import org.apache.solr.common.SolrInputField;
+import org.apache.solr.request.SolrQueryRequest;
+import org.apache.solr.response.SolrQueryResponse;
+import org.apache.solr.schema.IndexSchema;
+import org.apache.solr.update.AddUpdateCommand;
+import static org.apache.solr.update.processor.NestedUpdateProcessorFactory.NestedFlag;
+
+public class NestedUpdateProcessor extends UpdateRequestProcessor {
+  public static final String splitChar = ".";
+  private EnumSet<NestedFlag> fields;
```
dsmiley: Although I do think EnumSet / enums are pretty nifty JDK utilities, in this case two booleans would be more clear?

solr/core/src/java/org/apache/solr/update/processor/NestedUpdateProcessor.java
```diff
@@ -0,0 +1,85 @@
...
+import org.apache.solr.common.SolrInputDocument;
+import org.apache.solr.common.SolrInputField;
+import org.apache.solr.request.SolrQueryRequest;
+import org.apache.solr.response.SolrQueryResponse;
+import org.apache.solr.schema.IndexSchema;
+import org.apache.solr.update.AddUpdateCommand;
+import static org.apache.solr.update.processor.NestedUpdateProcessorFactory.NestedFlag;
+
+public class NestedUpdateProcessor extends UpdateRequestProcessor {
+  public static final String splitChar = ".";
```
dsmiley: nitpick: "splitChar" name suggests we are going to "split" on something but we don't.  I think "PATH_SEP_CHAR" is a better constant name.  If it suits the usage details it could be of type character but if not leave as String.

solr/core/src/java/org/apache/solr/update/processor/NestedUpdateProcessorFactory.java
```diff
@@ -0,0 +1,72 @@
...
+
+  private EnumSet<NestedFlag> fields;
+  private static final List<String> allowedConfFields = Arrays.stream(NestedFlag.values()).map(e -> e.toString().toLowerCase(Locale.ROOT)).collect(Collectors.toList());
+
+  public UpdateRequestProcessor getInstance(SolrQueryRequest req, SolrQueryResponse rsp, UpdateRequestProcessor next ) {
+    return new NestedUpdateProcessor(req, rsp, fields, next);
+  }
+
+  @Override
+  public void init( NamedList args )
```
dsmiley: Maybe lets not have any configuration at all -- let the schema be the guide.  If the schema contains `IndexSchema.PATH_FIELD_NAME` then we should populate it, if not then don't.  Ditto for `PARENT_FIELD_NAME`.  The getInstance method of the factory can detect if there is any work to do and if none then return "next", thus avoiding overhead.

solr/core/src/test/org/apache/solr/update/TestNestedUpdateProcessor.java
```diff
@@ -0,0 +1,158 @@
...
+import org.apache.solr.response.SolrQueryResponse;
+import org.apache.solr.schema.IndexSchema;
+import org.apache.solr.servlet.SolrRequestParsers;
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+
+import static org.apache.solr.update.processor.NestedUpdateProcessor.splitChar;
```
dsmiley: After Yonik gave some sage advice on this topic once, I now think tests ought not to refer to constants in the tested code.  This way in the future if we change our minds on what those constants refer to, we're then realizing we may break back-compat.

solr/core/src/test/org/apache/solr/update/TestNestedUpdateProcessor.java
```diff
@@ -0,0 +1,158 @@
...
+    assertJQ(req("q", IndexSchema.PATH_FIELD_NAME + ":children",
+        "fl","*",
+        "sort","id asc",
+        "wt","json"),
+        childrenTests);
+  }
+
+  @Test
+  public void testDeeplyNestedURPFieldNameException() throws Exception {
+    final String errMsg = "contains: '" + splitChar + "' , which is reserved for the nested URP";
```
dsmiley: Lets consider rewording the error message so that the user is able to understand it without having to know about URPs.  I suggest: "The field/relationship name 'children.a' to a nested document cannot have a '.' as it interferes with internal processing".

Come to think of it, lets use '/' as this is not a valid field name char whereas '.' is?

solr/core/src/java/org/apache/solr/update/processor/NestedUpdateProcessor.java
moshebla: Yes, though in the future if more meta-data is to be added it will be a lot more painful to add more fields

moshebla: Replaced the EnumSet with two booleans as you proposed.

moshebla:
I added a test for "id"less children, which is currently because the JSON loader does not support it yet, as discussed in [SOLR-12362](https://issues.apache.org/jira/browse/SOLR-12362)

solr/core/src/test/org/apache/solr/update/TestNestedUpdateProcessor.java
```diff
@@ -0,0 +1,189 @@
...
+
+  @Test
+  public void testDeeplyNestedURPFieldNameException() throws Exception {
+    final String errMsg = "contains: '" + PATH_SEP_CHAR + "' , which is reserved for the nested URP";
+    thrown.expect(SolrException.class);
+    indexSampleData(errDoc);
+    thrown.expectMessage(errMsg);
+  }
+
+  private void indexSampleData(String cmd) throws Exception {
```
dsmiley: I'm skeptical we need to write this much code for something routine.  See `updateJ()`; can you just call that?

solr/core/src/test/org/apache/solr/update/TestNestedUpdateProcessor.java
```diff
@@ -0,0 +1,189 @@
...
+      "                    \"id\": \"3\",\n" +
+      "                    \"foo_s\": \"Bar\"\n" +
+      "                }\n" +
+      "            ]\n" +
+      "        }\n" +
+      "    }\n" +
+      "}";
+
+  @Rule
+  public ExpectedException thrown = ExpectedException.none();
```
dsmiley: This is neat; I didn't know about this utility.  But take note of `SolrTestCaseJ4.assertQEx(...)` especially the one that takes the exception message.  You could use that?

solr/core/src/java/org/apache/solr/update/processor/NestedUpdateProcessor.java
```diff
@@ -0,0 +1,96 @@
...
+  private void processDocChildren(SolrInputDocument doc, String rootId, String fullPath) {
+    int childNum = 0;
+    for(SolrInputField field: doc.values()) {
+      for(Object val: field) {
+        if(val instanceof SolrInputDocument) {
+          if(field.getName().contains(PATH_SEP_CHAR)) {
+            throw new SolrException(SolrException.ErrorCode.BAD_REQUEST, "Field name: '" + field.getName()
+                + "' contains: '" + PATH_SEP_CHAR + "' , which is reserved for the nested URP");
+          }
+          final String jointPath = Objects.isNull(fullPath) ? field.getName(): String.join(PATH_SEP_CHAR, fullPath, field.getName());
```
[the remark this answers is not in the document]
moshebla: Do you suggest using StringBuilder instead?

solr/core/src/test/org/apache/solr/update/TestNestedUpdateProcessor.java
moshebla: It seems like if I use updateJ(), I do not build the SolrRequest have manually any more, which will only complicate the test. You will probably be able to see it when I push the new commits.

solr/core/src/test/org/apache/solr/update/TestNestedUpdateProcessor.java
moshebla: Sorry, kind of missed that method.
My bad.

solr/core/src/java/org/apache/solr/update/processor/NestedUpdateProcessor.java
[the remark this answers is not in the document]
dsmiley: Simply use Java's String concatenation syntax:  `fullPath + PATH_SEP_CHAR + field.getName()`  BTW this translates internally to a StringBuilder though that's an implementation detail.

solr/core/src/java/org/apache/solr/update/processor/NestedUpdateProcessor.java
```diff
@@ -0,0 +1,96 @@
...
+      setPathField(sdoc, fullPath);
+    }
+    if (storeParent) {
+      setParentKey(sdoc, parent);
+    }
+    processDocChildren(sdoc, rootId, fullPath);
+  }
+
+  private String generateChildUniqueId(String rootId, String childPath, int childNum) {
+    return String.join(PATH_SEP_CHAR, rootId, childPath, Integer.toString(childNum));
```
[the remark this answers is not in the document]
dsmiley: See my point?

solr/core/src/java/org/apache/solr/update/processor/NestedUpdateProcessor.java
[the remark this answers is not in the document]
moshebla: Oh nice,
changed it

solr/core/src/java/org/apache/solr/update/processor/NestedUpdateProcessor.java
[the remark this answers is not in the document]
moshebla: Would label be the key of the document?

solr/core/src/java/org/apache/solr/update/processor/NestedUpdateProcessorFactory.java
```diff
@@ -0,0 +1,120 @@
...
+    }
+    if (storeParent) {
+      setParentKey(sdoc, parent);
+    }
+    processDocChildren(sdoc, fullPath);
+  }
+
+  private String generateChildUniqueId(String parentId, String childKey, int childNum) {
+    // combines parentId with the child's key and childNum. e.g. "10/footnote/1"
+    return String.join(PATH_SEP_CHAR, parentId, childKey, Integer.toString(childNum));
```
dsmiley: Glad you fixed the comment.
Another dubious use of of String.join
What did you think of my suggestion to use some other separator char (not PATH_SEP_CHAR) over here.  My first suggestion was a pound symbol.  Though it wouldn't show when in a URL... maybe a comma.  But we could stay with a '/'; I just thought it might be nice to make a distinction between the separator between parent/child and the separator to append the child sequence/num.
We ought to surface this to the JIRA as it's likely to get input.

solr/core/src/java/org/apache/solr/schema/IndexSchema.java
```diff
@@ -107,6 +107,8 @@
   public static final String LUCENE_MATCH_VERSION_PARAM = "luceneMatchVersion";
   public static final String MAX_CHARS = "maxChars";
   public static final String NAME = "name";
+  public static final String PARENT_FIELD_NAME = "_NEST_PARENT_";
```
dsmiley: In one of my earliest review comments I suggested renaming both constants to start with `_NEST_` so that they are more clearly grouped together in their purpose.  WDYT?

solr/core/src/java/org/apache/solr/update/processor/NestedUpdateProcessorFactory.java
```diff
@@ -0,0 +1,120 @@
...
+    if(storePath) {
+      setPathField(sdoc, fullPath);
+    }
+    if (storeParent) {
+      setParentKey(sdoc, parent);
+    }
+    processDocChildren(sdoc, fullPath);
+  }
+
+  private String generateChildUniqueId(String parentId, String childKey, int childNum) {
```
dsmiley: It occurred to me that this feature only occurs for labelled child docs, not anonymous ones.  And furthermore only if these two special fields are in the schema.  We'll have to remember to document that in the ref guide when we get to that point.

solr/core/src/test/org/apache/solr/update/TestNestedUpdateProcessor.java
```diff
@@ -0,0 +1,171 @@
...
+    final String expectedId = parentId + PATH_SEP_CHAR + childKey + PATH_SEP_CHAR + "0";
+    SolrInputDocument noIdChildren = sdoc("id", "1", "children", sdocs(sdoc("id", "2", "name_s", "Yaz"), sdoc("id", parentId, "name_s", "Jazz", childKey, sdoc("names_s", "Gaz"))));
+    UpdateRequestProcessor nestedUpdate = new NestedUpdateProcessorFactory().getInstance(req(), null, null);
+    AddUpdateCommand cmd = new AddUpdateCommand(req());
+    cmd.solrDoc = noIdChildren;
+    nestedUpdate.processAdd(cmd);
+    cmd.clear();
+    List children = (List) noIdChildren.get("children").getValues();
+    SolrInputDocument idLessChild = (SolrInputDocument)((SolrInputDocument) children.get(1)).get(childKey).getValue();
+    assertTrue("Id less child did not get an Id", idLessChild.containsKey("id"));
```
dsmiley: This is fine but FYI doing a doc.toString() and doing a simple assertEquals on that String is often more thorough and easy to read and less code.  That said, sometimes it can be brittle to unrelated changes.

solr/core/src/test/org/apache/solr/update/TestNestedUpdateProcessor.java
```diff
@@ -0,0 +1,171 @@
...
+    assertJQ(req("q", IndexSchema.PATH_FIELD_NAME + ":children",
+        "fl","*",
+        "sort","id asc",
+        "wt","json"),
+        childrenTests);
+  }
+
+  @Test
+  public void testDeeplyNestedURPChildrenWoId() throws Exception {
+    final String parentId = "3";
```
dsmiley: I think this doesn't test the recursive/layered nature of the ID generation.  Change the test to only have an ID at the root, then lets see that the rest of the IDs are what they ought to be?

solr/core/src/java/org/apache/solr/schema/IndexSchema.java
moshebla: Sounds good to me :+1:

solr/core/src/test/org/apache/solr/update/TestNestedUpdateProcessor.java
moshebla: Sure thing.

solr/core/src/java/org/apache/solr/update/processor/NestedUpdateProcessorFactory.java
```diff
@@ -0,0 +1,126 @@
...
+          break;
+        }
+        final String fieldName = field.getName();
+
+        if(fieldName.contains(PATH_SEP_CHAR)) {
+          throw new SolrException(SolrException.ErrorCode.BAD_REQUEST, "Field name: '" + fieldName
+              + "' contains: '" + PATH_SEP_CHAR + "' , which is reserved for the nested URP");
+        }
+        final String sChildNum = isSingleVal ? SINGLE_VALUE_CHAR: String.valueOf(childNum);
+        final String lastPath = fieldName + NUM_SEP_CHAR + sChildNum + NUM_SEP_CHAR;
```
dsmiley: I'm a bit confused why SINGLE_VALUE_CHAR was introduced.  Is this so that a DocTransformer can differentiates between a one-ary array and a single doc?  I suppose so.  The choice of 's' was confusing as I thought it was an attempt to pluralize something at first but now I think you intended an abbreviation for "singular".  Could we simply use the empty string in this case?

Also, please define lastPath & jointPath later closer to when they are used, which is immediately prior to calling processChildDoc.

Why the trailing NUM_SEP_CHAR in lastPath?

solr/core/src/java/org/apache/solr/update/processor/NestedUpdateProcessorFactory.java
```diff
@@ -0,0 +1,126 @@
...
+  }
+
+  private static boolean shouldStoreDocPath(IndexSchema schema) {
+    return schema.getFields().containsKey(IndexSchema.NEST_PATH_FIELD_NAME);
+  }
+}
+
+class NestedUpdateProcessor extends UpdateRequestProcessor {
+  private static final String PATH_SEP_CHAR = "/";
+  private static final String NUM_SEP_CHAR = ",";
```
dsmiley: It appears most any special char we use would be URL encoded so I think might as well use '#' for NUM_SEP_CHAR, which seems like the ideal choice.

solr/core/src/java/org/apache/solr/update/processor/NestedUpdateProcessorFactory.java
```diff
@@ -0,0 +1,126 @@
...
+      setPathField(sdoc, fullPath);
+    }
+    if (storeParent) {
+      setParentKey(sdoc, parent);
+    }
+    processDocChildren(sdoc, fullPath);
+  }
+
+  private String generateChildUniqueId(String parentId, String childKey, String childNum) {
+    // combines parentId with the child's key and childNum. e.g. "10/footnote/1"
```
dsmiley: the comment is obsolete again ;-)

solr/core/src/java/org/apache/solr/update/processor/NestedUpdateProcessorFactory.java
moshebla: The trailing num sep char is there so we can query all documents in the specified _NEST_PATH_: path#*#

solr/core/src/java/org/apache/solr/update/processor/NestedUpdateProcessorFactory.java
```diff
@@ -72,8 +72,8 @@ public void processAdd(AddUpdateCommand cmd) throws IOException {
   }
 
   private void processDocChildren(SolrInputDocument doc, String fullPath) {
-    int childNum = 0;
     for(SolrInputField field: doc.values()) {
+      int childNum = 0;
```
dsmiley: If the test did not show this bug, lets tweak the test to show it.  Yeah?

solr/core/src/java/org/apache/solr/update/processor/NestedUpdateProcessor.java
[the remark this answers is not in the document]
dsmiley: Yes; my use of "label" means the key of the doc, aka the pseudo-fieldname linking the parent to child.

solr/core/src/java/org/apache/solr/update/processor/NestedUpdateProcessorFactory.java
```diff
@@ -0,0 +1,126 @@
...
+        }
+        final String sChildNum = isSingleVal ? SINGULAR_VALUE_CHAR : String.valueOf(childNum);
+        SolrInputDocument cDoc = (SolrInputDocument) val;
+        if(!cDoc.containsKey(uniqueKeyFieldName)) {
+          String parentDocId = doc.getField(uniqueKeyFieldName).getFirstValue().toString();
+          cDoc.setField(uniqueKeyFieldName, generateChildUniqueId(parentDocId, fieldName, sChildNum));
+        }
+        final String lastPath = fieldName + NUM_SEP_CHAR + sChildNum + NUM_SEP_CHAR;
+        final String jointPath = fullPath == null ? lastPath : fullPath + PATH_SEP_CHAR + lastPath;
+        processChildDoc((SolrInputDocument) val, doc, jointPath);
```
dsmiley: Can you please add a comment showing what an example multi-layer (child of child of root) might look like?  Also FWIW "jointPath" as a variable name confuses me; I just don't know what the origin of that choice is.  Perhaps childDocPath?

solr/core/src/test/org/apache/solr/update/TestNestedUpdateProcessor.java
dsmiley: come to think of it, in this whole test file I don't see a test for what the `_NEST_PATH_` field is supposed to look like?  That's kinda important!  If you modify this to test via doc.toString it'd cover it.  Granted `testDeeplyNestedURPChildren()` or some new method that ought to test this as well.

solr/core/src/java/org/apache/solr/update/processor/NestedUpdateProcessorFactory.java
moshebla: I'll add a test with another key holding childDocs to ensure this bug does not resurface

solr/core/src/test/org/apache/solr/update/TestNestedUpdateProcessor.java
moshebla: I have just added a new commit with tests to ensure this :)

solr/core/src/test/org/apache/solr/update/TestNestedUpdateProcessor.java
```diff
@@ -120,25 +122,41 @@ public void before() throws Exception {
...
+    assertJQ(req("q", IndexSchema.NEST_PATH_FIELD_NAME + ":*" + PATH_SEP_CHAR + "grandChild" + NUM_SEP_CHAR + "*",
         "fl","*",
         "sort","id desc",
         "wt","json"),
-        "/response/docs/[0]/id=='" + grandChildId + "'");
+        tests);
   }
 
   @Test
   public void testDeeplyNestedURPChildren() throws Exception {
```
dsmiley: This test tests the search behavior more so than literally what the URP is doing.  Can you make this more of a unit test around the result of the URP without actually indexing/searching anything?  And I would much prefer simpler test assertions that check a complete string value instead of making reference to many variables/constants that need to be concatenated.  This makes it plainly clear what the nest path will be; no mental gymnastics are needed to chase down vars/constants to figure it out.  I've mentioned before Yonik's advise on avoiding some constants in tests as it helps tests make us aware if in the future we might have a backwards-breaking change; so there are virtues to this way of thinking.  It would make this easier to review too.

moshebla: I added a new sanity unit test for this URP

PR #411, opened 2018-06-26
Debugging PriorityQueue.java

rsaavedraf:
The change I'm proposing eliminates a problem by properly checking the validity of maxSize itself, before even computing heapSize=maxSize+1.

In the constructor, when maxSize has a value == Integer.MAX_VALUE, then heapSize = maxSize+1 ends up being negative, more exactly -2147483648 (aka. Integer.MIN_VALUE.)  This is quite a bug because then the if statement checking whether heapSize is larger than ArrayUtil.MAX_ARRAY_LENGTH ends up false, so the IllegalArgumentException is never thrown. Yet the code in PriorityQueue.java fails immediately afterwards when reaching "new Object[heapSize]" because of heapSize being negative, causing a NegativeArraySize exception.

We saw this problem with our software which was using (correction--> ) Solr 6.6.1 in cloud mode.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/util/PriorityQueue.java
+++ b/lucene/core/src/java/org/apache/lucene/util/PriorityQueue.java
@@ -50,14 +50,15 @@ public abstract class PriorityQueue<T> implements Iterable<T> {
       // We allocate 1 extra to avoid if statement in top()
       heapSize = 2;
     } else {
+
+      if ((maxSize < 0) || (maxSize >= ArrayUtil.MAX_ARRAY_LENGTH)) {
+        // Throw exception to prevent confusing OOME:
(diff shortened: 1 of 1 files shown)


mikemccand:
OK I merged this change into Lucene master & 7.x, and added a simple unit test.

Thanks @rsaavedraf!

rsaavedraf:
Thanks as well @mikemccand !
Happily tweeting about this small change being accepted :)
https://twitter.com/RaulSaavedra6/status/1012389058146861056

mikemccand:
Great, thanks @rsaavedraf!

tflobbe:
Already committed: 154952f47d8f4ae23637c42792cc0713c009bde4

PR #412, opened 2018-07-02
SOLR-12533 Collection collection fails if metrics are called during core creation

gezapeti:


gezapeti:
The change was committed. Thanks!

PR #416, opened 2018-07-12
WIP: SOLR-12519

moshebla:


solr/core/src/java/org/apache/solr/response/transform/DeeplyNestedChildDocTransformerFactory.java
```diff
@@ -0,0 +1,367 @@
...
+ * Optionally you can provide a "childFilter" param to filter out which child documents should be returned and a
+ * "limit" param which provides an option to specify the number of child documents
+ * to be returned per parent document. By default it's set to 10.
+ *
+ * Examples -
+ * [child parentFilter="fieldName:fieldValue"]
+ * [child parentFilter="fieldName:fieldValue" childFilter="fieldName:fieldValue"]
+ * [child parentFilter="fieldName:fieldValue" childFilter="fieldName:fieldValue" limit=20]
+ */
+public class DeeplyNestedChildDocTransformerFactory extends TransformerFactory {
```
dsmiley: FWIW I'm not convinced we need a distinct class from ChildDocTransformer.  I think CDT is fine... the part that changes is what follows obtaining the matching child doc iterator.   That could split out to two different methods -- plain/flat (current) and nested with ancestors (new).

solr/core/src/java/org/apache/solr/response/transform/DeeplyNestedChildTransformerBase.java
```diff
@@ -0,0 +1,139 @@
...
+    return lastPath.split(NUM_SEP_CHAR);
+  }
+
+  protected static String getSolrFieldString(Object fieldVal, FieldType fieldType) {
+    return fieldVal instanceof IndexableField
+        ? fieldType.toExternal((IndexableField)fieldVal)
+        : fieldVal.toString();
+  }
+
+  protected static class NullFilteringArrayList<T> extends ArrayList<T> {
```
dsmiley: Um; this ArrayList subclass seems undesirable to me; I'm not sure yet why it's used (though I'm sure you have your reasons)... but maybe there could be some change to the logic to avoid the need for this thing.

solr/core/src/java/org/apache/solr/response/transform/DeeplyNestedChildDocTransformerFactory.java
```diff
@@ -0,0 +1,367 @@
...
+
+  public DeeplyNestedFilterChildDocTransformer( String name, final BitSetProducer parentsFilter,
+                              final SolrQueryRequest req, final Query childFilterQuery, Pattern pathPattern, int limit) {
+    super(name, parentsFilter, req, limit);
+    this.childFilterQuery = childFilterQuery;
+    this.nestPathMatcher = pathPattern;
+  }
+
+  @Override
+  public void transform(SolrDocument rootDoc, int docid) {
```
dsmiley: This is a going to be a real challenge to come up with something that is understandable (doesn't appear *too* complicated).  Right now it appears too complicated to me; it's hard to review.  Tonight I think I'll propose some pseudocode on how to structure it.

solr/core/src/java/org/apache/solr/response/transform/DeeplyNestedChildDocTransformerFactory.java
moshebla: Sure thing, I will merge the two transformers

solr/core/src/java/org/apache/solr/response/transform/DeeplyNestedChildTransformerBase.java
moshebla: I used this since the paths in the documents are stored as array indexes. In case some documents from the array were filtered, I insert null values, so the child docs that matched the filter will be in the same array index as they were originally. Inside the iterator method, the null values are filtered, so they don't get written by the response writer.

solr/core/src/java/org/apache/solr/response/transform/DeeplyNestedChildDocTransformerFactory.java
moshebla: Sure thing, any help would be welcome.
This is a pretty rough draft, so I will try to work on this ASAP before I add more logic to the transformer.

dsmiley:
Before I look further I want to mention 2 things:
* There needn't be separation between the non-nested and nested algorithm.  We could simply use the nested algorithm but special-case when the SortedDocValues we get is null due to non-existent path.  In that event, all docs get added to the root, anonymously (unlabelled).  I suspect doing it this way is less code and will be sufficiently clear but we'll see?
* My pseudocode had a TODO about double-resolving the path in order to get the label at the time the docs are added to the parent.  If we make the pending data structure a Multimap<Integer,Multimap<String,SolrDocument>> then at the time we read the document and get the path we could store the relationship at this time.  Or perhaps you have a better idea.  I thought of putting the label into a temporary field of the SolrDocument; though I'm not sure I like that any better; maybe less.

moshebla:
In my implementation I have used a Multimap<String,SolrDocument> for the pending data, so we can decide if we put the number in the array of the parent's key. That way, There can be an option to add all parents if the child documents parent is inside a list, or only the child doc's direct parent. This can be done by the way we store the pending child doc's key, with or with out the array index.

dsmiley:
When you refer to the "array index", what do you mean?  Do you mean a DocValue ord?  Do you mean the '#2' or whatever child index?  AFAIK this code shouldn't care about child index.

moshebla:
I was referring to the '#2'.
Imagine a query for a social network, where all the posts comments made by a specific user are required, including the ones he commented on, which would not ,atch the ChildFilter commnent.author:Joe. To display these comments we would need to bring all the hierarchy above.
In another scenario one might want to get statistics of how many comments, including replies Joe commented on each post. This can be done using just the Path, excluding the array index, so isAncestor returns true.

moshebla:
I will push another commit soon.

dsmiley:
> Imagine a query for a social network, where all the posts comments made by a specific user are required, including the ones he commented on, which would not match the ChildFilter comment.author:Joe.

Assuming a comment on another comment is a parent/child relationship (i.e. the recursive comments are ancestors), wouldn't our doc transformer return those ancestors?  We're expressly writing it to return those ancestors so it would.  The ancestor path ord IDs will refer to paths distinguishing the comment occurrences, e.g. might resolve to a path looking something like `comment#3/comment#9/comment#1/` and thus we don't mix up which comment is on which.

dsmiley:
I think this issue should only be about returning a nested hierarchy.  That at least seems clear -- something everyone would want.  But computing stats... wow thats a large scope increase that deserves its own issue.  Lets table that.

moshebla:
Sure thing it won't mix up the comments, but what if the user also requests all the comments which in the same thread, thus being in the same array just a path above?

dsmiley:
> what if the user also requests all the comments which in the same thread, thus being in the same array just a path above?

Sounds like a distinct issue.  It'd add some complexity... for example the very first child doc ID might be the 2nd child of its parent.  We'd need to grab the first doc of the parent, which would have a docID earlier than we have even seen.

moshebla:
OK, We'll scratch that for now, and discuss this in a separate issue.

moshebla:
I found this nasty bug in [AtomicUpdateDocumentMerger#isAtomicUpdate](https://github.com/apache/lucene-solr/pull/416/commits/7cd1dbc3208746541bd80f2d2656b6464fecd23d#diff-50d2f2a4580a649dc2a08749061504a0R78), where single child docs(non array) were mistaken for an atomicUpdate.

dsmiley:
RE isAtomicUpdate:  Good catch!  Please add a test to call this; perhaps it's just one additional line to an existing test.

moshebla:
I added a new test [TestDeeplyNestedChildDocTransformer#testSingularChildFilterJSON](https://github.com/apache/lucene-solr/pull/416/commits/df77f5b9cd32fdd2d607d8e705481e8f17544ae7#diff-f10d284d1a6b88916e43de7c01f46bc1R113) to test this.

solr/core/src/java/org/apache/solr/response/transform/DeeplyNestedChildDocTransformer.java
```diff
@@ -0,0 +1,214 @@
...
+    } catch (IOException e) {
+      rootDoc.put(getName(), "Could not fetch child Documents");
+    }
+  }
+
+  void addChildToParent(SolrDocument parent, SolrDocument child, String label) {
+    // lookup leaf key for these children using path
+    // depending on the label, add to the parent at the right key/label
+    // TODO: unfortunately this is the 2nd time we grab the paths for these docs. resolve how?
+    String trimmedPath = trimSuffixFromPaths(getLastPath(label));
```
dsmiley: I'm a bit confused here.  Firstly, it's curious to see getLastPath(label) -- since label should not have a path if it is a label.  A label is a plain string key/label joining a parent to child like "comment".  Secondly, why the trimSuffixFromPaths... I don't get the point.  As I've said, comments of input/output sample values can help a ton in knowing what's going on.

solr/core/src/test-files/solr/collection1/conf/schema15.xml
```diff
@@ -567,7 +567,17 @@
   <field name="_root_" type="string" indexed="true" stored="true"/>
   <!-- required for NestedUpdateProcessor -->
   <field name="_nest_parent_" type="string" indexed="true" stored="true"/>
-  <field name="_nest_path_" type="string" indexed="true" stored="true"/>
+  <field name="_nest_path_" type="descendants_path" indexed="true" multiValued="false" docValues="true" stored="false" useDocValuesAsStored="true"/>
```
dsmiley: I think useDocValuesAsStored=false.  This is internal and shouldn't be returned when someone does `fl=*`

solr/core/src/test/org/apache/solr/response/transform/TestDeeplyNestedChildDocTransformer.java
```diff
@@ -0,0 +1,163 @@
...
+  @After
+  public void after() throws Exception {
+    assertU(delQ("*:*"));
+    assertU(commit());
+  }
+
+  @Test
+  public void testParentFilterJSON() throws Exception {
+    indexSampleData(10);
+    String[] tests = new String[] {
```
dsmiley: As I've suggested in a previous issue, I think it's simpler and more complete to test by evaluating the entire result as a string compared to an expected string.  Otherwise, as a reviewer I'm asking myself what assertions would be useful for this query that you didn't think of.  I think this test philosophy here is especially valuable because fundamentally you're building this interesting structured response.  Essentially everything in the response is pertinent to the test -- contains very little immaterial noise.

solr/core/src/java/org/apache/solr/response/transform/DeeplyNestedChildDocTransformer.java
```diff
@@ -0,0 +1,214 @@
...
+
+          // Is this doc a direct ancestor of another doc we've seen?
+          boolean isAncestor = pendingParentPathsToChildren.containsKey(fullDocPath);
+
+          // Do we need to do anything with this doc (either ancestor or a matched the child query)
+          if (isAncestor || children.exists(docId)) {
+            // load the doc
+            SolrDocument doc = DocsStreamer.convertLuceneDocToSolrDoc(docFetcher.doc(docId),
+                schema, new SolrReturnFields());
+            doc.setField(NEST_PATH_FIELD_NAME, fullDocPath);
```
dsmiley: This approach kinda requires we remove such internal fields later.

solr/core/src/java/org/apache/solr/response/transform/DeeplyNestedChildDocTransformer.java
```diff
@@ -0,0 +1,214 @@
...
+      }
+    } catch (IOException e) {
+      rootDoc.put(getName(), "Could not fetch child Documents");
+    }
+  }
+
+  void addChildToParent(SolrDocument parent, SolrDocument child, String label) {
+    // lookup leaf key for these children using path
+    // depending on the label, add to the parent at the right key/label
+    // TODO: unfortunately this is the 2nd time we grab the paths for these docs. resolve how?
```
dsmiley: I think this TODO is not an issue based on your placement of the path in the doc.

solr/core/src/test-files/solr/collection1/conf/schema15.xml
moshebla: Sure thing, will fix ASAP

solr/core/src/java/org/apache/solr/response/transform/DeeplyNestedChildDocTransformer.java
```diff
@@ -0,0 +1,214 @@
...
+    String trimmedPath = trimSuffixFromPaths(getLastPath(label));
+    if (!parent.containsKey(trimmedPath) && (label.contains(NUM_SEP_CHAR) && !label.endsWith(NUM_SEP_CHAR))) {
+      List<SolrDocument> list = new ArrayList<>();
+      parent.setField(trimmedPath, list);
+    }
+    parent.addField(trimmedPath, child);
+  }
+
+  void addChildToParent(SolrDocument parent, SolrDocument child) {
+    String docPath = getSolrFieldString(child.getFirstValue(NEST_PATH_FIELD_NAME), schema.getFieldType(NEST_PATH_FIELD_NAME));
```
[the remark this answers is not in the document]
moshebla: Makes sense, I actually changed it so we don't fetch the path twice, but I guess you're right

solr/core/src/java/org/apache/solr/response/transform/ChildDocTransformerFactory.java
```diff
@@ -70,7 +78,8 @@ public DocTransformer create(String field, SolrParams params, SolrQueryRequest r
     }
 
     String parentFilter = params.get( "parentFilter" );
-    if( parentFilter == null ) {
+    boolean buildHierarchy = params.getBool("hierarchy", false);
+    if( parentFilter == null && !buildHierarchy) {
```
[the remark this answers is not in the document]
moshebla: Sure, if we want this change to apply to the regular ChildDocumentTransformer.

solr/core/src/java/org/apache/solr/response/transform/DeeplyNestedChildDocTransformer.java
moshebla: We don't want the # and trailing number to be added to the fieldName represented in the document hierarchy, so trimSuffixFromPath is used.

solr/core/src/java/org/apache/solr/response/transform/ChildDocTransformerFactory.java
```diff
@@ -61,6 +65,10 @@
  */
 public class ChildDocTransformerFactory extends TransformerFactory {
 
+  public static final String PATH_SEP_CHAR = "/";
+  public static final String NUM_SEP_CHAR = "#";
+  private static final String sRootFilter = "*:* NOT " + NEST_PATH_FIELD_NAME + ":*";
```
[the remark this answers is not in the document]
dsmiley: Oh I forgot.  Perhaps instead create the Lucene Query directly (rather than create a String that needs to be parsed). Use a BooleanQuery with a MUST_NOT clause around DocValuesFieldsExistsQuery for the nest path field?   And comment this only applies to schemas that are enabled for this enhanced nesting.

solr/core/src/java/org/apache/solr/response/transform/ChildDocTransformerFactory.java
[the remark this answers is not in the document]
dsmiley: if sRootFilter needs to depend on a nest field, and if nest fields aren't required, I guess we can't relax this constraint.

solr/core/src/java/org/apache/solr/response/transform/DeeplyNestedChildDocTransformer.java
dsmiley: Ok.  So I propose renaming the "label" variable to be "nestPath" since that's what it seems to be.  getLastPath is fine.  trimSuffixFromPaths should probable be renamed to trimLastChildIndex.  Then, "trimmedPath" variable would in fact be the "label" so should be named as such.

solr/core/src/java/org/apache/solr/response/transform/DeeplyNestedChildDocTransformer.java
dsmiley: It's probably easiest to remove it later, probably at the point we add children to document.  Or try the Multimap<String<MultiMap<String,SolrDocument>> idea I had, wherein the intermediate String is the label.

moshebla: Perhaps we can Store everything in a Multimap<String,Pair<String,SolrDocument>>, thus saving each document bonded to its full path?

dsmiley: A Pair would be fine too; it's only slightly more bulky than an intermediate Multimap.

moshebla: It seems like a MultiMap<String, Pair<String,Pair>> is even more efficient, since each path is unique, while the inner MultiMap will store each item in each child label as another array.

dsmiley: I expected the intermediate string is a *label* like “comment” not a path. A label will be shared by all children under that label.

dsmiley: Oh, ok.  Perhaps have a separate standalone Set of complete paths of "single child" elements.  that tracks parent path to child labels in which that label contains a unique entry.  WDYT?

dsmiley: To be clear, I just mean a `Set<String>` of paths like `foo#1/bar#9/author#`  (author is single valued)

moshebla:
I tried optimizing the lookup and insertion of child documents.
Hopefully I'll get more time tomorrow to get the tests up to scratch.

solr/core/src/java/org/apache/solr/response/transform/DeeplyNestedChildDocTransformer.java
```diff
@@ -108,12 +103,11 @@ public void transform(SolrDocument rootDoc, int rootDocId) {
       final LeafReaderContext leafReaderContext = searcher.getIndexReader().leaves().get(seg);
       final SortedDocValues segPathDocValues = DocValues.getSorted(leafReaderContext.reader(), NEST_PATH_FIELD_NAME);
 
-      Multimap<String,SolrDocument> pendingParentPathsToChildren = ArrayListMultimap.create();
+      Map<String, Multimap<String, SolrDocument>> pendingParentPathsToChildren = new HashMap<>();
```
dsmiley: Ah good catch; yes a Map and not MultiMap at outer level once we add the label intermediate

solr/core/src/java/org/apache/solr/response/transform/DeeplyNestedChildDocTransformer.java
```diff
@@ -132,54 +126,49 @@ public void transform(SolrDocument rootDoc, int rootDocId) {
...
-            // if this path has pending child docs, add them.
-            if (isAncestor) {
-              addChildrenToParent(doc, pendingParentPathsToChildren.get(fullDocPath));
-              pendingParentPathsToChildren.removeAll(fullDocPath); // no longer pending
+            if(isAncestor) {
+              // if this path has pending child docs, add them.
+              addChildrenToParent(doc, pendingParentPathsToChildren.remove(fullDocPath)); // no longer pending
             }
+            pendingParentPathsToChildren.computeIfAbsent(parentDocPath, x -> ArrayListMultimap.create())
+                .put(trimIfSingleDoc(getLastPath(fullDocPath)), doc); // multimap add (won't replace)
```
dsmiley: I believe we simply want the label here; maybe add a method getLeafLabel() that takes a path and returns the leaf label.  It would trim off any "#" with digits no matter if it's a single doc or not.

solr/core/src/java/org/apache/solr/response/transform/DeeplyNestedChildDocTransformer.java
```diff
@@ -132,54 +126,49 @@ public void transform(SolrDocument rootDoc, int rootDocId) {
...
-  void addChildToParent(SolrDocument parent, SolrDocument child, String label) {
-    // lookup leaf key for these children using path
-    // depending on the label, add to the parent at the right key/label
-    // TODO: unfortunately this is the 2nd time we grab the paths for these docs. resolve how?
-    String trimmedPath = trimSuffixFromPaths(getLastPath(label));
-    if (!parent.containsKey(trimmedPath) && (label.contains(NUM_SEP_CHAR) && !label.endsWith(NUM_SEP_CHAR))) {
-      List<SolrDocument> list = new ArrayList<>();
-      parent.setField(trimmedPath, list);
+  void addChildrenToParent(SolrDocument parent, Multimap<String, SolrDocument> children) {
+    for(String childLabel: children.keySet()) {
```
dsmiley: there is likely a way to iterate over Map.Entry or similar so you don't have to "get" each in the next line.

solr/core/src/java/org/apache/solr/response/transform/DeeplyNestedChildDocTransformer.java
```diff
@@ -190,8 +179,18 @@ private String getLastPath(String path) {
     return path.substring(path.lastIndexOf(PATH_SEP_CHAR.charAt(0)) + 1);
   }
 
-  private String trimSuffixFromPaths(String path) {
-    return path.replaceAll("#\\d|#", "");
+  private String trimIfSingleDoc(String path) {
```
dsmiley: I'll assume this is still WIP as we discussed wanting to use a Set<String> of child docs, and thus we wouldn't trim conditionally for single doc.

moshebla: Do we really need a set though?
If we trim only the array docs we can then easily figure out at insert time whether the item is a single doc or an array, optimizing the insert time, and removing the need for a set

dsmiley: Oh gotcha -- sure.  It does complicate an explanation of what that String is, as it won't simply be a label, but that's okay.  Please add comments where we declare the field to explain this stuff. (e.g.. what the outer String is (with example), what the intermediate string is (with 2 examples)).

solr/core/src/java/org/apache/solr/response/transform/DeeplyNestedChildDocTransformer.java
moshebla: it seems like MultiMap entries are not returned as the collections for each key, but are instead returned as entries with each key added to it. It seems more efficient to add the whole array at once, reducing the amount of lookups that have to be made for each child. We could do this in one simple look up.

solr/core/src/java/org/apache/solr/response/transform/ChildDocTransformerFactory.java
```diff
@@ -67,7 +73,9 @@
 
   public static final String PATH_SEP_CHAR = "/";
   public static final String NUM_SEP_CHAR = "#";
-  private static final String sRootFilter = "*:* NOT " + NEST_PATH_FIELD_NAME + ":*";
+  private static final BooleanQuery rootFilter = new BooleanQuery.Builder()
+      .add(new BooleanClause(new MatchAllDocsQuery(), BooleanClause.Occur.MUST))
+      .add(new BooleanClause(new WildcardQuery(new Term(NEST_PATH_FIELD_NAME, new BytesRef("*"))), BooleanClause.Occur.MUST_NOT)).build();
```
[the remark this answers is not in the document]
moshebla: My bad, I'll fix this in the next commit.

solr/core/src/java/org/apache/solr/response/transform/DeeplyNestedChildDocTransformer.java
dsmiley: Ah, I see (I didn't look at Multimap's iteration options when I wrote that).  Your code here is good.

solr/core/src/test/org/apache/solr/response/transform/TestDeeplyNestedChildDocTransformer.java
moshebla: I have just updated this test, hopefully it is a lot better now.

solr/core/src/test/org/apache/solr/response/transform/TestDeeplyNestedChildDocTransformer.java
```diff
@@ -0,0 +1,227 @@
...
+    assertU(commit());
+  }
+
+  private static String id() {
+    return "" + counter.incrementAndGet();
+  }
+
+  private static String grandChildDocTemplate(int id) {
+    int docNum = id / 8; // the index of docs sent to solr in the AddUpdateCommand. e.g. first doc is 0
+    return "SolrDocument{id=stored,indexed,tokenized,omitNorms,indexOptions=DOCS<id:" + id + ">, type_s=[stored,indexed,tokenized,omitNorms,indexOptions=DOCS<type_s:" + types[docNum % types.length] + ">], name_s=[stored,indexed,tokenized,omitNorms,indexOptions=DOCS<name_s:" + names[docNum % names.length] + ">], " +
```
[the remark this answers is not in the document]
dsmiley: Another option, perhaps easiest, is to pre-process to Solr document before calling toString.  This could remove "noise" fields like root.  If an IndexableField is there then replace with stringValue().

solr/core/src/java/org/apache/solr/response/transform/DeeplyNestedChildDocTransformer.java
```diff
@@ -0,0 +1,224 @@
...
+import org.apache.solr.search.DocList;
+import org.apache.solr.search.SolrDocumentFetcher;
+import org.apache.solr.search.SolrIndexSearcher;
+import org.apache.solr.search.SolrReturnFields;
+
+import static org.apache.solr.response.transform.ChildDocTransformerFactory.NUM_SEP_CHAR;
+import static org.apache.solr.response.transform.ChildDocTransformerFactory.PATH_SEP_CHAR;
+import static org.apache.solr.schema.IndexSchema.NEST_PATH_FIELD_NAME;
+
+class DeeplyNestedChildDocTransformer extends DocTransformer {
```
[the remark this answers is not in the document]
moshebla: Should I just delete the old transformer and make this one the new ChildDocTransformer?
Or should I give it a new name like ChildDocHierarchyTransformer?

solr/core/src/java/org/apache/solr/response/transform/DeeplyNestedChildDocTransformer.java
```diff
@@ -0,0 +1,224 @@
...
+   */
+  String lookupParentPath(String currDocPath) {
+    // chop off leaf (after last '/')
+    // if child of leaf then return null (special value)
+    int lastPathIndex = currDocPath.lastIndexOf(PATH_SEP_CHAR);
+    return lastPathIndex == -1 ? null: currDocPath.substring(0, lastPathIndex);
+  }
+
+  private String getPathByDocId(int segDocId, SortedDocValues segPathDocValues) throws IOException {
+    int numToAdvance = segPathDocValues.docID()==-1?segDocId: segDocId - (segPathDocValues.docID());
```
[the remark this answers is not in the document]
moshebla: Yes,
because if we have not set the docId (the method returns -1), the iterator will advance by one doc too many.

solr/core/src/java/org/apache/solr/response/transform/DeeplyNestedChildDocTransformer.java
```diff
@@ -0,0 +1,224 @@
...
+
+  @Override
+  public String getName()  {
+    return name;
+  }
+
+  @Override
+  public String[] getExtraRequestFields() {
+    // we always need the idField (of the parent) in order to fill out it's children
+    return new String[] { idField.getName() };
```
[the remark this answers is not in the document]
moshebla: FieldType idFt = idField.getType();

    String rootIdExt = getSolrFieldString(rootDoc.getFirstValue(idField.getName()), idFt);

Doesn't this mean we need the root document's ID?

solr/core/src/test/org/apache/solr/response/transform/TestDeeplyNestedChildDocTransformer.java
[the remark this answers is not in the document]
moshebla: Would this method go in the TestUtils or the specific test class?

solr/core/src/java/org/apache/solr/response/transform/DeeplyNestedChildDocTransformer.java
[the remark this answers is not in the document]
moshebla: Perhaps since _root_ is added to every document, we could use that field instead of the ID field?
This is just a thought that popped into my head.

solr/core/src/java/org/apache/solr/response/transform/DeeplyNestedChildDocTransformer.java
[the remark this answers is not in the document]
dsmiley: Replace/update ChildDocTransformer, I think.

solr/core/src/java/org/apache/solr/response/transform/DeeplyNestedChildDocTransformer.java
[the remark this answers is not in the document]
dsmiley: Oh right; I forgot we needed an ID to do the child doc query limited by this parent ID.  Please add a comment.  I don't think we should bother with root.  I guess ChildDocTransformer might break if the id field is not stored but did have docValues.  That's a shame; it deserves a TODO.

solr/core/src/test/org/apache/solr/response/transform/TestDeeplyNestedChildDocTransformer.java
```diff
@@ -168,35 +172,57 @@ private static String id() {
...
-        "_root_=stored,indexed,tokenized,omitNorms,indexOptions=DOCS<_root_:" + id + ">, " +
-        "ingredients=[SolrDocument{id=stored,indexed,tokenized,omitNorms,indexOptions=DOCS<id:" + (id + 6) + ">, name_s=[stored,indexed,tokenized,omitNorms,indexOptions=DOCS<name_s:cocoa>], _nest_parent_=stored,indexed,tokenized,omitNorms,indexOptions=DOCS<_nest_parent_:" + (id + 5)+ ">, " +
-        "_root_=stored,indexed,tokenized,omitNorms,indexOptions=DOCS<_root_:" + id + ">}, " +
-        "SolrDocument{id=stored,indexed,tokenized,omitNorms,indexOptions=DOCS<id:" + (id + 7) + ">, name_s=[stored,indexed,tokenized,omitNorms,indexOptions=DOCS<name_s:cocoa>], _nest_parent_=stored,indexed,tokenized,omitNorms,indexOptions=DOCS<_nest_parent_:" + (id + 5) + ">, " +
-        "_root_=stored,indexed,tokenized,omitNorms,indexOptions=DOCS<_root_:" + id + ">}]}]}";
+    return "SolrDocument{id="+ id + ", type_s=[" + types[docNum % types.length] + "], name_s=[" + names[docNum % names.length] + "], " +
```
dsmiley: It's a shame to have all this embedded ID calculation business. During cleaning can they be removed (both "id" and nest parent and root) and we still have enough distinguishing characteristics of the docs to know which is which?  Seems that way.  It adds a lot of noise.

solr/core/src/java/org/apache/solr/response/transform/DeeplyNestedChildDocTransformer.java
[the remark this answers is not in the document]
moshebla: Would the transformer need to support the old method of adding childDocuments to the _childDocuments_ field?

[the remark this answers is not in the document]
moshebla: ?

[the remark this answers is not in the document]
dsmiley: Yes; it can do both easily enough I think?  A separate method could take over for the existing/legacy case.

solr/core/src/test/org/apache/solr/response/transform/TestDeeplyNestedChildDocTransformer.java
dsmiley: Keeping one ID is fine; we certainly don't need additional ones.  Maybe consider using letters or names for IDs instead of incrementing counters.  Anything to help make reading a doc/child structure more readily apparent.  Anything to reduce string interpolation here is also a win IMO.

solr/core/src/test/org/apache/solr/response/transform/TestChildDocTransformer.java
```diff
@@ -66,34 +66,34 @@ public void testAllParams() throws Exception {
   private void testChildDoctransformerXML() {
     String test1[] = new String[] {
         "//*[@numFound='1']",
-        "/response/result/doc[1]/doc[1]/str[@name='id']='2'" ,
-        "/response/result/doc[1]/doc[2]/str[@name='id']='3'" ,
-        "/response/result/doc[1]/doc[3]/str[@name='id']='4'" ,
-        "/response/result/doc[1]/doc[4]/str[@name='id']='5'" ,
-        "/response/result/doc[1]/doc[5]/str[@name='id']='6'" ,
-        "/response/result/doc[1]/doc[6]/str[@name='id']='7'"};
+        "/response/result/doc[1]/arr[@name='_childDocuments_']/doc[1]/str[@name='id']='2'" ,
```
moshebla: One major difference is the way nested XML is returned, since now it has a key it belongs to.
Thought it would make the review process easier for you if I pointed these differences out.

solr/core/src/test/org/apache/solr/response/transform/TestChildDocTransformer.java
```diff
@@ -242,10 +242,10 @@ private void testChildDocNonStoredDVFields() throws Exception {
         "fl", "*,[child parentFilter=\"subject:parentDocument\"]"), test1);
 
     assertJQ(req("q", "*:*", "fq", "subject:\"parentDocument\" ",
-        "fl", "subject,[child parentFilter=\"subject:parentDocument\" childFilter=\"title:foo\"]"), test2);
+        "fl", "id,_childDocuments_,subject,intDvoDefault,[child parentFilter=\"subject:parentDocument\" childFilter=\"title:foo\"]"), test2);
```
moshebla: Another major difference is that now since child documents are stored as fields, the user needs to explicitly add them to the list of return fields. I have no opinion regarding this issue, but this might pose a problem to some use cases. Perhaps documenting this would be enough?

solr/core/src/test/org/apache/solr/response/transform/TestChildDocTransformerHierarchy.java
```diff
@@ -32,7 +32,7 @@
 import org.junit.BeforeClass;
 import org.junit.Test;
 
-public class TestDeeplyNestedChildDocTransformer extends SolrTestCaseJ4 {
+public class TestChildDocTransformerHierarchy extends SolrTestCaseJ4 {
```
moshebla: These "hierarchy" tests are in a separate class since they require a different schema and configuration to be loaded upon start-up.

solr/core/src/test/org/apache/solr/response/transform/TestChildDocTransformer.java
dsmiley: ah; interesting.  It's logical.  Is this only needed for anonymous child docs (thus \_childDocuments\_ or any/all possible relationship names that aren't necessarily just at the root level but anywhere in the hierarchy?  Perhaps this is where that "anonChildDocs" ought to come into play again for backwards-compatibility sake?  Well perhaps not... someone who is using anonymous child docs today will not have the nested field metadata and thus the old logic will kick in and ensure child documents are added as it was; right?

moshebla: This is what I would expect would happen. I am going to try it to check this works as expected. Would not want to suddenly change this for everyone :worried:

moshebla: I had to add a small addition and indeed, the behaviour was as expected.
I added a test to ensure there is backwards compatibility with the old XML format [here](https://github.com/apache/lucene-solr/blob/2573b89e465de615623084145cab7d17e9cb8a07/solr/core/src/test/org/apache/solr/response/transform/TestChildDocTransformer.java#L100).

moshebla: I was thinking about what you said, and perhaps we could add the old format using if the user did not index the field meta-data. I prefer having a flag, since the user can query separate two collections and get the same output if desired, instead of having to deal with two different XML formats.

dsmiley: I think either the user hasn't yet started using the new key'ed/labeled style of child documents, or they have updated completely.  It's a migration to a new way you either do or don't do (and perhaps one day will not have a choice).

moshebla: Of course, one day this will probably get removed. Right now there is a way to provide the old XML format.

moshebla:
I hope this is in a good enough shape for review.
There was an overlap between this ticket and the XMLLoader, so I hope I did not miss anything.

dsmiley:
This is pretty close; it has come far.  I've applied the patch and started to make some manipulations.  I just have some questions, which I'll ask inline.  Please don't push new changes as it'd be hard to integrate.  I could either post a patch or... maybe I should push a feature branch to my fork on GH which would display diffs from yours.  Hmm; does GitHub allow you to show a diff between your feature branch and another feature branch on another fork?

solr/core/src/test/org/apache/solr/update/TestNestedUpdateProcessor.java
```diff
@@ -107,8 +107,8 @@ public void testDeeplyNestedURPGrandChild() throws Exception {
     };
     indexSampleData(jDoc);
 
-    assertJQ(req("q", IndexSchema.NEST_PATH_FIELD_NAME + ":*/grandChild#*",
-        "fl","*",
+    assertJQ(req("q", IndexSchema.NEST_PATH_FIELD_NAME + ":*/grandChild",
+        "fl","*, _nest_path_",
```
dsmiley: This change (and others here) to a test of an URP that isn't modified in this issue underscores my point made previously having the test for that URP be more of a unit test of what the URP produces (test the SolrInputDocument), and _not_ executing queries.  I'm not saying it was wrong to make this change in this issue but just want you to reflect on the ramifications of these choices.

solr/core/src/test/org/apache/solr/response/transform/TestChildDocTransformerHierarchy.java
```diff
@@ -0,0 +1,249 @@
...
+import org.apache.lucene.index.IndexableField;
+import org.apache.solr.SolrTestCaseJ4;
+import org.apache.solr.common.SolrDocument;
+import org.apache.solr.request.SolrQueryRequest;
+import org.apache.solr.response.BasicResultContext;
+import org.junit.After;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+public class TestChildDocTransformerHierarchy extends SolrTestCaseJ4 {
```
dsmiley: I'm glad you have split out this test.  I want to further create a new "schema-nest.xml" for our tests that explicitly deal with this new nested document stuff.  This test will use it, as well as TestNestedUpdateProcessor.  schema15.xml can be reverted to as it was.  TestChildDocTransformer can continue to use the legacy schema and as-such we can observe that our additions to the underlying code don't disturb anyone who is using it that is unaware of the new nested stuff.

solr/core/src/test/org/apache/solr/response/transform/TestChildDocTransformer.java
dsmiley: I'm really not a fan of {{anonChildDocs}} flag; I regret I conjured up the idea.  If we have "nest" schema fields, the user wants nested documents (including field/label association), if the schema doesn't it ought to work as it used to.  I think this is straight-forward to reason about and document.

solr/core/src/java/org/apache/solr/response/transform/ChildDocTransformerFactory.java
```diff
@@ -70,109 +73,59 @@ public DocTransformer create(String field, SolrParams params, SolrQueryRequest r
...
-            docFetcher.decorateDocValueFields(solrChildDoc, childDocNum, dvFieldsToReturn);
-          }
-          doc.addChildDocument(solrChildDoc);
-        }
-      }
-      
-    } catch (IOException e) {
-      doc.put(name, "Could not fetch child Documents");
+  protected static String buildHierarchyChildFilterString(String queryString) {
+    List<String> split = StrUtils.splitSmart(queryString, ':');
```
dsmiley: FYI I've rewritten this method to not use splitting & joining in cases like this which can use "indexOf"/"lastIndexOf".   I've also increased it's robustness to more complicated query examples of multiple conditions.  

For now I think we shouldn't document this; let it be kinda a secret feature until we can query (in q/fq) in like-kind.

solr/core/src/java/org/apache/solr/response/transform/ChildDocTransformer.java
```diff
@@ -0,0 +1,244 @@
...
+    // chop off leaf (after last '/')
+    // if child of leaf then return null (special value)
+    int lastPathIndex = currDocPath.lastIndexOf(PATH_SEP_CHAR);
+    return lastPathIndex == -1 ? null: currDocPath.substring(0, lastPathIndex);
+  }
+
+  private static String getPathByDocId(int segDocId, SortedDocValues segPathDocValues) throws IOException {
+    int numToAdvance = segPathDocValues.docID()==-1?segDocId: segDocId - (segPathDocValues.docID());
+    assert numToAdvance >= 0;
+    assert segPathDocValues.advanceExact(segDocId);
```
dsmiley: Woah; we want to call advanceExact even if assertions are disabled -- like, you know, in production/real-world; not just tests.  Thankfully, Lucene/Solr's randomized test infrastructure actually randomly disables assertions and thus would catch this (eventually) if I didn't.

solr/core/src/java/org/apache/solr/response/transform/ChildDocTransformerFactory.java
```diff
@@ -61,6 +57,13 @@
  */
 public class ChildDocTransformerFactory extends TransformerFactory {
 
+  public static final String PATH_SEP_CHAR = "/";
```
dsmiley: I don't see why these are declared as Strings and not `char`.  It's clumsy later to do `charAt(0)`

solr/core/src/test-files/solr/collection1/conf/schema15.xml
```diff
@@ -567,7 +567,17 @@
   <field name="_root_" type="string" indexed="true" stored="true"/>
   <!-- required for NestedUpdateProcessor -->
   <field name="_nest_parent_" type="string" indexed="true" stored="true"/>
-  <field name="_nest_path_" type="string" indexed="true" stored="true"/>
+  <field name="_nest_path_" type="descendants_path" indexed="true" multiValued="false" docValues="true" stored="false" useDocValuesAsStored="false"/>
+  <fieldType name="descendants_path" class="solr.SortableTextField">
+    <analyzer type="index">
+      <charFilter class="solr.PatternReplaceCharFilterFactory" pattern="(^.*.*$)" replacement="$0/"/>
```
dsmiley: Can you please provide an example input String (here in GH) that has multiple levels and comment out it looks like when it's done?  I know how to read regexps but I need to stare at them a bit to figure them out, so lets make it easier to read/review.

moshebla: Sure thing

moshebla:
It has been a while since I last worked on this,
and after the merge with your branch some comments were overwritten.
Do you remember what changes are still needed?

dsmiley:
I'm glad you added the comments to the analysis regexps.  I don't know why there are two `.*` (dot-star) in the first regexp.

In JIRA I mentioned the "limit" needs to be re-implemented.  Other than that... I don't recall anything.  It'd be good to scan over for any nocommitts or TODOs that look pressing.  I think it's probably about ready otherwise.

solr/core/src/test/org/apache/solr/response/transform/TestChildDocTransformer.java
```diff
@@ -81,8 +81,8 @@ private void testChildDoctransformerXML() {
 
     String test3[] = new String[] {
         "//*[@numFound='1']",
-        "/response/result/doc[1]/doc[1]/str[@name='id']='3'" ,
-        "/response/result/doc[1]/doc[2]/str[@name='id']='5'" };
+        "/response/result/doc[1]/doc[1]/str[@name='id']='5'" ,
```
moshebla: This had to be changed since we initialise the for loop in ChildDocTransformer using the limit param, so the sort is docID descending now.
Thought it is important to point it out, and perhaps this should also be documented.

solr/core/src/java/org/apache/solr/response/transform/ChildDocTransformer.java
```diff
@@ -0,0 +1,253 @@
...
+  }
+
+  /**
+   *
+   * @param segDocBaseId base docID of the segment
+   * @param RootId docID if the current root document
+   * @param lastDescendantId lowest docID of the root document's descendant
+   * @return the docID to loop and to not surpass limit of descendants to match specified by query
+   */
+  private int calcLimitIndex(int segDocBaseId, int RootId, int lastDescendantId) {
```
moshebla: The reason the sort is docID descending, is because this method initialises the index to the highest docID to match exhaust the limit, if possible, or the last descendant document.

solr/core/src/test/org/apache/solr/response/transform/TestChildDocTransformer.java
dsmiley: I suspect you are simply using the "sort" word inappropriately.  "sort" is related to ordering, but we didn't change the order.  We did change the "window" or "offset", if you will, into the docs to examine.  Ordering/sort hasn't changed.

moshebla: Yes,
I meant the order in which the documents are counted.
My bad.

dsmiley: Again, the order hasn't changed ;-)

moshebla: since we initialize the document using from the highest index to exhaust the limit if possible, it means some documents may be skipped. If we have 3 child docs: 1 , 2, 3, and the limit is set to "2", only doc 2 and 3 will be added if we initialise the index to the one that will exhaust the limit. Another way is simply to count the number of matching docs so far, and continue onto the next root doc if reached(return from the transform method).

moshebla: I guess I was not specific enough beforehand.

dsmiley: "If we have 3 child docs" ....   -- the explanation in that sentence is perfectly clear, I think.  The next sentence confused me, but whatever.

moshebla: The next sentence is a proposal for a different way to limit the number of child documents.
Incase the current logic is sufficient, then we have no need to worry about it.

solr/core/src/java/org/apache/solr/response/transform/ChildDocTransformer.java
```diff
@@ -224,9 +225,29 @@ private static String getPathByDocId(int segDocId, SortedDocValues segPathDocVal
...
-        : fieldVal.toString();
+  /**
+   *
+   * @param segDocBaseId base docID of the segment
+   * @param RootId docID if the current root document
+   * @param lastDescendantId lowest docID of the root document's descendant
+   * @return the docID to loop and to not surpass limit of descendants to match specified by query
+   */
+  private int calcLimitIndex(int segDocBaseId, int RootId, int lastDescendantId) {
+    int i = segDocBaseId + RootId - 1; // the child document with the highest docID
```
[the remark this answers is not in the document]
moshebla: Changed and noted :)

moshebla:
I have simplified the calcDocIdToIterateFrom method, and tests seem to pass :)

solr/core/src/java/org/apache/solr/response/transform/ChildDocTransformer.java
```diff
@@ -227,27 +225,28 @@ private static String getPathByDocId(int segDocId, SortedDocValues segPathDocVal
...
+   * @param RootDocId docID if the current root document
+   * @param lowestChildDocId lowest docID of the root document's descendant
+   * @return the docID to loop and not surpass limit of descendants to match specified by query
    */
-  private int calcLimitIndex(int segDocBaseId, int RootId, int lastDescendantId) {
-    int i = segDocBaseId + RootId - 1; // the child document with the highest docID
-    final int prevSegRootId = segDocBaseId + lastDescendantId;
-    assert prevSegRootId < i; // previous rootId should be smaller then current RootId
+  private int calcDocIdToIterateFrom(int lowestChildDocId, int RootDocId) {
+    assert lowestChildDocId < RootDocId; // first childDocId should be smaller then current RootId
```
dsmiley: if the root doc has no children, this assertion would fail, right?  Hmm; should be tested we don't blow up/fail.

moshebla: Yes it would,
I'll add a test to ensure its behaviour.

moshebla:
I added extra logic to return as soon as we are able to determine that the root doc has no child documents.

solr/core/src/java/org/apache/solr/response/transform/ChildDocTransformer.java
```diff
@@ -87,7 +87,12 @@ public void transform(SolrDocument rootDoc, int rootDocId) {
       final int segBaseId = leafReaderContext.docBase;
       final int segRootId = rootDocId - segBaseId;
       final BitSet segParentsBitSet = parentsFilter.getBitSet(leafReaderContext);
-      final int segPrevRootId = segParentsBitSet.prevSetBit(segRootId - 1); // can return -1 and that's okay
+      final int segPrevRootId = rootDocId==0? -1: segParentsBitSet.prevSetBit(segRootId - 1); // can return -1 and that's okay
+
+      if(segPrevRootId == (rootDocId - 1)) {
```
dsmiley: you are comparing a segment local ID with a global ID which is incorrect.  You should refer to segRootId.  This is why I'm particular about using "seg" nomenclature in a body of code that deals with both segment and global IDs -- it makes it at least easier to identify such an error.  It's difficult to get tests to detect this; we'd need to commit some docs up front to cause more segments to be created than many tests will do.

solr/core/src/test/org/apache/solr/response/transform/TestChildDocTransformerHierarchy.java
```diff
@@ -291,4 +302,15 @@ private static String generateDocHierarchy(int i) {
               "}\n" +
             "}";
   }
+
+  private static String IndexWoChildDocs() {
```
dsmiley: First, never start a method name with an uppercase letter (at least in Java).  But secondly, I suggest inlining this to the point of use since you're only using it in one place.  I know this is a stylistic point but I find it harder to read code that refers to a bunch of other things in other places that I have to go read -- it's disjointed; disrupts reading flow.  Of course a good deal of that is normal (calling string.whatever) but here it's only used for this test.

moshebla: Oops how embarrassing, guess I was in a hurry :(.

solr/core/src/java/org/apache/solr/response/transform/ChildDocTransformer.java
dsmiley: You altered line 90 in response to my comment.  I'm referring to line 92 -- `if(segPrevRootId == (rootDocId - 1))`, where my comment is.

Interesting though.... line 90 is different from the line of code I committed to the feature branch.  That line on the feature branch is:
`final int segPrevRootId = segParentsBitSet.prevSetBit(segRootId - 1); // can return -1 and that's okay`
Notice there is no conditional, but there is in your version in this PR.  BitSet.prevSetBit will return -1 if given a 0 input (so says it's documentation).  That's what I was trying to say in my comment on the line of code.  Why did you change it?

moshebla: If segRootId is 0, segParentsBitSet.prevSetBit(-1) throws an assertion error, since the index has to be >= 0.
I will fix line 92

dsmiley: Ooooh, good catch.

Lets enhance the tests in this file a bit to help us give confidence that we're using docIDs correctly (and help avoid future enhancers/modifiers from introducing similar bugs).  Here's what I propose:  in the @BeforeClass, if random().nextBoolean(), add some nested docs -- using one of your existing nested document adding methods.  And randomly do a commit() to flush the segment.  Later in the test methods we need to add a filter query that will exclude those docs.  One way to do this is to ensure these first docs have some field we can exclude.  Another way might be knowing the maximum uniqueKey ID you can query by prior to the test starting, and then adding a filter query with a range saying the uniqueKey must be at least this value.  Make sense?

moshebla: Yes,
I have added this conditional to the test.
Let's hope I did not misread your intentions :).

solr/core/src/test/org/apache/solr/response/transform/TestChildDocTransformerHierarchy.java
```diff
@@ -40,22 +43,52 @@
...
+      filterOtherSegments = "{!frange l=" + randomDocTopId + " incl=false}idInt";
+    } else {
+      filterOtherSegments = "*:*";
+    }
   }
 
   @After
   public void after() throws Exception {
-    clearIndex();
+    if (!useSegments) {
```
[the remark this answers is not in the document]
dsmiley: You could keep this if you want (use BeforeClass but have delete query in after()).  But if you do, it could be simplified: Don't conditionally call "clearIndex"; simply always do this delete.  And there's a simpler overloaded version -- `delQ(queryhere)`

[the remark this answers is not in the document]
moshebla: So I could simply use delQ using the filter,
great :-).

solr/core/src/test/org/apache/solr/response/transform/TestChildDocTransformerHierarchy.java
```diff
@@ -264,7 +309,7 @@ private static Object cleanIndexableField(Object field) {
   }
 
   private static String grandChildDocTemplate(int id) {
-    int docNum = id / 8; // the index of docs sent to solr in the AddUpdateCommand. e.g. first doc is 0
+    int docNum = (id / sumOfDocsPerNestedDocument) % numberOfDocsPerNestedTest; // the index of docs sent to solr in the AddUpdateCommand. e.g. first doc is 0
```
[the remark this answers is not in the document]
moshebla: the modulo is added to filter the docs in the other segments, and then calculate the i that was passed, when constructing the nested document. This then ensures the child transformer did not fail. If we can be satisfied by only testing the ids, which does not seem as bullet-proof to me, this could be removed, and only the ids will be tested.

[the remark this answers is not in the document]
moshebla: I tried simplifying the calculation a bit,  [here's a link](https://github.com/apache/lucene-solr/pull/416/files#diff-9fe0ab006f82be5c6a07d5bb6dbc6da0R299).

solr/core/src/test/org/apache/solr/response/transform/TestChildDocTransformerHierarchy.java
```diff
@@ -0,0 +1,346 @@
...
+      // create random segments
+      final int numOfDocs = 10;
+      for(int i = 0; i < numOfDocs; ++i) {
+        updateJ(generateDocHierarchy(i), params("update.chain", "nested"));
+        if(random().nextBoolean()) {
+          assertU(commit());
+        }
+      }
+      assertU(commit());
+      randomDocTopId = counter.get();
```
dsmiley: I think this line should be at the end of this method, executed always.  I can see it works where it is now, executed conditionally, but I think it'd be clearer if not done in a condition.

solr/core/src/java/org/apache/solr/response/transform/ChildDocTransformer.java
```diff
@@ -0,0 +1,257 @@
...
+      }
+
+      // we'll need this soon...
+      final SortedDocValues segPathDocValues = DocValues.getSorted(leafReaderContext.reader(), NEST_PATH_FIELD_NAME);
+
+      // the key in the Map is the document's ancestors key(one above the parent), while the key in the intermediate
+      // MultiMap is the direct child document's key(of the parent document)
+      Map<String, Multimap<String, SolrDocument>> pendingParentPathsToChildren = new HashMap<>();
+
+      IndexSchema schema = searcher.getSchema();
```
dsmiley: Master has changes recently committed related to document fetching.  It'll simplify this nicely here.  Can you please sync up?

moshebla: Will do

moshebla:
Fixed SOLRJ tests which failed limit assertion for ChildDocTransformer

solr/core/src/test/org/apache/solr/response/transform/TestChildDocTransformer.java
```diff
@@ -92,6 +97,18 @@ private void testChildDoctransformerXML() {
     assertQ(req("q", "*:*", "fq", "subject:\"parentDocument\" ",
         "fl", "subject,[child parentFilter=\"subject:parentDocument\" childFilter=\"title:foo\"]"), test2);
 
+    try(SolrQueryRequest req = req("q", "*:*", "fq", "subject:\"parentDocument\" ",
```
dsmiley: Please include a comment explaining what it is you're testing here (i.e. what is the point of this particular test); it's non-obvious to me.
This test tests manually... but it could be done more concisely using the XPath style assertions done elsewhere in this file and most tests.  For example if you want to test the number of child documents, it'd be something like this: "count(/response/result/doc[1]/doc)=2" and include a numFound check.

solr/core/src/test/org/apache/solr/response/transform/TestChildDocTransformerHierarchy.java
```diff
@@ -238,6 +238,17 @@ public void testSingularChildFilterJSON() throws Exception {
         tests);
   }
 
+  @Test
+  public void testNonRootChildren() throws Exception {
+    indexSampleData(numberOfDocsPerNestedTest);
+    assertJQ(req("q", "test_s:testing",
+        "sort", "id asc",
+        "fl", "*,[child childFilter='lonely/lonelyGrandChild/test2_s:secondTest' parentFilter='_nest_path_:\"lonely/\"']",
```
dsmiley: Aha; I see you're requiring that a custom parentFilter be present in order to use the childDocFilter on non-root docs.  I suppose that's fine; it something that could be enhanced in the future to avoid that inconvenience & limitation(*).  Since this is now required, the transformer should throw an error to the user if the doc to be transformed isn't in that filter.  For example:  "The document to be transformed must be matched by the parentFilter of the child transformer" with BAD_REQUEST flag as it's user-error.  And add a test for this.

(*) An example of it being a limitation is this:  Say the child docs are all a "comment" in nature; and thus are recursive.  The top query "q" might match certain comments of interest.  And we want all children of those comments returned hierarchically.  The parentFilter would end up having to match all comments, but that would prevent returning child comments and thus blocking the whole idea altogether :-(

To "fix" this limitation, we'd not constrain transformed docs to those in the parentFilter, and we wouldn't insist on any special parentFilter.  In the loop that builds the hierarchy, when fetching the path, we'd need to skip over the current doc if it doesn't descend from that of the doc to be transformed.  Seems pretty straight-forward.

moshebla: Perhaps I could try and remove the parentFilter requirement all together for hierarchy based queries?

solr/core/src/java/org/apache/solr/response/transform/ChildDocTransformer.java
```diff
@@ -0,0 +1,263 @@
...
+
+      SolrDocumentFetcher docFetcher = searcher.getDocFetcher();
+      final int lastChildId = segBaseId + segPrevRootId + 1;
+      // Loop each child ID up to the parent (exclusive).
+      for (int docId = calcDocIdToIterateFrom(lastChildId, rootDocId); docId < rootDocId; ++docId) {
+
+        // get the path.  (note will default to ANON_CHILD_KEY if schema is not nested or empty string if blank)
+        String fullDocPath = getPathByDocId(docId - segBaseId, segPathDocValues);
+
+        if(isNestedSchema && !fullDocPath.contains(transformedDocPath)) {
```
moshebla: Perhaps a better way to do this would be building a new Filter for every transformed doc e.g. `_nest_path_:transformedDocPath`?
I am not quite sure of the performance overhead such technique would impose,
WDYT?

dsmiley: > Perhaps a better way to do this ...

I think it would be slow and add complexity if we did that.

string.contains(...) doesn't seem right here; shouldn't it be startsWith or endsWith?  Will probably need to special-case empty transformedDocPath (rootDocPath).

I think it's possible to not need "isNestedSchema", though it's not a big deal.

solr/core/src/java/org/apache/solr/response/transform/ChildDocTransformer.java
```diff
@@ -109,9 +109,14 @@ public void transform(SolrDocument rootDoc, int rootDocId) {
       // Loop each child ID up to the parent (exclusive).
       for (int docId = calcDocIdToIterateFrom(lastChildId, rootDocId); docId < rootDocId; ++docId) {
 
-        // get the path.  (note will default to ANON_CHILD_KEY if not in schema or oddly blank)
+        // get the path.  (note will default to ANON_CHILD_KEY if schema is not nested or empty string if blank)
         String fullDocPath = getPathByDocId(docId - segBaseId, segPathDocValues);
 
+        if(isNestedSchema && !fullDocPath.contains(transformedDocPath)) {
+          // is not a descendant of the transformed doc, return fast.
+          return;
```
dsmiley: woah; shouldn't this be "continue"?  We should have a test that would fail on this bug.  All it would take would be an additional child doc that is not underneath the input root/transformed doc but follows it (as provided on input).  Some of the first docs we iterate over here might not descend from rootDocId

solr/core/src/java/org/apache/solr/response/transform/ChildDocTransformer.java
```diff
@@ -99,6 +96,9 @@ public void transform(SolrDocument rootDoc, int rootDocId) {
 
       // we'll need this soon...
       final SortedDocValues segPathDocValues = DocValues.getSorted(leafReaderContext.reader(), NEST_PATH_FIELD_NAME);
+      // passing a different SortedDocValues obj since the child documents which come after are of smaller docIDs,
+      // and the iterator can not be reversed.
+      final String transformedDocPath = getPathByDocId(segRootId, DocValues.getSorted(leafReaderContext.reader(), NEST_PATH_FIELD_NAME));
```
dsmiley: Can you call this rootDocPath since we refer to this doc as the "root doc" elsewhere here?  I can see why you chose this name.  You could add a comment that the "root doc" is the input doc we are adding information to, and is usually but not necessarily the root of the block of documents (i.e. the root doc may itself be a child doc of another doc).

solr/core/src/java/org/apache/solr/response/transform/ChildDocTransformer.java
moshebla: Yep, you're right.
I'll investigate further to see why a test did not fail because of this.

solr/core/src/java/org/apache/solr/response/transform/ChildDocTransformer.java
moshebla: Sure thing.
Done :-)

solr/core/src/java/org/apache/solr/response/transform/ChildDocTransformer.java
moshebla: Added another query to [TestChildDocumentHierarchy#testNonRootChildren](https://github.com/apache/lucene-solr/pull/416/files#diff-9fe0ab006f82be5c6a07d5bb6dbc6da0R243).
This test failed before I changed the return to continue(previous commit), and passes using the latest.

solr/core/src/java/org/apache/solr/response/transform/ChildDocTransformer.java
```diff
@@ -123,6 +124,16 @@ public void transform(SolrDocument rootDoc, int rootDocId) {
 
         // Do we need to do anything with this doc (either ancestor or matched the child query)
         if (isAncestor || childDocSet == null || childDocSet.exists(docId)) {
+
+          if(limit != -1) {
+            if(!isAncestor) {
+              if(matches == limit) {
+                continue;
+              }
+              ++matches;
```
dsmiley: I think matches should be incremented if it's in childDocSet (includes childDocSet being null).  Wether it's an ancestor or not doesn't matter I think.  You could pull out a new variable isInChildDocSet.  Or I suppose simply consider all a match; what I see what you just did as I write this; that's fine too.

solr/core/src/test/org/apache/solr/response/transform/TestChildDocTransformerHierarchy.java
```diff
@@ -124,10 +124,11 @@ public void testParentFilterLimitJSON() throws Exception {
 
     assertJQ(req("q", "type_s:donut",
         "sort", "id asc",
-        "fl", "id, type_s, toppings, _nest_path_, [child limit=1]",
+        "fl", "id, type_s, lonely, lonelyGrandChild, test_s, test2_s, _nest_path_, [child limit=1]",
```
dsmiley: To my point I wrote in JIRA:  It's sad that when I see this I have no idea if it's right/wrong without having to go look at indexSampleData then think about it.  No?  (this isn't a critique of you in particular; lots of tests including some I've written look like the current tests here).   imagine one doc with some nested docs, all of which only have their ID.  Since they only have their ID, it's not a lot of literal text in JSON.  The BeforeClass unmatched docs cold use negative IDs to easily know who's who.  Any way if you would rather leave this as a "TODO" for another day then I understand.

moshebla: Leaving this as a TODO for another day sounds like a decent option.

moshebla:
was commited in commit 5a0e7a615a9b1e7ac97c6b0f9e5604dcc1aeb03f

PR #417, opened 2018-07-12
Fixes SOLR-12550: ConcurrentUpdateSolrClient doesn't respect timeouts for commits and optimize

morissm:


solr/solrj/src/test/org/apache/solr/client/solrj/impl/ConcurrentUpdateSolrClientTest.java
```diff
@@ -346,4 +347,25 @@ public OutcomeCountingConcurrentUpdateSolrClient build() {
...
     }
   }
+
+  /**
+   * Test that connection timeout information is passed to the HttpSolrClient that handle non add operations
+   */
+  @Test(timeout = 1000)
+  public void testConnectionTimeoutOnCommit() throws IOException, SolrServerException {
+    // 240.0.0.0/4 is reserved for future use by the IANA and is in effect a black hole
+    try (ConcurrentUpdateSolrClient client = new ConcurrentUpdateSolrClient.Builder("http://240.0.0.1:8983/")
```
shalinmangar: Please use `SolrTestCaseJ4.DEAD_HOST` instead of a real networking address so that it fails fast. After SOLR-14050, the security policy for tests allows connections to loopback only.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/ConcurrentUpdateSolrClient.java
```diff
@@ -121,9 +121,11 @@ protected ConcurrentUpdateSolrClient(String solrServerUrl,
   
   protected ConcurrentUpdateSolrClient(Builder builder) {
     this.internalHttpClient = (builder.httpClient == null);
-    this.client = new HttpSolrClient.Builder(builder.baseSolrUrl)
-        .withHttpClient(builder.httpClient)
-        .build();
+    HttpSolrClient.Builder httpClientBuilder = new HttpSolrClient.Builder(builder.baseSolrUrl)
+        .withHttpClient(builder.httpClient);
+    httpClientBuilder.connectionTimeoutMillis = builder.connectionTimeoutMillis;
+    httpClientBuilder.socketTimeoutMillis = builder.socketTimeoutMillis;
```
BiM31: Maybe you should use with* methods and do the build as the previous version.

```
        .withConnectionTimeout(builder.connectionTimeoutMillis)
        .withSocketTimeout(builder.socketTimeoutMillis)
```

HoustonPutman:
So I think this is built off of a really old master. We should probably make sure these timeouts are also respected by `ConcurrentUpdateHttp2SolrClient`

solr/solrj/src/java/org/apache/solr/client/solrj/impl/ConcurrentUpdateSolrClient.java
morissm: Right. I've updated the pull request accordingly.

solr/solrj/src/test/org/apache/solr/client/solrj/impl/ConcurrentUpdateSolrClientTest.java
morissm: You are right, this test fails on current master. (This PR is quite ancient.) 

Unfortunately, DEAD_HOST does not work either. For this test to actually test the right condition, the connection needs to time out and not fail on connection. 

I have updated the pull request with another solution that works within current security policy and fails quite fast (1 ms).

morissm:
> So I think this is built off of a really old master. We should probably make sure these timeouts are also respected by `ConcurrentUpdateHttp2SolrClient`

I took a look at `ConcurrentUpdateHttp2SolrClient` while I was adressing the issues raised by the other reviewers. 

I have not conducted any tests but I don't think it would exhibit the same issue. Its builder works differently and reuses `UpdateShardHandler`'s httpClient which is correctly configured.

PR #423, opened 2018-07-19
Solr 12357 

nsoft:
Premptive Creation of collections in Time Routed Aliases

### DIFF
--- a/solr/core/src/java/org/apache/solr/update/processor/TimeRoutedAliasUpdateProcessor.java
+++ b/solr/core/src/java/org/apache/solr/update/processor/TimeRoutedAliasUpdateProcessor.java
@@ -19,17 +19,23 @@ package org.apache.solr.update.processor;
 
 import java.io.IOException;
 import java.lang.invoke.MethodHandles;
+import java.text.ParseException;
 import java.time.Instant;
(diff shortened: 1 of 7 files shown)


nsoft:
bleh, can't seem to get rid of the commit with the wrong message. Maybe this should get converted to a simple patch file...

solr/core/src/java/org/apache/solr/update/processor/TimeRoutedAliasUpdateProcessor.java
```diff
@@ -272,7 +349,8 @@ private String findTargetCollectionGivenTimestamp(Instant docTimestamp) {
         return entry.getValue(); //found it
       }
     }
-    return null; //not found
+    throw new SolrException(SolrException.ErrorCode.BAD_REQUEST,
+        "Doc " + id + " couldn't be routed with " + timeRoutedAlias.getRouteField() + "=" + docTimestamp);
```
dsmiley: I think the error should say it's too old for the oldest collection

fsparv: That would be clearer, I agree

PR #425, opened 2018-07-23
WIP SOLR-12555: refactor tests to use expectThrows

barrotsteindev:
Refactor tests to use expectThrows instead of the old try catch method.

### DIFF
--- a/solr/core/src/test/org/apache/solr/TestDistributedSearch.java
+++ b/solr/core/src/test/org/apache/solr/TestDistributedSearch.java
@@ -910,13 +910,11 @@ public class TestDistributedSearch extends BaseDistributedSearchTestCase {
 
     //SOLR 3161 ensure shards.qt=/update fails (anything but search handler really)
     // Also see TestRemoteStreaming#testQtUpdateFails()
-    try {
-      ignoreException("isShard is only acceptable");
-      // query("q","*:*","shards.qt","/update","stream.body","<delete><query>*:*</query></delete>");
-      // fail();
-    } catch (SolrException e) {
-      //expected
-    }
+
+    //SolrException e = expectThrows(SolrException.class, () -> {
+    //  ignoreException("isShard is only acceptable");
+    //  query("q","*:*","shards.qt","/update","stream.body","<delete><query>*:*</query></delete>");
+    //});
     unIgnoreException("isShard is only acceptable");
 
     // test debugging
@@ -1212,41 +1210,34 @@ public class TestDistributedSearch extends BaseDistributedSearchTestCase {
   private void validateCommonQueryParameters() throws Exception {
     ignoreException("parameter cannot be negative");
 
-    try {
+    SolrException e1 = expectThrows(SolrException.class, () -> {
       SolrQuery query = new SolrQuery();
       query.setParam("start", "non_numeric_value").setQuery("*");
       QueryResponse resp = query(query);
-      fail("Expected the last query to fail, but got response: " + resp);
-    } catch (SolrException e) {
-      assertEquals(ErrorCode.BAD_REQUEST.code, e.code());
-    }
+    });
+    assertEquals(ErrorCode.BAD_REQUEST.code, e1.code());
 
-    try {
+    SolrException e2 = expectThrows(SolrException.class, () -> {
       SolrQuery query = new SolrQuery();
       query.setStart(-1).setQuery("*");
       QueryResponse resp = query(query);
-      fail("Expected the last query to fail, but got response: " + resp);
-    } catch (SolrException e) {
-      assertEquals(ErrorCode.BAD_REQUEST.code, e.code());
-    }
+    });
+    assertEquals(ErrorCode.BAD_REQUEST.code, e2.code());
 
-    try {
+    SolrException e3 = expectThrows(SolrException.class, () -> {
       SolrQuery query = new SolrQuery();
       query.setRows(-1).setStart(0).setQuery("*");
       QueryResponse resp = query(query);
-      fail("Expected the last query to fail, but got response: " + resp);
-    } catch (SolrException e) {
-      assertEquals(ErrorCode.BAD_REQUEST.code, e.code());
-    }
+    });
+    assertEquals(ErrorCode.BAD_REQUEST.code, e3.code());
 
-    try {
+    SolrException e4 = expectThrows(SolrException.class, () -> {
       SolrQuery query = new SolrQuery();
       query.setParam("rows", "non_numeric_value").setQuery("*");
       QueryResponse resp = query(query);
-      fail("Expected the last query to fail, but got response: " + resp);
-    } catch (SolrException e) {
-      assertEquals(ErrorCode.BAD_REQUEST.code, e.code());
-    }
+    });
+    assertEquals(ErrorCode.BAD_REQUEST.code, e4.code());
+
     resetExceptionIgnores();
   }
 }
--- a/solr/core/src/test/org/apache/solr/cloud/BasicZkTest.java
+++ b/solr/core/src/test/org/apache/solr/cloud/BasicZkTest.java
@@ -150,7 +150,9 @@ public class BasicZkTest extends AbstractZkTestCase {
     zkController.getZkClient().setData("/configs/conf1/solrconfig.xml", new byte[0], true);
  
     // we set the solrconfig to nothing, so this reload should fail
-    SolrException e = expectThrows(SolrException.class, () -> {
+    SolrException e = expectThrows(SolrException.class,
+        "The reloaded SolrCore did not pick up configs from zookeeper",
+        () -> {
       ignoreException("solrconfig.xml");
       h.getCoreContainer().reload(h.getCore().getName());
     });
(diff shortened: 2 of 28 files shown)


solr/core/src/test/org/apache/solr/cloud/BasicZkTest.java
```diff
@@ -150,15 +150,13 @@ public void testBasic() throws Exception {
     zkController.getZkClient().setData("/configs/conf1/solrconfig.xml", new byte[0], true);
  
     // we set the solrconfig to nothing, so this reload should fail
-    try {
+    SolrException e = expectThrows(SolrException.class, () -> {
       ignoreException("solrconfig.xml");
       h.getCoreContainer().reload(h.getCore().getName());
-      fail("The reloaded SolrCore did not pick up configs from zookeeper");
```
gerlowskija: One downside I've noticed in moving towards the `expectThrows` pattern is that we lose a lot of the nice error messages that would appear if no exception was thrown at all.

To help with this, I added another version of `expectThrows`, which takes an additional argument: a String message to display if no exception was thrown.  (Link to that new method here: https://github.com/apache/lucene-solr/blob/master/lucene/test-framework/src/java/org/apache/lucene/util/LuceneTestCase.java#L2672)

If you're revising this PR (I noticed you marked it as WIP), consider using that version of `expectThrows` where it makes sense, to preserve the better error messages.

solr/core/src/test/org/apache/solr/TestDistributedSearch.java
```diff
@@ -910,13 +910,11 @@ public void test() throws Exception {
...
     //SOLR 3161 ensure shards.qt=/update fails (anything but search handler really)
     // Also see TestRemoteStreaming#testQtUpdateFails()
-    try {
-      ignoreException("isShard is only acceptable");
-      // query("q","*:*","shards.qt","/update","stream.body","<delete><query>*:*</query></delete>");
-      // fail();
-    } catch (SolrException e) {
-      //expected
-    }
+
```
gerlowskija: [0] Did you intend to leave this commented out?  (I know this PR is a WIP, so totally possible you're aware of this already.  Just wanted to mention it in-case).

solr/core/src/test/org/apache/solr/cloud/BasicZkTest.java
```diff
@@ -150,15 +150,13 @@ public void testBasic() throws Exception {
...
-    } catch(SolrException e) {
-      resetExceptionIgnores();
-      assertTrue(e.getMessage().contains("Unable to reload core [collection1]"));
-      assertTrue(e.getCause().getMessage().contains("Error loading solr config from solrconfig.xml"));
-    }
+    });
+    resetExceptionIgnores();
```
gerlowskija: [0] This isn't behavior you introduced, but I would've expected this to be in a `finally` block.

As I read this, if an IOException or anything else gets triggered by reloading the core, our exception-ignoring will never be reset.

I'm not asking you to change anything here.  Just musing aloud.  Will look into it.

solr/core/src/test/org/apache/solr/cloud/OverseerStatusTest.java
```diff
@@ -58,14 +59,13 @@ public void test() throws Exception {
...
-      fail("Split shard for non existent collection should have failed");
-    } catch (Exception e) {
-      // expected because we did not correctly specify required params for split
-    }
+    HttpSolrClient.RemoteSolrException e = expectThrows(HttpSolrClient.RemoteSolrException.class, () -> CollectionAdminRequest
+        .splitShard("non_existent_collection")
+        .setShardName("non_existent_shard")
+        .process(cluster.getSolrClient())
+    );
+    assertTrue("expected failure to be caused by non existent collection", e.getMessage().contains("Could not find collection"));
```
gerlowskija: [-1] I agree with your intention here (it's a good assertion to add), but I'd rather keep non-refactor changes out of this PR.  Just in the interest of changing the test-behavior as little as possible for this refactor JIRA.  Would you mind taking out this assertion?

solr/core/src/test/org/apache/solr/cloud/SolrXmlInZkTest.java
```diff
@@ -139,16 +139,13 @@ public void testNotInZkFallbackLocal() throws Exception {
 
   @Test
   public void testNotInZkOrOnDisk() throws Exception {
-    try {
+    SolrException e = expectThrows(SolrException.class, () -> {
```
gerlowskija: [-1] I think the `closeZK()` here should stay in a finally, to make sure it executes even if other unexpected exceptions pop up.

So it'd have to look something like:

```
try {
    SolrException e = expectThrows(SolrException.class, () -> {
    ...
    }
    assertTrue(.....);
} finally {
    closeZk();
}
```

It is a little weird to have both the try-finally, and the expectThrows.  If you think it's not even worth it to add in the expectThrows in a case like this, I could see that argument.  I could go either way on it really. Our hands are tied unfortunately, as we've gotta have that finally-block though.

solr/core/src/test/org/apache/solr/cloud/TestAuthenticationFramework.java
```diff
@@ -78,17 +79,11 @@ public void testBasics() throws Exception {
...
-        fail("Should've returned a 401 error");
-      }
-    } finally {
-      MockAuthenticationPlugin.expectedUsername = null;
-      MockAuthenticationPlugin.expectedPassword = null;
-    }
+    HttpSolrClient.RemoteSolrException e = expectThrows(HttpSolrClient.RemoteSolrException.class,
+        this::collectionCreateSearchDeleteTwice);
+    assertTrue("Should've returned a 401 error", e.getMessage().contains("Error 401"));
+    MockAuthenticationPlugin.expectedUsername = null;
```
gerlowskija: [-1] I think we need to keep the `MockAuthenticationPlugin` calls within a "finally"-block.  See the related comment in SolrXmlInZkTest.java above.

solr/core/src/test/org/apache/solr/cloud/TestPullReplicaErrorHandling.java
```diff
@@ -151,12 +151,14 @@ public void testCantConnectToPullReplica() throws Exception {
...
-      } catch (SolrServerException e) {
-        //expected
-      }
+
+      SolrServerException e = expectThrows(SolrServerException.class, () -> {
+        try(HttpSolrClient pullReplicaClient = getHttpSolrClient(s.getReplicas(EnumSet.of(Replica.Type.PULL)).get(0).getCoreUrl())) {
+          pullReplicaClient.query(new SolrQuery("*:*")).getResults().getNumFound();
+        }
+      });
+      assertTrue("Shouldn't be able to query the pull replica", e.getMessage().contains("refused connection"));
```
gerlowskija: [-1] I like the idea, and agree this is a good assertion to add, but I'd rather not add it in this PR since it's intended to be a pure refactor.  Would you mind removing this assertion?

solr/core/src/test/org/apache/solr/cloud/TestSolrCloudWithSecureImpersonation.java
```diff
@@ -223,37 +223,26 @@ private String getExpectedHostExMsg(String user) {
...
-    }
-    catch (HttpSolrClient.RemoteSolrException ex) {
-      // FixMe: this should return an exception about the host being invalid,
-      // but a bug (HADOOP-11077) causes an NPE instead.
-      //assertTrue(ex.getMessage().contains(getExpectedHostExMsg("noHosts")));
-    }
+    HttpSolrClient.RemoteSolrException e = expectThrows(HttpSolrClient.RemoteSolrException.class,
+        () -> solrClient.request(getProxyRequest("noHosts","bar"))
+    );
+    assertTrue(e.getMessage().contains(getExpectedHostExMsg("noHosts")));
```
gerlowskija: [-1] I'd prefer this assertion remain commented out (and the attached comment stick around as well).

I'm really paranoid about introducing actual test changes in with the refactor-only JIRA.  If you'd like to see it fixed, I'm happy to review it as a part of a different JIRA/PR.  But I'd rather not have the two mix.

solr/core/src/test/org/apache/solr/TestDistributedSearch.java
barrotsteindev: Oops my bad, I'll remove it ASAP

barrotsteindev: Yes, since it was commented out beforehand, I just changed the test to use the new syntax, but I figured there was a reason it was commented out.

solr/core/src/test/org/apache/solr/cloud/BasicZkTest.java
barrotsteindev: Perhaps we should try and change it, though we have to make sure some other tests do not depend on this bug.

solr/core/src/test/org/apache/solr/cloud/OverseerStatusTest.java
barrotsteindev: Sure thing

barrotsteindev:
Hopefully this is satisfactory after the last commit changes.

gerlowskija:
Thanks for all the work Bar.  I'm going to run the tests throughout the day and I hope to get this merged this evening or this weekend.

PR #426, opened 2018-07-28
SOLR-12595: CloudSolrClient.Builder accepts ZK connection strings

vpranckaitis:
Added additional constructor to `CloudSolrClient.Builder` which accepts ZooKeeper connection strings.


solr/solrj/src/test/org/apache/solr/client/solrj/impl/CloudSolrClientBuilderTest.java
```diff
@@ -63,22 +66,20 @@ public void testSeveralZkHostsSpecifiedSingly() throws IOException {
   }
   
   @Test
-  public void testSeveralZkHostsSpecifiedTogether() throws IOException {
```
vpranckaitis: This test was exactly the same as the one above, just with differently formatted newlines.

PR #430, opened 2018-08-02
SOLR-12485: support labelled children in xml documents

moshebla:


solr/core/src/test/org/apache/solr/update/AddBlockUpdateTest.java
```diff
@@ -501,6 +501,73 @@ public void testXML() throws IOException, XMLStreamException {
        
   }
 
+  @Test
+  public void testXMLLabeledChildren() throws IOException, XMLStreamException {
```
dsmiley: This is a copy-paste of the previous test, but as-such doesn't test that these labels are retained in any way.  In this test, I think you just need to use the XMLLoader to get a SolrInputDocument, and at that point you can test that there is a field "test" with these two child documents as values.

I do like that in this test you've displayed the syntax.  I was expecting something like 
```
<doc>
<field name='id'>p1</field>
<field name='comment'>
  <doc>
    <field name='id'>c1</field>
    <field name='text'>I like this</field>
  </doc>
</field>
</doc>
```
What do you think of that syntax?  The intention is to make it super clear that the child document is attached as a field to the parent document, which is also reflected in the internal API, and so it's consistent.  I haven't looked at the details of XMLLoader yet to judge how approachable this is.  Putting a "name" attribute on doc isn't bad but makes the "field" relationship un-obvious.

solr/solrj/src/java/org/apache/solr/client/solrj/util/ClientUtils.java
```diff
@@ -72,7 +72,9 @@ public static void writeXML( SolrInputDocument doc, Writer writer ) throws IOExc
       for( Object v : field ) {
         String update = null;
 
-        if (v instanceof Map) {
+        if(v instanceof SolrInputDocument) {
```
dsmiley: Good catch!

solr/core/src/test/org/apache/solr/update/AddBlockUpdateTest.java
moshebla: Yes I agree. I have just pushed 2 new commits which address both the syntax and the test additions you requested.

solr/solrj/src/java/org/apache/solr/client/solrj/util/ClientUtils.java
```diff
@@ -114,17 +121,34 @@ private static void writeVal(Writer writer, String name, Object v, String update
 
     if (update == null) {
       if (v != null) {
-        XML.writeXML(writer, "field", v.toString(), "name", name );
+        if(v instanceof SolrInputDocument) {
+          OutputStream os = SolrInputDocumentXmlOutput((SolrInputDocument) v);
```
dsmiley: 1st; it's not generally valid to call OutputString.toString() (or Writer.toString() for that matter) unless the declared implementation expressly defines what it does.  I am aware the impl/subclass you've chosen does but the declared type here, OutputStream, does not.  It would be better if that "solrInputDocumentXmlOutput" method returns a String since that's all we're going to do here any way.

But moreover, I'm concerned that the approach here buffers child documents instead of writing directly to a stream.  The buffering happens for each depth again, and gets larger.  It seems this deficiency is mostly due to the API contract of XML.writeXML which is not recursive -- it wants a string value and thus forces you to materialize that up front.  We could choose to just accept this as _perhaps_ not a big deal.  Or, what if XML had an inner interface:  (lambda friendly)
```
  interface Writable {
    void write(Writer w) throws IOException ;
  }
```
And the writeXML method was overloaded with a variant that took that instead of the Value.  To avoid duplicating similar methods, the existing ones could be convenience methods to this one with a simple lambda that wrote the string, possibly escaping it first if needed.  WDYT?

solr/solrj/src/java/org/apache/solr/client/solrj/util/ClientUtils.java
```diff
@@ -114,17 +121,34 @@ private static void writeVal(Writer writer, String name, Object v, String update
...
+          XML.writeXML(writer, "field", false, os.toString(), "name", name, "update", update);
+        }
         XML.writeXML(writer, "field", v.toString(), "name", name, "update", update);
       }
     }
   }
 
+  private static OutputStream SolrInputDocumentXmlOutput(SolrInputDocument v) throws IOException {
```
dsmiley: Never start a method name with an uppercase letter.

solr/core/src/java/org/apache/solr/handler/loader/XMLLoader.java
```diff
@@ -404,6 +404,7 @@ public SolrInputDocument readDoc(XMLStreamReader parser) throws XMLStreamExcepti
     StringBuilder text = new StringBuilder();
     String name = null;
     boolean isNull = false;
+    boolean isLabeledChildDoc = false;
```
dsmiley: I think this won't handle deeply nested cases.  Please enhance the test to have another layer of docs and lets see what happens.

moshebla: I just added a test which seems to pass.
Hopefully I did not miss anything when I wrote it. :)

solr/solrj/src/java/org/apache/solr/common/util/XML.java
```diff
@@ -101,8 +101,12 @@ public final static void writeXML(Writer out, String tag, String val) throws IOE
     }
   }
 
-  /** does NOT escape character data in val, must already be valid XML */
   public final static void writeUnescapedXML(Writer out, String tag, String val, Object... attrs) throws IOException {
+    writeUnescapedXML(out, tag, (writer1) -> writer1.write(val), attrs);
+  }
+
+  /** does NOT escape character data in val, must already be valid XML */
+  public final static void writeUnescapedXML(Writer out, String tag, Writable val, Object... attrs) throws IOException {
```
dsmiley: The method name should not include "unescaped".  Wether it's escaped or not is up to the Writable.  "val" variable name should be valWritable or some-such.  The javadoc comment about what the method does belongs on the method above which you added.

solr/solrj/src/java/org/apache/solr/common/util/XML.java
```diff
@@ -118,16 +122,16 @@ public final static void writeUnescapedXML(Writer out, String tag, String val, O
...
       out.write('/');
       out.write(tag);
       out.write('>');
     }
   }
 
-  /** escapes character data in val if shouldEscape is true*/
-  public final static void writeXML(Writer out, String tag, boolean shouldEscape, String val, Object... attrs) throws IOException {
+  /** escapes character data in val */
+  public final static void writeXML(Writer out, String tag, String val, Object... attrs) throws IOException {
```
dsmiley: Lets make this work in terms of the existing Writable based impl?  (less code)

solr/core/src/java/org/apache/solr/handler/loader/XMLLoader.java
```diff
@@ -453,7 +454,12 @@ public SolrInputDocument readDoc(XMLStreamReader parser) throws XMLStreamExcepti
               }
               break;
             }
-            doc.addField(name, v);
+            if(!isLabeledChildDoc){
+              // do not add empty data if a labeled child document was added
```
dsmiley: Maybe this could be clarified some; I find it confusing.  If at this point we're *not* a labelled child doc, then yes add a value (it won't be empty; yet the comment suggests it is).  Maybe your comment goes in the "else" section?  It may be simpler to read if we avoid the negation -- do an if(isLabeledChildDoc) and else not labelled child doc.  It appears setting isLabeledChildDoc=false is a "reset" action of sorts; maybe say that.

solr/solrj/src/java/org/apache/solr/common/util/XML.java
moshebla: Of course, I have a commit I haven't pushed yet. waiting for ant precommit to finish running.

solr/core/src/java/org/apache/solr/handler/loader/XMLLoader.java
moshebla: I added this since for some odd reason after parsing the child document, the parser would return a string with 2 spaces(the indent) and add that to the field value list.
Yes setting the isLabeledChildDoc=false is a way to reset so the behaviour with non child document values won't change.

moshebla:
Hopefully the new method names are clear enough.
I naming is not one of my strong points. :worried:

solr/solrj/src/java/org/apache/solr/common/util/XML.java
```diff
@@ -208,7 +179,13 @@ public static void escape(String str, Writer out, String[] escapes) throws IOExc
...
   }
 
+  @FunctionalInterface
   public interface Writable {
-    void write(Writer w) throws IOException ;
+    void write(Writer w) throws IOException;
+  }
+
+  @FunctionalInterface
+  public interface Escapable<T, U> {
```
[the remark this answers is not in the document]
moshebla: Perhaps since this is a utility some user will have a use case to escape their own attributes, like we did with the data in the child document. So I added this.

solr/solrj/src/java/org/apache/solr/common/util/XML.java
```diff
@@ -101,83 +102,53 @@ public final static void writeXML(Writer out, String tag, String val) throws IOE
...
+  /** does NOT escape character data in val or attributes, must already be valid XML */
   public final static void writeUnescapedXML(Writer out, String tag, String val, Object... attrs) throws IOException {
-    writeUnescapedXML(out, tag, (writer1) -> writer1.write(val), attrs);
+    writeUnescapedAttrXML(out, tag, (writer1) -> writer1.write(val), attrs);
   }
 
-  /** does NOT escape character data in val, must already be valid XML */
-  public final static void writeUnescapedXML(Writer out, String tag, Writable val, Object... attrs) throws IOException {
+  /** does NOT escape character data in attributes, must already be valid XML */
+  public final static void writeUnescapedAttrXML(Writer out, String tag, Writable valWritable, Object... attrs) throws IOException {
```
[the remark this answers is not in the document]
moshebla: I changed it to writeElementXML, since when using null values this method's signature conflicts with the existing writeXML.

solr/solrj/src/java/org/apache/solr/common/util/XML.java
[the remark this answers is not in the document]
dsmiley: It's somewhat unusual to have pre-escaped content, I think.  Case in point: the "unescape" method, judging from "find usages" in my IDE, is only used in tests.  Furthermore, I think it's not a good idea to introduce an interface or class to serve a use case that does not yet exist, nor do we particularly think it's going to happen.

solr/solrj/src/java/org/apache/solr/common/util/XML.java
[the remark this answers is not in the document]
dsmiley: I see your point.  I'm not sure it's worth introducing another name for it though... it'd only be an issue in a couple places, and this class ought not to be used by others generally even though it's possible.  I'm not even sure I like writeElementXML because it's name is suggestive the other similar methods aren't Elements when they all are (since why aren't they named similarly).  If you really want to differentiate by name, perhaps having a writeEmptyXML method for callers who expressly pass null (again; that's only 2 places)

solr/solrj/src/java/org/apache/solr/common/util/XML.java
```diff
@@ -101,82 +102,56 @@ public final static void writeXML(Writer out, String tag, String val) throws IOE
...
-      out.write('<');
-      out.write('/');
-      out.write(tag);
-      out.write('>');
-    }
+    writeXML(out, tag, val, attrs.entrySet().stream().flatMap((entry) -> Stream.of(entry.getKey(), entry.getValue())).toArray());
   }
 
-  private static void escape(char [] chars, int offset, int length, Writer out, String [] escapes) throws IOException{
+  public static void escape(char [] chars, int offset, int length, Writer out, String [] escapes) throws IOException{
```
dsmiley: Why make this public?  It seems rather internal: it isn't even documented; you have to really read the code to know what it does and what "escapes" is.

moshebla: Sorry forgot to revert this change. I was playing around with the code. Will switch it back to private.

moshebla:
Hopefully your previous requests were addressed in these commits.

dsmiley:
This is looking really good Moshe; I'll commit later today after I do precommit & tests

dsmiley:
RE your last commit, why change `(String)null` to `(XML.Writeable)null`?  The latter exposes internal details/abstractions to a calling class that ought not to care?  (can simply change this back)

moshebla:
I was thinking it is a bit more efficient, but I guess that small gain in performance does not justify the exposure of a more internal API.

PR #432, opened 2018-08-06
LUCENE-8438: RAMDirectory speed improvements and cleanup

dweiss:
PR for the issue.


lucene/core/src/java/org/apache/lucene/store/ByteBuffersDataOutput.java
```diff
@@ -0,0 +1,488 @@
...
+import org.apache.lucene.util.BytesRef;
+import org.apache.lucene.util.RamUsageEstimator;
+import org.apache.lucene.util.UnicodeUtil;
+
+public final class ByteBuffersDataOutput extends DataOutput implements Accountable {
+  private final static ByteBuffer EMPTY = ByteBuffer.allocate(0);
+  private final static List<ByteBuffer> EMPTY_LIST = Arrays.asList(EMPTY);
+  private final static byte [] EMPTY_BYTE_ARRAY = {};
+
+  public final static IntFunction<ByteBuffer> ALLOCATE_BB_ON_HEAP = (size) -> {
```
uschindler: Just write `public final static IntFunction<ByteBuffer> ALLOCATE_BB_ON_HEAP = ByteBuffer::allocate;`, no lambda needed. The same applies at other places in the PR, but this looks really strange.

dweiss: Ah, sure. It's just easier for me to see what the argument is sometimes. I'll correct.

uschindler: There are other places with unneeded lambdas, but that's just a preference. I am fine with both.

dweiss: I'll try to pay attention to that, thanks for looking!

lucene/core/src/java/org/apache/lucene/store/ByteBuffersDirectory.java
```diff
@@ -0,0 +1,237 @@
...
+            fileName,
+            array.length);
+        return new ByteArrayIndexInput(inputName, array, 0, array.length);
+      };
+
+  public static final BiFunction<String, ByteBuffersDataOutput, IndexInput> OUTPUT_AS_MANY_BUFFERS_LUCENE = 
+      (fileName, output) -> {
+        List<ByteBuffer> bufferList = output.toBufferList();
+        int chunkSizePower;
+        bufferList.add(ByteBuffer.allocate(0));
```
uschindler: why do we need the empty buffer

dweiss: Lucene's implementation makes an assumption about this, actually (if multiple buffers are used). Try to remove it and run the tests, you'll see.

uschindler: I think the 0-byte buffer is only needed for buffers with exactly the chunkSize, see the comment:

```java
    // we always allocate one more buffer, the last one may be a 0 byte one
    final int nrBuffers = (int) (length >>> chunkSizePower) + 1;
```

So the last one is a 0-bytes buffer, if the total size is actually exactly 2^x. Maybe we can fix this, but this made calculations more easy.

uschindler: What exception did you get and where?

dweiss: I know -- it's the wraparound case in case pos falls just outside of the block boundary. Simplifies some logic later on, I think it's harmless.

ctargett:
I think this can be closed since from the commits here the branch was merged with master back in Aug 2018, and the corresponding Jira issue is resolved.

PR #433, opened 2018-08-08
SOLR-12357 Premptive creation of collections in Time Routed Aliases

nsoft:
new PR to clean up the strange merge conflicts

### DIFF
--- a/solr/core/src/java/org/apache/solr/cloud/api/collections/TimeRoutedAlias.java
+++ b/solr/core/src/java/org/apache/solr/cloud/api/collections/TimeRoutedAlias.java
@@ -17,6 +17,7 @@
 
 package org.apache.solr.cloud.api.collections;
 
+import java.text.ParseException;
 import java.time.Instant;
 import java.time.ZoneOffset;
 import java.time.format.DateTimeFormatter;
@@ -171,6 +172,13 @@ public class TimeRoutedAlias {
         }
       } catch (Exception e) {
         throw new SolrException(BAD_REQUEST, "bad " + TimeRoutedAlias.ROUTER_AUTO_DELETE_AGE + ", " + e, e);
+      }
+    }
+    if (preemptiveCreateMath != null) {
+      try {
+        new DateMathParser().parseMath(preemptiveCreateMath);
+      } catch (ParseException e) {
+        throw new SolrException(BAD_REQUEST, "Invalid date math for preemptiveCreateMath:" + preemptiveCreateMath);
       }
     }
 
--- a/solr/core/src/java/org/apache/solr/cloud/api/collections/TimeRoutedAlias.java
+++ b/solr/core/src/java/org/apache/solr/cloud/api/collections/TimeRoutedAlias.java
@@ -63,6 +63,7 @@ public class TimeRoutedAlias {
   public static final String ROUTER_START = ROUTER_PREFIX + "start";
   public static final String ROUTER_INTERVAL = ROUTER_PREFIX + "interval";
   public static final String ROUTER_MAX_FUTURE = ROUTER_PREFIX + "maxFutureMs";
+  public static final String ROUTER_PREEMPTIVE_CREATE_WINDOW = ROUTER_PREFIX + "preemptiveCreateWindow";
   public static final String ROUTER_AUTO_DELETE_AGE = ROUTER_PREFIX + "autoDeleteAge";
   public static final String CREATE_COLLECTION_PREFIX = "create-collection.";
   // plus TZ and NAME
@@ -84,6 +85,7 @@ public class TimeRoutedAlias {
   public static final List<String> OPTIONAL_ROUTER_PARAMS = Collections.unmodifiableList(Arrays.asList(
       ROUTER_MAX_FUTURE,
       ROUTER_AUTO_DELETE_AGE,
+      ROUTER_PREEMPTIVE_CREATE_WINDOW,
       TZ)); // kinda special
 
   static Predicate<String> PARAM_IS_PROP =
@@ -126,6 +128,7 @@ public class TimeRoutedAlias {
   private final String routeField;
   private final String intervalMath; // ex: +1DAY
   private final long maxFutureMs;
+  private final String preemptiveCreateWindow;
   private final String autoDeleteAgeMath; // ex: /DAY-30DAYS  *optional*
   private final TimeZone timeZone;
 
@@ -141,6 +144,7 @@ public class TimeRoutedAlias {
 
     //optional:
     maxFutureMs = params.getLong(ROUTER_MAX_FUTURE, TimeUnit.MINUTES.toMillis(10));
+    preemptiveCreateWindow = params.get(ROUTER_PREEMPTIVE_CREATE_WINDOW); // no default
     autoDeleteAgeMath = params.get(ROUTER_AUTO_DELETE_AGE); // no default
     timeZone = TimeZoneUtils.parseTimezone(aliasMetadata.get(CommonParams.TZ));
 
@@ -189,6 +193,10 @@ public class TimeRoutedAlias {
     return maxFutureMs;
   }
 
+  public String getPreemptiveCreateWindow() {
+    return preemptiveCreateWindow;
+  }
+
   public String getAutoDeleteAgeMath() {
     return autoDeleteAgeMath;
   }
@@ -204,6 +212,7 @@ public class TimeRoutedAlias {
         .add("routeField", routeField)
         .add("intervalMath", intervalMath)
         .add("maxFutureMs", maxFutureMs)
+        .add("preemptiveCreateWindow", preemptiveCreateWindow)
         .add("autoDeleteAgeMath", autoDeleteAgeMath)
         .add("timeZone", timeZone)
         .toString();
--- a/solr/core/src/java/org/apache/solr/cloud/api/collections/TimeRoutedAlias.java
+++ b/solr/core/src/java/org/apache/solr/cloud/api/collections/TimeRoutedAlias.java
@@ -63,7 +63,7 @@ public class TimeRoutedAlias {
   public static final String ROUTER_START = ROUTER_PREFIX + "start";
   public static final String ROUTER_INTERVAL = ROUTER_PREFIX + "interval";
   public static final String ROUTER_MAX_FUTURE = ROUTER_PREFIX + "maxFutureMs";
-  public static final String ROUTER_PREEMPTIVE_CREATE_WINDOW = ROUTER_PREFIX + "preemptiveCreateWindow";
+  public static final String ROUTER_PREEMPTIVE_CREATE_MATH = ROUTER_PREFIX + "preemptiveCreateMath";
   public static final String ROUTER_AUTO_DELETE_AGE = ROUTER_PREFIX + "autoDeleteAge";
   public static final String CREATE_COLLECTION_PREFIX = "create-collection.";
   // plus TZ and NAME
@@ -85,7 +85,7 @@ public class TimeRoutedAlias {
   public static final List<String> OPTIONAL_ROUTER_PARAMS = Collections.unmodifiableList(Arrays.asList(
       ROUTER_MAX_FUTURE,
       ROUTER_AUTO_DELETE_AGE,
-      ROUTER_PREEMPTIVE_CREATE_WINDOW,
+      ROUTER_PREEMPTIVE_CREATE_MATH,
       TZ)); // kinda special
 
   static Predicate<String> PARAM_IS_PROP =
@@ -128,7 +128,7 @@ public class TimeRoutedAlias {
   private final String routeField;
   private final String intervalMath; // ex: +1DAY
   private final long maxFutureMs;
-  private final String preemptiveCreateWindow;
+  private final String preemptiveCreateMath;
   private final String autoDeleteAgeMath; // ex: /DAY-30DAYS  *optional*
   private final TimeZone timeZone;
 
@@ -144,7 +144,9 @@ public class TimeRoutedAlias {
 
     //optional:
     maxFutureMs = params.getLong(ROUTER_MAX_FUTURE, TimeUnit.MINUTES.toMillis(10));
-    preemptiveCreateWindow = params.get(ROUTER_PREEMPTIVE_CREATE_WINDOW); // no default
+    // the date math configured is an interval to be subtracted from the most recent collection's time stamp
+    preemptiveCreateMath = params.get(ROUTER_PREEMPTIVE_CREATE_MATH) != null ?
+        "-" + params.get(ROUTER_PREEMPTIVE_CREATE_MATH) : null;
     autoDeleteAgeMath = params.get(ROUTER_AUTO_DELETE_AGE); // no default
     timeZone = TimeZoneUtils.parseTimezone(aliasMetadata.get(CommonParams.TZ));
 
@@ -194,7 +196,7 @@ public class TimeRoutedAlias {
   }
 
   public String getPreemptiveCreateWindow() {
-    return preemptiveCreateWindow;
+    return preemptiveCreateMath;
   }
 
   public String getAutoDeleteAgeMath() {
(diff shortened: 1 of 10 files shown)


solr/core/src/java/org/apache/solr/cloud/api/collections/TimeRoutedAlias.java
```diff
@@ -141,6 +145,9 @@ public TimeRoutedAlias(String aliasName, Map<String, String> aliasMetadata) {
 
     //optional:
     maxFutureMs = params.getLong(ROUTER_MAX_FUTURE, TimeUnit.MINUTES.toMillis(10));
+    // the date math configured is an interval to be subtracted from the most recent collection's time stamp
+    preemptiveCreateMath = params.get(ROUTER_PREEMPTIVE_CREATE_MATH) != null ?
```
dsmiley: Perhaps this should detect if it already starts with a plus or minus and if so removes it, thus making it more friendly to people who like to include the operator?

solr/core/src/test/org/apache/solr/update/processor/TimeRoutedAliasUpdateProcessorTest.java
```diff
@@ -322,6 +325,104 @@ public void testSliceRouting() throws Exception {
...
+    final int numShards = 1 ;
+    final int numReplicas = 1 ;
+    CollectionAdminRequest.createTimeRoutedAlias(alias, "2017-10-23T00:00:00Z", "+1DAY", timeField,
+        CollectionAdminRequest.createCollection("_unused_", configName, numShards, numReplicas)
+            .setMaxShardsPerNode(numReplicas)).setPreemptiveCreateWindow("3HOUR")
+        .process(solrClient);
+
+    // cause some collections to be created
+    assertUpdateResponse(solrClient.add(alias,
+        sdoc("id","1","timestamp_dt", "2017-10-25T00:00:00Z")
```
dsmiley: To reduce the overhead of the test, can we start one one collection, maybe two, before getting into pre-emption?
And this test should very likely be annotated with \@Slow

solr/core/src/test/org/apache/solr/update/processor/TimeRoutedAliasUpdateProcessorTest.java
```diff
@@ -322,6 +325,104 @@ public void testSliceRouting() throws Exception {
...
+    assertNumDocs("2017-10-23", 1);
+    assertNumDocs("2017-10-24", 1);
+    assertNumDocs("2017-10-25", 4);
+    assertNumDocs("2017-10-26", 0);
+
+    // now test with pre-create window longer than time slice, and forcing multiple creations.
+    CollectionAdminRequest.setAliasProperty(alias)
+        .addProperty(TimeRoutedAlias.ROUTER_PREEMPTIVE_CREATE_MATH, "3DAY").process(solrClient);
+
+    Thread.sleep(1000); // a moment to be sure the alias change has taken effect
```
dsmiley: curious; has this been shown to be necessary?  I can imagine it might be... since a new TRA URP isn't doing to do a ZK sync to view the alias state.

solr/core/src/test/org/apache/solr/update/processor/TimeRoutedAliasUpdateProcessorTest.java
```diff
@@ -343,8 +444,31 @@ public void testSliceRouting() throws Exception {
...
+    // and alias to be aware of collection
+    long start = System.nanoTime(); // mumble mumble precommit mumble mumble...
+    while (!cluster.getSolrClient().getZkStateReader().getAliases().getCollectionAliasListMap().get(alias).contains(collection)) {
+      if (NANOSECONDS.toSeconds(System.nanoTime() - start) > 10) {
+        fail("took over 10 seconds after collection creation to update aliases");
+      } else {
+        try {
+          Thread.sleep(500);
+        } catch (InterruptedException e) {
+          e.printStackTrace();
```
dsmiley: -- and return or throw.  Interrupt means stop what you're doing.

solr/core/src/test/org/apache/solr/update/processor/TimeRoutedAliasUpdateProcessorTest.java
```diff
@@ -322,6 +325,104 @@ public void testSliceRouting() throws Exception {
...
+    assertNumDocs("2017-10-25", 1);
+
+    // cause some collections to be created
+
+    ModifiableSolrParams params = params();
+    assertUpdateResponse(add(alias, Arrays.asList(
+        sdoc("id", "2", "timestamp_dt", "2017-10-24T00:00:00Z"),
+        sdoc("id", "3", "timestamp_dt", "2017-10-25T00:00:00Z"),
+        sdoc("id", "4", "timestamp_dt", "2017-10-23T00:00:00Z"),
+        sdoc("id", "5", "timestamp_dt", "2017-10-25T23:00:00Z")), // should cause preemptive creation
```
dsmiley: lets also add another document after this, one that would also cause pre-emptive creation if it weren't for there already being a request in-progress

solr/core/src/java/org/apache/solr/update/processor/TimeRoutedAliasUpdateProcessor.java
```diff
@@ -167,59 +167,17 @@ private String getAliasName() {
...
-      if (!updated) { // thus we didn't make progress...
-        // this is not expected, even in known failure cases, but we check just in case
-        throw new SolrException(SolrException.ErrorCode.SERVER_ERROR,
-            "We need to create a new time routed collection but for unknown reasons were unable to do so.");
-      }
-      // then retry the loop ...
-    } while(true);
-    assert targetCollection != null;
-
+    String candidateCollection = findCandidateCollectionGivenTimestamp(docTimestampToRoute, cmd.getPrintableId());
```
dsmiley: You can move this line to immediately before its first use of the result.  Presently, the maxFutureTime check is inbetween which breaks up the natural flow.
Hmm; even the "updateParsedCollectionAliases()" call can move down.
Finally, some newlines here & there would help separate separate steps.

solr/core/src/java/org/apache/solr/update/processor/TimeRoutedAliasUpdateProcessor.java
```diff
@@ -94,13 +92,15 @@
   private final SolrCmdDistributor cmdDistrib;
   private final CollectionsHandler collHandler;
   private final SolrParams outParamsToLeader;
+  @SuppressWarnings("FieldCanBeLocal")
   private final CloudDescriptor cloudDesc;
 
   private List<Map.Entry<Instant, String>> parsedCollectionsDesc; // k=timestamp (start), v=collection.  Sorted descending
   private Aliases parsedCollectionsAliases; // a cached reference to the source of what we parse into parsedCollectionsDesc
   private SolrQueryRequest req;
+  private ExecutorService preemptiveCreationExecutor;
```
dsmiley: Since it will be nulled out in another thread, we ought to declare this as volatile.  I know this is being a bit pedantic since I don't think it'd be a real problem.

solr/core/src/java/org/apache/solr/update/processor/TimeRoutedAliasUpdateProcessor.java
```diff
@@ -230,6 +188,95 @@ public void processAdd(AddUpdateCommand cmd) throws IOException {
...
+        case ASYNC_PREEMPTIVE:
+          // Note: creating an executor and throwing it away is slightly expensive, but this is only likely to happen
+          // once per hour/day/week (depending on time slice size for the TRA). If the executor were retained, it
+          // would need to be shut down in a close hook to avoid test failures due to thread leaks which is slightly
+          // more complicated from a code maintenance and readability stand point. An executor must used instead of a
+          // thread to ensure we pick up the proper MDC logging stuff from ExecutorUtil. T
+          if (preemptiveCreationExecutor == null) {
+            DefaultSolrThreadFactory threadFactory = new DefaultSolrThreadFactory("TRA-preemptive-creation");
+            preemptiveCreationExecutor = newMDCAwareSingleThreadExecutor(threadFactory);
+            preemptiveCreationExecutor.execute(() -> {
```
dsmiley: the code executed in the new thread should not call maintain() since I see you had to make maintain more complicated to tell if it's being called from "async".  It can call: 
```
      final String mostRecentCollName = parsedCollectionsDesc.get(0).getValue();
      createCollectionAfter(mostRecentCollName);
```

solr/core/src/java/org/apache/solr/update/processor/TimeRoutedAliasUpdateProcessor.java
```diff
@@ -230,6 +188,95 @@ public void processAdd(AddUpdateCommand cmd) throws IOException {
...
+  private CreationType requiresCreateCollection(Instant routeTimestamp,  String preemptiveCreateMath) {
+    final Instant mostRecentCollTimestamp = parsedCollectionsDesc.get(0).getKey();
+    final Instant nextCollTimestamp = timeRoutedAlias.computeNextCollTimestamp(mostRecentCollTimestamp);
+    if (!routeTimestamp.isBefore(nextCollTimestamp)) {
+      // current document is destined for a collection that doesn't exist, must create the destination
+      // to proceed with this add command
+      return SYNCHRONOUS;
+    }
+
+    if (isBlank(preemptiveCreateMath)) {
```
dsmiley: nitpick: this method might flow better if you check for _not_ blank, then see if should return ASYNC_PREEMPTIVE, and at the very end return NONE.  This would mean exactly 3 "return" statements of each type, which reads better IMO.

solr/core/src/test/org/apache/solr/update/processor/TimeRoutedAliasUpdateProcessorTest.java
```diff
@@ -392,29 +393,73 @@ public void testPreemptiveCreation() throws Exception {
...
     assertNumDocs("2017-10-28", 0);
 
     QueryResponse resp;
     resp = solrClient.query(alias, params(
         "q", "*:*",
         "rows", "10"));
-    assertEquals(7, resp.getResults().getNumFound());
+    assertEquals(8, resp.getResults().getNumFound());
+
+    assertUpdateResponse(add(alias, Arrays.asList(
```
dsmiley: Shouldn't we use `addDocsAndCommit` here and the other spots where multiple docs are added at a time?  I know you're passing params but it's empty.

Otherwise, tests look good!

solr/core/src/test/org/apache/solr/update/processor/TimeRoutedAliasUpdateProcessorTest.java
nsoft: I think this comment is obsolete, carried over from earlier versions?

solr/core/src/test/org/apache/solr/update/processor/TimeRoutedAliasUpdateProcessorTest.java
nsoft: addDocsAndCommit contains a lot of logic I don't really want, especially the shuffling of the input documents!

solr/core/src/java/org/apache/solr/update/processor/TimeRoutedAliasUpdateProcessor.java
```diff
@@ -437,45 +461,46 @@ protected void doClose() {
...
+            // this is not expected, even in known failure cases, but we check just in case
+            throw new SolrException(SolrException.ErrorCode.SERVER_ERROR,
+                "We need to create a new time routed collection but for unknown reasons were unable to do so.");
+          }
+          // then retry the loop ... have to do find again in case other requests also added collections
+          // that were made visible when we called updateParsedCollectionAliases()
+          targetCollectionDesc = findCandidateGivenTimestamp(docTimestamp, printableId);
+          break;
+        default:
+          throw new SolrException(SolrException.ErrorCode.SERVER_ERROR, "Unknown creation type while adding " +
```
dsmiley: I wouldn't waste your virtual breath (lines of code with thoughtful explanation) on a default case of an enum switch.    Another approach I like is to `assert ENUMNAME.values().length == 3;` before the switch which will be caught at test time.  I forget but if java still insists we have a default then throw IllegalStateException or something like that in a one-liner without explaination.  (keep it short & sweet)

solr/core/src/java/org/apache/solr/update/processor/TimeRoutedAliasUpdateProcessor.java
```diff
@@ -437,45 +461,46 @@ protected void doClose() {
...
-        return targetCollection; // thus we don't need another collection
+  private String createAllRequiredCollections( Instant docTimestamp, String printableId,
+                                               Map.Entry<Instant, String> targetCollectionDesc) {
+    do {
+      switch(typeOfCreationRequired(docTimestamp, targetCollectionDesc.getKey())) {
+        case NONE:
+          return targetCollectionDesc.getValue(); // we don't need another collection
+        case ASYNC_PREEMPTIVE:
+          // can happen when preemptive interval is longer than one time slice
+          preemptiveAsync(() -> createNextCollection(this.parsedCollectionsDesc.get(0).getValue()));
```
dsmiley: bug: do not refer to this.parsedCollectionsDesc from the lambda runnable as it is not thread-safe (and I don't think we should try to make it so.  the input should be gathered in the immediate line prior (thus in the main thread) so that it's already resolved.

solr/core/src/java/org/apache/solr/update/processor/TimeRoutedAliasUpdateProcessor.java
```diff
@@ -198,32 +209,32 @@ private String createCollectionsIfRequired(Instant docTimestamp, String targetCo
...
-            });
+            // It's important not to add code between here and the prior call to findCandidateGivenTimestamp()
+            // in processAdd() that invokes updateParsedCollectionAliases(). Doing so would update parsedCollectionsDesc
+            // and create a race condition. We are relying on the fact that get(0) is returning the head of the parsed
+            // collections that existed when candidateCollectionDesc was created. If this class updates it's notion of
+            // parsedCollectionsDesc since candidateCollectionDesc was chosen, we could create collection n+2
+            // instead of collection n+1.
+
+            // This line does not block and the document can be added immediately
+            preemptiveAsync(() -> createNextCollection(this.parsedCollectionsDesc.get(0).getValue()));
```
dsmiley: bug: do not refer to this.parsedCollectionsDesc from the lambda runnable as it is not thread-safe (and I don't think we should try to make it so.  the input should be gathered in the immediate line prior (thus in the main thread) so that it's already resolved.

To make this bug harder (you did it twice), you could change preemptiveAsync() to take the collection name instead of taking a lambda.  It could then be named createNextCollectionAsync

solr/core/src/java/org/apache/solr/update/processor/TimeRoutedAliasUpdateProcessor.java
```diff
@@ -165,31 +165,42 @@ private String getAliasName() {
...
       cmdDistrib.distribAdd(cmd, Collections.singletonList(targetLeaderNode), new ModifiableSolrParams(outParamsToLeader));
     }
   }
 
-
-  private String createCollectionsIfRequired(Instant docTimestamp, String targetCollection, String printableId) {
+  /**
+   * Create any required collections and return the name of the collection to which the current document should be sent.
+   *
+   * @param docTimestamp the date for the document taken from the field specified in the TRA config
```
dsmiley: These parameter-level docs are fine but FYI don't feel that you have to do this, especially for private methods, and especially for obvious parameters (i.e. you don't *have* to document all; you could just do some at your discretion, or none).  This is subjective of course; I'm sharing my opinion and not insisting on anything.  My preference leans towards... "if you having something helpful to say then say it, but avoid writing the obvious" (for various reasons)

Oh look... you don't need to pass printableId since you could get it from AddUpdateCommand if needed.  printableId is a wart on these method signatures IMO.

PR #435, opened 2018-08-11
SOLR-12591: add tests to ensure ParseDateFieldUpdateProcessor compati…

barrotsteindev:
…blity with ExtractionDateUtil

### DIFF
--- a/solr/core/src/java/org/apache/solr/update/processor/ParseDateFieldUpdateProcessorFactory.java
+++ b/solr/core/src/java/org/apache/solr/update/processor/ParseDateFieldUpdateProcessorFactory.java
@@ -26,6 +26,7 @@ import java.time.ZoneOffset;
 import java.time.format.DateTimeFormatter;
 import java.time.format.DateTimeFormatterBuilder;
 import java.time.format.DateTimeParseException;
+import java.time.format.ResolverStyle;
 import java.time.temporal.TemporalAccessor;
 import java.time.temporal.TemporalQueries;
 import java.util.Collection;
@@ -120,6 +121,16 @@ public class ParseDateFieldUpdateProcessorFactory extends FieldMutatingUpdatePro
       protected Object mutateValue(Object srcVal) {
         if (srcVal instanceof CharSequence) {
           String srcStringVal = srcVal.toString();
+          // trim single quotes around date if present
+          // see issue #5279  (Apache HttpClient)
+          int stringValLen = srcStringVal.length();
+          if (stringValLen > 1
+              && srcStringVal.startsWith("'")
+              && srcStringVal.endsWith("'")
+          ) {
+            srcStringVal = srcStringVal.substring(1, stringValLen - 1);
+          }
+
           for (Map.Entry<String,DateTimeFormatter> format : formats.entrySet()) {
             DateTimeFormatter parser = format.getValue();
             try {
@@ -159,8 +170,9 @@ public class ParseDateFieldUpdateProcessorFactory extends FieldMutatingUpdatePro
     Collection<String> formatsParam = args.removeConfigArgs(FORMATS_PARAM);
     if (null != formatsParam) {
       for (String value : formatsParam) {
-        DateTimeFormatter formatter = new DateTimeFormatterBuilder().parseCaseInsensitive()
-            .appendPattern(value).toFormatter(locale).withZone(defaultTimeZone);
+        DateTimeFormatter formatter = new DateTimeFormatterBuilder().parseLenient().parseCaseInsensitive()
+            .appendPattern(value).toFormatter(locale)
+            .withResolverStyle(ResolverStyle.LENIENT).withZone(defaultTimeZone);
         validateFormatter(formatter);
         formats.put(value, formatter);
       }
--- a/solr/core/src/test-files/solr/collection1/conf/solrconfig-parsing-update-processor-chains.xml
+++ b/solr/core/src/test-files/solr/collection1/conf/solrconfig-parsing-update-processor-chains.xml
@@ -109,12 +109,12 @@
     </processor>
   </updateRequestProcessorChain>
 
-  <updateRequestProcessorChain name="parse-date-extraction-util-formats">
+  <updateRequestProcessorChain name="parse-date-patterns-from-extract-contrib">
     <processor class="solr.ParseDateFieldUpdateProcessorFactory">
       <str name="defaultTimeZone">UTC</str>
       <str name="locale">en</str>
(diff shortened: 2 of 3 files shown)


solr/core/src/test/org/apache/solr/update/processor/ParsingFieldUpdateProcessorsTest.java
```diff
@@ -896,13 +899,125 @@ public void testCascadingParsers() throws Exception {
...
+
+  public void testRfc1123() throws IOException {
+    assertParsedDate("Fri, 07 Oct 2005 13:14:15 GMT", Date.from(inst20051007131415()), "parse-date-extraction-util-formats");
+  }
+
+  public void testRfc1036() throws IOException {
+    assertParsedDate("Friday, 07-Oct-05 13:14:15 GMT", Date.from(inst20051007131415()), "parse-date-extraction-util-formats");
+  }
+
+  @AwaitsFix(bugUrl = "https://issues.apache.org/jira/browse/SOLR-12593")
```
barrotsteindev: This seems to pass as is, although I am still hesitant, perhaps I missed something?
Would be nice if another set of eyes looked at this one.

solr/core/src/java/org/apache/solr/update/processor/ParseDateFieldUpdateProcessorFactory.java
```diff
@@ -120,6 +121,16 @@ public UpdateRequestProcessor getInstance(SolrQueryRequest req,
       protected Object mutateValue(Object srcVal) {
         if (srcVal instanceof CharSequence) {
           String srcStringVal = srcVal.toString();
+          // trim single quotes around date if present
```
dsmiley: Oh yeah.  It's unfortunate such ugly data comes in as such... but it's at least cheap & safe to do this so I guess it's okay.

solr/core/src/test-files/solr/collection1/conf/solrconfig-parsing-update-processor-chains.xml
```diff
@@ -109,6 +109,21 @@
     </processor>
   </updateRequestProcessorChain>
 
+  <updateRequestProcessorChain name="parse-date-extraction-util-formats">
```
dsmiley: Soon there will be no such "date extraction util" so I suggest another name.  Hmmm.  Perhaps "parse-date-patterns-from-extract-contrib".  Yeah it's long but no big deal.

solr/core/src/test-files/solr/collection1/conf/solrconfig-parsing-update-processor-chains.xml
```diff
@@ -109,6 +109,21 @@
...
+    <processor class="solr.ParseDateFieldUpdateProcessorFactory">
+      <str name="defaultTimeZone">UTC</str>
+      <str name="locale">en</str>
+      <arr name="format">
+        <str>yyyy-MM-dd['T'[HH:mm:ss['.'SSS]['Z'</str>
```
dsmiley: After working with you on the previous issue, I now know that `'Z'` isn't appropriate if the defaultTimeZone is configurable, since as-such it's a literal that is ignored -- and the defaultTimeZone would take effect which might not be UTC.  We know 'z' is right now.

solr/core/src/java/org/apache/solr/update/processor/ParseDateFieldUpdateProcessorFactory.java
```diff
@@ -159,8 +170,9 @@ public void init(NamedList args) {
     Collection<String> formatsParam = args.removeConfigArgs(FORMATS_PARAM);
     if (null != formatsParam) {
       for (String value : formatsParam) {
-        DateTimeFormatter formatter = new DateTimeFormatterBuilder().parseCaseInsensitive()
-            .appendPattern(value).toFormatter(locale).withZone(defaultTimeZone);
+        DateTimeFormatter formatter = new DateTimeFormatterBuilder().parseLenient().parseCaseInsensitive()
```
dsmiley: I see you're hard-coding leniency and not making it an option.  It's fine, I think.  I'm not sure why someone would expressly want it to be strict (would want an error), not for an application of parsing like we see here.

solr/core/src/test/org/apache/solr/update/processor/ParsingFieldUpdateProcessorsTest.java
dsmiley: This annotation should not be present.  It was not present in `SOLR-12561.patch`.  It is present today in TestExtractionDateUtil for the string "Thu Nov 13 04:35:51 AKST 2008" and you already have a test for that -- testAKSTZone.  Since the URP is not afflicted with the bug causing me to add that AwaitsFix, we don't need the annotation here in this test class at all.  If you want to try to see if the bug is gone, you need to set this system property like so: `-Dtests.timezone=America/Metlakatla`.  If you don't, Lucene/Solr tests choose a random timezone and it's not particularly likely to get this specific one.

solr/core/src/java/org/apache/solr/update/processor/ParseDateFieldUpdateProcessorFactory.java
barrotsteindev: I guess it can be added as an option,
should I work on it?

dsmiley: nah; it's fine.  Less to document :-)

solr/core/src/test/org/apache/solr/update/processor/ParsingFieldUpdateProcessorsTest.java
barrotsteindev: ran the tests with that particular flag and it seems to pass, so I removed the annotation.

barrotsteindev:
Hopefully the last commit got everything sorted out. :)

solr/core/src/java/org/apache/solr/update/processor/ParseDateFieldUpdateProcessorFactory.java
barrotsteindev: Less work is always welcome 👍

dsmiley:
(At the moment I'm just waiting to hear back from Yetus, the build bot, on tests & precommit)

barrotsteindev:
Oh nice, hopefully the tests will pass :)

barrotsteindev:
Closing this PR,
as it was merged.

PR #436, opened 2018-08-15
SOLR-12652: Remove SolrMetricManager.overridableRegistryName method

petersomogyi:


petersomogyi:
SOLR-12652 was committed to master. Thanks for reviewing!

PR #438, opened 2018-08-21
SOLR-12593: Remove date parsing functionality from extraction contrib

barrotsteindev:


solr/server/solr/configsets/_default/conf/solrconfig.xml
```diff
@@ -1146,6 +1146,11 @@
       <str>yyyy-MM-dd HH:mm[:ss[.SSS]][z</str>
       <str>yyyy-MM-dd HH:mm[:ss[,SSS]][z</str>
       <str>yyyy-MM-dd</str>
+      <str>EEE, dd MMM yyyy HH:mm:ss z</str>
+      <str>EEEE, dd-MMM-yy HH:mm:ss z</str>
+      <str>EEE MMM d HH:mm:ss yyyy</str>
+      <str>EEE MMM d hh:mm:ss z yyyy</str>
```
dsmiley: The "hh" is erroneous.  I'm surprised this even works; perhaps "lenient" allows os to get away with it?  I was hoping to keep the added patterns more minimal.  You recently added "parse-date-patterns-from-extract-contrib" to solrconfig-parsing-update-processor-chains which is a better guide.

solr/server/solr/configsets/_default/conf/solrconfig.xml
```diff
@@ -1146,6 +1146,11 @@
...
+      <str>yyyy-MM-dd hh:mm:ss</str>
```
dsmiley: This pattern could be removed if we add optional brackets to an existing pattern.

barrotsteindev: Right, it could be joined with "yyyy-MM-dd" as "yyyy-MM-dd [hh:mm:ss]".
Or should I replace the "hh" bit with "HH"?

solr/server/solr/configsets/_default/conf/solrconfig.xml
barrotsteindev: Should I just try and copy that part of the config to the default one?

solr/server/solr/configsets/_default/conf/solrconfig.xml
dsmiley: Close; it'd be this:  `yyyy-MM-dd[ hh:mm:ss]`  (notice the bracket starts immediately after the 'dd'; otherwise we'd be wrongly insisting on a trailing space.  

And yes, always use "HH" and not "hh" unless the pattern contains an AM/PM designator (and these don't).

solr/server/solr/configsets/_default/conf/solrconfig.xml
dsmiley: Yeah, that's what I thought you were going to do.  You've separately tested that the ExtractionDateUtils tests also work on these patterns in the last issue, so we want them to be the same; otherwise due to bugs (like your 'hh') we might inadvertently screw that up.  If needed, modify/tweak one of those tests/configs.

barrotsteindev:
Used the patterns from parse-date-patterns-from-extract-contrib,
which seem to work just fine :).

solr/server/solr/configsets/_default/conf/solrconfig.xml
```diff
@@ -1141,11 +1141,11 @@
   <updateProcessor class="solr.ParseDoubleFieldUpdateProcessorFactory" name="parse-double"/>
   <updateProcessor class="solr.ParseDateFieldUpdateProcessorFactory" name="parse-date">
     <arr name="format">
-      <str>yyyy-MM-dd'T'HH:mm[:ss[.SSS]][z</str>
```
dsmiley: You just changed the default date patterns to not parse patterns it parsed before.  We need to be cautious!  For example it used to parse a comma to separate the milliseconds.  And the 'T' was optional.

Whatever patterns exist here ought to be identical in documentation & test configs (which yes is annoying to keep in sync but shouldn't change often).

barrotsteindev: So should I keep the old patterns in addition to the ones from SOLR-12591?

dsmiley: By "the old patterns" do you mean the _current_ patterns we see in the default config?  Yes.  You can tweak an existing pattern to parse more things, but don't remove the ability to parse something it would have parsed before!  And yes add/merge the additional ones from SOLR-12591.  By "merge" I mean modify some existing patterns to match some of the ones in SOLR-12591.

barrotsteindev:
@dsmiley , I am having trouble figuring out which parts of the ref guide I should edit,
could you give me a hand as I am fairly new to this? 😅
Apologies, I am very new to all of this.

dsmiley:
Docs: `solr/solr-ref-guide/src/uploading-data-with-solr-cell-using-apache-tika.adoc`

Alternatively, perhaps leave that part to me.  Or take a 1st cut at it and I edit from there, which I'm likely to do any way.  Either way.

RE the default config:  I'm concerned we're not testing the default config well.  Here's what I recommend:  in solrconfig-parsing-update-processor-chains.xml.  I think there should be one date config that is intentionally identical to the `_default` configSet's configuration of this URP (and should have a name indicating this, like "parse-date-default-configSet").  All the other configurations of this URP in the file should have exactly one pattern and is only there to test a particular condition -- i.e. the effects of a locale or an interesting pattern or something like that.   I realize this recommendation would mean stomping over a bit of the changes in a previous issue, e.g. removing `parse-date-patterns-from-extract-contrib`, but the test code should stay the same, I think, notwithstanding referring to a new/renamed parse-date processor config per test.

To me, the merging of patterns in ExtractDateUtils into the default URP patterns seems distinct enough from removing ExtractDateUtils that it ought to occur in its own issue.

BTW I was looking at DateTimeFormatter.RFC_1123_DATE_TIME and noticed the leading day of week is optional.  We should update our pattern accordingly, and test it's optionality.  It'd be nice if we had a feature like `<str>RFC_1123_DATE_TIME</str>` where we could refer to a well known pattern by it's well-known name and in turn get the pre-compiled instance in DateTimeFormatter.  But that's not in scope here.

solr/contrib/extraction/src/test-files/extraction/solr/collection1/conf/solrconfig.xml
```diff
@@ -43,6 +43,25 @@
...
 
+  <updateProcessor class="solr.ParseDateFieldUpdateProcessorFactory" name="parse-date">
+    <arr name="format">
+      <str>yyyy-MM-dd['T'[HH:mm[:ss[.SSS]][z</str>
+      <str>yyyy-MM-dd['T'[HH:mm[:ss[,SSS]][z</str>
+      <str>yyyy-MM-dd HH:mm[:ss[.SSS]][z</str>
+      <str>yyyy-MM-dd HH:mm[:ss[,SSS]][z</str>
+      <str>EEE MMM d [HH:mm:ss ][z ]yyyy</str>
+      <str>EEEE, dd-MMM-yy HH:mm:ss [z</str>
+      <str>[EEE, ]dd MMM yyyy HH:mm[:ss] [z</str>
```
barrotsteindev: > BTW I was looking at DateTimeFormatter.RFC_1123_DATE_TIME and noticed the leading day of week is optional.

I also had a little peek and noticed the second of the minute is also optional, here is the code snippet that initializes the default RFC_1123_DATE_TIME DateTimeFromatter:

`.optionalStart()
  .appendLiteral(':')
  .appendValue(SECOND_OF_MINUTE, 2)
  .optionalEnd()`

barrotsteindev:
> Alternatively, perhaps leave that part to me. Or take a 1st cut at it and I edit from there, which I'm likely to do any way. Either way.

I read the doc file and could not come up with the right words for the job.
Perhaps this part is better left for you, and I could learn from the final patch so I can do these kind of tasks next time. :)

solr/server/solr/configsets/_default/conf/solrconfig.xml
```diff
@@ -1141,11 +1141,13 @@
...
-      <str>yyyy-MM-dd'T'HH:mm[:ss[,SSS]][z</str>
+      <str>yyyy-MM-dd['T'[HH:mm[:ss[.SSS]][z</str>
+      <str>yyyy-MM-dd['T'[HH:mm[:ss[,SSS]][z</str>
       <str>yyyy-MM-dd HH:mm[:ss[.SSS]][z</str>
       <str>yyyy-MM-dd HH:mm[:ss[,SSS]][z</str>
-      <str>yyyy-MM-dd</str>
+      <str>EEE MMM d [HH:mm:ss ][z ]yyyy</str>
+      <str>EEEE, dd-MMM-yy HH:mm:ss [z</str>
```
dsmiley: These last two patterns are RFC-1036 and RFC-1123.  Neither should have an optional timezone -- as seen in 
`ExtractDateUtils.PATTERN_RFC1036`, `ExtractDateUtils.PATTERN_RFC1123`, `DateTimeFormatter.RFC_1123_DATE_TIME`, and I looked at RFC-1036 spec as well.

Why did you make the time of day optional in ASCTIME?  I don't see that in ExtractDateUtils.  And can you add a test we parse "Sun Nov  6 08:49:37 1994"  (example date from RFC-2616 which is HTTP/1.1 spec which lists RFC-1123, RFC-1036, and asctime() -- the origin of why we see these particular patterns in ExtractDateUtils, borrowed from ApacheHttpClient).  The double-space before the single digit day is deliberate.  It may be necessary to use the 'p' (pad modifier) as specified in DateTimeFormatter.  I've seen conflicting information from internet searches on the "day" portion of asctime() as either 2-digit or 1; so it'd be good to test that either work.  "Leniency" will hopefully ensure one pattern works without needing to add more variations.

Good catch on noticing the seconds is optional in RFC-1123!

Can you reverse the order of these last 3 patterns?  Based on RFC-2616 (HTTP/1.1), this is the order defined by preference.

BTW if we really did want RFC-1123 & RFC-1036 patterns to have an optional timezone, then it would need to be specified differently than how you did it.  You put the optional start bracket to the right of the space when it would need to be to the left of it.

Obviously, all tweaks we do to these patterns need to be redone between 
* `solr/server/solr/configsets/_default/conf/solrconfig.xml` 
* `solr/core/src/test- files/solr/collection1/conf/solrconfig-parsing-update-processor-chains.xml`
* `solr/contrib/extraction/src/test-files/extraction/solr/collection1/conf/solrconfig.xml`
Probably elsewhere; I can check before committing.

barrotsteindev: I made time of day optional in asciitime so we can remove this configuration <str>yyyy-MM-dd</str>, which uses same without a time of day.
I will remove the optional timezone in RFC-1123 & RFC-1036, guess it is my bad.

barrotsteindev: Oh, now I recall why I added the optional time zone to asctime.
When the optional time zone is added, the pattern covers AKS and EDT date formats as well.
IMO, it might be beneficial to keep it in the default config.

barrotsteindev: >And can you add a test we parse "Sun Nov 6 08:49:37 1994"

BTW, I used "Fri Oct 7 13:14:15 2005" instead, since it uses the same date format, but there is a helper method to generate this instant in ParsingFieldUpdateProcessorsTest#inst20051007131415.
Hope it is OK.

barrotsteindev:
Added test for RFC-2616 (ParsingFieldUpdateProcessorsTest#testRfc2616) :-).

solr/core/src/test/org/apache/solr/update/processor/ParsingFieldUpdateProcessorsTest.java
```diff
@@ -992,6 +992,14 @@ public void testLenient() throws IOException {
     assertParsedDate("Friday Oct 7 13:14:15 2005", Date.from(inst20051007131415()), "parse-date-patterns-default-config");
   }
 
+  public void testRfc2616() throws Exception {
```
dsmiley: Should be named testAsctime since that's the format of this input you are testing.  When I mentioned RFC2616, that is a spec that in turn refers to multiple formats (including asctime).

Also, like I pointed out, need to test with 2 spaces left of a single digit day.

solr/core/src/test/org/apache/solr/update/processor/ParsingFieldUpdateProcessorsTest.java
```diff
@@ -992,6 +992,14 @@ public void testLenient() throws IOException {
...
+    assertParsedDate("Fri Oct 7 13:14:15 2005" , Date.from(inst20051007131415()), "parse-date-patterns-default-config");
+  }
+
+  public void testRfc2616Leniency() throws Exception {
```
dsmiley: Should be named testAsctimeLeniency (see my previous comment).  Glad to see this test works.

'course there  are already some tests with similar names... so feel free to combine some.  For example the "testLenient" method is an Ansi C test of leniency; by all means combine tests into one test method as appropriate.

dsmiley:
Another little detail I see that's different is the default locale.  ParseDateFieldUpdateProcessorFactory uses ROOT but ExtractDateUtils uses ENGLISH.  AFAIK, such a difference would appear when parsing timezones like AKST.  I could be wrong.  Can we see if this is indeed a problem?  Ideally a test would demonstrate if it is or not.  If we do find an issue here... then we could simply always configure the locale in the configs (my preference) or we could change the internal default of `ParseDateFieldUpdateProcessorFactory`

barrotsteindev:
Luckily for us the locale does not seem to cause a problem.
Unfortunately, the double space before single day digit test did not pass, forcing me to change the asctime configuration to "**<str>EEE MMM [ ]d HH:mm:ss [z ]yyyy</str>**",
which results in the test passing.

PR #442, opened 2018-08-29
Fixing an edge case bug when overriding a default PostingsSolrHighligher

Apmats:
Encountered this edge case issue. The impact of this should be limited but in our case (using a default PostingsSolrHighlighter) this was causing a weird issue.

The issue here is that, when passing the parameter hl.method=unified, we enter this switch, and end up in the UNIFIED case. But then we check whether the preconfigured highlighter is an instance of UnifiedSolrHighlighter. 
If the preconfigured highlighter is a PostingsSolrHighlighter, which is implemented for depracation reasons as a UnifiedSolrHighlighter, then that highlighter is used instead of a unified one (and effectively the hl.method parameter is ignored).

In our case, the additional parameters that the PostingsSolrHighlighter brings with it were causing an issue with highlighting down the line.

### DIFF
--- a/solr/core/src/java/org/apache/solr/handler/component/HighlightComponent.java
+++ b/solr/core/src/java/org/apache/solr/handler/component/HighlightComponent.java
@@ -189,7 +189,7 @@ public class HighlightComponent extends SearchComponent implements PluginInfoIni
 
     switch (method) {
       case UNIFIED:
-        if (solrConfigHighlighter instanceof UnifiedSolrHighlighter) {
+        if (solrConfigHighlighter.getClass() == UnifiedSolrHighlighter.class) {
           return solrConfigHighlighter;
         }
         return new UnifiedSolrHighlighter(); // TODO cache one?


janhoy:
Thanks for contributing this PR. This project requires that you open a JIRA issue at http://issues.apache.org/jira/browse/SOLR and link the two. After you do that, please link the two by adding the JIRA id to the title of this PR. I'm adding our new PR template to the description of this PR, please read through the checklist and check those that apply.

solr/core/src/java/org/apache/solr/handler/component/HighlightComponent.java
```diff
@@ -189,7 +189,7 @@ protected SolrHighlighter getHighlighter(SolrParams params) {
 
     switch (method) {
       case UNIFIED:
-        if (solrConfigHighlighter instanceof UnifiedSolrHighlighter) {
+        if (solrConfigHighlighter.getClass() == UnifiedSolrHighlighter.class) {
```
dsmiley: We must allow users to subclass UnifiedSolrHighlighter.  The solution to the conundrum here is to check that the class _also_  NOT a subclass of the PostingsSolrHighlighter.

Apmats: Ah, yes, you are indeed correct.

Apmats:
> I'd like to better understand why you are using the PostingsSolrHighlighter at all. I think it will be gone in 9.0. You _should_ be able to do what you need to do with the UnifiedSolrHighlighter.

I'm no longer working on the project where the issue was encountered, but I do think the highlighter was changed, partly to deal with this. I just wanted to avoid having someone waste time debugging the same issue before they switch.

Since this is being removed soon we can simply ignore this I believe, unless you think otherwise, in which case I'll go through the additional steps (correct code and open an issue to track this).

PR #443, opened 2018-08-30
SOLR-12722: add fl param to ChildDocTransformer

moshebla:


solr/core/src/test/org/apache/solr/response/transform/TestChildDocTransformerHierarchy.java
```diff
@@ -297,6 +297,35 @@ public void testNoChildren() throws Exception {
         "/response/docs/[0]/type_s==cake");
   }
 
+  @Test
+  public void testChildReturnFields() throws Exception {
+    indexSampleData(numberOfDocsPerNestedTest);
+
+    assertJQ(req("q", "test_s:testing",
+        "sort", "id asc",
+        "fl", "*, [child childFilter='lonely/lonelyGrandChild/test2_s:secondTest' fl='*, _nest_path_']",
```
dsmiley: Lets leave `_nest_path_` as internal and not mention it since testing "fl" has nothing to do with this field or nested documents.  So instead use some other field, foo_s or whatever.  And can you move this test to TestChildDocTransformer.java as it's not related to hierarchy?

moshebla: Sure thing,
just pushed a new commit :).

solr/solr-ref-guide/src/transforming-result-documents.adoc
```diff
@@ -137,6 +137,7 @@ When using this transformer, the `parentFilter` parameter must be specified, and
 
 * `childFilter` - query to filter which child documents should be included, this can be particularly useful when you have multiple levels of hierarchical documents (default: all children)
 * `limit` - the maximum number of child documents to be returned per parent document (default: 10)
+* `fl` - the field list which the transformer is to return.
```
dsmiley: ... "There is a further limitation in which the fields here should be a subset of those specified by the top level fl param."  And say defaults to the top level "fl" ?

solr/core/src/java/org/apache/solr/response/transform/ChildDocTransformerFactory.java
```diff
@@ -106,9 +107,12 @@ public DocTransformer create(String field, SolrParams params, SolrQueryRequest r
       }
     }
 
+    String childReturnFields = params.get("fl");
+    SolrReturnFields childSolrReturnFields = childReturnFields==null? new SolrReturnFields(): new SolrReturnFields(childReturnFields ,req);
```
dsmiley: Shouldn't this default to the top level "fl"?  I think so.  Even if it didn't do this before, it in-effect behaves this way and it makes sense (to me) that it would.

moshebla: That makes sense,
just pushed a commit :).

moshebla:
I tried bypassing the stack overflow error from the original commit.
Hope this is good enough.

dsmiley:
Oh right; we'd get an infinite loop as the ChildDocTransformerFactory is already being called by SolrReturnFields!  Ouch!

Sorry but I don't like your solution of trying to parse/strip the child transformer "fl"; it's very error-prone to do such things.

Lets define a ThreadLocal<Boolean> that allows us to know if we're being asked to create this transformer from within a recursive call to SolrReturnFields.  When we see this happening, we can return a no-op transformer (see RawValueTransformerFactory.NoopFieldTransformer which could be made public and moved to an inner class of DocTransformer).  This thread local would be set & cleared in a try-finally, and this logic would be in some utility method to keep it from being distracting to the caller.  In general I avoid ThreadLocals as it's a singleton design pattern which is usually bad, but it can be used sparingly/judiciously.

BTW it appears that doc transformers will not be executed on the child docs.  The transformer is not invoking them, and I suspect they won't be invoked on the way out to the client.  Can you check/test if this is true?  If I'm correct, this is a pre-existing bug/limitation that would be good to mark as a TODO; or you can try to address here since it's related (since you before couldn't specify what 'fl' was therefore it wasn't possible to specify a transformer until now).

solr/core/src/java/org/apache/solr/response/transform/ChildDocTransformer.java
```diff
@@ -131,6 +131,12 @@ public void transform(SolrDocument rootDoc, int rootDocId) {
 
           // load the doc
           SolrDocument doc = searcher.getDocFetcher().solrDoc(docId, childReturnFields);
```
moshebla: Perhaps this should be add to DocFetcher#solrDoc in the future?
This is just a thought.

moshebla:
BTW, I was skimming through DocTransformer, and noticed DocTransformer#needsSolrIndexSearcher.
Shouldn't this return true for ChildDocTransformer, since it searches the docId provided in the underlying index, and will fail if given a negative value?

solr/core/src/java/org/apache/solr/response/transform/ChildDocTransformer.java
dsmiley: I was thinking the same thing!  But yeah; out of scope for this PR.  CC @ErickErickson

solr/solrj/src/test/org/apache/solr/client/solrj/SolrExampleTests.java
```diff
@@ -1831,7 +1831,7 @@ public void testChildDoctransformer() throws IOException, SolrServerException {
       
       q = new SolrQuery("q", "*:*", "indent", "true");
       q.setFilterQueries(parentFilter);
-      q.setFields("id,[child parentFilter=\"" + parentFilter +
+      q.setFields("id, level_i, [child parentFilter=\"" + parentFilter +
```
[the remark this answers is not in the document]
moshebla: These tests fail since the recent changes made in this branch result the child doc transformer defaulting to the top fl. Beforehand, the default child fl was "*", which does not filter "level_i" from child docs. Should we default to "*" in lucene version < 8, and change this behaviour only once Solr 8 arrives?

solr/core/src/java/org/apache/solr/response/transform/ChildDocTransformer.java
```diff
@@ -132,6 +134,12 @@ public void transform(SolrDocument rootDoc, int rootDocId) {
...
+          if(childReturnFields.getTransformer() != null) {
+            if(childReturnFields.getTransformer().context != this.context) {
```
dsmiley: maybe this line should only check if context is null?  By checking if the context is different, it's suggestive there may already be a context but if that were true then we ought to reinstate the former context when done.  I suspect it's always going to be null.  Can you look?

solr/core/src/java/org/apache/solr/response/transform/ChildDocTransformerFactory.java
```diff
@@ -106,9 +123,20 @@ public DocTransformer create(String field, SolrParams params, SolrQueryRequest r
       }
     }
 
+    String childReturnFields = params.get("fl");
+    SolrReturnFields childSolrReturnFields;
+    if(childReturnFields != null) {
+      childSolrReturnFields = new SolrReturnFields(childReturnFields, req);
+    } else if(req.getSchema().getDefaultLuceneMatchVersion().major < 8) {
```
dsmiley: Nice

moshebla:
Closing this PR as it was committed to master in commit e4f256be15ca44f12a4aecb32c13d1ab2617cc00

PR #444, opened 2018-08-30
SOLR-12708: Aggregate failures from downstream async jobs; add error …

manokovacs:
…handling for RestoreCmd

### DIFF
--- a/solr/core/src/java/org/apache/solr/cloud/api/collections/RestoreCmd.java
+++ b/solr/core/src/java/org/apache/solr/cloud/api/collections/RestoreCmd.java
@@ -32,6 +32,9 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.Properties;
 import java.util.Set;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 
 import org.apache.solr.client.solrj.cloud.DistributedQueue;
 import org.apache.solr.client.solrj.cloud.autoscaling.PolicyHelper;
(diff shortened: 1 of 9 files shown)


solr/core/src/java/org/apache/solr/cloud/api/collections/RestoreCmd.java
```diff
@@ -271,7 +278,37 @@ public void call(ClusterState state, ZkNodeProps message, NamedList results) thr
           propMap.put(ASYNC, asyncId);
         }
         ocmh.addPropertyParams(message, propMap);
-        ocmh.addReplica(clusterState, new ZkNodeProps(propMap), new NamedList(), null);
+        final NamedList addResult = new NamedList();
+        ocmh.addReplica(clusterState, new ZkNodeProps(propMap), addResult, () -> {
+          countDownLatch.countDown();
```
tflobbe: I think we should move this to the end of the callback, or there can be a race condition

solr/core/src/java/org/apache/solr/cloud/api/collections/RestoreCmd.java
```diff
@@ -271,7 +278,37 @@ public void call(ClusterState state, ZkNodeProps message, NamedList results) thr
...
+              results.add("success", success);
+            }
+            success.addAll((NamedList) addResult.get("success"));
+          }
+        });
+      }
+
+      boolean allIsDone = countDownLatch.await(1, TimeUnit.HOURS);
+      if (!allIsDone) {
+        throw new TimeoutException("Initial replicas were not created within 10 minutes. Timing out.");
```
tflobbe: Outdated error, this is now 1 hour

solr/core/src/test-files/solr/configsets/cloud-minimal-faulty/conf/schema.xml
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8" ?>
```
tflobbe: Maybe there is a way to not have to create this new configset? either use the Config API, or the configset API, or maybe just updating one file? or maybe set a system property that one of the existing configsets will use to something invalid?

janhoy:
Closing since JIRA is closed

PR #446, opened 2018-09-05
LUCENE-8890: Improve parallel iteration of two lists of same length.

salsolatragus:
The class `BooleanWeight` takes a `BooleanQuery` (a list of `BooleanClause`s) as input and maintains a list of weights corresponding to the clauses. The clauses and the weights are iterated in parallel in various places throughout the class. At these code locations, it is not obvious that these two lists always have the same length, i.e., that the parallel iteration is safe. Moreover, the parallel iteration is not well supported by the Java language, which is why this operation is implemented differently throughout the code.

This patch joins the two lists to enable parallel iteration without managing two separate lists. This makes the code’s intent more obvious and prevents bugs due to the lists getting out of sync by a future change.

This problem was identified by our automated detector [MUDetect](https://github.com/stg-tud/MUDetect/). We would very much appreciate your feedback on our patch. Thank you very much!


lucene/core/src/java/org/apache/lucene/search/BooleanWeight.java
```diff
@@ -40,28 +39,40 @@
   /** The Similarity implementation. */
   final Similarity similarity;
   final BooleanQuery query;
-  
-  final ArrayList<Weight> weights;
+
+  private static class WeightedBooleanClause {
+    BooleanClause clause;
```
jpountz: I \*think\* we only need the BooleanClause#Occur here?

lucene/core/src/java/org/apache/lucene/search/BooleanWeight.java
```diff
@@ -40,28 +39,40 @@
...
+    Weight weight;
```
jpountz: let's make both members final?

salsolatragus: Done.

lucene/core/src/java/org/apache/lucene/search/BooleanWeight.java
salsolatragus: The code also uses some convenience methods on `BooleanClause` (`isRequired()`, `isScoring()`, and `isProhibited()`). Those are only checking the `BooleanCause#Occur`, but these checks are used elsewhere, so we should not inline them, I suppose. Therefore, we need the entire clause.

PR #451, opened 2018-09-14
LUCENE-8496: Add selective indexing to BKD/POINTS codec

nknize:
PR for [LUCENE-8496](https://issues.apache.org/jira/browse/LUCENE-8496) enabling fields to use the first `numIndexDimensions` from `numDataDimensions` while creating the BKD index.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/codecs/lucene60/Lucene60FieldInfosFormat.java
+++ b/lucene/core/src/java/org/apache/lucene/codecs/lucene60/Lucene60FieldInfosFormat.java
@@ -114,7 +114,7 @@ public final class Lucene60FieldInfosFormat extends FieldInfosFormat {
       Throwable priorE = null;
       FieldInfo infos[] = null;
       try {
-        CodecUtil.checkIndexHeader(input,
+        int version = CodecUtil.checkIndexHeader(input,
                                    Lucene60FieldInfosFormat.CODEC_NAME, 
                                    Lucene60FieldInfosFormat.FORMAT_START, 
                                    Lucene60FieldInfosFormat.FORMAT_CURRENT,
@@ -149,14 +149,13 @@ public final class Lucene60FieldInfosFormat extends FieldInfosFormat {
             attributes = lastAttributes;
           }
           lastAttributes = attributes;
-          int packedDimensionCount = input.readVInt();
-          int pointDataDimensionCount = 0x0000FFFF & packedDimensionCount;
-          int pointIndexDimensionCount = packedDimensionCount >>> 16;
-          if (pointIndexDimensionCount == 0) {
-            pointIndexDimensionCount = pointDataDimensionCount;
-          }
+          int pointDataDimensionCount = input.readVInt();
           int pointNumBytes;
+          int pointIndexDimensionCount = pointDataDimensionCount;
           if (pointDataDimensionCount != 0) {
+            if (version >= Lucene60FieldInfosFormat.FORMAT_SELECTIVE_INDEXING) {
+              pointIndexDimensionCount = input.readVInt();
+            }
             pointNumBytes = input.readVInt();
           } else {
             pointNumBytes = 0;
@@ -292,13 +291,9 @@ public final class Lucene60FieldInfosFormat extends FieldInfosFormat {
         output.writeByte(docValuesByte(fi.getDocValuesType()));
         output.writeLong(fi.getDocValuesGen());
         output.writeMapOfStrings(fi.attributes());
-        // max number of dimensions is <= 8; so for backcompat we can use the integer value to pack number of data and index dimensions
-        // where first two bytes are data dimension count, second two bytes are index dimension count
-        int pointDataDimensionCount = 0x0000FFFF & fi.getPointDataDimensionCount();
-        int pointIndexDimensionCount = fi.getPointIndexDimensionCount();
-        int packedDimensionCount = (pointIndexDimensionCount << 16) | pointDataDimensionCount;
-        output.writeVInt(packedDimensionCount);
-        if (pointDataDimensionCount != 0) {
+        output.writeVInt(fi.getPointDataDimensionCount());
+        if (fi.getPointDataDimensionCount() != 0) {
+          output.writeVInt(fi.getPointIndexDimensionCount());
           output.writeVInt(fi.getPointNumBytes());
         }
       }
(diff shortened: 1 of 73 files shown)


lucene/core/src/java/org/apache/lucene/util/bkd/BKDWriter.java
```diff
@@ -981,15 +988,15 @@ public long finish(IndexOutput out) throws IOException {
...
+    PathSlice[] sortedPointWriters = new PathSlice[numDataDims];
 
     // This is only used on exception; on normal code paths we close all files we opened:
     List<Closeable> toCloseHeroically = new ArrayList<>();
 
     boolean success = false;
     try {
       //long t0 = System.nanoTime();
-      for(int dim=0;dim<numDims;dim++) {
+      for(int dim=0;dim<numDataDims;dim++) {
```
iverase: I think we should only create sorted point writers for indexed dims. There is no point sort data dims as it only creates overhead when writing the tree.

lucene/core/src/java/org/apache/lucene/util/bkd/BKDReader.java
```diff
@@ -52,10 +53,17 @@
...
     version = CodecUtil.checkHeader(in, BKDWriter.CODEC_NAME, BKDWriter.VERSION_START, BKDWriter.VERSION_CURRENT);
-    numDims = in.readVInt();
+    int packedDims = in.readVInt();
+    if (version >= BKDWriter.VERSION_LEAF_STORES_BOUNDS) {
+      numDataDims = packedDims & 0x0000FFFF;
+      numIndexDims = packedDims >>> 16;
+    } else {
+      numDataDims = packedDims;
+      numIndexDims = numDataDims;
+    }
```
iverase: Should we create a new version of the tree for this change and not mix it up with VERSION_LEAF_STORES_BOUNDS?

lucene/core/src/test/org/apache/lucene/util/bkd/TestBKD.java
```diff
@@ -1063,7 +1064,7 @@ public void testWastedLeadingBytes() throws Exception {
 
     Directory dir = newFSDirectory(createTempDir());
     int numDocs = 100000;
-    BKDWriter w = new BKDWriter(numDocs+1, dir, "tmp", numDims, bytesPerDim, 32, 1f, numDocs, true);
+    BKDWriter w = new BKDWriter(numDocs+1, dir, "tmp", numDims, numDims, bytesPerDim, 32, 1f, numDocs, true);
```
iverase: Couldn't we add a mode on the random tests where dataDim != indexedDim. The test can only check that the data dims are properly propagated to the leaf nodes.

iverase: I have just seen there is already a test using different index and data dimensions, still it would be worthy to check if possible that data dimensions are properly propagated.

lucene/core/src/java/org/apache/lucene/codecs/lucene60/Lucene60FieldInfosFormat.java
```diff
@@ -149,9 +149,14 @@ public FieldInfos read(Directory directory, SegmentInfo segmentInfo, String segm
             attributes = lastAttributes;
           }
           lastAttributes = attributes;
-          int pointDimensionCount = input.readVInt();
+          int packedDimensionCount = input.readVInt();
+          int pointDataDimensionCount = 0x0000FFFF & packedDimensionCount;
+          int pointIndexDimensionCount = packedDimensionCount >>> 16;
+          if (pointIndexDimensionCount == 0) {
+            pointIndexDimensionCount = pointDataDimensionCount;
+          }
```
jpountz: could we use a new version number instead to handle backcompat?

lucene/core/src/java/org/apache/lucene/util/bkd/BKDReader.java
```diff
@@ -52,10 +53,17 @@
   /** Caller must pre-seek the provided {@link IndexInput} to the index location that {@link BKDWriter#finish} returned */
   public BKDReader(IndexInput in) throws IOException {
     version = CodecUtil.checkHeader(in, BKDWriter.CODEC_NAME, BKDWriter.VERSION_START, BKDWriter.VERSION_CURRENT);
-    numDims = in.readVInt();
+    int packedDims = in.readVInt();
+    if (version >= BKDWriter.VERSION_LEAF_STORES_BOUNDS) {
+      numDataDims = packedDims & 0x0000FFFF;
+      numIndexDims = packedDims >>> 16;
```
jpountz: let's read two vints instead?

lucene/core/src/java/org/apache/lucene/util/bkd/BKDWriter.java
nknize: I made the change but then I remembered that the leaf nodes (which write all dimensions) require even the non indexed dimensions be sorted for compression (in order to compute the common prefix). I could explore delaying that sort operation only when its needed, but I think the recursion still needs sorted point writers passed through PathSlice[] so this is still necessary?

lucene/core/src/test/org/apache/lucene/util/bkd/TestBKD.java
iverase: Thanks @nknize! 
Would it be possible to add the following assertion in the method for IntersectVisitor#visit. That would signal that we are propagating the index and data dimensions to the method?

```
   byte[][] expectedDocs = docValues[docID];
   for (int dim =0; dim < numDataDims; dim++) {
        assertTrue(FutureArrays.compareUnsigned(packedValue, dim * numBytesPerDim, dim * numBytesPerDim + numBytesPerDim, expectedDocs[dim], 0, numBytesPerDim) == 0);
    }
```

nknize:
Updated the PR and attached an update patch to [LUCENE-8496](https://issues.apache.org/jira/browse/LUCENE-8496) along with a patch that modifies `LatLonShape.Triangle` encoding to take advantage of selective indexing. I think this is close. Wouldn't mind another round of review along with thoughts on whether we should commit this for a Lucene 7.6 release or hold off for 8.0.

lucene/core/src/test/org/apache/lucene/util/bkd/TestBKD.java
iverase: I have realised that the above assertion is incorrect for multivalued docs so please discard it. Maybe we can just assert the number of bytes in packedValue.

iverase:
I have only one comment for this version that I think we should address. We are allocation two byte arrays for each leaf we are visiting and I think we can prevent it. Apart from that it looks there!

Regarding the Lucene version I have no strong opinion. Because it is a backward compatible change I think there is no harm to commit it to Lucene 7.x branch as well.

nknize:
closing with commit 1118299c338253cea09640acdc48dc930dc27fda

PR #455, opened 2018-09-17
SOLR-12638

moshebla:


moshebla:
Currently Some Solr Core tests do not pass.
I will work on those, since this is a WIP.
The test that concerns me the most is ConvertedLegacyTest, which fails only if VersionInfo#lookupVersion is called.

dsmiley:
ConvertedLegacyTest fails in part because line 306 adds with overwrite=false -- a bit of a dubious Solr feature that is probably not properly compatible with the UpdateLog which makes various assumptions about the uniqueKey being unique.  I'll email the dev list to see what others think but I'm inclined to think overwrite=false ought to be explicitly forbidden with an UpdateLog in place.  That ancient legacy test can use a config that doesn't have an updateLog.

solr/core/src/java/org/apache/solr/update/processor/DistributedUpdateProcessor.java
```diff
@@ -1184,7 +1196,16 @@ protected boolean versionAdd(AddUpdateCommand cmd) throws IOException {
 
         // TODO: possibly set checkDeleteByQueries as a flag on the command?
         doLocalAdd(cmd);
-        
+
+        if(lastKnownVersion != null && req.getSchema().isUsableForChildDocs() &&
```
dsmiley: This may very well be right but can you tell me why you added this delete here at this line and why the delete is version-dependent?

moshebla: This delete ensures the updated block is the latest block, deleting all previous ones.

moshebla:
Currently the following update operations are supported for child documents:

- add
- set
- remove

solr/core/src/test/org/apache/solr/update/processor/AtomicUpdateBlockTest.java
```diff
@@ -181,6 +182,180 @@ public void testBlockRealTimeGet() throws Exception {
...
+    assertU(commit());
+
+    assertJQ(req("qt","/get", "id","1", "fl","id, cat_ss, child1, [child]")
+        ,"=={\"doc\":{'id':\"1\"" +
+            ", cat_ss:[\"aaa\",\"ccc\"], child1:[{\"id\":\"2\",\"cat_ss\":[\"child\"]}]" +
+            "       }}"
+    );
+
+    doc = sdoc("id", "1",
+        "cat_ss", ImmutableMap.of("set", Arrays.asList("aaa", "bbb")),
```
dsmiley: Minor: in cases where the JDK has alternatives to Guava, use them.  Here, it's `Collections.singletonMap` I recall.

solr/core/src/java/org/apache/solr/update/processor/AtomicUpdateDocumentMerger.java
```diff
@@ -461,5 +466,33 @@ private static boolean isChildDoc(Object obj) {
...
+        return;
+      }
+    }
+  }
+
+  private static boolean isDerivedFromDoc(SolrInputDocument fullDoc, SolrInputDocument subDoc) {
+    for(SolrInputField subSif: subDoc) {
+      String fieldName = subSif.getName();
+      if(!fullDoc.containsKey(fieldName)) return false;
+      Collection<Object> fieldValues = subDoc.getFieldValues(fieldName);
```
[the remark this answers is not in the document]
moshebla: Oh yes you are correct.
I will get to this and test this ASAP.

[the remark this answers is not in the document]
moshebla: fieldValues was supposed to be:
`fullDoc.getFieldValues(fieldName);`

solr/core/src/java/org/apache/solr/update/processor/AtomicUpdateDocumentMerger.java
```diff
@@ -461,5 +466,33 @@ private static boolean isChildDoc(Object obj) {
...
+  private static void removeChildDoc(Collection original, SolrInputDocument docToRemove) {
+    for(SolrInputDocument doc: (Collection<SolrInputDocument>) original) {
+      if(isDerivedFromDoc(doc, docToRemove)) {
+        original.remove(doc);
+        return;
+      }
+    }
+  }
+
+  private static boolean isDerivedFromDoc(SolrInputDocument fullDoc, SolrInputDocument subDoc) {
```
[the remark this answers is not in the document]
moshebla: Exactly,
I will add Javadocs for this.

solr/core/src/java/org/apache/solr/update/processor/AtomicUpdateDocumentMerger.java
```diff
@@ -461,5 +466,33 @@ private static boolean isChildDoc(Object obj) {
...
+      removeChildDoc(original, (SolrInputDocument) toRemove);
+    } else {
+      original.remove(getNativeFieldValue(fieldName, toRemove));
+    }
+  }
+
+  private static void removeChildDoc(Collection original, SolrInputDocument docToRemove) {
+    for(SolrInputDocument doc: (Collection<SolrInputDocument>) original) {
+      if(isDerivedFromDoc(doc, docToRemove)) {
+        original.remove(doc);
```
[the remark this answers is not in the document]
moshebla: I was thinking that if doc {"id": "4"} was supplied, then the **whole** child doc with that id(4) is to be removed.
Or perhaps you have a different scenario in mind, where a field of a a particular child document is to be removed while the rest of the child document is left intact?

solr/core/src/java/org/apache/solr/update/processor/AtomicUpdateDocumentMerger.java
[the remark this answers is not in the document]
dsmiley: I see the new docs; still it's unclear without the context of our conversation.  Should "subDoc" be "partialDoc"?  How is it that the fullDoc is "the document to be tested" when the return value is "wether subDoc is a subset of fullDoc"?  Isn't the subject of the test subDoc?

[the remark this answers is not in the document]
moshebla: Just pushed a new commit to address your comments.

solr/core/src/java/org/apache/solr/handler/component/RealTimeGetComponent.java
```diff
@@ -639,12 +650,30 @@ public static SolrInputDocument getInputDocument(SolrCore core, BytesRef idBytes
...
-        if (onlyTheseNonStoredDVs != null) {
-          docFetcher.decorateDocValueFields(sid, docid, onlyTheseNonStoredDVs);
-        } else {
-          docFetcher.decorateDocValueFields(sid, docid, docFetcher.getNonStoredDVsWithoutCopyTargets());
+        ensureDocDecorated(onlyTheseNonStoredDVs, sid, docid, docFetcher, resolveBlock || schema.hasExplicitField(IndexSchema.NEST_PATH_FIELD_NAME));
+        SolrInputField rootField;
+        if(resolveBlock && schema.isUsableForChildDocs() && (rootField = sid.getField(IndexSchema.ROOT_FIELD_NAME))!=null) {
+          // doc is part of a nested structure
+          ModifiableSolrParams params = new ModifiableSolrParams()
+              .set("q", core.getLatestSchema().getUniqueKeyField().getName()+ ":" +rootField.getFirstValue())
```
dsmiley: It seems the LocalSolrQueryRequest here is a dummy needed to satisfy some of the methods below.  This threw me; there should be comments and/or choice of var names (e.g. dummyReq) to reflect this.  "q" isn't needed; just the "fl".  It seems we don't even need the "fl" here since that can be supplied as the first parameter to SolrReturnFields, which seems better if it works.

solr/core/src/java/org/apache/solr/handler/component/RealTimeGetComponent.java
```diff
@@ -661,6 +690,21 @@ public static SolrInputDocument getInputDocument(SolrCore core, BytesRef idBytes
     return sid;
   }
 
+  private static void ensureDocDecorated(Set<String> onlyTheseNonStoredDVs, SolrDocumentBase doc, int docid, SolrDocumentFetcher docFetcher) throws IOException {
```
dsmiley: I suggest renaming these methods `ensureDocDecorated` since it's what it calls.

solr/core/src/java/org/apache/solr/update/processor/DistributedUpdateProcessor.java
```diff
@@ -1360,24 +1385,47 @@ boolean getUpdatedDocument(AddUpdateCommand cmd, long versionOnUpdate) throws IO
...
 
 
-    cmd.solrDoc = docMerger.merge(sdoc, oldDoc);
+    SolrInputDocument mergedDoc;
+    if(idField == null || blockDoc == null) {
+      // create a new doc by default if an old one wasn't found
+      mergedDoc = docMerger.merge(sdoc, new SolrInputDocument());
+    } else {
+      if(isNestedSchema && req.getSchema().hasExplicitField(IndexSchema.NEST_PATH_FIELD_NAME) &&
+          blockDoc.containsKey(IndexSchema.ROOT_FIELD_NAME) && !id.utf8ToString().equals(blockDoc.getFieldValue(IndexSchema.ROOT_FIELD_NAME))) {
```
dsmiley: I don't think we can assume id.utf8ToString() is correct.  I think we have to consult the corresponding FieldType to get the "external value".  Also, cast blockDoc.getFieldValue as a String to make it clear we expected it to be one.

solr/core/src/java/org/apache/solr/update/processor/DistributedUpdateProcessor.java
```diff
@@ -1360,24 +1385,47 @@ boolean getUpdatedDocument(AddUpdateCommand cmd, long versionOnUpdate) throws IO
...
+        SolrInputDocument oldDoc = RealTimeGetComponent.getInputDocument(cmd.getReq().getCore(), id, null,
+            false, null, true, false);
+        mergedDoc = docMerger.merge(sdoc, oldDoc);
+        String docPath = (String) mergedDoc.getFieldValue(IndexSchema.NEST_PATH_FIELD_NAME);
+        List<String> docPaths = StrUtils.splitSmart(docPath, PATH_SEP_CHAR);
+        SolrInputField replaceDoc = blockDoc.getField(docPaths.remove(0).replaceAll(PATH_SEP_CHAR + "|" + NUM_SEP_CHAR, ""));
```
dsmiley: The logic here (not just this line) is non-obvious to me.  There are no comments.  Please add comments and maybe refactor out a method.  The use of replaceAll with a regexp is suspicious to me.  None of the tests you added triggered a breakpoint within the docPaths loop below.  Needs more testing or possible error.

dsmiley:
As a general point, I feel we should prefer "nested" terminology instead of "block".  If we were purely working within Lucene then I think "block" might be okay but at the Solr layer people see this stuff as "nested" docs.

solr/core/src/test/org/apache/solr/update/processor/AtomicUpdateBlockTest.java
```diff
@@ -59,39 +52,147 @@ public void before() {
...
-        ,"/response/numFound==0"
+    assertU(commit());
+
+    assertJQ(req("q", "_root_:1"),
+        "/response/numFound==11");
+
+    assertJQ(req("q", "string_s:child", "fl", "*", "rows", "1000000"),
+        "/response/numFound==10");
+
+    // ensure updates work when block has more than 10 children
```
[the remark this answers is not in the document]
moshebla: This is to ensure the limit=-1 is passed to childDocumentTransformer, to retrieve all the block's child documents

solr/core/src/java/org/apache/solr/update/AddUpdateCommand.java
```diff
@@ -262,6 +263,11 @@ private void flattenAnonymous(List<SolrInputDocument> unwrappedDocs, SolrInputDo
     flattenAnonymous(unwrappedDocs, currentDoc, false);
   }
 
+  public String getRouteFieldVal() {
```
dsmiley: I'm skeptical of this method.  It's name seems innocent enough looking at the code here.  But then also consider some collections have a "router.field" and this method is named in such a way that one would think this method returns that field's value... yet it does not.  Some callers put this into a variable named "id" or similar.  Given that, I propose you remove it but incorporate the logic into getHashableId which seems the proper place for it.  It _is_ the hashable Id... the hashable ID of a nested doc is it's root.

But... I do also wonder if we need this at all.  Somewhere Solr already has code that looks at \_route\_ and acts on that if present.  Perhaps the code path for an atomic update isn't doing this yet but should do it?  Then we wouldn't need this change to AddUpdateCommand.

moshebla: After reading through the code it seems like _route_ is currently only used in delete commands and queries.
This can be seen in DistributedUpdateProceesor#doDeleteByQuery
`ModifiableSolrParams outParams = new ModifiableSolrParams(filterParams(req.getParams()));
      outParams.set(DISTRIB_UPDATE_PARAM, DistribPhase.TOLEADER.toString());
      outParams.set(DISTRIB_FROM, ZkCoreNodeProps.getCoreUrl(
          zkController.getBaseUrl(), req.getCore().getName()));

      SolrParams params = req.getParams();
      String route = params.get(ShardParams._ROUTE_);
      Collection<Slice> slices = coll.getRouter().getSearchSlices(route, params, coll);

      List<Node> leaders =  new ArrayList<>(slices.size());
      for (Slice slice : slices) {
        String sliceName = slice.getName();
        Replica leader;
        try {
          leader = zkController.getZkStateReader().getLeaderRetry(collection, sliceName);
        } catch (InterruptedException e) {
          throw new SolrException(ErrorCode.SERVICE_UNAVAILABLE, "Exception finding leader for shard " + sliceName, e);
        }`
Perhaps this should be moved to addUpdateCommand(perhaps even UpdateCommand)
or to a method that will be accessible and visible so we do not have this confusion in the future?

dsmiley: What code duplication?

I think we need to standardize/harmonize route a bit.  Notice UpdateCommand has setRoute & getRoute.  Lets initialize route in the constructor here by the presence of \_route\_ in the params.  Then lets not look for _route_ in params later since we know we can get it here.  Then I think `org.apache.solr.update.processor.DistributedUpdateProcessor#setupRequest(java.lang.String, org.apache.solr.common.SolrInputDocument)` can be removed and instead insist everyone call the overloaded version that takes a route, and each caller looks up the route from the command.  It's not clear to me if "null" actually should be passed to route in any circumstance.

solr/core/src/java/org/apache/solr/handler/component/RealTimeGetComponent.java
```diff
@@ -609,9 +618,10 @@ public static SolrInputDocument getInputDocument(SolrCore core, BytesRef idBytes
    * @param resolveFullDocument In case the document is fetched from the tlog, it could only be a partial document if the last update
    *                  was an in-place update. In that case, should this partial document be resolved to a full document (by following
    *                  back prevPointer/prevVersion)?
+   * @param resolveBlock Check whether the document is part of a block. If so, return the whole block.
    */
   public static SolrInputDocument getInputDocument(SolrCore core, BytesRef idBytes, AtomicLong versionReturned, boolean avoidRetrievingStoredFields,
-      Set<String> onlyTheseNonStoredDVs, boolean resolveFullDocument) throws IOException {
+      Set<String> onlyTheseNonStoredDVs, boolean resolveFullDocument, boolean resolveBlock) throws IOException {
     SolrInputDocument sid = null;
```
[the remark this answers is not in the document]
dsmiley: @moshebla I'm concerned about this method having so many parameters... it's a code smell.
resolveBlock & resolveChildren booleans... not sure how they differ.  Isn't resolveChildren enough?  if resolveChildren true, then isn't effectively resolveFullDocument also true?  (if so the constraint could be documented and enforced with an assertion).  

versionReturned is a dubious parameter I'm skeptical needs to be here.  I'm aware you didn't add it, but the number of parameters is getting troublingly long with your changes; it's good to review what's needed.

solr/core/src/java/org/apache/solr/handler/component/RealTimeGetComponent.java
```diff
@@ -609,9 +618,11 @@ public static SolrInputDocument getInputDocument(SolrCore core, BytesRef idBytes
    * @param resolveFullDocument In case the document is fetched from the tlog, it could only be a partial document if the last update
    *                  was an in-place update. In that case, should this partial document be resolved to a full document (by following
    *                  back prevPointer/prevVersion)?
+   * @param resolveRootDoc Check whether the document is part of a nested hierarchy. If so, return the whole hierarchy.
+   * @param resolveChildren Check whether the document has child documents. If so, return the document including its children.
    */
   public static SolrInputDocument getInputDocument(SolrCore core, BytesRef idBytes, AtomicLong versionReturned, boolean avoidRetrievingStoredFields,
-      Set<String> onlyTheseNonStoredDVs, boolean resolveFullDocument) throws IOException {
+      Set<String> onlyTheseNonStoredDVs, boolean resolveFullDocument, boolean resolveRootDoc, boolean resolveChildren) throws IOException {
```
moshebla: resolveRootDoc -> resolves the whole hierarchy from the root doc regardless of whether the docId provided was the root document.
resolveChildren -> resolves the specified document's child documents, regardless of whether the document was is a root document or not.

solr/core/src/java/org/apache/solr/handler/component/RealTimeGetComponent.java
```diff
@@ -639,12 +651,32 @@ public static SolrInputDocument getInputDocument(SolrCore core, BytesRef idBytes
...
+        if((resolveChildren || resolveRootDoc) && schema.isUsableForChildDocs() && rootField!=null) {
+          // doc is part of a nested structure
+          String id = resolveRootDoc? (String) rootField.getFirstValue(): (String) sid.getField(idField.getName()).getFirstValue();
+          ModifiableSolrParams params = new ModifiableSolrParams()
+              .set("fl", "*, _nest_path_, [child]")
+              .set("limit", "-1");
+          SolrQueryRequest nestedReq = new LocalSolrQueryRequest(core, params);
+          final BytesRef rootIdBytes = new BytesRef(id);
+          final int rootDocId = searcher.getFirstMatch(new Term(idField.getName(), rootIdBytes));
+          final DocTransformer childDocTransformer = TransformerFactory.defaultFactories.get("child").create("child", params, nestedReq);
```
dsmiley: no, use `core.getTransformerFactory("child")`

solr/core/src/java/org/apache/solr/update/processor/AtomicUpdateDocumentMerger.java
```diff
@@ -259,9 +264,8 @@ public boolean doInPlaceUpdateMerge(AddUpdateCommand cmd, Set<String> updatedFie
     SolrInputDocument oldDocument = RealTimeGetComponent.getInputDocument
       (cmd.getReq().getCore(), idBytes,
        null, // don't want the version to be returned
-       true, // avoid stored fields from index
        updatedFields,
-       true); // resolve the full document
+       RealTimeGetComponent.Resolution.FULL_DOC); // get the partial document for the in-place update
```
dsmiley: it is not a partial document; the original comment was correct.  Now with an enum, we don't even need a comment as the enum has a self explanatory name.

solr/core/src/java/org/apache/solr/handler/component/RealTimeGetComponent.java
```diff
@@ -1164,6 +1225,31 @@ public void processGetUpdates(ResponseBuilder rb) throws IOException
...
+   *  <p>
+   *    Lookup strategy for {@link #getInputDocument(SolrCore, BytesRef, AtomicLong, Set, Resolution)}.
+   *  </p>
+   *  <ul>
+   *    <li>{@link #FULL_DOC}</li>
+   *    <li>{@link #DOC_CHILDREN}</li>
+   *    <li>{@link #FULL_HIERARCHY}</li>
+   *  </ul>
+   */
+  public static enum Resolution {
```
dsmiley: Suggested new names:  "DOC", "DOC_WITH_CHILDREN", "ROOT_WITH_CHILDREN".
Resolution is now only used where there is no "in place" so no need to mention partial documents.

moshebla:
I rebased using the latest master.
I also changed the tests so _root_ is not hard-coded into childless documents, since SOLR-5211 was committed to master and ensures this behaviour.

dsmiley:
What does "so _root_ is not hard-coded into childless documents" mean?  If you think SOLR-5211 is missing something, can you please post a patch file to the issue and explain what you're adding?

solr/core/src/test/org/apache/solr/cloud/AtomicUpdateBlockShardedTest.java
```diff
@@ -130,7 +130,7 @@ public void doNestedInplaceUpdateTest() throws Exception {
     // for now,  we know how ranges will be distributed to shards.
     // may have to look it up in clusterstate if that assumption changes.
 
-    SolrInputDocument doc = sdoc("id", "a", "level_s", "root", "_root_", "a");
+    SolrInputDocument doc = sdoc("id", "a", "level_s", "root");
```
moshebla: @dsmiley, this is what I meant by "remove hard coded _root_.
I added these since my logic depended on chidless documents having _root_.
When SOLR-5211 was merged, I just rebased on master and removed these hardcoded _root_ fields, which are now automatically generated by Solr.

moshebla:
Merged with latest changes from master

moshebla:
Resolved conflicts with master

dsmiley:
* The change to `DocTermOrds` appears to be erroneous related to getting up to date with master.  Same for `the-terms-component.adoc`
* AddUpdateCommand.getHashableId calls getRootIdUsingRouteParam which calls getHashableId -- seems like a recipe for an infinite loop.  Even if it turns out to never happen, lets improve this.

Just curious; what's your opinion on the refactor of DistributedUpdateProcessor?  Did it help make your edits here bettor or worse or same?

More feedback to follow... will run tests.

moshebla:
>     * The change to `DocTermOrds` appears to be erroneous related to getting up to date with master.  Same for `the-terms-component.adoc`
> 
>     * AddUpdateCommand.getHashableId calls getRootIdUsingRouteParam which calls getHashableId -- seems like a recipe for an infinite loop.  Even if it turns out to never happen, lets improve this.
> 
> 
> Just curious; what's your opinion on the refactor of DistributedUpdateProcessor? Did it help make your edits here bettor or worse or same?
> 
> More feedback to follow... will run tests.

I find it easier to scroll through DistributedUpdateProcessor, now that all the zk logic has been moved to a separate class. :+1:

solr/core/src/java/org/apache/solr/update/processor/DistributedUpdateProcessor.java
```diff
@@ -815,31 +871,7 @@ protected void versionDeleteByQuery(DeleteUpdateCommand cmd) throws IOException
     vinfo.blockUpdates();
     try {
 
-      if (versionsStored) {
```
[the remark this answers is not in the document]
moshebla: It's simply a refactoring.

solr/core/src/test-files/solr/collection1/conf/solrconfig-block-atomic-update.xml
```diff
@@ -0,0 +1,64 @@
...
+      <str name="dir">${solr.ulog.dir:}</str>
+    </updateLog>
+
+    <commitWithin>
+      <softCommit>${solr.commitwithin.softcommit:true}</softCommit>
+    </commitWithin>
+
+  </updateHandler>
+
+  <updateProcessor class="solr.NestedUpdateProcessorFactory" name="nested-processor"/>
```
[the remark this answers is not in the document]
moshebla: The one major difference in this schema that sets it apart from most schemas is the fact that `_root_` is indexed and stored.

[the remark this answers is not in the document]
dsmiley: Could that be avoided then?  Less test files is a good thing.   If can’t be then this distinction should be documented

solr/core/src/test/org/apache/solr/cloud/AtomicUpdateBlockShardedTest.java
```diff
@@ -0,0 +1,187 @@
...
+      SolrDocument results = (SolrDocument) routeRsp.getResponse().get("doc");
+      assertNotNull("RTG should find doc because _route_ was set to the root documents' ID", results);
+      assertEquals("b", results.getFieldValue("id"));
+
+      // assert all docs are indexed under the same root
+      assertEquals(0, client.query(new ModifiableSolrParams().set("q", "-_root_:a")).getResults().size());
+
+      // assert all docs are indexed inside the same block
+      QueryResponse rsp = client.query(params("qt","/get", "id","a", "fl", "*, [child]"));
+      SolrDocument val = (SolrDocument) rsp.getResponse().get("doc");
```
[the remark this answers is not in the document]
moshebla: In each iteration there the "inplace_updatable_int" is increased,
and another doc is added under the `grandChildren` field.
I could generate a method which takes the number if iterations executed as an input and constructs the string for the document which should be returned.
Would that be any better though?
I am not sure.
WDYT?

solr/core/src/java/org/apache/solr/update/processor/DistributedUpdateProcessor.java
```diff
@@ -645,32 +660,78 @@ boolean getUpdatedDocument(AddUpdateCommand cmd, long versionOnUpdate) throws IO
...
     }
 
 
-    cmd.solrDoc = docMerger.merge(sdoc, oldDoc);
+    SolrInputDocument mergedDoc;
+    if(idField == null || nestedDoc == null) {
+      // create a new doc by default if an old one wasn't found
+      mergedDoc = docMerger.merge(sdoc, new SolrInputDocument());
+    } else {
+      if(isDeeplyNestedSchema && nestedDoc.containsKey(IndexSchema.ROOT_FIELD_NAME) &&
```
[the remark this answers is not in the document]
moshebla: I separated this part into a few methods, some have  in-line comments,
while others have javaDocs

solr/core/src/java/org/apache/solr/update/processor/DistributedUpdateProcessor.java
```diff
@@ -645,32 +660,126 @@ boolean getUpdatedDocument(AddUpdateCommand cmd, long versionOnUpdate) throws IO
     }
     
     // full (non-inplace) atomic update
+    final boolean isDeeplyNestedSchema = req.getSchema().isUsableForChildDocs() && req.getSchema().hasExplicitField(IndexSchema.NEST_PATH_FIELD_NAME);
     SolrInputDocument sdoc = cmd.getSolrInputDocument();
     BytesRef id = cmd.getIndexedId();
-    SolrInputDocument oldDoc = RealTimeGetComponent.getInputDocument(cmd.getReq().getCore(), id);
+    SolrInputDocument completeHierarchy = RealTimeGetComponent.getInputDocument(cmd.getReq().getCore(), id, RealTimeGetComponent.Resolution.ROOT_WITH_CHILDREN);
```
dsmiley: I think "oldDocWithChildren" would be a better name.  The semantics that it's the old document is important.

solr/core/src/java/org/apache/solr/update/processor/DistributedUpdateProcessor.java
```diff
@@ -1032,6 +1151,24 @@ protected void assertNotFinished() {
     finished = true;
   }
 
+  /**
+   *
+   * @return whether this update changes a value of a block
+   */
+  public static boolean shouldRefreshUlogCaches(AddUpdateCommand cmd) {
```
dsmiley: Can we make this logic more efficient by adding a field to AddUpdateCommand that declares (a) is this update an atomic update, and furthermore, did it involve nested documents?  Perhaps two separate `Boolean`s, null meaning not yet calculated.  By this juncture in the code I think we've already determined these things and I'd like to avoid potential overhead per document going into Solr.  Lambda's can add up.

typo "update.q" => "update"

solr/core/src/test-files/solr/collection1/conf/solrconfig-block-atomic-update.xml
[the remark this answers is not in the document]
dsmiley: BTW your comment here is about the schema but this review thread is on solrconfig.xml which I think could be a stock/basic one instead of yet another test config file.

solr/core/src/test/org/apache/solr/cloud/NestedShardedAtomicUpdateTest.java
```diff
@@ -0,0 +1,191 @@
...
+    SolrInputDocument doc = sdoc("id", "a", "level_s", "root");
+
+    final SolrParams params = new ModifiableSolrParams().set("update.chain", "nested-rtg").set("wt", "json").set("_route_", "a");
+
+    int which = (params.get("_route_").hashCode() & 0x7fffffff) % clients.size();
+    SolrClient aClient = clients.get(which);
+
+    indexDoc(aClient, params, doc);
+
+    aClient.commit();
```
dsmiley: again; some of these commits are superfluous and would better be randomized so exercise different internal code paths.

solr/core/src/test/org/apache/solr/cloud/NestedShardedAtomicUpdateTest.java
```diff
@@ -0,0 +1,191 @@
...
+      // assert all docs are indexed inside the same block
+      QueryResponse rsp = client.query(params("qt","/get", "id","a", "fl", "*, [child]"));
+      SolrDocument val = (SolrDocument) rsp.getResponse().get("doc");
+      assertEquals("a", val.getFieldValue("id"));
+      List<SolrDocument> children = (List) val.getFieldValues("children");
+      assertEquals(1, children.size());
+      SolrDocument childDoc = children.get(0);
+      assertEquals("b", childDoc.getFieldValue("id"));
+      List<SolrDocument> grandChildren = (List) childDoc.getFieldValues("grandChildren");
+      assertEquals(++i, grandChildren.size());
```
dsmiley: aha; I see now.  This is kind of a strange loop... looping over clients and also incrementing an 'i' that is used to try a different doc that has various number of children.  I think the loop would be clearer if you were looping over the 'i' values, and then when you need a client then grab one randomly (since it doesn't matter which client you use).  Also maybe use "idIndex" instead of "i" if you also think that's clearer.

solr/core/src/test/org/apache/solr/cloud/NestedShardedAtomicUpdateTest.java
```diff
@@ -0,0 +1,191 @@
...
+    doc = sdoc("id", "a", "children", map("add", sdocs(sdoc("id", "b", "level_s", "child"))));
+
+    indexDocAndRandomlyCommit(aClient, params, doc);
+
+    doc = sdoc("id", "b", "grandChildren", map("add", sdocs(sdoc("id", ids[0], "level_s", "grand_child"))));
+
+    indexDocAndRandomlyCommit(aClient, params, doc);
+
+    int i = 0;
+    for (SolrClient client : clients) {
```
dsmiley: same feedback as earlier

solr/core/src/test/org/apache/solr/update/processor/NestedAtomicUpdateTest.java
```diff
@@ -0,0 +1,659 @@
...
+  }
+
+  @Test
+  public void testBlockAtomicAdd() throws Exception {
+
+    SolrInputDocument doc = sdoc("id", "1",
+        "cat_ss", new String[] {"aaa", "ccc"},
+        "child1", sdoc("id", "2", "cat_ss", "child")
+    );
+    addDoc(adoc(doc), "nested-rtg");
```
dsmiley: I believe this is outdated -- Nested URP is already in the default chain

solr/solr-ref-guide/src/the-terms-component.adoc
```diff
@@ -301,3 +301,5 @@ See the section <<distributed-requests.adoc#configuring-the-shardhandlerfactory,
 
 `shards.qt`::
 Specifies the request handler Solr uses for requests to shards.
+
+Same as with regular distributed search, the `shards` parameter is subject to a host whitelist that has to be configured in the component init parameters using the configuration key `shardsWhitelist` and the list of hosts as values. In the same way as with distributed search, the whitelist will be populated to all live nodes by default when running in SolrCloud mode. If you need to disable this feature for backwards compatibility, you can set the system property `solr.disable.shardsWhitelist=true`.
```
dsmiley: this snuck in from a bad merge

solr/core/src/java/org/apache/solr/update/processor/DistributedUpdateProcessor.java
```diff
@@ -645,32 +660,126 @@ boolean getUpdatedDocument(AddUpdateCommand cmd, long versionOnUpdate) throws IO
...
         throw new SolrException(ErrorCode.CONFLICT, "Document not found for update.  id=" + cmd.getPrintableId());
       }
     } else {
-      oldDoc.remove(CommonParams.VERSION_FIELD);
+      completeHierarchy.remove(CommonParams.VERSION_FIELD);
     }
 
 
-    cmd.solrDoc = docMerger.merge(sdoc, oldDoc);
+    SolrInputDocument mergedDoc;
```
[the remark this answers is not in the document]
moshebla: Hmmm fair point.
I'll try and move the logic there.

solr/core/src/java/org/apache/solr/update/processor/DistributedUpdateProcessor.java
```diff
@@ -645,32 +660,126 @@ boolean getUpdatedDocument(AddUpdateCommand cmd, long versionOnUpdate) throws IO
...
+    if(idField == null || completeHierarchy == null) {
+      // create a new doc by default if an old one wasn't found
+      mergedDoc = docMerger.merge(sdoc, new SolrInputDocument());
+    } else {
+      if(isDeeplyNestedSchema && !isRootDoc(sdoc, completeHierarchy)) {
```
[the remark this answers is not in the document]
moshebla: The user does send the root, only it is done using the [`_route_`](https://issues.apache.org/jira/browse/SOLR-12638?focusedCommentId=16658534&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-16658534) parameter.

solr/solr-ref-guide/src/the-terms-component.adoc
moshebla: Oops,
I'll take care of this.

solr/core/src/test-files/solr/collection1/conf/solrconfig-block-atomic-update.xml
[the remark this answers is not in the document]
moshebla: Since this feature requires updateLog to be enabled, I used solrconfig-tlog.xml and added a few fields to schema-nest.xml to ensure signatureField URP works.

solr/core/src/java/org/apache/solr/update/processor/AtomicUpdateDocumentMerger.java
```diff
@@ -83,6 +86,17 @@ public static boolean isAtomicUpdate(final AddUpdateCommand cmd) {
...
     return false;
   }
+
+  /**
+   *
+   * @param sdoc doc to be updated
+   * @param rootDoc the root document
+   * @return whether sdoc's id equals to rootDoc's id
+   */
+  public boolean isRootDoc(SolrInputDocument sdoc, SolrInputDocument rootDoc) {
```
dsmiley: This method's signature is so weird and its implementation is short enough that I wonder if you could simply just inline it and we'd be happier?

solr/core/src/java/org/apache/solr/update/processor/DistributedUpdateProcessor.java
```diff
@@ -663,118 +661,36 @@ boolean getUpdatedDocument(AddUpdateCommand cmd, long versionOnUpdate) throws IO
...
         SolrInputDocument sdocWithChildren = RealTimeGetComponent.getInputDocument(cmd.getReq().getCore(),
             id, RealTimeGetComponent.Resolution.DOC_WITH_CHILDREN);
-        // get path of document to be updated
-        String updatedDocPath = (String) sdocWithChildren.getFieldValue(IndexSchema.NEST_PATH_FIELD_NAME);
-        // get the SolrInputField containing the document which the AddUpdateCommand updates
-        SolrInputField sifToReplace = getFieldFromHierarchy(completeHierarchy, updatedDocPath);
-        // update SolrInputField, either appending or replacing the updated document
-        updateDocInSif(sifToReplace, sdocWithChildren, sdoc);
-        mergedDoc = completeHierarchy;
+        mergedDoc = docMerger.mergeNonRoot(sdoc, oldRootDocWithChildren, sdocWithChildren);
```
dsmiley: `mergeChildDoc` would probably be clearer than `mergeNonRoot`?  It's usually clearer to avoid negatives when we can say clearly what it _is_ versus what it _is not_.

solr/core/src/java/org/apache/solr/update/processor/AtomicUpdateDocumentMerger.java
moshebla: > * Can solrconfig-block-atomic-update.xml be deleted?

Yes :+1: 

>     * I didn't notice where you are using/testing the signatureField; I only noticed your edit to add the schema portion.

I started using solrconfig-tlog.xml, which uses the signature field URP.
It is now required that the schema have the required signatureFields.

moshebla:
> * Can solrconfig-block-atomic-update.xml be deleted?

Yes :+1: 

>     * I didn't notice where you are using/testing the signatureField; I only noticed your edit to add the schema portion.

I started using solrconfig-tlog.xml, which uses the signature field URP.
It is now required that the schema have the required signatureFields.

solr/core/src/java/org/apache/solr/update/processor/DistributedUpdateProcessor.java
moshebla: Sure thing.

moshebla:
> Does RTG handle [child] finally? I'm not sure that was tested if it does (it needs a test if it works).

Yes, it is tested in the NestedAtomicUpdateTest test suite.

dsmiley:
Could the necessary parts of sharded-block-schema.xml be folded into schema-nest.xml (thus one less schema) or do you view them as too dissimilar?
Can you please propose CHANGES.txt wording?

moshebla:
> Can you please propose CHANGES.txt wording?

I will try and get to it tomorrow.

I have yet to address your other [proposal](https://github.com/apache/lucene-solr/pull/455#discussion_r270084153), I will also try to get that done.

moshebla:
> Could the necessary parts of sharded-block-schema.xml be folded into schema-nest.xml (thus one less schema) or do you view them as too dissimilar?

schema-nest.xml has a copy field which copies the doc's id to an int field `id_i`, which makes it more difficult to test the distribution between shards in the tests.

moshebla:
> Can you please propose CHANGES.txt wording?

SOLR-12638: Atomic Updates for nested documents. This enables atomic updates for nested documents, without the need to supply the whole nested hierarchy (which would be overwritten if absent). This is done by fetching the whole document hierarchy, updating the specific doc in the path that is to be updated, removing the old document hierarchy and indexing the new one with the atomic update merged into it.

dsmiley:
For detecting which document got where, consider the `[shard]` doc transformer https://lucene.apache.org/solr/guide/7_7/transforming-result-documents.html

> SOLR-12638: Atomic Updates for nested documents. This enables atomic updates for nested documents, without the need to supply the whole nested hierarchy (which would be overwritten if absent). This is done by fetching the whole document hierarchy, updating the specific doc in the path that is to be updated, removing the old document hierarchy and indexing the new one with the atomic update merged into it.

Thanks for taking a stab at this.  Sometimes I think the "atomic"-ness gets more visibility in features/naming than the "partial-update"-ness that goes along with it, which is IMO more in the minds of those that use these features.  I could be wrong.  Maybe instead of "Atomic Updates" we say "Atomic/Partial Updates"?  You could also mention that `[child]` now works in RTG.

OH; we're missing updates to the Ref Guide!  See `updating-parts-of-documents.adoc`.  You could also add a link & mention in the indexing side of the page on nested docs you did.

solr/core/src/java/org/apache/solr/schema/IndexSchema.java
```diff
@@ -1959,6 +1959,21 @@ public boolean isUsableForChildDocs() {
...
+   * defined as a {@link NestPathField}
+   * @lucene.internal
+   */
+  public boolean savesChildDocRelations() {
+    //TODO make this boolean a field so it needn't be looked up each time?
+    if (!isUsableForChildDocs()) {
+      return false;
+    }
+    FieldType nestPathType = getFieldTypeNoEx(NEST_PATH_FIELD_NAME);
+    return null != nestPathType && nestPathType.getClass().equals(NestPathField.class);
```
dsmiley: if you do "instanceof NestPathField" it allows advanced users to subclass which is nicer.

moshebla:
> Thanks for taking a stab at this. Sometimes I think the "atomic"-ness gets more visibility in features/naming than the "partial-update"-ness that goes along with it, which is IMO more in the minds of those that use these features. I could be wrong. Maybe instead of "Atomic Updates" we say "Atomic/Partial Updates"? You could also mention that `[child]` now works in RTG.

SOLR-12638: Partial/Atomic Updates for nested documents. This enables atomic updates for nested documents, without the need to supply the whole nested hierarchy (which would be overwritten if absent). This is done by fetching the whole document hierarchy, updating the specific doc in the path that is to be updated, removing the old document hierarchy and indexing the new one with the atomic update merged into it. `[child]` Doc Transformer now also work with RealTimeGet

moshebla:
> OH; we're missing updates to the Ref Guide! See `updating-parts-of-documents.adoc`. You could also add a link & mention in the indexing side of the page on nested docs you did.

Could I submit that as another patch/PR?
Sometimes I find all the changes overwhelming, especially in longer PRs such as this one.

dsmiley:
Another PR is fine.  I attached the patch version of this to JIRA and enabled patch review.  I'm looking forward to seeing this one done!

PR #457, opened 2018-09-20
SOLR-12791: Add Metrics reporting for AuthenticationPlugin

janhoy:
First partial patch

### DIFF
--- a/solr/core/src/java/org/apache/solr/core/CoreContainer.java
+++ b/solr/core/src/java/org/apache/solr/core/CoreContainer.java
@@ -563,6 +563,10 @@ public class CoreContainer {
 
     securityConfHandler = isZooKeeperAware() ? new SecurityConfHandlerZk(this) : new SecurityConfHandlerLocal(this);
     reloadSecurityProperties();
+    if (authenticationPlugin != null) {
+      authenticationPlugin.plugin.initializeMetrics(metricManager, SolrInfoBean.Group.node.toString(), metricTag, "/authentication/" + authenticationPlugin.plugin.getName());
+      pkiAuthenticationPlugin.initializeMetrics(metricManager, SolrInfoBean.Group.node.toString(), metricTag, "/authentication/" + pkiAuthenticationPlugin.getName());
+    }
     this.backupRepoFactory = new BackupRepositoryFactory(cfg.getBackupRepositoryPlugins());
 
     createHandler(ZK_PATH, ZookeeperInfoHandler.class.getName(), ZookeeperInfoHandler.class);
--- a/solr/core/src/java/org/apache/solr/core/SolrInfoBean.java
+++ b/solr/core/src/java/org/apache/solr/core/SolrInfoBean.java
@@ -34,7 +34,7 @@ public interface SolrInfoBean {
    * Category of Solr component.
    */
   enum Category { CONTAINER, ADMIN, CORE, QUERY, UPDATE, CACHE, HIGHLIGHTER, QUERYPARSER, SPELLCHECKER,
-    SEARCHER, REPLICATION, TLOG, INDEX, DIRECTORY, HTTP, OTHER }
+    SEARCHER, REPLICATION, TLOG, INDEX, DIRECTORY, HTTP, SECURITY, OTHER }
 
   /**
    * Top-level group of beans or metrics for a subsystem.
--- a/solr/core/src/java/org/apache/solr/security/AuthenticationPlugin.java
+++ b/solr/core/src/java/org/apache/solr/security/AuthenticationPlugin.java
@@ -20,15 +20,43 @@ import javax.servlet.FilterChain;
 import javax.servlet.ServletRequest;
 import javax.servlet.ServletResponse;
 import java.io.Closeable;
+import java.util.Arrays;
 import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+
+import com.codahale.metrics.Counter;
+import com.codahale.metrics.Meter;
+import com.codahale.metrics.MetricRegistry;
+import com.codahale.metrics.Timer;
+import org.apache.solr.core.SolrInfoBean;
+import org.apache.solr.metrics.SolrMetricManager;
+import org.apache.solr.metrics.SolrMetricProducer;
 
 /**
  * 
  * @lucene.experimental
  */
-public abstract class AuthenticationPlugin implements Closeable {
+public abstract class AuthenticationPlugin implements Closeable, SolrInfoBean, SolrMetricProducer {
 
   final public static String AUTHENTICATION_PLUGIN_PROP = "authenticationPlugin";
+
+  // Metrics
+  private Set<String> metricNames = ConcurrentHashMap.newKeySet();
+  private MetricRegistry registry;
+
+  protected String registryName;
+  protected SolrMetricManager metricManager;
+  protected Meter numErrors = new Meter();
+  protected Meter numTimeouts = new Meter();
+  protected Counter requests = new Counter();
+  protected Timer requestTimes = new Timer();
+  protected Counter totalTime = new Counter();
+  protected Counter numAuthenticated = new Counter();
+  protected Counter numPassThrough = new Counter();
+  protected Counter numWrongCredentials = new Counter();
+  protected Counter numMissingCredentials = new Counter();
+  protected Counter numInvalidCredentials = new Counter();
 
   /**
    * This is called upon loading up of a plugin, used for setting it up.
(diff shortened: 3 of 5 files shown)


solr/core/src/java/org/apache/solr/core/CoreContainer.java
```diff
@@ -798,6 +798,12 @@ private void reloadSecurityProperties() {
     SecurityConfHandler.SecurityConfig securityConfig = securityConfHandler.getSecurityConfig(false);
     initializeAuthorizationPlugin((Map<String, Object>) securityConfig.getData().get("authorization"));
     initializeAuthenticationPlugin((Map<String, Object>) securityConfig.getData().get("authentication"));
+    if (authenticationPlugin != null && authenticationPlugin.plugin.getMetricRegistry() == null) {
```
sigram: Is this second check necessary? we know that just after the plugin was created its metricRegistry is null, it's set only after `initializeMetrics` has been called.

solr/test-framework/src/java/org/apache/solr/cloud/SolrCloudTestCase.java
```diff
@@ -420,4 +425,68 @@ public static void ensureRunningJettys(int nodeCount, int timeoutSeconds) throws
     cluster.waitForAllNodes(timeoutSeconds);
   }
 
+
+  static final List<String> AUTH_METRICS_KEYS = Arrays.asList("errors", "timeouts", "requests", "authenticated",
```
sigram: I think it would be better to put this whole section in a subclass of SolrCloudTestCase, eg. SolrCloudAuthTestCase - there are only two tests that need this stuff but there are hundreds of tests that extend this base class and don't need this.

solr/core/src/java/org/apache/solr/core/CoreContainer.java
janhoy: Yea, I think you're right.

solr/test-framework/src/java/org/apache/solr/cloud/SolrCloudTestCase.java
janhoy: Yep, a separate test base class for auth tests makes sense. Good idea. We could then push down some other utility methods from the tests as well, I see some duplication a few places.

solr/core/src/java/org/apache/solr/core/SolrInfoBean.java
```diff
@@ -34,7 +34,7 @@
    * Category of Solr component.
    */
   enum Category { CONTAINER, ADMIN, CORE, QUERY, UPDATE, CACHE, HIGHLIGHTER, QUERYPARSER, SPELLCHECKER,
-    SEARCHER, REPLICATION, TLOG, INDEX, DIRECTORY, HTTP, OTHER }
+    SEARCHER, REPLICATION, TLOG, INDEX, DIRECTORY, HTTP, SECURITY, OTHER }
```
janhoy: I chose a new category because over time I'd like to add metrics also for Authorization plugins and Auditlog plugins (all components registered in security.json). An alternative could have been `CONTAINER` I guess?

solr/core/src/java/org/apache/solr/security/AuthenticationPlugin.java
```diff
@@ -20,16 +20,44 @@
...
+import org.apache.solr.core.SolrInfoBean;
+import org.apache.solr.metrics.SolrMetricManager;
+import org.apache.solr.metrics.SolrMetricProducer;
 
 /**
  * 
  * @lucene.experimental
  */
-public abstract class AuthenticationPlugin implements Closeable {
+public abstract class AuthenticationPlugin implements Closeable, SolrInfoBean, SolrMetricProducer {
```
janhoy: I chose to add the `SolrMetricsProducer` on the base class instead of letting each Auth plugin choose to implement it or not. The effect will be that plugins that do not change their code to increment counters will be logged as 0 for these counters. But the `requests` count, `requestTimes` stats, `totalTime` and `errors` will still be logged by the base class on behalf of all plugins even if they don't increment the other counters.

solr/core/src/java/org/apache/solr/security/AuthenticationPlugin.java
```diff
@@ -52,11 +80,73 @@
...
+   * It does some standard metrics counting.
+   */
+  public final boolean authenticate(ServletRequest request, ServletResponse response, FilterChain filterChain) throws Exception {
+    Timer.Context timer = requestTimes.time();
+    requests.inc();
+    try {
+      return doAuthenticate(request, response, filterChain);
+    } catch(Exception e) {
+      numErrors.mark();
+      throw e;
```
janhoy: Is this re-throw safe?

solr/core/src/java/org/apache/solr/security/AuthenticationPlugin.java
```diff
@@ -52,11 +80,73 @@
...
   }
 
+  @Override
+  public void initializeMetrics(SolrMetricManager manager, String registryName, String tag, final String scope) {
+    this.metricManager = manager;
+    this.registryName = registryName;
+    // Metrics
+    registry = manager.registry(registryName);
+    numErrors = manager.meter(this, registryName, "errors", getCategory().toString(), scope);
+    numTimeouts = manager.meter(this, registryName, "timeouts", getCategory().toString(), scope);
```
janhoy: No plugins currently log timeouts. I think it should perhaps be removed and instead re-introduced if we see that some auth plugins with external network depenedencies feel the need for logging number of (external) timeouts.

solr/core/src/java/org/apache/solr/security/AuthenticationPlugin.java
```diff
@@ -52,11 +80,73 @@
...
+    requests = manager.counter(this, registryName, "requests", getCategory().toString(), scope);
+    numAuthenticated = manager.counter(this, registryName, "authenticated", getCategory().toString(), scope);
+    numPassThrough = manager.counter(this, registryName, "passThrough", getCategory().toString(), scope);
+    numWrongCredentials = manager.counter(this, registryName, "failWrongCredentials", getCategory().toString(), scope);
```
janhoy: If auth does not succeed we log either wrong credentials, invalid credentials (e.g. malformed, if basicAuth header does not contain a ':'), missing credentials (if no header at all). Not all plugins will be able to distinguish between all these, such as hadoop plugin which I raised a question about in the JIRA. Should we have a more coarse-grained metric on this? Perhaps these are enough: `numAuthenticated`, `numPassThrough`, `numFailedAuth` and `numErrors`?

solr/core/src/java/org/apache/solr/security/AuthenticationPlugin.java
```diff
@@ -52,11 +80,73 @@
...
+  }
+  
+  @Override
+  public String getName() {
+    return this.getClass().getName();
+  }
+
+  @Override
+  public String getDescription() {
+    return this.getClass().getName();
```
janhoy: I did not want to extend the API surface of the Auth plugins to require a description, Any suggestions of how to obtain better description from somewhere else?

solr/core/src/java/org/apache/solr/core/SolrInfoBean.java
sigram: SECURITY makes more sense imho.

solr/core/src/java/org/apache/solr/security/AuthenticationPlugin.java
sigram: +1 to remove and add it later as needed.

solr/core/src/java/org/apache/solr/security/AuthenticationPlugin.java
sigram: I think that malformed credentials should be simply reported as errors.

solr/core/src/java/org/apache/solr/security/AuthenticationPlugin.java
sigram: Perhaps it should describe at least what the base class' function is, in this case `"Authentication Plugin " + this.getClass().getSimpleName()`. This way subclasses would automatically report the base function and their specific class name (or they could override it if needed).

janhoy: Done

solr/core/src/java/org/apache/solr/security/AuthenticationPlugin.java
janhoy: Done

PR #458, opened 2018-09-24
SOLR-12799: Allow Authentication Plugins to intercept internode requests on a per-request basis

janhoy:
Framework support for per-request decision on whether to secure inter-node requests


solr/core/src/java/org/apache/solr/security/BasicAuthPlugin.java
```diff
@@ -172,6 +192,22 @@ public void closeRequest() {
...
   }
 
+  @Override
+  protected boolean interceptInternodeRequest(HttpRequest httpRequest, HttpContext httpContext) {
+    if (forwardCredentials) {
+      if (httpContext instanceof HttpClientContext) {
+        HttpClientContext httpClientContext = (HttpClientContext) httpContext;
+        if (httpClientContext.getUserToken() instanceof BasicAuthUserPrincipal) {
+          BasicAuthUserPrincipal principal = (BasicAuthUserPrincipal) httpClientContext.getUserToken();
+          String userPassBase64 = Base64.encodeBase64String((principal.getName() + ":" + principal.getPassword()).getBytes(StandardCharsets.UTF_8));
```
janhoy: UTF-8 is hardcoded, as it is also hardcoded on the backend side. Could alternatively have sent a `charset=UTF-8` on the `WWW-authenticate` header from the server to be explicit, but see no real need for that in the context of this issue.

solr/core/src/java/org/apache/solr/security/BasicAuthPlugin.java
```diff
@@ -181,7 +217,54 @@ public boolean getBlockUnknown(){
...
+    @Override
+    public int hashCode() {
+      return Objects.hash(username, password);
+    }
+
+    @Override
+    public String toString() {
+      return new ToStringBuilder(this)
+          .append("username", username)
+          .append("pwd", password)
```
janhoy: Hmm, perhaps the password should not be part of the `toString`, in case this object is printed in some error logs...

solr/core/src/test/org/apache/solr/security/BasicAuthIntegrationTest.java
```diff
@@ -232,6 +229,17 @@ public void testBasicAuth() throws Exception {
...
+      
+      addDocument("harry","HarryIsUberCool","id", "5");
+
+      // Validate forwardCredentials
+      assertEquals(1, executeQuery(params("q", "id:5"), "harry", "HarryIsUberCool").getResults().getNumFound());
+      // TODO: Validate that this requests were handled by PKI plugin
+      executeCommand(baseUrl + authcPrefix, cl, "{set-property : { forwardCredentials: true}}", "harry", "HarryIsUberCool");
+      verifySecurityStatus(cl, baseUrl + authcPrefix, "authentication/forwardCredentials", "true", 20, "harry", "HarryIsUberCool");
+      assertEquals(1, executeQuery(params("q", "id:5"), "harry", "HarryIsUberCool").getResults().getNumFound());
+      // NOCOMMIT: Validate that sub requests were handled by BasicAuth plugin
```
janhoy: After committing SOLR-12791 we can test this by asserting that metrics has been incremented for Basic and not for PKI

solr/core/src/java/org/apache/solr/security/BasicAuthPlugin.java
janhoy: Redacted password with `*****`

solr/core/src/test/org/apache/solr/security/BasicAuthIntegrationTest.java
janhoy: Done

janhoy:
@bbl232 If you're using BasicAuthPlugin then, yes, this feature may help you. Feel free to build 8.0.0-SNAPSHOT and test. Backporting should be easy.

bbl232:
@janhoy yeah we're planning on using the BasicAuthPlugin... this seems like a fairly large flaw unless we're just misconfiguring solr cloud somehow? IDK if there is a way to make the node handling the request evaluate the authorization rules without the collection being present on the node? I'll play around with pulling and building 8.0.0-SNAPSHOT today, though we're planning on using this in production so it would be awesome to get a point release / hotfix for 7.6.0... especially because we're pulling and using the official solr-docker image to run the cluster

PR #463, opened 2018-10-04
SOLR-12833: Use timed-out lock in DistributedUpdateProcessor

jefferyyuan:


tflobbe:
This code was already committed b9a966e

PR #464, opened 2018-10-07
SOLR-12555: refactor tests in package org.apache.solr.search

barrotsteindev:
Tests are refactored to use expectThrows instead of try/catch.

### DIFF
--- a/solr/core/src/test/org/apache/solr/search/QueryEqualityTest.java
+++ b/solr/core/src/test/org/apache/solr/search/QueryEqualityTest.java
@@ -1190,14 +1190,8 @@ public class QueryEqualityTest extends SolrTestCaseJ4 {
     assertFuncEquals("gte(foo_i,2)", "gte(foo_i,2)");
     assertFuncEquals("eq(foo_i,2)", "eq(foo_i,2)");
 
-    boolean equals = false;
-    try {
-      assertFuncEquals("eq(foo_i,2)", "lt(foo_i,2)");
-      equals = true;
-    } catch (AssertionError e) {
-      //expected
-    }
-    assertFalse(equals);
+    expectThrows(AssertionError.class, "expected error, functions are not equal",
+        () -> assertFuncEquals("eq(foo_i,2)", "lt(foo_i,2)"));
   }
 
   public void testChildField() throws Exception {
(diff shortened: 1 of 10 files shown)


solr/core/src/test/org/apache/solr/search/QueryEqualityTest.java
```diff
@@ -1190,14 +1190,8 @@ public void testCompares() throws Exception {
...
 
-    boolean equals = false;
-    try {
-      assertFuncEquals("eq(foo_i,2)", "lt(foo_i,2)");
-      equals = true;
-    } catch (AssertionError e) {
-      //expected
-    }
-    assertFalse(equals);
+    expectThrows(AssertionError.class, "expected error, functions are not equal",
```
gerlowskija: [0] Not suggesting you change it here, but....kindof weird that there's just not an `assertFuncNotEquals`

solr/core/src/test/org/apache/solr/search/QueryEqualityTest.java
```diff
@@ -1214,29 +1208,23 @@ public void testPayloadScoreQuery() throws Exception {
...
-      assertQueryEquals
-          ("payload_check"
-              , "{!payload_check f=foo_dpf payloads=2}one"
-              , "{!payload_check f=foo_dpf payloads=2}two"
-          );
-      fail("queries should not have been equal");
-    } catch(AssertionFailedError e) {
-      assertTrue("queries were not equal, as expected", true);
-    }
+    expectThrows(AssertionFailedError.class, "queries were not equal, as expected",
```
gerlowskija: [-1] I think this exception message here is backwards.  This assertion fails if the queries _were_ equal, but the message indicates that the problem is that they were !=.  Using the message from the original `fail()` invocation would probably work better here.

solr/core/src/test/org/apache/solr/search/QueryEqualityTest.java
```diff
@@ -1214,29 +1208,23 @@ public void testPayloadScoreQuery() throws Exception {
...
-      assertQueryEquals
-          ("payload_score"
-              , "{!payload_score f=foo_dpf v=query func=min includeSpanScore=false}"
-              , "{!payload_score f=foo_dpf v=query func=min includeSpanScore=true}"
-          );
-      fail("queries should not have been equal");
-    } catch(AssertionFailedError e) {
-      assertTrue("queries were not equal, as expected", true);
-    }
+    expectThrows(AssertionFailedError.class, "queries were not equal, as expected",
```
gerlowskija: [-1] I think this exception message here is backwards.  This assertion fails if the queries _were_ equal, but the message implies that the problem is that they were !=.  Using the message from the original `fail()` invocation would probably work better here ("queries should not have been equal")

solr/core/src/test/org/apache/solr/search/QueryEqualityTest.java
```diff
@@ -1260,16 +1248,14 @@ public void testBoolQuery() throws Exception {
...
-          ("bool"
-              , "{!bool must='{!lucene}foo_s:a'}"
-              , "{!bool should='{!lucene}foo_s:a'}"
-          );
-      fail("queries should not have been equal");
-    } catch(AssertionFailedError e) {
-      assertTrue("queries were not equal, as expected", true);
-    }
+
+    expectThrows(AssertionFailedError.class, "queries were not equal, as expected",
```
gerlowskija: [-1] ditto re: wrong String message in `expectThrows` here

solr/core/src/test/org/apache/solr/search/TestExtendedDismaxParser.java
```diff
@@ -656,45 +656,38 @@ public void testAliasingBoost() throws Exception {
...
-      } catch (SolrException e) {
-        assertTrue(e.getCause().getMessage().contains("Field aliases lead to a cycle"));
-      }
-      
-      try {
-        h.query(req("defType","edismax", "q","who:(Zapp Pig)", "qf","text", "f.who.qf","name","f.name.qf","myalias", "f.myalias.qf","who"));
-        fail("Cyclic alising not detected");
-      } catch (SolrException e) {
-        assertTrue(e.getCause().getMessage().contains("Field aliases lead to a cycle"));
+      } catch (SolrException ex) {
```
gerlowskija: [Q] Is there a reason that this example also couldn't be changed into an `expectThrows`?

solr/core/src/test/org/apache/solr/search/TestRealTimeGet.java
```diff
@@ -414,56 +391,44 @@ public void testOptimisticLocking() throws Exception {
...
-    } catch (SolrException se) {
-      assertEquals(409, se.code());
-    }
+    // deleting the previous version should now fail
+    se = expectThrows(SolrException.class, "deleting the previous version should now fail",
+        () -> deleteAndGetVersion("1", params("_version_", Long.toString(version))));
+    assertEquals(409, se.code());
 
-    version = version2;
+    final long prevVersion = version2;
```
gerlowskija: [0] I'm still somewhat leery of changing how the version variables are used here.  I agree with what seems like your intent here - that `final` variables often make it much easier to reason about Java code.  But with how flaky the tests are, I'd rather not introduce such changes here.

barrotsteindev: The only problem is that if the variables are not final, they can not be used inside the lambda that is passed to expectThrows.
Should I just revert these changes?

gerlowskija: Oof, yeah good point, that would be a problem.  No need to revert. I'll just need to pay particular attention to this class when testing things out.

solr/core/src/test/org/apache/solr/search/TestExtendedDismaxParser.java
barrotsteindev: My bad,
I sort of missed that one.
Just pushed a new commit fixing it though.

gerlowskija:
Thanks again for the work Bar.  I've merged this to `master`.  Should see this PR get closed out shortly.

PR #465, opened 2018-10-09
SOLR-7896: Add a login page to Admin UI, with initial support for Basic Auth

janhoy:
First cut for basic auth


solr/webapp/web/js/angular/app.js
```diff
@@ -365,12 +370,62 @@ solrAdminApp.config([
...
+      var headers = rejection.headers();
+      var wwwAuthHeader = headers['www-authenticate'];
+      sessionStorage.setItem("auth.wwwAuthHeader", wwwAuthHeader);
+      sessionStorage.setItem("auth.statusText", rejection.statusText);
+      var authDataHeader = headers['X-Solr-AuthData'];
+      if (authDataHeader !== null) {
+        sessionStorage.setItem("auth.config", authDataHeader);
+      }
+      if ($location.path() === '/login') {
+        sessionStorage.setItem("auth.location", "/");
```
janhoy: Normally users will be redirected to Login page from whatever page that tried to do a restricted API call to Solr, and will be redirected back to that page after login. However, if user explicitly navigates to Login tab, this makes sure he'll be redirected to Dashboard instead of remaining on the Login page.

solr/webapp/web/partials/login.html
```diff
@@ -0,0 +1,80 @@
...
+      <br/>
+      <form name="logoutForm" ng-submit="logout()" role="form" ng-show="isLoggedIn()">
+        <div class="form-actions">
+          <button type="submit" class="btn btn-danger">Logout</button>
+        </div>
+      </form>
+    </div>
+
+  </div>
+
```
janhoy: The idea is that future authentication schemes can add their UI here and code hooks to `login.js` and thus have a very simple extension mechanism. While the ideal would be a true pluggable experience for 3rd party auth to also work, I have not attempted that in this iteration. Ideas of how to achieve that are welcome!

PR #471, opened 2018-10-16
SOLR-8335 HdfsLockFactory should eventually release index after crash.

manokovacs:


solr/test-framework/src/java/org/apache/solr/SolrIgnoredThreadsFilter.java
```diff
@@ -60,6 +61,11 @@ public boolean reject(Thread t) {
       return true;
     }
 
+    if (threadName.startsWith(HdfsLockFactory.HDFS_LOCK_UPDATE_PREFIX)) {
```
risdenk: This doesn't seem right. We should be able to shutdown our threads correctly...

solr/core/src/test/org/apache/solr/update/TestInPlaceUpdatesStandalone.java
```diff
@@ -18,6 +18,7 @@
```
risdenk: Unnecessary changes in this file?

solr/core/src/test/org/apache/solr/store/hdfs/HdfsLockFactoryTest.java
```diff
@@ -17,46 +17,173 @@
...
+  private HdfsDirectory dir;
+  private Path lockPath;
+  private HdfsLockFactory.HdfsLock lock = null;
+  private HdfsLockFactory.HdfsLock lock2 = null;
+  private Exception latestException = null;
+  private List<HdfsLockFactory.HdfsLock> allLocks = new LinkedList<>();
+  private static MiniDFSCluster.DataNodeProperties dataNodeProperties;
+
+  @Rule
+  public TestRuleRestoreSystemProperties p = new TestRuleRestoreSystemProperties(LOCK_HOLD_TIMEOUT_KEY);
```
risdenk: Is this necessary?

PR #473, opened 2018-10-17
SOLR-12878 - Use the cached SolrIndexSearch.fieldInfos instead of recreating one each time FacetFieldProcessorByHashDV is constructed

tpunder:
Also this updates one of the collect methods to use the `advanceExact` method to match the other collect methods.


tflobbe:
I understand all this code was merged already, right?

tpunder:
Yes this has been merged.

PR #478, opened 2018-10-18
Query Source Tracker custom component

tirthmehta1994:


tirthmehta1994:
Hi @vthacker , I have created the issue here: https://issues.apache.org/jira/browse/SOLR-12903

tirthmehta1994:
@vthacker it would be great if you could let me know of some updates here. Thanks.

PR #486, opened 2018-10-30
SOLR-12801: Fix the tests, remove BadApples and AwaitsFix annotations…

markrmiller:
…, improve env for test development.

SOLR-12804: Remove static modifier from Overseer queue access.

SOLR-12896: Introduce more checks for shutdown and closed to improve clean close and shutdown. (Partial)

SOLR-12897: Introduce AlreadyClosedException to clean up silly close / shutdown logging. (Partial)

SOLR-12898: Replace cluster state polling with ZkStateReader#waitFor. (Partial)

SOLR-12923: The new AutoScaling tests are way to flaky and need special attention. (Partial)

SOLR-12932: ant test (without badapples=false) should pass easily for developers. (Partial)


solr/core/src/test-files/solr/solr.xml
```diff
@@ -40,12 +40,12 @@
...
     <int name="hostPort">${hostPort:8983}</int>
     <str name="hostContext">${hostContext:solr}</str>
-    <int name="zkClientTimeout">${solr.zkclienttimeout:30000}</int>
+    <int name="zkClientTimeout">${solr.zkclienttimeout:45000}</int> <!-- This should be high by default - dc's are expensive -->
     <bool name="genericCoreNodeNames">${genericCoreNodeNames:true}</bool>
-    <int name="leaderVoteWait">${leaderVoteWait:10000}</int>
-    <int name="leaderConflictResolveWait">${leaderConflictResolveWait:180000}</int>
-    <int name="distribUpdateConnTimeout">${distribUpdateConnTimeout:45000}</int>
-    <int name="distribUpdateSoTimeout">${distribUpdateSoTimeout:340000}</int>
+    <int name="leaderVoteWait">${leaderVoteWait:15000}</int>   <!-- We are running tests - the default should be low, not like production -->
```
vthacker: The older explicitly set value for  leaderVoteWait was 10000 and now it's increased to 15000 but the comment seems to indicate that we wanted to lower it.

Just wanted to make sure it isn't a typo

solr/test-framework/src/java/org/apache/solr/cloud/MiniSolrCloudCluster.java
```diff
@@ -265,10 +281,30 @@ public MiniSolrCloudCluster(int numServers, Path baseDir, String solrXml, JettyC
     solrClient = buildSolrClient();
   }
 
-  private void waitForAllNodes(int numServers, int timeout) throws IOException, InterruptedException {
+  private void waitForAllNodes(int numServers, int timeoutSeconds) throws IOException, InterruptedException {
```
vthacker: Some Javadocs that we could add here :

"This method wait till all Solr JVMs ( Jettys ) are running . It waits upto the timeout (in seconds) for the JVMs to be up before throwing IllegalStateException"

solr/core/src/test-files/solr/solr.xml
markrmiller: No, didn't necessarily want to lower it, just explain why it's set so low compared to what would be recommended in production (minutes).

solr/test-framework/src/java/org/apache/solr/cloud/MiniSolrCloudCluster.java
markrmiller: Cool, I'll add, as well as a note about how you don't need to use this after starting the cluster with the builder - just when adding new jetties.

solr/contrib/analytics/src/test/org/apache/solr/analytics/legacy/facet/LegacyFieldFacetExtrasCloudTest.java
```diff
@@ -42,9 +42,8 @@
   static ArrayList<ArrayList<Integer>> intDoubleTestStart; 
   static ArrayList<ArrayList<Integer>> intStringTestStart; 
   
-  @BeforeClass
-  public static void beforeClass() throws Exception {
-    cleanIndex();
+  @Before
+  public void beforeClass() throws Exception {
```
jefferyyuan: Maybe change the method name? should not be beforeClass : )
- also in other cases :)

PR #487, opened 2018-11-02
LUCENE-8557: LeafReader.getFieldInfos should always return the same instance

tpunder:


tpunder:
This has been merged.

PR #491, opened 2018-11-05
Update for multiple term's suggestion scoring

UtsavVanodiya7:
While calculating coefficient, used all suggestion's query term position. Tried to make better scoring for suggestion.


atris:
Hi @UtsavVanodiya7 ,

Thanks for raising the PR. I took a brief look and it is a bit hard to grok the exact improvements that you are suggesting.

Could you file a JIRA and share your proposal in the description of the same? Also, please highlight the limitations in the status quo that you would like to fix, the use cases and demonstrable benefits.

PR #493, opened 2018-11-06
SOLR-12964: Make use of DocValuesIterator.advanceExact() instead of the advance()/docID() pattern

tpunder:


tpunder:
This has been merged.

PR #495, opened 2018-11-08
LUCENE-8464: Implement ConstantScoreScorer#setMinCompetitiveScore

cbismuth:
> We should make it so the iterator returns `NO_MORE_DOCS` after `setMinCompetitiveScore` is called with a value that is greater than the constant score.

See [LUCENE-8464](https://issues.apache.org/jira/browse/LUCENE-8464).


cbismuth:
@jpountz PR is ready for review, thank you.

lucene/core/src/java/org/apache/lucene/search/ConstantScoreScorer.java
```diff
@@ -58,6 +58,14 @@ public float getMaxScore(int upTo) throws IOException {
     return score;
   }
 
+  @Override
+  public void setMinCompetitiveScore(float minScore) throws IOException {
+    float minScoreDown = Math.nextDown(minScore);
```
jimczi: Since we tie-break on doc id and collect in doc id order I don't think you need to take the previous float here.
The `TopScoreCollector` sets the minimum score as the next float after the score of the worst top document so you should not need to change the value here.

lucene/core/src/test/org/apache/lucene/search/TestConstantScoreScorer.java
```diff
@@ -0,0 +1,104 @@
...
+  static class TestConstantScoreScorerIndex implements AutoCloseable {
+
+    private final Directory directory;
+    private final RandomIndexWriter writer;
+    private final IndexReader reader;
+
+    TestConstantScoreScorerIndex() throws IOException {
+      directory = newDirectory();
+
+      writer = new RandomIndexWriter(random(), directory);
```
jimczi: you need to use a merger that retains doc id order. Something like:
```
RandomIndexWriter w = new RandomIndexWriter(random(), dir, newIndexWriterConfig().setMergePolicy(
        // retain doc id order
        newLogMergePolicy(random().nextBoolean())));
```

lucene/core/src/test/org/apache/lucene/search/TestConstantScoreScorer.java
```diff
@@ -0,0 +1,104 @@
...
+      for (String VALUE : VALUES) {
+        Document doc = new Document();
+        doc.add(newTextField(FIELD, VALUE, Field.Store.YES));
+        writer.addDocument(doc);
+      }
+      reader = writer.getReader();
```
jimczi: The random index writer can commit segments randomly so you need to force merge to 1 segments to ensure that all documents are in a single segment:
```
writer.forceMerge(1);
```

lucene/core/src/test/org/apache/lucene/search/TestConstantScoreScorer.java
cbismuth: Done in 0ad2f781790f35c6b7f16531db182b743d13a7c0, thank you.

lucene/core/src/test/org/apache/lucene/search/TestConstantScoreScorer.java
cbismuth: Done in 8d7d5aefd477dc86a35cc3eb137a8a45534db285, thank you.

Could you please explain me why docs have to be in a single segment to ensure this test passes?

lucene/core/src/java/org/apache/lucene/search/ConstantScoreScorer.java
cbismuth: Without this, the `TestPrefixRandom#testPrefixes` test fails with `score = 1.0` and `minScore = 1.0000001`, so documents are not collected. What do you think of it?

```
java.lang.AssertionError: Collected: 21 but scorer: -1

	at __randomizedtesting.SeedInfo.seed([C3B30196BD4F380C:91676D9F44FA34A5]:0)
	at org.apache.lucene.search.AssertingLeafCollector.collect(AssertingLeafCollector.java:48)
	at org.apache.lucene.search.Weight$DefaultBulkScorer.scoreAll(Weight.java:263)
	at org.apache.lucene.search.Weight$DefaultBulkScorer.score(Weight.java:214)
	at org.apache.lucene.search.AssertingBulkScorer.score(AssertingBulkScorer.java:81)
	at org.apache.lucene.search.AssertingBulkScorer.score(AssertingBulkScorer.java:65)
	at org.apache.lucene.search.IndexSearcher.search(IndexSearcher.java:652)
	at org.apache.lucene.search.AssertingIndexSearcher.search(AssertingIndexSearcher.java:72)
	at org.apache.lucene.search.IndexSearcher.search(IndexSearcher.java:443)
	at org.apache.lucene.search.IndexSearcher.search(IndexSearcher.java:567)
	at org.apache.lucene.search.IndexSearcher.searchAfter(IndexSearcher.java:419)
	at org.apache.lucene.search.IndexSearcher.search(IndexSearcher.java:430)
	at org.apache.lucene.search.TestPrefixRandom.assertSame(TestPrefixRandom.java:135)
	at org.apache.lucene.search.TestPrefixRandom.testPrefixes(TestPrefixRandom.java:125)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at com.carrotsearch.randomizedtesting.RandomizedRunner.invoke(RandomizedRunner.java:1742)
	at com.carrotsearch.randomizedtesting.RandomizedRunner$8.evaluate(RandomizedRunner.java:935)
	at com.carrotsearch.randomizedtesting.RandomizedRunner$9.evaluate(RandomizedRunner.java:971)
	at com.carrotsearch.randomizedtesting.RandomizedRunner$10.evaluate(RandomizedRunner.java:985)
	at org.apache.lucene.util.TestRuleSetupTeardownChained$1.evaluate(TestRuleSetupTeardownChained.java:49)
	at org.apache.lucene.util.AbstractBeforeAfterRule$1.evaluate(AbstractBeforeAfterRule.java:45)
	at org.apache.lucene.util.TestRuleThreadAndTestName$1.evaluate(TestRuleThreadAndTestName.java:48)
	at org.apache.lucene.util.TestRuleIgnoreAfterMaxFailures$1.evaluate(TestRuleIgnoreAfterMaxFailures.java:64)
	at org.apache.lucene.util.TestRuleMarkFailure$1.evaluate(TestRuleMarkFailure.java:47)
	at com.carrotsearch.randomizedtesting.rules.StatementAdapter.evaluate(StatementAdapter.java:36)
	at com.carrotsearch.randomizedtesting.ThreadLeakControl$StatementRunner.run(ThreadLeakControl.java:368)
	at com.carrotsearch.randomizedtesting.ThreadLeakControl.forkTimeoutingTask(ThreadLeakControl.java:817)
	at com.carrotsearch.randomizedtesting.ThreadLeakControl$3.evaluate(ThreadLeakControl.java:468)
	at com.carrotsearch.randomizedtesting.RandomizedRunner.runSingleTest(RandomizedRunner.java:944)
	at com.carrotsearch.randomizedtesting.RandomizedRunner$5.evaluate(RandomizedRunner.java:830)
	at com.carrotsearch.randomizedtesting.RandomizedRunner$6.evaluate(RandomizedRunner.java:880)
	at com.carrotsearch.randomizedtesting.RandomizedRunner$7.evaluate(RandomizedRunner.java:891)
	at org.apache.lucene.util.AbstractBeforeAfterRule$1.evaluate(AbstractBeforeAfterRule.java:45)
	at com.carrotsearch.randomizedtesting.rules.StatementAdapter.evaluate(StatementAdapter.java:36)
	at org.apache.lucene.util.TestRuleStoreClassName$1.evaluate(TestRuleStoreClassName.java:41)
	at com.carrotsearch.randomizedtesting.rules.NoShadowingOrOverridesOnMethodsRule$1.evaluate(NoShadowingOrOverridesOnMethodsRule.java:40)
	at com.carrotsearch.randomizedtesting.rules.NoShadowingOrOverridesOnMethodsRule$1.evaluate(NoShadowingOrOverridesOnMethodsRule.java:40)
	at com.carrotsearch.randomizedtesting.rules.StatementAdapter.evaluate(StatementAdapter.java:36)
	at com.carrotsearch.randomizedtesting.rules.StatementAdapter.evaluate(StatementAdapter.java:36)
	at com.carrotsearch.randomizedtesting.rules.StatementAdapter.evaluate(StatementAdapter.java:36)
	at org.apache.lucene.util.TestRuleAssertionsRequired$1.evaluate(TestRuleAssertionsRequired.java:53)
	at org.apache.lucene.util.TestRuleMarkFailure$1.evaluate(TestRuleMarkFailure.java:47)
	at org.apache.lucene.util.TestRuleIgnoreAfterMaxFailures$1.evaluate(TestRuleIgnoreAfterMaxFailures.java:64)
	at org.apache.lucene.util.TestRuleIgnoreTestSuites$1.evaluate(TestRuleIgnoreTestSuites.java:54)
	at com.carrotsearch.randomizedtesting.rules.StatementAdapter.evaluate(StatementAdapter.java:36)
	at com.carrotsearch.randomizedtesting.ThreadLeakControl$StatementRunner.run(ThreadLeakControl.java:368)
	at java.lang.Thread.run(Thread.java:748)

NOTE: reproduce with: ant test  -Dtestcase=TestPrefixRandom -Dtests.method=testPrefixes -Dtests.seed=C3B30196BD4F380C -Dtests.slow=true -Dtests.badapples=true -Dtests.locale=lv-LV -Dtests.timezone=Asia/Oral -Dtests.asserts=true -Dtests.file.encoding=UTF-8
NOTE: test params are: codec=Asserting(Lucene80), sim=Asserting(org.apache.lucene.search.similarities.AssertingSimilarity@3da71b7e), locale=lv-LV, timezone=Asia/Oral
NOTE: Linux 4.15.0-38-generic amd64/Oracle Corporation 1.8.0_181 (64-bit)/cpus=4,threads=1,free=114149608,total=124780544
NOTE: All tests run in this JVM: [TestPrefixRandom]
Disconnected from the target VM, address: '127.0.0.1:40741', transport: 'socket'

Process finished with exit code 255
```

cbismuth:
Thanks a lot @jimczi for the very quick and very helpful review, I've made changes to improve this PR.

lucene/core/src/test/org/apache/lucene/search/TestConstantScoreScorer.java
jimczi: Because you check only one segment in the test:
https://github.com/apache/lucene-solr/blob/8d7d5aefd477dc86a35cc3eb137a8a45534db285/lucene/core/src/test/org/apache/lucene/search/TestConstantScoreScorer.java#L94
... which is fine if you force merge to 1 segment.

cbismuth: Oh yes, I understand, thank you :+1:

lucene/core/src/java/org/apache/lucene/search/ConstantScoreScorer.java
cbismuth: After a tiny debugging session, it seems we're passing through the `AssertingScorable#setMinCompetitiveScore` which not takes the previous float.

https://github.com/apache/lucene-solr/blob/master/lucene/test-framework/src/java/org/apache/lucene/search/AssertingScorable.java#L41-L44

jimczi: Yes this is because you need to add an indirection in `ConstantScoreScorer#iterator`. Something like:
````
 @Override
  public DocIdSetIterator iterator() {
    // add indirection so that if 'it' is updated then it will
    // be taken into account
    return new DocIdSetIterator() {
      @Override
      public int nextDoc() throws IOException {
        return doc = disi.nextDoc();
      }

      @Override
      public int docID() {
        return doc;
      }

      @Override
      public long cost() {
        return disi.cost();
      }

      @Override
        return doc = disi.advance(target);
      }
    };
  }
````
You replace `disi` with an empty one when the min score is greater than the query score so you  cannot return it directly there. You also need to save the current doc to make sure that you don't lost it when you reset the disi.

Comment (jimczi) [reply lucene/core/src/java/org/apache/lucene/search/ConstantScoreScorer.java]: And for the record, [here](https://github.com/apache/lucene-solr/blob/8d7d5aefd477dc86a35cc3eb137a8a45534db285/lucene/core/src/java/org/apache/lucene/search/TopScoreDocCollector.java#L244) is where we set the minimum score to the next float. The `AssertingScorable` just delegate the call to the underlying scorer.

Comment (cbismuth) [reply lucene/core/src/java/org/apache/lucene/search/ConstantScoreScorer.java]: Thanks **a lot** for your time, these are my Lucene baby steps. I'm currently debugging to understand why we need to add an indirection in `ConstantScoreScorer#iterator`.

I understand that we need to save the current doc to not lose it. But even when I don't save it, the indirection you suggest make the test pass and I don't understand why.

Comment (jimczi) [reply lucene/core/src/java/org/apache/lucene/search/ConstantScoreScorer.java]: > I understand that we need to save the current doc to not lose it. But even when I don't save it, the indirection you suggest make the test pass and I don't understand why. 

That's not related to this test. When `minCompetitiveScore` is called the doc id iterator should stay on the current doc and only return NO_MORE_DOCS when it is advanced (nextDoc or advance). You can add a test to check this by calling `minCompetitiveScore` in the test and then check that the iterator still return the current doc until it is advanced.

Comment (cbismuth) [reply lucene/core/src/java/org/apache/lucene/search/ConstantScoreScorer.java]: Yes, thank you, I'll add an assertion in the dedicated test :+1:

**Comment by cbismuth** [comment]
Please ignore my previous comment as some tests are broken when the scorer has an approximation (see [here](https://github.com/apache/lucene-solr/blob/1acfca5ebcc4eb8600fe0fc0def160f610866f72/lucene/core/src/java/org/apache/lucene/search/Weight.java#L267-L273) and [here](https://github.com/apache/lucene-solr/blob/910a0231f6fc668426056e31d43e293248ff5ce1/lucene/test-framework/src/java/org/apache/lucene/search/AssertingLeafCollector.java#L48)).

I'm working on fixing them.

**Comment by cbismuth** [comment]
Great! Exactly what I did except declaring a static inner class, I’ll declare it :+1: thanks!

File: lucene/core/src/java/org/apache/lucene/search/ConstantScoreScorer.java [hunk 4]
```diff
@@ -70,7 +98,7 @@ public TwoPhaseIterator twoPhaseIterator() {
 
   @Override
   public int docID() {
-    return disi.docID();
+    return twoPhaseIterator != null ? twoPhaseIterator.approximation().docID() : doc;
```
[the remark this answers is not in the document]
cbismuth: Yes, it broke some tests.

cbismuth:
Well you're right, without a static wrapper (see 19e156cedd343b27214538aec58f24e1d66105da) I can't assign approximation to `DocIdSetIterator.empty()` after setting `minCompetitiveScore`.

I'll update this PR with your suggestions and I let you know, thanks a lot @jimczi for your help!

cbismuth:
Hi @jimczi, PR is up-to-date and ready for review. All Lucene tests are green on my side.

As you'll see, I'd like to suggest a slightly different implementation to keep `approximation` class attribute `null` when no two phase iterator has been passed to the `ConstantScoreScorer` constructor. Feel free to let me now if you prefer to keep your proposal, thanks a lot.

cbismuth:
I think I've understood my mistake and it should be fixed now.

I've added a random condition in tests to check the second `ConstantScoreScorer` constructor. Even though tests pass, I'm not very confident with my [two phase iterator implementation](https://github.com/apache/lucene-solr/pull/495/files#diff-de7ae2ad20402371f11166a7486dade7R119), could you please give it a review? Thank you.

lucene/core/src/java/org/apache/lucene/search/ConstantScoreScorer.java
```diff
@@ -49,15 +100,27 @@ public ConstantScoreScorer(Weight weight, float score, DocIdSetIterator disi) {
...
   }
 
   @Override
   public float getMaxScore(int upTo) throws IOException {
     return score;
   }
 
+  @Override
+  public void setMinCompetitiveScore(float minScore) throws IOException {
+    if (minScore > score) {
```
jimczi: See my comment above but it should be enough to reset `approximation.delegate` if you wrap the twoPhaseIterator correctly.

cbismuth:
Thank you @jimczi.

I've added parametrized tests with phrase and term queries to check the two constructors of the `ConstantScoreScorer` class.
And I also came back to your original proposal (with `this.disi = this.approximation`) as it's the best fit to reset the iterator.

I hope it's a good one!

lucene/core/src/java/org/apache/lucene/search/ConstantScoreScorer.java
```diff
@@ -70,7 +120,7 @@ public TwoPhaseIterator twoPhaseIterator() {
 
   @Override
   public int docID() {
-    return disi.docID();
```
jimczi: You can keep `disi.docID()` here and move `doc` to the `DocIdSetIteratorWrapper` ?

lucene/core/src/test/org/apache/lucene/search/TestConstantScoreScorer.java
```diff
@@ -0,0 +1,141 @@
...
+import org.apache.lucene.util.LuceneTestCase;
+import org.junit.Test;
+
+import static org.apache.lucene.search.BooleanClause.Occur;
+import static org.apache.lucene.search.DocIdSetIterator.NO_MORE_DOCS;
+import static org.hamcrest.CoreMatchers.equalTo;
+
+public class TestConstantScoreScorer extends LuceneTestCase {
+  private static final String FIELD = "f";
+  private static final String[] VALUES = new String[]{"foo", "bar", "foo bar", "azerty"};
```
jimczi: Can you add documents that match `foo` and `bar` but not as a phrase query to test the two phase matching ?

lucene/core/src/test/org/apache/lucene/search/TestConstantScoreScorer.java
```diff
@@ -0,0 +1,141 @@
...
+
+  @ParametersFactory(argumentFormatting = "query=%s")
```
jimczi: Can you make explicit test case instead ? We don't have permutations to tests here so being explicit should simplify the reading ?

lucene/core/src/java/org/apache/lucene/search/ConstantScoreScorer.java
cbismuth: Totally, fixed in 114013f86a7a3b12aa70e7ee6bea4d1211a1448b, thanks :+1:

lucene/core/src/test/org/apache/lucene/search/TestConstantScoreScorer.java
cbismuth: Sure! Done in faa0c77adfbc0bd5e296a93975347c8753a91dbf.

lucene/core/src/test/org/apache/lucene/search/TestConstantScoreScorer.java
cbismuth: Yes, it's part of faa0c77adfbc0bd5e296a93975347c8753a91dbf.

cbismuth:
I'll default the score mode to `ScoreMode.TOP_SCORES` in all 45 constructor callers. And then, I'll try to change it to a more sensible value based on the caller context.

cbismuth:
Lucene tests are all green on my side, I'll launch Solr ones later this day.

I'm not at ease with these two changes [here](https://github.com/apache/lucene-solr/pull/495/files#diff-4c736ed817068ac157cd867fa1dc6418R769) and [there](https://github.com/apache/lucene-solr/pull/495/files#diff-4c736ed817068ac157cd867fa1dc6418R847). I'm not sure I can _easily_ retrieve the score mode from the `CachingWrapperWeight` class, therefore I passed `ScoreMode.TOP_SCORE` to not prevent from setting min competitive score.

What do you think about it?

lucene/core/src/test/org/apache/lucene/search/TestPointQueries.java
```diff
@@ -1903,9 +1903,6 @@ public void testRangeOptimizesIfAllPointsMatch() throws IOException {
       upperBound[i] = value[i] + random().nextInt(1);
     }
     Query query = IntPoint.newRangeQuery("point", lowerBound, upperBound);
-    Weight weight = searcher.createWeight(query, ScoreMode.COMPLETE_NO_SCORES, 1);
-    Scorer scorer = weight.scorer(searcher.getIndexReader().leaves().get(0));
-    assertEquals(DocIdSetIterator.all(1).getClass(), scorer.iterator().getClass());
```
jimczi: The iterator is not wrapped when the score mode is set to `COMPLETE_NO_SCORES` so you don't need to change this assertion anymore ?

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
```diff
@@ -766,7 +766,7 @@ public ScorerSupplier scorerSupplier(LeafReaderContext context) throws IOExcepti
       return new ScorerSupplier() {
         @Override
         public Scorer get(long LeadCost) throws IOException {
-          return new ConstantScoreScorer(CachingWrapperWeight.this, 0f, disi);
+          return new ConstantScoreScorer(CachingWrapperWeight.this, 0f, ScoreMode.TOP_SCORES, disi);
```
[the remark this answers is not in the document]
cbismuth: I see, [query caching is related to filters](https://wiki.apache.org/lucene-java/LuceneFAQ?highlight=%28query%29%7C%28cache%29#Can_I_cache_search_results_with_Lucene.3F), so no score, thanks :+1:

Fixed in d114c02d4cd035b2c0d4480225f8064698784a7e.

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
```diff
@@ -844,7 +844,7 @@ public BulkScorer bulkScorer(LeafReaderContext context) throws IOException {
         return null;
       }
 
-      return new DefaultBulkScorer(new ConstantScoreScorer(this, 0f, disi));
+      return new DefaultBulkScorer(new ConstantScoreScorer(this, 0f, ScoreMode.TOP_SCORES, disi));
```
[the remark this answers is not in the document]
cbismuth: Same here, part of d114c02d4cd035b2c0d4480225f8064698784a7e.

lucene/core/src/test/org/apache/lucene/search/TestConjunctionDISI.java
```diff
@@ -234,12 +234,12 @@ public void testConjunction() throws IOException {
           case 0:
             // simple iterator
             sets[i] = set;
-            iterators[i] = new ConstantScoreScorer(new FakeWeight(), 0f, anonymizeIterator(new BitDocIdSet(set).iterator()));
+            iterators[i] = new ConstantScoreScorer(new FakeWeight(), 0f, ScoreMode.TOP_SCORES, anonymizeIterator(new BitDocIdSet(set).iterator()));
```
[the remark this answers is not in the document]
cbismuth: Totally, allow to visit all results and score isn't used in test :+1:

Fixed in d5c82d855c152fc958574076b5d14ae1cbf6158b.

cbismuth:
Ant `precommit` task and Lucene test are all green.

lucene/core/src/test/org/apache/lucene/search/TestConstantScoreScorer.java
```diff
@@ -0,0 +1,215 @@
...
+      writer.forceMerge(1);
+
+      reader = writer.getReader();
+      writer.close();
+    }
+
+    ConstantScoreScorer constantScoreScorer(Query query, float score, ScoreMode scoreMode) throws IOException {
+      IndexSearcher searcher = newSearcher(reader);
+      Weight weight = searcher.createWeight(new ConstantScoreQuery(query), scoreMode, 1);
+      LeafReaderContext context = searcher.getIndexReader().leaves().get(0);
```
jpountz: let's assert that leaves.size() is 1?

cbismuth: Sure, fixed in c5a83179ebd4a2775e24bd6518ea445cc2cb700e, thanks :+1:

PR #497, opened 2018-11-12
LUCENE-8026: ExitableDirectoryReader does not instrument points

cbismuth:
> This means it cannot interrupt range or geo queries.

See [LUCENE-8026](https://issues.apache.org/jira/browse/LUCENE-8026).

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/ExitableDirectoryReader.java
+++ b/lucene/core/src/java/org/apache/lucene/index/ExitableDirectoryReader.java
@@ -211,13 +211,15 @@ public class ExitableDirectoryReader extends FilterDirectoryReader {
      * or if {@link Thread#interrupted()} returns true.
      */
     private void checkAndThrow() {
-      if (calls++ % MAX_CALLS_BEFORE_QUERY_TIMEOUT_CHECK == 0 && queryTimeout.shouldExit()) {
-        throw new ExitingReaderException("The request took too long to intersect point values. Timeout: "
-            + queryTimeout.toString()
-            + ", PointValues=" + in
-        );
-      } else if (Thread.interrupted()) {
-        throw new ExitingReaderException("Interrupted while intersecting point values. PointValues=" + in);
+      if (calls++ % MAX_CALLS_BEFORE_QUERY_TIMEOUT_CHECK == 0) {
+        if (queryTimeout.shouldExit()) {
+          throw new ExitingReaderException("The request took too long to intersect point values. Timeout: "
+              + queryTimeout.toString()
+              + ", PointValues=" + in
+          );
+        } else if (Thread.interrupted()) {
+          throw new ExitingReaderException("Interrupted while intersecting point values. PointValues=" + in);
+        }
       }
     }
 
--- a/lucene/core/src/java/org/apache/lucene/index/ExitableDirectoryReader.java
+++ b/lucene/core/src/java/org/apache/lucene/index/ExitableDirectoryReader.java
@@ -129,12 +129,12 @@ public class ExitableDirectoryReader extends FilterDirectoryReader {
      */
     private void checkAndThrow() {
       if (queryTimeout.shouldExit()) {
-        throw new ExitingReaderException("The request took too long to iterate over terms. Timeout: "
+        throw new ExitingReaderException("The request took too long to iterate over point values. Timeout: "
             + queryTimeout.toString()
             + ", PointValues=" + in
         );
       } else if (Thread.interrupted()) {
-        throw new ExitingReaderException("Interrupted while iterating over terms. PointValues=" + in);
+        throw new ExitingReaderException("Interrupted while iterating over point values. PointValues=" + in);
       }
     }
 
--- a/lucene/core/src/java/org/apache/lucene/index/ExitableDirectoryReader.java
+++ b/lucene/core/src/java/org/apache/lucene/index/ExitableDirectoryReader.java
@@ -112,13 +112,12 @@ public class ExitableDirectoryReader extends FilterDirectoryReader {
   /**
    * Wrapper class for another PointValues implementation that is used by ExitableFields.
    */
-  public static class ExitablePointValues extends PointValues {
+  private static class ExitablePointValues extends PointValues {
 
     private final PointValues in;
     private final QueryTimeout queryTimeout;
 
-    /** Constructor **/
-    public ExitablePointValues(PointValues in, QueryTimeout queryTimeout) {
+    private ExitablePointValues(PointValues in, QueryTimeout queryTimeout) {
       this.in = in;
       this.queryTimeout = queryTimeout;
       checkAndThrow();
--- a/lucene/core/src/java/org/apache/lucene/index/ExitableDirectoryReader.java
+++ b/lucene/core/src/java/org/apache/lucene/index/ExitableDirectoryReader.java
@@ -210,38 +210,44 @@ public class ExitableDirectoryReader extends FilterDirectoryReader {
      * Throws {@link ExitingReaderException} if {@link QueryTimeout#shouldExit()} returns true,
      * or if {@link Thread#interrupted()} returns true.
      */
-    private void checkAndThrow() {
+    private void checkAndThrowWithSampling() {
       if (calls++ % MAX_CALLS_BEFORE_QUERY_TIMEOUT_CHECK == 0) {
-        if (queryTimeout.shouldExit()) {
-          throw new ExitingReaderException("The request took too long to intersect point values. Timeout: "
-              + queryTimeout.toString()
-              + ", PointValues=" + in
-          );
-        } else if (Thread.interrupted()) {
-          throw new ExitingReaderException("Interrupted while intersecting point values. PointValues=" + in);
-        }
+        checkAndThrow();
+      }
+    }
+
+    private void checkAndThrow() {
+      if (queryTimeout.shouldExit()) {
+        throw new ExitingReaderException("The request took too long to intersect point values. Timeout: "
+            + queryTimeout.toString()
+            + ", PointValues=" + in
+        );
+      } else if (Thread.interrupted()) {
+        throw new ExitingReaderException("Interrupted while intersecting point values. PointValues=" + in);
       }
     }
 
     @Override
     public void visit(int docID) throws IOException {
-      checkAndThrow();
+      checkAndThrowWithSampling();
       in.visit(docID);
     }
 
     @Override
     public void visit(int docID, byte[] packedValue) throws IOException {
-      checkAndThrow();
+      checkAndThrowWithSampling();
       in.visit(docID, packedValue);
     }
 
     @Override
     public PointValues.Relation compare(byte[] minPackedValue, byte[] maxPackedValue) {
+      checkAndThrow();
       return in.compare(minPackedValue, maxPackedValue);
     }
 
     @Override
     public void grow(int count) {
+      checkAndThrow();
       in.grow(count);
     }
   }
--- a/lucene/core/src/java/org/apache/lucene/index/ExitableDirectoryReader.java
+++ b/lucene/core/src/java/org/apache/lucene/index/ExitableDirectoryReader.java
@@ -193,15 +193,15 @@ public class ExitableDirectoryReader extends FilterDirectoryReader {
     }
   }
 
-  public static class ExitableIntersectVisitor implements PointValues.IntersectVisitor {
+  private static class ExitableIntersectVisitor implements PointValues.IntersectVisitor {
 
-    public static final int MAX_CALLS_BEFORE_QUERY_TIMEOUT_CHECK = 10;
+    private static final int MAX_CALLS_BEFORE_QUERY_TIMEOUT_CHECK = 10;
 
     private final PointValues.IntersectVisitor in;
     private final QueryTimeout queryTimeout;
     private int calls;
 
-    public ExitableIntersectVisitor(PointValues.IntersectVisitor in, QueryTimeout queryTimeout) {
+    private ExitableIntersectVisitor(PointValues.IntersectVisitor in, QueryTimeout queryTimeout) {
       this.in = in;
       this.queryTimeout = queryTimeout;
     }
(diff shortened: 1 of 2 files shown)


lucene/core/src/java/org/apache/lucene/index/ExitableDirectoryReader.java
```diff
@@ -100,13 +109,97 @@ public CacheHelper getCoreCacheHelper() {
...
+      checkAndThrow();
+    }
+
+    /**
+     * Throws {@link ExitingReaderException} if {@link QueryTimeout#shouldExit()} returns true,
+     * or if {@link Thread#interrupted()} returns true.
+     */
+    private void checkAndThrow() {
+      if (queryTimeout.shouldExit()) {
+        throw new ExitingReaderException("The request took too long to iterate over terms. Timeout: "
```
mikemccand: Maybe change `to iterate over terms` to `to iterate over points`?

lucene/core/src/java/org/apache/lucene/index/ExitableDirectoryReader.java
```diff
@@ -100,13 +109,97 @@ public CacheHelper getCoreCacheHelper() {
...
+            + queryTimeout.toString()
+            + ", PointValues=" + in
+        );
+      } else if (Thread.interrupted()) {
+        throw new ExitingReaderException("Interrupted while iterating over terms. PointValues=" + in);
```
mikemccand: Same here?

lucene/core/src/test/org/apache/lucene/index/TestExitableDirectoryReader.java
```diff
@@ -160,5 +160,78 @@ public void testExitableFilterIndexReader() throws Exception {
 
     directory.close();
   }
+
+  /**
+   * Tests timing out of PointValues queries
+   *
+   * @throws Exception on error
+   */
+  @Ignore("this test relies on wall clock time and sometimes false fails")
```
mikemccand: I wonder if we could make a mock clock and then have deterministic control and be able to re-enable these tests?

lucene/core/src/java/org/apache/lucene/index/ExitableDirectoryReader.java
```diff
@@ -100,13 +109,97 @@ public CacheHelper getCoreCacheHelper() {
...
+      }
+    }
+
+    @Override
+    public void intersect(IntersectVisitor visitor) throws IOException {
+      checkAndThrow();
+      in.intersect(visitor);
```
mikemccand: A lot of time/effort can be spent in this (recursive) intersect call -- should we also wrap the `IntersectVisitor` and sometimes check for timeout?

lucene/core/src/java/org/apache/lucene/index/ExitableDirectoryReader.java
cbismuth: I missed this one ... yes :+1:

lucene/core/src/java/org/apache/lucene/index/ExitableDirectoryReader.java
cbismuth: Same here.

lucene/core/src/java/org/apache/lucene/index/ExitableDirectoryReader.java
cbismuth: I think it would be better, yes. I'll suggest a change like this, thank you.

lucene/core/src/test/org/apache/lucene/index/TestExitableDirectoryReader.java
cbismuth: I'll try to do so, yes, thanks :+1:

lucene/core/src/java/org/apache/lucene/index/ExitableDirectoryReader.java
cbismuth: Fixed in 6c6e45be3b645814b9871db3b33b799a7c31296d.

lucene/core/src/java/org/apache/lucene/index/ExitableDirectoryReader.java
cbismuth: Fixed in 6c6e45be3b645814b9871db3b33b799a7c31296d.

lucene/core/src/test/org/apache/lucene/index/TestExitableDirectoryReader.java
cbismuth: Fixed in 2b3dfbcba8944dd48c22c341699da528c8525f91.

lucene/core/src/java/org/apache/lucene/index/ExitableDirectoryReader.java
cbismuth: Fixed in 5587aa2c904ba135da11a89f47cfa035901b046d.

lucene/core/src/java/org/apache/lucene/index/ExitableDirectoryReader.java
```diff
@@ -100,13 +109,157 @@ public CacheHelper getCoreCacheHelper() {
...
+     * Throws {@link ExitingReaderException} if {@link QueryTimeout#shouldExit()} returns true,
+     * or if {@link Thread#interrupted()} returns true.
+     */
+    private void checkAndThrow() {
+      if (calls++ % maxCallsBeforeQueryTimeoutCheck == 0 && queryTimeout.shouldExit()) {
+        throw new ExitingReaderException("The request took too long to intersect point values. Timeout: "
+            + queryTimeout.toString()
+            + ", PointValues=" + in
+        );
+      } else if (Thread.interrupted()) {
```
jpountz: should this also not be called everytime and moved under the `if (calls++ % maxCallsBeforeQueryTimeoutCheck == 0)` test?

jpountz: also sampling makes sense for the two visit() methods which are called in tight loops, but less for relate and grow, should we only sample visit() calls?

lucene/core/src/java/org/apache/lucene/index/ExitableDirectoryReader.java
```diff
@@ -100,13 +109,157 @@ public CacheHelper getCoreCacheHelper() {
...
+      this.queryTimeout = queryTimeout;
+      this.maxCallsBeforeQueryTimeoutCheck = maxCallsBeforeQueryTimeoutCheck;
+    }
+
+    /**
+     * Throws {@link ExitingReaderException} if {@link QueryTimeout#shouldExit()} returns true,
+     * or if {@link Thread#interrupted()} returns true.
+     */
+    private void checkAndThrow() {
+      if (calls++ % maxCallsBeforeQueryTimeoutCheck == 0 && queryTimeout.shouldExit()) {
```
jpountz: :+1: to sampling checks to reduce overhead.

lucene/core/src/test/org/apache/lucene/index/TestExitableDirectoryReader.java
```diff
@@ -152,13 +150,130 @@ public void testExitableFilterIndexReader() throws Exception {
...
+    // Set a fairly high timeout value (1 second) and expect the query to complete in that time frame.
+    // Not checking the validity of the result, all we are bothered about in this test is the timing out.
+    directoryReader = DirectoryReader.open(directory);
+    exitableDirectoryReader = new ExitableDirectoryReader(directoryReader, inifiniteQueryTimeout());
+    reader = new TestReader(getOnlyLeafReader(exitableDirectoryReader));
+    searcher = new IndexSearcher(reader);
+    searcher.search(query, 10);
+    reader.close();
+
+    // Set a really low timeout value (1 millisecond) and expect an Exception
```
jpountz: comment says 1 milli but code says immediate

lucene/core/src/test/org/apache/lucene/index/TestExitableDirectoryReader.java
```diff
@@ -152,13 +150,130 @@ public void testExitableFilterIndexReader() throws Exception {
...
+      public boolean shouldExit() {
+        return true;
+      }
+
+      @Override
+      public boolean isTimeoutEnabled() {
+        return true;
+      }
+    };
+  }
```
jpountz: :+1:

lucene/core/src/java/org/apache/lucene/index/ExitableDirectoryReader.java
```diff
@@ -100,13 +109,157 @@ public CacheHelper getCoreCacheHelper() {
...
+    public ExitableIntersectVisitor(PointValues.IntersectVisitor in, QueryTimeout queryTimeout) {
+      this(in, queryTimeout, DEFAULT_MAX_CALLS_BEFORE_QUERY_TIMEOUT_CHECK);
+    }
+
+    public ExitableIntersectVisitor(PointValues.IntersectVisitor in,
+                                    QueryTimeout queryTimeout, int maxCallsBeforeQueryTimeoutCheck) {
+      this.in = in;
+      this.queryTimeout = queryTimeout;
+      this.maxCallsBeforeQueryTimeoutCheck = maxCallsBeforeQueryTimeoutCheck;
+    }
```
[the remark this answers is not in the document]
cbismuth: Yes, fixed in dfd1403e02b270ba0712c0a40ee14dbf9e7719fd.

lucene/core/src/java/org/apache/lucene/index/ExitableDirectoryReader.java
cbismuth: My bad, outer condition fixed in 9e11f76b2ac7b9cda9f1c78ad0c1f3a5be09f435 and only sample `visit` methods fixed in 3d09426a81277a5344435f43a1942570c1a85f37.

lucene/core/src/test/org/apache/lucene/index/TestExitableDirectoryReader.java
cbismuth: Totally, comments fixed in 45996c885d3e52edcc21d47e40db443895982af5.

lucene/core/src/java/org/apache/lucene/index/ExitableDirectoryReader.java
```diff
@@ -100,13 +109,150 @@ public CacheHelper getCoreCacheHelper() {
...
+  }
+
+  public static class ExitableIntersectVisitor implements PointValues.IntersectVisitor {
+
+    public static final int MAX_CALLS_BEFORE_QUERY_TIMEOUT_CHECK = 10;
+
+    private final PointValues.IntersectVisitor in;
+    private final QueryTimeout queryTimeout;
+    private int calls = 0;
```
mikemccand: Minor: you don't need the `= 0` -- java does that for you by default.

cbismuth: Yes, removed in 11fd3de51c31ef4c0e2a660ce22df018d194ce7e, thank you.

lucene/core/src/java/org/apache/lucene/index/ExitableDirectoryReader.java
jpountz: I was thinking of keeping the checkAndThrow in other methods too, just without sampling. Though I guess most time is going to be spent in visit() anyway so the current version isn't bad.

cbismuth: My bad, fixed in 7eef7e9a8213714bdc3a90aa7fa5c5db010d5151.

jpountz:
`ant precommit` is complaining about some missing javadocs, could you address this? (the solution to some of these issues might be to make classes private rather than adding docs)

lucene/core/src/java/org/apache/lucene/index/ExitableDirectoryReader.java
```diff
@@ -100,13 +109,156 @@ public CacheHelper getCoreCacheHelper() {
...
 
+  /**
+   * Wrapper class for another PointValues implementation that is used by ExitableFields.
+   */
+  public static class ExitablePointValues extends PointValues {
+
+    private final PointValues in;
+    private final QueryTimeout queryTimeout;
+
+    public ExitablePointValues(PointValues in, QueryTimeout queryTimeout) {
```
jpountz: I'm getting precommit failures about missing javadocs for this constructor, aren't you?

cbismuth: Issue found: I forgot to add a `set -e` statement in my tiny bash script below to stop on first failure.

Let me run another `precommit` Ant task and push.

```bash
#!/usr/bin/env bash

set -e

cd ${HOME}/git/lucene-solr
find . -name "*.lck" -delete
ant clean compile precommit
ant -f lucene/build.xml test
```

cbismuth: That's it, documentation fixed and `precommit` task really green :sweat_smile: thanks!

pzygielo: But isn't it worse now? precommit is satisfied, but nothing more than (obvious) noise was added.

cbismuth: I'm not sure to understand, the `set -e` flag prevents the second Ant command line from being run if the `precommit` task fails, and therefore the latest output in my terminal is a build failure.

The `set -x` (debug option) flag would add a lot noise.

lucene/core/src/java/org/apache/lucene/index/ExitableDirectoryReader.java
```diff
@@ -117,6 +117,7 @@ public CacheHelper getCoreCacheHelper() {
     private final PointValues in;
     private final QueryTimeout queryTimeout;
 
+    /** Constructor **/
```
pzygielo: My comment
> But isn't it worse now? precommit is satisfied, but nothing more than (obvious) noise was added.

was for this line.

cbismuth: My bad, I though I used this class in tests, but I didn't. I've turned it `private` in 6f48d4f851b4eccf4c3d8c4cfbe6d540e13b4d85, thanks :+1:

PR #500, opened 2018-11-17
LUCENE-8517: do not wrap FixedShingleFilter with conditional in TestR…

msokolov:
…andomChains

### DIFF
--- a/lucene/test-framework/src/java/org/apache/lucene/analysis/ValidatingTokenFilter.java
+++ b/lucene/test-framework/src/java/org/apache/lucene/analysis/ValidatingTokenFilter.java
@@ -181,8 +181,8 @@ public final class ValidatingTokenFilter extends TokenFilter {
 
   /**
    * Prints details about consumed tokens stored in any ValidatingTokenFilters in the input chain
-   * @param in
-   * @param out
+   * @param in the input token stream
+   * @param out the output print stream
    */
   public static void dumpValidatingTokenFilters(TokenStream in, PrintStream out) {
     if (in instanceof TokenFilter) {
--- a/lucene/test-framework/src/java/org/apache/lucene/analysis/ValidatingTokenFilter.java
+++ b/lucene/test-framework/src/java/org/apache/lucene/analysis/ValidatingTokenFilter.java
@@ -56,7 +56,7 @@ public final class ValidatingTokenFilter extends TokenFilter {
   private final OffsetAttribute offsetAtt = getAttribute(OffsetAttribute.class);
   private final CharTermAttribute termAtt = getAttribute(CharTermAttribute.class);
 
-  // record all the Tokens seen so they can be dumped on failure
+  // record the last MAX_DEBUG_TOKENS tokens seen so they can be dumped on failure
   private final List<Token> tokens = new LinkedList<>();
 
   private final String name;
--- a/lucene/test-framework/src/java/org/apache/lucene/analysis/ValidatingTokenFilter.java
+++ b/lucene/test-framework/src/java/org/apache/lucene/analysis/ValidatingTokenFilter.java
@@ -17,7 +17,11 @@
 package org.apache.lucene.analysis;
 
 import java.io.IOException;
+import java.io.PrintStream;
 import java.util.HashMap;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 
 import org.apache.lucene.analysis.tokenattributes.CharTermAttribute;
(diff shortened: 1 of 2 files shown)


lucene/test-framework/src/java/org/apache/lucene/analysis/TokenStreamUtil.java
```diff
@@ -0,0 +1,41 @@
...
+
+  private TokenStreamUtil() {
+  }
+
+  /**
+   * Prints details about consumed tokens stored in any ValidatingTokenFilters in the input chain
+   * @param in
+   * @param out
+   */
+  public static void dumpValidatingTokenFilters(TokenStream in, PrintStream out) {
```
romseygeek: I think this method can just be a private method on ValidatingTokenFilter?  No need for the extra class

lucene/test-framework/src/java/org/apache/lucene/analysis/ValidatingTokenFilter.java
```diff
@@ -50,6 +53,9 @@
   private final OffsetAttribute offsetAtt = getAttribute(OffsetAttribute.class);
   private final CharTermAttribute termAtt = getAttribute(CharTermAttribute.class);
 
+  // record all the Tokens seen so they can be dumped on failure
+  private final List<Token> tokens = new ArrayList<>();
```
romseygeek: I slightly worry here that we are exposing ourselves to spurious out-of-memory failures during testRandomStrings() or similar; can we maybe limit this to just the last 200 tokens or something like that?

romseygeek: This also needs to be emptied on `reset()`

msokolov: Oh I hadn't realized RandomChains could create such large strings that would be a concern!  Yes, I don't think it's helpful to have even 200 tokens printed out. In general your interest here will be in the final few tokens, so I think we can limit to 20 or so.

lucene/test-framework/src/java/org/apache/lucene/analysis/TokenStreamUtil.java
msokolov: Yes - I started from a broader project including the ability to inject arbitrary filters into any token stream. I was thinking of `ToDotTokenFilter`, but simplified it along the way since I didn't need that for debugging here. I'll move into VTF and create this later as needed.

msokolov:
Ah, I see. I wonder if it would be a huge effort to support building with
JDK11 at least? IMO one could skip 9 and 10 since even openjdk is not
"supporting" them with bug fixes, right? Anyway I'd be happy to help out if
that would be welcome

uschindler:
Hi,

Currently a Lucene release should be done with Java 8. Therefore all checks like precommit have to be done with Java 8, otherwise we can’t be sure that it really works with Java 8.

Lucene can for sure be built with Java 9 or later, but some developer-central tasks cannot be executed for various reasons (that are not necessary for building from source). The problem here is with the ECJ tool that is used for precommit checks. It’s the Eclipse compiler and this one needs a version that support the module system. Unfortunately there were some problems with updating it as it broke some parts of the checks (I tried it a while back, mabye it’s better now). Other tools that won’t work are the Javadocs linter, because the format of Javadocs differs in Java 9+ and maintaining 2 different HTML parsers is too hard (because the Javadocs in the release JAR/ZIP/TGZ are Java 8 anyways).

As a contributor/committer of Lucene you have to use Java 8 at least for the quality checks so we can be sure that all is fine with your code! As an end user, you can build lucene with any version later – this is also tested with Jenkins.

Uwe

lucene/test-framework/src/java/org/apache/lucene/analysis/ValidatingTokenFilter.java
```diff
@@ -50,6 +56,9 @@
   private final OffsetAttribute offsetAtt = getAttribute(OffsetAttribute.class);
   private final CharTermAttribute termAtt = getAttribute(CharTermAttribute.class);
 
+  // record all the Tokens seen so they can be dumped on failure
```
mikemccand: Update comment to `record last MAX_DEBUG_TOKENS...`?

msokolov: oh good, thanks. I pushed an edit

msokolov:
BTW - I built w/Java 8 and precommit / test are passing for me now. @romseygeek if you are planning to address the apparent graph-handling bug in FixedShingleFilter, I can remove the inhibition on making it conditional in TestRandomChains, or we could open an issue to track it and push this fix with a JIRA tagged in a comment? Separately, I might take a look at that problem, but I'd like to separate that from this test framework fix.

PR #501, opened 2018-11-18
SOLR-5211

moshebla:


solr/core/src/java/org/apache/solr/update/AddUpdateCommand.java
```diff
@@ -96,12 +96,21 @@ public SolrInputDocument getSolrInputDocument() {
    * Any changes made to the returned Document will not be reflected in the SolrInputDocument, or future calls to this
    * method.
    * Note that the behavior of this is sensitive to {@link #isInPlaceUpdate()}.
+   * @param withBlockId If true, then block id is forcibly added to the doc
    */
-   public Document getLuceneDocument() {
+   Document getLuceneDocument(boolean withBlockId) {
```
dsmiley: can we call this addRootField ?

dsmiley: actually I'm not sure we actually need to overload getLuceneDocument with this boolean after all.  if "isInPlaceUpdate()" then don't add root but otherwise add it.  That'll be fine; right?

solr/core/src/java/org/apache/solr/update/AddUpdateCommand.java
```diff
@@ -194,20 +203,22 @@ public String getHashableId() {
...
-      if(isVersion) sdoc.setField(CommonParams.VERSION_FIELD, version);
-      // TODO: if possible concurrent modification exception (if SolrInputDocument not cloned and is being forwarded to replicas)
-      // then we could add this field to the generated lucene document instead.
+      addBlockId(sdoc, rootId);
     }
 
     return () -> all.stream().map(sdoc -> DocumentBuilder.toDocument(sdoc, req.getSchema())).iterator();
   }
 
+  private void addBlockId(SolrInputDocument sdoc, String rootId) {
```
dsmiley: Can we rename this addRootField?  and separate out the version part to addVersionField which should _not_ be called by this method; the caller should.  addVersionField only needs to be called for the nested case, not by getLuceneDocument.

solr/core/src/java/org/apache/solr/update/AddUpdateCommand.java
```diff
@@ -194,20 +203,22 @@ public String getHashableId() {
...
+    boolean isVersion = version != 0;
+    sdoc.setField(IndexSchema.ROOT_FIELD_NAME, rootId);
+    if(isVersion) sdoc.setField(CommonParams.VERSION_FIELD, version);
```
dsmiley: needs a comment as said in JIRA why we add the version here

solr/core/src/java/org/apache/solr/update/DirectUpdateHandler2.java
```diff
@@ -975,6 +976,14 @@ private void updateDocOrDocValues(AddUpdateCommand cmd, IndexWriter writer) thro
     }
   }
 
+  boolean forciblyAddBlockId() {
```
dsmiley: this is here in part so that we can test _not_ doing this (i.e. current behavior)... but I don't think it's worth it.  If we want to _not_ do it conditionally then it could be via luceneMatchVersion but I don't care either way as it's minor to add it unconditionally.

solr/core/src/java/org/apache/solr/update/AddUpdateCommand.java
dsmiley: although it may be a slight bit of scope creep; I think this logic would be more clear why it's needed if it merely propagated the SolrInputField from the root doc to the child docs.  My understanding is that this will wind up being equivalent to what happens now, which is very non-obvious; it took a bunch of digging on my part to see.

solr/core/src/java/org/apache/solr/update/DirectUpdateHandler2.java
moshebla: The change described above causes BlockUpdateTest#testLegacyBlockProcessing to fail.
Should I remove it, I noticed you mentioned this being a burden going forward, or should I create a new solrconfig and test case to test this works in older solr versions?
Wouldn't this be a problem if _root_ is not defined in the collection's schema?

dsmiley: Come to think of it, who cares if root is populated or not for "old" schemas?  Lets just do it always; no conditional test.  If someone doesn't want root to be populated, that same person probably doesn't have child docs and should just as well omit the root field from their schema.

dsmiley:
Also please rename "block" terminology to "nested" where possible.

solr/core/src/java/org/apache/solr/update/AddUpdateCommand.java
```diff
@@ -194,20 +197,33 @@ public String getHashableId() {
       return null; // caller should call getLuceneDocument() instead
     }
 
-    String rootId = getHashableId();
-
-    boolean isVersion = version != 0;
+    final String rootId = getHashableId();
+    final boolean hasVersion = version != 0;
+    final SolrInputField versionSif = hasVersion? solrDoc.get(CommonParams.VERSION_FIELD): null;
```
dsmiley: What I meant was that *if* the solrDoc has a \_version\_ then we can copy it to all nested children.  We need not examine this.version at all.  At least this is my understanding of reading the code; it'd be good to ensure tests pass.

dsmiley:
Nice.
Do we even need the test config / schema at all?  If they don't add anything, we shouldn't have them.  Also I see there is still some "Block" terminology added here we can avoid to prefer "Nested".

solr/core/src/test/org/apache/solr/search/TestPseudoReturnFields.java
```diff
@@ -126,7 +126,7 @@ public void testAllRealFields() throws Exception {
               ,"//result/doc/str[@name='ssto']"
               ,"//result/doc/str[@name='subject']"
               
-              ,"//result/doc[count(*)=4]"
+              ,"//result/doc[count(*)=5]"
```
moshebla: The numbers were changed since now _root_ is also added to the doc,
increasing the field count by 1.

solr/core/src/test/org/apache/solr/search/TestReload.java
```diff
@@ -36,13 +36,13 @@ public void testGetRealtimeReload() throws Exception {
...
 
-    assertJQ(req("qt","/get","id","1")
+    assertJQ(req("qt","/get","id","1", "fl", "id,_version_")
         ,"=={'doc':{'id':'1','_version_':" + version + "}}"
     );
 
     h.reload();
 
-    assertJQ(req("qt","/get","id","1")
+    assertJQ(req("qt","/get","id","1", "fl", "id,_version_")
```
moshebla: _root_ is also added to the doc so I added fl param to filter it,
since, IMO, testing _root_ seemed out of scope for this test.

solr/core/src/test/org/apache/solr/BlockUpdateTest.java
```diff
@@ -0,0 +1,125 @@
...
+import org.apache.solr.common.SolrInputDocument;
+import org.apache.solr.common.params.CommonParams;
+import org.junit.BeforeClass;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+
+import static org.hamcrest.CoreMatchers.is;
+
+public class BlockUpdateTest extends SolrJettyTestBase {
```
dsmiley: Here's that "Block" terminology again.  Can the tests here merge with another test for nested docs? (one you made?)

moshebla: Um I think it would be a problem since this test uses either a schema with or without _root_ declared.
I could rename it though.
Perhaps RootFieldUpdateTest would suffice?

moshebla: Right,
I renamed it to RootFieldTest.

solr/core/src/test/org/apache/solr/cloud/TestCloudPseudoReturnFields.java
```diff
@@ -194,7 +194,7 @@ public void testAllRealFields() throws Exception {
       SolrDocumentList docs = assertSearch(params("q", "*:*", "rows", "10", "fl",fl));
       // shouldn't matter what doc we pick...
       for (SolrDocument doc : docs) {
-        assertEquals(fl + " => " + doc, 4, doc.size());
+        assertEquals(fl + " => " + doc, 5, doc.size());
```
moshebla: Doc size increased by 1 since _root_ is also returned by queries.

solr/core/src/test/org/apache/solr/cloud/TestRandomFlRTGCloud.java
```diff
@@ -70,6 +70,9 @@
 
   /** Always included in fl so we can vet what doc we're looking at */
   private static final FlValidator ID_VALIDATOR = new SimpleFieldValueValidator("id");
+
+  /** Since nested documents are not tested, when _root_ is declared in schema, it is always the same as id */
+  private static final FlValidator ROOT_VALIDATOR = new RenameFieldValueValidator("id" , "_root_");
```
moshebla: Added a validator for _root_, which is now added automatically since the schema used here declares _root_.

moshebla:
Closing since this issue has a commit in master.

PR #503, opened 2018-11-21
LUCENE-8571: Don't block on FrozenBufferedUpdates#apply during IW#processEvents

s1monw:
While indexing we try to apply frozen deletes packages concurrently
on indexing threads if necessary. This is done in an opaque way via
IndexWriter#processEvents. Yet, when we commit or refresh we have to
ensure we apply all frozen update packages before we return.
Today we execute the apply method in a blocking fashion which is unnecessary
when we are in a IndexWriter#processEvents context, we block indexing
threads while they could just continue since it's already being applied.
We also might wait in BufferedUpdatesStream when we apply all necessary updates
were we can continue with other work instead of waiting.
This change also tries to apply the packages that are not currently applied
first in order to not unnecessarily block.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/FrozenBufferedUpdates.java
+++ b/lucene/core/src/java/org/apache/lucene/index/FrozenBufferedUpdates.java
@@ -30,6 +30,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.IntConsumer;
 
 import org.apache.lucene.index.DocValuesUpdate.BinaryDocValuesUpdate;
@@ -83,6 +84,7 @@ final class FrozenBufferedUpdates {
 
   /** Counts down once all deletes/updates have been applied */
   public final CountDownLatch applied = new CountDownLatch(1);
+  private final ReentrantLock applyLock = new ReentrantLock();
 
   /** How many total documents were deleted/updated. */
   public long totalDelCount;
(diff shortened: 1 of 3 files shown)


s1monw:

s1monw:
I recommend to look at the diff with `?w=1` -> https://github.com/apache/lucene-solr/pull/503/files?w=1

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
```diff
@@ -2607,7 +2607,9 @@ synchronized long publishFrozenUpdates(FrozenBufferedUpdates packet) {
     // Do this as an event so it applies higher in the stack when we are not holding DocumentsWriterFlushQueue.purgeLock:
     eventQueue.add(w -> {
       try {
-        packet.apply(w);
+        // we call tryApply here since we don't want to block if a refresh or a flush is already applying the
+        // packet. The flush will retry this packet anyway to ensure all of the are applied
```
jpountz: s/all of the/all of them/

lucene/core/src/java/org/apache/lucene/index/FrozenBufferedUpdates.java
```diff
@@ -214,150 +216,174 @@ public FrozenBufferedUpdates(InfoStream infoStream, BufferedUpdates updates, Seg
...
     }
+    return false;
+  }
 
-    long startNS = System.nanoTime();
+  /** Translates a frozen packet of delete term/query, or doc values
+   *  updates, into their actual docIDs in the index, and applies the change.  This is a heavy
+   *  operation and is done concurrently by incoming indexing threads.
+   *  */
+  boolean forceApply(IndexWriter writer) throws IOException {
```
mikemccand: Can we make this `void`?  It always just `return true` now right?

s1monw:
pushed

msokolov:
I'm curious if you were able to measure a good improvement from this? I guess a large delete concurrent with indexing lots of docs would stress it?

s1monw:
@msokolov I measures a pretty signficant lock contention on this lock without my change. with the change it went down significantly. Refreshes got 30% faster for that benchmark (small docs with 25% collision on the IDs and soft-deletes enabled). There was some minor indexing throughput improvements but nothing major.

PR #504, opened 2018-11-22
First cut of indenting JSON formatted field value if indent=true

epugh:


epugh:
I've run "ant test" and while there are some failures, they don't appear to be related to this feature.  I'm testing it now compiling and running Solr.

epugh:
I added some nicer indenting.  For a JSON block like:
```json
{
"am I a teapot":true,
"lyrics":[
"I'm a little teapot","Short and stout","Here is my handle","Here is my spout"
]
}
```
with `indent=false`:
noindent

However, with 'indent=true`:
indented

I'd love some feedback, as I am relying on Jackson to parse the XML or JSON string, and their is some assumptions on indentation levels that I made...

solr/core/src/java/org/apache/solr/response/transform/RawValueTransformerFactory.java
```diff
@@ -79,7 +84,18 @@ public DocTransformer create(String display, SolrParams params, SolrQueryRequest
...
 
     if(apply) {
-      return new RawTransformer( field, display );
+      boolean indent = req.getParams().getBool("indent", false);
+      ObjectMapper mapper = null;
+      if (indent) {
+        if (applyToWT.equalsIgnoreCase("json")) {
+          mapper = new ObjectMapper();
+        }
+        else {
```
nsoft: There's a log more WT's possible? https://lucene.apache.org/solr/guide/7_6/response-writers.html#response-writers probably don't want to format all of then as xml, Although I admit I haven't yet figured out when applyToWT gets set, this else makes me nervous. Use of a switch would seem better, especially if we later want to indent some other format (hopefully someday wt=yaml...). Also what about wt=geojson? that's json too. (not sure if the existing code handles :[geojson] though...)

epugh:
This whole PR may be too icky to really press forward with.  Will think about it some more if htere is a better route.

PR #510, opened 2018-11-27
LUCENE-8573: Use FutureArrays#mismatch in BKDWriter

cbuescher:
As [LUCENE-8573](https://issues.apache.org/jira/browse/LUCENE-8573) describes, BKDWriter uses loops in many places to find the first mismatching byte between multiple arrays. This change replaces that with FutureArrays#mismatch.


jpountz:
Thanks @cbuescher, the change looks good. I'll merge it provided tests pass.

uschindler:
Perfect. So it's even more performant with Java 9+.
Thanks, Christian

PR #513, opened 2018-12-04
LUCENE-8590: Optimize DocValues update datastructures

s1monw:
Today we are using a LinkedHashMap to buffer doc-values updates in
BufferedUpdates. This on the one hand uses an Object based datastructure
and on the other requires re-encoding the data into a more compact representation
once the BufferedUpdates are frozen. This change uses a more compact represenation
for the updates already in the BufferedUpdates in a parallel-array like datastructure
that can be reused in FrozenBufferedDeletes. It also adds an much simpler to use
API to consume the updates and allows for internal memory optimization for common
case updates.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/BufferedUpdates.java
+++ b/lucene/core/src/java/org/apache/lucene/index/BufferedUpdates.java
@@ -19,16 +19,15 @@ package org.apache.lucene.index;
 
 import java.util.ArrayList;
 import java.util.HashMap;
-import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.atomic.AtomicInteger;
-import java.util.concurrent.atomic.AtomicLong;
-import java.util.function.IntFunction;
 
 import org.apache.lucene.index.DocValuesUpdate.BinaryDocValuesUpdate;
 import org.apache.lucene.index.DocValuesUpdate.NumericDocValuesUpdate;
 import org.apache.lucene.search.Query;
+import org.apache.lucene.util.Accountable;
+import org.apache.lucene.util.Counter;
 import org.apache.lucene.util.RamUsageEstimator;
 
 /** Holds buffered deletes and updates, by docID, term or query for a
@@ -42,7 +41,7 @@ import org.apache.lucene.util.RamUsageEstimator;
 // instance on DocumentWriterPerThread, or via sync'd code by
 // DocumentsWriterDeleteQueue
 
-class BufferedUpdates {
+class BufferedUpdates implements Accountable {
 
   /* Rough logic: HashMap has an array[Entry] w/ varying
      load factor (say 2 * POINTER).  Entry is object w/ Term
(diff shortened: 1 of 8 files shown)


s1monw:
@mikemccand can you take a look?

lucene/core/src/java/org/apache/lucene/index/BufferedUpdates.java
```diff
@@ -288,15 +186,24 @@ void clear() {
...
     deleteDocIDs.clear();
-    numericUpdates.clear();
-    binaryUpdates.clear();
     numTermDeletes.set(0);
-    numNumericUpdates.set(0);
-    numBinaryUpdates.set(0);
-    bytesUsed.set(0);
+    numFieldUpdates.set(0);
+    fieldUpdates.clear();
+    bytesUsed.addAndGet(-bytesUsed.get());
```
shaie: Unrelated to this PR, but it feels like `Counter` could expose a `reset()` method and internally set the value to 0, instead of everyone doing this juggling.

lucene/core/src/java/org/apache/lucene/index/FieldUpdatesBuffer.java
```diff
@@ -0,0 +1,235 @@
...
+
+/**
+ * This class efficiently buffers numeric and binary field updates and stores
+ * terms, values and metadata in a memory efficient way without creating large amounts
+ * of objects. Update terms are stored without de-duplicating the update term.
+ * In general we try to optimize for several use-cases. For instance we try to use constant
+ * space for update terms field since the common case always updates on the same field. Also for docUpTo
+ * we try to optimize for the case when updates should be applied to all docs ie. docUpTo=Integer.MAX_VALUE.
+ * In other cases each update will likely have a different docUpTo.
+ * Along the same lines this impl optimizes the case when all updates have a value. Lastly, the soft_deletes case
```
shaie: The last sentence about "soft_deletes" reads too cumbersome to me. Can we just say "Lastly, if all updates share the same value for a field, the value is stored only once."? Or something along this.

lucene/core/src/java/org/apache/lucene/index/FieldUpdatesBuffer.java
```diff
@@ -0,0 +1,235 @@
...
+        bytesUsed.addAndGet((newSize-numericValues.length) * Long.BYTES);
+        numericValues = array;
+      }
+      numericValues[ord] = value;
+    }
+  }
+
+  void addNoValue(Term term, int docUpTo) {
+    termValues.append(term.bytes);
+    String field = term.field;
```
shaie: Maybe inline this, like you do in the other `addUpdate`?

lucene/core/src/java/org/apache/lucene/index/FieldUpdatesBuffer.java
```diff
@@ -0,0 +1,235 @@
...
+ * of objects. Update terms are stored without de-duplicating the update term.
+ * In general we try to optimize for several use-cases. For instance we try to use constant
+ * space for update terms field since the common case always updates on the same field. Also for docUpTo
+ * we try to optimize for the case when updates should be applied to all docs ie. docUpTo=Integer.MAX_VALUE.
+ * In other cases each update will likely have a different docUpTo.
+ * Along the same lines this impl optimizes the case when all updates have a value. Lastly, the soft_deletes case
+ * where all values for a specific field is shared this also stores numeric values only once if all updates share
+ * the same value.
+ */
+final class FieldUpdatesBuffer {
```
jpountz: This would be a bit easier to read for me if you introduced tiny abstractions around an (initially null) array + a default value, as all conditions make the code a bit hard to follow.

lucene/core/src/java/org/apache/lucene/index/BufferedUpdates.java
s1monw: agreed I will open a followup.

lucene/core/src/java/org/apache/lucene/index/FieldUpdatesBuffer.java
s1monw: I will look into this!

lucene/core/src/java/org/apache/lucene/index/FieldUpdatesBuffer.java
```diff
@@ -0,0 +1,235 @@
...
+  // we use a very simple approach and store the update term values without de-duplication
+  // which is also not a common case to keep updating the same value more than once...
+  // we might pay a higher price in terms of memory in certain cases but will gain
+  // on CPU for those. We also save on not needing to sort in order to apply the terms in order
+  // since by definition we store them in order.
+  private final BytesRefArray termValues;
+  private final BytesRefArray byteValues; // this will be null if we are buffering numerics
+  private int[] docsUpTo = null;
+  private long[] numericValues; // this will be null if we are buffering binaries
+  private boolean[] hasValues;
```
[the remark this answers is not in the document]
s1monw: yeah so I wasn't sure I can explore it

s1monw:
>  Any sense of how much of RAM reduction this is over current impl?
I ran a very simple measurement on top of `BufferedUpdates` adding 10k random updates with a constant seed. Here are the results:

Setup | master | patch | Percentage
------------ | ------------- | ------------ | -------------
RamBytesUsed Numeric/SameField/RandomValue/SameDocUpTo |  2783514 bytes | 549930 bytes | 19%
RamBytesUsed Numeric/SameField/RandomValue/RandomDocUpTo | 2783514 bytes | 593294 bytes | 21%
RamBytesUsed Numeric/SameField/SameValue | 2783514 bytes | 469546 bytes | 16%
RamBytesUsed Numeric/SameField/SameValue/RandomDocUpTo| 2783514 bytes | 512910 bytes | 18%

mikemccand:
Wow, that's a massive reduction!

shaie:
> Wow, that's a massive reduction!

Indeed! Well done @s1monw!

lucene/core/src/java/org/apache/lucene/index/BufferedUpdates.java
```diff
@@ -288,15 +184,24 @@ void clear() {
...
-    return deleteTerms.size() > 0 || deleteDocIDs.size() > 0 || deleteQueries.size() > 0 || numericUpdates.size() > 0 || binaryUpdates.size() > 0;
+    return deleteTerms.size() > 0 || deleteDocIDs.size() > 0 || deleteQueries.size() > 0 || numFieldUpdates.get() > 0;
+  }
+
+  @Override
+  public long ramBytesUsed() {
+    return bytesUsed.get() + fieldUpdatesBytesUsed.get();
+  }
+
+  public void clearDeletedDocIds() {
```
jpountz: no need for the public modifier?

lucene/core/src/java/org/apache/lucene/index/FieldUpdatesBuffer.java
```diff
@@ -0,0 +1,281 @@
...
+ * In general we try to optimize for several use-cases. For instance we try to use constant
+ * space for update terms field since the common case always updates on the same field. Also for docUpTo
+ * we try to optimize for the case when updates should be applied to all docs ie. docUpTo=Integer.MAX_VALUE.
+ * In other cases each update will likely have a different docUpTo.
+ * Along the same lines this impl optimizes the case when all updates have a value. Lastly, if all updates share the
+ * same value for a numeric field we only store the value once.
+ */
+final class FieldUpdatesBuffer {
+  private static final long SELF_SHALLOW_SIZE = RamUsageEstimator.shallowSizeOfInstance(FieldUpdatesBuffer.class);
+  private static final long STRING_SHALLOW_SIZE = RamUsageEstimator.shallowSizeOf(String.class);
```
jpountz: I think you meant `shallowSizeOfInstance` here too?

lucene/core/src/java/org/apache/lucene/index/FieldUpdatesBuffer.java
```diff
@@ -0,0 +1,281 @@
...
+    docsUpTo = new int[] {docUpTo};
+    if (initialValue.hasValue == false) {
+      hasValues = new FixedBitSet(1);
+      bytesUsed.addAndGet(hasValues.ramBytesUsed());
+    }
+    this.isNumeric = isNumeric;
+    byteValues = isNumeric ? null : new BytesRefArray(bytesUsed);
+  }
+
+  private long sizeOfString(String string) {
```
jpountz: static?

lucene/core/src/java/org/apache/lucene/index/FieldUpdatesBuffer.java
```diff
@@ -0,0 +1,281 @@
...
+      docsUpTo[ord] = docUpTo;
+    }
+
+    if (hasValue == false || hasValues != null) {
+      if (hasValues == null) {
+        hasValues = new FixedBitSet(ord+1);
+        hasValues.set(0, ord);
+        bytesUsed.addAndGet(hasValues.ramBytesUsed());
+      } else {
+        FixedBitSet fixedBitSet = FixedBitSet.ensureCapacity(hasValues, ord+1);
```
jpountz: You might want to do something like this instead. Otherwise this is going to reuse the long[] but create a new FixedBitSet wrapper for every new update.

```java
if (fixedBitSet.length() <= ord) {
  FixedBitSet fixedBitSet = FixedBitSet.ensureCapacity(hasValues, ArrayUtil.oversize(ord + 1, 1));
  bytesUsed.addAndGet(fixedBitSet.ramBytesUsed()-hasValues.ramBytesUsed());
  hasValues = fixedBitSet;
}
```

lucene/core/src/java/org/apache/lucene/index/FieldUpdatesBuffer.java
```diff
@@ -0,0 +1,281 @@
...
+        if (numericValues.length == 1) {
+          Arrays.fill(array, 1, ord, numericValues[0]);
+        }
+        bytesUsed.addAndGet((array.length-numericValues.length) * Long.BYTES);
+        numericValues = array;
+      }
+      numericValues[ord] = value;
+    }
+  }
+  void addNoValue(Term term, int docUpTo) {
```
jpountz: insert newline on top of this line

lucene/core/src/java/org/apache/lucene/index/FieldUpdatesBuffer.java
```diff
@@ -0,0 +1,281 @@
...
+        return bufferedUpdate;
+      } else {
+        return null;
+      }
+    }
+  }
+
+  private static int getArrayIndex(int arrayLength, int index) {
+    assert arrayLength == 1 || arrayLength > index : "illegal array index length: " + arrayLength + " index: " + index;
+    return Math.min(arrayLength-1, index);
```
jpountz: haha nice trick

lucene/core/src/java/org/apache/lucene/index/FieldUpdatesBuffer.java
```diff
@@ -0,0 +1,280 @@
...
+      docsUpTo[ord] = docUpTo;
+    }
+
+    if (hasValue == false || hasValues != null) {
+      if (hasValues == null) {
+        hasValues = new FixedBitSet(ord+1);
+        hasValues.set(0, ord);
+        bytesUsed.addAndGet(hasValues.ramBytesUsed());
+      } else {
+        FixedBitSet fixedBitSet = FixedBitSet.ensureCapacity(hasValues, ArrayUtil.oversize(ord + 1, 1));
```
jpountz: you need to put this under an `if (fixedBitSet.length() <= ord)`, otherwise we will likely resize most of the time?

s1monw: ++ good catch

lucene/core/src/java/org/apache/lucene/index/BufferedUpdates.java
```diff
@@ -288,15 +184,24 @@ void clear() {
...
+
+  @Override
+  public long ramBytesUsed() {
+    return bytesUsed.get() + fieldUpdatesBytesUsed.get();
+  }
+
+  void clearDeletedDocIds() {
+    deleteDocIDs.clear();
+    bytesUsed.addAndGet(-deleteDocIDs.size() * BufferedUpdates.BYTES_PER_DEL_DOCID);
   }
```
bringyou: The `byteUsed` will add nothing if we clear `deleteDocIDs` first, cuz its size is always zero.
Maybe a typo here?

s1monw: that's a bug indeed. It might not be a huge deal since we don't do much with the bytesUsed after this was called but it should still be fixed. Do you want to open a PR for this?

bringyou: I'm glad to do so😊

PR #515, opened 2018-12-05
LUCENE-8593: Specialize single value numeric DV updates

s1monw:
The case when all values are the the same on a numeric field update
is common for soft_deletes. With the new infrastucture for buffering
DV updates we can gain an easy win by specializing the applied updates
if all values are the same.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/DocValuesFieldUpdates.java
+++ b/lucene/core/src/java/org/apache/lucene/index/DocValuesFieldUpdates.java
@@ -425,4 +425,52 @@ abstract class DocValuesFieldUpdates implements Accountable {
       return hasValue;
     }
   }
+
+  static abstract class SingleValueDocValuesFieldUpdates extends DocValuesFieldUpdates {
+
+    protected SingleValueDocValuesFieldUpdates(int maxDoc, long delGen, String field, DocValuesType type) {
+      super(maxDoc, delGen, field, type);
+    }
+
+    @Override
+    void add(int doc, long value) {
+      assert longValue() == value;
+      super.add(doc);
+    }
+
+    @Override
+    void add(int doc, BytesRef value) {
+      assert binaryValue().equals(value);
+      super.add(doc);
+    }
+
+    @Override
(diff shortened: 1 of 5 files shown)


lucene/core/src/java/org/apache/lucene/index/DocValuesFieldUpdates.java
```diff
@@ -425,4 +425,51 @@ final boolean hasValue() {
...
+    @Override
+    void add(int doc, long value) {
+      super.add(doc);
+      assert longValue() == value;
+    }
+
+    @Override
+    void add(int doc, BytesRef value) {
+      super.add(doc);
+      assert binaryValue().equals(value);
```
dnhatn: I think this assert should be moved up.

lucene/core/src/java/org/apache/lucene/index/DocValuesFieldUpdates.java
```diff
@@ -425,4 +425,51 @@ final boolean hasValue() {
...
+  static abstract class SingleValueDocValuesFieldUpdates extends DocValuesFieldUpdates {
+
+    protected SingleValueDocValuesFieldUpdates(int maxDoc, long delGen, String field, DocValuesType type) {
+      super(maxDoc, delGen, field, type);
+    }
+
+    @Override
+    void add(int doc, long value) {
+      super.add(doc);
+      assert longValue() == value;
```
dnhatn: I think this assert should be moved up.

lucene/core/src/java/org/apache/lucene/index/DocValuesFieldUpdates.java
```diff
@@ -425,4 +425,51 @@ final boolean hasValue() {
...
+      assert binaryValue().equals(value);
+    }
+
+    @Override
+    void add(int docId, Iterator iterator) {
+      throw new UnsupportedOperationException();
+    }
+
+    protected abstract BytesRef binaryValue();
+    protected abstract long longValue();
```
dnhatn: nit: extra line.

PR #516, opened 2018-12-06
LUCENE-8594: DV update are broken for updates on new field

s1monw:
A segmemnt written with Lucene70Codec failes if it ties to update
a DV field that didn't exist in the index before it was upgraded to
Lucene80Codec. We bake the DV format into the FieldInfo when it's used
the first time and therefor never go to the codec if we need to update.
yet on a field that didn't exist before and was added during an indexing
operation we have to consult the coded and get an exception.
This change fixes this issue and adds the relevant bwc tests.

### DIFF
--- a/lucene/backward-codecs/src/java/org/apache/lucene/codecs/lucene70/Lucene70Codec.java
+++ b/lucene/backward-codecs/src/java/org/apache/lucene/codecs/lucene70/Lucene70Codec.java
@@ -55,7 +55,8 @@ public class Lucene70Codec extends Codec {
   private final SegmentInfoFormat segmentInfosFormat = new Lucene70SegmentInfoFormat();
   private final LiveDocsFormat liveDocsFormat = new Lucene50LiveDocsFormat();
   private final CompoundFormat compoundFormat = new Lucene50CompoundFormat();
-  
+  private final DocValuesFormat defaultDVFormat = DocValuesFormat.forName("Lucene70");
+
   private final PostingsFormat postingsFormat = new PerFieldPostingsFormat() {
     @Override
     public PostingsFormat getPostingsFormatForField(String field) {
@@ -66,7 +67,7 @@ public class Lucene70Codec extends Codec {
   private final DocValuesFormat docValuesFormat = new PerFieldDocValuesFormat() {
     @Override
     public DocValuesFormat getDocValuesFormatForField(String field) {
-      throw new IllegalStateException("This codec should only be used for reading, not writing");
+      return defaultDVFormat;
     }
   };
   
--- a/lucene/backward-codecs/src/test/org/apache/lucene/index/TestBackwardsCompatibility.java
+++ b/lucene/backward-codecs/src/test/org/apache/lucene/index/TestBackwardsCompatibility.java
@@ -1538,7 +1538,7 @@ public class TestBackwardsCompatibility extends LuceneTestCase {
       assertEquals(ndvcf.longValue(), ndvf.longValue()*2);
     }
   }
-  
+
   private void assertBinaryDocValues(LeafReader r, String f, String cf) throws IOException {
     BinaryDocValues bdvf = r.getBinaryDocValues(f);
     BinaryDocValues bdvcf = r.getBinaryDocValues(cf);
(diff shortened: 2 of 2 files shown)


s1monw:
this was found in https://github.com/elastic/elasticsearch/pull/36286

lucene/backward-codecs/src/test/org/apache/lucene/index/TestBackwardsCompatibility.java
```diff
@@ -1587,6 +1590,76 @@ public void testDocValuesUpdates() throws Exception {
...
+            BinaryDocValues binaryDocValues = leafReader.getBinaryDocValues("new_binary");
+            int doc;
+            while ((doc = postings.nextDoc()) != DocIdSetIterator.NO_MORE_DOCS) {
+              found = true;
+              assertEquals(doc, binaryDocValues.nextDoc());
+              assertEquals(doc, numericDocValues.nextDoc());
+              assertEquals(1, numericDocValues.longValue());
+              assertEquals(toBytes(1), binaryDocValues.binaryValue());
+            }
+          }
```
jpountz: for completeness, add `assertEquals(DocIdSetIterator.NO_MORE_DOCS, binaryDocValues.nextDoc()); assertEquals(DocIdSetIterator.NO_MORE_DOCS, numericDocValues.nextDoc());`?

lucene/backward-codecs/src/test/org/apache/lucene/index/TestBackwardsCompatibility.java
```diff
@@ -1587,6 +1590,76 @@ public void testDocValuesUpdates() throws Exception {
...
+    IndexWriter writer = new IndexWriter(dir, conf);
+    // introduce a new field that we later update
+    writer.addDocument(Arrays.asList(new StringField("id", "" + Integer.MAX_VALUE, Field.Store.NO),
+        new NumericDocValuesField("new_numeric", 1),
+        new BinaryDocValuesField("new_binary", toBytes(1))));
+    writer.updateNumericDocValue(new Term("id", "1"), "new_numeric", 1);
+    writer.updateBinaryDocValue(new Term("id", "1"), "new_binary", toBytes(1));
+
+    writer.commit();
+    Callable assertDV = () -> {
```
jpountz: s/Callable/Runnable/ since you don't care about the return value/

s1monw: it's throws an exception hence the callable

lucene/backward-codecs/src/test/org/apache/lucene/index/TestBackwardsCompatibility.java
s1monw: 👍

s1monw: well this doesn't work since we have an other doc with a value in this field I guess and it might be merged into a single segment. which we also force

s1monw:
@jpountz I pushed changes

lucene/backward-codecs/src/java/org/apache/lucene/codecs/lucene70/Lucene70Codec.java
```diff
@@ -66,7 +67,7 @@ public PostingsFormat getPostingsFormatForField(String field) {
   private final DocValuesFormat docValuesFormat = new PerFieldDocValuesFormat() {
     @Override
     public DocValuesFormat getDocValuesFormatForField(String field) {
-      throw new IllegalStateException("This codec should only be used for reading, not writing");
+      return defaultDVFormat;
```
mikemccand: Ahh this is because the DV updates must also be written with the old codec?

jpountz: I don't think it's a requirement, but I like using the old format better:
 - it's probably easier to debug, as all fields would be using the same format
 - given that PFDVFormat creates different files per format, forcing the new format would increase the number of files for old segments

PR #521, opened 2018-12-09
LUCENE-8598: Improve field updates packed values

s1monw:
DocValuesFieldUpdats are using compact settings for packet ints that causes
dramatic slowdowns when the updates are finished and sorted. Moving to the default
accepted overhead ratio yields up to 4x improvements in applying updates. This change
also improves the packing of numeric values since we know the value range in advance and
can choose a different packing scheme in such a case.
Overall this change yields a good performance improvement since 99% of the times of applying
DV field updates are spend in the sort method which essentially makes applying the updates
4x faster.

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -266,7 +266,10 @@ Optimizations
 
 * LUCENE-8590: BufferedUpdates now uses an optimized storage for buffering docvalues updates that
   can safe up to 80% of the heap used compared to the previous implementation and uses non-object
-  based datastructures. (Simon Willnauer, Mike McCandless, Shai Erera, Adrien Grant)
+  based datastructures. (Simon Willnauer, Mike McCandless, Shai Erera, Adrien Grand)
+
+* LUCENE-8598: Moved to the default accepted overhead ratio for packet ints in DocValuesFieldUpdats
+  yields an up-to 4x performance improvement when applying doc values updates. (Simon Willnauer, Adrien Grand)
 
 Other
 
--- a/lucene/core/src/java/org/apache/lucene/index/NumericDocValuesFieldUpdates.java
+++ b/lucene/core/src/java/org/apache/lucene/index/NumericDocValuesFieldUpdates.java
@@ -19,6 +19,7 @@ package org.apache.lucene.index;
 import org.apache.lucene.document.NumericDocValuesField;
 import org.apache.lucene.util.BytesRef;
 import org.apache.lucene.util.RamUsageEstimator;
+import org.apache.lucene.util.packed.AbstractPagedMutable;
 import org.apache.lucene.util.packed.PackedInts;
 import org.apache.lucene.util.packed.PagedGrowableWriter;
 import org.apache.lucene.util.packed.PagedMutable;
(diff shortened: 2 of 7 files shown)


lucene/core/src/test/org/apache/lucene/index/Benchmark.java
```diff
@@ -0,0 +1,94 @@
...
+import org.apache.lucene.document.Field;
+import org.apache.lucene.document.NumericDocValuesField;
+import org.apache.lucene.document.StringField;
+import org.apache.lucene.store.Directory;
+import org.apache.lucene.store.FSDirectory;
+import org.apache.lucene.util.BytesRef;
+import org.apache.lucene.util.NullInfoStream;
+import org.apache.lucene.util.TestUtil;
+
+public class Benchmark {
```
s1monw: NOTE: I included this benchmark in the initial version to show what I benchmarked

lucene/core/src/java/org/apache/lucene/index/NumericDocValuesFieldUpdates.java
```diff
@@ -53,14 +55,28 @@ BytesRef binaryValue() {
...
+   this(delGen, field, Long.MIN_VALUE, Long.MAX_VALUE, maxDoc);
+  }
 
-  public NumericDocValuesFieldUpdates(long delGen, String field, int maxDoc) {
+  NumericDocValuesFieldUpdates(long delGen, String field, long minValue, long maxValue, int maxDoc) {
     super(maxDoc, delGen, field, DocValuesType.NUMERIC);
-    values = new PagedGrowableWriter(1, PAGE_SIZE, 1, PackedInts.FAST);
+    assert minValue <= maxValue : "minValue must be <= maxValue [" + minValue + " > " + maxValue + "]";
+    long max = maxValue - minValue;
+    if (max < 0) { // pretty large range - fall back to growable writer
```
jpountz: It feels a bit weird to optimize when using 64 bits per value but not 63, or 62? Not sure if you are aware but `PackedInts.unsignedBitsRequired` can help avoid special-casing when max-min overflows.

lucene/core/src/java/org/apache/lucene/index/NumericDocValuesFieldUpdates.java
```diff
@@ -53,14 +55,28 @@ BytesRef binaryValue() {
...
+      values = new PagedGrowableWriter(1, PAGE_SIZE, 1, PackedInts.DEFAULT);
+      this.minValue = 0;
+    } else {
+      int bitsPerValue = PackedInts.bitsRequired(max);
+      values = new PagedMutable(1, PAGE_SIZE, bitsPerValue, PackedInts.DEFAULT);
```
jpountz: since we are sorting in the end, all pages will likely have the same number of bits per value in the end anyway, should we move to PackedInts.getMutable instead?

s1monw: not sure we can since we don't know how many values we actually have?

s1monw:
> t's interesting that you found out that most of the time is spent with packed ints when sorting, likely because of swaps: the sorting impl that is being used (InPlaceMergeSorter) is the impl that performs the higher number of swaps due to the constraint to run in-place. Maybe we should look into spending a bit more memory on sorting to avoid having so many swaps, eg. with TimSorter (see ArrayTimSorter for an example).

regarding TimSorter, it looks pretty complex and I wonder how well it would work with the subclasses and what needs to be implemented. Maybe we can explore this in a separate change?

lucene/CHANGES.txt
```diff
@@ -268,6 +268,9 @@ Optimizations
   can safe up to 80% of the heap used compared to the previous implementation and uses non-object
   based datastructures. (Simon Willnauer, Mike McCandless, Shai Erera, Adrien Grant)
 
+* LUCENE-8598: Moved to the default accepted overhead ratio for packet ints in DocValuesFieldUpdats
+  yields an up-to 4x performance improvement when applying doc values updates. (Simon Willnauer, Adrien Grant)
```
jpountz: typo in my name

PR #522, opened 2018-12-10
LUCENE-8599: Use sparse bitset to store docs in SingleValueDocValuesFieldUpdates

s1monw:
Using a sparse bitset in SingleValueDocValuesFieldUdpates allows storing
which documents have an update much more efficient and prevents the need
to sort the docs array altogether that showed to be a significant bottleneck
in LUCENE-8598. Using the spares bitset yields another 10x performance improvement
in applying updates versus the changes proposed in LUCENE-8598.


jpountz:
For the record, DocIdSetBuilder needs to sort as well if the number of collected docs is low, but it does it with a pretty fast radix sort.

s1monw:
@jpountz  I will get this in as is and then explore what we can do as a followup with DocIdSetBuilder. We do need additional stats to make efficient use of it IMO and this already yields a significant improvement.

jpountz:
Sure!

PR #525, opened 2018-12-12
LUCENE-8585: Index-time jump-tables for DocValues

tokee:
First complete attempt of LUCENE-8585, with all jump-tables at index-time and with the not-used-anymore lucene70 codec classes moved to backward-codecs.


lucene/core/src/java/org/apache/lucene/codecs/lucene80/IndexedDISI.java
```diff
@@ -0,0 +1,536 @@
...
+    // If the distance between the current position and the target is < 8 longs
+    // there is no sense in using rank
+    if (targetWordIndex - disi.wordIndex < RANK_BLOCK_LONGS) {
+      return;
+    }
+
+    // Resolve the rank as close to targetInBlock as possible (maximum distance is 8 longs)
+    // Note: rankOrigoOffset is tracked on block open, so it is absolute (e.g. don't add origo)
+    final int rankIndex = targetInBlock >> RANK_BLOCK_BITS; // 8 longs: 2^3 * 2^6 = 512
+    disi.slice.seek(disi.rankOrigoOffset + rankIndex*Short.BYTES);
```
jpountz: Seeking back and forth between ranks and the bitset could be slow with NIOFSDirectory for the same reason as the jump table. Maybe we should put the rank index at the beginning of blocks and load it up-front in an in-memory short[]?

lucene/core/src/java/org/apache/lucene/codecs/lucene80/Lucene80DocValuesProducer.java
```diff
@@ -0,0 +1,1413 @@
...
+      if (this.block != block) {
+        int bitsPerValue;
+        do {
+          // If the needed block is the one directly following the current block, it is cheaper to avoid the cache
+          if (entry.jumpTableOffset != -1 && block != this.block+1) {
+//            System.out.println("Jumptabling to index=" + index +
+//                ", current(offset=" + data.getFilePointer() + ", length=" + data.length() +
+//                "), jumpTable(offset=" + entry.jumpTableOffset + ", jumpTableOffset=" + block*Long.BYTES + ", absolute=" +
+//                (entry.jumpTableOffset+block*Long.BYTES) + ")");
+            blockEndOffset = slice.readLong(entry.jumpTableOffset+block*Long.BYTES-entry.valuesOffset)-entry.valuesOffset;
```
jpountz: This one is reading alternatively from the jump table and from actual data too. Should we have two slices?

lucene/core/src/java/org/apache/lucene/codecs/lucene80/IndexedDISI.java
```diff
@@ -0,0 +1,536 @@
...
+    }
+    if (blockCardinality > 0) {
+      jumps = addJumps(jumps, out.getFilePointer()-origo, totalCardinality, jumpBlockIndex, prevBlock+1);
+      flush(prevBlock, buffer, blockCardinality, out);
+      buffer.clear(0, buffer.length());
+      prevBlock++;
+    }
+    final int lastBlock = prevBlock == -1 ? 0 : prevBlock;
+//    jumps = addJumps(jumps, out.getFilePointer(), totalCardinality, lastBlock, lastBlock+1);
+    // NO_MORE_DOCS is stored explicitly
```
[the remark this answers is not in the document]
tokee: Currently (as you note further down this might be changed), the jump-tables are not always used. So we still need `NO_MORE_DOCS` for small-stepping through the bits?

lucene/core/src/java/org/apache/lucene/codecs/lucene80/IndexedDISI.java
```diff
@@ -0,0 +1,536 @@
...
+      jumps[b] = jump;
+    }
+    return jumps;
+  }
+
+  // Flushes the offet & index jump-table for blocks. This should be the last data written to out
+  private static void flushBlockJumps(long[] jumps, int blockCount, IndexOutput out, long origo) throws IOException {
+    if (blockCount == 1) { // A single jump is just wasted space so we ignore that
+      blockCount = 0;
+    }
```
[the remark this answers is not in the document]
tokee: The `jumpTableOffset` that is set on construction is `final`, so the branch `if (jumpTableOffset != -1L && blockIndex >= (block >> 16)+2 && blockIndex < jumpTableEntryCount) {` should be JITted away if the jump-table is disabled (by having offset `-1`). It was done explicitly to spare the indirection of a jump-table lookup when it was more costly than the base operation.

lucene/core/src/java/org/apache/lucene/codecs/lucene80/IndexedDISI.java
```diff
@@ -0,0 +1,536 @@
...
+    for (int i = 0 ; i < blockCount ; i++) {
+      out.writeLong(jumps[i]);
+    }
+    // The number of blocks are written as the last data in the IndexedDISI structure.
+    // This makes it possible to infer the jumpTableOffset: lastPos - #blocks * Long.BYTES
+    // As there are at most 32k blocks, the count is stored as a short
+    out.writeShort((short) blockCount);
```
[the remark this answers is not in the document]
tokee: Whoops. Of course it should. I'll change that.

lucene/core/src/java/org/apache/lucene/codecs/lucene80/IndexedDISI.java
```diff
@@ -0,0 +1,536 @@
...
+  }
+
+  /**
+   * This constructor allows to pass the slice directly in case it helps reuse.
+   * see eg. Lucene80 norms producer's merge instance.
+   * @param slice backing data.
+   * @param cost normally the number of logical docIDs.
+   */
+  IndexedDISI(IndexInput slice, long cost) throws IOException {
+    this.slice = slice;
```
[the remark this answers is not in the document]
tokee: Very good point. I'll implement it as you suggest.

lucene/core/src/java/org/apache/lucene/codecs/lucene80/IndexedDISI.java
```diff
@@ -0,0 +1,536 @@
...
+    }
+    boolean found = block == targetBlock && method.advanceExactWithinBlock(this, target);
+    this.doc = target;
+    return found;
+  }
+
+  private void advanceBlock(int targetBlock) throws IOException {
+    final int blockIndex = targetBlock >> 16;
+    // If the destination block is 2 blocks or more ahead, we use the jump-table.
+    if (jumpTableOffset != -1L && blockIndex >= (block >> 16)+2 && blockIndex < jumpTableEntryCount) {
```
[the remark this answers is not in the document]
tokee: I don't know. It is hard to measure such small things when the index size is large. The check for `jumpTableOffset` should be JITted away and as the address for the next block is already fetched as part of reading the block headers, I estimated that skipping the lookup in the jump-table would be faster.

lucene/core/src/java/org/apache/lucene/codecs/lucene80/IndexedDISI.java
tokee: This is a hard one and I am not sure we can find a single solution that works best for all scenarios. And there are so many possibilities:

1. Allocate a slice for DENSE-rank: We already use many slices.
1. Make DENSE-rank and block-jump share the same slice. This thrashes the block-jump cache when a DENSE-rank lookup is needed, but not the other way as the need for skipping to a new block means that the DENSE block has been exhausted.
1. Don't have a special slice for DENSE-rank. A DENSE block is about 2 * 8192 bytes, so jumps within the first half would not mean thrashing.
1. Load all 128 shorts up front: Wasted if skips are < 512 docIDs or there is only a single (or possibly very few) lookups in the DENSE block.
1. Increase `RANK_BLOCK_SIZE`: A fair suggestion as the current 1 rank/8 longs is somewhat arbitrary. The concrete choice comes down to how large the average advance is and we have little knowledge there (at least I don't). It is still "only" a mitigation of the potentially-wasted-overhead problem.
1. Use heuristics: Load the 128 shorts on first advance within the DENSE block, if the advance is > `RANK_BLOCK_SIZE` and < X, where we choose X so that we guess there will only be a single advance (maybe 32768?). Makes the logic more complicated and branchy.
1. Interleave the ranking with the bits. This would make the math more complicated and probably require division and/or hurt single-docID-stepping. I'll think a bit about this and see if there is a way to do it cleanly.

jpountz: > Load all 128 shorts up front: Wasted if skips are < 512 docIDs or there is only a single (or possibly very few) lookups in the DENSE block.

This would be wasted indeed, but I'd expect reading in a byte[256] to be super cheap?

jpountz: A 8th option could be to drop the rank index entirely and decrease the size of blocks so that within-block skipping wouldn't be necessary anymore.

tokee: Sorry, yes, I should have distinguished `short` read one-at-a-time from `byte[256]`. If that is indeed super cheap then I am leaning towards #4 (with byte[256]-read) or #6. With #4 preferred for simplicity.

Decreasing the size of blocks is a valid possibility and it is cheaper than before as they do not need to be iterated for long advances. I am not too positive about this idea as it changes the dynamic for `SPARSE`, where there will be wasted bits in the docIDs, and as it increases the block jump-table length. Also, what should the block-size be to get to a "reasonable" `DENSE`? 1024 bits? 2048? I am afraid that going so low would make the relative overhead from switching blocks noticeable higher.

So, should I try and implement #4? It's quite simple to do.

jpountz: +1 to try and implement 4

tokee: As stated on the JIRA the pre-loading of rank is now implemented. It had the secondary effect of making the code cleaner. Should we need to change to `#6` heuristics, it can be done without too much trouble.

lucene/test-framework/src/java/org/apache/lucene/index/BaseDocValuesFormatTestCase.java
```diff
@@ -1204,6 +1204,12 @@ public void testRandomSortedBytes() throws IOException {
   }
 
   private void doTestNumericsVsStoredFields(double density, LongSupplier longs) throws Exception {
+    doTestNumericsVsStoredFields(density, longs, 256);
+    // TODO: 200000 docs are needed to test jumps properly (see LUCENE-8585), but that is quite slow (few minutes).
+    // Maybe it can be nightly?
+    //doTestNumericsVsStoredFields(density, longs, 200000);
```
tokee: Should we add a `if (TEST_NIGHTLY) ...`-switch here to to enable the 200K test case? It would increase coverage for jumps a great deal.

lucene/core/src/java/org/apache/lucene/codecs/lucene80/IndexedDISI.java
```diff
@@ -0,0 +1,542 @@
...
+    }
+    boolean found = block == targetBlock && method.advanceExactWithinBlock(this, target);
+    this.doc = target;
+    return found;
+  }
+
+  private void advanceBlock(int targetBlock) throws IOException {
+    final int blockIndex = targetBlock >> 16;
+    // If the destination block is 2 blocks or more ahead, we use the jump-table.
+    if (jumpTable != null && blockIndex >= (block >> 16)+2 && blockIndex < jumpTableEntryCount) {
```
jpountz: This seems to mean that if you call `advance(target)` on a target that is greater than the higher docID of the last block, then you will resort to a linear scan. Can we fix it?

As a side note, `jumpTable != null` seems to be covered by `blockIndex < jumpTableEntryCount` given that jumpTable is only null if jumpTableEntryCount is zero?

lucene/core/src/java/org/apache/lucene/codecs/lucene80/IndexedDISI.java
```diff
@@ -0,0 +1,542 @@
...
+   * If the distance between the current position and the target is > 8 words, the rank cache will
+   * be used to guarantee a worst-case of 1 rank-lookup and 7 word-read-and-count-bits operations.
+   * Note: This does not guarantee a skip up to target, only up to nearest rank boundary. It is the
+   * responsibility of the caller to iterate further to reach target.
+   * @param disi standard DISI.
+   * @param target the wanted docID for which to calculate set-flag and index.
+   * @throws IOException if a DISI seek failed.
+   */
+  private static void rankSkip(IndexedDISI disi, int target) throws IOException {
+    final int targetInBlock = target & 0xFFFF;       // Lower 16 bits
```
jpountz: should we compute it after the below if statement, it's not used otherwise?

lucene/core/src/java/org/apache/lucene/codecs/lucene80/IndexedDISI.java
```diff
@@ -0,0 +1,542 @@
...
+    // Resolve the rank as close to targetInBlock as possible (maximum distance is 8 longs)
+    // Note: rankOrigoOffset is tracked on block open, so it is absolute (e.g. don't add origo)
+    final int rankIndex = targetInBlock >> RANK_BLOCK_BITS; // 8 longs: 2^3 * 2^6 = 512
+
+    final int rank =
+        (disi.rankTable[rankIndex<<1] & 0xFF) << 8 |
+        (disi.rankTable[(rankIndex<<1)+1] & 0xFF);
+
+    // Position the counting logic just after the rank point
+    final int rankAlignedWordIndex = rankIndex << RANK_BLOCK_BITS >> 6;
```
jpountz: `<< (RANK_BLOCK_BITS - 6)` so that the compiler can optimize it? It doesn't seem like you need to clean up the higher bits?

lucene/core/src/java/org/apache/lucene/codecs/lucene80/Lucene80DocValuesConsumer.java
```diff
@@ -0,0 +1,663 @@
...
+          writer.add(encode.get(v));
+        }
+      }
+    }
+    writer.finish();
+  }
+
+  // Returns the offset to the jump-table for vBPV
+  private long writeValuesMultipleBlocks(SortedNumericDocValues values, long gcd) throws IOException {
+    long[] offsets = new long[ArrayUtil.oversize(100, Long.BYTES)]; // 100 blocks = 1.6M values
```
jpountz: should we start with ArrayUtil.oversize(1, Long.BYTES) to be more likely to exercise the growth logic in tests?

lucene/core/src/java/org/apache/lucene/codecs/lucene80/Lucene80DocValuesProducer.java
```diff
@@ -0,0 +1,1430 @@
...
+    final long mul;
+    final int mask;
+
+    long block = -1;
+    long delta;
+    long offset;
+    long blockEndOffset;
+    LongValues values;
+
+    VaryingBPVReader(NumericEntry entry) throws IOException {
```
jpountz: this looks unused

lucene/core/src/java/org/apache/lucene/codecs/lucene80/Lucene80DocValuesProducer.java
```diff
@@ -0,0 +1,1430 @@
...
+      this(entry,
+          data.randomAccessSlice(entry.valuesOffset, entry.valuesLength),
+          data.randomAccessSlice(entry.valuesOffset, entry.valuesLength));
```
jpountz: let's make them not overlap?

lucene/core/src/java/org/apache/lucene/codecs/lucene80/Lucene80DocValuesProducer.java
```diff
@@ -0,0 +1,1430 @@
...
+    }
+
+    long getLongValue(long index) throws IOException {
+      final long block = index >>> shift;
+      if (this.block != block) {
+        int bitsPerValue;
+        do {
+          // If the needed block is the one directly following the current block, it is cheaper to avoid the cache
+          if (entry.valueJumpTableOffset != -1 && block != this.block+1) {
+            blockEndOffset = rankSlice.readLong(entry.valueJumpTableOffset +block*Long.BYTES-entry.valuesOffset)-
```
jpountz: can we start the slice at `entry.valueJumpTableOffset` to avoid adding it every time (related to the above comment)

lucene/core/src/java/org/apache/lucene/codecs/lucene80/IndexedDISI.java
```diff
@@ -0,0 +1,542 @@
...
+    this.cost = cost;
+
+    this.jumpTableEntryCount = jumpTableEntryCount;
+    //slice.seek(slice.length()-Short.BYTES);
+    if (jumpTableEntryCount <= 0) {
+      jumpTable = null;
+    } else {
+      long jumpTableOffset = slice.length()-jumpTableEntryCount*Long.BYTES;
+      jumpTable = slice.randomAccessSlice(jumpTableOffset, slice.length()-jumpTableOffset);
+    }
```
jpountz: you will need to support reuse for jumpTable as well. Doc values are special because they get pulled when writing postings in order to index impacts, and we don't want to create a new slice for every indexed term. I believe the test that tests slice creation only doesn't fail because we don't create jump tables if there aren't many docs.

lucene/core/src/test/org/apache/lucene/index/TestDocValues.java
```diff
@@ -123,8 +124,72 @@ public void testNumericField() throws Exception {
...
-  
-  /** 
+
+  // The LUCENE-8585 jump-tables enables O(1) skipping of IndexedDISI blocks, DENSE block lookup
+  // and numeric multi blocks. This test focuses on testing these jumps.
+  @Slow
+  public void testNumericFieldJumpTables() throws Exception {
+    Directory dir = newDirectory();
+    IndexWriter iw = new IndexWriter(dir, newIndexWriterConfig(null));
+    final int maxDoc = atLeast(3*65536); // Must be above 3*65536 to trigger IndexedDISI block skips
```
jpountz: Why do we have this test here as opposed to TestLucene80DocValuesFormat?

Given that you are indexing so many documents and don't care about testing indexing as much as you care about checking iterators, you should make sure that the index writer config isn't too crazy and enforce sane values of maxBufferedDocs, RAMBufferSizeMB, and a sane merge policy (eg. not the alcoholic one). See for instance TestLucene70DocValuesFormat#doTestSortedNumericBlocksOfVariousBitsPerValue.

lucene/test-framework/src/java/org/apache/lucene/index/BaseDocValuesFormatTestCase.java
jpountz: Let's not do that in the base test case as this will be too slow with some codecs such as the SimpleTextCodec. What we have done before for formats that have branches that only get triggered with many documens is that we added tests to their own test file, ie. TestLucene80DocValuesFormat.

jpountz: By the way if we want to test such large numbers of documents, we should avoid RandomIndexWriter (use IndexWriter instead) and set reasonable values of maxBufferedDocs+RAMSizeBufferMB+mergePolicy instead of relying on random values that are set by LuceneTestCase#newIndexWriterConfig and make indexing slow (but are useful to increase coverage of flushing+merging)

jpountz:
I forgot to mention: precommit complains about unused imports.

lucene/core/src/java/org/apache/lucene/codecs/lucene80/IndexedDISI.java
```diff
@@ -0,0 +1,542 @@
...
+ *
+ * Each long entry consists of 2 logical parts:
+ *
+ * The first 31 bits hold the index (number of set bits in the blocks) up to just before the
+ * wanted block. The next 33 bits holds the offset in bytes into the underlying slice.
+ * As there is a maximum of 2^16 blocks, it follows that the maximum size of any block must
+ * not exceed 2^17 bits to avoid overflow. This is currently the case, with the largest
+ * block being DENSE and using 2^16 + 288 bits, and is likely to continue to hold as using
+ * more than double the amount of bits is unlikely to be an efficient representation.
+ * The cache overhead is numDocs/1024 bytes.
```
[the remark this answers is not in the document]
tokee: Nice catch. That was me mixing bits & bytes and arriving at too harsh a requirement, making the representation needlessly complicated: All block types are < 2^17 *bits*, but the offset is in *bytes*, so we do not require 2^16 * 2^17 bits to hold it; only 2^16 * 2^17 / 2^3 = 30 bits.

With the offset-requirement lowered from 33 to 30 bits, it is much more natural to represent offset & index as two ints. This will be in the next commit.

lucene/core/src/java/org/apache/lucene/codecs/lucene80/IndexedDISI.java
```diff
@@ -0,0 +1,542 @@
...
+    return flushBlockJumps(jumps, lastBlock, out, origo);
+  }
+
+  // Adds entries to the offset & index jump-table for blocks
+  private static long[] addJumps(long[] jumps, long offset, long index, int startBlock, int endBlock) {
+    jumps = ArrayUtil.grow(jumps, endBlock +1);
+    final long jump = (index << BLOCK_INDEX_SHIFT) | offset;
+    for (int b = startBlock; b < endBlock; b++) {
+      jumps[b] = jump;
+    }
```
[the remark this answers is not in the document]
tokee: Yes, except that the code was changed to use pars of ints, to split index and offset.

lucene/core/src/java/org/apache/lucene/codecs/lucene80/IndexedDISI.java
tokee: It worked before because jumpTables are relative to the length of the slice, not the current position. Anyway, now I understand what is happening. Re-use of jumpTables is in the next commit.

lucene/core/src/java/org/apache/lucene/codecs/lucene80/IndexedDISI.java
```diff
@@ -0,0 +1,542 @@
...
+      buffer.set(doc & 0xFFFF);
+      blockCardinality++;
+      prevBlock = block;
+    }
+    if (blockCardinality > 0) {
+      jumps = addJumps(jumps, out.getFilePointer()-origo, totalCardinality, jumpBlockIndex, prevBlock+1);
+      flush(prevBlock, buffer, blockCardinality, out);
+      buffer.clear(0, buffer.length());
+      prevBlock++;
+    }
```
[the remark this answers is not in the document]
tokee: The `lastBlock`handles that so that the jumpTable will have its length shortened. But it does have the unintended consequence that advancing into trailing empty blocks will not use jumpTables.

The check is `blockIndex < jumpTableEntryCount` in `IndexedDISI.advanceBlock` and I _think_ it can be optimized by seeking to the last jumpTable entry. I'll have to look at that later though.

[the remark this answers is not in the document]
tokee: Ah... The `writeBitset`-method takes an `DocIDSetIterator`, which has no concept of the range of documents, only the ones which are present. It is not possible to make jump-table entries for trailing blank blocks as there are none at write time.

I have changed `IndexedDISI.advanceBlock` to handle jumps outside of the known range by jumping to the last known entry, which will always be the `NO_MORE_DOCS`-entry.

lucene/core/src/java/org/apache/lucene/codecs/lucene80/IndexedDISI.java
```diff
@@ -0,0 +1,601 @@
...
+
+  /**
+   * @param in backing data.
+   * @param offset starting offset for blocks in backing data.
+   * @param length the number of bytes in the backing data.
+   * @param jumpTableEntryCount the number of blocks convered by the jump-table.
+   * @param cost normally the number of logical docIDs.
+   */
+  IndexedDISI(IndexInput in, long offset, long length, int jumpTableEntryCount, byte denseRankPower, long cost) throws IOException {
+    this(in.slice("docs", offset, length),
```
jpountz: it _seems_ to include the jump table in the end, can we make this slice only cover blocks, but not the jump table?

lucene/core/src/java/org/apache/lucene/codecs/lucene80/IndexedDISI.java
```diff
@@ -0,0 +1,626 @@
...
+   *                       This should be stored in meta and used when creating an instance of IndexedDISI.
+   * @throws IOException if there was an error writing to out.
+   * @return the number of jump-table entries following the blocks, -1 for no entries.
+   *         This should be stored in meta and used when creating an instance of IndexedDISI.
+   */
+  static short writeBitSet(DocIdSetIterator it, IndexOutput out, byte denseRankPower) throws IOException {
+    final long origo = out.getFilePointer(); // All jumps are relative to the origo
+    if (denseRankPower < 7 || denseRankPower > 15) {
+      denseRankPower = -1; // Disabled
+    }
```
jpountz: let's not be lenient and fail instead?

lucene/core/src/java/org/apache/lucene/codecs/lucene80/IndexedDISI.java
```diff
@@ -0,0 +1,626 @@
...
+   *                            This must match the number returned by {@link #writeBitSet(DocIdSetIterator, IndexOutput, byte)}.
+   * @param denseRankPower the number of docIDs covered by each rank entry in DENSE blocks, expressed as {@code 2^denseRankPower}.
+   *                       This must match the power given in {@link #writeBitSet(DocIdSetIterator, IndexOutput, byte)}
+   * @param cost normally the number of logical docIDs.
+   */
+  IndexedDISI(IndexInput blockSlice, RandomAccessInput jumpTable, int jumpTableEntryCount, byte denseRankPower, long cost) throws IOException {
+    this.slice = blockSlice;
+    this.jumpTable = jumpTable;
+    this.jumpTableEntryCount = jumpTableEntryCount;
+    this.denseRankPower = denseRankPower < 7 || denseRankPower > 15 ? -1 : denseRankPower;
```
jpountz: same here, should we fail values that make little sense instead?

lucene/core/src/java/org/apache/lucene/codecs/lucene80/Lucene80DocValuesConsumer.java
jpountz: can you reduce the initial size so that testing exercises this line of code?

lucene/test-framework/src/java/org/apache/lucene/index/BaseDocValuesFormatTestCase.java
```diff
@@ -1216,7 +1219,7 @@ private void doTestNumericsVsStoredFields(double density, LongSupplier longs) th
     doc.add(dvField);
     
     // index some docs
-    int numDocs = atLeast(300);
+    int numDocs = atLeast((int) (minDocs*1.172));
```
jpountz: why this 1.172 factor?

tokee: Yeah, that should have been explained. It is because I made the minDocs variable. Originally the test was fixed with 256 as minDocs. `256*1.172=300`, so with this trick the `numDocs/minDocs`-factor stayed constant for the old tests. I have probably been overthinking this. Should I set it to something less puzzling like 1.5 or 2?

tokee:
Closing pull request as this has been committed to master.

PR #526, opened 2018-12-12
LUCENE-8608: Extract utility class to iterate over terms docs

s1monw:
Today we re-implement the same algorithm in various places
when we want to consume all docs for a set/list of terms. This
caused serious slowdowns for instance in the case of applying
updates fixed in LUCENE-8602. This change extracts the common
usage and shares the interation code including logic to reuse
Terms and PostingsEnum instances as much as possble and adds
tests for it.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/FrozenBufferedUpdates.java
+++ b/lucene/core/src/java/org/apache/lucene/index/FrozenBufferedUpdates.java
@@ -86,7 +86,7 @@ final class FrozenBufferedUpdates {
                                    // only have Queries and doc values updates
   private final InfoStream infoStream;
 
-  public FrozenBufferedUpdates(InfoStream infoStream, BufferedUpdates updates, SegmentCommitInfo privateSegment) throws IOException {
+  public FrozenBufferedUpdates(InfoStream infoStream, BufferedUpdates updates, SegmentCommitInfo privateSegment) {
     this.infoStream = infoStream;
     this.privateSegment = privateSegment;
     assert updates.deleteDocIDs.isEmpty();
@@ -468,9 +468,6 @@ final class FrozenBufferedUpdates {
                                             long delGen,
                                             boolean segmentPrivateDeletes) throws IOException {
 
-    TermsEnum termsEnum = null;
-    PostingsEnum postingsEnum = null;
-
     // TODO: we can process the updates per DV field, from last to first so that
     // if multiple terms affect same document for the same field, we add an update
     // only once (that of the last term). To do that, we can keep a bitset which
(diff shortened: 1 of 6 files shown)


lucene/core/src/java/org/apache/lucene/index/FrozenBufferedUpdates.java
```diff
@@ -821,4 +759,93 @@ public String toString() {
...
+    TermDocsIterator(LeafReader reader, boolean sortedTerms) {
+      this(reader::terms, sortedTerms);
+    }
+
+    private TermDocsIterator(TermsProvider provider, boolean sortedTerms) {
+      this.sortedTerms = sortedTerms;
+      this.provider = provider;
+    }
+
+    private void nextField(String field) throws IOException {
```
mikemccand: Maybe simply name `setField` or `changeField`?  `nextField` made me think it was somehow following along in sorted field order or something.

lucene/core/src/java/org/apache/lucene/index/FrozenBufferedUpdates.java
```diff
@@ -821,4 +759,93 @@ public String toString() {
...
+          termsEnum = null;
+        }
+      }
+    }
+
+    DocIdSetIterator nextTerm(String field, BytesRef term) throws IOException {
+      nextField(field);
+      if (termsEnum != null) {
+        if (sortedTerms) {
+          // in the sorted case we can take advantage of the "seeking forward" property
```
mikemccand: Maybe we should confirm the incoming `term` is `>=` the last term, under assert?

lucene/core/src/java/org/apache/lucene/index/FrozenBufferedUpdates.java
```diff
@@ -821,4 +759,93 @@ public String toString() {
...
+          // this allows us depending on the term dict impl to reuse data-strucutres internally
```
mikemccand: `data-strucutres` -> `data-structures`

lucene/core/src/java/org/apache/lucene/index/FrozenBufferedUpdates.java
```diff
@@ -821,4 +759,93 @@ public String toString() {
...
+          // which speed up iteration over terms and docs significantly.
+          int cmp = term.compareTo(readerTerm);
+          if (cmp < 0) {
+            return null;
```
mikemccand: Maybe add comment `// requested term does not exist in this segment`?

lucene/core/src/test/org/apache/lucene/index/TestFrozenBufferedUpdates.java
```diff
@@ -0,0 +1,104 @@
...
+import org.apache.lucene.util.BytesRefIterator;
+import org.apache.lucene.util.Counter;
+import org.apache.lucene.util.FixedBitSet;
+import org.apache.lucene.util.LuceneTestCase;
+import org.apache.lucene.index.FrozenBufferedUpdates.TermDocsIterator;
+import org.apache.lucene.util.TestUtil;
+
+public class TestFrozenBufferedUpdates extends LuceneTestCase {
+
+  public void testTermDocsIterator() throws IOException {
```
mikemccand: Nice random test :)

s1monw:
@mikemccand I pushed changes. I think it's ready

PR #527, opened 2018-12-13
LUCENE-8609: Allow getting consistent docstats from IndexWriter

s1monw:
Today we have #numDocs() and #maxDoc() on IndexWriter. This is enough
to get all stats for the current index but it's subject to concurrency
and might return numbers that are not consistent ie. some cases can
return maxDoc < numDocs which is undesirable. This change adds a getDocStats()
method to index writer to allow fetching consistent numbers for these stats.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
+++ b/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
@@ -1131,18 +1131,6 @@ public class IndexWriter implements Closeable, TwoPhaseCommit, Accountable,
     return analyzer;
   }
 
-  /** Returns total number of docs in this index, including
-   *  docs not yet flushed (still in the RAM buffer),
-   *  not counting deletions.
-   *  @see #numDocs
-   *  @deprecated use {@link #getDocStats()} instead
-   *  */
-  @Deprecated
-  public synchronized int maxDoc() {
-    ensureOpen();
-    return docWriter.getNumDocs() + segmentInfos.totalMaxDoc();
-  }
-
   /** If {@link SegmentInfos#getVersion} is below {@code newVersion} then update it to this value.
    *
    * @lucene.internal */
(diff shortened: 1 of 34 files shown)


mikemccand:
LGTM

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
```diff
@@ -5289,4 +5289,48 @@ final synchronized boolean segmentCommitInfoExist(SegmentCommitInfo sci) {
...
+   * Returns accurate {@link DocStats} form this writer. This is equivalent to calling {@link #numDocs()} and {@link #maxDoc()}
+   * but is not subject to race-conditions. The numDoc for instance can change after maxDoc is fetched that causes numDocs to be
+   * greater than maxDoc which makes it hard to get accurate document stats from IndexWriter.
+   */
+  public synchronized DocStats getDocStats() {
+    ensureOpen();
+    int numDocs = docWriter.getNumDocs();
+    int maxDoc = numDocs;
+    for (final SegmentCommitInfo info : segmentInfos) {
+      maxDoc = info.info.maxDoc();
```
dnhatn: `=` -> `+=`.

lucene/core/src/test/org/apache/lucene/index/TestIndexWriter.java
```diff
@@ -3147,7 +3162,7 @@ public void testSoftUpdateDocuments() throws IOException {
     for (SegmentCommitInfo info : writer.cloneSegmentInfos()) {
      numSoftDeleted += info.getSoftDelCount();
     }
-    assertEquals(writer.maxDoc() - writer.numDocs(), numSoftDeleted);
+    assertEquals(writer.getDocStats().maxDoc - writer.getDocStats().numDocs, numSoftDeleted);
```
dnhatn: maybe use a single docStats?

lucene/core/src/test/org/apache/lucene/index/TestIndexWriter.java
```diff
@@ -3300,7 +3315,7 @@ public int numDeletesToMerge(SegmentCommitInfo info, int delCount, IOSupplier<Co
     for (SegmentCommitInfo info : writer.cloneSegmentInfos()) {
       numSoftDeleted += info.getSoftDelCount() + info.getDelCount();
     }
-    assertEquals(writer.maxDoc() - writer.numDocs(), numSoftDeleted);
+    assertEquals(writer.getDocStats().maxDoc - writer.getDocStats().numDocs, numSoftDeleted);
```
dnhatn: maybe use a single docStats?

PR #519, opened 2018-12-08
WIP: SOLR-12955: rebase and split changes to smaller commits

barrotsteindev:

PR #528, opened 2018-12-13
SOLR-12955 2

barrotsteindev:


barrotsteindev:
Tackling this again,
according to @dsmiley's input:

> I'm not sure if the "base" functionality needed to be in a new class... I could imagine simply adding subclasses and leave DUP named as it is with it's base functionality.

as stated in pr [519](https://github.com/apache/lucene-solr/pull/519)

solr/core/src/java/org/apache/solr/update/processor/DistributedZkUpdateProcessor.java
```diff
@@ -58,6 +75,127 @@ String getCollectionName(CloudDescriptor cloudDesc) {
     return cloudDesc.getReplicaType();
   }
 
+  @Override
+  public void doDeleteByQuery(DeleteUpdateCommand cmd) throws IOException {
+    // even in non zk mode, tests simulate updates from a leader
```
dsmiley: this particular comment is I think intended for the standalone mode

solr/core/src/java/org/apache/solr/update/processor/DistributedStandaloneUpdateProcessor.java
```diff
@@ -56,8 +59,15 @@ public void doDeleteByQuery(DeleteUpdateCommand cmd) throws IOException {
...
 
+  @Override
+  protected void postProcessDeleteByQuery(DeleteUpdateCommand cmd,
+                                          List<SolrCmdDistributor.Node> replicas, DocCollection coll) throws IOException {
+    // no op
+  }
+
   @Override
   boolean isLeader(UpdateCommand cmd) {
+    updateCommand = cmd;
```
dsmiley: ouch this piece is ugly.  a method like "is..." (or get...)  should not have side effects.

solr/core/src/java/org/apache/solr/update/processor/DistributedZkUpdateProcessor.java
```diff
@@ -223,6 +289,7 @@ private String getLeaderUrlZk(String id) {
 
   @Override
   boolean isLeader(UpdateCommand cmd) {
+    updateCommand = cmd;
```
dsmiley: again; this line makes me frown

solr/core/src/java/org/apache/solr/update/processor/DistributedUpdateProcessor.java
```diff
@@ -1417,45 +1407,7 @@ private void doDeleteById(DeleteUpdateCommand cmd) throws IOException {
...
-      params.set(DISTRIB_FROM, ZkCoreNodeProps.getCoreUrl(
-          zkController.getBaseUrl(), req.getCore().getName()));
-
-      if (req.getParams().get(UpdateRequest.MIN_REPFACT) != null) {
-        // TODO: Kept for rolling upgrades only. Remove in Solr 9
-        params.add(UpdateRequest.MIN_REPFACT, req.getParams().get(UpdateRequest.MIN_REPFACT));
-      }
-      cmdDistrib.distribDelete(cmd, nodes, params, false, rollupReplicationTracker, leaderReplicationTracker);
-    }
+    postProcessDeleteById(cmd);
```
dsmiley: I wonder if we actually need this postPorcessDeleteById(cmd) method... could the subclass put its logic after calling super.doDeleteById()?

barrotsteindev: Well I could try,
I didn't want to introduce too many changes to the logic. Would changing this and letting the full sold test suite run be of assurance that the change is safe?

solr/core/src/java/org/apache/solr/update/processor/DistributedStandaloneUpdateProcessor.java
barrotsteindev: since this is a package private method, perhaps it should be renamed to setupRequest(like the protected methods) just that this once takes an update command( Distributed#setupRequest(UpdateCommand cmd) ). It should be stated in the java doc that the method returns whether this node is the leader for the supplied update command(cmd).
WDYT?

dsmiley: +1 to setupRequest name.  I don't think it needs to return "isLeader"; it should set this and return void.

solr/core/src/java/org/apache/solr/update/processor/DistributedUpdateProcessor.java
dsmiley: I can understand why you took the current approach.  Lets put this off towards near the end to see if it's straight-forward or not.  Maybe it won't be and we shouldn't do it.

solr/core/src/java/org/apache/solr/update/processor/DistributedStandaloneUpdateProcessor.java
barrotsteindev: Perhaps we should have another isLeader method that returns a boolean, since the current method is used by other classes in the package to setup the request and determine whether the core is the leader for the specified update command.
I'll work on a patch ASAP.

barrotsteindev:
addressed comments about previous PR and re-based on latest master

dsmiley:
Looks good.  More to do?

barrotsteindev:
Oh yes,
I still have a few Todo's I added that I'd like to tackle.
I won't have time for more today,
Hopefully I'll be able to get these done by the end of the week.

solr/core/src/java/org/apache/solr/update/processor/DistributedUpdateProcessor.java
```diff
@@ -1250,23 +1250,10 @@ private UpdateCommand fetchFullUpdateFromLeader(AddUpdateCommand inplaceAdd, lon
...
-            + "failed since we're not in cloud mode.");
-      }
-      Replica leader;
-      try {
-        leader = zkController.getZkStateReader().getLeaderRetry(collection, cloudDesc.getShardId());
-      } catch (InterruptedException e) {
-        throw new SolrException(ErrorCode.SERVER_ERROR, "Exception during fetching from leader.", e);
-      }
-      leaderUrl = leader.getCoreUrl();
+    String leaderUrl = getLeaderUrl(id);
```
dsmiley: I'm not sure how this diff is a refactoring (keeps semantics identical).  For example... the exception message isn't the same.

barrotsteindev: Hmmm, if it is crucial to keep the exception message identical,
I could throw the exceptions inside getLeaderUrl(the abstract method).

solr/core/src/java/org/apache/solr/update/processor/DistributedStandaloneUpdateProcessor.java
```diff
@@ -42,15 +53,36 @@ public DistributedStandaloneUpdateProcessor(SolrQueryRequest req,
...
   }
 
+  @Override
+  public void processCommit(CommitUpdateCommand cmd) throws IOException {
+
+    assert TestInjection.injectFailUpdateRequests();
+
+    updateCommand = cmd;
+
+    CompletionService<Exception> completionService = new ExecutorCompletionService<>(req.getCore().getCoreContainer().getUpdateShardHandler().getUpdateExecutor());
```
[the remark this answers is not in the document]
barrotsteindev: I guess I missed the fact that replicas in standalone mode are always NRT,
right?

solr/core/src/java/org/apache/solr/update/processor/DistributedStandaloneUpdateProcessor.java
```diff
@@ -0,0 +1,121 @@
...
+  @Override
+  protected void doDeleteById(DeleteUpdateCommand cmd) throws IOException {
+    isLeader = getNonZkLeaderAssumption(req);
+    super.doDeleteById(cmd);
+  }
+
+  @Override
+  protected void doDeleteByQuery(DeleteUpdateCommand cmd) throws IOException {
+    // even in non zk mode, tests simulate updates from a leader
+    isLeader = getNonZkLeaderAssumption(req);
```
dsmiley: I think just call setupRequest instead of getNonZkLeaderAssumption?  setupRequest calls that and sets isLeader.  This comment goes for all/most of the methods here.

solr/core/src/java/org/apache/solr/update/processor/DistributedStandaloneUpdateProcessor.java
```diff
@@ -0,0 +1,121 @@
...
+    super.doDeleteByQuery(cmd, null, null);
+  }
+
+  @Override
+  public void finish() throws IOException {
```
dsmiley: isn't this what a base finish() will do?  so why subclass.  If there really is a reason then a comment here would help

solr/core/src/java/org/apache/solr/update/processor/DistributedUpdateProcessor.java
dsmiley: I looked again and it seems fine.  I could be the diff I was looking at confused the matter.

solr/core/src/java/org/apache/solr/update/processor/DistributedStandaloneUpdateProcessor.java
barrotsteindev: Well, previously, the code was:
  ```
@Override
public void finish() throws IOException {
    assert ! finished : "lifecycle sanity check";
    finished = true;
    
    if (zkEnabled) doFinish();
    
    if (next != null && nodes == null) next.finish();
  }
```
So nodes was always set to null when zkEnabled was false(see DistributedUpdateProcessor#setupRequest on master).
So, I just split the two, zkProc calls doFinish, while StandaloneProc calls next.finish(),
and the base class only changes isFinished and asserts it's not true:
```
@Override
  public void finish() throws IOException {
    assert ! finished : "lifecycle sanity check";
    finished = true;
  }
```

solr/core/src/java/org/apache/solr/update/processor/DistributedStandaloneUpdateProcessor.java
barrotsteindev: I think it can also be used in the zkProcessor,
I'll investigate further.

solr/core/src/java/org/apache/solr/update/processor/DistributedStandaloneUpdateProcessor.java
barrotsteindev: I'm still trying to dig in further and see why the last line (`if (next != null && nodes == null) next.finish();`) was put in an if and not an else if statement.
nodes is supposed to be the replicas of the current nodes, and when zkEnabled was false it was set to null. I am still no sure why this was in a completely separate if statement.
@dsmiley , WDYT?

dsmiley: nevermind; just leave this be then.  Lets keep our changes to more clear/obvious ones.

solr/core/src/java/org/apache/solr/update/processor/DistributedStandaloneUpdateProcessor.java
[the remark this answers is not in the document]
barrotsteindev: Hopefully I addressed your concerns with latest commit.

dsmiley:
* DUP.postProcessAdd's javadoc is confusing; maybe drop it.  It says "This method can be overridden to tamper with the cmd" but I'm looking at where we override it and I don't see any tampering to this.updateCommand.
* DUP: Lets remove the Standalone subclass in favor of making the base class do the standalone functionality; I think it's not really doing much of anything other than relying on the base impl.  There's practically no functionality (real code) in the standalone class right now so it's kind of pointless.
* DistributedZkUpdateProcessor's constructor that takes a docMerger isn't called externally so lets have just one constructor

dsmiley:
I like that the zk version of processCommit completely overrides the superclass since the logic is much more complicated.  But it's a little unfortunate to call super.processCommit which will have some redundant setupRequest.  One slight change would be for these method calls to change back to call localCommit() as they did before, but also put the cmd.setVersion logic at the front of localCommit's logic since this is always done first.

dsmiley:
Man this is a tangled mess we are trying to unravel!  I wish someone undertook this years ago.

Lets not rename DUP#processDelete as it is an URP defined method we override, and I doubt it'd worth it to expand the change to URP.

postProcessDeleteByQuery can be renamed to doDistribDeleteByQuery, and postProcessDeleteById can be renamed doDistribDeleteById.

The flow of doDeleteById & doDeleteByQuery are different, which is a bit annoying. It's hard for me to make much sense of it right now.  I know the ZK version entirely overrides processCommit but I think there the flow was more clear, at least to me.  But on the delete side, I find it confusing that the ZK version overrides doDeleteByQuery. ... gotta go for now

Let me know when you're done with the issue and I can try some thing too.  It's such a time sink unfortunately.

barrotsteindev:
> postProcessDeleteByQuery can be renamed to doDistribDeleteByQuery, and postProcessDeleteById can be renamed doDistribDeleteById.

I pushed a commit with these requested changes.

> Let me know when you're done with the issue and I can try some thing too. It's such a time sink unfortunately.

I can not seem to come up with a way to deal with doDeleteById & doDeleteByQuery, and I do not have a lot of time to sit and think about this, so I will lay off this issue for a while.
Feel free to have a go at it as well,
perhaps you might find a neater solution.

barrotsteindev:
Hey @dsmiley ,
Have you had time to look around?
Should I upload a patch in the meantime so all tests will be run(I did run the Solr test suite which seems to pass).
Thanks in advance.

barrotsteindev:
Rebased on master.
Also ran the full Solr test suite and made sure all tests pass.
@dsmiley , do you reckon we could get this one merged into master?

dsmiley:
@barrotsteindev Can you please update to master again?

barrotsteindev:
I'm on it.
There were so many changes,
hopefully I'll be able to push an up to date version tonight.

dsmiley:
@yonik do you have any overall thoughts on this PR?  In a nutshell: DURP has a lot of branching related to being in cloud mode vs standalone.  It's complex and it hampers readability.  This PR aims to resolve that by separating them out to separate classes.

barrotsteindev:
Updated to latest master,
while moving all read only functionality to DistributedZkUpdateProcessor,
since it is only supported in SolrCloud.

dsmiley:
So long as the tests pass; I'm inclined to just commit this.  I'll run them tonight.

PR #529, opened 2018-12-20
LUCENE-8617: Use SimpleFSDirectory on non-default FS

marschall:
Currently we use MMapDirectory on a non-default file system on Unix.
This will almost always fail as only the default file system can
provide mmap().

Switch to SimpleFSDirectory instead for non-default file systems
independent of the operating system.


MichaelKunze:
Any chance this can be checked and merged? Testing lucene with memory filesystems won't work otherwise. See https://github.com/marschall/memoryfilesystem/issues/113

MichaelKunze:
Any reason for closing this? Are you not considering this a problem?

rmuir:
Please, see the discussion on the associated JIRA issue.

PR #531, opened 2019-01-01
SOLR-12768

moshebla:


dsmiley:
Thanks for helping out!
It's insufficient to simply prepend a `*` since it would match other labels that share the same suffix.  For example: `childFilter='ingredients/name\_s:cocoa'` would get transformed into a query that might match a label `secretingredients` but we don't want it to.  I think a solution is to build an OR query where one clause has a `*/` prepended to it, and the other has nothing modified and so matches the term exactly.  WDYT?

moshebla:
> I think a solution is to build an OR query where one clause has a */ prepended to it, and the other has nothing modified and so matches the term exactly. WDYT?

Yes that seems correct, and all added unit tests pass.
I just pushed a new commit implementing this.

dsmiley:
Looks good.  I thought you were going to test with something to show we don't screw it up?  You could search for 'redients' which would have returned something before but now should not.

I wonder... hmmm... maybe the nest path should always start with a '/'?  It would seem more correct since paths in general usually start with one, and it would also make implementing this case simpler.

moshebla:
> I wonder... hmmm... maybe the nest path should always start with a '/'? It would seem more correct since paths in general usually start with one, and it would also make implementing this case simpler.

Uploaded a new commit with the above requested change.
I also fixed TestNestedUpdateProcessor and TestChildDocTransformerHierarchy.

moshebla:
Perhaps we could test whether "foo" is a defined field in the current collection?
I'll try and test this on Sunday as I have got to go.
I personally prefer keeping the API as simple as possible(I'm a believer of the KISS principle).

dsmiley:
> Perhaps we could test whether "foo" is a defined field in the current collection?

Eh; that sounds too much like a "guess what the user might mean" kind of solution.  Even if "foo" is in the schema, it doesn't prevent an element "foo" from being used.

> I personally prefer keeping the API as simple as possible(I'm a believer of the KISS principle).

At the expense of ambiguity?  Trade-offs.

dsmiley:
Code changes seem good; I'll look closer and run tests.

PathHierarchyTokenizerFactory _might_ eventually be used at query time if we had a syntax to say "get me all my ancestors".  But given it's conditional based on syntax, it can't easily go into the query analyzer, so I think it's better to leave the query time chain simple/direct and we do this stuff when looking at the syntax.  I know this has evolved recently from where we were.  In this issue I removed PathHierarchyTokenizer from the index side because I feel it's better to err on lighter-weight indexing at the expense of slower queries since I think it's very likely a very low cost for what I think are typical use-cases.

RE issue to discuss the syntax: yes a subtask of SOLR-12298.  no hurry on that one

moshebla:
Closing since this ticket has been resolved.

PR #533, opened 2019-01-14
LUCENE-8636: TestPointQueries and long execution times

dweiss:
Applied Adrien's feedback.


atris:
Is this redundant now? Should we close it?

PR #535, opened 2019-01-15
LUCENE-8639: Prevent new threadstates from being created while we cut over to a new delete queue

s1monw:
This prevents an edge case where suddenly a lot of threads start indexing
while we carry over sequence ids from the previous to the new delete queue.
We now lock creation of new thread states for a very short time until we created and assigned
a new delete queue.


mikemccand:
LGTM -- sneaky.

mikemccand:
I like the test case!!

PR #540, opened 2019-01-18
minor syntax corrections in CSS files of Angular UI

saschaszott:
There are some minor syntax bugs in the CSS files of the Angular UI module.


saschaszott:
this PR needs to be splitted into two PRs

PR #545, opened 2019-01-21
SOLR-13157: Convert Unicode double quotes to standard quotes

jnyryan:
https://issues.apache.org/jira/browse/SOLR-13157

Updated the dismax parser to find any Unicode double quotes and replace them with a standard double quotes.


ctargett:
Added Jira issue ID to the title, that allows Jira to pick up changes, merges, commits from GH.

PR #553, opened 2019-01-29
SOLR-9515: Update to Hadoop 3

risdenk:
Builds on existing branch from @markrmiller. Updated to master branch and made sure that HDFS tests are passing. Added comments to a few files on the PR as to why it was changed.


solr/core/src/test/org/apache/hadoop/http/HttpServer2.java
```diff
@@ -0,0 +1,1685 @@
+/*
```
risdenk: This file is copied directly from Apache Hadoop to allow integration tests to run under Jetty 9.4.

I couldn't find a way to get the shaded Hadoop dependencies (new in Hadoop 3) to work since they shaded too much (like javax.servlet). This made our integration with hadoop-auth break.

I have no idea if copying HttpServer2 in here will break other things. I tested with JDK 8 and so far this looks good.

risdenk: @uschindler I know you have a lot of experience with JDK 9+ and packaging. Is this even allowed to have a separate package `org.apache.hadoop...` inside the Solr test directory? Will it cause issues with other JDK versions?

solr/core/src/test/org/apache/solr/handler/export/TestExportWriter.java
```diff
@@ -28,14 +28,14 @@
...
 
+import com.fasterxml.jackson.databind.ObjectMapper;
 import org.apache.lucene.util.TestUtil;
 import org.apache.solr.SolrTestCaseJ4;
 import org.apache.solr.common.SolrInputDocument;
 import org.apache.solr.common.util.SuppressForbidden;
 import org.apache.solr.common.util.Utils;
 import org.apache.solr.request.SolrQueryRequest;
 import org.apache.solr.schema.SchemaField;
-import org.codehaus.jackson.map.ObjectMapper;
```
risdenk: Test used to use old Jackson class. Replaced with newer one and removed old Jackson pulled in by Hadoop.

solr/core/src/test/org/apache/solr/response/TestCustomDocTransformer.java
```diff
@@ -26,7 +26,6 @@
 import org.apache.solr.request.SolrQueryRequest;
 import org.apache.solr.response.transform.DocTransformer;
 import org.apache.solr.response.transform.TransformerFactory;
-import org.bouncycastle.util.Strings;
```
risdenk: Test used to use old bouncycastle dependency which isn't needed anymore from Hadoop. Switched to use builtin Java split.

solr/core/src/test/org/apache/solr/security/JWTAuthPluginTest.java
```diff
@@ -44,7 +44,6 @@
 import org.junit.Before;
 import org.junit.BeforeClass;
 import org.junit.Test;
-import org.mortbay.util.ajax.JSON;
```
risdenk: Avoid using old mortbay utilities for converting JSON. Uses existing Solr Utils to convert from JSON string.

solr/test-framework/src/java/org/apache/solr/util/BadHdfsThreadsFilter.java
```diff
@@ -29,16 +29,18 @@ public boolean reject(Thread t) {
       return true;
     } else if (name.startsWith("org.apache.hadoop.hdfs.PeerCache")) { // SOLR-7288
       return true;
+    } else if (name.endsWith("StatisticsDataReferenceCleaner")) {
```
risdenk: I have not tracked down why this is necessary yet.

solr/test-framework/src/java/org/apache/solr/util/BadHdfsThreadsFilter.java
```diff
@@ -29,16 +29,18 @@ public boolean reject(Thread t) {
...
+      return true;
     } else if (name.startsWith("LeaseRenewer")) { // SOLR-7287
       return true;
     } else if (name.startsWith("org.apache.hadoop.fs.FileSystem$Statistics")) { // SOLR-11261
       return true;
     } else if (name.startsWith("ForkJoinPool.")) { // JVM built in pool
       return true;
+    } else if (name.startsWith("ForkJoinPool-")) { // JVM built in pool
```
risdenk: I have not tracked down why this is necessary yet.

lucene/tools/src/groovy/check-source-patterns.groovy
```diff
@@ -149,6 +149,7 @@ ant.fileScanner{
     exclude(name: 'lucene/benchmark/temp/**')
     exclude(name: '**/CheckLoggingConfiguration.java')
     exclude(name: 'lucene/tools/src/groovy/check-source-patterns.groovy') // ourselves :-)
+    exclude(name: 'solr/core/src/test/org/apache/hadoop/**')
```
risdenk: Needed to skip checking the copied HttpServer2 code from Apache Hadoop.

solr/core/build.xml
```diff
@@ -25,6 +25,7 @@
 
   <property name="forbidden-tests-excludes" value="
     org/apache/solr/internal/**
+    org/apache/hadoop/**
```
risdenk: Needed to skip checking the copied HttpServer2 code from Apache Hadoop.

risdenk:
Would appreciate a review - @markrmiller, @uschindler, @sigram, @ctargett and whoever else is interested.

uschindler:
Hi,
I don't fully understand why you need the clone of the HttpServer2.java source code. Isn't this included in hadoop anyways. Ot has this some special reason like replacing a shaded version inside hadoop. It's gppd that we have removed the mortbay jetty now from test dependencies, so I have the feeling this does something like this to prevent another version of jetyt to be included.

So please explain!

As long as Solr uses classpath instead of modules this won't bring any issues - if the classpath order for running tests is fine.

risdenk:
@uschindler sorry I definitely didn't explain it as well as I could have. Hadoop moved from old mortba Jetty to Jetty 9.3. Solr is on Jetty 9.4. Session management changed between Jetty 9.3 and 9.4. For the integration tests with Hadoop, the classpath has Jetty 9.4. Adding Jetty 9.3 isn't possible since then the new HTTP2 tests don't work. The smallest change I could make was to copy HttpServer2 from Hadoop and fix the session management sections for Jetty 9.4. The changes are similar to https://issues.apache.org/jira/browse/HADOOP-14930 which looked at upgrading Hadoop to Jetty 9.4.

So yes your understanding is correct in that we can't have both Jetty 9.3 and 9.4 on the Solr test classpath. The testing I have done shows that the Hadoop integration tests work with Jetty 9.4 with the copied/patched HttpServer2 code.

risdenk:
ping @hgadre since you filled SOLR-9761 - I know I didn't get to all the subtasks under SOLR-9761 but would appreciate any thoughts you might have.

uschindler:
@risdenk Thanks. That's exactly what I expected. Thanks for explanation. I am fine with that solution (because it's test only, so we don't deliver that to the end user for use in production).

To me it's only important to clarify: If we want to go that route, it's important to have our own test classes BEFORE the hadoop imports on classpath. We have to make sure that this also applies for the Eclipse/NetBeans/... Development integrations!

So I am fine with the approach, but we need to make the build system be informed about that.

We also should not disable our source code and license checks on the forked Hadoop classes! If there are violations, we should maybe fix them at least in the code we imported.

risdenk:
@uschindler yea so I tried fixing a lot of the failures. Some up the issues I ran into:
* Changing LOG -> log to match the source patterns check broke other Hadoop classes. It turns out Hadoop uses the LOG since it is public outside the HttpServer2 class. I couldn't find a way to suppress just the source check for that file.
* Forbiddenapis (even with SuppressForbidden) kept saying that https://github.com/apache/hadoop/blob/branch-3.2/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/http/HttpServer2.java#L1500 was invalid due to https://github.com/apache/lucene-solr/blob/master/lucene/tools/forbiddenApis/servlet-api.txt#L18. I tried to add SuppressForbidden at the class, static class, method, and override methods to fix the issue, but the errors still came back.

So with the above two issues that is how I ended up suppressing the checks on the HttpServer2 class. The other thought I had is that it would be nice to not edit HttpServer2 too much in case we need to copy changes back from Apache Hadoop again.

I am open to suggestions on how to move forward without suppressing the checks.

risdenk:
> To me it's only important to clarify: If we want to go that route, it's important to have our own test classes BEFORE the hadoop imports on classpath. We have to make sure that this also applies for the Eclipse/NetBeans/... Development integrations!

@uschindler any suggestions on how to accomplish this? I understand what you mean but not sure how to go about it.

solr/core/src/test/org/apache/solr/cloud/hdfs/HdfsTestUtil.java
```diff
@@ -67,6 +68,14 @@ public static MiniDFSCluster setupClass(String dir, boolean safeModeTesting, boo
     LuceneTestCase.assumeFalse("HDFS tests were disabled by -Dtests.disableHdfs",
       Boolean.parseBoolean(System.getProperty("tests.disableHdfs", "false")));
 
+    // Checks that commons-lang3 FastDateFormat works with configured locale
```
risdenk: Currently certain locales like "ja-JP-u-ca-japanese-x-lvariant-JP" cause `FastDateFormat` from commons-lang3 to fail.

risdenk:
So far I've loaded the project in IntelliJ (`ant intellij`) and Eclipse (`ant eclipse`). In both cases the patched `HttpServer2` class was used to run the tests. I'll try with Netbeans next.

uschindler:
Hi,

> * Forbiddenapis (even with SuppressForbidden) kept saying that https://github.com/apache/hadoop/blob/branch-3.2/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/http/HttpServer2.java#L1500 was invalid due to https://github.com/apache/lucene-solr/blob/master/lucene/tools/forbiddenApis/servlet-api.txt#L18. I tried to add SuppressForbidden at the class, static class, method, and override methods to fix the issue, but the errors still came back.

That's indeed strange. The reason here is that it overrides a method that is forbidden. So the only chance is to add @SuppressForbidden to the inner static class "RequestQuoter". I will do a check and maybe open bug in forbiddenapis. I will do some tries once I have time.

risdenk:
@uschindler thanks - any concerns with merging this to the master branch before you have a chance to check? I'd be interested in getting this change into Solr 8.0 if possible.

risdenk:
Squashed the multiple commits into one and added CHANGES.txt entry. Planning to push this to master and 8.x in a little bit.

solr/core/src/test/org/apache/solr/cloud/hdfs/HdfsTestUtil.java
risdenk: Turns out only locale "ja-JP-u-ca-japanese-x-lvariant-JP" causes this failure. Emailed commons-user list about this. No response yet. I have not seen any failures in the past ~24 hours with this fix put in place.

risdenk:
> I'll try with Netbeans next.
I tried with Netbeans 8.x and tests ran successfully. Netbeans 10.x hung on my machine and didn't track that down.

risdenk:
Pushed to master. Will wait a few hours and then push to other branches.

risdenk:
@uschindler Want to make sure you don't think I am trying to avoid fixing the forbiddenapis/source check stuff. I plan to fix it but since it is test only and want to make sure this cleanup ends up in 8.0 if possible.

risdenk:
Rebased on master and updated the workaround for BlockPoolSlice. Think that also found the issue with threads not being cleaned up. Running through all tests now.

uschindler:
+1

risdenk:
Well naively adding shutdown of the thread pool results in a lot of `RejectedExecutionException`s since the pool is static. Shutting down is per the lifetime of one BlockPoolSlice. The lifetime of the Hadoop `addReplicaThreadPool` needs some more thought.  I don't know if it is completely possible to fix this only in BlockPoolSlice. Planning to leave it for now as a bad thread.

solr/core/src/test/org/apache/hadoop/hdfs/server/datanode/fsdataset/impl/BlockPoolSlice.java
```diff
@@ -0,0 +1,1045 @@
+/*
```
risdenk: Copied straight from Hadoop code base with minor modifications to fix SecurityManager and ForkJoinPool integration. Details in SOLR-9515 and HDFS-14251

solr/core/src/test/org/apache/solr/cloud/hdfs/HdfsTestUtil.java
```diff
@@ -62,73 +63,88 @@
...
   public static MiniDFSCluster setupClass(String dir, boolean haTesting) throws Exception {
     return setupClass(dir, haTesting, true);
   }
-  
+
+  /**
+   * Checks that commons-lang3 FastDateFormat works with configured locale
+   */
+  @SuppressForbidden(reason="Call FastDateFormat.format same way Hadoop calls it")
+  private static void checkFastDateFormat() {
```
risdenk: commons-lang3 has an issue with the locale `ja-JP-u-ca-japanese-x-lvariant-JP` but this will catch if there are other locales with issues.

solr/core/src/test/org/apache/solr/cloud/hdfs/HdfsTestUtil.java
```diff
@@ -62,73 +63,88 @@
...
+    } catch (ArrayIndexOutOfBoundsException e) {
+      LuceneTestCase.assumeNoException("commons-lang3 FastDateFormat doesn't work with " +
+          Locale.getDefault().toLanguageTag(), e);
+    }
+  }
+
+  /**
+   * Hadoop fails to generate locale agnostic ids - Checks that generated string matches
+   */
+  private static void checkGeneratedIdMatches() {
```
risdenk: Hadoop generates filename ids with `String.format` without any Locale set and then uses regex to try to match it. This fails on locales `th-TH-u-nu-thai-x-lvariant-TH` and `hi-IN`. This check matches what Hadoop is doing to make sure there is a match.

solr/core/src/test/org/apache/solr/cloud/hdfs/HdfsTestUtil.java
```diff
@@ -62,73 +63,88 @@
...
+    checkGeneratedIdMatches();
 
-    savedLocale = Locale.getDefault();
-    // TODO: we HACK around HADOOP-9643
-    Locale.setDefault(Locale.ENGLISH);
-    
     if (!HA_TESTING_ENABLED) haTesting = false;
-    
-    
-    // keep netty from using secure random on startup: SOLR-10098
```
risdenk: Not needed after netty upgrade.

solr/core/src/test/org/apache/solr/cloud/hdfs/HdfsTestUtil.java
```diff
@@ -239,4 +253,24 @@ public static String getURI(MiniDFSCluster dfsCluster) {
...
+   * By default in JDK9+, the ForkJoinWorkerThreadFactory does not give SecurityManager permissions
+   * to threads that are created. This works around that with a custom thread factory.
+   * See SOLR-9515 and HDFS-14251
+   * Used in org.apache.hadoop.hdfs.server.datanode.fsdataset.impl.BlockPoolSlice
+   */
+  public static class HDFSForkJoinThreadFactory implements ForkJoinPool.ForkJoinWorkerThreadFactory {
+    @Override
+    public ForkJoinWorkerThread newThread(ForkJoinPool pool) {
+      ForkJoinWorkerThread worker = new SecurityManagerWorkerThread(pool);
+      worker.setName("solr-hdfs-threadpool-" + worker.getPoolIndex());
```
risdenk: Name used in BadHdfsThreadFilter since this pool is not shutdown correctly by BlockPoolSlice and not easily worked around. See SOLR-9515 and HDFS-14251

solr/core/build.xml
risdenk: Also now catches the BlockPoolSlice code

risdenk:
Squashed commits back down to a single commit to make it easier to commit.

risdenk:
Planning to commit to master (again) and make sure the world doesn't fall apart (again). I'll keep and eye on Jenkins builds. If successful, I'll look at 8.x/8.0 this weekend.

risdenk:
Build https://jenkins.thetaphi.de/job/Lucene-Solr-master-Linux/23610/ includes this master commit and was BUILD SUCCESSFUL - Total time: 75 minutes 44 seconds

PR #556, opened 2019-01-31
LUCENE-8673: Use radix partitioning when merging dimensional points

iverase:
Radix selection implementation when merging BKD segments


lucene/core/src/java/org/apache/lucene/util/bkd/BKDRadixSelector.java
```diff
@@ -0,0 +1,350 @@
...
+    // We did not have any points to sort on memory,
+    // compute the partition point from the common prefix
+    byte[] partition = new byte[bytesPerDim];
+    for (int i = 0; i < bytesPerDim; i++) {
+      partition[i] = (byte)partitionBucket[i];
+    }
+    return partition;
+  }
+
+  private byte[] heapSort(HeapPointWriter writer, PointWriter left, PointWriter right, int dim, int k, int leftCounter, int commonPrefix) throws IOException {
```
jpountz: s/Sort/Select/?

lucene/core/src/java/org/apache/lucene/util/bkd/BKDRadixSelector.java
```diff
@@ -0,0 +1,350 @@
...
+  }
+
+  private int getBucket(byte[] value, int dim, int commonPrefix) {
+    return value[dim * bytesPerDim + commonPrefix] & 0xff;
+  }
+
+  /**
+   * Reusable on-heap intro  selector.
+   */
+  private static class HeapRadixSelector extends IntroSelector {
```
jpountz: This isn't using a radix approach, maybe just call it "HeapSelector"?

lucene/core/src/java/org/apache/lucene/util/bkd/BKDRadixSelector.java
```diff
@@ -0,0 +1,350 @@
...
+    @Override
+    protected void setPivot(int i) {
+      writer.getPackedValueSlice(i, pivot);
+      pivotDoc = writer.docIDs[i];
+    }
+
+    @Override
+    protected int compare(int i, int j) {
+      writer.getPackedValueSlice(i, bytesRef1);
+      writer.getPackedValueSlice(j, bytesRef2);
```
jpountz: we can nest these two calls within the following if statement?

lucene/core/src/java/org/apache/lucene/util/bkd/BKDRadixSelector.java
jpountz: and rename the member too

lucene/core/src/java/org/apache/lucene/util/bkd/BKDRadixSelector.java
```diff
@@ -0,0 +1,350 @@
...
+    try (PointReader reader = data.getReader(from, to)) {
+      if (iteration == 0) {
+        // we specialise this case
+        reader.buildHistogram(dim * bytesPerDim + commonPrefix, histogram[commonPrefix]);
+      } else {
+        while (reader.next()) {
+          byte[] packedValue = reader.packedValue();
+          if (hasCommonPrefix(packedValue, dim, commonPrefix)) {
+            int bucket = packedValue[dim * bytesPerDim + commonPrefix] & 0xff;
+            histogram[commonPrefix][bucket]++;
```
jpountz: maybe we should also write these values that have the right prefix to a temporary OfflinePointWriter so that the next iteration has less data to look at?

lucene/core/src/java/org/apache/lucene/util/bkd/PointReader.java
```diff
@@ -35,47 +33,11 @@
...
-        rightCount++;
-        if (doClearBits) {
-          rightTree.clear(ord);
-        }
-      } else {
-        left.append(packedValue, ord, docID);
-      }
-    }
+  /** Build histogram of the document at the provided byte position */
+  public abstract void buildHistogram(int bytePosition, int[] histogram) throws IOException;
```
jpountz: does this optimization really help compared to computing the histogram on top of the point reader?

iverase: It actually does quite a bit (you don't need to copy every packedValue to a byte array). Note that I am using the same strategy that the method `markOrds` is using in the current strategy.

jpountz: Agreed, but I was hoping that we could further simplify. :)

Maybe another way to make it work would be to have something like `byte byteAt(index index)` instead of `buildHistogram`, which has the benefit that it would not only be useful for the first recursion?

iverase: This method has been removed in the latest version. Since we are buffering the Offline data, it is actually faster to compute it from the point reader.

lucene/core/src/java/org/apache/lucene/util/bkd/PointWriter.java
```diff
@@ -19,24 +19,30 @@
...
  *
- *  @lucene.internal */
-public interface PointWriter extends Closeable {
-  /** Add a new point */
-  void append(byte[] packedValue, long ord, int docID) throws IOException;
+ *  @lucene.internal
+ *  */
+public interface PointWriter<T extends PointReader> extends Closeable {
+  /** Add a new point from byte array*/
+  void append(byte[] packedValue, int docID) throws IOException;
```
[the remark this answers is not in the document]
iverase: This method is used when we spill offline the incoming points on the BKD writer. We can wrap there the incoming byte[] into a BytesRef

lucene/core/src/java/org/apache/lucene/util/bkd/OfflinePointReader.java
```diff
@@ -74,55 +69,68 @@ public OfflinePointReader(Directory tempDir, String tempFileName, int packedByte
...
     }
+
     name = tempFileName;
 
     long seekFP = start * bytesPerDoc;
     in.seek(seekFP);
     countLeft = length;
-    packedValue = new byte[packedBytesLength];
-    this.longOrds = longOrds;
+    if (reusableBuffer != null) {
```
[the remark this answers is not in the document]
iverase: I am actually thinking if we should make those constructors from the readers protected. They should be always be constructed from the writers.

lucene/core/src/java/org/apache/lucene/util/bkd/BKDRadixSelector.java
```diff
@@ -0,0 +1,433 @@
...
+    }
+
+    assert leftCount + rightCount + histogram[commonPrefix][partitionBucket[commonPrefix]] == to - from;
+
+    if (commonPrefix == bytesSorted - 1) {
+      if (deltaPoints != null) {
+        deltaPoints.destroy();
+      }
+      // we are done, lets break data around. No need to sort on heap
+      return partition(points, left, right, from, to, partitionPoint, dim, null, commonPrefix, partitionPoint - from - leftCount);
```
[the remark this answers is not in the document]
iverase: This method populates the provided left and right PointWriters and it needs all the points.

lucene/core/src/java/org/apache/lucene/util/bkd/BKDRadixSelector.java
```diff
@@ -0,0 +1,433 @@
...
+    long length;
+    if (deltaPoints == null) {
+      currentPoints =  points;
+      start = from;
+      length = to - from;
+    } else {
+      currentPoints =  deltaPoints;
+      start = 0;
+      length = deltaPoints.count;
+    }
```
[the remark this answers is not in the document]
iverase: Because at the end of the recursion we need all points so we need both. Maybe we can find a more elegant way to handle this situation.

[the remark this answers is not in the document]
iverase: I have changed the algorithm so we partition data as we recurse. It does simplify the code.

lucene/core/src/java/org/apache/lucene/util/bkd/BKDWriter.java
```diff
@@ -194,40 +169,28 @@ protected BKDWriter(int maxDoc, Directory tempDir, String tempFileNamePrefix, in
...
     // time in the recursion, we hold the number of points at that level, plus
     // all recursive halves (i.e. 16 + 8 + 4 + 2) so the memory usage is 2X
     // what that level would consume, so we multiply by 0.5 to convert from
-    // bytes to points here.  Each dimension has its own sorted partition, so
-    // we must divide by numDims as wel.
+    // bytes to points here.  In addition the radix partitioning may sort on memory
+    // double of this size so we multiply by another 0.5.
 
-    maxPointsSortInHeap = (int) (0.5 * (maxMBSortInHeap * 1024 * 1024) / (bytesPerDoc * numDataDims));
+    maxPointsSortInHeap = (int) (0.25 * (maxMBSortInHeap * 1024 * 1024) / (bytesPerDoc));
```
jpountz: I understand the first 0.5 factor due to how recursion levels add up, but not the second one.

lucene/core/src/java/org/apache/lucene/util/bkd/BKDRadixSelector.java
```diff
@@ -0,0 +1,311 @@
...
+    this.tempFileNamePrefix = tempFileNamePrefix;
+  }
+
+  /**
+   *
+   * Method to partition the input data. It returns the value of the dimension where
+   * the split happens. The method destroys the original writer.
+   *
+   */
+  public byte[] select(PointWriter points, PointWriter left, PointWriter right, long from, long to, long partitionPoint, int dim) throws IOException {
```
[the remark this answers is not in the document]
iverase: At the moment no, but one of the improvements it can be done is to re-use the idea of path slices on the previous versions and keep just the first `HeapPointWriter`. Then you do not need to copy bytes between heap objects and just adjust the start end end points for the different levels.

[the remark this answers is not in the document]
iverase: Of course this needs a bit of work because left and right `PointWriters` should then be created inside this object.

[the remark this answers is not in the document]
jpountz: Right, let's play with this idea on a follow-up PR. +1 to keep from and to for now then.

lucene/core/src/java/org/apache/lucene/util/bkd/BKDWriter.java
iverase: This is due to the fact that BKDRadixSelector can build a buffer up to `2 * maxPointsSortInHeap`. Thinking about it, it might not be ideal(we are not using all available heap) but I have in mind a follow up issue where this will be handled better so I will keep it like that.

iverase:
Change has been pushed to master in commit 5d1d6448b91abdebbc68b54b062340c5baa88f4c

Change has been pushed to 8.x in commit a843b3aa35fba2bab82caa2fbdcbcefa61fd40cd

PR #560, opened 2019-02-03
SOLR-13213: Search Components cannot modify "shards" parameter

janhoy:


janhoy:
Marked as "not a bug" in JIRA

PR #562, opened 2019-02-04
Don't create a LeafCollector when the Scorer for the leaf is null

msokolov:


lucene/core/src/java/org/apache/lucene/search/IndexSearcher.java
```diff
@@ -638,21 +638,16 @@ protected void search(List<LeafReaderContext> leaves, Weight weight, Collector c
...
     for (LeafReaderContext ctx : leaves) { // search each subreader
-      final LeafCollector leafCollector;
-      try {
-        leafCollector = collector.getLeafCollector(ctx);
-      } catch (CollectionTerminatedException e) {
-        // there is no doc of interest in this reader context
-        // continue with the following leaf
-        continue;
-      }
       BulkScorer scorer = weight.bulkScorer(ctx);
```
mikemccand: Should we also update javadocs for `Collector.getLeafCollector` that it will (may?) not be called when query can determine there will be no hits in this segment?

jpountz:
I had considered something like that in the past but didn't like the fact that it made things slower with queries whose bulk scorer is costly to create (eg. range or multi-term) and collectors that can skip entire segments by raising a `CollectionTerminatedException` in `getLeafCollector`. Maybe we could have a way to construct bulk scorers in two steps similarly to `Weight#scorerSupplier` for regular scorers so that we could check whether a bulk scorer might have any hits without paying the full construction price. I expect this would only help if a leaf collector is expensive to construct, with which collector are you seeing slowdowns because Lucene is constructing leaf collectors on segments that don't have matches?

msokolov:
Oh good point! I hadn't realized that creating a scorer could be the more
expensive path. Actually I mostly just saw this as a cleanup with only
slighlt positive or negligble speed gain. Our Collectors are not that
expensive to create, so I think this is not really needed and we should not
do it

lucene/test-framework/src/java/org/apache/lucene/search/QueryUtils.java
```diff
@@ -40,9 +39,9 @@
 import org.apache.lucene.util.LuceneTestCase;
 import org.apache.lucene.util.Version;
 
-import static junit.framework.Assert.assertEquals;
-import static junit.framework.Assert.assertFalse;
-import static junit.framework.Assert.assertTrue;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
```
atris: Does this change need to be part of this CR?

msokolov: It's not required. I did it because junit.framework.Assert is deprecated.

PR #564, opened 2019-02-05
prorated early termination

msokolov:


msokolov:
Abandoning as I plan to post a better alternative that achieves the same result without the random behavior.

PR #569, opened 2019-02-08
LUCENE-8687: Optimise radix partitioning for points on heap

iverase:
This PR revives the idea of path slices so we do not have to hold many copies of the same data on memory.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/util/bkd/BKDRadixSelector.java
+++ b/lucene/core/src/java/org/apache/lucene/util/bkd/BKDRadixSelector.java
@@ -78,7 +78,7 @@ public final class BKDRadixSelector {
 
   /**
    *  It uses the provided {@code points} from the given {@code from} to the given {@code to}
-   *  to populate the {@code partitionSlices} array holder (length > 1) with two path slices
+   *  to populate the {@code partitionSlices} array holder (length &gt; 1) with two path slices
    *  so the path slice at position 0 contains {@code partition - from} points
    *  where the value of the {@code dim} is lower or equal to the {@code to -from}
    *  points on the slice at position 1.
--- a/lucene/core/src/java/org/apache/lucene/util/bkd/BKDRadixSelector.java
+++ b/lucene/core/src/java/org/apache/lucene/util/bkd/BKDRadixSelector.java
@@ -146,7 +146,8 @@ public final class BKDRadixSelector {
         int endIndex  = (commonPrefixPosition > bytesPerDim) ? bytesPerDim :  commonPrefixPosition;
         int j = FutureArrays.mismatch(commonPrefix, startIndex, endIndex, bytesRef1.bytes, bytesRef1.offset + dim * bytesPerDim + startIndex, bytesRef1.offset + dim * bytesPerDim + endIndex);
         if (j == 0) {
-          return dimCommonPrefix;
+          commonPrefixPosition = dimCommonPrefix;
+          break;
         } else if (j == -1) {
           if (commonPrefixPosition > bytesPerDim) {
             //tie-break on docID
--- a/lucene/core/src/java/org/apache/lucene/util/bkd/BKDRadixSelector.java
+++ b/lucene/core/src/java/org/apache/lucene/util/bkd/BKDRadixSelector.java
@@ -48,7 +48,7 @@ public final class BKDRadixSelector {
   private final int maxPointsSortInHeap;
   //reusable buffer
   private final byte[] offlineBuffer;
-  //holder for offlinePartition points
+  //holder for partition points
   private final int[] partitionBucket;
   // scratch object to move bytes around
   private final BytesRef bytesRef1 = new BytesRef();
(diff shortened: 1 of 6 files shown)


lucene/core/src/java/org/apache/lucene/util/bkd/BKDRadixSelector.java
```diff
@@ -82,30 +82,40 @@ public BKDRadixSelector(int numDim, int bytesPerDim, int maxPointsSortInHeap, Di
    * the split happens. The method destroys the original writer.
    *
    */
-  public byte[] select(PointWriter points, PointWriter left, PointWriter right, long from, long to, long partitionPoint, int dim) throws IOException {
+  public byte[] select(PathSlice points, PathSlice[] slices, long from, long to, long partitionPoint, int dim, int dimCommonPrefix) throws IOException {
```
jpountz: can you document in javadocs that `slices` should have a length of 2 and how it gets filled?

lucene/core/src/java/org/apache/lucene/util/bkd/BKDRadixSelector.java
```diff
@@ -176,7 +188,7 @@ private int findCommonPrefix(OfflinePointWriter points, long from, long to, int
         histogram[commonPrefix][bucket]++;
       }
     }
-    //Count left points and record the partition point
+    //Count left points and record the offlinePartition point
```
jpountz: bad search/replace?

lucene/core/src/java/org/apache/lucene/util/bkd/BKDRadixSelector.java
```diff
@@ -82,30 +82,40 @@ public BKDRadixSelector(int numDim, int bytesPerDim, int maxPointsSortInHeap, Di
...
 
-    //if all equals we just partition the data
-    if (commonPrefix ==  bytesSorted) {
-      partition(offlinePointWriter, left,  right, null, from, to, dim, commonPrefix - 1, partitionPoint);
-      return partitionPointFromCommonPrefix();
+    try (PointWriter left = getPointWriter(partitionPoint - from, "left" + dim);
+         PointWriter right = getPointWriter(to - partitionPoint, "right" + dim)) {
+      slices[0] = new PathSlice(left, 0, partitionPoint - from);
+      slices[1] = new PathSlice(right, 0, to - partitionPoint);
+      //if all equals we just offlinePartition the data
```
jpountz: bad search/replace?

lucene/core/src/java/org/apache/lucene/util/bkd/BKDRadixSelector.java
```diff
@@ -196,33 +208,45 @@ private int findCommonPrefix(OfflinePointWriter points, long from, long to, int
...
+  private int getMaxPointsSortInHeap(PointWriter left, PointWriter right) {
+    long pointsUsed = 0;
+    if (left instanceof HeapPointWriter) {
+      pointsUsed += left.count();
+    }
+    if (right instanceof HeapPointWriter) {
+      pointsUsed += right.count();
+    }
+    assert maxPointsSortInHeap >= pointsUsed;
+    return maxPointsSortInHeap - (int) pointsUsed;
```
jpountz: can we check this cast?

lucene/core/src/java/org/apache/lucene/util/bkd/BKDRadixSelector.java
```diff
@@ -294,18 +335,37 @@ protected int byteAt(int i, int k) {
...
-      }
-    }
     byte[] partition = new byte[bytesPerDim];
     points.getPackedValueSlice(partitionPoint, bytesRef1);
     System.arraycopy(bytesRef1.bytes, bytesRef1.offset + dim * bytesPerDim, partition, 0, bytesPerDim);
     return partition;
   }
+
+  PointWriter getPointWriter(long count, String desc) throws IOException {
+    if (count <= maxPointsSortInHeap / 2) {
```
jpountz: why do we divide by 2? can you add a comment?

iverase: We divided by 2 because as we recurse, we have two `heapPointWriter` as maximum, so they should not hold more than half of the `maxPointsSortInHeap`.

lucene/core/src/java/org/apache/lucene/util/bkd/BKDRadixSelector.java
iverase: I have changed the logic a bit and I am using the `maxSize` on the `HeapPointwriter` to calculate the offset to move into heap the selection.

lucene/core/src/java/org/apache/lucene/util/bkd/BKDRadixSelector.java
iverase: I have relaxed the condition and it should be bigger than 1. Javadocs added

iverase:
Pushed on master: 9db39ab0ca09d7eeece4c4bc055f7591eb789a00

Pushed on branch_8x: 793441475e837a3b8a40c74993783a73e27750b4

PR #575, opened 2019-02-15
SOLR-13235: Split Collections API Ref Guide page

ctargett:
This PR takes the massive collections-api.adoc page for the Collections API and splits it into several smaller, more focused, child pages.

The following new pages will be added:

- Cluster and Node Management (cluster-node-management.adoc): Define properties for the entire cluster; check the status of a cluster; remove replicas from a node; utilize a newly added node; add or remove roles for a node.
- Collection Management (collection-management.adoc): Create, list, reload and delete collections; set collection properties; migrate documents to another collection; rebalance leaders; backup and restore collections.
- Collection Aliasing (collection-aliasing.adoc): Create, list or delete collection aliases; set alias properties.
- Shard Management (shard-management.adoc): Create and delete a shard; split a shard into two or more additional shards; force a shard leader.
- Replica Management (replica-management.adoc): Add or delete a replica; set replica properties; move a replica to a different node.

Remaining on the Collections API page is the detail about how to make asynchronous API requests, since it applies to all commands. Intro text on collections-api.adoc is also changed to reflect it's new role.

Link references to specific API commands throughout the Ref Guide are also updated.

The page still "lives" in the same place under "SolrCloud Configuration and Parameters", where it has been in the page hierarchy. A broader re-org of the pages under the SolrCloud section is warranted, but is out of scope for now.

Also not included in this PR is any new content for v2-style API commands or missing params or examples, and those are out of scope for this change.


solr/solr-ref-guide/src/collection-management.adoc
```diff
@@ -0,0 +1,752 @@
...
+
+`createNodeSet`::
+Allows defining the nodes to spread the new collection across. The format is a comma-separated list of node_names, such as `localhost:8983_solr,localhost:8984_solr,localhost:8985_solr`.
++
+If not provided, the CREATE operation will create shard-replicas spread across all live Solr nodes.
++
+Alternatively, use the special value of `EMPTY` to initially create no shard-replica within the new collection and then later use the <<replica-management.adoc#addreplica,ADDREPLICA>> operation to add shard-replicas when and where required.
+
+`createNodeSet.shuffle`::
+Controls wether or not the shard-replicas created for this collection will be assigned to the nodes specified by the `createNodeSet` in a sequential manner, or if the list of nodes should be shuffled prior to creating individual replicas.
```
gerlowskija: [0] typo: wether/whether

ctargett:
> The only classification I thought twice about was your choice in putting `REBALANCELEADERS` in the collection-mgmt page, instead of the cluster-mgmt page. I might've put it on the cluster-mgmt page, since it affects all collections on the cluster. (It's a bit like `BALANCESHARDUNIQUE` that way) But I see the argument the other way too. 🤷‍♂️

Interesting. I put it with the collection management commands because it seemed it only works on a single collection (it has a required `collection` param to define the collection name), and the example just shows it working on a single collection.

Do you know more about how it works on all collections?

gerlowskija:
> Do you know more about how it works ...

No.  In fact, apparently I know less.  Not sure what I was thinking.  Looking closer at the docs, and trying it out, it doesn't work the way I thought it did.  You're right that it only works on a single collection.  Ignore my original comment, sorry.

ctargett:
This branch has significantly diverged from master, and I spent a little bit of time trying to straighten it out but it would take more time than I think it's worth.

Since this was reviewed and got a +1, instead of trying to update the branch and update the PR for another round, I'm going to close this PR and just make my changes locally to master and branch_8x and push them up that way.

PR #576, opened 2019-02-16
LUCENE-8631: Longest-Matching for User words in Nori Tokenizer

gritmind:
The issue is described in [LUCENE-8631](https://issues.apache.org/jira/browse/LUCENE-8631). 

The existing nori tokenizer is able to do Longest-Matching for system words, but not for user words. The reason why it doesn’t work for user words is because their large costs severely mitigate bigram costs. Longest-Matching is a natural behavior in that system words, whose costs were generated from large Korean corpus, can be applied. Also, it is needed when to search exact words. So, we want it also apply to user words.

Thankfully, Jim Ferenczi, who is a nori developer, gave to me some suggestions to fix it (ref. [LUCENE-8631](https://issues.apache.org/jira/browse/LUCENE-8631)). I followed the “Implement longest-match first in the dictionary” because it is simplest and efficient way. As Jim mentioned, it provides not only Longest-Matching, but also speed-up because it retains a single path for user words.

The idea is, whenever to iterate the input stream by character step, the longest word is only stored and its subset words are ignored. There are two steps to implement this. First, before break in the for-loop, we call at once add function with posAheadMax info. Second, we remember global_posAheadMax during while-loop and use it to avoid  storing subset words at different positions. 

For checking how robust is the implementation, we tested using about 100,000 Korean words, where subset words are included, and all test cases were passed. 

Yeongsu Kim
high_yeongsu@wemakeprice.com
Wemakeprice Inc., South Korea
https://www.wemakeprice.com


lucene/analysis/nori/src/java/org/apache/lucene/analysis/ko/KoreanTokenizer.java
```diff
@@ -514,6 +522,9 @@ private void parse() throws IOException {
     // Index of the last character of unknown word:
     int unknownWordEndIndex = -1;
 
+    /// Largest global posAhead of user word
+    int global_posAheadMax = -1;
```
jimczi: Why is it needed ?

lucene/analysis/nori/src/java/org/apache/lucene/analysis/ko/KoreanTokenizer.java
```diff
@@ -651,20 +662,37 @@ private void parse() throws IOException {
...
         for(int posAhead=pos;;posAhead++) {
           final int ch = buffer.get(posAhead);
-          if (ch == -1) {
-            break;
-          }
-          if (userFST.findTargetArc(ch, arc, arc, posAhead == pos, userFSTReader) == null) {
-            break;
+          if (ch == -1 || userFST.findTargetArc(ch, arc, arc, posAhead == pos, userFSTReader) == null) {
+            if (anyMatches
+              && posAheadMax > global_posAheadMax) {
```
jimczi: What are you trying to avoid with this check ? We should always add the longest dictionary term, no ?

jimczi: I also think it would be more readable if you move this code after the for loop with a comment explaining what it does (e.g.: add the longest matching user term)

lucene/analysis/nori/src/java/org/apache/lucene/analysis/ko/KoreanTokenizer.java
gritmind: Please refer to the below (1) comment.

lucene/analysis/nori/src/java/org/apache/lucene/analysis/ko/KoreanTokenizer.java
gritmind: First, I really appreciate your feedback (@jimczi). I updated my code based on your comments. [Here](https://github.com/apache/lucene-solr/pull/576/commits/0e8dd66a521eb479dca79a98c1c8d321e24addbe) is new version.

## 1.

Why do we need a variable,`global_posAheadMax` and a condition `posAheadMax > global_posAheadMax)`?

Without them, the longest matching might not work, if the first characters of the query are alphabetic and/or numeric and the rest are more than two subset user words.

Let me explain with a concrete example.
Let assume we have user words such as {'대한민국', '날씨', '대한민국날씨', '21세기대한민국', '세기'}. When the query is '대한민국날씨', then the analysis result is '대한민국날씨'. Longest matching is successfully applied to this case. However, when the query is '21세기대한민국', the analysis result is ['21', '세기', '대한민국']. Unfortunately, the longest matching is not working to this case. 

Why?

Numeric and alphabetic characters (rather than hangul characters) always become tokens as unknown words (,which can be a part of a viterbi search path) even though they are also in user words and/or known words. (It might be related to the condition, `characterDefinition.isInvoke(firstCharacter)`).

Given this fact, '21' in '21세기대한민국' can be a token, and token combination ['21', '세기', '대한민국'] is generated. Since there are two user words such as '세기', and '대한민국', the cost of this token combination is far lower than the cost of longest token, ['21세기대한민국']. (the cost of each user is -100,000). So, ['21', '세기', '대한민국'] win against ['21세기대한민국'].

To avoid this, we ensure that after a user word (e.g. '21세기대한민국') is added, its subset words (e.g. '세기', '대한민국') could not be added. To implement this, always remember the maximum index of added user words via `global_posAheadMax`, and use the condition `posAheadMax > global_posAheadMax)` for avoiding adding subset words.

## 2.

As you mentioned, I updated some codes for readability. 
* re-naming variables based on java style
* moving the code after the for loop with a longest matching comment
* replacing 'max_int' function with  'Math.max' function

For checking the updated code, we again tested it with 100,000 Korean word cases, where subset words are included, and all test cases were passed.

## 3.

I added two unit test cases for longest matching in `TestKoreanTokenizer`. (Also, added some user words in `userdict.txt`.)

Thank you :)

lucene/analysis/nori/src/java/org/apache/lucene/analysis/ko/KoreanTokenizer.java
jimczi: Got it, thanks for explaining. Can you rename with camel case for consistency: `userWordMaxPosAhead` ?

lucene/analysis/nori/src/java/org/apache/lucene/analysis/ko/KoreanTokenizer.java
jimczi: Got it, thanks for explaining

gritmind:
I renamed `globalMaxPosAhead ` to `userWordMaxPosAhead`. It looks better. Thank you @jimczi.

lucene/analysis/nori/src/java/org/apache/lucene/analysis/ko/KoreanTokenizer.java
gritmind: Yeap!

jimczi:
I merged this change in master and 8x: https://gitbox.apache.org/repos/asf?p=lucene-solr.git;a=commit;h=b1f870a4164769df62b24af63048aa2f9b21af47. Thanks @gritmind!

PR #579, opened 2019-02-19
LUCENE-8681: prorated early termination

msokolov:
This adds support for prorated early termination in TopFieldsCollector, a new unit test that demonstrates that, and a convenience method for creating a CollectorManager that uses that. It also refactors IndexSearcher to use the new convenience method and cleans up some leftover javadocs about a search parameter that was removed. See LUCENE-8681

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/IndexSearcher.java
+++ b/lucene/core/src/java/org/apache/lucene/search/IndexSearcher.java
@@ -94,11 +94,6 @@ public class IndexSearcher {
     final long maxRamBytesUsed = Math.min(1L << 25, Runtime.getRuntime().maxMemory() / 20);
     DEFAULT_QUERY_CACHE = new LRUQueryCache(maxCachedQueries, maxRamBytesUsed);
   }
-  /**
-   * By default we count hits accurately up to 1000. This makes sure that we
-   * don't spend most time on computing hit counts
-   */
-  private static final int TOTAL_HITS_THRESHOLD = 1000;
 
   final IndexReader reader; // package private for testing!
   
@@ -401,7 +396,7 @@ public class IndexSearcher {
 
       @Override
       public TopScoreDocCollector newCollector() throws IOException {
-        return TopScoreDocCollector.create(cappedNumHits, after, TOTAL_HITS_THRESHOLD);
+        return TopScoreDocCollector.create(cappedNumHits, after, TerminationStrategy.HIT_COUNT);
       }
 
       @Override
@@ -518,7 +513,7 @@ public class IndexSearcher {
     final Sort rewrittenSort = sort.rewrite(this);
 
     final CollectorManager<TopFieldCollector, TopFieldDocs> manager =
-        TopFieldCollector.createManager(rewrittenSort, cappedNumHits, after, TOTAL_HITS_THRESHOLD, Integer.MAX_VALUE);
+        TopFieldCollector.createManager(rewrittenSort, cappedNumHits, after, TerminationStrategy.HIT_COUNT);
 
     TopFieldDocs topDocs = search(query, manager);
     if (doDocScores) {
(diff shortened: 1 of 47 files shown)


lucene/core/src/java/org/apache/lucene/search/TopFieldCollector.java
```diff
@@ -165,11 +169,35 @@ public void collect(int doc) throws IOException {
...
+      return totalHits;
+    }
+
+    private int prorateForSegment(int topK, LeafReaderContext leafCtx) {
+        // prorate number of hits to collect based on proportion of documents in this leaf (segment).
+        // p := probability of a top-k document (or any document) being in this segment
+        double p = (double) leafCtx.reader().numDocs() / leafCtx.parent.reader().numDocs();
+        // m := expected number of the topK results in this segment
+        double m = p * topK;
+        // Increase N to include a bound to ensure the probability of missing a doc is very small
```
mikemccand: Maybe add a short description describing the math / distribution here, explaining the equation below?

lucene/core/src/test/org/apache/lucene/search/TestTopFieldCollectorEarlyTermination.java
```diff
@@ -234,13 +242,189 @@ public void testCanEarlyTerminateOnPrefix() {
...
+      writer.addDocument(numberedDocument(i, i));
+      if (random().nextInt(numDocs / 10) == 0) {
+        // Make about 10 random-sized segments
+        writer.commit();
+      }
+    }
+    reader = DirectoryReader.open(writer);
+  }
+
+  public void testProratedEarlyTermination() throws IOException {
```
mikemccand: Do we explicitly test passing `Integer.MAX_VALUE` for `perSegmentMargin` somewhere here?

lucene/core/src/java/org/apache/lucene/search/TopFieldCollector.java
```diff
@@ -165,11 +169,35 @@ public void collect(int doc) throws IOException {
...
+    /** The total number of documents that matched this query; may be a lower bound in case of early termination. */
+    @Override
+    public int getTotalHits() {
+      return totalHits;
+    }
+
+    private int prorateForSegment(int topK, LeafReaderContext leafCtx) {
+        // prorate number of hits to collect based on proportion of documents in this leaf (segment).
+        // p := probability of a top-k document (or any document) being in this segment
+        double p = (double) leafCtx.reader().numDocs() / leafCtx.parent.reader().numDocs();
```
mikemccand: Do we need to guard for divide-by-zero here?

lucene/core/src/test/org/apache/lucene/search/TestTopFieldCollectorEarlyTermination.java
msokolov: hmm, explicitly .. no. It's the default now, so tested implicitly in many cases, but an explicit test would be good to demonstrate all still works OK even when index distribution is skewed: I'll add that.

lucene/core/src/java/org/apache/lucene/search/TopFieldCollector.java
msokolov: Oh yes we do. I guess that would be an empty index. This also made me thing about deleted docs. It would be better to compute this ratio using livedocs I think? Basically we want to use numbers that correspond to what will be collected. Can we easily know the number of live docs in a segment and in the index?

lucene/core/src/java/org/apache/lucene/search/IndexSearcher.java
```diff
@@ -525,25 +517,8 @@ private TopFieldDocs searchAfter(FieldDoc after, Query query, int numHits, Sort
...
-        int i = 0;
-        for (TopFieldCollector collector : collectors) {
-          topDocs[i++] = collector.topDocs();
-        }
-        return TopDocs.merge(rewrittenSort, 0, cappedNumHits, topDocs, true);
-      }
-
-    };
+    final CollectorManager<TopFieldCollector, TopFieldDocs> manager =
+        TopFieldCollector.createManager(rewrittenSort, cappedNumHits, after, TOTAL_HITS_THRESHOLD, Integer.MAX_VALUE);
```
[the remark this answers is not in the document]
msokolov: Yes, except for plumbing through the new `perSegmentMargin` parameter. I just moved the code so I could re-use.

lucene/core/src/java/org/apache/lucene/search/TopFieldCollector.java
```diff
@@ -100,8 +101,8 @@ private static boolean canEarlyTerminateOnPrefix(Sort searchSort, Sort indexSort
     final Sort sort;
     final FieldValueHitQueue<Entry> queue;
 
-    public SimpleFieldCollector(Sort sort, FieldValueHitQueue<Entry> queue, int numHits, int totalHitsThreshold) {
-      super(queue, numHits, totalHitsThreshold, sort.needsScores());
+    public SimpleFieldCollector(Sort sort, FieldValueHitQueue<Entry> queue, int numHits, int totalHitsThreshold, int perSegmentMargin) {
```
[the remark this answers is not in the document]
msokolov: This whole class is private; I don't think its javadocs would be visible outside this file.

lucene/core/src/java/org/apache/lucene/search/TopFieldCollector.java
mikemccand: Ahh yes we should use `numDocs` -- you can get this from `leafCtx.reader().numDocs()`.

msokolov: I added divide-by-zero check, and I think we were always using `numDocs()`, so this should be good.

PR #580, opened 2019-02-19
LUCENE-8700: IndexWriter.yield()

msokolov:
This provides a new `IndexWriter.yield()` method, and adds testing for it to `RandomIndexWriter` which will now use `yield()` to flush concurrently during `commit()`. I also added a flag for disabling this behavior. I'm not sure if it's needed though, since no tests failed I pulled out the flushing logic from `DocumentupdateDocument` into a standalone method `flushPending` - I initially wanted to refactor and share but the code doesn't lend itself to that without adding arguments to these methods, and it wasn't clear the result would be better.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/DocumentsWriter.java
+++ b/lucene/core/src/java/org/apache/lucene/index/DocumentsWriter.java
@@ -473,6 +473,23 @@ final class DocumentsWriter implements Closeable, Accountable {
     return seqNo;
   }
 
+  void flushPending() throws IOException {
+    boolean hasEvents = preUpdate();
+    final ThreadState perThread = flushControl.obtainAndLock();
+    final DocumentsWriterPerThread flushingDWPT;
+    try {
+      // This must happen after we've pulled the ThreadState because IW.close
+      // waits for all ThreadStates to be released:
+      ensureOpen();
+      ensureInitialized(perThread);
+      assert perThread.isInitialized();
+      flushingDWPT = flushControl.checkout(perThread, false);
+    } finally {
+      perThreadPool.release(perThread);
+    }
+    postUpdate(flushingDWPT, hasEvents);
+  }
+
   long updateDocument(final Iterable<? extends IndexableField> doc, final Analyzer analyzer,
                       final DocumentsWriterDeleteQueue.Node<?> delNode) throws IOException {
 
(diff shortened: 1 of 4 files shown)


s1monw:
@msokolov can you explain more what the usecase of this method is. I am not sure I am following the logic entirely and I need to better understand what this is about to review it.

msokolov:
HI @s1monw there was some context in the mailing list, sorry I did not copy here. The use case here is batch indexing followed by a commit. When one has lots of RAM and no need to commit frequently for visibility (eg offline testing, or some nightly background process), it may be more performant to index without committing, accumulating fairly large in-memory buffers before flushing, and then at the end of a large batch, commit(). The problem with this case today is that we do not make use of any parallel I/O when flushing unless calls are being made to `updateDocument()` and the like. I considered calling that with an empty document, but I was concerned that it might have some undesirable side effects?

s1monw:
Would `IndexWriter#flushNextBuffer()` help here?

msokolov:
In our specific case, we create a small index ~3G for some integration tests that can fit entirely in RAM, and we index all the documents, using large buffers with IndexWriter, using multiple indexing threads, and only when we are done do we commit. Adding this, together with some logic in the caller to supply the threads, reduced our time to commit from ~10 minutes to ~5 minutes on one host (we are sorting the index, so it takes a while), and the situation is even more embarassing on hosts with more cores (where we use more indexing threads).

msokolov:
bq. Would `IndexWriter#flushNextBuffer()` help here?

oh -- comments flying past each other -- I'll check; I did not know about that

msokolov:
Oh, yes - that looks like it would handle this case. Maybe we do not need this `yield()` method then. I'll give it a try - thanks!

s1monw:
> Oh, yes - that looks like it would handle this case. Maybe we do not need this `yield()` method then. I'll give it a try - thanks!

give it a go, if it's tricky we can still extend it.

msokolov:
OK, so `IndexWriter#flushNextBuffer()` seems to do just what we need; I get basically the same behavior by just replacing `yield()` with that. The other notable thing in this patch is having `RandomIndexWriter.commit()` sometimes call that during tests. I see there are some explicit tests of `flushNextBuffer` in `IndexWriterTest`, but I wonder if you see any value in adding this concurrent flushing to lots more tests?

mikemccand:
> but I wonder if you see any value in adding this concurrent flushing to lots more tests?

+1

This might help uncover concurrency bugs in `IndexWriter`.

s1monw:
superseded by #607 @mikemccand can you take a look

PR #581, opened 2019-02-20
LUCENE-3041: QueryVisitor

romseygeek:
This commit adds a visit() method to Query, allowing you to traverse query trees and analyse
their structure and the terms they use for matching.


lucene/core/src/java/org/apache/lucene/search/QueryVisitor.java
```diff
@@ -0,0 +1,123 @@
...
+ * @see Query#visit(QueryVisitor)
+ */
+public interface QueryVisitor {
+
+  /**
+   * Called by leaf queries that match on a specific term
+   * @param query the leaf query visited
+   * @param term  the term the query will match on
+   */
+  void visitLeaf(Query query, Term term);
```
s1monw: I do wonder if we should make it simpler and pass `String field, BytesRef value` instead of term. I am ok with keeping it this way.

lucene/core/src/java/org/apache/lucene/search/QueryVisitor.java
```diff
@@ -0,0 +1,123 @@
...
+   */
+  default void visitLeaf(Query query) {}
+
+  /**
+   * Called by leaf queries that match against a set of terms defined by a predicate
+   * @param query               the leaf query
+   * @param field               the field the query matches against
+   * @param predicateSupplier   a supplier for a predicate that will select matching terms
+   */
+  default void visitLeaf(Query query, String field, Supplier<Predicate<BytesRef>> predicateSupplier) {}
```
s1monw: I find this generalization odd. I think we should not add this without a user of this API. This is a general comment on this PR. It's hard to tell if it's sufficient if we don't have a clear user of the API? Should we move this to a branch and then iterate on it also with a user of it. ie. deprecate extractTerms and implement it with this?

lucene/core/src/java/org/apache/lucene/search/TermInSetQuery.java
```diff
@@ -122,6 +123,18 @@ public Query rewrite(IndexReader reader) throws IOException {
     return super.rewrite(reader);
   }
 
+  @Override
+  public void visit(QueryVisitor visitor) {
+    visitor.visitLeaf(this, field, () -> {
```
s1monw: why isn't this just calling the visitor with the individual terms?

romseygeek:
Replacing extractTerms() with this is a good idea.  We can iterate on this PR for the time being, I think?

romseygeek:
I've pushed some changes:
* removed the Supplier generalization; this was intended to help with highlighting AutomatonQuery implementations, but let's deal with that when we need to
* replaced the usage of extractTerms() with appropriate visitors - the nice thing here is that because most of the methods on QueryVisitor have default implementations, you can just do `Query.visit(terms::add)` as a replacement; but you can also see immediate improvements in things like QueryTermExtractor in highlighter, which no longer needs its big tree of instanceof checks

lucene/core/src/java/org/apache/lucene/search/QueryVisitor.java
```diff
@@ -0,0 +1,91 @@
...
+package org.apache.lucene.search;
+
+import org.apache.lucene.index.Term;
+
+/**
+ * Interface to allow recursion through a query tree
+ *
+ * @see Query#visit(QueryVisitor)
+ */
+public interface QueryVisitor {
```
s1monw: I think this should have `@FunctionalInterface` annotation or alternatively `extends Consumer<Term>`?

lucene/core/src/java/org/apache/lucene/search/QueryVisitor.java
```diff
@@ -0,0 +1,91 @@
...
+  default void visitLeaf(Query query) {}
+
+  /**
+   * Pulls a visitor instance for visiting matching child clauses of a query
+   *
+   * The default implementation returns {@code this}
+   *
+   * @param parent  the query visited
+   */
+  default QueryVisitor getMatchingVisitor(Query parent) {
```
s1monw: this corresponds to `BooleanClause.Occur` maybe we should just have a method `default void visitLeaf(Query query, BooleanClause.Occur occur) {}` instead of 4 default methods?

s1monw: oh yeah and a `matchesTerm(Term, Occur)` as well.

s1monw: now that I looked further down I think I see what you are trying to do here. I wonder if it makes sense to reduce to a single `getSubVisitor(BooleanClause.Occur occur, Query parent)` this seems more straight forward?

lucene/core/src/java/org/apache/lucene/search/QueryVisitor.java
```diff
@@ -0,0 +1,91 @@
...
+   * @param parent  the query visited
+   */
+  default QueryVisitor getNonMatchingVisitor(Query parent) {
+    return NO_OP;
+  }
+
+  /**
+   * A QueryVisitor implementation that collects no terms
+   */
+  QueryVisitor NO_OP = term -> { };
```
s1monw: not sure we need this, can we just inline `term -> {}`

lucene/sandbox/src/java/org/apache/lucene/search/intervals/IntervalQuery.java
```diff
@@ -147,7 +153,7 @@ public IntervalWeight(Query query, float boost, ScoreMode scoreMode) {
 
     @Override
     public void extractTerms(Set<Term> terms) {
```
s1monw: should this method be removed too?

lucene/core/src/java/org/apache/lucene/search/QueryVisitor.java
romseygeek: I've removed it

lucene/core/src/java/org/apache/lucene/search/QueryVisitor.java
romseygeek: ++, have pushed a change to implement this

lucene/sandbox/src/java/org/apache/lucene/search/intervals/IntervalQuery.java
romseygeek: Let's make removing extractTerms() completely a separate issue?

lucene/core/src/java/org/apache/lucene/search/QueryVisitor.java
romseygeek: I've made it a functional interface, but I'm now wondering if this would be better done as an abstract class with all default implementations?  For example, if we want to add leaf methods that accept Automata, or doc values ranges, or points, or similar - it's targeted at terms currently, but that seems overly restrictive.  Plus an abstract class allows you to call super methods, whereas now you always have to remember to deal with eg `MUST_NOT` clauses if you override `getSubVisitor()`

romseygeek:
Pushed a change that tries moving things to an abstract base class, and adds a helper method for the common case where we just want to collect terms (ie an extractTerms() replacement)

romseygeek:
I made a couple more changes:
* Rather than having a `NO_OP` visitor, I changed the contract so that `getSubVisitor()` returns `null` if queries should not continue visiting.  This enables us to shortcut traversal, for example if we're checking for the presence of a particular query type in a tree then we can stop as soon as we've found a single instance.  I also added a method to `QueryUtils` to check that queries do the right thing when `getSubVisitor()` returns `null`.
* I added a `matchesAutomaton` method back in, and refactored `MultiTermHighlighting` in the unified highlighter to use it - it's much cleaner and easier to follow now, and makes it much easier to handle custom queries.

lucene/core/src/test/org/apache/lucene/search/TestQueryVisitor.java
```diff
@@ -0,0 +1,226 @@
...
+import org.apache.lucene.search.spans.SpanNearQuery;
+import org.apache.lucene.search.spans.SpanQuery;
+import org.apache.lucene.search.spans.SpanTermQuery;
+import org.apache.lucene.util.LuceneTestCase;
+
+import static org.hamcrest.CoreMatchers.equalTo;
+
+public class TestQueryVisitor extends LuceneTestCase {
+
+  private static final Query query = new BooleanQuery.Builder()
```
dsmiley: Can you show a BoostQuery wrapping a BooleanQuery that has a BoostQuery around a TermQuery :-).  Your test below would then be able to show you can do the inherited multiplication.

lucene/core/src/java/org/apache/lucene/search/QueryVisitor.java
```diff
@@ -0,0 +1,87 @@
...
+  public void matchesTerm(Term term) { }
+
+  /**
+   * Called by leaf queries that match terms against an Automaton
+   * @param query               the leaf query
+   * @param field               the field to match against
+   * @param isBinary            {@code true} if the Automaton is built to match against byte[] rather than char[]
+   * @param automatonSupplier   a supplier for the built Automaton
+   */
+  public void matchesAutomaton(Query query, String field, boolean isBinary, Supplier<Automaton> automatonSupplier) {}
```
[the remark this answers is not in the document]
romseygeek: The whole Automaton API is still kind of trappy, but I think that's a larger issue.  I think the flag is the least invasive way of doing things for now.

lucene/core/src/java/org/apache/lucene/search/TermInSetQuery.java
```diff
@@ -135,6 +138,20 @@ private boolean equalsTo(TermInSetQuery other) {
         termData.equals(other.termData);
   }
 
+  @Override
+  public void visit(QueryVisitor visitor) {
+    visitor.matchesAutomaton(this, field, true, this::toAutomaton);
```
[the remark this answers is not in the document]
romseygeek: ++, will add a comment

lucene/highlighter/src/java/org/apache/lucene/search/uhighlight/MultiTermHighlighting.java
```diff
@@ -17,180 +17,124 @@
...
-class MultiTermHighlighting {
+final class MultiTermHighlighting {
   private MultiTermHighlighting() {
   }
 
   /**
    * Extracts MultiTermQueries that match the provided field predicate.
    * Returns equivalent automata that will match terms.
    */
-  public static CharacterRunAutomaton[] extractAutomata(Query query,
```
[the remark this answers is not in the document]
romseygeek: It's still used in PhraseHelper; I think we can rework UnifiedHighlighter in a follow-up, along with things like removing extractWeight(), given that this PR is already quite big?

lucene/core/src/test/org/apache/lucene/search/TestQueryVisitor.java
romseygeek: ++

lucene/core/src/java/org/apache/lucene/search/TermInSetQuery.java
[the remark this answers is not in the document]
romseygeek: This should probably use a DaciukMihov automaton, as it's just fixed strings.

romseygeek:
Thanks for the review @dsmiley!  I pushed some changes in response.

lucene/core/src/java/org/apache/lucene/search/QueryVisitor.java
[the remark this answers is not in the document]
s1monw: I really dont' like that this name implies a question and then it doesn't have a return type. I think we should rather call it `consumeTerm` / `consumeAutomaton` instead. I agree with @dsmiley if we can put this boolean on the Automaton it would be great. There is also the issue of the Supplier which I think is a design smell I am really torn on this method to be honest

lucene/core/src/java/org/apache/lucene/search/PhraseQuery.java
```diff
@@ -284,6 +284,16 @@ public Query rewrite(IndexReader reader) throws IOException {
     }
   }
 
+  @Override
+  public void visit(QueryVisitor visitor) {
+    for (Term term : terms) {
+      QueryVisitor v = visitor.getSubVisitor(BooleanClause.Occur.MUST, this);
```
s1monw: don't pull this more than once?

lucene/core/src/java/org/apache/lucene/search/QueryVisitor.java
```diff
@@ -0,0 +1,87 @@
...
+   * Returns {@code null} if the child clauses should not be visited
+   *
+   * The default implementation returns {@code this}, unless {@code occur} is equal
+   * to {@link BooleanClause.Occur#MUST_NOT} in which case it returns
+   * {@code null}
+   *
+   * @param occur   the relationship between the parent and its children
+   * @param parent  the query visited
+   */
+  public QueryVisitor getSubVisitor(BooleanClause.Occur occur, Query parent) {
```
s1monw: not sure I like the null here. if we keep this logic I think it should return an `Optional<QueryVisitor>` or always return non-null

lucene/core/src/java/org/apache/lucene/search/MultiPhraseQuery.java
```diff
@@ -203,6 +203,22 @@ public Query rewrite(IndexReader reader) throws IOException {
...
 
+  @Override
+  public void visit(QueryVisitor visitor) {
+    QueryVisitor v = visitor.getSubVisitor(BooleanClause.Occur.MUST, this);
+    if (v == null) {
+      return;
+    }
+    for (Term[] terms : termArrays) {
+      for (Term term : terms) {
+        QueryVisitor sv = v.getSubVisitor(BooleanClause.Occur.SHOULD, this);
```
s1monw: why do you pull this more than once?

lucene/core/src/java/org/apache/lucene/search/DisjunctionMaxQuery.java
```diff
@@ -237,6 +237,16 @@ public Query rewrite(IndexReader reader) throws IOException {
     return super.rewrite(reader);
   }
 
+  @Override
+  public void visit(QueryVisitor visitor) {
+    for (Query q : disjuncts) {
+      QueryVisitor v = visitor.getSubVisitor(BooleanClause.Occur.SHOULD, this);
```
s1monw: same here why pull more than once?

lucene/queries/src/java/org/apache/lucene/queries/CommonTermsQuery.java
```diff
@@ -129,7 +130,18 @@ public Query rewrite(IndexReader reader) throws IOException {
     collectTermStates(reader, leaves, contextArray, queryTerms);
     return buildQuery(maxDoc, contextArray, queryTerms);
   }
-  
+
+  @Override
+  public void visit(QueryVisitor visitor) {
+    for (Term term : terms) {
+      QueryVisitor v = visitor.getSubVisitor(Occur.SHOULD, this);
```
s1monw: again why pull more than once?

lucene/sandbox/src/java/org/apache/lucene/search/CoveringQuery.java
```diff
@@ -109,6 +109,17 @@ public Query rewrite(IndexReader reader) throws IOException {
     return super.rewrite(reader);
   }
 
+  @Override
+  public void visit(QueryVisitor visitor) {
+    for (Query query : queries) {
+      QueryVisitor v = visitor.getSubVisitor(BooleanClause.Occur.SHOULD, this);
```
s1monw: pulled more than once?

lucene/sandbox/src/java/org/apache/lucene/search/TermAutomatonQuery.java
```diff
@@ -492,4 +492,15 @@ public Query rewrite(IndexReader reader) throws IOException {
     // TODO: we could maybe also rewrite to union of PhraseQuery (pull all finite strings) if it's "worth it"?
     return this;
   }
+
+  @Override
+  public void visit(QueryVisitor visitor) {
+    for (BytesRef term : termToID.keySet()) {
+      QueryVisitor v = visitor.getSubVisitor(BooleanClause.Occur.SHOULD, this);
```
s1monw: pulled more than once?

lucene/sandbox/src/java/org/apache/lucene/search/intervals/IntervalQuery.java
s1monw: I agree if we can get rid of it, lets do it. Not sure what purpose it serves still here. If we can't can we deprecate?

lucene/sandbox/src/java/org/apache/lucene/search/intervals/MinimumShouldMatchIntervalsSource.java
```diff
@@ -85,9 +85,14 @@ public MatchesIterator matches(String field, LeafReaderContext ctx, int doc) thr
   }
 
   @Override
-  public void extractTerms(String field, Set<Term> terms) {
+  public void visit(String field, QueryVisitor visitor) {
+    Query parent = new IntervalQuery(field, this);
     for (IntervalsSource source : sources) {
-      source.extractTerms(field, terms);
+      QueryVisitor v = visitor.getSubVisitor(BooleanClause.Occur.SHOULD, parent);
```
s1monw: pull only once?

lucene/core/src/java/org/apache/lucene/search/QueryVisitor.java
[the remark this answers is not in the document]
s1monw: here is a different suggestion: maybe we can make `FuzzyQuery` subclass `AutomatonQuery` and then simply add a method `visitAutomaton(AutomatonQuery query) { visiteLeaf(query); }` then we can remove this method an users can just pull the automaton, field and if it's binary from `AutomatonQuery`.  This would be a much cleaner API. We don't even have to add an API for this (there is only one users) we can just do an `instanceof` check?

lucene/sandbox/src/java/org/apache/lucene/search/intervals/IntervalQuery.java
romseygeek: +1 to deprecate and remove in a followup

lucene/core/src/java/org/apache/lucene/search/QueryVisitor.java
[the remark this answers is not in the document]
romseygeek: ++

lucene/core/src/java/org/apache/lucene/search/PhraseQuery.java
romseygeek: The advantage of pulling multiple times is the combination with the null check, meaning we can short-circuit; it's particularly useful on something like BooleanQuery or DisjunctionQuery, where if you're say looking to see if there's a single instance of a particular term or query type, you can stop looking as soon as you find it rather than having to continue traversing through all the possibilities

lucene/core/src/java/org/apache/lucene/search/QueryVisitor.java
romseygeek: I'm not sure I like `Optional`...  We use `null` elsewhere in the codebase to signal "stop doing stuff here", for example in `Weight.scorer()`, so it seems an established pattern to me?

lucene/core/src/java/org/apache/lucene/search/PhraseQuery.java
s1monw: [I think you need to read this. Maybe twice.](https://en.wikipedia.org/wiki/Program_optimization)

romseygeek: I took a shortcut :p

More seriously, I came across issues like this in Luwak when dealing with super-massive queries, but I suppose there are other ways around it, like throwing an exception or whatever.  I can change it back to returning an empty visitor if you really don't like it.

s1monw: you can enforce this by adding a boolean return to `consumeTerm` that we can respect. that would be a clear API to short-circuit.

romseygeek:
* I've taken Simon's suggestion of removing special-casing for Automaton, and things still seem to work (although we may need to add hooks back in to allow handling of custom multi-term queries).
* I've renamed matchesTerm() to consumesTerm(), and changed the sig to include the Query doing the consuming (otherwise we can't tell if a leaf query is a TermQuery or something else).
* QueryVisitor.EMPTY_VISITOR is back, nulls are gone
* I've also removed Weight.extractTerms() entirely - I can revert this last commit if it makes the patch too big, but it does increase the number of lines removed by a lot, which is nice :)

lucene/core/src/java/org/apache/lucene/search/PhraseQuery.java
javanna: I agree with Simon. I find it counter intuitive to get the sub-visitor for the same query more than once. if we do need that short-circuiting maybe it can be made more explicit in the API.

Probably because I am not yet very familiar with the change here, but I have a hard time following where things happen here: we pull the sub-visitor from the same parent visitor multiple times, yet what gets returned may be null at some point. Does that mean that the parent visitor holds state that changes as we consume terms through its sub-visitors?

lucene/core/src/java/org/apache/lucene/search/BoostQuery.java
```diff
@@ -104,6 +104,14 @@ public Query rewrite(IndexReader reader) throws IOException {
     return super.rewrite(reader);
   }
 
+  @Override
+  public void visit(QueryVisitor visitor) {
+    QueryVisitor v = visitor.getSubVisitor(BooleanClause.Occur.MUST, this);
+    if (v != null) {
```
javanna: should you rather check for empty visitor here now? I guess we can assume that getSubVisitor never returns null?

lucene/core/src/java/org/apache/lucene/search/ConstantScoreQuery.java
```diff
@@ -64,6 +64,14 @@ public Query rewrite(IndexReader reader) throws IOException {
     return super.rewrite(reader);
   }
 
+  @Override
+  public void visit(QueryVisitor visitor) {
+    QueryVisitor v = visitor.getSubVisitor(BooleanClause.Occur.FILTER, this);
+    if (v != null) {
```
javanna: same as above, can this be null now?

lucene/core/src/java/org/apache/lucene/search/IndexOrDocValuesQuery.java
```diff
@@ -109,16 +107,20 @@ public Query rewrite(IndexReader reader) throws IOException {
     return this;
   }
 
+  @Override
+  public void visit(QueryVisitor visitor) {
+    QueryVisitor v = visitor.getSubVisitor(BooleanClause.Occur.MUST, this);
+    if (v != null) {
```
javanna: can this be null?

lucene/core/src/java/org/apache/lucene/search/MultiTermQueryConstantScoreWrapper.java
```diff
@@ -231,4 +231,12 @@ public boolean isCacheable(LeafReaderContext ctx) {
 
     };
   }
+
+  @Override
+  public void visit(QueryVisitor visitor) {
+    QueryVisitor v = visitor.getSubVisitor(Occur.FILTER, this);
+    if (v != null) {
```
javanna: redundant null check?

lucene/core/src/java/org/apache/lucene/search/NGramPhraseQuery.java
```diff
@@ -77,6 +77,14 @@ public Query rewrite(IndexReader reader) throws IOException {
     return builder.build();
   }
 
+  @Override
+  public void visit(QueryVisitor visitor) {
+    QueryVisitor v = visitor.getSubVisitor(BooleanClause.Occur.MUST, this);
+    if (v != null) {
+      phraseQuery.visit(v);
+    }
```
javanna: redundant null check?

lucene/core/src/java/org/apache/lucene/search/QueryVisitor.java
```diff
@@ -0,0 +1,80 @@
...
+      public void consumesTerm(Query query, Term term) {
+        termSet.add(term);
+      }
+    };
+  }
+
+  /**
+   * A QueryVisitor implementation that does nothing
+   */
+  public static QueryVisitor EMPTY_VISITOR = new QueryVisitor() {};
```
javanna: this should be final?

lucene/core/src/java/org/apache/lucene/search/spans/FieldMaskingSpanQuery.java
```diff
@@ -104,6 +106,14 @@ public Query rewrite(IndexReader reader) throws IOException {
     return super.rewrite(reader);
   }
 
+  @Override
+  public void visit(QueryVisitor visitor) {
+    QueryVisitor v = visitor.getSubVisitor(BooleanClause.Occur.MUST, this);
+    if (v != null) {
```
javanna: redundant null check?

lucene/core/src/java/org/apache/lucene/search/QueryVisitor.java
```diff
@@ -0,0 +1,80 @@
...
+ */
+public abstract class QueryVisitor {
+
+  /**
+   * Called by leaf queries that match on a specific term
+   *
+   * @param query the leaf query
+   * @param term  the term the query will match on
+   */
+  public void consumesTerm(Query query, Term term) { }
```
javanna: consume or consumes ?

lucene/core/src/java/org/apache/lucene/search/FuzzyQuery.java
```diff
@@ -146,6 +147,10 @@ public boolean getTranspositions() {
     return transpositions;
   }
 
+  public Automaton getAutomaton() {
```
dsmiley: The name of this method is totally trappy.  We are in fact *building* an automaton... or if you prefer toAutomaton.

dsmiley:
I can understand that the default implementation of Query.visit is to call visitLeaf on itself for the benefit of backwards compatibility.  That said, I think it's trappy for all the custom Query's out there.  On master branch it ought to be an abstract method.

lucene/core/src/java/org/apache/lucene/search/DisjunctionMaxQuery.java
```diff
@@ -237,6 +228,16 @@ public Query rewrite(IndexReader reader) throws IOException {
...
+      if (v != null) {
```
javanna: you can remove the null check?

lucene/core/src/java/org/apache/lucene/search/spans/SpanBoostQuery.java
```diff
@@ -93,6 +95,14 @@ public Query rewrite(IndexReader reader) throws IOException {
     return super.rewrite(reader);
   }
 
+  @Override
+  public void visit(QueryVisitor visitor) {
+    QueryVisitor v = visitor.getSubVisitor(BooleanClause.Occur.MUST, this);
+    if (v != null) {
```
javanna: remove the null check?

lucene/core/src/java/org/apache/lucene/search/spans/SpanContainQuery.java
```diff
@@ -128,6 +120,18 @@ public Query rewrite(IndexReader reader) throws IOException {
...
   }
 
+  @Override
+  public void visit(QueryVisitor visitor) {
+    QueryVisitor v1 = visitor.getSubVisitor(BooleanClause.Occur.MUST, this);
+    if (v1 != null) {
+      big.visit(v1);
+    }
+    QueryVisitor v2 = visitor.getSubVisitor(BooleanClause.Occur.MUST, this);
+    if (v2 != null) {
```
javanna: remove the null check here and above?

lucene/core/src/java/org/apache/lucene/search/spans/SpanMultiTermQueryWrapper.java
```diff
@@ -121,7 +122,15 @@ public String toString(String field) {
   public Query rewrite(IndexReader reader) throws IOException {
     return rewriteMethod.rewrite(reader, query);
   }
-  
+
+  @Override
+  public void visit(QueryVisitor visitor) {
+    QueryVisitor v = visitor.getSubVisitor(Occur.MUST, this);
+    if (v != null) {
```
javanna: remove the null check?

lucene/core/src/java/org/apache/lucene/search/spans/SpanNearQuery.java
```diff
@@ -265,6 +259,16 @@ public Query rewrite(IndexReader reader) throws IOException {
     return super.rewrite(reader);
   }
 
+  @Override
+  public void visit(QueryVisitor visitor) {
+    for (SpanQuery clause : clauses) {
+      QueryVisitor v = visitor.getSubVisitor(BooleanClause.Occur.MUST, this);
+      if (v != null) {
```
javanna: null check?

lucene/core/src/java/org/apache/lucene/search/spans/SpanNotQuery.java
```diff
@@ -209,7 +205,20 @@ public Query rewrite(IndexReader reader) throws IOException {
...
-    /** Returns true iff <code>o</code> is equal to this. */
+
+  @Override
+  public void visit(QueryVisitor visitor) {
+    QueryVisitor i = visitor.getSubVisitor(BooleanClause.Occur.MUST, this);
+    if (i != null) {
+      include.visit(i);
+    }
+    QueryVisitor x = visitor.getSubVisitor(BooleanClause.Occur.MUST_NOT, this);
+    if (x != null) {
```
javanna: null check here and above

lucene/core/src/java/org/apache/lucene/search/spans/SpanOrQuery.java
```diff
@@ -88,6 +89,16 @@ public Query rewrite(IndexReader reader) throws IOException {
     return super.rewrite(reader);
   }
 
+  @Override
+  public void visit(QueryVisitor visitor) {
+    for (SpanQuery q : clauses) {
+      QueryVisitor v = visitor.getSubVisitor(BooleanClause.Occur.SHOULD, this);
+      if (v != null) {
```
javanna: here too

lucene/core/src/java/org/apache/lucene/search/spans/SpanPositionCheckQuery.java
```diff
@@ -126,6 +122,14 @@ public Query rewrite(IndexReader reader) throws IOException {
     return super.rewrite(reader);
   }
 
+  @Override
+  public void visit(QueryVisitor visitor) {
+    QueryVisitor v = visitor.getSubVisitor(BooleanClause.Occur.MUST, this);
+    if (v != null) {
```
javanna: another one

lucene/highlighter/src/test/org/apache/lucene/search/uhighlight/TestUnifiedHighlighterMTQ.java
```diff
@@ -1051,6 +1043,14 @@ public SpanWeight createWeight(IndexSearcher searcher, ScoreMode scoreMode, floa
       return originalQuery.createWeight(searcher, scoreMode, boost);
     }
 
+    @Override
+    public void visit(QueryVisitor visitor) {
+      QueryVisitor v = visitor.getSubVisitor(BooleanClause.Occur.MUST, this);
+      if (v != null) {
```
javanna: null check

lucene/queries/src/java/org/apache/lucene/queries/function/FunctionScoreQuery.java
```diff
@@ -119,6 +119,14 @@ public Query rewrite(IndexReader reader) throws IOException {
     return new FunctionScoreQuery(rewritten, source);
   }
 
+  @Override
+  public void visit(QueryVisitor visitor) {
+    QueryVisitor v = visitor.getSubVisitor(BooleanClause.Occur.MUST, this);
+    if (v != null) {
```
javanna: null check

lucene/queries/src/java/org/apache/lucene/queries/payloads/PayloadScoreQuery.java
```diff
@@ -86,6 +87,14 @@ public Query rewrite(IndexReader reader) throws IOException {
     return super.rewrite(reader);
   }
 
+  @Override
+  public void visit(QueryVisitor visitor) {
+    QueryVisitor v = visitor.getSubVisitor(BooleanClause.Occur.MUST, this);
+    if (v != null) {
```
javanna: null check

lucene/queries/src/java/org/apache/lucene/queries/payloads/SpanPayloadCheckQuery.java
```diff
@@ -77,6 +78,14 @@ public Query rewrite(IndexReader reader) throws IOException {
     return super.rewrite(reader);
   }
 
+  @Override
+  public void visit(QueryVisitor visitor) {
+    QueryVisitor v = visitor.getSubVisitor(BooleanClause.Occur.MUST, this);
+    if (v != null) {
```
javanna: null check

lucene/sandbox/src/java/org/apache/lucene/search/CoveringQuery.java
```diff
@@ -109,6 +107,17 @@ public Query rewrite(IndexReader reader) throws IOException {
...
+      if (v == null) {
```
javanna: I think you can remote this null check

lucene/sandbox/src/java/org/apache/lucene/search/intervals/ConjunctionIntervalsSource.java
```diff
@@ -56,9 +56,14 @@ public String toString() {
   }
 
   @Override
-  public void extractTerms(String field, Set<Term> terms) {
+  public void visit(String field, QueryVisitor visitor) {
+    Query parent = new IntervalQuery(field, this);
     for (IntervalsSource source : subSources) {
-      source.extractTerms(field, terms);
+      QueryVisitor v = visitor.getSubVisitor(BooleanClause.Occur.MUST, parent);
+      if (v == null) {
```
javanna: null check here too

lucene/sandbox/src/java/org/apache/lucene/search/intervals/DifferenceIntervalsSource.java
```diff
@@ -83,8 +83,16 @@ public String toString() {
...
-  public void extractTerms(String field, Set<Term> terms) {
-    minuend.extractTerms(field, terms);
+  public void visit(String field, QueryVisitor visitor) {
+    IntervalQuery q = new IntervalQuery(field, this);
+    QueryVisitor m = visitor.getSubVisitor(BooleanClause.Occur.MUST, q);
+    if (m != null) {
+      minuend.visit(field, m);
+    }
+    QueryVisitor s = visitor.getSubVisitor(BooleanClause.Occur.MUST_NOT, q);
+    if (s != null) {
```
javanna: null check here and above

lucene/sandbox/src/java/org/apache/lucene/search/intervals/DisjunctionIntervalsSource.java
```diff
@@ -84,9 +85,14 @@ public String toString() {
   }
 
   @Override
-  public void extractTerms(String field, Set<Term> terms) {
+  public void visit(String field, QueryVisitor visitor) {
+    Query parent = new IntervalQuery(field, this);
     for (IntervalsSource source : subSources) {
-      source.extractTerms(field, terms);
+      QueryVisitor v = visitor.getSubVisitor(BooleanClause.Occur.SHOULD, parent);
+      if (v == null) {
```
javanna: here you want to check for empty visitor instead I think

lucene/sandbox/src/java/org/apache/lucene/search/intervals/ConjunctionIntervalsSource.java
javanna: I guess you want to replace this with a check for empty visitor

lucene/sandbox/src/java/org/apache/lucene/search/intervals/MinimumShouldMatchIntervalsSource.java
```diff
@@ -85,9 +85,14 @@ public MatchesIterator matches(String field, LeafReaderContext ctx, int doc) thr
...
+      if (v == null) {
```
javanna: check for empty visitor?

lucene/core/src/java/org/apache/lucene/search/QueryVisitor.java
```diff
@@ -0,0 +1,80 @@
...
+   * {@link #EMPTY_VISITOR}
+   *
+   * @param occur   the relationship between the parent and its children
+   * @param parent  the query visited
+   */
+  public QueryVisitor getSubVisitor(BooleanClause.Occur occur, Query parent) {
+    if (occur == BooleanClause.Occur.MUST_NOT) {
+      return EMPTY_VISITOR;
+    }
+    return this;
```
javanna: I think I had missed above that this returns itself, which kind of answers my question around where the state is. I do find this a bit hard to follow, is `getSubVisitor` used only for short-circuiting based on the occur argument? Are there cases where we return something that differs from empty or this?

romseygeek: Have a look in QueryVisitorTests to see an example of a more complicated visitor, which builds a tree of conjunction/disjunction nodes and then selects minimum sets of terms that must be present in a document for that doc to match.

lucene/core/src/java/org/apache/lucene/search/QueryVisitor.java
```diff
@@ -0,0 +1,80 @@
...
+   * @param query the leaf query
+   * @param term  the term the query will match on
+   */
+  public void consumesTerm(Query query, Term term) { }
+
+  /**
+   * Called by leaf queries that do not match on terms
+   * @param query the query
+   */
+  public void visitLeaf(Query query) { }
```
javanna: Not knowing all the internals of this and their potential future usages, I am looking at the API mainly and I have a couple of questions:

- should all compound queries have their corresponding `visit` method which calls `getSubVisitor` for each of their sub-query?

- when should `visitLeaf` be used compared to `consumeTerms`? I understand from the javadocs above that there are queries which are not compound queries but don't hold terms either.  If so what would query visitors generally do in their `visitLeaf` impl? I think I am getting to whether `consumesTerm` and `visitLeaf` could be combined together somehow.

romseygeek: Yes, all compound queries should pull a sub visitor and call `visit` on each of their sub queries with it.  And `visitLeaf` is for queries that don't match on terms.  Maybe `visitTermLeaf` might be a better name than `consumesTerm`?  In future we might want to add `visitBKDRange` or something similar - for example, to implement range query rewriting like Adrien mentioned on LUCENE-8708

lucene/core/src/java/org/apache/lucene/search/QueryVisitor.java
uschindler: @s1monw I know it for sure, you were once a user of the original version of this APIon the Lucene issue - using InvocationVisitor to rewrite output of a QueryParser. I am still using this API for the same customer (with a slightly improved InvocationVisitor using MethodHandles and a better dispatch algorithm). I have to figure out what I can share about this customer, but the overlaps are there, so it's great to get this at some point.
What I like here is that this is hard-typed, but on the other hand I don't really like that visiting takes care of field names and terms: Not all queries are dealing with that. So it should be possible to have both ways: Just walk through a query tree in general, but if you are only interested in terms, it might be better to have it like proposed here. It should just be both possiblities (abstraction from terms and fields).

lucene/core/src/java/org/apache/lucene/search/QueryVisitor.java
javanna: makes sense, so I take it that such usage will become production code in the future and not only test code. is the parent argument needed ?

lucene/core/src/java/org/apache/lucene/search/QueryVisitor.java
javanna: > Maybe `visitTermLeaf` might be a better name than `consumesTerm`?

I would rather look for a way to make `visitLeafQuery` clearer, but it may be my own issue with its naming. I find `consumeTerms` clear.

lucene/core/src/java/org/apache/lucene/search/QueryVisitor.java
javanna: is the query argument needed? I also wonder, do terms need to be provided one by one, or could the term argument be a varargs ? just an idea, not sure it makes sense

s1monw: I though about this too, I think a varArg API would be great but I think we also need a Iterator based API then? maybe not. Still varArg would be great we can implement early termination there easily on per-visitor and don't need to iterate everywhere.

s1monw: also +1 for singular vs. plural

lucene/core/src/java/org/apache/lucene/search/QueryVisitor.java
s1monw: ok

lucene/core/src/test/org/apache/lucene/search/TestBooleanQuery.java
```diff
@@ -764,15 +763,31 @@ public void testExtractTerms() throws IOException {
...
+      Term expected;
 
-    Set<Term> matchingTerms = new HashSet<>();
-    searcher.createWeight(searcher.rewrite(bq), ScoreMode.COMPLETE_NO_SCORES, 1).extractTerms(matchingTerms);
-    assertEquals(new HashSet<>(Arrays.asList(a, b, c)), matchingTerms);
+      @Override
+      public QueryVisitor getSubVisitor(Occur occur, Query parent) {
+        switch (occur) {
+          case SHOULD:
+            expected = a; break;
```
s1monw: can you move the break on an extra line please

lucene/core/src/test/org/apache/lucene/search/TestBooleanQuery.java
```diff
@@ -764,15 +763,31 @@ public void testExtractTerms() throws IOException {
...
+          case MUST:
+            expected = b; break;
+          case FILTER:
+            expected = c; break;
+          case MUST_NOT:
+            expected = d;
```
s1monw: please add a break; and make sure we throw ISE in the default case

lucene/core/src/java/org/apache/lucene/search/QueryVisitor.java
romseygeek: The tricky part with varArgs is what the semantics are - is it always the equivalent of a SHOULD or can PhraseQuery use it too, in which case it's a MUST?

lucene/core/src/java/org/apache/lucene/search/FuzzyQuery.java
romseygeek: ++, changed

lucene/core/src/java/org/apache/lucene/search/QueryVisitor.java
s1monw: I do remember what you are referring to. I think this can be a follow up to this change?

lucene/core/src/java/org/apache/lucene/search/QueryVisitor.java
s1monw: doesn't this depend on the SubVisitor here?

lucene/core/src/java/org/apache/lucene/search/QueryVisitor.java
uschindler: If the visitor API works in general, adding more methods to improve usability (maybe using functional interfaces) would be a great win. I agree, can be done later.

The most important parts are covered here, so I am fine. I will still look at the code this evening to check if I find some problems. Most of the patch here deals with extractTerms, the actual changes in API are minor...

lucene/core/src/java/org/apache/lucene/search/QueryVisitor.java
uschindler: I hate "Optional" unless we get value types in Java.

lucene/core/src/java/org/apache/lucene/search/QueryVisitor.java
romseygeek: > so I take it that such usage will become production code in the future and not only test code

Yes - I have plans for luwak :) . I've added some test code to show how the `parent` argument can be used as well.

lucene/core/src/java/org/apache/lucene/search/QueryVisitor.java
```diff
@@ -0,0 +1,81 @@
...
+
+/**
+ * Allows recursion through a query tree
+ *
+ * @see Query#visit(QueryVisitor)
+ */
+public abstract class QueryVisitor {
+
+  /**
+   * Called by leaf queries that match on a specific term
```
javanna: "on a specific term" or "on specific terms" ?

lucene/core/src/java/org/apache/lucene/search/QueryVisitor.java
```diff
@@ -0,0 +1,81 @@
...
+   *
+   * @param query  the leaf query
+   * @param terms  the terms the query will match on
+   */
+  public void consumeTerms(Query query, Term... terms) { }
```
javanna: do we ever use the query argument? Is it needed?

romseygeek: I think we do want it, otherwise you can't tell the difference between for example a TermQuery and a SpanTermQuery

lucene/core/src/java/org/apache/lucene/search/QueryVisitor.java
romseygeek: ++

lucene/core/src/java/org/apache/lucene/search/QueryVisitor.java
```diff
@@ -0,0 +1,81 @@
...
+import java.util.Set;
+
+import org.apache.lucene.index.Term;
+
+/**
+ * Allows recursion through a query tree
+ *
+ * @see Query#visit(QueryVisitor)
+ */
+public abstract class QueryVisitor {
```
s1monw: this is an abstract class but doesn't mark anything abstract, that's odd. I also wonder why it's not an interface?

dsmiley:
I'm not sure if my comment got lost somewhere but I strongly feel that Query.visit should be an abstract method.  Each subclass ought to determine what to do; we don't know a suitable default choice.  By having a default choice, we let things slip through... for example this patch has forgotten about the surround query parser's Query subclasses, and forgetting this means that these queries aren't highlighted by the UnifiedHighlighter.  However on the backport to 8x I can see the need to implement a default method for backwards compatibility.

s1monw:
>  Query.visit should be an abstract 

+1

lucene/core/src/java/org/apache/lucene/search/QueryVisitor.java
romseygeek: It's not an interface so that you can call `super.getSubVisitor` to get the default behaviour, but there's no reason for it to be abstract any more.  I'll change.

romseygeek:
> Query.visit should be an abstract

Sure, I'll push some changes.

romseygeek:
Adding an explicit implementation on every class here has made me rethink how we're dealing with things like `TermInSetQuery`, which currently doesn't call `consumeTerms` at all; this is copied from its `extractTerms` implementation, which did nothing so as to protect highlighters, but I think this is incorrect.  One idea would be to change the `Query.visit` method to take a `Predicate<String>` field selector; that way highlighters that aren't interested in the field that a `TermInSetQuery` is applied to can avoid dealing with its large set of terms, but we can still access those terms if need be.

romseygeek:
I made Query.visit() abstract, and changed it to take a field selector string - I think this makes a lot more sense now?

lucene/classification/src/java/org/apache/lucene/classification/utils/NearestFuzzyQuery.java
```diff
@@ -327,4 +329,9 @@ private boolean equalsTo(NearestFuzzyQuery other) {
         Objects.equals(fieldVals, other.fieldVals);
   }
 
+  @Override
+  public void visit(QueryVisitor visitor, Predicate<String> fieldSelector) {
```
s1monw: what's the advantage over asking the visitor if the field is accepted? I also wonder if it can't just be an implementation detail of a specific visitor if a field is of interest or not? It can be a simple delegate?

romseygeek: As in, add an `acceptsField` method to QueryVisitor?  What I'm trying to do is to avoid heavy computations in queries like TermInSetQuery when the visitor isn't actually interested in the terms from that query's field.

s1monw: > As in, add an acceptsField method to QueryVisitor?

yes that's what I mean. I am just trying to think what we would do if we'd have another predicate in the future?

romseygeek: ++, will change

lucene/core/src/java/org/apache/lucene/search/QueryVisitor.java
```diff
@@ -0,0 +1,82 @@
...
+import java.util.function.Predicate;
+
+import org.apache.lucene.index.Term;
+
+/**
+ * Allows recursion through a query tree
+ *
+ * @see Query#visit(QueryVisitor, Predicate)
+ */
+public class QueryVisitor {
```
dsmiley: Shouldn't this be abstract?  Even if it doesn't *have* to be; it's a signal we expect it to be subclassed.

lucene/core/src/java/org/apache/lucene/search/QueryVisitor.java
```diff
@@ -0,0 +1,82 @@
...
+
```
dsmiley: What's the state of wether or not we have a way of consuming Automata from the visitor?  Is this a TODO?  If it is, lets have the TODO here.

lucene/core/src/java/org/apache/lucene/search/QueryVisitor.java
romseygeek: It was, but @s1monw objected - I think I'm more in agreement with you here though, passing an unsubclassed QueryVisitor to visit() would most likely be a user error.

lucene/core/src/java/org/apache/lucene/search/QueryVisitor.java
romseygeek: ++ will add one

lucene/core/src/java/org/apache/lucene/search/QueryVisitor.java
s1monw: fair enough lets make it abstract

romseygeek:
I think this is ready, so the next question is on backporting - are we OK to add an abstract method to `Query` and remove one from `Weight` in a point release?  I'd say it was fine, as these are both expert APIs and they'll generate compile-time errors for people upgrading.

s1monw:
LGTM - I think we can backport it. I wonder if it's worth it to not break BWC give the low level API we touch. I do wonder about removing extractTerms, that might not be so low level. Maybe we can keep it for BWC and deprecate?

dsmiley:
I think we should backport to 8x and retain BWC.  These may be "advanced" but lets not needlessly annoy all the users out there with custom Queries.

s1monw:
ship it. @dsmiley do you wanna approve it?

PR #582, opened 2019-02-20
Change heap array init index to 1

hanbj:
change heap array init index to 1


ctargett:
Removed Jira link to LUCENE-582 in the title, as that issue was resolved in 2008. Since this PR is #582, I suspect the PR and issue ID were confused.

I cannot find a corresponding Jira issue for this in our system, I'd recommend to @hanbj to create one at https://issues.apache.org/jira/projects/LUCENE.

jpountz:
@hanbj It is done this way in order to avoid wasting one sentinel object.

PR #583, opened 2019-02-21
SOLR-13151

moshebla:


solr/core/src/java/org/apache/solr/cloud/api/collections/CategoryRoutedAlias.java
```diff
@@ -119,13 +119,19 @@ public void validateRouteValue(AddUpdateCommand cmd) throws SolrException {
...
     String dataValue = String.valueOf(cmd.getSolrInputDocument().getFieldValue(getRouteField()));
+    String mustMatch = aliasMetadata.get(ROUTER_MUST_MATCH);
     String candidateCollectionName = buildCollectionNameFromValue(dataValue);
     List<String> cols = getCollectionList(this.parsedAliases);
 
     if (cols.contains(candidateCollectionName)) {
       return;
     }
 
+    if (mustMatch != null && !candidateCollectionName.matches(mustMatch)) {
```
gus-asf: We should use a pre-compiled expression. When the CRA is created this expression should be precompiled and stored as a field.

solr/core/src/test/org/apache/solr/update/processor/CategoryRoutedAliasUpdateProcessorTest.java
```diff
@@ -233,6 +233,54 @@ public void testMissingRequiredParams() throws Exception {
     assertTrue("Create Alias should fail since not all required params were supplied", e.getMessage().contains("Not all required params were supplied"));
   }
 
+  @Slow
+  @Test
+  public void testMustMatch() throws Exception {
+    String configName = getSaferTestName();
+    createConfigSet(configName);
+    final String regexEnding = "_solr";
```
gus-asf: This test would be clearer if the regex were expressed as a single string.

solr/solrj/src/java/org/apache/solr/client/solrj/request/CollectionAdminRequest.java
```diff
@@ -1652,15 +1654,21 @@ public CreateCategoryRoutedAlias setMaxCardinality(int maxCardinality) {
...
     public SolrParams getParams() {
       ModifiableSolrParams params = (ModifiableSolrParams) super.getParams();
       params.add(CommonParams.NAME, aliasName);
       params.add(ROUTER_TYPE_NAME, "category");
       params.add(ROUTER_FIELD, routerField);
+      params.add(ROUTER_MAX_CARDINALITY, maxCardinality.toString());
 
-      if (maxCardinality != null) {
-        params.add(ROUTER_MAX_CARDINALITY, maxCardinality.toString());
+      if (mustMatch != null) {
```
gus-asf: Reading this on github, so maybe I'm reading it wrong, but won't mustMatch always start out null and this check won't ever pass?

moshebla: It only passes if setMustMatch was called,
since it sets mustMatch.

solr/core/src/test/org/apache/solr/update/processor/CategoryRoutedAliasUpdateProcessorTest.java
moshebla: On it.

PR #584, opened 2019-02-22
SOLR-10751: PULL Replicas don’t replicate empty index

tflobbe:
This is still WIP. Should also solve SOLR-11094 (and maybe SOLR-12100). This PR also includes changes to TestTlogReplica, which I’m going to separate to a different commit


tflobbe:
This shouldn't be a PR yet. I'll open a new one when this is closer to being merged

PR #587, opened 2019-02-25
LUCENE-8707: Add LatLonShape and XYShape distance query

iverase:
LatLonShape query distance based heavily in LatLonPoint implementation


iverase:
I went through another iteration and built those queries using the Component2D abstraction. I think it is looking pretty good, LatLonShape supports circles that goes over the date line.

@nknize @jpountz feedback would be great!

lucene/sandbox/src/java/org/apache/lucene/geo/WGS84Circle2D.java
```diff
@@ -0,0 +1,285 @@
...
+    double[] vectorAP = new double[] {centerLon - aX, centerLat - aY};
+    double[] vectorAB = new double[] {bX - aX, bY - aY};
+
+    double magnitudeAB = vectorAB[0] * vectorAB[0] + vectorAB[1] * vectorAB[1];
+    double dotProduct = vectorAP[0] * vectorAB[0] + vectorAP[1] * vectorAB[1];
+
+    double distance = dotProduct / magnitudeAB;
+
+    if (distance < 0 || distance > dotProduct)
+    {
```
jpountz: the rest of the code base puts brackets on the previous line

lucene/sandbox/src/java/org/apache/lucene/document/LatLonShapeDistanceQuery.java
```diff
@@ -0,0 +1,115 @@
...
+
+    double alat = GeoEncodingUtils.decodeLatitude(scratchTriangle.aY);
+    double alon = GeoEncodingUtils.decodeLongitude(scratchTriangle.aX);
+    double blat = GeoEncodingUtils.decodeLatitude(scratchTriangle.bY);
+    double blon = GeoEncodingUtils.decodeLongitude(scratchTriangle.bX);
+    double clat = GeoEncodingUtils.decodeLatitude(scratchTriangle.cY);
+    double clon = GeoEncodingUtils.decodeLongitude(scratchTriangle.cX);
+
+    switch (queryRelation) {
+      case INTERSECTS: return circle2D.relateTriangle(alon, alat, blon, blat, clon, clat) != Relation.CELL_OUTSIDE_QUERY;
```
jpountz: Are we expecting INTERSECTS to be the common case? If yes it's a bit disappointing that it it checks all corners in the common case, when it could stop after one corner falls within the circle? In any case I don't think we need to handle it now, but we should at least leave a TODO?

lucene/sandbox/src/java/org/apache/lucene/document/LatLonShapeDistanceQuery.java
```diff
@@ -0,0 +1,115 @@
...
+      case WITHIN: return circle2D.relateTriangle(alon, alat, blon, blat, clon, clat) == Relation.CELL_INSIDE_QUERY;
+      case DISJOINT: return circle2D.relateTriangle(alon, alat, blon, blat, clon, clat) == Relation.CELL_OUTSIDE_QUERY;
+      default: throw new IllegalArgumentException("Unsupported query type :[" + queryRelation + "]");
```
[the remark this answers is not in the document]
iverase: CONTAINS needs to follow a different logic that is implemented in `withinTriangle`. In that case we need to take not account if the edges belong to the original shape or not.

lucene/sandbox/src/java/org/apache/lucene/document/LatLonShapeDistanceQuery.java
iverase: Are you thinking in having a specialise method for intersects so we can break the logic as soon as possible? This logic is similar to the logic implemented in Polygon2D, an improvement here should be applicable to that logic as well.

jpountz: Yes something like that indeed.

lucene/core/src/java/org/apache/lucene/geo/Circle2D.java
```diff
@@ -0,0 +1,463 @@
...
+    // indexed "triangle" is a triangle:
+    return relateIndexedTriangle(minX, maxX, minY, maxY, ax, ay, bx, by, cx, cy);
+  }
+
+  @Override
+  public WithinRelation withinTriangle(double minX, double maxX, double minY, double maxY,
+                                       double ax, double ay, boolean ab, double bx, double by, boolean bc, double cx, double cy, boolean ca) {
+    // short cut, lines and points cannot contain this type of shape
+    if ((ax == bx && ay == by) || (ax == cx && ay == cy) || (bx == cx && by == cy)) {
+      return WithinRelation.DISJOINT;
```
jpountz: couldn't it be NOTWITHIN in some cases?

iverase: Strictly speaking yes but a point or a line would never return CANDIDATE for a circle. I guess I am using disjoint because is the cheaper answer (the shape is actually ignored) but I agree this is probably wrong.

PR #590, opened 2019-02-26
SOLR-13152

moshebla:
make maxCardinality part of createCRACmd constructor para,s and validate CRA params during construction of alias.


gus-asf:
This lacks documentation, several additional validation checks and V2 api support, but in the interest of time I will handle these.

PR #597, opened 2019-03-04
[SOLR-13272] feat(facet/interval): support json facet requests for interval facet

apoorvprecisely:


solr/core/src/java/org/apache/solr/search/facet/FacetIntervalProcessor.java
```diff
@@ -0,0 +1,154 @@
...
+    rangeList = new ArrayList<>();
+    otherList = new ArrayList<>(3);
+
+    Comparable low = start;
+    Comparable loop_end = this.end;
+    List<Range> ranges = parseInterval(interval);
+    for (Range range : ranges) {
+      rangeList.add(range);
+    }
+    createOtherList(loop_end);
```
gd-spb-e1m: why begin/start isn't used to creating other?

solr/core/src/java/org/apache/solr/search/facet/FacetIntervalProcessor.java
```diff
@@ -0,0 +1,154 @@
...
+  }
+
+  private List<Range> parseInterval(Object input) {
+    //interval :[{key:"set1",value:"[0,10]"}]
+    //@todo apoorv handle exception, sort orders
+    List<Map<String, String>> intervals = (List<Map<String, String>>) input;
```
gd-spb-e1m: Why not handle singular ```interval :{key:"set1",value:"[0,10]"}``` ? 
Can we have ```key``` optional? thus if we omit it, repeating ```value``` is boilerplate. 
Also, why ```value```? I'd rather call it ```range```
Another question: shouldn't we employ json to carry those intervals? (might not be good idea at all)

solr/core/src/java/org/apache/solr/search/facet/FacetIntervalProcessor.java
```diff
@@ -0,0 +1,154 @@
...
+    List<Range> ranges = new ArrayList<>();
+    for (Map<String, String> interval : intervals) {
+      String key = interval.get("key");
+      if (interval == null) {
```
gd-spb-e1m: Null check right after deferencing? Brave!

solr/core/src/java/org/apache/solr/search/facet/FacetIntervalProcessor.java
```diff
@@ -0,0 +1,154 @@
...
+      }
+      try {
+        end = calc.getValue(endStr.toString());
+      } catch (SolrException e) {
+        throw new SolrException(SolrException.ErrorCode.BAD_REQUEST, String.format(Locale.ROOT, "Invalid end interval for key '%s': %s", key, e.getMessage()), e);
+      }
+      if (start != null && end != null && start.compareTo(end) > 0) {
+        throw new SolrException(SolrException.ErrorCode.BAD_REQUEST, "Start is higher than end in interval for key: " + key);
+      }
+      if (interval.get("key") == null) key = startStr.toString();
```
gd-spb-e1m: why not full interval range expression?

solr/core/src/java/org/apache/solr/search/facet/FacetIntervalProcessor.java
apoorvprecisely: the code block that populates `otherList` doesn't use `start` and `end`
I just refactored it, I can change it to what it was earlier

solr/core/src/java/org/apache/solr/search/facet/FacetIntervalProcessor.java
apoorvprecisely: - I'll make change to allow passing a single interval also, but keeping it an array would be consistent, the key is `intervals` only, my comment on :69 is wrong
- If key is not present the value of key defaults to the lower limit of `value`
- will change it to range
- Didn't get your question, 
this is the request format that I've assumed
```
json.facet={pubyear:{type : interval,field : pubyear_i,intervals:[{key:"2000-2200",value:"[2000,2200]"}]}}
```

solr/core/src/java/org/apache/solr/search/facet/FacetIntervalProcessor.java
apoorvprecisely: my bad, will change it

solr/core/src/java/org/apache/solr/search/facet/FacetIntervalProcessor.java
apoorvprecisely: This is what `key` defaults to in the older apis, keeping whole interval range expression would make more sense, will change

apoorvprecisely:
@gd-spb-e1m for syntax, just for the sake of PR I assumed something like this
```
json.facet={pubyear:{type : interval,field : pubyear_i,intervals:[{key:"2000-2200",value:"[2000,2200]"}]}}
```
I'll add test cases, since I raise this PR based on a lot of assumptions on the request format I didn't add any yet.
How can we arrive at a request format? 
I tried to keep the format similar to the older interval json apis

solr/core/src/java/org/apache/solr/search/facet/FacetIntervalProcessor.java
chatman: >  Another question: shouldn't we employ json to carry those intervals? (might not be good idea at all)
Might make it verbose for simple ranges.

>  Also, why value? I'd rather call it range
+1

solr/core/src/java/org/apache/solr/search/facet/FacetRange.java
```diff
@@ -299,8 +323,127 @@ private void createRangeList() throws IOException {
       actual_end = null;
     }
   }
-  
-  
+
+  private List<Range> parseInterval(Object input) {
+    //intervals :[{key:"set1",value:"[0,10]"}]
+    //@todo apoorv handle exception, sort orders
```
munendrasn: could you please validate the input? What is the `sort orders` here?

solr/core/src/test/org/apache/solr/search/facet/TestJsonFacets.java
```diff
@@ -3225,6 +3224,58 @@ public void testDomainErrors() throws Exception {
 
   }
 
+  @Test
+  public void testIntervalFacets() throws Exception {
```
munendrasn: Please add test cases to cover `other`, `include` and refinement

solr/core/src/test/org/apache/solr/search/facet/TestJsonFacets.java
```diff
@@ -49,8 +49,8 @@
 //   TestCloudJSONFacetJoinDomain for random field faceting tests with domain modifications
 //   TestJsonFacetRefinement for refinement tests
 
-@LuceneTestCase.SuppressCodecs({"Lucene3x","Lucene40","Lucene41","Lucene42","Lucene45","Appending"})
-public class TestJsonFacets extends SolrTestCaseHS {
+@LuceneTestCase.SuppressCodecs({"Lucene3x","Lucene40","Lucene41","FST50","Direct","Lucene42","Lucene45","Appending","BlockTreeOrds","FSTOrd50"})
```
munendrasn: any particular reason for suppressing this change?

munendrasn:
This is merged

PR #601, opened 2019-03-08
Adding reader settings for moving fst offheap

jainankitk:
This change adds setting for moving fst offheap, specifically address [JIRA-8671](https://issues.apache.org/jira/browse/LUCENE-8671)

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ project's git repository now:
 - Lucene: <https://gitbox.apache.org/repos/asf/lucene.git>
 - Solr: <https://gitbox.apache.org/repos/asf/solr.git>
 
-Development for buxfixes of 8.11.x releases remains on branch
+Development for bugfixes of 8.11.x releases remains on branch
 `branch_8_11` in the shared repository:
 
 - https://gitbox.apache.org/repos/asf/lucene-solr.git


lucene/core/src/java/org/apache/lucene/store/Directory.java
```diff
@@ -48,6 +49,12 @@
  * @see FilterDirectory
  */
 public abstract class Directory implements Closeable {
+
+  /**
+   * Reader only attributes for this directory.
+   */
+  private Map<String, String> readerAttributes;
```
s1monw: I don't think `Directory` is the right abstraction to hold this. I suggest we add an overload to DirectoryReader.open that looks like this:

```Java
  public static DirectoryReader open(final Directory directory, final Map<String, String> attributes) throws IOException;

//
  public static DirectoryReader open(final IndexCommit commit, final Map<String, String> attributes) throws IOException;
```
that allows us to have a dedicated map of attributes per open call. Subsequent reopens then inherit the attributes form the reader. You can also then make this a property of `DirecotryReader` I guess.

lucene/core/src/java/org/apache/lucene/index/IndexWriterConfig.java
```diff
@@ -482,6 +483,15 @@ public IndexWriterConfig setCommitOnClose(boolean commitOnClose) {
     return this;
   }
 
+  /**
+   * Sets read time attributes that will be loaded into SegmentReadState while
+   * opening the index
+   */
+  public IndexWriterConfig setReaderAttributes(Map<String, String> readerAttributes) {
+    this.readerAttributes = readerAttributes;
```
s1monw: this one should be an unmodifiable map, you should also make a copy for the attributes passed in.

lucene/core/src/java/org/apache/lucene/codecs/blocktree/FieldReader.java
```diff
@@ -88,10 +97,13 @@
 
     if (indexIn != null) {
       final IndexInput clone = indexIn.clone();
+      //System.out.println("start=" + indexStartFP + " field=" + fieldInfo.name);
       clone.seek(indexStartFP);
-      // Initialize FST offheap if index is MMapDirectory and
-      // docCount != sumDocFreq implying field is not primary key
-      if (clone instanceof ByteBufferIndexInput && this.docCount != this.sumDocFreq) {
+
+      final boolean fstOffHeap = readerAttributes != null && "true".equals(readerAttributes.get(FST_OFFHEAP_ATTRIBUTE));
```
s1monw: don't we want to have a per-field setting? I mean something like:

`"true".equals(readerAttributes.get("blocktree.fst." + fieldInfo.name + ".offheap"));`

this would be more flexible no?

jainankitk: That was the plan initially, but given that non-PK eligible fields are offheap by default. Having single setting instead of per field setting makes more sense.
```
If false:
Only non-PK fields offheap
If true:
All fields offheap
```

lucene/core/src/java/org/apache/lucene/index/IndexWriterConfig.java
jainankitk: Good point. Will make the change. Thanks!

lucene/core/src/java/org/apache/lucene/store/Directory.java
jainankitk: >I don't think Directory is the right abstraction to hold this. 

Is that because single directory can be opened in multiple contexts?

>Subsequent reopens then inherit the attributes form the reader. You can also then make this a property of DirecotryReader I guess.

How is the attribute map passed to SegmentReadState in this approach? Also, IndexWriterConfig need not be changed with this approach, right?

lucene/core/src/java/org/apache/lucene/codecs/blocktree/FieldReader.java
s1monw: why is this? I mean there might be fields that can easily be off-heap while another one should be on-heap?

lucene/core/src/java/org/apache/lucene/store/Directory.java
s1monw: > Is that because single directory can be opened in multiple contexts?
I think there are multiple reasons I think it should be either bound to a single reader or writer. ie two calls to DirectoryReader.open can return readers with different fields on-heap even if they share the same directory.

> How is the attribute map passed to SegmentReadState in this approach? Also, IndexWriterConfig need not be changed with this approach, right?

I don't think IWC needs to change. I think such a map can be passed down to `SegmentCoreReaders` and then passed on via `SegmentReadState` I am not sure why you think we need to change a lot of things?

lucene/core/src/java/org/apache/lucene/codecs/blocktree/FieldReader.java
jainankitk: >why is this? I mean there might be fields that can easily be off-heap while another one should be on-heap?

Having per field setting is unnecessary cognitive load for the user, given that non PK eligible fields are offheap by default. So, users want PK eligible fields either offheap or onheap.
Initially, I did try it with [this patch](https://issues.apache.org/jira/secure/attachment/12958117/offheap_generic_settings.patch), but found it to be unnecessary. There can be dynamic fields as well, for which you need special keyword, which needs to account for all combinations of offheap and onheap.

lucene/core/src/java/org/apache/lucene/store/Directory.java
jainankitk: >I think such a map can be passed down to SegmentCoreReaders and then passed on via SegmentReadState I am not sure why you think we need to change a lot of things?

Maybe, better way is to add `readerAttributes` in `SegmentInfo`, and populate that during `StandardDirectoryReader.open`?

s1monw:
I tired a different way today and I wonder what you think of this #606 (note I am currently getting a 404 on it - seems a github issue)

PR #602, opened 2019-03-08
docu change: use class TopDocs instead of Hits

saschaszott:
`Hits` was removed in Lucene 3.0

### DIFF
--- a/lucene/queries/src/java/org/apache/lucene/queries/mlt/MoreLikeThis.java
+++ b/lucene/queries/src/java/org/apache/lucene/queries/mlt/MoreLikeThis.java
@@ -92,9 +92,9 @@ import org.apache.lucene.util.PriorityQueue;
  * Reader target = ... // orig source of doc you want to find similarities to
  * Query query = mlt.like( target);
  * 
- * Hits hits = is.search(query);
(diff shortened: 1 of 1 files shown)


lucene/queries/src/java/org/apache/lucene/queries/mlt/MoreLikeThis.java
```diff
@@ -92,9 +92,9 @@
  * Reader target = ... // orig source of doc you want to find similarities to
  * Query query = mlt.like( target);
  * 
- * Hits hits = is.search(query);
- * // now the usual iteration thru 'hits' - the only thing to watch for is to make sure
- * //you ignore the doc if it matches your 'target' document, as it should be similar to itself
+ * TopDocs topDocs = is.search(query, 10);
```
atris: Could you add documentation around the second parameter? It seems a bit unclear

janhoy: @saschaszott ?

saschaszott: added comment

PR #606, opened 2019-03-12
LUCENE-8671: Allow more fine-grained control over off-heap term dictionaries

s1monw:
This change allows to control if term-dictionaries are loaded off-heap or on
heap on a per reader basis. None NRT readers will access all term dictionaries
off heap including ID fields while readers that require fast ID access like all
readers used within index writer will by default only load non-ID like fields
heap. Additionally, IOContext has now the ability to specify if ram-usage should be
minimized and can control the off vs. on-heap decisions on a per-reader basis.

Off heap term dictionaries are still only used if the index input is memory mapped.


lucene/core/src/java/org/apache/lucene/store/ByteBuffersIndexInput.java
```diff
@@ -193,5 +193,10 @@ private void ensureOpen() {
     if (in == null) {
       throw new AlreadyClosedException("Already closed.");
     }
-  }  
+  }
+
+  @Override
+  public boolean isMMapped() {
+    return true;
```
dweiss: Hi Simon. Whether this should return true depends on what byte buffers are used? The same applies to ByteBufferIndexInput, actually... I don't think you can generally tell whether the ByteBuffers the input operates on come from a mmap call or from somewhere else (even direct buffers don't have to be a result of mmap).

s1monw: it's rather mmapped or in ram so `true` is fine?

dweiss: Err... I didn't follow this change too closely. I thought the meaning of this is to distinguish between memory-mapped (off-heap) vs. heap-based storage?

s1monw: I will clarify the API.

s1monw: I removed this API went down a different route

s1monw:
@mikemccand I agree - I close this one and try to break it into smaller chunks and different solutions. see #610

PR #607, opened 2019-03-13
Concurrently flush next buffer during commit in RandomIndexWriter

s1monw:
This is a spinn-off from `LUCENE-8700` that is satisfied by IndexWriter#flushNextBuffer.
The idea here is to additionally call flushNextBuffer in RandomIndexWriter for better
test coverage.

### DIFF
--- a/lucene/test-framework/src/java/org/apache/lucene/index/RandomIndexWriter.java
+++ b/lucene/test-framework/src/java/org/apache/lucene/index/RandomIndexWriter.java
@@ -21,6 +21,7 @@ import java.io.IOException;
 import java.util.Arrays;
 import java.util.Iterator;
 import java.util.Random;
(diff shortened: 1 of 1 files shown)


s1monw:
> Do you think it effectively tests the same thing as flushing in parallel on several threads?

I agree it's a bit different. I think it's sufficient to do it this way to make sure we exercise this code in a concurrent fashion. WDYT?

msokolov:
Yes, I guess before I was primarily focused on getting maximum throughput, but for testing purposes this seems to exercise the main point of concern, which I guess would be the potential for some conflict between flushing threads and the main thread. I guess it's conceivable that some infrequent random locking issue would be more likely to arise with more threads contending, but it's kind of a theoretical issue, and with enough randomness and time, all things occur :)

PR #610, opened 2019-03-19
LUCENE-8671: Load FST off-heap if reader is not opened from an index writer

s1monw:
Today we never load FSTs of ID-like fields off-heap since we need
very fast access for updates. Yet, a reader that is not loaded from
an index wirter can also leave the FST on disk. This change adds
this information to SegmentReadState to allow the postiings format
to make this decision without configuration.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/codecs/blocktree/FieldReader.java
+++ b/lucene/core/src/java/org/apache/lucene/codecs/blocktree/FieldReader.java
@@ -63,10 +63,11 @@ public final class FieldReader extends Terms implements Accountable {
   final BlockTreeTermsReader parent;
 
   final FST<BytesRef> index;
+  final boolean isFSTOffHeap;
   //private boolean DEBUG;
 
   FieldReader(BlockTreeTermsReader parent, FieldInfo fieldInfo, long numTerms, BytesRef rootCode, long sumTotalTermFreq, long sumDocFreq, int docCount,
-              long indexStartFP, int longsSize, IndexInput indexIn, BytesRef minTerm, BytesRef maxTerm) throws IOException {
+              long indexStartFP, int longsSize, IndexInput indexIn, BytesRef minTerm, BytesRef maxTerm, boolean openedFromWriter) throws IOException {
     assert numTerms > 0;
     this.fieldInfo = fieldInfo;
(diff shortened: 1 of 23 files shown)


s1monw:
@mikemccand FYI

lucene/core/src/java/org/apache/lucene/codecs/blocktree/FieldReader.java
```diff
@@ -91,7 +92,8 @@
       clone.seek(indexStartFP);
       // Initialize FST offheap if index is MMapDirectory and
       // docCount != sumDocFreq implying field is not primary key
-      if (clone instanceof ByteBufferIndexInput && this.docCount != this.sumDocFreq) {
+      isFSTOffHeap = clone instanceof ByteBufferIndexInput && ((this.docCount != this.sumDocFreq) || openedFromWriter == false);
```
mikemccand: Is the idea here that it's only `IndexWriter` that needs fast ID lookups (since it uses this when deleting docs by ID field)?  E.g. apps that open `IndexReader` themselves (outside of `IndexWriter`) don't need fast ID lookups by default?

lucene/core/src/java/org/apache/lucene/index/SegmentReadState.java
```diff
@@ -49,23 +49,29 @@
    *  {@link IndexFileNames#segmentFileName(String,String,String)}). */
   public final String segmentSuffix;
 
+  /**
+   * True iff this SegmentReadState is opened from an index writer.
```
mikemccand: s/`index writer`/`IndexWriter`

lucene/core/src/java/org/apache/lucene/codecs/blocktree/FieldReader.java
jpountz: I guess some users need fast lookups on read-only indices, but this proposal looks like a good default to me.

s1monw: @mikemccand you assumption is correct. @jpountz I will work on making it possible for users to  configure it down the road but I wanted to break this one out to make it simpler to iterate and not mix concerns.

PR #613, opened 2019-03-20
LUCENE-8671: Expose FST off/on-heap options on Lucene50PostingsFormat

s1monw:
Before we can expose options to configure this postings format
on a per-reader basis we need to expose the option to load the terms
index FST off or on heap on the postings format. This already allows to
change the default in a per-field posting format if an expert user
wants to change the defaults. This essentially provides the ability to change
defaults globally while still involving some glue code.


s1monw:
@mikemccand can you look at this too pls

lucene/core/src/java/org/apache/lucene/codecs/blocktree/FieldReader.java
```diff
@@ -84,21 +85,18 @@
     // if (DEBUG) {
     //   System.out.println("BTTR: seg=" + segment + " field=" + fieldInfo.name + " rootBlockCode=" + rootCode + " divisor=" + indexDivisor);
     // }
-
     rootBlockFP = (new ByteArrayDataInput(rootCode.bytes, rootCode.offset, rootCode.length)).readVLong() >>> BlockTreeTermsReader.OUTPUT_FLAGS_NUM_BITS;
-
+    // Initialize FST offheap if index is MMapDirectory and
+    // docCount != sumDocFreq implying field is not primary key
+    isFSTOffHeap = ((this.docCount != this.sumDocFreq) || openedFromWriter == false) && loadFSTOffHeap;
```
mikemccand: Should we instead pass in the `FSTLoadMode`  enum down here?  And if it's AUTO then apply these defaulting heuristics?

This way the codec/caller could decide to leave FST off heap even if its a primary key field or its opened from writer?

s1monw: sure we can do that

lucene/core/src/java/org/apache/lucene/codecs/lucene50/Lucene50PostingsFormat.java
```diff
@@ -389,6 +390,33 @@
...
+    OFF_HEAP,
+    /**
+     * Never read FSTs from disk ie. all fields FSTs are loaded into memory
+     */
+    ON_HEAP,
+    /**
+     * Always read FSTs from disk.
+     * An exception is made for ID fields in an IndexWriter context which are always loaded into memory.
+     * This is useful to guarantee best update performance even if a non MMapDirectory is used which required for
+     * the {@link FSTLoadMode#AUTO} mode is used.
```
mikemccand: Two `is used` in the last sentence here?

Maybe include a warning here that this does not look at `Directory` type and may put FST off-heap even for buffered directory implementations?

PR #616, opened 2019-03-22
LUCENE-8734: Record nextWriteGens in SegmentCommitInfo

dnhatn:
When opening an IndexWriter, we will set nextWriteDelGen,
nextWriteFieldInfosGen, and nextWriteDocValuesGen beyond what we have in
the directory to avoid double writes in cases where the previous IW was
not closed gracefully or the new IW is opened with a previous commit
(not the latest commit). We need to persist these values; otherwise,
they will be lost when the current IW is closed.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/SegmentCommitInfo.java
+++ b/lucene/core/src/java/org/apache/lucene/index/SegmentCommitInfo.java
@@ -91,16 +91,25 @@ public class SegmentCommitInfo {
    * @param docValuesGen
    *          DocValues generation number (used to name doc-values updates files)
    */
-  public SegmentCommitInfo(SegmentInfo info, int delCount, int softDelCount, long delGen, long fieldInfosGen, long docValuesGen) {
+  public SegmentCommitInfo(SegmentInfo info, int delCount, int softDelCount,
+                           long delGen, long nextWriteDelGen,
+                           long fieldInfosGen, long nextWriteFieldInfosGen,
+                           long docValuesGen, long nextWriteDocValuesGen) {
     this.info = info;
     this.delCount = delCount;
     this.softDelCount = softDelCount;
+
+    assert delGen < nextWriteDelGen : delGen + " >= " + nextWriteDelGen;
     this.delGen = delGen;
-    this.nextWriteDelGen = delGen == -1 ? 1 : delGen + 1;
+    this.nextWriteDelGen = nextWriteDelGen;
+
+    assert fieldInfosGen < nextWriteFieldInfosGen : fieldInfosGen + " >= " + nextWriteFieldInfosGen;
     this.fieldInfosGen = fieldInfosGen;
-    this.nextWriteFieldInfosGen = fieldInfosGen == -1 ? 1 : fieldInfosGen + 1;
+    this.nextWriteFieldInfosGen = nextWriteFieldInfosGen;
+
+    assert docValuesGen < nextWriteDocValuesGen : docValuesGen + " >= " + nextWriteDocValuesGen;
     this.docValuesGen = docValuesGen;
-    this.nextWriteDocValuesGen = docValuesGen == -1 ? 1 : docValuesGen + 1;
+    this.nextWriteDocValuesGen = nextWriteDocValuesGen;
   }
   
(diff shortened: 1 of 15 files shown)


dnhatn:
@s1monw Can you please have a look?

jpountz:
Do we actually need to record these generations, or could we initialize them from existing commits at IndexWriter creation time, similarly to how we carry over generation numbers with `OpenMode.CREATE` (https://github.com/apache/lucene-solr/blob/bca22d58e2d126ec6d349d375d3ea028892104e1/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java#L788-L789), and rely on `IndexFileDeleter` to handle the case that IndexWriter was not closed gracefully?

s1monw:
> Do we actually need to record these generations, ....

That was my first reaction too. Yet, it's not enough in the scenario that the test case shows which I think is realistic in order to roll back to an existing commit point. Let me give you an example:

```
iw = new IW(keepAllCommits)
iw.addDocument(<Id:1>)
iw. softUpdateDocument(<Id:1>) -> writes _0.dvd
iw.commit() <commit1, dvGen: -1 // no updates>
iw. softUpdateDocument(<Id:1>)  -> writes _0_1.dvd
iw.commit() <commit2, dvGen: 1>
iw.close();
iw = new IW(<commit1: dvGen -1>) --> IFD updates SegmentCommitInfo to <dvGen: -1, nextDvWriteGen: 2> and deletes <commit2, dvGen:1>
iw.close(); // now only one commit is available 

iw = new IW(<commit1: dvGen -1>) --> IFD updates SegmentCommitInfo to <dvGen: -1, nextDvWriteGen: 1> 
iw. softUpdateDocument(<Id:1>)  -> writes _0_1.dvd and crashes since the file already exists.

```

I hope this makes sense?

lucene/core/src/java/org/apache/lucene/index/SegmentCommitInfo.java
```diff
@@ -91,16 +91,25 @@
    * @param docValuesGen
    *          DocValues generation number (used to name doc-values updates files)
    */
-  public SegmentCommitInfo(SegmentInfo info, int delCount, int softDelCount, long delGen, long fieldInfosGen, long docValuesGen) {
+  public SegmentCommitInfo(SegmentInfo info, int delCount, int softDelCount,
```
s1monw: I think we should use the  `setNext*Gen` setters instead of this. We need them anyway and then we have only one way to set these values. It might also make sense to assert that the next*Gen is > than the actual gen in these setters.

lucene/core/src/java/org/apache/lucene/index/SegmentInfos.java
```diff
@@ -124,7 +124,10 @@
   public static final int VERSION_72 = 8;
   /** The version that recorded softDelCount */
   public static final int VERSION_74 = 9;
-  static final int VERSION_CURRENT = VERSION_74;
+  /** The version that recorded nextWriteDocValuesGen */
+  public static final int VERSION_77 = 10;
```
s1monw: hmm should this be `VERSION_81` instead?

s1monw:
@dnhatn actually making this change:

```diff
diff --git a/lucene/core/src/java/org/apache/lucene/store/FilterDirectory.java b/lucene/core/src/java/org/apache/lucene/store/FilterDirectory.java
index 7b2a6c702d..c2faaf4d0c 100644
--- a/lucene/core/src/java/org/apache/lucene/store/FilterDirectory.java
+++ b/lucene/core/src/java/org/apache/lucene/store/FilterDirectory.java
@@ -117,6 +117,6 @@ public abstract class FilterDirectory extends Directory {
 
   @Override
   public Set<String> getPendingDeletions() throws IOException {
-    return super.getPendingDeletions();
+    return in.getPendingDeletions();
   }
 }
```

seems to have fixed the issue since it should take the deleted files into account. Maybe we should just fix the issue with FilterDirectory @henningandersen found [here](https://github.com/elastic/elasticsearch/pull/40345)

dnhatn:
@s1monw Yeah, the fix from @henningandersen does fix the problem. I am happy to go with that change :).

PR #617, opened 2019-03-22
SOLR-13342: Remove dom4j fromj Solr

risdenk:
* Removes dom4j from core dependencies
* Adds to it velocity contrib
* Replaces dom4j in test with Java xml handling
* Updates licenses/notices that had referenced dom4j historically to their new versions.


risdenk:
Depending if `SOLR-13335 - Upgrade to velocity 2.0 and velocity-tools 3.0` or this goes in first, will need to fix the other one since they both touch Velocity dependencies.

PR #618, opened 2019-03-24
#Lucene-2562: Make Luke a Lucene/Solr Module

mocobeta:
This pull request integrates Luke (https://github.com/DmitryKey/luke) into Lucene as a submodule.
See https://issues.apache.org/jira/browse/LUCENE-2562

Binary distribution package's directory sturucture is:
```
[lucene-9.0.0-SNAPSHOT]$ tree luke
luke/
├── lib
│   ├── log4j-api-2.11.2.jar
│   └── log4j-core-2.11.2.jar
├── lucene-luke-9.0.0-SNAPSHOT.jar
├── luke.bat
└── luke.sh
```

For developers, Luke can be launched by this command after checking out Lucene source code.
```
[lucene-solr]$ ant -f lucene/luke/build.xml run
```


uschindler:
Hi @mocobeta - I have some smaller changes while going through the whole code, but its relatively hard to all describe them in the issue or here. Would it be possible to get write access to your repositoy, so we can both work on finalizing it? Alternatively I can push your branch to Apache's GIT, but you have no push access there, so I'd propose to finalize this on Github and then merge it to Lucene "master".

It's just some smaller changes (wrong use of a charset, an obsolete log4j-1.2-api dependency, and maybe some smaller changes I'd like to do). I can also help with fixing the log4j issues.

mocobeta:
@uschindler I added you as a Collaborator of this repository. Can you please confirm if you can push to here?

FYI, I pushed another branch https://github.com/mocobeta/lucene-solr-mirror/tree/jira/lucene-2562-luke-swing-4-logger, including  my current changes around loggers.

- Remove log4j configuration files
- Configurate log4j programatically
- Remove dependencies for sl4j and log4j-1.2-api

([diff](https://github.com/mocobeta/lucene-solr-mirror/commit/2594f735142d332c9780f3dfc8d1b7e9e7bd4225))

If you agree with the changes above, I will merge it to this branch.

uschindler:
Hi, 
the error message in the main method happens because your LoggerFactory is not initialized BEFORE the first logger is created. main() runs after the "static final" logger. There are 2 ways to fix this:
- Remove logging from the main class
- put the LoggerFactory stuff into a `static { LoggerFactory.initializeContext(); }` *before* the logger declaration in the class.
It's just a chicken&egg problem.

For the tests, the easiest is to just remove all log factories. It doe snot matter where the output goes, let's just use defaults.

uschindler:
Could you just merge the branch? I also commited 2 things here: The removal of log4j1 and I cleaned up the TextAreaPrintStream.

mocobeta:
I merged the two branches and resolved some conflicts. Now, log4j configuration files were gone.

uschindler:
Hi, seems to work. I will fix the log4j error message in a moment. There is now also again obsolete license. I fixed it already locally. We're coming to an end :-)

The tests are also working now, although I have a whitespace in my home folder.

uschindler:
Hi I fixed the warning message. The reason was a bit different than I thougt: You were getting the default LogContext before you configured your own and this caused the error.

I changed it to detect "test mode" (when we run tests, a special system property is set) and use the default context then. I have to check how this behaves in GUIs, but it should work. Otherwise I will push a further change.

uschindler:
I did a bit of better log setup, now it does no longer need a detection of test environment. In the GUI main class the logger factory is initialized with GUI logging.

I was also talking with Robert Muir, he suggested to maybe remove Log4J completely and use java.util.logging instead, so we do not need to add the additional security permissions (we actually want prevent all code from getting context classloader in lucene, as this is a no-go for a library). But for now I am happy.

mocobeta:
Thanks, I fetched the changes. It also works for me!

mocobeta:
Hi @uschindler,

> I was also talking with Robert Muir, he suggested to maybe remove Log4J completely and use java.util.logging instead, so we do not need to add the additional security permissions (we actually want prevent all code from getting context classloader in lucene, as this is a no-go for a library).

I will look into it this weekend. Still I have no idea how I implement TextAreaAppender with java.util.logging for now, please let me know if you have any suggestions.

uschindler:
I fixed the compiler warnings. While doing this (especially because the eclipse compiler complained heavily), I also moved some of the "auxiliary classes" to be inner classes.

Especially with GUIs its a problem if you have multiple classes in the same java file next to each other like: 

```java
public class Main {
 // main class code
}

class Auxiliary {
  // code
}
```

This brings problems with Javac when you change code (it wont compile). This type of classes was used in Java 1.0, but are bad code style nowadays.

It's better to just move those classes as static inner classes:

```java
public class Main {
 // main class code

 static class Auxiliary {
    // code
  }
}
```

Lucene wanted to add a special precommit task to detect that, but it's actually not easy to do with pure regular expressions. I moved some of them, but it looks like there are more (it's often Model classes in Factories or Providers).

uschindler:
> Hi @uschindler,
> 
> > I was also talking with Robert Muir, he suggested to maybe remove Log4J completely and use java.util.logging instead, so we do not need to add the additional security permissions (we actually want prevent all code from getting context classloader in lucene, as this is a no-go for a library).
> 
> I will look into it this weekend. Still I have no idea how I implement TextAreaAppender with java.util.logging for now, please let me know if you have any suggestions.

Hi, I don't think we should do this now. Jul has several problems like locales (the INFO/ERROR/...) messages are translated to your locale and this cannot be changed without using your own formatter. So IMHO, I am happy with the state now, I just wanted to throw some ideas in. Another way would be to only use SLF4J and just implement the "appender" there.

Uwe

P.S.: BTW, implementing logging to the textarea is as simple as with log4j. Just implement your own jul.Handler.

mocobeta:
> I fixed the compiler warnings. While doing this (especially because the eclipse compiler complained heavily), I also moved some of the "auxiliary classes" to be inner classes.

Thank you!  I looked through the code again and moved all other auxiliary classes to static inner classes in the same manner. 🐣

mocobeta:
I made a small change: simply rename the ant target `launch` to `run` (and remove the alias `run`).

mocobeta:
> So IMHO, I am happy with the state now, I just wanted to throw some ideas in. Another way would be to only use SLF4J and just implement the "appender" there.

AFAIK, it does not seem to be easy - for me - to implement the Logger from scratch without any existing logging framework, JUL or Log4J. (I just checked SLF4J documentation and its SimpleLogger binding, this does not support appenders.)
I'd like to keep it as is, if you are OK with that.

uschindler:
Hi @mocobeta,
to me all looks fine. Let's wait a day or two and let us merge it into Lucene master, so others can comment. We can then proceed with more improvements separately.

So I am +1 to merge!

There are some smaller UI issues, especially with font sized and wrapping on Windows with 150% HiDPI: when building a custom analyzer the text of the actual components is huge and it may be better to show tokenfilters as a larger listview instead of one line. If you want to change this over the weekend, go ahead, but that does not limit us, as code is fine:

image

One other small improvement: It shows the Lucene version as title (in the form like saved in index), so it does not contain "dev"/"snapshot"/... Thats fine for the title of main window and about box. In the "AboutForm", it would be good to add another line of text next to credits with the full version string (Solr also shows this): `LucenePackage.get().getImplementationVersion()` - if it's not null (inside IDEs it may be null when no JAR file is built; this command reads the JAR file, so add proper null check). 

I just wanted to say: Eclipse IDE works fine after running "ant eclipse". I also built the lucene.zip/tgz file and was able to start luke from there. forbiddenapis works, "ant veroify" is also fine. Javadocs look OK.

mocobeta:
Thanks for the confirmation.

> There are some smaller UI issues, especially with font sized and wrapping on Windows with 150% HiDPI: 

Through the development of Luke so far I've learned that the fine adjustment for a Swing application is slightly difficult just as web apps' CSS design fix, because the look & feels vary depending on the platform (OS and display resolution). Currently I have Windows tablet (Surface Pro) and Linux box but no Mac OS, so I cannot confirm that component/font size or layout changes do not broke Mac users' experience - right now. Actually it could be the biggest headache to developers  :-/
Anyway I'd like to leave them as they are for present use.

> In the "AboutForm", it would be good to add another line of text next to credits with the full version string (Solr also shows this): LucenePackage.get().getImplementationVersion()

I added "Implementation Version" to the About dialog:

Screenshot from 2019-04-06 17-22-02

> So I am +1 to merge!

Thank you so much, @uschindler, for your support and patience so far! 🙌  🙌 🙌

uschindler:
I noticed, when I open the about box, the text area is by default now scrolled to bottom:

image

Thanks for all your work, it's really fantastic to get Luke finally into Lucene's Default Download package.

mocobeta:
Hi,
I got done with comprehensive interface testing on my Windows/Linux PCs, fixed bugs, made some minor improvements. 
Seems all functionalities are fine.
- Launch from .bat/.sh - OK
- READ access (Open/Browse/Search indexes) - OK
- WRITE access (Create new index, Add/Delete documents)  - OK
- Other index operations (Optimize/Check/Repair indexes) - OK
- Analysis (Built-in/Custom analyzers) - OK
- Logs (File/TextArea) - OK
- Popup dialogs (Help/Error messages) - OK

Also the branch passes unit tests, precommit task.
To me, this branch is ready to be merged 🆗

uschindler:
Hi @mocobeta, 
+1 to merge it soon. Please check your private mail, we have some good news for you.

I will soon push this branch "as is" to Apache's GIT repository, so we have the whole history on that issue also preserved in Apache's GIT. Once we are ready to merge, we can do this together.

IMHO, we should merge it with squashing, so we have a single commit including changes.txt updates at the end. This is also the reason why I pushed the branch with all history to ASF.

mocobeta:
This was merged to Apache Lucene master and branch_8x.

PR #619, opened 2019-03-25
LUCENE-8735: Avoid FileAlreadyExistsException on windows.

henningandersen:
FilterDirectory.getPendingDeletions() did not delegate the call, which
resulted in a new IndexWriter on same directory not considering pending
delete files. This could in turn result in a FileAlreadyExistsException
when running windows.

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -7,6 +7,10 @@ http://s.apache.org/luceneversions
 
 API Changes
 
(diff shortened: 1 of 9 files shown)


s1monw:
this change LGTM - @henningandersen can you add a changes.txt entry. once this is in I will take care of the backporting process.

s1monw:
thanks @henningandersen

PR #622, opened 2019-03-26
Update Tessellator logic to label if triangle edges belongs to the original polygon.

iverase:
This information is then encoded and triangles are decoded using LatLonShape.decodeTriangle.


iverase:
Moved to #771

PR #626, opened 2019-03-28
SOLR-13353: Add SolrCli AuthTool test

risdenk:


solr/core/src/test/org/apache/solr/util/AuthToolTest.java
```diff
@@ -0,0 +1,76 @@
...
+
+  @After
+  public void tearDown() throws Exception {
+    super.tearDown();
+    org.apache.commons.io.FileUtils.deleteDirectory(dir.toFile());
+  }
+
+  @Test
+  public void testEnableAuth() throws Exception {
+    Path solrIncludeFile = Files.createFile(dir.resolve("solrIncludeFile.txt"));
```
janhoy: You could skip the setup() and tearDown() and instead create the tmpdir in the test method. No need to delete it, it should be deleted automatically on test suite termination.

risdenk: Yea I took the same pattern from `UtilsToolTest`. I figured someone will come along and want to add more tests to AuthToolTest and need the directory created.

janhoy: That's ok

PR #627, opened 2019-03-29
LUCENE-8746: Make EdgeTree (aka ComponentTree) support different type of components

iverase:
Refactoring of `EdgeTree` class to support geometry collections


iverase:
I have opened #770 which probably supersedes this one

iverase:
See #878

PR #630, opened 2019-03-30
SOLR-13344: Admin UI inaccessible with RuleBasedAuthorizationPlugin

janhoy:


solr/core/src/java/org/apache/solr/servlet/HttpSolrCall.java
```diff
@@ -550,6 +550,7 @@ public Action call() throws IOException {
   private boolean shouldAuthorize() {
     if(PublicKeyHandler.PATH.equals(path)) return false;
     //admin/info/key is the path where public key is exposed . it is always unsecured
+    if ("/".equals(path)) return false; // Static Admin UI files must always be served
```
gerlowskija: This looks good to me.  I was worried about str comparison on the path being brittle, but I'm reassured that we already do this for authentication.

My only question: In JIRA you linked to the corresponding logic for authentication [here|https://github.com/apache/lucene-solr/blob/6222abf4480ee15bad2e67942263f563ce5ee434/solr/core/src/java/org/apache/solr/servlet/SolrDispatchFilter.java#L468].  For authentication, we check against the paths `/` and `/solr/`, but here you're only checking against `/`.  Is there a reason for that?

(Will probably cross-post this to JIRA as well)

janhoy: Yea. The LoadAdminUIServlet maps to "/" but then there is a rewriteRule to /solr/. But when I tested this patch locally, /solr/ worked as well, that's why I kept it so. But the check should probably be consistent between auth and autz.

PR #631, opened 2019-04-02
LUCENE-8750: implement setMissingValue for ValueSource sortFields

msokolov:


lucene/core/src/java/org/apache/lucene/search/DoubleValuesSource.java
```diff
@@ -428,6 +428,16 @@ public DoubleValuesSource rewrite(IndexSearcher searcher) throws IOException {
       this.producer = producer;
     }
 
+    @Override
+    public void setMissingValue(Object missingValue) {
+      if (missingValue instanceof Number && getType() == Type.CUSTOM) {
+        this.missingValue = missingValue;
```
romseygeek: I think type will always be `Type.CUSTOM` here?

msokolov: hm, I forget why I thought I needed that check. I think you're right.  I'll just double-check and remove.

> The setMissingValue() API is a bit nasty in general, isn't it? Perhaps we should revisit this and clean up type-safety etc in a separate issue.

Yeah -- its ideas about type-safety are a little draconian. I didn't want to start making changes there just yet, but can follow up after.

I also didn't particularly like having to setMissingValue both on the SortField, and internally on the ComparatorSource, but I'm not sure how to clean that up short of eliminating setMissingValue and moving the missingValue to getSortField() and the SortField constructors, which seems drastic. Or maybe the ComparatorSource should not be static and should read the missingValue from its enclosing ValuesSource?

msokolov: OK - PR updated, removing unneeded check for SortField type

romseygeek: > eliminating setMissingValue and moving the missingValue to getSortField()

That's exactly what I think we should do, but it's absolutely a much bigger change!

romseygeek:
Looks good @msokolov, I'll merge this in tomorrow.

msokolov:
Thanks @Alan Woodward <alan@flax.co.uk>

PR #632, opened 2019-04-03
LUCENE-8752: Add a Kuromoji dictionary entry for Japanese new era 令和(Reiwa)

mocobeta:
Corresponding Lucene issue: https://issues.apache.org/jira/browse/LUCENE-8752

As of May 1st, 2019, Japanese era '元号' (Gengo) will be set to '令和' (Reiwa). 

Currently '令和' is splitted up to '令' and '和' by `JapaneseTokenizer`. It should be tokenized as one word so that Japanese texts including era names are searched as users expect. 
Because the default Kuromoji dictionary (mecab-ipadic) is not maintained since 2007, this applies a patch to the source CSV file before building the dictionary.


mocobeta:
This was merged to master & branch_8x.

PR #633, opened 2019-04-03
LUCENE-8753 UniformSplit PostingsFormat

bruno-roustant:


lucene/codecs/src/java/org/apache/lucene/codecs/uniformsplit/IntersectBlockReader.java
```diff
@@ -0,0 +1,315 @@
...
+  public SeekStatus seekCeil(BytesRef text) {
+    throw new UnsupportedOperationException();
+  }
+
+  @Override
+  public void seekExact(BytesRef term, TermState state) {
+    throw new UnsupportedOperationException();
+  }
+
+  //TODO this approach is a hack to reuse complex code in AutomatonTermsEnum.  How to make ATE's logic more reusable?
```
dsmiley: @rmuir this is the hack.  The conundrum I ask in the comment:  "How to make ATE's logic more reusable"?  Your last comment in JIRA seemed to indicate you didn't think AutomatonTermsEnum should be modified at all.  The changes to ATE in this patch are trivial and are avoidable if we choose not to care much about "isLinearState" -- fine by me.  Nonetheless the hack is here (IntersectBlockReader) not ATE, in how ATE is subclassed just to call its logic.  I'm not sure if you've seen the code yet.

lucene/codecs/src/java/org/apache/lucene/codecs/lucene50/DeltaBaseTermStateSerializer.java
```diff
@@ -0,0 +1,206 @@
...
+ * {@link TermState} serializer which encodes each file pointer as a delta relative
+ * to a base file pointer. It differs from {@link Lucene50PostingsWriter#encodeTerm}
+ * which encodes each file pointer as a delta relative to the previous file pointer.
+ * <p>
+ *   It automatically sets the base file pointer to the first valid file pointer for
+ *   doc start FP, pos start FP, pay start FP. These base file pointers have to
+ *   be {@link #resetBaseStartFP() reset} by the caller before starting to write
+ *   a new block.
+ * <p>
+ *   It belongs to the lucene50 package because it accesses the package private
```
dsmiley: @rmuir Any thoughts on this?  Is it fine where it is, in the lucene50 package?  It is not strictly tied to UniformSplit although U.S. uses it; so I suppose it's fine here.  Or perhaps IntBlockTermState could be made completely public, marked as lucene.internal of course, and moved to codecs package?

lucene/codecs/src/java/org/apache/lucene/codecs/uniformsplit/IntersectBlockReader.java
dsmiley: Any opinion @jpountz ?

lucene/codecs/src/java/org/apache/lucene/codecs/lucene50/DeltaBaseTermStateSerializer.java
jpountz: I haven't looked in details, can't it reuse Lucene50PostingsWriter instead?

lucene/codecs/src/java/org/apache/lucene/codecs/uniformsplit/IntersectBlockReader.java
jpountz: Maybe fork the part that you need from AutomatonTermsEnum instead?

lucene/core/src/java/org/apache/lucene/codecs/lucene80/Lucene80Codec.java
```diff
@@ -91,7 +91,11 @@ public Lucene80Codec() {
    *             flushed/merged segments.
    */
   public Lucene80Codec(Mode mode) {
-    super("Lucene80");
+    this("Lucene80", mode);
+  }
+  
+  protected Lucene80Codec(String name, Mode mode) {
+    super(name);
```
jpountz: can you use FilterCodec instead?

bruno-roustant: > Use FilterCodec instead?
Yes! Good point. I'll do that.

> Use Lucene50PostingsWriter instead?
DeltaBaseTermStateSerializer writes each TermState with file pointers delta encoded relatively to a base pointer (which is the block base file pointer). And it does only that thing. This is different from Lucene50PostingsWriter.encodeTerm which writes the TermState with file pointers delta encoded relatively to the previous TermState written.
This logic to delta encode relatively to the base pointer allows us to read TermStates in random access, no sequential reading required. This allows us to make a binary search inside the block itself and read only one TermState.
To summarize: No, we cannot use Lucene50PostingsWriter.encodeTerm. That said DeltaBaseTermStateSerializer could be located elsewhere, the only issue is that it writes IntBlockTermState to stay compatible and still use Lucene50PostingsReader.postings() which requires IntBlockTermState because it casts internally the provided BlockTermState.

lucene/codecs/src/java/org/apache/lucene/codecs/uniformsplit/IntersectBlockReader.java
dsmiley: Okay.  Trade-offs... I usually prefer to re-use code instead of duplicating code.  This would duplicate a lot of complexities in ATE; and you find that better?

lucene/test-framework/src/java/org/apache/lucene/codecs/uniformsplit/UniformSplitTestingCodec.java
```diff
@@ -0,0 +1,41 @@
...
+
+package org.apache.lucene.codecs.uniformsplit;
+
+import org.apache.lucene.codecs.PostingsFormat;
+import org.apache.lucene.codecs.lucene50.Lucene50StoredFieldsFormat;
+import org.apache.lucene.codecs.lucene80.Lucene80Codec;
+
+/**
+ * Codec that chooses {@link UniformSplitPostingsFormat}. The problem with doing
+ * -Dtests.postingsformat=UniformSplitTesting instead is that it's awkward to disable this on some Lucene tests.
```
dsmiley: I think we don't need this testing codec?  I read your javadoc comments here but there is no case yet where we want to disable uniformsplit (or any other particular PostingsFormat for that matter).

lucene/codecs/src/java/org/apache/lucene/codecs/uniformsplit/IntersectBlockReader.java
```diff
@@ -291,23 +293,263 @@ public void seekExact(BytesRef term, TermState state) {
     throw new UnsupportedOperationException();
   }
 
-  //TODO this approach is a hack to reuse complex code in AutomatonTermsEnum.  How to make ATE's logic more reusable?
-  protected static class AutomatonNextTermCalculator extends AutomatonTermsEnum {
+  //This is a copy of AutomatonTermsEnum.  Since it's an inner class, the outer class can
```
dsmiley: @jpountz as you asked, I forked AutomatonTermsEnum into UniformSplit's IntersectBlockReader, and undid the minor change in ATE.

lucene/codecs/src/java/org/apache/lucene/codecs/uniformsplit/IntersectBlockReader.java
jpountz: Yes I like it a bit better, it avoids exposing internals of ATE and makes it easier to evolve both features independently.

lucene/codecs/src/java/org/apache/lucene/codecs/uniformsplit/BlockReader.java
```diff
@@ -0,0 +1,535 @@
...
+      if (readLineInBlock() == null) {
+        // No more terms for the block.
+        return SeekStatus.END;
+      }
+      TermBytes lineTermBytes = blockLine.getTermBytes();
+      BytesRef lineTerm = lineTermBytes.getTerm();
+      assert lineTerm.offset == 0;
+
+      // Equivalent to comparing with BytesRef.compareTo(),
+      // but faster since we start comparing from min(comparisonOffset, suffixOffset).
```
jpountz: you might want to use `Arrays.compareUnsigned` instead of comparing bytes one by one

lucene/core/src/java/org/apache/lucene/codecs/lucene80/Lucene80Codec.java
jpountz: Has this binary search within blocks proved useful? In general I don't have objections to adding new file formats, but I'm a bit worried when they use implementation details of existing file formats instead of using APIs, as this might make these other formats harder to evolve.

bruno-roustant: Two points here:

The binary search inside the block allowed the posting format to improve the seek performance significantly without increasing the size of the FST. Indeed we can have larger blocks for same seek speed, so the FST size is smaller since there are less leaves.

We do use the API for Lucene50PostingsReader.postings(). This is a public method, but unfortunately it internally casts the TermState param to an internal IntBlockTermState. Maybe it would be cleaner for this API to have a public interface for IntBlockTermState, or make it public since its fields are accessed directly. That way we wouldn't have to have a class in this package.

lucene/test-framework/src/java/org/apache/lucene/codecs/uniformsplit/UniformSplitTestingCodec.java
bruno-roustant: Hum, you're right. I'll remove this Codec, yes.

lucene/core/src/java/org/apache/lucene/codecs/lucene80/Lucene80Codec.java
dsmiley: > Maybe it would be cleaner for this API to have a public interface for IntBlockTermState, or make it public since its fields are accessed directly.

+1!

bruno-roustant: @jpountz about (private) IntBlockTermState cast in public Lucene50PostingsReader.postings(), should I create a separate JIRA issue? I think it deserves it.

bruno-roustant: I have created LUCENE-8906 (Lucene50PostingsReader.postings() casts BlockTermState param to private IntBlockTermState) to discuss this specific issue separately.

lucene/codecs/src/java/org/apache/lucene/codecs/uniformsplit/BlockReader.java
bruno-roustant: There is an optimization with comparisonOffset in addition to the regular byte comparison. I have to benchmark to determine if Arrays.compareUnsigned is faster anyway.

dsmiley:
Moved to #828 so I am closing this.

PR #635, opened 2019-04-04
SOLR-13371 improve security chapters in refguide

janhoy:


solr/solr-ref-guide/src/authentication-and-authorization-plugins.adoc
```diff
@@ -1,4 +1,4 @@
-= Authentication and Authorization Plugins
+= Configuring Authentication, Authorization and Auditlogging plugins
```
janhoy: I have not renamed the file itself since it WILL break a ton of URL links and I'm not sure how to efficiently do it in asciidoc. I first named the page "Configuring security plugins in security.json" but then changed the name to make it more searchable and so that it will still pop up for "auth" in the title-search in online refGuide.

However, if we have figured out how to do redirects from old pages to new ones, we could rename the adoc file as well?

solr/solr-ref-guide/src/authentication-and-authorization-plugins.adoc
```diff
@@ -81,7 +84,7 @@ This example also defines `security.json` on the command line, but you can also
 
 [WARNING]
 ====
-Depending on the authentication and authorization plugin that you use, you may have user information stored in `security.json`. If so, we highly recommend that you implement access control in your ZooKeeper nodes. Information about how to enable this is available in the section <<zookeeper-access-control.adoc#zookeeper-access-control,ZooKeeper Access Control>>.
+Whenever you use any security plugins and store `security.json` in ZooKeeper, we highly recommend that you implement access control in your ZooKeeper nodes. Information about how to enable this is available in the section <<zookeeper-access-control.adoc#zookeeper-access-control,ZooKeeper Access Control>>.
```
janhoy: I felt that it is important to secure ZK even if you don't have sensitive data in security.json. For the simple reason that if Zk is wide open then *anyone* could disable or change auth*

solr/solr-ref-guide/src/authentication-and-authorization-plugins.adoc
```diff
@@ -118,20 +121,13 @@ An authentication plugin consists of two parts:
...
-* <<hadoop-authentication-plugin.adoc#hadoop-authentication-plugin,Hadoop Authentication Plugin>>
-* <<jwt-authentication-plugin.adoc#jwt-authentication-plugin,JWT Authentication Plugin>>
+See <<securing-solr.adoc#securing-solr,Securing Solr>> for a list of available authentication plugins.
 
 == Authorization
 
 An authorization plugin can be written for Solr by extending the {solr-javadocs}/solr-core/org/apache/solr/security/AuthorizationPlugin.html[AuthorizationPlugin] interface.
 
-=== Loading a Custom Plugin
+=== Enabling an authorization Plugin
```
janhoy: Sub header more in line with the one for auth. Plugin does not need to be custom...

solr/solr-ref-guide/src/solrcloud-configuration-and-parameters.adoc
```diff
@@ -1,5 +1,5 @@
 = SolrCloud Configuration and Parameters
-:page-children: setting-up-an-external-zookeeper-ensemble, using-zookeeper-to-manage-configuration-files, zookeeper-access-control, collections-api, parameter-reference, command-line-utilities, solrcloud-with-legacy-configuration-files, configsets-api
+:page-children: setting-up-an-external-zookeeper-ensemble, using-zookeeper-to-manage-configuration-files, collections-api, parameter-reference, command-line-utilities, solrcloud-with-legacy-configuration-files, configsets-api
```
janhoy: A page can only be child of ONE parent. I feel zookeeper-ACL fits better under securing-solr than here

solr/solr-ref-guide/src/securing-solr.adoc
```diff
@@ -19,16 +19,52 @@
...
-* <<enabling-ssl.adoc#enabling-ssl,Enabling SSL>>
-* If using SolrCloud, <<zookeeper-access-control.adoc#zookeeper-access-control,ZooKeeper Access Control>>
-* <<audit-logging.adoc#audit-logging,Audit logging>> for recording an audit trail
-
 [WARNING]
 ====
 No Solr API, including the Admin UI, is designed to be exposed to non-trusted parties. Tune your firewall so that only trusted computers and people are allowed access. Because of this, the project will not regard e.g., Admin UI XSS issues as security vulnerabilities. However, we still ask you to report such issues in JIRA.
 ====
+
+=== Encryption with TLS (SSL) certificates
```
janhoy: Here I skip level-2 headers since that means page-break in the PDF. We get a compile-warning but it was the only way I could get the more detailed securing-solr page to stick together. @ctargett any other tricks we could use instead?

janhoy:
Friendly reminder @ctargett (or anyone else who care about ref guide 😊)

janhoy:
I'm prepared to merge this in time for the 8.1 guide. Any last comments before I merge?

solr/solr-ref-guide/src/authentication-and-authorization-plugins.adoc
ctargett: I prefer keeping the "auth" in the title, for the same reasons. I don't think we've worked out how to do redirects well - we don't really even have a functional 404 page. It's fine to keep the filename the same since it includes most of the reworked title, in my opinion, so I'm not worried about keeping it the way it is already. (We should figure out redirects, but it doesn't have to be done here.)

Do you mean to use "Auditlogging" in the title though? It's a bit inconsistent in these edits - sometimes it's "auditlogging" and sometimes "audit logging". I feel like it should be "audit logging" wherever possible. An exception would be when referring to a class, and then I'll guess it would be something like "AuditLoggingPlugin".

I think we could drop "Plugins" from the page title to be honest. The page is about configuring authentication, authorization and audit logging - the fact that's done with plugins doesn't need to be in the page title IMO. If you decide to keep "plugins" in the title, though, it should be capitalized so we maintain our standard of headline case for page and section titles.

solr/solr-ref-guide/src/securing-solr.adoc
ctargett: It's not just a compile warning - we would fail the build for sections out of order if we could. Even though I know it breaks the page in the PDF, we need to use the levels in order.

After 8.1 I intend to propose we drop the PDF as our main release artifact, which means we could change the format type we use, which would totally change all of this page break behavior (the "article" format doesn't page-break any sections at all).

solr/solr-ref-guide/src/securing-solr.adoc
```diff
@@ -17,18 +17,44 @@
...
+
+When planning how to secure Solr, you should consider which of the available features or approaches are right for you:
+
+=== Encryption with TLS (SSL) certificates
+
+Ecrypting traffic to/from Solr and between Solr nodes prevents sensitive data to be leaked out on the network. TLS is also normally a requirement to prevent credential sniffing when using Authentication.
+
+See the page <<enabling-ssl.adoc#enabling-ssl,Enabling TLS (SSL)>> for details.
+
+=== Authentication, Authorization and Audit logging
```
ctargett: "logging" needs to be capitalized.

solr/solr-ref-guide/src/securing-solr.adoc
```diff
@@ -17,18 +17,44 @@
...
+
+See chapter <<authentication-and-authorization-plugins.adoc#authentication-and-authorization-plugins,Configuring Authentication, Authorization and Auditlogging plugins>> to learn how to work with the `security.json` file.
```
ctargett: This is an example of the inconsistency I mentioned in my comment on the `authentication-and-authorization-plugins.adoc` page. The title of the section uses "Audit logging", but right below it the link to the page uses `Auditlogging`.

solr/solr-ref-guide/src/securing-solr.adoc
```diff
@@ -17,18 +17,44 @@
...
+
+[#securing-solr-auth-plugins]
+==== Authentication plugins
```
ctargett: "plugins" here needs to be capitalized.

solr/solr-ref-guide/src/securing-solr.adoc
```diff
@@ -17,18 +17,44 @@
...
+
+Authentication makes sure you know the identity of your users. The authentication plugins shipping with Solr are:
+
+* <<kerberos-authentication-plugin.adoc#kerberos-authentication-plugin,Kerberos Authentication Plugin>>
+* <<basic-authentication-plugin.adoc#basic-authentication-plugin,Basic Authentication Plugin>>
+* <<hadoop-authentication-plugin.adoc#hadoop-authentication-plugin,Hadoop Authentication Plugin>>
+* <<jwt-authentication-plugin.adoc#jwt-authentication-plugin,JWT Authentication Plugin>>
+
+==== Authorization plugins
```
ctargett: Capitalize "plugins"

solr/solr-ref-guide/src/securing-solr.adoc
```diff
@@ -17,18 +17,44 @@
...
+
+Authorization makes sure that only users with the necessary roles/permissions can access any given resource. The authorization plugins shipping with Solr are:
+
+* <<rule-based-authorization-plugin.adoc#rule-based-authorization-plugin,Rule-Based Authorization Plugin>>
+
+==== Audit logging plugins
```
ctargett: Capitalize "logging" and "plugins".

solr/solr-ref-guide/src/securing-solr.adoc
```diff
@@ -17,18 +17,44 @@
...
+
+Audit logging will record an audit trail of incoming reqests to your cluster, such as users being denied access to admin APIs. Learn more about audit logging and how to implement an audit logger plugin here in the <<audit-logging.adoc#audit-logging,Audit Logging>> chapter.
+
+=== Securing Zookeeper traffic
```
ctargett: Capitalize "traffic"

solr/solr-ref-guide/src/solrcloud-configuration-and-parameters.adoc
ctargett: Yes, I agree, it's a good move.

solr/solr-ref-guide/src/authentication-and-authorization-plugins.adoc
```diff
@@ -17,17 +17,17 @@
 // specific language governing permissions and limitations
 // under the License.
 
-Solr has security frameworks for supporting authentication and authorization of users. This allows for verifying a user's identity and for restricting access to resources in a Solr cluster.
+Solr has security frameworks for supporting authentication, authorization and auditing of users. This allows for verifying a user's identity and for restricting access to resources in a Solr cluster.
 
-Solr includes some plugins out of the box, and additional plugins can be developed using the authentication and authorization frameworks described below.
+Solr includes some plugins out of the box, and additional plugins can be developed using the authentication, authorization and auditlogging frameworks described below.
```
ctargett: Here's another example of the inconsistency using "audit logging" or "auditlogging".

solr/solr-ref-guide/src/authentication-and-authorization-plugins.adoc
ctargett: True, good point. +1

solr/solr-ref-guide/src/authentication-and-authorization-plugins.adoc
```diff
@@ -103,7 +106,7 @@ An authentication plugin consists of two parts:
 . Server-side component, which intercepts and authenticates incoming requests to Solr using a mechanism defined in the plugin, such as Kerberos, Basic Auth or others.
 . Client-side component, i.e., an extension of `HttpClientConfigurer`, which enables a SolrJ client to make requests to a secure Solr instance using the authentication mechanism which the server understands.
 
-=== Enabling a Plugin
+=== Enabling an authentication Plugin
```
ctargett: Capitalize "authentication"

solr/solr-ref-guide/src/authentication-and-authorization-plugins.adoc
ctargett: +1 to the change, but capitalize "authorization".

solr/solr-ref-guide/src/authentication-and-authorization-plugins.adoc
```diff
@@ -118,20 +121,13 @@ An authentication plugin consists of two parts:
...
 
-=== Available Authentication Plugins
-
-Solr has the following implementations of authentication plugins:
-
-* <<kerberos-authentication-plugin.adoc#kerberos-authentication-plugin,Kerberos Authentication Plugin>>
-* <<basic-authentication-plugin.adoc#basic-authentication-plugin,Basic Authentication Plugin>>
-* <<hadoop-authentication-plugin.adoc#hadoop-authentication-plugin,Hadoop Authentication Plugin>>
-* <<jwt-authentication-plugin.adoc#jwt-authentication-plugin,JWT Authentication Plugin>>
+See <<securing-solr.adoc#securing-solr,Securing Solr>> for a list of available authentication plugins.
```
ctargett: I'm not sure about this change (and note you did it for the authorization section too, so this comment applies there too). The pages for the plugins are sub-pages of this one, so navigation-wise it makes more sense to bring people deeper into the hierarchy from here instead of bouncing them one level up just so they can go a level back down.

If the change was made to make it easier to maintain the list and not have to repeat it across pages, one option would be to insert a tag around the original list, and then define an include directive here to pull the tagged section from the other page to this one. That allows maintenance in a single place but inclusion wherever it's relevant.

The syntax is explained here: https://asciidoctor.org/docs/user-manual/#by-tagged-regions. Insert tags like this:

```
// tag::<tag-name>[]
content you want to be tagged
// end::<tag-name>[]
```
Wherever you want to put the tagged content, you use a syntax like this:

```
include::<filename.adoc>[tag=<tag-name>]
```

solr/solr-ref-guide/src/authentication-and-authorization-plugins.adoc
```diff
@@ -152,11 +148,14 @@ All of the content in the `authorization` block of `security.json` would be pass
 The authorization plugin is only supported in SolrCloud mode. Also, reloading the plugin isn't yet supported and requires a restart of the Solr installation (meaning, the JVM should be restarted, not simply a core reload).
 ====
 
-=== Available Authorization Plugins
+See <<securing-solr.adoc#securing-solr,Securing Solr>> for a list of available authorization plugins.
+
+[#configuring-audit-logging]
+== Audit logging
```
ctargett: Capitalize "logging"

ctargett:
Overall I'm +1 for the changes, but had a few comments before it's committed. Thanks for doing this, and sorry it took me so long to review it.

janhoy:
I pushed a commit that I think addresses all of your comments above (c600f7ae9ce950dfee659eefe74a5c5df409be0a). Feel free to merge or I can merge if there are no further comments.

ctargett:
I don't know how to update PRs from forks outside the main lucene-solr repo or I would fix this myself, but there are 2 problems causing the build to fail. Both of these are in `authentication-and-authorization-plugins.adoc`:

* The change to the section heading from "Authentication Plugins" to "Authentication" duplicates section title (it appears in `solr-control-script-reference.adoc`), so one of them needs to change.
* I recommended the tag & include macro for the plugin lists but when you did it, you kept the link to the parent Securing Solr page, which triggered this bug: https://github.com/asciidoctor/asciidoctor/issues/2200. Essentially, a page can't have both an `include` to bring in a section of a page and also link to that same page. The bug is technically fixed, but we can't use that version yet due to SOLR-12786. Consequently the link to the Securing Solr page in the Audit Logging section is actively broken. Since the link only goes to a sentence or two that point people at the Audit Logging page, why not just point people at the Audit Logging page?

PR #636, opened 2019-04-04
SOLR-10409 - Added support for rows=x in /export to limit number of docs exported (With unit tests)

b-pradeeprao:
This lets user add rows=x  parameter to request to limit number of docs they get from /export query

rows=x
{
x=0 => returns numFound=0 and no docs in results - this is existing behavior (I guess, unintended)
x=undefined or x>=numFound => will not have any effect on existing behavior 
x < numFound => will return first x docs only (in specified sorted order ) in results.
}

Note Reg. Unit Testing.
Given **DOCUMENT_BATCH_SIZE = 30000;** in **ExportWriter.java:87** , ideal testing should be done on index with >30k docs. I have tested all edge cases with the batch_sizes to 100 and 1.

### DIFF
--- a/solr/core/src/java/org/apache/solr/handler/export/ExportWriter.java
+++ b/solr/core/src/java/org/apache/solr/handler/export/ExportWriter.java
@@ -91,6 +91,7 @@ public class ExportWriter implements SolrCore.RawWriter, Closeable {
   final SolrQueryResponse res;
   FieldWriter[] fieldWriters;
   int totalHits = 0;
+  int maxDocs=0;
   FixedBitSet[] sets = null;
   PushWriter writer;
   private String wt;
@@ -188,8 +189,16 @@ public class ExportWriter implements SolrCore.RawWriter, Closeable {
       }
     }
     SolrParams params = req.getParams();
+    
+    maxDocs = totalHits;
+    
+    if ( params.get("rows") != null) {
+      int rows = Integer.parseInt(params.get("rows"));
+      //rows=0 has an existing side effect of returning numFound=0 and no docs in /export component. So following logic effectively kicks in when rows>0
+      maxDocs = ( rows >= 0 && rows < totalHits ) ? rows:totalHits;
+    }
+    
     String fl = params.get("fl");
-
     String[] fields = null;
 
     if (fl == null) {
@@ -251,13 +260,13 @@ public class ExportWriter implements SolrCore.RawWriter, Closeable {
         destinationArr[++outDocsIndex] = s;
       }
     }
-
     return outDocsIndex;
   }
 
-  protected void addDocsToItemWriter(List<LeafReaderContext> leaves, IteratorWriter.ItemWriter writer, SortDoc[] docsToExport, int outDocsIndex) throws IOException {
+  protected void addDocsToItemWriter(List<LeafReaderContext> leaves, IteratorWriter.ItemWriter writer, SortDoc[] docsToExport, int outDocsIndex, int numDocsToExclude) throws IOException {
+    int indexLimit = numDocsToExclude>=0 ? 0:Math.abs(numDocsToExclude);
     try {
-      for (int i = outDocsIndex; i >= 0; --i) {
+      for (int i = outDocsIndex; i >= indexLimit; --i) {
         SortDoc s = docsToExport[i];
         writer.add((MapWriter) ew -> {
           writeDoc(s, leaves, ew);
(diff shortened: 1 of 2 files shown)


solr/core/src/java/org/apache/solr/handler/export/ExportWriter.java
```diff
@@ -91,6 +91,7 @@
   final SolrQueryResponse res;
   FieldWriter[] fieldWriters;
   int totalHits = 0;
+  int maxDocs=0;
```
diegoceccarelli: minor style: readd the spaces before and after the `=`

solr/core/src/java/org/apache/solr/handler/export/ExportWriter.java
```diff
@@ -188,8 +189,16 @@ public void write(OutputStream os) throws IOException {
       }
     }
     SolrParams params = req.getParams();
+    
+    maxDocs = totalHits;
+    
+    if ( params.get("rows") != null) {
```
diegoceccarelli: you can use 
```
Integer rows = params.getInt("rows");
if (rows != null) {
   maxDocs = ( rows >= 0 && rows < totalHits ) ? rows : totalHits;
}
```

solr/core/src/java/org/apache/solr/handler/export/ExportWriter.java
```diff
@@ -188,8 +189,16 @@ public void write(OutputStream os) throws IOException {
...
+      int rows = Integer.parseInt(params.get("rows"));
+      //rows=0 has an existing side effect of returning numFound=0 and no docs in /export component. So following logic effectively kicks in when rows>0
+      maxDocs = ( rows >= 0 && rows < totalHits ) ? rows:totalHits;
+    }
```
diegoceccarelli: what happens when `rows` is not specified?

solr/core/src/java/org/apache/solr/handler/export/ExportWriter.java
```diff
@@ -251,13 +260,13 @@ protected int transferBatchToArrayForOutput(SortQueue queue, SortDoc[] destinati
         destinationArr[++outDocsIndex] = s;
       }
     }
-
     return outDocsIndex;
   }
 
-  protected void addDocsToItemWriter(List<LeafReaderContext> leaves, IteratorWriter.ItemWriter writer, SortDoc[] docsToExport, int outDocsIndex) throws IOException {
+  protected void addDocsToItemWriter(List<LeafReaderContext> leaves, IteratorWriter.ItemWriter writer, SortDoc[] docsToExport, int outDocsIndex, int numDocsToExclude) throws IOException {
```
diegoceccarelli: you are changing the interface here, are you sure you really really need to do that?

solr/core/src/test/org/apache/solr/handler/export/TestExportWriter.java
```diff
@@ -347,7 +347,288 @@ public void testIndexOrder() throws Exception {
...
+    String descExpectedResult = resultPrefix;   
+    int id = maxDocs;
+    for(int i=expRows-1; i>0; i--) {
+      descExpectedResult+= "{\n\"id\":\"" + --id + "\"}\n" ;
+    }
+    descExpectedResult+= "{\n\"id\":\"" + --id + "\"}]}}" ;
+    
+    //in reverse indexing order
+    String s2 =  h.query(req("q", "*:*", "qt", "/export", "fl", "id", "sort", "intdv asc", "rows", Integer.toString(rows)));
+    assertJsonEquals(s2, descExpectedResult);
```
diegoceccarelli: This `assertJsonEquals` just do a String comparison, and it has been introduced by the exporter tests, can i suggest instead to use `assertJQ` instead? 

examples: 

https://github.com/apache/lucene-solr/blob/0bd1911db6de9f38f74fc61398bd1fc3f42037a2/solr/core/src/test/org/apache/solr/update/processor/TestDocBasedVersionConstraints.java#L65

audun:
This functionality would be useful for me too. However, I think rows=0 should still result in 0 rows, because that is how the rows parameter in other solr queries work.

I agree with all of Diego's comments. @b-pradeeprao If you don't mind, I can fix them and perhaps we can get it merged?

b-pradeeprao:
@audun as mentioned above, that indeed is the behavior here. 

**x=0 => returns numFound=0 and no docs in results - this is existing behavior (I guess, unintended)**

Also, I've just seen all these comments - let me try to fix them -

PR #637, opened 2019-04-07
LUCENE-8754: Prevent ConcurrentModificationException in SegmentInfo

s1monw:
In order to prevent ConcurrentModificationException this change makes
an unmodifiable copy on write for all maps in SegmentInfo. MergePolicies
can access these maps without synchronization and cause exceptions if
it's modified in the merge thread.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/SegmentInfo.java
+++ b/lucene/core/src/java/org/apache/lucene/index/SegmentInfo.java
@@ -335,7 +335,7 @@ public final class SegmentInfo {
     // the same instance. Technically that's an unsafe publication but IW design would require
     // significant changes to prevent this. On the other hand, since we expose the map in getAttributes()
     // it's a good design to make it unmodifiable anyway.
-    attributes = Collections.unmodifiableMap(new HashMap<>(Objects.requireNonNull(newMap)));
+    attributes = Collections.unmodifiableMap(newMap);
     return oldValue;
   }
   
--- a/lucene/core/src/java/org/apache/lucene/index/SegmentInfo.java
+++ b/lucene/core/src/java/org/apache/lucene/index/SegmentInfo.java
@@ -325,6 +325,7 @@ public final class SegmentInfo {
    * <p>
    * If a value already exists for the field, it will be replaced with the new
    * value.
+   * This method make a copy on write for every attribute change.
    */
   public String putAttribute(String key, String value) {
     HashMap<String, String> newMap = new HashMap<>(attributes);
--- a/lucene/core/src/java/org/apache/lucene/index/SegmentInfo.java
+++ b/lucene/core/src/java/org/apache/lucene/index/SegmentInfo.java
@@ -20,6 +20,7 @@ package org.apache.lucene.index;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Locale;
 import java.util.Map;
@@ -68,7 +69,7 @@ public final class SegmentInfo {
 
   private Map<String,String> diagnostics;
   
-  private final Map<String,String> attributes;
+  private Map<String,String> attributes;
 
   private final Sort indexSort;
 
@@ -86,7 +87,7 @@ public final class SegmentInfo {
   Version minVersion;
 
   void setDiagnostics(Map<String, String> diagnostics) {
-    this.diagnostics = Objects.requireNonNull(diagnostics);
+    this.diagnostics = Collections.unmodifiableMap(new HashMap<>(Objects.requireNonNull(diagnostics)));
   }
 
   /** Returns diagnostics saved into the segment when it was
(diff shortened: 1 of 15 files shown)


s1monw:

lucene/core/src/java/org/apache/lucene/index/SegmentInfo.java
```diff
@@ -326,7 +327,16 @@ public String getAttribute(String key) {
...
-    return attributes.put(key, value);
+    HashMap<String, String> newMap = new HashMap<>(attributes);
+    String oldValue = newMap.put(key, value);
+    // we make a full copy of this to prevent concurrent modifications to this in the toString method
+    // this method is only called when a segment is written but the SegmentInfo might be exposed
+    // in running merges which can cause ConcurrentModificationExceptions if we modify / share
+    // the same instance. Technically that's an unsafe publication but IW design would require
+    // significant changes to prevent this. On the other hand, since we expose the map in getAttributes()
+    // it's a good design to make it unmodifiable anyway.
+    attributes = Collections.unmodifiableMap(new HashMap<>(Objects.requireNonNull(newMap)));
```
jpountz: the requireNonNull call doesn't help here?

lucene/core/src/java/org/apache/lucene/index/SegmentInfo.java
```diff
@@ -68,7 +69,7 @@
 
   private Map<String,String> diagnostics;
   
-  private final Map<String,String> attributes;
+  private Map<String,String> attributes;
```
jpountz: Do we need to make it volatile so that toString sees the up-to-date reference to the attributes?

lucene/core/src/java/org/apache/lucene/index/SegmentInfo.java
```diff
@@ -111,12 +112,12 @@ public SegmentInfo(Directory dir, Version version, Version minVersion, String na
...
     this.isCompoundFile = isCompoundFile;
     this.codec = codec;
-    this.diagnostics = Objects.requireNonNull(diagnostics);
+    this.diagnostics = Collections.unmodifiableMap(new HashMap<>(Objects.requireNonNull(diagnostics)));
     this.id = id;
     if (id.length != StringHelper.ID_LENGTH) {
       throw new IllegalArgumentException("invalid id: " + Arrays.toString(id));
     }
-    this.attributes = Objects.requireNonNull(attributes);
+    this.attributes = Collections.unmodifiableMap(new HashMap<>(Objects.requireNonNull(attributes)));
```
jpountz: Nit: now that we are copying the input attributes, we should change the few call sites that pass a `new HashMap<>()` as attributes to pass a `Collections.emptyMap()` instead?

s1monw: do we have some? I did a quick look and I didn't find any

lucene/core/src/java/org/apache/lucene/index/SegmentInfo.java
s1monw: I don't think this is helping anybody. This entire class is mutable and accessed without locking. I think we are making the situation as good as it gets without lock since we always safe-publish with a new instance. There are many other things like `maxDoc` and `isCompoundFile` that have the same issues.

lucene/core/src/java/org/apache/lucene/index/SegmentInfo.java
s1monw: sure

s1monw:
@jpountz I pushed new commits

lucene/core/src/java/org/apache/lucene/index/SegmentInfo.java
```diff
@@ -326,7 +327,16 @@ public String getAttribute(String key) {
...
+    HashMap<String, String> newMap = new HashMap<>(attributes);
+    String oldValue = newMap.put(key, value);
+    // we make a full copy of this to prevent concurrent modifications to this in the toString method
+    // this method is only called when a segment is written but the SegmentInfo might be exposed
+    // in running merges which can cause ConcurrentModificationExceptions if we modify / share
+    // the same instance. Technically that's an unsafe publication but IW design would require
+    // significant changes to prevent this. On the other hand, since we expose the map in getAttributes()
+    // it's a good design to make it unmodifiable anyway.
+    attributes = Collections.unmodifiableMap(newMap);
+    return oldValue;
```
mikemccand: Can you fix the javadocs for this class to make it clear that every `putAttribute` call causes whole attributes `HashMap` to be cloned?

s1monw:
@mikemccand pushed changes

lucene/core/src/java/org/apache/lucene/index/SegmentInfo.java
jpountz: IndexWriter#mergeInit, IndexWriter#addIndexes, IndexSplitter#split seem to do that

uschindler: One thing: once master is on Java 11, this should use: https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Map.html#copyOf(java.util.Map)

We may simplify more code doing that as Java 11 has now unmodifiable maps using a static initializer like Guava once had with ImmutableMap. Same for lists and sets.

s1monw: ++ I will be able to help here

PR #638, opened 2019-04-08
LUCENE-8756: Fix MLT like text with custom frequencies

ollik1:
When an analyzer with custom term frequencies is used with MLT like
texts, the custom term frequencies are incorrectly omitted and a fixed
frequency of 1 is used instead.

This commit fixes the issue by using `TermFrequencyAttribute` to get
the term frequencies instead of using fixed 1. Also adds test cases
for them mentioned issue.


lucene/queries/src/test/org/apache/lucene/queries/mlt/TestMoreLikeThis.java
```diff
@@ -427,5 +438,70 @@ public void testMultiFieldShouldReturnPerFieldBooleanQuery() throws Exception {
...
+        return new TokenStreamComponents(tokenizer, foo(filt));
+      }
+
+      TokenStream foo(TokenStream input) {
+        return new TokenFilter(input) {
+
+          @Override
+          public boolean incrementToken() throws IOException {
+            final CharTermAttribute termAtt = addAttribute(CharTermAttribute.class);
+            final TermFrequencyAttribute tfAtt = addAttribute(TermFrequencyAttribute.class);
```
mikemccand: Normally `addAttribute` calls must happen at the `TokenFilter` construction time -- I think the test only works because the `AttributeFactory` has already added these attributes.  Maybe move them up above?

lucene/queries/src/test/org/apache/lucene/queries/mlt/TestMoreLikeThis.java
```diff
@@ -427,5 +438,70 @@ public void testMultiFieldShouldReturnPerFieldBooleanQuery() throws Exception {
...
+    Analyzer analyzer = new Analyzer() {
+
+      @Override
+      protected TokenStreamComponents createComponents(String fieldName) {
+        MockTokenizer tokenizer = new MockTokenizer(MockTokenizer.WHITESPACE, false, 100);
+        MockTokenFilter filt = new MockTokenFilter(tokenizer, MockTokenFilter.EMPTY_STOPSET);
+        return new TokenStreamComponents(tokenizer, foo(filt));
+      }
+
+      TokenStream foo(TokenStream input) {
```
mikemccand: Maybe give this method a better name?  Maybe `addCustomTokenFilter` or something?

lucene/queries/src/test/org/apache/lucene/queries/mlt/TestMoreLikeThis.java
ollik1: 👍

lucene/queries/src/test/org/apache/lucene/queries/mlt/TestMoreLikeThis.java
ollik1: My bad, forgot this

PR #639, opened 2019-04-10
Solve the problem of highlighting Chinese inaccurately.

cuckooM:
I found a bug. When I am highlighting Chinese, extra phrases are highlighted. 
I solved it according to my understanding. 
I added a test case to simulate the problem when highlighting Chinese text.


janhoy:
Thanks for contributing. You way want to discuss your problem in the mailing list and confirm it is a bug, and then invite developers to take a look at your PR.

PR #640, opened 2019-04-10
LUCENE-8671: Introduce Reader attributes

s1monw:
Reader attributes allows a per IndexReader configuration of codec internals.
For instance this allows a per reader configuration if FSTs are loaded into memory or are left
on disk.


s1monw:
it's still a bit rough and misses javadocs etc. but I wanted to get the idea out.

lucene/core/src/java/org/apache/lucene/codecs/blocktree/BlockTreeTermsReader.java
```diff
@@ -75,6 +74,41 @@
...
+     * Automatically make the decision if FSTs are read from disk depending if the segment read from an MMAPDirectory
+     * An exception is made for ID fields in an IndexWriter context which are always loaded into memory.
+     */
+    AUTO
+  }
+
+  /** Attribute key for fst mode. */
+  public static final String FST_MODE_KEY = "blocktree.terms.fst.mode";
+  /** Attribute key for the ID field. ID fields are forced on-heap if set **/
+  public static final String ID_FIELD = "blocktree.terms.fst.id_field"; // TODO should this only happen in a writer context?
```
mikemccand: It's a little odd to have the notion of an "id field" down this low in Lucene -- maybe instead we could parameterize the keys by field name?  So e.g. `blocktree.terms.fst.$field` would be set to the `FSTLoadMode` enum values?  This would also give users per-field control.

lucene/core/src/java/org/apache/lucene/codecs/blocktree/BlockTreeTermsReader.java
```diff
@@ -156,6 +190,8 @@ public BlockTreeTermsReader(PostingsReaderBase postingsReader, SegmentReadState
       seekDir(termsIn);
       seekDir(indexIn);
 
+      FSTLoadMode fstLoadMode = FSTLoadMode.valueOf(state.readerAttributes.getOrDefault(FST_MODE_KEY, FSTLoadMode.AUTO.toString()));
```
mikemccand: Do we get a nice exception if the string is invalid here?  Can we fix the exception to state which key we were looking up when we got the invalid `FSTLoadMode` enum?

lucene/core/src/java/org/apache/lucene/codecs/blocktree/BlockTreeTermsReader.java
s1monw: yeah I had this before, I can go back to this for sure.

lucene/core/src/java/org/apache/lucene/codecs/blocktree/BlockTreeTermsReader.java
s1monw: I added one.

s1monw:
@jpountz thanks for the review @mikemccand I pushed updates for your comments

luyuncheng:
LGTM

PR #647, opened 2019-04-15
SOLR-12638: Added docs in the ref-guide for nested atomic updates

moshebla:


moshebla:
This PR adds docs for nested/partial updates #455 .

solr/solr-ref-guide/src/indexing-nested-documents.adoc
```diff
@@ -51,6 +51,13 @@ With the exception of in-place updates, the whole block must be updated or delet
   Consequently, the field need not be defined in the schema and probably shouldn't be as it would be confusing.
   There is no child document field type, at least not yet.
 
+[NOTE]
```
dsmiley: Oh, I didn't know this or forgot.  Ugh; I wish this weren't so.  Since \_route_  is a required parameter for this feature, is the stored/docValue of \_root_  actually truly necessary?  Hmm; do we at least throw an error if this is neglected when a user tries to do a partial update to a child doc?

Any way, I would put this as a sub-bullet of the existing bullet of `\_root_`.  Having written a book, I know it's tempting to put all that FYI info as call-out tips/notes but it's usually better to integrate the information into the text.

solr/solr-ref-guide/src/updating-parts-of-documents.adoc
```diff
@@ -103,6 +103,108 @@ The resulting document in our collection will be:
 }
 ----
 
+=== Updating Child Documents
+
+Solr supports modifying, adding and removing child documents as part of atomic updates. +
+Schema and configuration requirements are detailed in
+<<indexing-nested-documents#schema-configuration, Indexing Nested Documents>>. +
```
dsmiley: -- and also updating-parts-of-documents.adoc heading "Field Storage", which I would even list _first_ since what we're describing here fits underneath that conceptually.

solr/solr-ref-guide/src/updating-parts-of-documents.adoc
```diff
@@ -103,6 +103,108 @@ The resulting document in our collection will be:
...
+           },
+           {
+            "id":"mydoc3",
+            "color":"blue",
+            "size": ["M"]
+           }]
+}
+----
+
+Documents inside nested structures can also be updated, provided `\_route_` is set to the root document's ID
```
dsmiley: RE "provided"... that wording suggests \_route_ is optional.  It's NOT, right?  Then instead say nothing or just remind the user what it should be.

solr/solr-ref-guide/src/indexing-nested-documents.adoc
moshebla: > /home/internet/Projects/lucene-solr/solr/package/solr-7.7.2-SNAPSHOT.tgz

Currently we only throw an error if the wrong route param was supplied when updating a non root doc.
Should we throw an error when no _route_ param is supplied?
I could work on that if it is deemed necessary.
WDYT?

dsmiley: I think it's okay to include a small modification here to error out if `_route_` is not provided.  I'm assuming it's required only for atomic updates of child documents -- that is, when the user provides a child document as the root of the update.  I don't think it is required when updating a root doc to manipulate (add/update/delete) a child doc?  I wonder if we can tell the difference -- when the user makes this mistake.  And if `_route_` is provided unnecessarily, ideally we would verify that its value is correct.

What I wonder about is what happens when `_root_` is not stored/docValues?

moshebla: > And if `_route_` is provided unnecessarily, ideally we would verify that its value is correct.
> 
> What I wonder about is what happens when `_root_` is not stored/docValues?

Well, as long as `_root_` is either stored or docValues, we should be able to verify whether the correct `_route_` was provided.

moshebla: Since by default if a doc was not found, [Solr  creates it](https://github.com/moshebla/lucene-solr/blob/SOLR-12638-docs/solr/core/src/java/org/apache/solr/update/processor/DistributedUpdateProcessor.java#L677), We can only tell if the user omitted the `_route_` param if the doc was routed to the correct shard.
I think it would be a lot more simple to require `_route_` for **all** nested atomic updates.
@dsmiley ,
WDYT?

dsmiley: I think the `_route_` param should not be required for all atomic updates.  It's kind of an edge case and I'd rather not burden the other case with something that's superfluous.

moshebla: Sorry,
perhaps I should have been clearer.
I meant for all nested atomic updates.
Only if the document which is being updated is a nested document, the `_route_` param should be mandatory.

dsmiley: Is stored/docValues on  the`_root_` field still required when the atomic update provides a root doc as the root of the update?  Also, I think `useDocValuesAsStored` is best set to "false", not that it's a big deal.  We probably didn't test that.

moshebla: Well, we probably could get away with using _route_ to figure out _root_, but that would prevent any method of validating the user passed the right _route_ param.

dsmiley: > I meant for all nested atomic updates. Only if the document which is being updated is a nested document, the _route_ param should be mandatory.

This is confusing to talk about.  I view "a nested document" as a nested tree of documents, both the root and child and all.  Any way, essentially, what I mean is that we probably shouldn't require this param if it's superfluous, and it superfluous if the root of an update is a root document.

dsmiley: > Well, we probably could get away with using route to figure out root, but that would prevent any method of validating the user passed the right route param.

Perhaps we can't validate it then; okay.

dsmiley:
so I think at this point it's simply some doc tweaks and no changes to the code?  Might you be able to do this in a day so I can incorporate it into 8.1?  The branch will be cut in some hours from how but I could get documentation like this in.

moshebla:
Sure thing, I'll do the docs right away and start making the code changes afterwards.

moshebla:
Pushed a commit which addresses ref-guide docs structure related requests.

dsmiley:
I committed this; though made some small edits.  Notably I left out in the first example to pass `_route_` as it was unnecessary in that one.
Thanks @moshebla !

PR #648, opened 2019-04-16
LUCENE-8766: Add Luwak as a lucene module

romseygeek:


jpountz:
This has been merged.

PR #651, opened 2019-04-18
LUCENE-8774: Performance improvement for update?optimize=true

johann-beessip:
For our use case we have a small number of documents with a large number of docValues.  During optimize we noticed a lot of time being spent in the FilterFieldInfos constructor, on the filterFields.contains(...) function.  

The small change from an ArrayList to a LinkedHashSet reduced our optimize time from 60min to < 1min.

This was added to branch 7_2 as that's the version we're still using, but the issue is still present in master:
https://github.com/apache/lucene-solr/blob/master/lucene/core/src/java/org/apache/lucene/codecs/perfield/PerFieldDocValuesFormat.java#L142


janhoy:
Closing since LUCENE-8774 was marked as duplicate

PR #652, opened 2019-04-23
LUCENE-8775: Tessellator: Improve the election of diagonals when splitting the polygon

iverase:
This PR introduces:

- Make sure that resulting polygons from splitting logic are valid CW polygons.

- Make sure an error is thrown if a polygon cannot be splitted.

- Adds several test to the tessellator.


lucene/sandbox/src/java/org/apache/lucene/geo/Tessellator.java
```diff
@@ -163,7 +163,19 @@ private static final Node eliminateHoles(final Polygon polygon, Node outerNode)
...
+    {
+      double diff = pNodeA.getX() - pNodeB.getX();
+      if (diff == 0) {
+        diff = pNodeA.getY() - pNodeB.getY();
+        if (diff == 0) {
+          //same hole node
+          double a = Math.min(pNodeA.previous.getY(), pNodeA.next.getY());
+          double b = Math.min(pNodeB.previous.getY(), pNodeB.next.getY());
+          diff = a - b;
+        }
```
jpountz: Why do we need to look at previous and next nodes? Is it for the case that a vertex is shared by multiple holes, or by a hole and the enclosing polygon?

lucene/sandbox/src/java/org/apache/lucene/geo/Tessellator.java
```diff
@@ -248,8 +261,21 @@ private static final Node fetchHoleBridge(final Node holeNode, final Node outerN
...
+  private static Node getBestConnector(Node connection, Node holeNode) {
+    Node candidate = connection;
+    Node next = connection.next;
+    do {
+      if (isVertexEquals(next, connection) && isLocallyInside(next, holeNode)) {
+        candidate = next;
+      }
+      next = next.next;
+    } while (next.next != connection);
+    return candidate;
```
jpountz: This can run in linear time with the number of vertices?

lucene/sandbox/src/java/org/apache/lucene/geo/Tessellator.java
iverase: This is the case when two holes share a vertex and that vertex is the left most one in each case. We need to order them so the one on the top is bigger than the one at the bottom. 

In oder to do that we use the neighbour points. Probably it is enough to just use the next point but this is to work even in the case where holes are in different orientation.

lucene/sandbox/src/java/org/apache/lucene/geo/Tessellator.java
iverase: Yes, it is 0(n). It does not seem to degrade the performance of my tests.

iverase:
The algorithm seems to have problems when a hole shares a vertex with the polygon. I added a change that reduces the likelihood of failing but there are still corner cases.

One way to address it is to merge those holes using the shared vertex before looking for the bridges. I have done so and it seems to work but performance is hurt so I need to find a better way of doing so.

Darune:
Any news on this ?

iverase:
I am using the hole bounding box to filter out candidate vertex to be share between a polygon and a hole. With this change all the shapes I am working with tessellate properly and performance looks ok.

iverase:
@nknize I think this PR is ready, finally changes are not very invasive and benefits are good as it seems to handle most of corner cases.

nknize:
Sorry for the delay on this @iverase 
This LGTM!

PR #653, opened 2019-04-24
SOLR-13425: Wrong color in horizontal definition list

janhoy:


ctargett:
Good catch, thanks.

This is caused by the customized HTML templates I introduced in SOLR-12746 (7.6). Where the 1st column of that style of horizontal list used to be in a `td` tag, now it's in a `th` tag.

An alternate change that would ensure that style of list always looks exactly the way it did before would be to modify the template at https://github.com/apache/lucene-solr/blob/master/solr/solr-ref-guide/src/_templates/_hdlist.html.slim#L9 to just change:

```th.hdlist1 class=('strong' if option? 'strong')```

to

```td.hdlist1 class=('strong' if option? 'strong')```

This would prevent us from possibly finding other things we have to fix in the CSS later on. What do you think?

janhoy:
Thanks. I updated the PR with your solution which is better. Tried to build the HTML and it looks good. Will merge soon.

janhoy:
I'll merge to branch_8x, but since refGuide for 8.0 is not yet released, can I merge to branch_8_0 as well?

ctargett:
Yes, I'd love it if you would merge to branch_8_0 - if you didn't, I was going to. Thanks for fixing this.

PR #657, opened 2019-04-27
LUCENE-8781: FST direct arc addressing

msokolov:
Adds direct-array arc addressing mode to FSTs, enabled via new Builder constructor arg. The new encoding is used when there is a large number of outgoing Arcs that are reasonably dense in the space of labels.

### DIFF
--- a/lucene/codecs/src/java/org/apache/lucene/codecs/blocktreeords/OrdsSegmentTermsEnum.java
+++ b/lucene/codecs/src/java/org/apache/lucene/codecs/blocktreeords/OrdsSegmentTermsEnum.java
@@ -1082,10 +1082,6 @@ public final class OrdsSegmentTermsEnum extends BaseTermsEnum {
       if (FST.targetHasArcs(arc)) {
         // System.out.println("  targetHasArcs");
         result.grow(1+upto);
-        if (arc.target < 0 || arc.target > Integer.MAX_VALUE) {
-          assert(arc.target >= 0);
-          assert(arc.target < Integer.MAX_VALUE);
-        }
         fr.index.readFirstRealTargetArc(arc.target, arc, fstReader);
 
         if (arc.bytesPerArc != 0) {
--- a/lucene/codecs/src/java/org/apache/lucene/codecs/blocktreeords/OrdsSegmentTermsEnum.java
+++ b/lucene/codecs/src/java/org/apache/lucene/codecs/blocktreeords/OrdsSegmentTermsEnum.java
@@ -1082,12 +1082,14 @@ public final class OrdsSegmentTermsEnum extends BaseTermsEnum {
       if (FST.targetHasArcs(arc)) {
         // System.out.println("  targetHasArcs");
         result.grow(1+upto);
-        
+        if (arc.target < 0 || arc.target > Integer.MAX_VALUE) {
+          assert(arc.target >= 0);
+          assert(arc.target < Integer.MAX_VALUE);
+        }
         fr.index.readFirstRealTargetArc(arc.target, arc, fstReader);
 
         if (arc.bytesPerArc != 0) {
           // System.out.println("  array arcs");
-
           int low = 0;
           int high = arc.numArcs-1;
           int mid = 0;
--- a/lucene/core/src/java/org/apache/lucene/util/fst/Builder.java
+++ b/lucene/core/src/java/org/apache/lucene/util/fst/Builder.java
@@ -69,6 +69,8 @@ public class Builder<T> {
 
   private final IntsRefBuilder lastInput = new IntsRefBuilder();
 
+  final boolean useDirectArcAddressing;
+
   // NOTE: cutting this over to ArrayList instead loses ~6%
   // in build performance on 9.8M Wikipedia terms; so we
   // left this as an array:
@@ -92,12 +94,12 @@ public class Builder<T> {
   BytesStore bytes;
 
   /**
-   * Instantiates an FST/FSA builder without any pruning. A shortcut
-   * to {@link #Builder(FST.INPUT_TYPE, int, int, boolean,
-   * boolean, int, Outputs, boolean, int)} with pruning options turned off.
+   * Instantiates an FST/FSA builder without any pruning. A shortcut to {@link
+   * #Builder(FST.INPUT_TYPE, int, int, boolean, boolean, int, Outputs, boolean, int, boolean)} with
+   * pruning options turned off.
    */
   public Builder(FST.INPUT_TYPE inputType, Outputs<T> outputs) {
-    this(inputType, 0, 0, true, true, Integer.MAX_VALUE, outputs, true, 15);
+    this(inputType, 0, 0, true, true, Integer.MAX_VALUE, outputs, true, 15, false);
   }
 
   /**
@@ -148,12 +150,13 @@ public class Builder<T> {
    */
   public Builder(FST.INPUT_TYPE inputType, int minSuffixCount1, int minSuffixCount2, boolean doShareSuffix,
                  boolean doShareNonSingletonNodes, int shareMaxTailLength, Outputs<T> outputs,
-                 boolean allowArrayArcs, int bytesPageBits) {
+                 boolean allowArrayArcs, int bytesPageBits, boolean useDirectArcAddressing) {
     this.minSuffixCount1 = minSuffixCount1;
     this.minSuffixCount2 = minSuffixCount2;
     this.doShareNonSingletonNodes = doShareNonSingletonNodes;
     this.shareMaxTailLength = shareMaxTailLength;
     this.allowArrayArcs = allowArrayArcs;
+    this.useDirectArcAddressing = useDirectArcAddressing;
     fst = new FST<>(inputType, outputs, bytesPageBits);
     bytes = fst.bytes;
     assert bytes != null;
--- a/lucene/core/src/java/org/apache/lucene/util/fst/Builder.java
+++ b/lucene/core/src/java/org/apache/lucene/util/fst/Builder.java
@@ -49,6 +49,10 @@ import org.apache.lucene.util.fst.FST.INPUT_TYPE; // javadoc
  */
 
 public class Builder<T> {
+
+  // The amount of Arc array oversizing used to enable direct addressing of Arcs by their labels
+  static final int DIRECT_ARC_LOAD_FACTOR = 4;
+
   private final NodeHash<T> dedupHash;
   final FST<T> fst;
   private final T NO_OUTPUT;
--- a/lucene/core/src/java/org/apache/lucene/util/fst/FST.java
+++ b/lucene/core/src/java/org/apache/lucene/util/fst/FST.java
@@ -83,13 +83,13 @@ public final class FST<T> implements Accountable {
 
   // We use this as a marker (because this one flag is
   // illegal by itself ...):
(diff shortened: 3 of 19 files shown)


lucene/codecs/src/java/org/apache/lucene/codecs/blocktreeords/OrdsSegmentTermsEnum.java
```diff
@@ -1082,12 +1082,14 @@ private InputOutput getByOutput(long targetOrd) throws IOException {
       if (FST.targetHasArcs(arc)) {
         // System.out.println("  targetHasArcs");
         result.grow(1+upto);
-        
+        if (arc.target < 0 || arc.target > Integer.MAX_VALUE) {
```
dweiss: What's the point of this block (and isn't it effectively dead code)?

lucene/core/src/java/org/apache/lucene/util/fst/Builder.java
```diff
@@ -69,6 +69,8 @@
 
   private final IntsRefBuilder lastInput = new IntsRefBuilder();
 
+  final boolean useDirectArcAddressing;
```
dweiss: private final (?)

lucene/core/src/java/org/apache/lucene/util/fst/FSTEnum.java
```diff
@@ -141,11 +141,65 @@ protected void doSeekCeil() throws IOException {
       //System.out.println("  cycle upto=" + upto + " arc.label=" + arc.label + " (" + (char) arc.label + ") vs targetLabel=" + targetLabel);
 
       if (arc.bytesPerArc != 0 && arc.label != -1) {
+        // Arcs are in an array
```
dweiss: I feel tempted to split those very large methods into smaller units (methods); it'd really make them clearer.

lucene/core/src/java/org/apache/lucene/util/fst/FSTEnum.java
```diff
@@ -281,10 +335,81 @@ protected void doSeekFloor() throws IOException {
...
+              fst.readNextRealArc(arc, in);
+              assert arc.isLast();
+              assert arc.label < targetLabel: "arc.label=" + arc.label + " vs targetLabel=" + targetLabel;
+              pushLast();
+              return;
+            }
+            arc.nextArc = arc.posArcsStart - arc.bytesPerArc * targetOffset;
+            fst.readNextRealArc(arc, in);
+            if (arc.label == targetLabel) {
+              // found -- copy pasta from below
```
dweiss: pasta is good, but paste is better.

lucene/core/src/test/org/apache/lucene/util/fst/Test2BFST.java
```diff
@@ -50,12 +50,12 @@ public void test() throws Exception {
 
     for(int iter=0;iter<1;iter++) {
       // Build FST w/ NoOutputs and stop when nodeCount > 2.2B
-      {
+        if (false) {
```
dweiss: should be reverted (introduces dead code block).

lucene/core/src/java/org/apache/lucene/util/fst/FST.java
```diff
@@ -632,36 +641,85 @@ long addNode(Builder<T> builder, Builder.UnCompiledNode<T> nodeIn) throws IOExce
     if (doFixedArray) {
       final int MAX_HEADER_SIZE = 11; // header(byte) + numArcs(vint) + numBytes(vint)
       assert maxBytesPerArc > 0;
-      // 2nd pass just "expands" all arcs to take up a fixed
-      // byte size
+      // 2nd pass just "expands" all arcs to take up a fixed byte size
+
+      // If more than 1/2 of the "slots" would be occupied, write an arc array that may have holes in it
```
dweiss: Is this "1/2" consistent with the condition below (4 * nodeIn.numArcs)? Also, I think it'd be better to have the proportion as at least a constant or (better?) a configurable setting on the builder (not necessarily a constructor attribute).

lucene/core/src/java/org/apache/lucene/util/fst/FST.java
```diff
@@ -632,36 +641,85 @@ long addNode(Builder<T> builder, Builder.UnCompiledNode<T> nodeIn) throws IOExce
...
       //System.out.println("write int @pos=" + (fixedArrayStart-4) + " numArcs=" + nodeIn.numArcs);
       // create the header
       // TODO: clean this up: or just rewind+reuse and deal with it
       byte header[] = new byte[MAX_HEADER_SIZE]; 
       ByteArrayDataOutput bad = new ByteArrayDataOutput(header);
       // write a "false" first arc:
-      bad.writeByte(ARCS_AS_FIXED_ARRAY);
-      bad.writeVInt(nodeIn.numArcs);
+      if (writeDirectly) {
+        bad.writeByte(ARCS_AS_DIRECT_ARRAY);
```
dweiss: Perhaps change constants to something reflecting reality here -- ARCS_AS_ARRAY_DENSE, ARCS_AS_ARRAY_WITH_GAPS?

lucene/core/src/java/org/apache/lucene/util/fst/FST.java
```diff
@@ -632,36 +641,85 @@ long addNode(Builder<T> builder, Builder.UnCompiledNode<T> nodeIn) throws IOExce
...
-        builder.bytes.skipBytes((int) (destPos - srcPos));
-        for(int arcIdx=nodeIn.numArcs-1;arcIdx>=0;arcIdx--) {
-          destPos -= maxBytesPerArc;
-          srcPos -= builder.reusedBytesPerArc[arcIdx];
-          //System.out.println("  repack arcIdx=" + arcIdx + " srcPos=" + srcPos + " destPos=" + destPos);
-          if (srcPos != destPos) {
-            //System.out.println("  copy len=" + builder.reusedBytesPerArc[arcIdx]);
-            assert destPos > srcPos: "destPos=" + destPos + " srcPos=" + srcPos + " arcIdx=" + arcIdx + " maxBytesPerArc=" + maxBytesPerArc + " reusedBytesPerArc[arcIdx]=" + builder.reusedBytesPerArc[arcIdx] + " nodeIn.numArcs=" + nodeIn.numArcs;
-            builder.bytes.copyBytes(srcPos, destPos, builder.reusedBytesPerArc[arcIdx]);
+      if (writeDirectly) {
```
dweiss: Huge methods... move to a submethod?

lucene/codecs/src/java/org/apache/lucene/codecs/blocktreeords/OrdsSegmentTermsEnum.java
msokolov: Thanks for the comments, and sorry for the messes you found! It looks as if I fumbled uploading the patch and pushed a slightly outdated version that was missing some final cleanups, removing debug code and the like. I think failed to keep changes in sync between this and the working version from our production system that still builds against 7.5. Anyway I'll address that soon.

lucene/core/src/java/org/apache/lucene/util/fst/Builder.java
msokolov: FST reads this so it can't be private. This is following the same pattern as some existing settings like `allowArrayArcs`, which by the way should probably be final but isn't.

lucene/core/src/java/org/apache/lucene/util/fst/FSTEnum.java
msokolov: I was thinking of introducing multiple Arc classes so that the (now three) different kinds of Arc decoding methods can be modeled explicitly with function overrides and the branching between different implementations done using virtual function dispatch rather than inline conditionals. That would likely subsume any refactoring here, so I'd like to postpone it, but I definitely agree with the temptation. Still, if you think it would make the review easier, I'd be happy to dive into that, but overall it will probably lead to more code changes making direct comparison difficult, and I was thinking to do a separate a refactoring commit with no behavior changes.

lucene/core/src/java/org/apache/lucene/util/fst/FST.java
msokolov: Right it should have said 1/4, and I agree a symbol of some kind is called for. I think I'll add a constant, because rIght now the Builder has no (other) settable members; in practice every setting is passed on the constructor: it's not actually a Builder in the builder-pattern sense, and adding a constructor arg will be kind of intrusive given there is not yet any call for it elsewhere.

lucene/core/src/java/org/apache/lucene/util/fst/FST.java
msokolov: OK

lucene/core/src/java/org/apache/lucene/util/fst/FSTEnum.java
msokolov: After all, this revision includes break-up into smaller functions; it's definitely easier to read, and I guess could feed into some restructuring into classes later.

dweiss: Yes, a full refactoring is probably worth another issue, I'm just saying those methods evolved into complex beasts.

lucene/core/src/java/org/apache/lucene/util/fst/FST.java
```diff
@@ -865,20 +939,32 @@ public int readNextArcLabel(Arc<T> arc, BytesReader in) throws IOException {
...
+        // arcs are in an array
+        if (arc.arcIdx >= 0) {
+          in.setPosition(arc.posArcsStart);
+          // point at next arc, -1 to skip flags
+          in.skipBytes((1 + arc.arcIdx) * arc.bytesPerArc + 1);
+        } else {
+          in.setPosition(arc.nextArc);
+          byte flags = in.readByte();
+          // skip missing arcs
+          while (flag(flags, BIT_MISSING_ARC)) {
```
mikemccand: It would be nice (separate issue) if we could track counters of how many "arc scans" (typically done by `seekCeil` from Lucene's terms dict) versus "direct lookups" are done, in a typical use case ... but that's a future issue.

msokolov: great idea - perhaps if we add some Arc abstractions that would provide a nice hook for such tracking/counting

lucene/codecs/src/java/org/apache/lucene/codecs/blocktreeords/OrdsSegmentTermsEnum.java
msokolov: @dweiss I think I addressed your comments and this is ready to be committed. If you agree, would you be able to help with that (I'm not a committer, so I need help!)?

dweiss:
Thanks Mike. I'll take a look and will get back to you.

dweiss:
It all looks good, Mike. There is one remaining thing -- to add a changes.txt entry in Lucene (under "improvements" section, I think). Add yourself as author of the patch, please.

msokolov:
Thanks Dawid; while adding the CHANGES entry, I realized that the patch was not in fact enabling the new encoding for suggesters! Somehow I had flipped the bit while testing/submitting patches. So I re-enabled that as well, and it is still enabled for postings.

dweiss:
Committed to master, thanks Mike!

PR #660, opened 2019-04-29
SOLR-13047: Add facet2D Streaming Expression

NazerkeBS:
@joel-bernstein, there are since only 2 levels, we don't call appendJson(...) function recursively. Also, in the fillTuples(..) I looped first the "x" bucket then "y" bucket and added to the tuple list which has a type of Tuple, but not sure.


NazerkeBS:
I have not tested yet because I was not sure whether the implementation is correct or not. I will write a test case for that and try locally.

joel-bernstein:
Ok, sounds good. When you post the test case it will be easy for me to pull down and test out.

PR #664, opened 2019-05-07
LUCENE-8785: Ensure threadstates are locked before iterating

s1monw:
Ensure new threadstates are locked before retrieving the
number of active threadstates. This causes assertion errors
and potentially broken field attributes in the IndexWriter when
IndexWriter#deleteAll is called while actively indexing.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/DocumentsWriter.java
+++ b/lucene/core/src/java/org/apache/lucene/index/DocumentsWriter.java
@@ -287,8 +287,8 @@ final class DocumentsWriter implements Closeable, Accountable {
     };
     try {
       deleteQueue.clear();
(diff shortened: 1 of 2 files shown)


dweiss:
Looks good to me, thanks Simon.

lucene/core/src/java/org/apache/lucene/index/DocumentsWriter.java
```diff
@@ -287,8 +287,8 @@ synchronized Closeable lockAndAbortAll() throws IOException {
     };
     try {
       deleteQueue.clear();
-      final int limit = perThreadPool.getMaxThreadStates();
       perThreadPool.lockNewThreadStates();
+      final int limit = perThreadPool.getMaxThreadStates();
```
mikemccand: Wow, subtle!

PR #667, opened 2019-05-08
LUCENE-8796: Use exponential search in IntArrayDocIdSetIterator#advance

javanna:
As described in https://issues.apache.org/jira/browse/LUCENE-8796 .

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/util/IntArrayDocIdSet.java
+++ b/lucene/core/src/java/org/apache/lucene/util/IntArrayDocIdSet.java
@@ -77,7 +77,7 @@ final class IntArrayDocIdSet extends DocIdSet {
       while(i + bound < length && docs[i + bound] < target) {
         bound *= 2;
       }
-      i = Arrays.binarySearch(docs, i + bound / 2, Math.min(i + bound, length), target);
+      i = Arrays.binarySearch(docs, i + bound / 2, Math.min(i + bound + 1, length), target);
       if (i < 0) {
         i = -1 - i;
       }
--- a/lucene/core/src/java/org/apache/lucene/util/IntArrayDocIdSet.java
+++ b/lucene/core/src/java/org/apache/lucene/util/IntArrayDocIdSet.java
@@ -74,6 +74,7 @@ final class IntArrayDocIdSet extends DocIdSet {
     public int advance(int target) throws IOException {
       int bound = 1;
       int offset = Math.max(0, i);
+      //given that we use this for small array only, this is very unlikely to overflow
       while(offset + bound < length && docs[offset + bound] < target) {
         bound *= 2;
       }
--- a/lucene/core/src/java/org/apache/lucene/util/IntArrayDocIdSet.java
+++ b/lucene/core/src/java/org/apache/lucene/util/IntArrayDocIdSet.java
@@ -52,7 +52,7 @@ final class IntArrayDocIdSet extends DocIdSet {
 
     private final int[] docs;
     private final int length;
-    private int i = -1;
+    private int i = 0;
     private int doc = -1;
 
     IntArrayDocIdSetIterator(int[] docs, int length) {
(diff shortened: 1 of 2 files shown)


lucene/core/src/java/org/apache/lucene/util/IntArrayDocIdSet.java
```diff
@@ -72,7 +72,12 @@ public int nextDoc() throws IOException {
 
     @Override
     public int advance(int target) throws IOException {
-      i = Arrays.binarySearch(docs, i + 1, length, target);
+      int bound = 1;
+      int offset = Math.max(0, i);
+      while(offset + bound < length && docs[offset + bound] < target) {
+        bound *= 2;
+      }
+      i = Arrays.binarySearch(docs, offset + bound / 2, Math.min(offset + bound, length), target);
```
jpountz: `bound/2` is generally the previous bound that we tested, except when `bound` is equal to 1. It won't break in that case since callers are not supposed to call advance on a target that is lte the current doc ID, but this might still make room for bugs?

lucene/core/src/java/org/apache/lucene/util/IntArrayDocIdSet.java
```diff
@@ -72,7 +72,12 @@ public int nextDoc() throws IOException {
 
     @Override
     public int advance(int target) throws IOException {
-      i = Arrays.binarySearch(docs, i + 1, length, target);
+      int bound = 1;
+      int offset = Math.max(0, i);
```
jpountz: it feels a bit wrong to me that i==-1 needs a special case.

lucene/core/src/java/org/apache/lucene/util/IntArrayDocIdSet.java
```diff
@@ -72,7 +72,12 @@ public int nextDoc() throws IOException {
...
+      while(offset + bound < length && docs[offset + bound] < target) {
```
jpountz: I think it's unlikely to happen in practice since we use it for small arrays, but should we protect against integer overflow? Or at least add a comment that explains why this can't overflow?

javanna:
Thanks @jpountz this should be ready, I posted [here](https://issues.apache.org/jira/browse/LUCENE-8796?focusedCommentId=16852876&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-16852876) the latest benchmark results. Let me know if there's anything else that I can do.

lucene/core/src/java/org/apache/lucene/util/IntArrayDocIdSet.java
```diff
@@ -67,16 +67,21 @@ public int docID() {
...
 
     @Override
     public int advance(int target) throws IOException {
-      i = Arrays.binarySearch(docs, i + 1, length, target);
+      int bound = 1;
+      //given that we use this for small arrays only, this is very unlikely to overflow
+      while(i + bound < length && docs[i + bound] < target) {
+        bound *= 2;
+      }
+      i = Arrays.binarySearch(docs, i + bound / 2, Math.min(i + bound, length), target);
```
jpountz: I think it should be `binarySearch(docs, i + bound / 2, Math.min(i + bound + 1, length), target);`? (maybe we need a test for this special case that the result is found at `i + bound` exactly?)

javanna: correct, the upper bound is exclusive in Arrays.binarySearch . Funnily enough, this does not seem to cause any problem, as the returned value is `- insertion point - 1` with which we do `i = -1 - i`. If the docId was not in the array, we'd return the docId at the hypothetical position where the docId would be if it was in the array. In reality it is there, and what we return is the proper value still :) I think the test coverage is already good enough, I have corrected this, but I am not sure how to test it given that it did not end up causing any bug.

javanna:
done thanks @jpountz

PR #674, opened 2019-05-13
SOLR-13499: Fix "Apache License, Version 2.0" spelling in pom.xml.template

don-vip:
Fix for [SOLR-13499](https://issues.apache.org/jira/browse/SOLR-13499).
There are many Java libraries licensed under "Apache License, Version 2.0" that do not use its official spelling.
This causes issues like https://issues.apache.org/jira/browse/MPIR-382: with every library defining its own spelling, it's difficult in large projects to have a clear view of all licenses in use.
This PR changes the license spelling to the official one, as advised by Maven developers.

### DIFF
--- a/dev-tools/maven/pom.xml.template
+++ b/dev-tools/maven/pom.xml.template
@@ -114,7 +114,7 @@
   </scm>
   <licenses>
     <license>
-      <name>Apache 2</name>
+      <name>Apache License, Version 2.0</name>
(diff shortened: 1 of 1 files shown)


gus-asf:
You will probably want to create an issue in the Apache Jira (https://issues.apache.org/jira/issues/?jql=project%20%3D%20SOLR) for this and then re-issue this PR with the JIRA number and title so it can be tracked and (if needed) discussed.

don-vip:
@gus-asf done

PR #675, opened 2019-05-14
SOLR-13350: Multi-threaded search through an index

chatman:
Creating a github pull request for easier review.


lucene/core/src/java/org/apache/lucene/search/MultiCollector.java
```diff
@@ -115,6 +116,11 @@ public ScoreMode scoreMode() {
     return scoreMode;
   }
 
+  // nocommit: need to raise a LUCENE jira for this?
+  public List<Collector> getCollectors() {
+    return Collections.unmodifiableList(Arrays.asList(collectors));
+  }
```
jpountz: Maybe we don't need this if you maintain sub collectors as a `List<Collector>` and only convert to a `MultiCollector` when you need to pass them to `IndexSearcher`?

chatman: It seems a bit tricky. I'm creating these wrapped MultiCollectors in the newCollector() method of a CollectorManager; in the reduce() method, I get these MultiCollectors back but I'm not sure how I can associate them to some storage (for List<Collector>) for each MultiCollector that I created in newCollector. Maybe, I can extend MultiCollector to associate additional data with each at the time of creation, but seemed ugly.

This is used here: https://github.com/apache/lucene-solr/blob/e3b082a7f80f1682f650b531627b1af90eebc375/solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java#L1700-L1716

chatman:
Newer PR is here: https://github.com/apache/lucene-solr/pull/1310

PR #676, opened 2019-05-14
LUCENE-8753: SharedTerms UniformSplit

bruno-roustant:
An extension of UniformSplit to share the terms between fields. The fields share the same FST dictionary in the block file. Each term occur only once in the block file, and the different term states for multiple fields are stored on each term line in the block file. So there is a single FST for multiple fields, which is generally much more compact when there a lot of fields.


bruno-roustant:
Moved to #828 so I am closing this.

PR #677, opened 2019-05-14
SOLR-13257: support for stable replica routing preferences

magibney:
`shards.preferences` parameter allows users to assert replica preferences based on inherent replica attributes. This PR integrates support for user-specified ordering behavior applied to replicas that are equivalent with respect to preferred inherent attributes. 

The extant implementation shuffles replicas randomly per-request. Propose to leave shuffling as the default, but add support for stable replica routing preferences (list rotation by dividend|hash % number equivalent replicas)

### DIFF
--- a/solr/core/src/java/org/apache/solr/handler/component/AffinityReplicaListTransformer.java
+++ b/solr/core/src/java/org/apache/solr/handler/component/AffinityReplicaListTransformer.java
@@ -1,0 +1,122 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.solr.handler.component;
+
+import java.lang.invoke.MethodHandles;
+import java.util.Arrays;
+import java.util.Comparator;
+import java.util.List;
+import java.util.ListIterator;
+import org.apache.solr.common.cloud.Replica;
+import org.apache.solr.common.params.SolrParams;
+import org.apache.solr.common.util.Hash;
+import org.apache.solr.request.SolrQueryRequest;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Allows better caching by establishing deterministic evenly-distributed replica routing preferences according to
+ * either explicitly configured hash routing parameter, or the hash of a query parameter (configurable, usually related
+ * to the main query).
+ */
+public class AffinityReplicaListTransformer implements ReplicaListTransformer {
+
+  private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+
+  private final int routingDividend;
+
+  public AffinityReplicaListTransformer(String hashVal) {
+    this.routingDividend = Math.abs(Hash.lookup3ycs(hashVal, 0, hashVal.length(), 0));
+  }
+
+  public AffinityReplicaListTransformer(int routingDividend) {
+    this.routingDividend = routingDividend;
+  }
+
+  /**
+   *
+   * @param dividendParam int param to be used directly for mod-based routing
+   * @param hashParam String param to be hashed into an int for mod-based routing
+   * @param req the request from which param values will be drawn
+   * @return null if specified routing vals are not able to be parsed properly
+   */
+  public static ReplicaListTransformer getInstance(String dividendParam, String hashParam, SolrQueryRequest req) {
+    SolrParams params = req.getOriginalParams();
+    String dividendVal;
+    if (dividendParam != null && (dividendVal = params.get(dividendParam)) != null && !dividendVal.isEmpty()) {
+      try {
+        return new AffinityReplicaListTransformer(Integer.parseInt(dividendVal));
+      } catch (NumberFormatException ex) {
+        log.warn("unable to parse int from " + dividendParam + "=" + dividendVal);
+      }
+    }
+    String hashVal;
+    if (hashParam != null && (hashVal = params.get(hashParam)) != null && !hashVal.isEmpty()) {
+      return new AffinityReplicaListTransformer(hashVal);
+    } else {
+      return null;
+    }
+  }
+
+  @Override
+  @SuppressWarnings({"unchecked", "rawtypes"})
+  public void transform(List<?> choices) {
+    int size = choices.size();
+    if (size > 1) {
+      int i = 0;
+      SortableChoice[] sortableChoices = new SortableChoice[size];
+      for (Object o : choices) {
+        sortableChoices[i++] = new SortableChoice(o);
+      }
+      Arrays.sort(sortableChoices, SORTABLE_CHOICE_COMPARATOR);
+      ListIterator iter = choices.listIterator();
+      i = routingDividend % size;
+      final int limit = i + size;
+      do {
+        iter.next();
+        iter.set(sortableChoices[i % size].choice);
+      } while (++i < limit);
+    }
+  }
+
+  private static final class SortableChoice {
+
+    private final Object choice;
+    private final String sortableCoreLabel;
+
+    private SortableChoice(Object choice) {
+      this.choice = choice;
+      if (choice instanceof Replica) {
+        this.sortableCoreLabel = ((Replica)choice).getCoreUrl();
+      } else if (choice instanceof String) {
+        this.sortableCoreLabel = (String)choice;
+      } else {
+        throw new IllegalArgumentException("can't handle type " + choice.getClass());
+      }
+    }
+
+  }
+
+  private static final Comparator<SortableChoice> SORTABLE_CHOICE_COMPARATOR = new Comparator<SortableChoice>() {
+
+    @Override
+    public int compare(SortableChoice o1, SortableChoice o2) {
+      return o1.sortableCoreLabel.compareTo(o2.sortableCoreLabel);
+    }
+  };
+
+}
--- a/solr/core/src/java/org/apache/solr/handler/component/AffinityReplicaListTransformer.java
+++ b/solr/core/src/java/org/apache/solr/handler/component/AffinityReplicaListTransformer.java
@@ -56,13 +56,9 @@ class AffinityReplicaListTransformer implements ReplicaListTransformer {
    */
   public static ReplicaListTransformer getInstance(String dividendParam, String hashParam, SolrQueryRequest req) {
     SolrParams params = req.getOriginalParams();
-    String dividendVal;
-    if (dividendParam != null && (dividendVal = params.get(dividendParam)) != null && !dividendVal.isEmpty()) {
-      try {
-        return new AffinityReplicaListTransformer(Integer.parseInt(dividendVal));
-      } catch (NumberFormatException ex) {
-        log.warn("unable to parse int from " + dividendParam + "=" + dividendVal);
-      }
+    Integer dividendVal;
+    if (dividendParam != null && (dividendVal = params.getInt(dividendParam)) != null) {
+      return new AffinityReplicaListTransformer(dividendVal);
     }
     String hashVal;
     if (hashParam != null && (hashVal = params.get(hashParam)) != null && !hashVal.isEmpty()) {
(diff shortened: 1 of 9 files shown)


janhoy:
Linking with SOLR-13257

solr/core/src/java/org/apache/solr/handler/component/HttpShardHandlerFactory.java
```diff
@@ -449,9 +556,83 @@ private static boolean hasReplicaType(Object o, String preferred) {
...
+    private final NodePreferenceRulesComparator replicaComp;
+    private final ReplicaListTransformer baseReplicaListTransformer;
+
+    public TopLevelReplicaListTransformer(NodePreferenceRulesComparator replicaComp, ReplicaListTransformer baseReplicaListTransformer) {
+      this.replicaComp = replicaComp;
+      this.baseReplicaListTransformer = baseReplicaListTransformer;
+    }
+
+    @Override
+    public void transform(List<?> choices) {
```
tflobbe: I may be very wrong here, but I think the output of this method would be the same (or at least the same in practice) if we first run the transformer on all the list (which essentially either shuffles it or uses the new affinity code to define the order) and then run the `choices.sort(replicaComp)`, since this last one is stable, and will maintain the order of equal elements.
If that's the case, this method could be much simpler, and we don't need to find the boundaries, etc, plus the transformer gets called only once instead of once per group.

solr/core/src/java/org/apache/solr/handler/component/HttpShardHandlerFactory.java
```diff
@@ -449,9 +556,83 @@ private static boolean hasReplicaType(Object o, String preferred) {
...
+              Arrays.toString(replicaComp.preferenceRules.toArray()));
+        }
+
+        // First, sort according to comparator rules.
+        try {
+          choices.sort(replicaComp);
+        } catch (IllegalArgumentException iae) {
+          throw new SolrException(
+              SolrException.ErrorCode.BAD_REQUEST,
+              iae.getMessage()
```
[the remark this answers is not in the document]
magibney: This sounds reasonable to me, but this specific section of code is actually directly carried over from the existing code. I'm only mentioning this because I'm inclined to proceed with more caution here, in case this was a considered choice (for instance, if iae.getMessage() is expected to provide all the necessary information?). I notice that the IAEs that actually get thrown in NodePreferenceRulesComparator.compare() have pretty specific messages, so perhaps a the full exception/stack trace was deemed to be unnecessary?

solr/core/src/java/org/apache/solr/handler/component/HttpShardHandlerFactory.java
```diff
@@ -197,6 +200,60 @@ private static boolean getDisableShardsWhitelist() {
...
+      } else {
+        throw new IllegalArgumentException("more than one routing scheme marked as default");
+      }
+    }
+  }
+
+  private void initReplicaListTransformers(NamedList args) {
+    String defaultRouting = null;
+    if (args != null) {
+      NamedList routingConfig = (NamedList)args.get("replicaRouting");
```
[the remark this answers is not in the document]
magibney: Agreed, will push this change shortly. The error message is actually the main reason for the existence of the getNamedList() method, and the message definitely applies when used in this place.

solr/core/src/java/org/apache/solr/handler/component/HttpShardHandlerFactory.java
```diff
@@ -197,6 +200,60 @@ private static boolean getDisableShardsWhitelist() {
     return Boolean.getBoolean(INIT_SOLR_DISABLE_SHARDS_WHITELIST);
   }
 
+  private static NamedList<?> getNamedList(Object val) {
+    if (val instanceof NamedList) {
+      return (NamedList<?>)val;
+    } else {
+      throw new IllegalArgumentException("Invalid config for replicaRouting: " + val);
```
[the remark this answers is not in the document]
magibney: Agreed, will push update shortly.

solr/core/src/java/org/apache/solr/handler/component/HttpShardHandlerFactory.java
magibney: I had to think about this for a minute, when I wrote it, and again now :-) ...

I think it actually _would_ be different. The key is that the affinity code is always based on the same absolute order; it's only the _rotation_ that changes. And if we apply the rotation to the overall list before applying other preferences, the stability of the sort doesn't help us.

Take for example a situation with replicas [A, B, C, D, E]. For the sake of argument, let's say that B and C will end up being tied for "most preferred", according to the inherent preference comparators. The "affinity" transformation, applied at the top level, would sort the replicas: [A, B, C, D, E], then rotate them according to some explicitly specified or hash-derived int param. The problem is that there's only one in 5 possible top-level rotations would result in C being ranked first: [C, D, E, A, B]. So the resulting order would be deterministic, but wouldn't distribute load evenly.

Applying the affinity rotation "in post" does end up distributing load evenly, since you're always rotating mod the number of _equivalent_ replicas per group.

solr/core/src/java/org/apache/solr/handler/component/HttpShardHandlerFactory.java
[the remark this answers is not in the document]
tflobbe: Ah, maybe. Makes sense, since you are just catching IAEs

solr/core/src/java/org/apache/solr/handler/component/HttpShardHandlerFactory.java
tflobbe: Ah, right, because the affinity just sets the rotation. But does it have to be? What's the difference between doing a rotations vs a shuffle with a seed? like:
`Collections.shuffle(replicas, new Random(hashValue));`

magibney: True, random-with-seed should work, and I'm open to that. One use case that would benefit from a list-rotation-based implementation (as opposed to random-with-seed) would be if you wanted to set up a tiered or grouped routing system.

For example, replication factor of 2, say you have two distinct types of users with different access patterns (that would result in different cache usage patterns). Rather than simply routing deterministically for a given user/request, you could choose one of exactly 2 routing params, bifurcating traffic depending on expected use. With list-rotation, you could specify `routingParam=0` or `routingParam=1` and get the desired behavior; random-with-seed might not work that way, depending on the exact seeds chosen.

Granted you could achieve this tiered/grouped routing in other ways, but with a list-rotation-based implementation it would be trivial. The cost, as you say, would be some additional complexity (code-wise, not performance-wise) in `HttpShardHandlerFactory`.

In fact, it's this type of use case that I had in mind when providing support for the ability to specify `dividend` directly (as opposed to always hashing the routing param, which would be similarly opaque and thus incompatible with a tiered/grouped routing strategy).

tflobbe: You mean, you want to make sure those two users get different affinity, right? vs with shuffle-with-seed, they could end up in the same nodes?
In any way, I think this PR looks good, we can improve things too in the future if we want.

magibney: >make sure those two users get different affinity, right? vs with shuffle-with-seed, they could end up in the same nodes?

Exactly, yes; that's the idea. Thanks!

solr/core/src/java/org/apache/solr/handler/component/AffinityReplicaListTransformer.java
```diff
@@ -0,0 +1,122 @@
...
+   * @param hashParam String param to be hashed into an int for mod-based routing
+   * @param req the request from which param values will be drawn
+   * @return null if specified routing vals are not able to be parsed properly
+   */
+  public static ReplicaListTransformer getInstance(String dividendParam, String hashParam, SolrQueryRequest req) {
+    SolrParams params = req.getOriginalParams();
+    String dividendVal;
+    if (dividendParam != null && (dividendVal = params.get(dividendParam)) != null && !dividendVal.isEmpty()) {
+      try {
+        return new AffinityReplicaListTransformer(Integer.parseInt(dividendVal));
```
cpoerschke: Might the `Integer SolrParams.getInt(String)` method be an alternative way of parsing the `dividendParam`?

solr/core/src/java/org/apache/solr/handler/component/AffinityReplicaListTransformerFactory.java
```diff
@@ -0,0 +1,66 @@
...
+ */
+package org.apache.solr.handler.component;
+
+import org.apache.solr.common.params.CommonParams;
+import org.apache.solr.common.params.ShardParams;
+import org.apache.solr.common.util.NamedList;
+import org.apache.solr.request.SolrQueryRequest;
+
+/**
+ *
```
cpoerschke: `AffinityReplicaListTransformer` being public but `AffinityReplicaListTransformerFactory` here not being public jumps out as surprising, I would have expected them both to have the same visibility or potentially for the factory to have more visibility that the class itself. And perhaps the javadocs could mention e.g. about the defaulting-to-Q-param behaviour or so. What do you think?

cpoerschke:
> ... Also, we should probably add some documentation on how to configure the replicaRouting in solr.xml

Might the https://github.com/apache/lucene-solr/blob/master/solr/solr-ref-guide/src/format-of-solr-xml.adoc#the-shardhandlerfactory-element section be a suitable place?

solr/core/src/java/org/apache/solr/handler/component/AffinityReplicaListTransformerFactory.java
magibney: Yes, that makes sense. I'll make `AffinityReplicaListTransformer` package-private (like `AffinityReplicaListTransformerFactory`, and `ShufflingReplicaListTransformer`). I'll flesh out the javadocs too.

magibney:
Indeed, I'll update the docs under the ShardHandlerFactory section. Thanks!

solr/core/src/java/org/apache/solr/handler/component/AffinityReplicaListTransformer.java
magibney: That would certainly be more succinct. I think the intention here was to be relaxed about parsing (i.e., if you get a garbage dividend param, fall back to hashing the hashParam -- the assumption being that we're doing our best to honor the user's request for _stability_, one way or another). Given that approach, recovering from SolrParams.getInt(String) would have required catching a SolrException, which felt weird to me for some reason.

Assuming a change to SolrParams.getInt(String), do you think it makes sense to just be strict, or to be forgiving and catch/log the SolrException that could result from a non-integer dividendParam?

(I'm generally in favor of strictness, but in this case it feels a little arbitrary, since param in question doesn't affect the ability to return accurate results ...)

cpoerschke: Yes, that's an interesting question, how to handle problematic input, in a forgiving or in a strict way.

I'd tend to favour strictness in this case probably since the parameters would be within the Solr client's control (i.e. they are unlikely to be direct user input) or they would be part of the configuration via default or invariant elements in the config. So the queries would then relatively obviously fail and the client code can be fixed. If the queries succeed (with warnings logged) then it might be less apparent to Solr's client that something is amiss i.e. that routing and performance and cache hits etc. did not happen as intended, despite it returning, as you say, accurate results.

A semi-related edge case might also be affinity ambiguity i.e. both a dividend param and a hash param being supplied, though that ambiguity could be eliminated via documenation too.

magibney: That makes sense; I agree that strict is the way to go here, for the reasons you state. I'll also address the priority of dividend vs. hash param in the forthcoming revised documentation.

cpoerschke:
Thanks @magibney for updating the pull request and splitting the updates into logical commit units, I found that really helpful when incrementally reviewing.

solr/core/src/java/org/apache/solr/handler/component/HttpShardHandlerFactory.java
```diff
@@ -230,7 +288,23 @@ public void init(PluginInfo info) {
...
+
+    this.defaultClient = new Http2SolrClient.Builder()
+        .connectionTimeout(connectionTimeout)
+        .idleTimeout(soTimeout)
+        .maxConnectionsPerHost(maxConnectionsPerHost).build();
+    this.defaultClient.addListenerFactory(this.httpListenerFactory);
+    this.loadbalancer = new LBHttp2SolrClient(defaultClient);
+    initReplicaListTransformers(getParameter(args, "replicaRouting", null, sb));
+
     log.debug("created with {}",sb);
```
cpoerschke: You clarify in the https://github.com/apache/lucene-solr/pull/677/commits/d82dc54fd625762008a5f8b7c69bcc4ee4f57203 commit message that the re-ordering here is because `sb` is being logged here but then it's also subsequently still used.

I wonder if an alternative change could be to 'relocate' the logging statement to the end of the `init` method? The logging of a 'created' message in the middle of the init seems surprising from a code comprehension perspective but also just after the `log.debug` there is the `r.setSeed` call and (theoretically at least) moving code from 'after' to 'before' that might make a difference.

magibney: Yes, this would definitely be better. There's no reason for the "created" logging statement to be in the middle of the method, and in fact its being there initially is what nearly caused me to bypass it with the additions in this patch! Just pushed a commit to address this, thanks!

PR #680, opened 2019-05-17
LUCENE-8803: Provide a FieldComparator to allow sorting by a feature from a FeatureField

colings86:
As described in https://issues.apache.org/jira/browse/LUCENE-8803

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/FeatureComparatorSource.java
+++ b/lucene/core/src/java/org/apache/lucene/search/FeatureComparatorSource.java
@@ -1,0 +1,131 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.lucene.search;
+
+import java.io.IOException;
+
+import org.apache.lucene.document.FeatureField;
+import org.apache.lucene.index.LeafReaderContext;
+import org.apache.lucene.index.PostingsEnum;
+import org.apache.lucene.index.Terms;
+import org.apache.lucene.index.TermsEnum;
+import org.apache.lucene.util.BytesRef;
+
+/** Provides a comparator that parses a feature field's values as float and sorts by descending value */
+public class FeatureComparatorSource extends FieldComparatorSource {
+
+  private final BytesRef featureName;
+  private final Float missingValue;
+
+  /**
+   * Creates a {@link FieldComparatorSource} that can be used to sort hits by
+   * the value of a particular feature in a {@link FeatureField}. Note that the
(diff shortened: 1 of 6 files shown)


lucene/core/src/java/org/apache/lucene/search/FeatureComparatorSource.java
```diff
@@ -0,0 +1,131 @@
...
+   * field name to use for the sort will be specified by the {@link SortField}.
+   *
+   * @param featureName The name of the feature to use for the sort value
+   * @param missingValue The value to use for documents that don't have a value
+   *   for the feature. If <code>null</code>, documents with no value for the
+   *   feature will be sorted as if their value is <code>0.0f</code>.
+   */
+  public FeatureComparatorSource(BytesRef featureName, Float missingValue) {
+    this.featureName = featureName;
+    this.missingValue = missingValue;
```
jpountz: FeatureField is mostly useful for queries, where we don't support default values. I don't think supporting default values at sort time is going to be useful.

lucene/core/src/java/org/apache/lucene/search/FeatureComparatorSource.java
```diff
@@ -0,0 +1,131 @@
...
+          currentReaderPostingsValues = null;
+        } else {
+          currentReaderPostingsValues = termsEnum.postings(currentReaderPostingsValues, PostingsEnum.FREQS);
+        }
+      }
+    }
+
+    private float getValueForDoc(int doc) throws IOException {
+      if (currentReaderPostingsValues != null
+          && (currentReaderPostingsValues.docID() == doc || currentReaderPostingsValues.advance(doc) == doc)) {
```
jpountz: this might call advance on a target that is less that the current doc (which is illegal) if getValueForDoc is called consecutively on two doc IDs that don't have any match in the postings in between them

lucene/core/src/java/org/apache/lucene/search/FeatureComparatorSource.java
```diff
@@ -0,0 +1,131 @@
...
+        return currentReaderPostingsValues.freq();
```
jpountz: Freqs don't store the raw feature value but an encoded version, you need to pass the frequency to `FeatureField#decodeFeatureValue`

lucene/core/src/java/org/apache/lucene/search/FeatureComparatorSource.java
```diff
@@ -0,0 +1,131 @@
...
+
+import org.apache.lucene.document.FeatureField;
+import org.apache.lucene.index.LeafReaderContext;
+import org.apache.lucene.index.PostingsEnum;
+import org.apache.lucene.index.Terms;
+import org.apache.lucene.index.TermsEnum;
+import org.apache.lucene.util.BytesRef;
+
+/** Provides a comparator that parses a feature field's values as float and sorts by descending value */
+public class FeatureComparatorSource extends FieldComparatorSource {
```
jpountz: Can you make it pkg-private and expose it via a factory method on the FeatureField class similarly to `LatLonDocValuesField#newDistanceSort`? Exposing it via a SortField rather than a ComparatorSource might also help with not having to support reverse order (only decreasing order makes sense for features I think?).

lucene/core/src/java/org/apache/lucene/search/FeatureComparatorSource.java
colings86: So we should always assume a value of `0.0f` if the value is missing then? I think that makes sense since the feature is stored as a term frequency so not having a value corresponds to having a term frequency of 0

jpountz: Yes, this was my reasoning indeed.

colings86:
@jpountz Thanks for reviewing. I pushed a commit to address your comments.

colings86:
@jpountz I added an entry to CHANGES.txt, thanks for the reviews

PR #665, opened 2019-05-07
Fixes SOLR-13539

thomaswoeckinger:
Introduce EmbeddedSolrTestBase for easy and simple integration tests, fixes are included for SOLR-13331 and SOLR-13347.

@gerlowskija Please review

PR #681, opened 2019-05-21
Fix SOLR-13347

thomaswoeckinger:
Test are included in https://github.com/apache/lucene-solr/pull/665

### DIFF
--- a/solr/core/src/java/org/apache/solr/update/TransactionLog.java
+++ b/solr/core/src/java/org/apache/solr/update/TransactionLog.java
@@ -102,6 +102,7 @@ public class TransactionLog implements Closeable {
   };
 
   public class LogCodec extends JavaBinCodec {
+
     public LogCodec(JavaBinCodec.ObjectResolver resolver) {
       super(resolver);
     }
@@ -128,7 +129,7 @@ public class TransactionLog implements Closeable {
     public CharSequence readExternString(DataInputInputStream fis) throws IOException {
       int idx = readSize(fis);
       if (idx != 0) {// idx != 0 is the index of the extern string
-      // no need to synchronize globalStringList - it's only updated before the first record is written to the log
+        // no need to synchronize globalStringList - it's only updated before the first record is written to the log
         return globalStringList.get(idx - 1);
       } else {// idx == 0 means it has a string value
         // this shouldn't happen with this codec subclass.
@@ -136,6 +137,25 @@ public class TransactionLog implements Closeable {
       }
     }
 
+    @Override
+    protected Object readObject(DataInputInputStream dis) throws IOException {
+      if (UUID == tagByte) {
+        return new java.util.UUID(dis.readLong(), dis.readLong());
+      }
+      return super.readObject(dis);
+    }
+
+    @Override
+    public boolean writePrimitive(Object val) throws IOException {
+      if (val instanceof java.util.UUID) {
+        java.util.UUID uuid = (java.util.UUID) val;
+        daos.writeByte(UUID);
+        daos.writeLong(uuid.getMostSignificantBits());
+        daos.writeLong(uuid.getLeastSignificantBits());
+        return true;
+      }
+      return super.writePrimitive(val);
+    }
   }
 
   TransactionLog(File tlogFile, Collection<String> globalStrings) {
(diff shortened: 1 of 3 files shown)


thomaswoeckinger:
@janhoy : plz review

janhoy:
Nice. I'm afraid we cannot merge without a test in this PR. Could of course let this PR depend on #665 and thus only add the one test needed..

thomaswoeckinger:
> Nice. I'm afraid we cannot merge without a test in this PR. Could of course let this PR depend on #665 and thus only add the one test needed..

Added unit test to prove functionality

solr/solrj/src/java/org/apache/solr/common/util/JavaBinCodec.java
```diff
@@ -100,6 +100,7 @@
           MAP_ENTRY_ITER = 17,
           ENUM_FIELD_VALUE = 18,
           MAP_ENTRY = 19,
+          UUID = 20,
```
noblepaul: We should not add new types unless absolutely required. I'm not sure `UUID` is requires a new type

thomaswoeckinger: As you can see in the provided stack trace, it is required that JavaBinCodec must handle UUID, because it is written to the tlog. If you have a suggestion how to marshal and unmarshal such a type without using an type identifier, plz tell me, i did not see such a possibility

noblepaul: why does the unmarshal need to know the type?

thomaswoeckinger: because it needs to the exact opposite operation of the marshalling, otherwise the object can not be constructed from the stream

noblepaul: If I need to index a `UUID` field , I send a `String` using JSON and it works just fine right? How does it know how to handle the `UUID` field? The point is, there will be so many different types of Objects we need to handle.  An example is Point/Shape fields. javabin may not be able to accommodate all of them. It makes the format more complicated

thomaswoeckinger: if you do this, you will index it as StringField not as UUIDField

noblepaul: It depends on the type of the field , right?

If I send a `String` to `myfield_uuid ` which is configured as a UUID type, Solr will internally convert it to UUID , right?

thomaswoeckinger: Now i see your point, you are talking about the usage of JavaBinCodec used by the request handler, this is different to the usage of transaction logs (where the JavaBinCodec is also used), which requires marshalling and unmarshalling of the UUID type

thomaswoeckinger: Starting point is TransactionLog.write(TransactionLog.java:371)

solr/core/src/java/org/apache/solr/update/TransactionLog.java
```diff
@@ -102,6 +102,9 @@ public Object resolve(Object o, JavaBinCodec codec) throws IOException {
   };
 
   public class LogCodec extends JavaBinCodec {
+
+    private static final byte UUID = 20;
```
noblepaul: There is a risk that somebody may accidently use the code `20` and break `LogCodec`. So, let's keep this static field in JavabinCodec and say "This is reserved to be used only in LogCodec"

thomaswoeckinger:
So ready to merge?

audun:
I think the style comments from the other PR applies here too, ie. don't include unrelated changed in the commit. That makes it so much easier to review and merge.

The commit message should also be more descriptive, as you can see from the rest of the commit messages in the project. 

If you look in the commit history and you see a commit message like "Fix SOLR-12941", it means you need to look up what 12941 is. Compare that with the actual commit message: "SOLR-12941: Fix IndexSizeTrigger to correctly work with "aboveBytes" and "splitMethod=link" parameters."

thomaswoeckinger:
Updated commit message to be more specific

thomaswoeckinger:
So ready to merge?

gerlowskija:
This looks good to me, but I'm reluctant to merge until Noble's had a chance to re-review and clear his -1.  He also has more context on this stuff so I'd prefer to give him the last word.  I'll ping him in JIRA, and see if he has time to take another look.  If not, I'll come back and finalize this soon.

PR #684, opened 2019-05-23
LUCENE-8809: Ensure release segment states

dnhatn:
A [failed test](https://github.com/elastic/elasticsearch/issues/30290) from Elasticsearch shows that refresh and rollback concurrently can leave segment states unreleased leads to leaking refCount of some SegmentReaders.

Relates https://github.com/elastic/elasticsearch/issues/30290

### DIFF
--- a/lucene/core/src/test/org/apache/lucene/index/TestIndexWriter.java
+++ b/lucene/core/src/test/org/apache/lucene/index/TestIndexWriter.java
@@ -39,6 +39,7 @@ import java.util.Map;
 import java.util.Random;
 import java.util.Set;
 import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.Semaphore;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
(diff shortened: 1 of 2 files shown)


dnhatn:
@s1monw Can you please have a look? Thank you!

lucene/core/src/test/org/apache/lucene/index/TestIndexWriter.java
```diff
@@ -3730,4 +3733,50 @@ public void testFlushWhileStartingNewThreads() throws IOException, InterruptedEx
...
+      }
+    });
+
+    try {
+      indexer.start();
+      refresher.start();
+      indexedDocs.acquire(1 + random().nextInt(100));
+      w.rollback();
+    } finally {
+      stopped.get();
```
martin-g: What is this doing here ?

dnhatn: It should be `stopped.set(true)`. Thanks for looking.

dnhatn:
@martin-g @s1monw Thanks for reviewing.

PR #687, opened 2019-05-28
LUCENE-8815: Adds a DoubleValues implementation for feature fields

colings86:
As described in https://issues.apache.org/jira/browse/LUCENE-8815


lucene/core/src/test/org/apache/lucene/document/TestFeatureDoubleValues.java
```diff
@@ -0,0 +1,186 @@
...
+
+import org.apache.lucene.index.IndexReader;
+import org.apache.lucene.index.LeafReaderContext;
+import org.apache.lucene.index.RandomIndexWriter;
+import org.apache.lucene.search.DoubleValues;
+import org.apache.lucene.store.Directory;
+import org.apache.lucene.util.LuceneTestCase;
+
+/*
+ * Test for sorting using a feature from a FeatureField.
```
jpountz: update javadocs?

colings86:
Thanks for reviewing @jpountz. I pushed a couple of commits to address your comments, could you look again?

lucene/core/src/java/org/apache/lucene/document/FeatureDoubleValuesSource.java
```diff
@@ -0,0 +1,132 @@
...
+   * @throws NullPointerException if {@code field} or {@code featureName} is null.
+   */
+  public FeatureDoubleValuesSource(String field, BytesRef featureName) {
+    this.field = Objects.requireNonNull(field);
+    this.featureName = Objects.requireNonNull(featureName);
+  }
+
+  @Override
+  public boolean isCacheable(LeafReaderContext ctx) {
+    return false;
```
jpountz: this one can return true

lucene/core/src/java/org/apache/lucene/document/FeatureDoubleValuesSource.java
```diff
@@ -0,0 +1,132 @@
...
+      return false;
+    }
+    FeatureDoubleValuesSource other = (FeatureDoubleValuesSource) obj;
+    return Objects.equals(field, other.field) &&
+        Objects.equals(featureName, other.featureName);
+  }
+
+  @Override
+  public String toString() {
+    return "FeatureDoubleValuesSource("+field+", "+featureName+")";
```
jpountz: The toString representation will not look nice due to how BytesRef#toString works. Maybe we should take the featureName as a String in the constructor to make sure it is a String and not random bytes, and either use the String here, or the BytesRef#utf8ToString representation of the BytesRef.

lucene/core/src/test/org/apache/lucene/document/TestFeatureDoubleValues.java
```diff
@@ -0,0 +1,249 @@
...
+import org.apache.lucene.index.RandomIndexWriter;
+import org.apache.lucene.search.DoubleValues;
+import org.apache.lucene.search.DoubleValuesSource;
+import org.apache.lucene.search.IndexSearcher;
+import org.apache.lucene.store.Directory;
+import org.apache.lucene.util.BytesRef;
+import org.apache.lucene.util.LuceneTestCase;
+
+/*
+ * Test for sorting using a feature from a FeatureDoubleValuesSource.
```
jpountz: This doesn't test sorting in particular, but FeatureDoubleValuesSource in general?

colings86:
@jpountz thanks for reviewing, I pushed a new commit

lucene/CHANGES.txt
```diff
@@ -41,6 +41,12 @@ Other
 
 New Features
 
+* LUCENE-8815: Provide a DoubleValues implementation for retrieving the value of features without
+  requiring a separate numeric field. The feature sort added previously has been updated to use this
+  new DoubleValues implementation. Note that as feature values are stored with only 8 bits of mantissa
```
jpountz: The SortField seems not to be using the new double values?

PR #688, opened 2019-05-28
LUCENE-8813: Ensure we never apply deletes from a closed DWPTDeleteQueue

s1monw:
Today we don't have a strong protection that we add and apply deletes / updates
on or from an already flushed delete queue. DWPTDeleteQueue instances are replaced
once we do a full flush in order to reopen an NRT reader or commit the IndexWriter.

In LUCENE-8813 we tripped an assert that used to protect us from such an situation
but it didn't take all cornercases from concurrent flushing into account. This change
adds a stronger protection and ensures that we neither apply a closed delete queue nor
add any updates or deletes to it.

This change also allows to speculativly freeze the global buffer that might return
null now if the queue has already been closed. This is now possible since we ensure that
we never see modifications to the queue after it's been closed and that happens right after
the last DWPT for the ongoing full flush is done flushing.


s1monw:
@dnhatn @mikemccand would you mind taking a look

lucene/core/src/java/org/apache/lucene/index/DocumentsWriterDeleteQueue.java
```diff
@@ -183,6 +187,7 @@ boolean anyChanges() {
   }
 
   void tryApplyGlobalSlice() {
+    ensureOpen();
```
dnhatn: I think we should call `ensureOpen` after acquiring `globalBufferLock`.

lucene/core/src/java/org/apache/lucene/index/DocumentsWriter.java
```diff
@@ -168,27 +168,28 @@ long updateDocValues(DocValuesUpdate... updates) throws IOException {
...
   
   /** If buffered deletes are using too much heap, resolve them and write disk and return true. */
-  private boolean applyAllDeletes(DocumentsWriterDeleteQueue deleteQueue) throws IOException {
+  private boolean applyAllDeletes() throws IOException {
+    final DocumentsWriterDeleteQueue deleteQueue = this.deleteQueue;
     if (flushControl.isFullFlush() == false // never apply deletes during full flush this breaks happens before relationship
+        && deleteQueue.isOpen() // if it's closed then it's already fully applied and we have a new delete queue
         && flushControl.getAndResetApplyAllDeletes()) {
-      if (deleteQueue != null) {
-        assert assertTicketQueueModification(deleteQueue);
```
danmuzi: Is this code not necessary anymore?

lucene/core/src/java/org/apache/lucene/index/DocumentsWriter.java
```diff
@@ -168,27 +168,28 @@ long updateDocValues(DocValuesUpdate... updates) throws IOException {
...
   }
   
   /** If buffered deletes are using too much heap, resolve them and write disk and return true. */
-  private boolean applyAllDeletes(DocumentsWriterDeleteQueue deleteQueue) throws IOException {
+  private boolean applyAllDeletes() throws IOException {
+    final DocumentsWriterDeleteQueue deleteQueue = this.deleteQueue;
     if (flushControl.isFullFlush() == false // never apply deletes during full flush this breaks happens before relationship
+        && deleteQueue.isOpen() // if it's closed then it's already fully applied and we have a new delete queue
         && flushControl.getAndResetApplyAllDeletes()) {
-      if (deleteQueue != null) {
```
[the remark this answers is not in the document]
s1monw: yeah it can't be null afaik

lucene/core/src/java/org/apache/lucene/index/DocumentsWriter.java
```diff
@@ -724,14 +727,15 @@ long flushAllThreads()
         if (infoStream.isEnabled("DW")) {
           infoStream.message("DW", Thread.currentThread().getName() + ": flush naked frozen global deletes");
         }
-        assertTicketQueueModification(flushingDeleteQueue);
+        assert assertTicketQueueModification(flushingDeleteQueue);
```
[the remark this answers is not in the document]
s1monw: yeah phew! luckily this one is not heavy

lucene/core/src/java/org/apache/lucene/index/DocumentsWriterDeleteQueue.java
```diff
@@ -200,44 +205,74 @@ void tryApplyGlobalSlice() {
...
+        assert anyChanges() == false : "we are closed but have changes";
         return null;
       }
     } finally {
       globalBufferLock.unlock();
     }
   }
 
+  private FrozenBufferedUpdates freezeGlobalBufferInternal(final Node<?> currentTail ) {
+    if (globalSlice.sliceTail != currentTail) {
```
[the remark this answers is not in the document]
s1monw: 👍

lucene/core/src/java/org/apache/lucene/index/DocumentsWriter.java
s1monw: we can't assert this here since we don't hold the lock that protects us from running into the same situation again. I therefore added the closing protection to the delete queue to replace this and improve the protection. makes sense?

danmuzi: Oh, I understand what you mean.
Thanks.

PR #689, opened 2019-05-28
SOLR-13489: Stop the leader from trying to rejoin the election on ses…

markrmiller:
…sion expiration and harden our zk reconnect code path.

### DIFF
--- a/solr/core/src/java/org/apache/solr/cloud/ZkController.java
+++ b/solr/core/src/java/org/apache/solr/cloud/ZkController.java
@@ -335,10 +335,11 @@ public class ZkController implements Closeable {
         new OnReconnect() {
 
           @Override
-          public void command() {
+          public void command() throws SessionExpiredException {
             log.info("ZooKeeper session re-connected ... refreshing core states after session expiration.");
             clearZkCollectionTerms();
             try {
+              // recreate our watchers first so that they exist even on any problems below
               zkStateReader.createClusterStateWatchersAndUpdate();
 
               // this is troublesome - we dont want to kill anything the old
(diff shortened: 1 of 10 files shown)


solr/core/src/java/org/apache/solr/cloud/ZkController.java
```diff
@@ -368,8 +368,15 @@ public void command() {
...
-
-              registerAllCoresAsDown(registerOnReconnect, false);
+              
+              try {
+                registerAllCoresAsDown(registerOnReconnect, false);
+              } catch (SessionExpiredException e) {
+                // zk has to reconnect
+                throw e;
+              } catch (Exception e) {
+                log.warn("Exception while trying to register all cores as DOWN", e);
```
anshumg: don't you want to re-throw the exception here?

markrmiller: I'll add a better comment to the code.

anshumg: I should've guessed :) I just thought that you missed handling more cases.

solr/core/src/java/org/apache/solr/cloud/ZkController.java
```diff
@@ -1903,21 +1922,20 @@ private ZkCoreNodeProps waitForLeaderToSeeDownState(
             " is already less than leader, so not waiting for leader to see down state.");
       } else {
 
-        log.info("Replica need to wait for leader to see down state.");
+        log.info("Replica should to wait for leader to see down state.");
```
anshumg: Replica `is` waiting ?

anshumg: Also, while you are at it, can you add more information e.g. specifying the replica ?

solr/solrj/src/java/org/apache/solr/common/cloud/ConnectionManager.java
```diff
@@ -155,27 +155,28 @@ public void process(WatchedEvent event) {
                 public void update(SolrZooKeeper keeper) {
                   try {
                     waitForConnected(Long.MAX_VALUE);
+                    
+                    log.info("Connection with ZooKeeper reestablished.");
```
anshumg: we already log this as debug in `waitForConnected`. Let's just switch that out to an info ?

PR #690, opened 2019-05-29
SOLR-13517: [ UX improvement ] Dashboard will now store query and filter parameters on page change a…

saurav534:
…s well

1) refreshing the search page wont loose query and filter parameters anymore
2) this will make the filter page url sharable and land directly with filter and query intact
3) changing between multiple tabs(on left side) wont loose the these params if you come back to the search page again
4) This is mostly helpful as dashboard users switch between multiple views while making queries, and this is irritating when the earlier parameters a lost
5) Though refreshing whole page on search page and navigating between pages wont loose the filters but navigating to other page and refreshing on that page would loose it.
6) Above can be handled if I pass the query params on every page. As of now I am storing it in rootScope and query params of search page only (to make it sharable and handle refresh)


ctargett:
This is interesting, I can't even remember how long I've been annoyed at the lack of stickiness and url portability in the search UI.

However, is there a Jira issue for this effort? PRs should be linked to a Jira issue: https://wiki.apache.org/solr/HowToContribute#Working_with_GitHub

saurav534:
@ctargett created a JIRA ticket for the same. Updated the PR accordingly.

PR #697, opened 2019-06-05
LUCENE-8833: Allow MMapDirectory subclasses to pre-load per IndexInput

s1monw:
It's useful to select if we preload a mmap on a per index input / file
basis. Subclasses should be able to extend MMapDirectory to make it
configurable.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/store/MMapDirectory.java
+++ b/lucene/core/src/java/org/apache/lucene/store/MMapDirectory.java
@@ -231,6 +231,10 @@ public class MMapDirectory extends FSDirectory {
   /** Creates an IndexInput for the file with the given name. */
   @Override
   public IndexInput openInput(String name, IOContext context) throws IOException {
+    return openInput(name, context, this.preload);
+  }
+
+  protected IndexInput openInput(String name, IOContext context, boolean preload) throws IOException {
     ensureOpen();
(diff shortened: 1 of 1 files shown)


lucene/core/src/java/org/apache/lucene/store/MMapDirectory.java
```diff
@@ -231,20 +231,24 @@ public final int getMaxChunkSize() {
...
       return ByteBufferIndexInput.newInstance(resourceDescription,
-          map(resourceDescription, c, 0, c.size()), 
+          map(resourceDescription, c, 0, c.size(), preload),
           c.size(), chunkSizePower, new ByteBufferGuard(resourceDescription, useUnmap ? CLEANER : null));
     }
   }
 
   /** Maps a file into a set of buffers */
-  final ByteBuffer[] map(String resourceDescription, FileChannel fc, long offset, long length) throws IOException {
+  final ByteBuffer[] map(String resourceDescription, FileChannel fc, long offset, long length, boolean preload) throws IOException {
```
mikemccand: It's somewhat confusing that the `preload` parameter shadows the instance variable; maybe rename the instance variable to `globalPreload` or `preloadDefault` or so to prevent future confusion?

s1monw:
superseded by #758

PR #698, opened 2019-06-05
LUCENE-8834: Cache the SortedNumericDocValues.docValueCount() value whenever it is used in a loop

tpunder:


lucene/backward-codecs/src/java/org/apache/lucene/codecs/lucene70/Lucene70DocValuesConsumer.java
```diff
@@ -158,7 +158,8 @@ void nextBlock() {
     long gcd = 0;
     Set<Long> uniqueValues = new HashSet<>();
     for (int doc = values.nextDoc(); doc != DocIdSetIterator.NO_MORE_DOCS; doc = values.nextDoc()) {
-      for (int i = 0, count = values.docValueCount(); i < count; ++i) {
+      final int docValueCount = values.docValueCount();
+      for (int i = 0, count = docValueCount; i < count; ++i) {
```
jpountz: could you leave it the way it was before, it already cached the docValueCount? Maybe this way of caching the docValueCount is a bit better by the way as it gives a smaller scope to the variable?

tpunder: Whoops, you are right.  I totally missed that in that case it was already caching the value.  I will revert and audit the other changes.

tpunder: @jpountz I reverted my changes in the case where it was already cached.  It looks like that mostly scopes the changes to Solr with the exception of the Lucene JoinUtil.

Do you have a caching style preference of doing it inline in the `for` expression vs a separate variable before the loop?

I've seen both ways used in the codebase.  For example in LatLonPointDistanceComparator.java:

```java
    if (doc == currentDocs.docID()) {
      setValues();
      int numValues = currentDocs.docValueCount();
      for (int i = 0; i < numValues; i++) {
        long encoded = currentValues[i];
        double docLatitude = decodeLatitude((int)(encoded >> 32));
        double docLongitude = decodeLongitude((int)(encoded & 0xFFFFFFFF));
        minValue = Math.min(minValue, SloppyMath.haversinSortKey(latitude, longitude, docLatitude, docLongitude));
      }
    }
```

Versus the example you pointed out of:

```java
for (int i = 0, count = values.docValueCount(); i < count; ++i) {
```

jpountz: I'm fine with both. One thing that is nice with inlining it in the for loop is that the variable gets a smaller scope.

tpunder: Fine by me.  I've updated the PR to use the inline style except in the 2 cases where the count is used outside of the loop.

PR #700, opened 2019-06-06
LUCENE-8835: Respect file extension when listing files form FileSwitchDirectory

s1monw:
FileSwitchDirectory splits file actions between 2 directories based
on file extensions. The extensions are respected on write operations
like delete or create but ignored when we list the content of the
directories. Until now we only deduplicated the contents on
Directory#listAll which can cause inconsistencies and hard to debug
errors due to double deletions in IndexWriter is a file is pending
delete in one of the directories but still shows up in the directory
listing form the other directory. This case can happen if both
directories point to the same underlying FS directory which is a
common usecase to split between mmap and noifs.

This change filters out files from directories depending on their
file extension to make sure files that are deleted in one directory
are not returned form another if they point to the same fs directory.


lucene/test-framework/src/java/org/apache/lucene/util/LuceneTestCase.java
```diff
@@ -1414,6 +1415,9 @@ private static BaseDirectoryWrapper newFSDirectory(Path f, LockFactory lf, boole
       }
 
       Directory fsdir = newFSDirectoryImpl(clazz, f, lf);
+      if (rarely()) {
+        fsdir = new FileSwitchDirectory(Set.of("fdt", "fdx", "tim"), fsdir, newFSDirectoryImpl(clazz, f, lf), true);
```
mikemccand: We could maybe randomize which extensions we add to the set here for better testing?

s1monw: I pushed a new commit

lucene/test-framework/src/java/org/apache/lucene/util/LuceneTestCase.java
```diff
@@ -1416,7 +1416,11 @@ private static BaseDirectoryWrapper newFSDirectory(Path f, LockFactory lf, boole
 
       Directory fsdir = newFSDirectoryImpl(clazz, f, lf);
       if (rarely()) {
-        fsdir = new FileSwitchDirectory(Set.of("fdt", "fdx", "tim"), fsdir, newFSDirectoryImpl(clazz, f, lf), true);
+        List<String> fileExtensoions =
```
mikemccand: `fileExtensoions` -> `fileExtensions`

PR #701, opened 2019-06-06
LUCENE-8836 Optimize DocValues TermsDict to continue scanning from the last position when possible

bruno-roustant:


rmuir:
can the optimization be implemented in a less-invasive manner?

dsmiley:
How best might the performance of DocValues be evaluated?  luceneutil has a great query perf benchmark but the only thing related to DocValues I noticed was sorting

bruno-roustant:
"can the optimization be implemented in a less-invasive manner?"
Well, the change optimizes both seekExact(ord) and seekCeil(term), and it handles calls to next() to keep track of the last accessed term. This change also reduces the binary search range in TermsDict.seekTermsIndex(), by comparing to the last accessed term.
 I already tried to be minimal (although I added comments) so I don't think I can do with less.


"How best might the performance of DocValues be evaluated?"
The approach I took was to run some Lucene tests while counting the total number of seeks and terms read in the IndexInput, with and without the optimization.
TestLucene70DocValuesFormat - the optimization saves 24% seeks and 15% term reads.
TestDocValuesQueries - the optimization adds 0.7% seeks and 0.003% term reads.
TestDocValuesRewriteMethod.testRegexps - the optimization saves 71% seeks and 82% term reads.

rmuir:
maybe the better approach is top-down: for example starting at high level sorting/faceting algorithms, do they really go through any trouble to resolve ordinals in sequential order or do they just resolve the top-N in some arbitrary order by document count. 

IMO if this kind of code goes through the trouble to e.g. sort the top-N by ordinal, retrieve all the term metadata, then sort it back, then optimizations here make sense. But based on that code we can also determine if e.g. its only necessary to optimize say `seekCeil` versus also `seekExact`. It may simplify the low level code.

If the high level code isn't actually doing work to resolve ordinals in ordinal-order, then it won't help to optimize seek. Merging code etc is already using `next()` which is clearly optimized for that case.

msokolov:
bq. The approach I took was to run some Lucene tests while counting the total number of seeks and terms read in the IndexInput, with and without the optimization.

The numbers look compelling. Is this a fair test though? Doesn't it ignore the cost added by the optimization? We now do some work saving last term,etc. It seems as if it could be justified by reductions in seeking, but a test that shows that holistically would be good, ideally targeting some common use case. The unit tests might be exercising fairly artifical edge cases? EG we probably don't want to be optimizing slow exact range query when the caller would be better off using a points field. For DocValues, typical use cases do probably involve sorting/scoring/grouping/aggregations. Do you have a use case you are targeting and can share results on?

bruno-roustant:
> Is this a fair test though? Doesn't it ignore the cost added by the optimization?

The measure does not ignore the cost added by the optimization. It counts all seeks in the IndexInput, including the ones for the optimization, and it counts the next-block-first-term read added by the optimization also (counted as both a seek and a term read). So the measure is fair in itself.

But it may be more judicious to measure with some other tests that are closer to common use cases. I'll try to find some other classes to measure.


> Do you have a use case you are targeting and can share results on?

My use-case is the following. We have a list of values (up to 200 values) we search in the DocValues index.
If the index size is rather close to the number of values we search, then it is more efficient to iterate sequentially the DocValues index with the TermsDict TermsEnum.next().
But if the index size is large, then it is more efficient to search the values, sorted in lexicographical order, by seeking them in the TermsDict. That's where the optimization works. In this case we gain at least 25% performance.

jpountz:
I have some interest in this change too so I ported it to the current codec in a way that looks a bit less invasive as suggested above: https://github.com/apache/lucene/pull/827

PR #702, opened 2019-06-06
LUCENE-8837 smokeTestRelease.py option --download-only

janhoy:


janhoy:
I have tested these changes during the 7.7.2 release

dev-tools/scripts/scriptutil.py
```diff
@@ -66,14 +70,19 @@ def on_or_after(self, other):
            (self.bugfix > other.bugfix or self.bugfix == other.bugfix and
            self.prerelease >= other.prerelease)))
 
+  def gt(self, other):
```
janhoy: This is used to check if a given version is greater than current. Will be used in releaseWizard script to be committed later

dev-tools/scripts/scriptutil.py
```diff
@@ -66,14 +70,19 @@ def on_or_after(self, other):
...
+           (self.major == other.major and self.minor > other.minor) or
+           (self.major == other.major and self.minor == other.minor and self.bugfix > other.bugfix))
+
   def is_back_compat_with(self, other):
     if not self.on_or_after(other):
       raise Exception('Back compat check disallowed for newer version: %s < %s' % (self, other))
     return other.major + 1 >= self.major
 
-def run(cmd):
+def run(cmd, cwd=None):
```
janhoy: Support for cwd in the run command is useful so that the cmd can be relative to some folder

dev-tools/scripts/scriptutil.py
```diff
@@ -99,6 +108,18 @@ def update_file(filename, line_re, edit):
     f.write(''.join(buffer))
   return True
 
+
+def check_ant():
```
janhoy: Moved from buildAndPushRelease. Now also returns the actual ANT version. Will be used in releaseWizard script for warning about certain Ant versions.

dev-tools/scripts/smokeTestRelease.py
```diff
@@ -14,35 +16,31 @@
 # limitations under the License.
 
 import argparse
-import os
```
janhoy: Mostly auto rearranged imports. I could leave them as they were but this is more organized :)

PR #703, opened 2019-06-07
LUCENE-8838: Remove support for Steiner points

iverase:
Remove unused Steiner points support and fail if a hole is reduced to a point when all points are coplanar.


nknize:
LGTM. Thx for the cleanup!

PR #705, opened 2019-06-07
LUCENE-8842: add a pull request template

ctargett:
This adds a template for pull requests to the top level of the project.

I'm not sure if the commented out section at the top that explains how to create an issue would appear to users, I was never able to find out from other examples I looked at. If it does not, we could modify the template to make some variation of that text appear to users and they could delete it (although it's likely some would not...probably not the end of the world).


risdenk:
I vote for `.github` instead of in the root, but either way is fine with me.

If you want an example here is the one for Apache Knox: https://raw.githubusercontent.com/apache/knox/master/.github/PULL_REQUEST_TEMPLATE.md

It ends up looking like this:

Screen Shot 2019-06-07 at 3 36 16 PM

I haven't played around with the commented out sections.

risdenk:
The Elasticsearch one (https://raw.githubusercontent.com/elastic/elasticsearch/master/.github/PULL_REQUEST_TEMPLATE.md) has comments and looks like this:

Editing
Screen Shot 2019-06-07 at 3 41 04 PM

Preview
Screen Shot 2019-06-07 at 3 39 54 PM

risdenk:
So I think what is there is good since it won't be visible when saved, but it is there by default when you open the PR.

dsmiley:
I also suggest adding to `/.github/`.  Otherwise +1 and thanks!

ctargett:
I added a `.github` folder and moved the template to it. I'll merge this shortly.

PR #706, opened 2019-06-07
LUCENE-8843: Only ignore IOException on dirs when invoking force

jasontedor:
Today in the method IOUtils#fsync we ignore IOExceptions when fsyncing a directory. However, the catch block here is too broad, for example it would be ignoring IOExceptions when we try to open a non-existent file. This commit addresses that by scoping the ignored exceptions only to the invocation of FileChannel#force. This prevents us from suppressing an exception in case we run into an unexpected issue when opening the file.

However, fsyncing directories on Windows is not possible. We always suppressed this by allowing that an AccessDeniedException is thrown when attemping to open the directory for reading. Yet, per the above, this suppression also allowed other IOExceptions to be suppressed, and that should be considered a bug (e.g., not only the directory not existing, but any filesystem error and other reasons that we might get an access denied there, like genuine permissions issues). Rather than relying on exceptions for flow control and continuing to suppress there, we simply return early if attempting to fsync a directory on Windows (we should not put this burden on the caller).

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/util/IOUtils.java
+++ b/lucene/core/src/java/org/apache/lucene/util/IOUtils.java
@@ -18,6 +18,7 @@ package org.apache.lucene.util;
 
 import java.io.BufferedReader;
 import java.io.Closeable;
+import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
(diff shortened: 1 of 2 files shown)


jasontedor:
The associated issue is [LUCENE-8843](https://issues.apache.org/jira/projects/LUCENE/issues/LUCENE-8843).

PR #707, opened 2019-06-09
Use append() when StringBuilder is being used, instead of regular concat.

KoenDG:
This PR is related to the following ticket:

https://issues.apache.org/jira/browse/LUCENE-8847

As stated in the ticket, this was done in my spare time, please do not review this PR during your working hours. More details in the Jira ticket.

This specific commit affects all points in the codebase where the argument of a StringBuilder.append() call is itself a regular String concatenation.
This defeats the purpose of using StringBuilder and also introduces an extra alloction.
These changes should avoid that.

It's a big commit, yet it's 1 automated change, performed by Intellij's static code analysis.

ant tests have run, succeeded on local machine.

### DIFF
--- a/lucene/analysis/nori/src/java/org/apache/lucene/analysis/ko/tokenattributes/PartOfSpeechAttributeImpl.java
+++ b/lucene/analysis/nori/src/java/org/apache/lucene/analysis/ko/tokenattributes/PartOfSpeechAttributeImpl.java
@@ -80,7 +80,7 @@ public class PartOfSpeechAttributeImpl extends AttributeImpl implements PartOfSp
       if (builder.length() > 0) {
         builder.append("+");
       }
-      builder.append(morpheme.surfaceForm + "/" + morpheme.posTag.name() + "(" + morpheme.posTag.description() + ")");
+      builder.append(morpheme.surfaceForm).append('/').append(morpheme.posTag.name()).append('(').append(morpheme.posTag.description()).append(')');
     }
     return builder.toString();
   }
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -83,6 +83,10 @@ Test Framework
 * LUCENE-8825: CheckHits now display the shard index in case of mismatch
   between top hits. (Atri Sharma via Adrien Grand)
 
+Other
+
+* LUCENE-8847: Code Cleanup: Remove StringBuilder.append with concatted strings
+
 ======================= Lucene 8.1.1 =======================
 (No Changes)
 
--- a/lucene/analysis/smartcn/src/java/org/apache/lucene/analysis/cn/smart/hhmm/BiSegGraph.java
+++ b/lucene/analysis/smartcn/src/java/org/apache/lucene/analysis/cn/smart/hhmm/BiSegGraph.java
@@ -222,7 +222,7 @@ class BiSegGraph {
     Collection<ArrayList<SegTokenPair>>  values = tokenPairListTable.values();
     for (ArrayList<SegTokenPair> segList : values) {
       for (SegTokenPair pair : segList) {
-        sb.append(pair + "\n");
+        sb.append(pair).append("\n");
       }
     }
     return sb.toString();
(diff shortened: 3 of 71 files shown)


lucene/analysis/nori/src/java/org/apache/lucene/analysis/ko/tokenattributes/PartOfSpeechAttributeImpl.java
```diff
@@ -80,7 +80,7 @@ private String displayMorphemes(Morpheme[] morphemes) {
       if (builder.length() > 0) {
         builder.append("+");
       }
-      builder.append(morpheme.surfaceForm + "/" + morpheme.posTag.name() + "(" + morpheme.posTag.description() + ")");
+      builder.append(morpheme.surfaceForm).append("/").append(morpheme.posTag.name()).append("(").append(morpheme.posTag.description()).append(")");
```
martin-g: If you want to (micro)optimize even further you can use `#append(char)` instead of `#append(String)` for single characters.

solr/contrib/dataimporthandler/src/test/org/apache/solr/handler/dataimport/TestVariableResolverEndToEnd.java
```diff
@@ -77,34 +77,11 @@ protected String generateConfig() {
...
-        " {ts '2017-02-18 12:34:56'} as SECOND_TS, " +
-        " '${dataimporter.functions.formatDate(FIRST.FIRST_TS, 'yyyy'" + thirdLocaleParam + ")}' as SECOND1_S,  " +
-        " 'FISH' AS MEAT, " +
-        " 'FRY' AS METHOD, " +
-        " 'SIRLOIN' AS CUTS, " +
-        " 'BEEF_CUTS' AS WHATKIND " +
-        "from DUAL " +
-        "WHERE 1=${FIRST.ID} " +
-        "ORDER BY SORT \"" +
-        ">\n");
```
vthacker: I think we should leave out test files? The above is just so much more readable

vthacker:
Overall this looks good! I think we should not make these changes for the test files?

Also can you please add an entry under `lucene/CHANGES.txt` under the 8.2 section. You might want to add an "Other" section like 8.1 . For the Lucene-NNNN can you just prefix the PR number for now. 
I'm going to check if that's going to be encouraged going forward or not on this [thread](http://mail-archives.apache.org/mod_mbox/lucene-dev/201906.mbox/%3C00414C73-F0CE-4286-9F48-774D45E21A29@cominvent.com%3E)

KoenDG:
Alright, I'll do it again but without the test files and an entry in the CHANGES.txt file

lucene/analysis/nori/src/java/org/apache/lucene/analysis/ko/tokenattributes/PartOfSpeechAttributeImpl.java
KoenDG: While true that it is a micro-optimization, I do believe it is a solid one. As in, a change that will perform better every single time. Not just in edge cases. And it is a very small improvement, but an improvement none the less. And this is only 1 subject.

Making these changes, running the tests, writing the PR took a good 30 minutes(mostly waiting on the tests). For something that isn't work on a feature. So I perfectly understand nobody wants to put their time in that and I just feel like seeing how much I can contribute in this fashion.

uschindler:
Can you add a Changes Entry unter 8.2 in Lucene, don't add in Solr (not needed there).

KoenDG:
2 questions:

Do I reference the one I made on the SOLR project or the one I made on the LUCENE project(not sure I should have made that one).

This PR is against master, which I seemed to understand to be going to 9.0 eventually or am I wrong there? This is why I'm wondering why it should be under 8.2

Running tests now before comitting.

uschindler:
I will manage the merging afterwards. The changes entry will be in 8.2 as this is the first version where it will appear.

Don't add a changes entry to solr. Use the new issue number LUCENE-8847 and put it into Lucene's changes list. I can do this for you, as I will need to merge to my local checkout anyways.

KoenDG:
Right, push done, tests succeed locally, suggestions and entry to CHANGES.txt under lucene/ taken into account.

PR #708, opened 2019-06-10
LUCENE-8775: Fix bug when selecting bridge and improve tests

iverase:
When selecting a bridge, we need to check for shared vertex in its own loop. In addition the test now check that the area of the polygon is equal to the area of the tessellation.


iverase:
This is a follow up of LUCENE-8775 so I will push shortly

PR #709, opened 2019-06-11
LUCENE-8850: Calculate the area of a polygon and throw error when values are invalid

iverase:
Compute area of a polygon and throw an error when values are invalid.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/geo/Polygon.java
+++ b/lucene/core/src/java/org/apache/lucene/geo/Polygon.java
@@ -52,6 +52,8 @@ public final class Polygon {
   public final double maxLon;
   /** winding order of the vertices */
   private final WindingOrder windingOrder;
+  /** Area of the polygon */
+  private final double area;
 
   /**
    * Creates a new Polygon from the supplied latitude/longitude array, and optionally any holes.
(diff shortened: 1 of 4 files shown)


lucene/core/src/java/org/apache/lucene/geo/Polygon.java
```diff
@@ -96,34 +99,50 @@ public Polygon(double[] polyLats, double[] polyLons, Polygon... holes) {
...
+
   }
 
   /** returns the number of vertex points */
   public int numPoints() {
     return polyLats.length;
   }
 
+  /** returns the area of the polygon */
+  public double area() {
```
dsmiley: I can imagine some asking what the units are.  And why not getArea?  Maybe this should be marked @lucene.internal or package local even if we don't want to expose it.

iverase: I rename the method to getArea() and make it protected. I have no use case outside of validation so it is ok to make it package private.

iverase:
One of the things I notice with this change is that it breaks some of the benchmarks, in particular it breaks this one as the polygon contain invalid holes:

https://home.apache.org/~mikemccand/geobench.html#search-polyMedium

jpountz:
What is the particular rule that breaks?

iverase:
Reading one of the polygons representing a London Borough it fails because it contains a hole where all points are coplanar.

lucene/core/src/java/org/apache/lucene/geo/Polygon.java
nknize: I agree w/ @dsmiley re: units.  The calculation is using the [shoelace method](https://en.wikipedia.org/wiki/Shoelace_formula) on vertices in decimal degrees. So the units are deg^2. I remember exploring this a little over a year ago in [LUCENE-8364](https://issues.apache.org/jira/browse/LUCENE-8364) and originally having area because it was "easy". I ended up removing it until we had support for indexing in other projections (e.g., [equal area](https://en.wikipedia.org/wiki/Map_projection#Equal-area)). I think we should wait until [LUCENE-8632](https://issues.apache.org/jira/browse/LUCENE-8632) lands and only provide this area calculation in [XYPolygon](lucene/sandbox/src/java/org/apache/lucene/geo/XYPolygon.java) for regular cartesian geometries in non decimal degrees.

iverase: I agree we should not provide this info to the user, +1 not to offer the calculation to the user.

The idea behind this change was validation of the polygon validation. The idea of this change came from the fact that I am using this calculation to check that a Polygon tessellation is correct and I realise that:

- Currently we compute the signed area to check the orientation of a polygon, if the area is 0 the polygon is consider CW, should we fail instead?

- I have seen polygons with one hole where the hole and polygon are the same, should we capture that situation and fail?

dsmiley: Please add a comment to the code to to link to the "shoelace method" that Nick points at.  I think it's important for non-trivial algorithms to incorporate references where the readers of this code can learn more.  Who knows... maybe some "velcro method" will come along and be superior and then some student looking at this code will point this out to us :-)

nknize: > * Currently we compute the signed area to check the orientation of a polygon, if the area is 0 the polygon is considered CW, should we fail instead?

+1 to fail in this situation since it is not a valid polygon. 


> * I have seen polygons with one hole where the hole and polygon are the same, should we capture that situation and fail?

+1 to fail in this situation as well.

PR #712, opened 2019-06-11
SOLR-13534: Dynamic loading of jars from a remote url

noblepaul:


noblepaul:
SOLR-13534 : Dynamic loading to support loading jars from a URL (noble)

solr/core/src/java/org/apache/solr/core/BlobRepository.java
```diff
@@ -44,16 +45,22 @@
...
 
+import static org.apache.solr.common.SolrException.ErrorCode.SERVER_ERROR;
+import static org.apache.solr.common.SolrException.ErrorCode.SERVICE_UNAVAILABLE;
+import static org.apache.solr.common.cloud.ZkStateReader.BASE_URL_PROP;
+
 /**
  * The purpose of this class is to store the Jars loaded in memory and to keep only one copy of the Jar in a single node.
  */
 public class BlobRepository {
+  private static final long MAX_JAR_SIZE = Long.parseLong(System.getProperty("runtme.lib.size", String.valueOf(5* 1024*1024)));
```
mortenboldt: Is `runtme.lib.size` really correct? Seems like it should have been `runtime.lib.size`

PR #715, opened 2019-06-12
LUCENE-7714 Add a range query that takes advantage of index sorting.

jtibshirani:
I’m opening this draft PR to get feedback on an approach to [LUCENE-7714](https://issues.apache.org/jira/browse/LUCENE-7714).

The PR adds the new query type `IndexSortDocValuesRangeQuery`, a range query that takes advantage of the fact that the index is sorted on the same field as the query. It performs binary search on the field's doc values to find the doc IDs at the lower and upper ends of the range.

The query can only be used if all of these conditions hold:
- The index is sorted, and its primary sort is on the same field as the query.
- The query field has `SortedNumericDocValues`.
- The segments must have at most one field value per document (otherwise we cannot easily
  determine the matching document IDs through a binary search).

I was hoping for feedback on the overall approach, and also had a few open questions:
- I wasn’t sure on the best way to structure the query. As it stands, it requires that segments are sorted correctly and contain at most one value per document. Perhaps we could introduce a wrapper query that can make the decision on a segment-by-segment basis: it would only use `IndexSortDocValueRangeQuery` if the right conditions are met, and otherwise fall back to a standard range query. This wrapper query would have similarities to `IndexOrDocValuesQuery`.
- Because doc values only support forward iteration, we need to recreate the comparators every time we backtrack in the binary search. I assumed this recreation would be expensive and experimented with some strategies to avoid it, such as starting with a shared binary search that checks both `lowerValue` and `upperValue`, then moving on to the two individual binary searches. However, these efforts didn’t show any performance improvements in my benchmarks. I plan to do more research around sparse + blocked doc values to understand the circumstances in which reloading can be more expensive.

**Benchmarking results**
I ingested part of the [http logs](https://github.com/elastic/rally-tracks/tree/master/http_logs) dataset (123M total documents) into an index sorted by `@timestamp`. Every document has a single `@timestamp`, although they are not unique across documents.

The following date ranges were tested:
- range with single point [897303051, 897303051], 124 docs
- small range (897633930, 897655999], ~41K docs
- medium range (897623930, 897655999],  ~5M docs
- large range (897259801, 897503930], ~21M docs

```
| 50th percentile service time |            range-single |   7.50658 |     ms |
| 90th percentile service time |            range-single |   7.81589 |     ms |
| 50th percentile service time |  range-optimized-single |   8.02532 |     ms |
| 90th percentile service time |  range-optimized-single |   8.38834 |     ms |

| 50th percentile service time |               range-small |   12.7203 |     ms |
| 90th percentile service time |               range-small |   13.3798 |     ms |
| 50th percentile service time |     range-optimized-small |   7.43383 |     ms |
| 90th percentile service time |     range-optimized-small |   7.67655 |     ms |

| 50th percentile service time |               range-medium |   20.8554 |     ms |
| 90th percentile service time |               range-medium |   22.1637 |     ms |
| 50th percentile service time |     range-optimized-medium |   8.38864 |     ms |
| 90th percentile service time |     range-optimized-medium |   8.66427 |     ms |

| 50th percentile service time |           range-large |   53.8697 |     ms |
| 90th percentile service time |           range-large |   60.2573 |     ms |
| 50th percentile service time | range-optimized-large |   7.81584 |     ms |
| 90th percentile service time | range-optimized-large |   8.02755 |     ms |
```


lucene/core/src/java/org/apache/lucene/document/IndexSortDocValuesRangeQuery.java
```diff
@@ -0,0 +1,279 @@
...
+        }
+
+        SortField sortField = indexSort.getSort()[0];
+        if (sortField.getField().equals(field) == false) {
+          throw new IllegalArgumentException("To use IndexSortDocValuesRangeQuery, the index must be sorted " +
+              "on the same field as the query field.");
+        }
+
+        SortedNumericDocValues sortedDocValues = reader.getSortedNumericDocValues(field);
+        NumericDocValues docValues = DocValues.unwrapSingleton(sortedDocValues);
```
atris: I am curious to understand why we cannot support multiple values here. Is it because of duplicate elimination in binary searches?

lucene/core/src/java/org/apache/lucene/document/IndexSortDocValuesRangeQuery.java
```diff
@@ -0,0 +1,279 @@
...
+      upperValue = temp;
+    }
+
+    // Perform a binary search to find the first document with value >= lowerValue.
+    ValueComparator comparator = loadComparator(sortField, lowerValue, context);
+    int low = 0;
+    int high = maxDoc - 1;
+
+    while (low <= high) {
+      int mid = (low + high) >>> 1;
```
atris: Let's ensure we take care of the known edge cases of binary ranges here. Maybe add some targeted tests?

lucene/core/src/java/org/apache/lucene/document/IndexSortDocValuesRangeQuery.java
```diff
@@ -0,0 +1,279 @@
...
+      int value = leafFieldComparator.compareTop(doc);
+      return direction * value;
+    };
+  }
+
+  /**
+   * A doc ID set iterator that wraps a delegate iterator and only returns doc IDs in
+   * the range [firstDocInclusive, lastDoc).
+   */
+  private static class BoundedDocSetIdIterator extends DocIdSetIterator {
```
atris: Do we need this class? Why not populate a DocSetIdIterator with just the relevant DocValues and return, much like what PointRangeQuery does?

lucene/core/src/java/org/apache/lucene/document/IndexSortDocValuesRangeQuery.java
```diff
@@ -0,0 +1,279 @@
...
+    // of the binary search to the result of the previous search.
+    comparator = loadComparator(sortField, upperValue, context);
+    low = firstDocIdInclusive;
+    high = maxDoc - 1;
+
+    while (low <= high) {
+      int mid = (low + high) >>> 1;
+      if (comparator.compare(mid) < 0) {
+        high = mid - 1;
+        comparator = loadComparator(sortField, upperValue, context);
```
atris: As mentioned above, can we ensure the cost of this iteration does not override the benefits?

lucene/core/src/java/org/apache/lucene/document/IndexSortDocValuesRangeQuery.java
```diff
@@ -0,0 +1,279 @@
...
+    // Perform a binary search to find the first document with value >= lowerValue.
+    ValueComparator comparator = loadComparator(sortField, lowerValue, context);
+    int low = 0;
+    int high = maxDoc - 1;
+
+    while (low <= high) {
+      int mid = (low + high) >>> 1;
+      if (comparator.compare(mid) <= 0) {
+        high = mid - 1;
+        comparator = loadComparator(sortField, lowerValue, context);
```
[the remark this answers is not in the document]
jimczi: Can you elaborate on this, the worst case runs in logarithmic time so I don't understand the concern.

lucene/core/src/java/org/apache/lucene/document/IndexSortDocValuesRangeQuery.java
jimczi: Because we don't need to create a bitset beforehand. Unlike points docvalues iterator access the values in the order of the internal document id so can use the original iterator there.

lucene/core/src/java/org/apache/lucene/document/IndexSortDocValuesRangeQuery.java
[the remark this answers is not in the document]
atris: My point was about the number of comparisons we do. In the worst case, the number of comparisons done would be log(n), and for each of them, we would reload the comparator. I am curious to understand the cost of loading the comparator for different scenarios i.e large n, dense docValues etc. I think we should run some tests where we take a large data set, force merge it to a single segment, sort it by a dense field and run a query which would require a large range.

lucene/core/src/java/org/apache/lucene/document/IndexSortDocValuesRangeQuery.java
atris: Got it, so the idea is to have the same original iterator wrapped in multiple BoundedDocSetIdIterators, each with a different range.

lucene/core/src/java/org/apache/lucene/document/IndexSortDocValuesRangeQuery.java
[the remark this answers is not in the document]
jtibshirani: @atris the issue description includes benchmarks on a realistic dataset with dense docvalues. I share your concern that this reload may be expensive and plan to run more benchmarks with different docvalues types (including sparse docvalues).

lucene/core/src/java/org/apache/lucene/document/IndexSortDocValuesRangeQuery.java
jtibshirani: When a document has multiple values, the index sort must choose one of them to decide where it should be placed. So a binary search may miss a document where the value we’ve sorted on is outside the range, but it also contains a value that falls within the range.

jtibshirani:
Thanks @atris and @jimczi for taking a look!

> However I wonder if we should expose this query as is or if we should use it only internally in the `IndexOrDocValuesQuery` ?

I agree that the query is not so helpful on its own. I was unsure about integrating it into `IndexOrDocValuesQuery`, since that deals with queries generally and this query is specifically for `long` ranges. Would `IndexOrDocValuesQuery` optionally accept a third query, and only run it if the segment is sorted and also contains `NumericDocValues`? That seemed a bit specific to add to a fairly general query type.

Another idea is for `IndexOrDocValuesRangeQuery` to accept a fallback range query, and delegate to it if the necessary conditions aren't met?

lucene/core/src/java/org/apache/lucene/document/IndexSortDocValuesRangeQuery.java
```diff
@@ -0,0 +1,279 @@
...
+   */
+  private interface ValueComparator {
+    int compare(int docID) throws IOException;
+  }
+
+  private static ValueComparator loadComparator(SortField sortField,
+                                                long topValue,
+                                                LeafReaderContext context) throws IOException {
+    @SuppressWarnings("unchecked")
+    FieldComparator<Long> fieldComparator = (FieldComparator<Long>) sortField.getComparator(1, 0);
```
jimczi: Since we only need to compare values from a single source I don't think we should use a FieldComparator here. Using the docvalues iterator directly should be possible so we don't need to add this extra indirection ?

jtibshirani:
> I also agree that IndexOrDocValuesQuery is too generic to add the logic there but we could start by adding the optimization in SortedNumericDocValuesField#newSlowRangeQuery and NumericDocValuesField#newSlowRangeQuery. Then we can discuss the best way to ensure that IndexOrDocValuesQuery choose the docvalues one when the optimization can be applied. Does that makes sense ?

Got it, I understand your recommendation now -- I will give that a try and push some new commits.

lucene/core/src/java/org/apache/lucene/document/IndexSortDocValuesRangeQuery.java
jtibshirani: I used a field comparator because it dealt with missing values (otherwise I would have to handle missing values explicitly and also make sure to use the right default missing value). This approach felt a bit more solid, but I don't have a strong preference.

jimczi: Got it thanks, we need to handle missing values too so +1 for this approach.

jtibshirani:
@jimczi I tried moving the optimization into `SortedNumericDocValuesRangeQuery`. I had a couple questions:
- It is now possible that `match` and `explain` will be slower for `SortedNumericDocValuesRangeQuery`, since we may perform a binary search instead of just checking a single document value. Perhaps I should override these methods and ensure we always use the old approach?
- I don't see a good way to integrate the optimization into `IndexOrDocValuesQuery`, any guidance would be very welcome.

jimczi:
We discussed offline with Julie and agreed that it doesn't really make sense to add this optimization in `IndexOrDocValuesQuery`. If range queries are fast with docvalues on a field that is used to sort the index then it will become useless to also index the field as a point. For multi-valued fields points will remain the best solution since the optimization on docvalues cannot work in this case but it should be enough to introduce the optim in  `SortedNumericDocValuesField#newSlowRangeQuery` and `NumericDocValuesField#newSlowRangeQuery` or in its own query if we don't want to mix.

> It is now possible that match and explain will be slower for SortedNumericDocValuesRangeQuery, since we may perform a binary search instead of just checking a single document value. Perhaps I should override these methods and ensure we always use the old approach?

Agreed, the old approach makes more sense for these cases that advance to a specific document directly.

jtibshirani:
Thanks @jimczi, I pushed another commit that makes sure to use the two-phase iterator for explain and match.

jtibshirani:
> Instead of adding this alternative in the NumericDocValuesField#slowRangeQuery we could add an alternative query in the SortedNumericDocValuesRangeQuery that would be used if the condition to run in the sorted case are not met (multi-valued field or a different default value for instance).

This makes the most sense to me in terms of design, I had actually considered it originally but wasn't totally sure about it!

I pushed some new changes that move the logic to a new query `IndexSortSortedNumericDocValuesRangeQuery` in sandbox. I undid the changes to the existing `SortedNumericDocValuesRangeQuery`. For now, the new query only handles `SortedNumericDocValues` fields, since I think that is the most important use case and it was a bit tricky to add support for `NumericDocValues` as well.

lucene/sandbox/src/java/org/apache/lucene/search/IndexSortSortedNumericDocValuesRangeQuery.java
```diff
@@ -0,0 +1,305 @@
...
+    FieldComparator<Long> fieldComparator = (FieldComparator<Long>) sortField.getComparator(1, 0);
+    fieldComparator.setTopValue(topValue);
+
+    LeafFieldComparator leafFieldComparator = fieldComparator.getLeafComparator(context);
+    int direction = sortField.getReverse() ? -1 : 1;
+
+    return doc -> {
+      int value = leafFieldComparator.compareTop(doc);
+      return direction * value;
+    };
```
jpountz: I think we need to do the comparison manually instead of using the comparator because of missing values. The comparator will consider documents that don't have a value and documents that have the missing value equal, which doesn't work for us.

lucene/sandbox/src/java/org/apache/lucene/search/IndexSortSortedNumericDocValuesRangeQuery.java
```diff
@@ -0,0 +1,305 @@
...
+    if (lowerValue == Long.MIN_VALUE && upperValue == Long.MAX_VALUE) {
+      return new DocValuesFieldExistsQuery(field);
+    }
+
+    Query rewrittenFallback = fallbackQuery.rewrite(reader);
+    if (rewrittenFallback == fallbackQuery) {
+      return this;
+    } else {
+      return new IndexSortSortedNumericDocValuesRangeQuery(
+          field, lowerValue, upperValue, fallbackQuery);
```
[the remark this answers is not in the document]
atris: Should we add a test for this?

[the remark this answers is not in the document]
jpountz: :+1:

lucene/sandbox/src/java/org/apache/lucene/search/IndexSortSortedNumericDocValuesRangeQuery.java
```diff
@@ -0,0 +1,305 @@
...
+   * The returned {@link DocIdSetIterator} makes sure to wrap the original docvalues to skip
+   * over documents with no value.
+   */
+  private DocIdSetIterator getDocIdSetIterator(SortField sortField,
+                                               LeafReaderContext context,
+                                               DocIdSetIterator delegate) throws IOException {
+    int maxDoc = context.reader().maxDoc();
+    if (maxDoc <= 0) {
+      return DocIdSetIterator.empty();
+    }
```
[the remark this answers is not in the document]
jtibshirani: 👍

lucene/sandbox/src/java/org/apache/lucene/search/IndexSortSortedNumericDocValuesRangeQuery.java
```diff
@@ -0,0 +1,305 @@
...
+  }
+
+  @Override
+  public Weight createWeight(IndexSearcher searcher, ScoreMode scoreMode, float boost) throws IOException {
+    Weight fallbackWeight = fallbackQuery.createWeight(searcher, scoreMode, boost);
+
+    return new ConstantScoreWeight(this, boost) {
+      @Override
+      public Scorer scorer(LeafReaderContext context) throws IOException {
+        SortedNumericDocValues values = context.reader().getSortedNumericDocValues(field);
```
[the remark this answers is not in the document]
jtibshirani: 👍

lucene/sandbox/src/java/org/apache/lucene/search/IndexSortSortedNumericDocValuesRangeQuery.java
```diff
@@ -0,0 +1,305 @@
...
+
+/**
+ * A range query that can take advantage of the fact that the index is sorted to speed up
+ * execution. If the index is sorted on the same field as the query, it performs binary
+ * search on the field's {@link SortedNumericDocValues} to find the documents at the lower
+ * and upper ends of the range.
+ * 
+ * This optimized execution strategy is only used if the following conditions hold:
+ * - The index is sorted, and its primary sort is on the same field as the query.
+ * - The segments must have at most one field value per document (otherwise we cannot easily
```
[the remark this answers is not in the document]
jtibshirani: 👍

lucene/sandbox/src/java/org/apache/lucene/search/IndexSortSortedNumericDocValuesRangeQuery.java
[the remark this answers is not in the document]
jtibshirani: Oops, thanks for catching this! Will add a test.

jtibshirani:
Thanks @jpountz for the review, it's now ready for another look.

jimczi:
I merged the pr manually, hence closing.

PR #716, opened 2019-06-12
LUCENE-8853: Try parsing original file extension from tmp file

s1monw:
FileSwitchDirectory failes if the tmp file are not in the same directory
as the file it's renamed to. This is correct behavior but breaks with
tmp files used with index sorting. This change tries best effort to find
the right extension directory if the file ends with `.tmp`


lucene/core/src/java/org/apache/lucene/store/FileSwitchDirectory.java
```diff
@@ -141,7 +147,14 @@ public static String getExtension(String name) {
     if (i == -1) {
       return "";
     }
-    return name.substring(i+1, name.length());
+    String ext = name.substring(i + 1);
+    if (ext.equals("tmp")) {
+      Matcher matcher = EXT_PATTERN.matcher(name);
+      if (matcher.find()) {
+        return matcher.group(1);
+      }
```
mikemccand: Hmm should we throw exception in `else` clause here?  Force callers to use consistent `tmp` file naming?

s1monw:
> Or, maybe we could instead require anywhere in Lucene that creates temp files to preserve the final extension in the naming of its temp file, and fix our atomic rename to enforce that if it's renaming a tmp file the extensions must match? It's not too many places where we create temp files?

I don't think we should have such a restriction just for FileSwitchDirectory. I'd rather deprecate it and build a better alternative for situations where we want to use NIOFS by default and only mmap some parts etc.

lucene/core/src/java/org/apache/lucene/store/FileSwitchDirectory.java
s1monw: that's difficult since it's ok to create a tmp file with a prefix of "foo" and a suffix of "bar" that will result in no match. I think we can only do the best effort on the rename end. I pushed a new commit to make the heuristic better. I thin that's all we can do.

s1monw:
@mikemccand @jpountz can you please take another look

lucene/core/src/java/org/apache/lucene/store/FileSwitchDirectory.java
mikemccand: OK

PR #721, opened 2019-06-15
LUCENE-8781: add FST array-with-gap addressing to Util.readCeilArc

msokolov:
This method got overlooked when I recently added the new FST encoding. Thanks to @dsmiley  for uncovering this.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/util/fst/Util.java
+++ b/lucene/core/src/java/org/apache/lucene/util/fst/Util.java
@@ -924,7 +924,7 @@ public final class Util {
   */
 
   /**
-   * Reads the first arc greater or equal that the given label into the provided
+   * Reads the first arc greater or equal than the given label into the provided
    * arc in place and returns it iff found, otherwise return <code>null</code>.
    * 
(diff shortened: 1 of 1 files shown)


dsmiley:
Thanks.  Did you try this out / test it?  I recommend running the tests while changing the default FST to use array-with-gap addressing.

msokolov:
Yes, naturally, since that was what this was intended to fix. I thought I had said so, but I see there's no such comment here. Anyway I had a seed that failed before and succeeded after. In fact you didn't need any special seed, since it always failed before. One thing that confuses me is that the suggesters also seem to use this method, yet they never failed in my testing. I'm running again just to be sure, then I'll push.

msokolov:
Yes, naturally, since that was what this was intended to fix. I thought I had said so, but I see there's no such comment here. Anyway I had a seed that failed before and succeeded after. In fact you didn't need any special seed, since it always failed before. One thing that confuses me is that the suggesters also seem to use this method, yet they never failed in my testing. I'm running again just to be sure, then I'll push.

PR #722, opened 2019-06-15
LUCENE-8863: enable loading external Kuromoji dictionary 

msokolov:
…abling 13-bit ids

I'm sharing this without tests, for comment, while I try to get the tests running. I'm having trouble with

```
    cd ~/workspace/lucene-solr/lucene/analysis/kuromoji; ant test-tools
```

first I had to make this patch, 

```
   <target name="test-tools" depends="compile-tools-tests">
-    <test-macro dataDir="src/tools/test" junit.classpath="tools.test.classpath"/>
+    <test-macro workDir="src/tools/test" junit.classpath="tools.test.classpath"/>
   </target>
```

now I'm getting this error:

```
/home/ANT.AMAZON.COM/sokolovm/workspace/lucene-solr/lucene/common-build.xml:964: Problem: failed to create task or type antlib:com.carrotsearch.junit4:pickseed
Cause: The name is undefined.
```

which all makes me wonder whether test-tools has become obsolete?

### DIFF
--- a/lucene/analysis/kuromoji/build.xml
+++ b/lucene/analysis/kuromoji/build.xml
@@ -136,8 +136,8 @@
      </compile>
   </target>
 
-  <target name="test-tools" depends="compile-tools-tests">
-    <test-macro dataDir="src/tools/test" junit.classpath="tools.test.classpath"/>
+  <target name="test-tools" depends="install-junit4-taskdef, compile-tools-tests">
+    <test-macro testsDir="${build.dir}/classes/tools-test" workDir="src/tools/test" junit.classpath="tools.test.classpath"/>
   </target>
 
   <target name="compile-test" depends="module-build.compile-test, compile-tools-tests"/>
--- a/lucene/analysis/kuromoji/src/java/org/apache/lucene/analysis/ja/dict/BinaryDictionary.java
+++ b/lucene/analysis/kuromoji/src/java/org/apache/lucene/analysis/ja/dict/BinaryDictionary.java
@@ -23,6 +23,8 @@ import java.io.IOException;
 import java.io.FileNotFoundException;
 import java.io.InputStream;
 import java.nio.ByteBuffer;
+import java.nio.file.Files;
+import java.nio.file.Paths;
 import java.nio.channels.Channels;
 import java.nio.channels.ReadableByteChannel;
 
@@ -37,6 +39,10 @@ import org.apache.lucene.util.IOUtils;
  */
 public abstract class BinaryDictionary implements Dictionary {
   
+  enum ResourceScheme {
+    CLASSPATH, FILE
+  }
+
   public static final String DICT_FILENAME_SUFFIX = "$buffer.dat";
   public static final String TARGETMAP_FILENAME_SUFFIX = "$targetMap.dat";
   public static final String POSDICT_FILENAME_SUFFIX = "$posDict.dat";
@@ -46,6 +52,8 @@ public abstract class BinaryDictionary implements Dictionary {
   public static final String POSDICT_HEADER = "kuromoji_dict_pos";
   public static final int VERSION = 1;
   
+  private final ResourceScheme resourceScheme;
+  private final String resourcePath;
   private final ByteBuffer buffer;
   private final int[] targetMapOffsets, targetMap;
   private final String[] posDict;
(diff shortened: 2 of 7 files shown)


lucene/analysis/kuromoji/src/tools/java/org/apache/lucene/analysis/ja/util/BinaryDictionaryWriter.java
```diff
@@ -126,7 +132,9 @@ public int put(String[] entry) {
     buffer.putShort(wordCost);
 
     if ((flags & BinaryDictionary.HAS_BASEFORM) != 0) {
-      assert baseForm.length() < 16;
+      if (baseForm.length() >= 16) {
```
rmuir: this isn't equivalent to the previous assert.

lucene/analysis/kuromoji/src/tools/java/org/apache/lucene/analysis/ja/util/BinaryDictionaryWriter.java
```diff
@@ -35,9 +35,11 @@
 import org.apache.lucene.analysis.ja.dict.BinaryDictionary;
 
 public abstract class BinaryDictionaryWriter {
+  private final static int ID_LIMIT = 8192;
+
   protected final Class<? extends BinaryDictionary> implClazz;
   protected ByteBuffer buffer;
-  private int targetMapEndOffset = 0, lastWordId = -1, lastSourceId = -1;
+  protected int targetMapEndOffset = 0, lastWordId = -1, lastSourceId = -1;
```
rmuir: what happened here? why is this protected now?

lucene/analysis/kuromoji/src/tools/java/org/apache/lucene/analysis/ja/util/BinaryDictionaryWriter.java
msokolov: Could you elaborate? I do see the exception message is wrong (should say >=), but I think the assertion is correct" `len < 16` iff `!(len >= 16)`

lucene/analysis/kuromoji/src/tools/java/org/apache/lucene/analysis/ja/util/BinaryDictionaryWriter.java
msokolov: my bad - this was left this way from some earlier debugging efforts. I'll undo

lucene/analysis/kuromoji/src/tools/java/org/apache/lucene/analysis/ja/util/BinaryDictionaryWriter.java
rmuir: I am referring to the exception message. this is one reason why the `assert` works well, such bugs are impossible, just as long as its run with `-ea` which the build file does :)

msokolov:
I posted an updated patch that addresses the comments, and also adds a test that exercises the builder with some test data. In order to test, I needed to add the ability to load an "external" dictionary; this enables loading a system dictionary from an alternate location on the classpath, or from an external file.

rmuir:
+1, thank you for taking the time to do these tests. Can we open a followup issue to fix whatever we need to fix so that your tests actually run every time with `ant test`?

I think its an important step so that we can support different dictionary variants and so on, but it doesn't need to block this PR.

lucene/analysis/kuromoji/build.xml
```diff
@@ -136,8 +136,8 @@
      </compile>
   </target>
 
-  <target name="test-tools" depends="compile-tools-tests">
-    <test-macro dataDir="src/tools/test" junit.classpath="tools.test.classpath"/>
+  <target name="test-tools" depends="install-junit4-taskdef, compile-tools-tests">
+    <test-macro testsDir="${build.dir}/classes/tools-test" workDir="src/tools/test" junit.classpath="tools.test.classpath"/>
```
rmuir: just an idea for the future: if we are fixing this tooling (e.g. replacing bogus asserts with real checks so it is usable like this PR does), maybe we should simply move all the `tools*` stuff and tests into the regular source tree?

msokolov: yeah, this felt kind of like a dusty corner where things would easily get lost. I'm just learning about the tools and modules concepts in lucene builds. It seems fine to have the builder available in the analysis jar.

lucene/analysis/kuromoji/src/java/org/apache/lucene/analysis/ja/dict/BinaryDictionary.java
```diff
@@ -46,13 +50,31 @@
...
   protected BinaryDictionary() throws IOException {
+    this(CLASSPATH);
+  }
+
+  /**
+   * @param resourcePrefix - where to load resources (dictionaries) from. Either "file:", followed
+   * by a filesystem directory and optionally a filename-prefix, or "classpath:", followed by a
+   * classpath location prefix, or just plain "classpath:" to use this class's name as the prefix.
+   */
+  protected BinaryDictionary(String resourcePrefix) throws IOException {
```
mocobeta: If the resourcePrefix parameter takes only "classpath:" or "file:", would it be possible to use enum here instead of arbitrary String (for safety)?.

mocobeta: And just wondering, why do we need "file:" prefix? I feel like that loading the dictionary data from arbitrary location could be problematic... (Am I missed anything?)

msokolov: enum is a good idea; I'll update the patch. I added "file:" to enable unit testing. We can't load a test dictionary from the classpath, I think. Another possibility would be to pass in dictionaries directly somehow. EG we could maybe have {{...DictionaryBuilder}} {{FST.save()}} to an in-memory buffer, but this seemed like a more intrusive change since we can already save to a set of files.

mocobeta: I missed the tests. OK, thanks!

msokolov:
Sorry for all the close/open/close PR noise. I was fiddling around because I don't seem to have permission to push from github, although I have "apache member" on my profile, and I can push from command line (with login prompt)

PR #723, opened 2019-06-15
SOLR-13545: AutoClose stream in ContentStreamUpdateRequest

colvinco:
# Description

As discussed on SOLR-13545, ContentStreamUpdateRequest doesn't close the InputStream it creates, after a change made in SOLR-12142

# Solution

Use try-with-resources to autoclose the stream

# Tests

I've updated existing tests that make update requests backed by file streams, so that they attempt to delete the file they have streamed to Solr. If the stream isn't closed, the test will fail on Windows, since the file will be locked. On Linux the file will appear to be deleted even if the stream isn't immediately closed.

I think the alternative would be to change the request classes to expose the InputStream so that it can be mocked or spied on, to explicitly check that close is called, but that also feels fragile.

# Checklist

- [X] I am authorized to contribute this code to the ASF and have removed any code I do not have a license to distribute.
- [] I have added documentation for the Ref Guide (for Solr changes only).

### DIFF
--- a/solr/solrj/src/test/org/apache/solr/client/solrj/SolrExampleTests.java
+++ b/solr/solrj/src/test/org/apache/solr/client/solrj/SolrExampleTests.java
@@ -25,6 +25,7 @@ import java.io.ByteArrayInputStream;
 import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.OutputStream;
 import java.lang.invoke.MethodHandles;
 import java.nio.ByteBuffer;
 import java.nio.file.Files;
@@ -53,6 +54,9 @@ import org.apache.solr.client.solrj.request.ContentStreamUpdateRequest;
 import org.apache.solr.client.solrj.request.LukeRequest;
 import org.apache.solr.client.solrj.request.MultiContentWriterRequest;
 import org.apache.solr.client.solrj.request.QueryRequest;
+import org.apache.solr.client.solrj.request.RequestWriter;
+import org.apache.solr.client.solrj.request.RequestWriter.ContentWriter;
+import org.apache.solr.client.solrj.request.RequestWriter.StreamCopyContentWriter;
 import org.apache.solr.client.solrj.request.StreamingUpdateRequest;
 import org.apache.solr.client.solrj.request.UpdateRequest;
 import org.apache.solr.client.solrj.response.FacetField;
@@ -78,6 +82,7 @@ import org.apache.solr.common.util.Pair;
 import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
+import org.mockito.Mockito;
 import org.noggit.JSONParser;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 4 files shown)


solr/solrj/src/test/org/apache/solr/client/solrj/SolrExampleTests.java
```diff
@@ -710,10 +714,18 @@ public void testContentStreamRequest() throws Exception {
     Assert.assertEquals(0, rsp.getResults().getNumFound());
 
     ContentStreamUpdateRequest up = new ContentStreamUpdateRequest("/update");
-    up.addFile(getFile("solrj/books.csv"), "application/csv");
+    // Create a copy of the file, which can be deleted after uploading.
+    // If the stream isn't closed, the file deletion will fail on Windows,
+    // though it will succeed on linux regardless
+    final File file = new File("temp/books_copy.csv");
```
dsmiley: I see you're creating a file relative to the current working directory.  That's a bad practice for tests generally.  Call `createTempFile(...)` instead (declared in the inheritance chain).

colvinco: Thanks, for some reason I thought that createTempFile() wouldn't work for what I was doing (too late at night). I've changed that

solr/solrj/src/test/org/apache/solr/client/solrj/SolrExampleTests.java
```diff
@@ -724,10 +736,18 @@ public void testStreamingRequest() throws Exception {
     client.commit();
     QueryResponse rsp = client.query( new SolrQuery( "*:*") );
     Assert.assertEquals(0, rsp.getResults().getNumFound());
+    // Create a copy of the file, which can be deleted after uploading.
+    // If the stream isn't closed, the file deletion will fail on Windows,
+    // though it will succeed on linux regardless
```
noblepaul: The test is really useful only in Windows? 

Can we just create a FakeInputStream and ensure that `close()` is called?

colvinco: Yes it is. While that can be done for the ContentSteamUpdateRequest, it can't be done for StreamingUpdateRequest without changing the class itself to expose the creation of the steam that it uses... And I didn't want to just go and do that. And it seemed logical to add a test for it, and to use the same logic for both, since they could both suffer from the same problem if changed in the future.

I could refactor both to use something common to stream and close, which would be more easily unit testable?

noblepaul: yes, if it is easily unit testable it is better. As most of us develop in non-windows machines, this test becomes essentially useless

dsmiley: We do have CI infrastructure that runs on Windows, so a test that only fails on Windows is not useless.  Nonetheless, it's better if it can test everywhere, though not essential in my mind if it's hard in this case.  Your call.

colvinco:
I saw @mkhludnev has patched the ContentStreamUpdateRequest and covered it in the test. So I don't mind dropping this totally.

I'll offer this as a way of testing that the StreamingUpdateRequest also closes its underlying stream... I made ContentStreamUpdateRequest and StreamingUpdateRequest both use a common ContentWriter, which can be tested on its own, and asserted that both of the request classes use that ContentWriter.

I don't like writing tests that are this closely coupled with the implementation, but it seemed like a reasonable compromise. But like I say, I don't mind dropping this

ctargett:
Confused what's happening with this PR - https://issues.apache.org/jira/browse/SOLR-13545 is closed, so can this be closed? Is there something here that should be spun into a new Jira issue for future contribution?

gd-spb-e1m:
> So I don't mind dropping this totally.

Yes. Please close it.

PR #725, opened 2019-06-17
LUCENE-8865: Use incoming thread for execution if IndexSearcher has an executor

s1monw:
Today we don't utilize the incoming thread for a search when IndexSearcher
has an executor. This thread is only idling but can be used to execute a search
once all other collectors are dispatched.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/IndexSearcher.java
+++ b/lucene/core/src/java/org/apache/lucene/search/IndexSearcher.java
@@ -30,6 +30,8 @@ import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 
 import org.apache.lucene.document.Document;
 import org.apache.lucene.index.DirectoryReader;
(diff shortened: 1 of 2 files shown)


s1monw:

lucene/core/src/java/org/apache/lucene/search/IndexSearcher.java
```diff
@@ -636,16 +636,45 @@ public TopFieldDocs reduce(Collection<TopFieldCollector> collectors) throws IOEx
...
       final Weight weight = createWeight(query, scoreMode, 1);
       final List<Future<C>> topDocsFutures = new ArrayList<>(leafSlices.length);
+
       for (int i = 0; i < leafSlices.length; ++i) {
         final LeafReaderContext[] leaves = leafSlices[i].leaves;
         final C collector = collectors.get(i);
-        topDocsFutures.add(executor.submit(new Callable<C>() {
-          @Override
-          public C call() throws Exception {
+        if (i == leafSlices.length-1) { // execute the last on the caller thread
```
atris: Nit: Spacing

lucene/core/src/java/org/apache/lucene/search/IndexSearcher.java
```diff
@@ -636,16 +636,45 @@ public TopFieldDocs reduce(Collection<TopFieldCollector> collectors) throws IOEx
       query = rewrite(query);
       final Weight weight = createWeight(query, scoreMode, 1);
       final List<Future<C>> topDocsFutures = new ArrayList<>(leafSlices.length);
+
```
atris: Nit: Unnecessary newline

lucene/core/src/java/org/apache/lucene/search/IndexSearcher.java
```diff
@@ -636,16 +636,45 @@ public TopFieldDocs reduce(Collection<TopFieldCollector> collectors) throws IOEx
...
+            @Override
+            public C get() {
+              return collector;
+            }
+
+            @Override
+            public C get(long timeout, TimeUnit unit) {
+              return get();
+            }
```
atris: Nice Change! I am curious to understand though -- did we see this as an issue, or did we think of this as an optimization? Of course, that is not a blocker to this change

s1monw: well I looked at this code accidentally investigating some idea and I figured it doesn't make sense to just block and wait instead of executing a search so if you are having a single segment you never pay the price of spawning a thread and you don't do any unnecessary work.

lucene/core/src/java/org/apache/lucene/search/IndexSearcher.java
```diff
@@ -636,16 +636,45 @@ public TopFieldDocs reduce(Collection<TopFieldCollector> collectors) throws IOEx
...
+
       for (int i = 0; i < leafSlices.length; ++i) {
         final LeafReaderContext[] leaves = leafSlices[i].leaves;
         final C collector = collectors.get(i);
-        topDocsFutures.add(executor.submit(new Callable<C>() {
-          @Override
-          public C call() throws Exception {
+        if (i == leafSlices.length-1) { // execute the last on the caller thread
+          search(Arrays.asList(leaves), weight, collector);
+          topDocsFutures.add(new Future<>() {
```
jpountz: use `CompletableFuture#completedFuture`?

lucene/core/src/java/org/apache/lucene/search/IndexSearcher.java
```diff
@@ -639,13 +639,15 @@ public TopFieldDocs reduce(Collection<TopFieldCollector> collectors) throws IOEx
       for (int i = 0; i < leafSlices.length; ++i) {
         final LeafReaderContext[] leaves = leafSlices[i].leaves;
         final C collector = collectors.get(i);
-        topDocsFutures.add(executor.submit(new Callable<C>() {
-          @Override
-          public C call() throws Exception {
+        if (i == leafSlices.length - 1) { // execute the last on the caller thread
```
mikemccand: Can we change `for` loop to only iterate up to `leafSlices.length - 1` and then do this afterwards?

uschindler: Yeah, would be better!

s1monw: I pushed new changes

PR #726, opened 2019-06-17
LUCENE-8632: New XYShape Field and Queries for indexing and searching general cartesian geometries

nknize:
The `LatLonShape` field and `LatLonShape` query classes added the ability to index and search geospatial geometries in the WGS-84 latitude, longitude coordinate reference system. The foundation for this capability is provided by the `Tessellator` that converts an array of vertices describing a `Point` `Line` or `Polygon` into a stream of 3 vertex triangles that are encoded as a seven dimension point and indexed using the BKD `POINT` structure. A nice property of the Tessellator is that `lat, lon` restrictions are artificial and really only bound by the API. 

This commit builds on top of / abstracts the `Tessellator`  `LatLonShape` and `LatLonShapeQuery` classes to provide the ability to index & search general cartesian (non WGS84 lat,lon restricted) geometry. It does so by introducing two new base classes: `ShapeField` and `ShapeQuery` that provide the indexing and search foundation for `LatLonShape` and the `LatLonShape` derived query classes (`LatLonShapeBoundingBoxQuery`, `LatLonShapeLineQuery`, `LatLonShapePolygonQuery`) and introducing a new `XYShape` factory class along with `XYShape` derived query classes (`XYShapeBoundingBoxQuery`, `XYShapeLineQuery`, `XYShapePolygonQuery`). The heart of the cartesian indexing is achieved through `XYShapeEncodingUtils` that converts the double precision vertices into an `integer` encoded seven dimension point (similar to LatLonShape).

The test framework is also further abstracted and extended to provide a full test suite for the new `XYShape` capability that works the same way as the `LatLonShape` test suite (but applied to non GIS geometries).


dsmiley:
This all sounds great; I like the use of the "XY" as a naming prefix to help clarify the coordinate system so we don't confuse it with LatLon.

In your mind... with the increase spatial code in core, are you still resolute in your opinion of Lucene-core having all this spatial code?  I mean... wow... it's not like Lucene core doesn't have a rich module system already.  Yeah I know this is kinda settled already but this just gnaws at me.

nknize:
With extensions like this (and possibly more to come) it certainly adds a lot to core. I'm not at all bound to the idea that `LatLonShape` **must** go in core. The `spatial` module is largely empty so I'm certainly open to relocating both `LatLonShape` and `XYShape` to the spatial module while keeping it dependency free and having only `LatLonPoint` in core.

dsmiley:
Yeah, lets do that then (another issue).  When the spatial stuff "graduates", as some say, out of sandbox.  This stuff here can stay in sandbox since it's where LatLonShape is now.

iverase:
Thanks Nick! 

I am having problems with Polygon and XYPolygon and  Line and XYLine and so on.  Are we keeping Polygon for backwards compatibility? It looks like it should be called LatLonPolygon for consistency.

nknize:
@iverase +1 for renaming `Polygon` and `Line` to `LatLonPolygon` and `LatLonLine` . It's cleaner that way.

lucene/sandbox/src/test/org/apache/lucene/document/TestXYShape.java
```diff
@@ -0,0 +1,89 @@
...
+
+  /** test we can search for a point with a standard number of vertices*/
+  public void testBasicIntersects() throws Exception {
+//    int numVertices = TestUtil.nextInt(random(), 50, 100);
+    Directory dir = newDirectory();
+    RandomIndexWriter writer = new RandomIndexWriter(random(), dir);
+
+    // add a random polygon document
+//    Polygon p = GeoTestUtil.createRegularPolygon(0, 90, atLeast(1000000), numVertices);
+//    XYPolygon xyp = new XYPolygon(p.getPolyLons(), p.getPolyLats(), XYShape.XYShapeType.INTEGER);
```
jpountz: is it a left over?

lucene/sandbox/src/java/org/apache/lucene/document/LatLonShape.java
```diff
@@ -54,14 +50,7 @@
...
  */
-public class LatLonShape {
-  static final int BYTES = Integer.BYTES;
-
-  protected static final FieldType TYPE = new FieldType();
-  static {
-    TYPE.setDimensions(7, 4, BYTES);
-    TYPE.freeze();
-  }
+public class LatLonShape extends ShapeField {
```
jpountz: do we actually need to extend it?

jpountz:
The renames make sense to me but maybe they should go into a separate change and only go to master since Polygon is a quite a high level user-facing API?

lucene/sandbox/src/test/org/apache/lucene/document/TestXYShape.java
nknize: Kind of... I went back and made this test slightly more randomized...

lucene/sandbox/src/java/org/apache/lucene/document/XYShape.java
```diff
@@ -0,0 +1,101 @@
...
+ * </ul>
+
+ * <b>WARNING</b>: Like {@link LatLonPoint}, vertex values are indexed with some loss of precision from the
+ * original {@code double} values.
+ * @see PointValues
+ * @see LatLonDocValuesField
+ *
+ * @lucene.experimental
+ */
+public class XYShape extends ShapeField {
```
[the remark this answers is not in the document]
nknize: We sure don't. Fixed.

lucene/sandbox/src/java/org/apache/lucene/document/LatLonShape.java
nknize: Not at all. Fixed.

nknize:
I went ahead and updated `XYPolygon`, `XYLine`, `XYPoint` API to take `float` instead of `double`. Internally we still use doubles for the tessellation but the user must pass floats to the Geometry so it's consistent with the encoding.

lucene/sandbox/src/java/org/apache/lucene/geo/Tessellator.java
```diff
@@ -119,17 +118,67 @@ private Tessellator() {}
...
+    // Calculate the tessellation using the doubly LinkedList.
+    List<Triangle> result = earcutLinkedList(polygon, outerNode, new ArrayList<>(), State.INIT, mortonOptimized);
+    if (result.size() == 0) {
+      throw new IllegalArgumentException("Unable to Tessellate shape [" + polygon + "]. Possible malformed shape detected.");
+    }
+
+    return result;
+  }
+
+  private static final Node createDoublyLinkedList(XYPolygon polygon, int startIndex, final WindingOrder windingOrder) {
```
iverase: We can remove this method and make the tessellate method call directly the other one in either case (Polygon and XYPolygon)

lucene/sandbox/src/java/org/apache/lucene/geo/XYRectangle2D.java
```diff
@@ -0,0 +1,252 @@
...
+import static java.lang.Integer.BYTES;
+import static org.apache.lucene.geo.GeoUtils.orient;
+import static org.apache.lucene.geo.XYEncodingUtils.decode;
+
+/**
+ * 2D rectangle implementation containing cartesian spatial logic.
+ *
+ * @lucene.internal
+ */
+public class XYRectangle2D {
```
iverase: There is a lot of common logic with Rectangle2D, would it be worthy to try to create a base class where all this common logic is?

lucene/sandbox/src/java/org/apache/lucene/geo/XYPolygon2D.java
```diff
@@ -0,0 +1,44 @@
...
+package org.apache.lucene.geo;
+
+/**
+ * 2D cartesian polygon implementation represented as a balanced interval tree of edges.
+ *
+ * @lucene.internal
+ */
+public class XYPolygon2D extends Polygon2D {
+
+  protected XYPolygon2D(XYPolygon polygon, XYPolygon2D holes) {
```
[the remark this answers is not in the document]
nknize: Yes, we can absolutely add this to `Polygon2D` but I think we'd only want to do that if we decide `XYShape` will live in `core`?  If we decide to put both `LatLonShape` and `XYShape` in the `spatial` module we'd have to subclass like this so core does not depend on classes in external modules.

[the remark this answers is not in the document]
iverase: I see, good call. 

I am gearing towards the idea that if we are going to have a spatial module, we should move everything there (even LatLonPoint). It is strange to have classes in two different places.

lucene/sandbox/src/java/org/apache/lucene/geo/XYRectangle2D.java
nknize: I think that's a good idea. This class is pretty much just for API purposed so I'll tease out the common logic and just have `Rectangle2D` derive what's common and implement the dateline crossing logic separately. Each constructor can handle their own encoding.

nknize: Updated `XYRectangle2D` to derive from `Rectangle2D` Worked out quite nicely and eliminated nearly all duplicate code.

iverase:
I am wondering if we can test the tessellator with XYPolygons as well. What I am thinking modified the current test and with the same polygons we have now, build them as XYPolygons and run the tessellation through them.

In a similar fashion we might want to test the encoding/decoding for XYShapes.

iverase:
Note to us: Currently XY shapes are under the geo package. I don't think that is correct and maybe we should add a different package (xy package?). I don't think we need it in this iteration but just writing it here so it is considered later on.

lucene/sandbox/src/java/org/apache/lucene/geo/Tessellator.java
```diff
@@ -142,6 +183,27 @@ private static final Node createDoublyLinkedList(final Polygon polygon, int star
...
+    XYPolygon[] holes = polygon.getHoles();
+    int nodeIndex = polygon.numPoints() ;
+    for(int i = 0; i < polygon.numHoles(); ++i) {
+      // create the doubly-linked hole list
+      Node list = createDoublyLinkedList(holes[i].getPolyX(), holes[i].getPolyY(), holes[i].getWindingOrder(), false, nodeIndex, WindingOrder.CCW);
+      // Determine if the resulting hole polygon was successful.
+      if(list != null) {
+        // Add the leftmost vertex of the hole.
+        holeList.add(fetchLeftmost(list));
+      }
```
iverase: You need to populate here `holeListPolygons` or you get an NPE downstream

lucene/sandbox/src/java/org/apache/lucene/geo/Tessellator.java
```diff
@@ -167,7 +228,10 @@ private static final Node eliminateHoles(final Polygon polygon, Node outerNode)
       }
       nodeIndex += holes[i].numPoints();
     }
+    return eliminateHoles(holeList, holeListPolygons, outerNode);
+  }
 
+  private static final Node eliminateHoles(List<Node> holeList, final Map holeListPolygons, Node outerNode) {
```
iverase: Map should have generics?

lucene/sandbox/src/java/org/apache/lucene/document/XYShape.java
```diff
@@ -0,0 +1,103 @@
...
+
+  /** create indexable fields for cartesian point geometry */
+  public static Field[] createIndexableFields(String fieldName, double x, double y) {
+    return new Field[] {new Triangle(fieldName,
+        encode(x), encode(y), encode(x), encode(y), encode(x), encode(y))};
+  }
+
+  /** create a query to find all cartesian shapes that intersect a defined bounding box **/
+  public static Query newBoxQuery(String field, QueryRelation queryRelation, double minX, double maxX, double minY, double maxY) {
+    return new XYShapeBoundingBoxQuery(field, queryRelation, minX, maxX, minY, maxY);
```
iverase: Should we have floats here instead?

lucene/sandbox/src/java/org/apache/lucene/document/XYShape.java
```diff
@@ -0,0 +1,103 @@
...
+          encode(line.getX(i)), encode(line.getY(i)),
+          encode(line.getX(j)), encode(line.getY(j)),
+          encode(line.getX(i)), encode(line.getY(i)));
+    }
+    return fields;
+  }
+
+  /** create indexable fields for cartesian point geometry */
+  public static Field[] createIndexableFields(String fieldName, double x, double y) {
+    return new Field[] {new Triangle(fieldName,
```
iverase: Should we have floats here instead?

lucene/sandbox/src/java/org/apache/lucene/document/ShapeField.java
```diff
@@ -0,0 +1,303 @@
...
+ * Finding all shapes that intersect a range (e.g., bounding box), or target shape, at search time is efficient.
+ * <p>
+ * This class defines the static methods for encoding the three vertices of a tessellated triangles as a seven dimension point.
+ * The coordinates are converted from double precision values into 32 bit integers so they are sortable at index time.
+ * <p>
+ *
+ * @lucene.experimental
+ */
+public abstract class ShapeField {
+  /** vertex coordinates are encoded as 4 byte integers */
```
iverase: This class is abstract but not implemented by any other.

lucene/sandbox/src/java/org/apache/lucene/geo/XYPolygon2D.java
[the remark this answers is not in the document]
dsmiley: I hear ya @iverase ... I'm glad to hear I'm not the only crazy one :-)

nknize:
Cleaned up the API a bit:

* refactors `Tessellator.Triangle` `getLat / getLon` methods to `getY / getX`, respectively
* Changed `XYShape.createIndexableFields` and `XYShape.newBoxQuery` to accept floats instead of doubles

Refactored `TestLatLonShapeEncoding` to derive from a new `BaseShapeEncodingTestCase` class along with a new `TestXYShapeEncoding` class to equally test the `XYShapeEncoding` logic.

I think this PR is just about ready to merge and continue iterating in sandbox?

nknize:
Closing: Merged in commit 0c09481374cab029d57b0f9b45994822c0dcd39b

PR #728, opened 2019-06-18
LUCENE-8867: Store point with cardinality for low cardinality leaves

iverase:
This pull request adds a new strategy for storing leaves with low cardinality and a new method to the IntersectsVisitor that can accept an array of docs. This can speed up searches when there are many equal points.


msokolov:
We discussed this today - thanks for the change! I'm not familiar with this area of the code, and not really qualified to assess it, but I did notice there are no tests with the patch. I guess we are supporting existing functionality, but are we sure that the existing tests are covering the different cardinalities?

iverase:
@msokolov I check that the new code is being hit in the tests so yes, it is covered.

iverase:
This PR has been splitter into #730 and #746 so I will close it.

PR #729, opened 2019-06-19
LUCENE-8862: Introduce Collector Level Memory Accounting

atris:
This commit introduces a mechanism to track the dynamic memory
utilization of Collectors and allow setting collector level limits
for memory usage. A new bitset collecting Collector is introduced
which supports the memory tracker functionality.


atris:
I wonder if we should add (maybe as a separate PR) a wrapper collector which allows an uber memory limit for all collectors combined (a similar CollectorManager too?)

lucene/core/src/java/org/apache/lucene/util/CollectorMemoryTracker.java
```diff
@@ -0,0 +1,53 @@
...
+    this.memoryUsage = new AtomicLong();
+    this.memoryLimit = memoryLimit;
+  }
+
+  @Override
+  synchronized public void updateBytes(long bytes) {
+    long currentMemoryUsage = memoryUsage.addAndGet(bytes);
+
+    if (currentMemoryUsage > memoryLimit) {
+      throw new IllegalStateException("Memory limit exceeded for " + name);
```
atris: Should this be CollectionEarlyTerminated?

atris:
Renamed DynamicMemoryTracker to MemoryTracker and removed redundant synchronized keyword

atris:
Any thoughts on this? It seems the least invasive way to add a useful functionality that can be used by many upstream users (ES and Solr?)

lucene/core/src/java/org/apache/lucene/search/MemoryAccountingBitsetCollector.java
```diff
@@ -0,0 +1,63 @@
...
+  FixedBitSet bitSet = new FixedBitSet(0);
+  int length = 0;
+  int docBase = 0;
+
+  MemoryAccountingBitsetCollector(CollectorMemoryTracker tracker) {
+    this.tracker = tracker;
+    tracker.updateBytes(bitSet.ramBytesUsed());
+  }
+
+  BitSet getBits() {
```
jpountz: this too?

lucene/core/src/java/org/apache/lucene/search/MemoryAccountingBitsetCollector.java
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  @Override
+  public void collect(int doc) {
+    bitSet.set(docBase + doc);
+  }
+
+  @Override
+  public ScoreMode scoreMode() {
+    return ScoreMode.COMPLETE;
```
jpountz: this should be COMPLETE_NO_SCORES

atris: Fixed, thanks

lucene/core/src/java/org/apache/lucene/search/MemoryAccountingBitsetCollector.java
atris: Not sure if I got your comment. This was redundant, hence removed

atris:
@jpountz  Thanks, moved to misc as suggested

lucene/core/src/java/org/apache/lucene/search/MemoryAccountingBitsetCollector.java
```diff
@@ -0,0 +1,58 @@
...
+
+/** Bitset collector which supports memory tracking */
+public class MemoryAccountingBitsetCollector extends SimpleCollector {
+
+  final CollectorMemoryTracker tracker;
+  FixedBitSet bitSet = new FixedBitSet(0);
+  int length = 0;
+  int docBase = 0;
+
+  MemoryAccountingBitsetCollector(CollectorMemoryTracker tracker) {
```
jpountz: do we expect users to call this constructor themselves? If yet it should be public otherwise they won't be able to create it?

jpountz:
I was thinking of the lucene/misc module rather than the misc package.

atris:
@jpountz Thanks for clarifying. I have moved both the collector and the tracker to misc module.

lucene/core/src/java/org/apache/lucene/util/MemoryTracker.java
```diff
@@ -0,0 +1,26 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.lucene.util;
+
+/**
+ * Tracks dynamic allocations/deallocations of memory for transient objects
+ */
+public interface MemoryTracker {
```
jpountz: can you move this one as well to misc?

atris:
@jpountz  Sorry, somehow missed moving the actual interface. Done now, let me know if it seems fine.

atris:
Thanks a ton!

jpountz:
`ant precommit` fails for me with this change, it needs a `package.html` file in `lucene/misc/src/java/org/apache/lucene/util`.

```
-documentation-lint:
     [echo] checking for broken html...
    [jtidy] FIXME: Broken HTML checks were disabled, as jtidy can't handle HTML5.
     [echo] Checking for broken links...
     [exec] 
     [exec] Crawl/parse...
     [exec] 
     [exec] Verify...
     [echo] Checking for missing docs...
     [exec] 
     [exec] build/docs/misc/org/apache/lucene/util/package-summary.html
     [exec]   no package description (missing package.html in src?)
     [exec] 
     [exec] Missing javadocs were found!
```

atris:
@jpountz Added the same, please see if it works now

PR #730, opened 2019-06-19
LUCENE-8868: New storing strategy for BKD tree leaves with low cardinality

iverase:
# Description

Currently if a leaf on the BKD tree contains only few values, then the leaf is treated the same way as it all values are different. It many cases it can be much more efficient to store the distinct values with the cardinality.

# Solution

The strategy is the following:
1. When writing a leaf block the cardinality is computed.
2. Perform some naive calculation to compute if it is better to store the leaf as a low cardinality leaf. The storage cost are calculated as follows:
low cardinality: leafCardinality * (packedBytesLength - prefixLenSum + 2) where two is the estimated size of storing the cardinality. This is an overestimation as in some cases you will only need one byte to store the cardinality.
High cardinality: count * (packedBytesLength - prefixLenSum). We are not taking into account the runlen compression.
3. If the tree has low cardinality then we set the compressed dim to -2. Note that -1 is when all values are equal.

# Tests

BKD tree is extensively tested already so there is no need to add new tests.


lucene/core/src/java/org/apache/lucene/util/bkd/BKDWriter.java
```diff
@@ -1028,17 +1035,43 @@ private void writeLeafBlockDocs(DataOutput out, int[] docIDs, int start, int cou
...
   }
 
-  private void writeLeafBlockPackedValues(DataOutput out, int[] commonPrefixLengths, int count, int sortedDim, IntFunction<BytesRef> packedValues) throws IOException {
+  private void writeLeafBlockPackedValues(DataOutput out, int[] commonPrefixLengths, int count, int sortedDim, IntFunction<BytesRef> packedValues, int leafCardinality) throws IOException {
     int prefixLenSum = Arrays.stream(commonPrefixLengths).sum();
     if (prefixLenSum == packedBytesLength) {
       // all values in this block are equal
       out.writeByte((byte) -1);
-    } else {
+    } else if (leafCardinality * (packedBytesLength - prefixLenSum + 2)  <= count * (packedBytesLength - prefixLenSum)) {
```
jpountz: Am I reading it right that you are counting 2 for the vint? I think you could make it 1 instead, the reasoning being that if you vints are 2 bytes on average, then it means than your runs are very long (vint start using 2 bytes when they are greater than 127) and so the sparse encoding is an obvious win.

lucene/core/src/java/org/apache/lucene/util/bkd/BKDReader.java
```diff
@@ -441,8 +441,15 @@ int readDocIDs(IndexInput in, long blockFP, int[] docIDs) throws IOException {
 
   void visitDocValues(int[] commonPrefixLengths, byte[] scratchDataPackedValue, byte[] scratchMinIndexPackedValue, byte[] scratchMaxIndexPackedValue,
                       IndexInput in, int[] docIDs, int count, IntersectVisitor visitor) throws IOException {
+    if (version >= BKDWriter.VERSION_LOW_CARDINALITY_LEAVES) {
+      visitDocValuesWithCardinality(commonPrefixLengths, scratchDataPackedValue, scratchMinIndexPackedValue, scratchMaxIndexPackedValue, in, docIDs, count, visitor);
+    } else {
+      visitDocValuesNoCardinality(commonPrefixLengths, scratchDataPackedValue, scratchMinIndexPackedValue, scratchMaxIndexPackedValue, in, docIDs, count, visitor);
+    }
```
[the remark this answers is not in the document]
iverase: Actually we cannot because in the previous version of the BKD tree the compressed dim byte was written after the leaf bounds. This is the reason I have to fork the version here.

lucene/core/src/java/org/apache/lucene/util/bkd/BKDWriter.java
iverase: Yes, I was too conservative .

iverase:
I had a look in how good is the formula to decide to use this optimisation and results are very interesting. I have only done it for 1D so far but it seems we are underestimating the compression in 1 dimension so the result is a bigger index.

The test has been done randomly ingesting 10M intPoint, in each iteration the cardinality has been  increased. The size of the index has been calculated after indexing the data single threaded and after force merge into one segment.

- Considering the size of vint of 1 Byte results are pretty bad for some cardinalities.  

```
leafCardinality * (packedBytesLength - prefixLenSum + 1)  <= count * (packedBytesLength - prefixLenSum)
```

image


- Considering the size of vint of 2 Byte results improve but still there is a region were index becomes bigger.

```
leafCardinality * (packedBytesLength - prefixLenSum + 2)  <= count * (packedBytesLength - prefixLenSum)
```

image
  
 - Considering the size of vint of 3 Byte results improve but still there is a region were index becomes bigger.

```
leafCardinality * (packedBytesLength - prefixLenSum + 3)  <= count * (packedBytesLength - prefixLenSum)
```

image

iverase:
I think the calculation was incorrect, we need to consider the runLen compression, which basically means that if totally efficient there is one less byte on the high cardinality side. Therefore the formula looks like:

```
leafCardinality * (packedBytesLength - prefixLenSum + 1)  <= count * (packedBytesLength - prefixLenSum - 1)
```

With these, plot look fine:

image

iverase:
I had another iteration, main things are:

1. Now the cost of runLen compression is computed exact.

2.  Cardinality is computed after the leaf has been sorted by the elected dimension. This computation does not copy arrays around.

iverase:
> I wonder whether we should apply the run-length compression on the sorted dimension in the 
> low-cardinality case as well, this could save some additional bytes, and might make the logic to 
> decide whether to use the low-cardinality or high-cardinality encoding a bit easier?

I don't think this would be so beneficial in the case of very low cardinality. Imagine that those few values only differ in the last Byte of the sorted dimension. For each value you add at least two bytes for the runLen compression and just save one when writing the sorted dimension.  All in all the final size of the leaf will be bigger and I think this approach should favour this case.

jpountz:
Fair enough.

> BKD tree is extensively tested already so there is no need to add new tests.

Do we already have tests for the case that leaves have lots of duplicates?

iverase:
Tests trigger the new approachbut it is true not as often as they should. I have added a new test that exercise the new code more often.

PR #731, opened 2019-06-19
LUCENE-8865: Move to executor in IndexSearcher

s1monw:
In order to simplify testing this change moves to use the Executor
interface instead of ExecutorService. This change also simplifies
customizing execute methods for use-cases that need to add additional
logic for forking to new threads. This change also adds a test for
the optimization added in LUCENE-8865.

This change is fully backwards compatible since ExecutorService implements
Executor.

### DIFF
--- a/lucene/core/src/test/org/apache/lucene/search/TestIndexSearcher.java
+++ b/lucene/core/src/test/org/apache/lucene/search/TestIndexSearcher.java
@@ -18,11 +18,14 @@ package org.apache.lucene.search;
 
 
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.List;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import org.apache.lucene.document.Document;
 import org.apache.lucene.document.Field.Store;
@@ -30,6 +33,7 @@ import org.apache.lucene.document.Field;
 import org.apache.lucene.document.SortedDocValuesField;
 import org.apache.lucene.document.StringField;
 import org.apache.lucene.index.IndexReader;
+import org.apache.lucene.index.LeafReaderContext;
 import org.apache.lucene.index.MultiReader;
 import org.apache.lucene.index.RandomIndexWriter;
 import org.apache.lucene.index.Term;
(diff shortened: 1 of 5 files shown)


s1monw:
@mikemccand  can you take a look

lucene/core/src/test/org/apache/lucene/search/TestIndexSearcher.java
```diff
@@ -237,4 +241,28 @@ public void testGetSlices() throws Exception {
...
+      protected LeafSlice[] slices(List<LeafReaderContext> leaves) {
+        ArrayList<LeafSlice> slices = new ArrayList<>();
+        for (LeafReaderContext ctx : leaves) {
+          slices.add(new LeafSlice(Arrays.asList(ctx)));
+        }
+        return slices.toArray(new LeafSlice[0]);
+      }
+    };
+    searcher.search(new MatchAllDocsQuery(), 10);
+    if (leaves.size() <= 1) {
```
atris: Should this be just == 1?

lucene/core/src/test/org/apache/lucene/search/TestIndexSearcher.java
```diff
@@ -237,4 +241,28 @@ public void testGetSlices() throws Exception {
...
+      assertEquals(0, numExecutions.get());
+    } else {
+      assertEquals(leaves.size()-1, numExecutions.get());
```
atris: Nit: Spacing

lucene/core/src/test/org/apache/lucene/search/TestIndexSearcher.java
```diff
@@ -237,4 +241,28 @@ public void testGetSlices() throws Exception {
     service.shutdown();
     IOUtils.close(r, dir);
   }
+
+  public void testOneSegmentExecutesOnTheCallerThread() throws IOException {
```
atris: Should we have more random testing, where we test different combination of slices? Maybe have a document distribution with random commit calls to create different segment distributions?

lucene/core/src/test/org/apache/lucene/search/TestIndexSearcher.java
s1monw: it's in theory possible to have 0 leaves. That's why I did that

lucene/core/src/test/org/apache/lucene/search/TestIndexSearcher.java
s1monw: RandomIndexWriter is used in this test which does that.

msokolov:
+1

atris:
+1, Looks good

PR #734, opened 2019-06-20
LUCENE-8857: Introduce Custom Tiebreakers in TopDocs#merge

atris:
This commit introduces custom tiebreakers which allows users to
specify custom tiebreakers when ordering hits to return. A
default tiebreaker is introduced for tie breaking on shard index
first and then docID.

### DIFF
--- a/lucene/MIGRATE.txt
+++ b/lucene/MIGRATE.txt
@@ -182,3 +182,9 @@ to concurrent changes.
 ## maxClausesCount moved from BooleanQuery To IndexSearcher (LUCENE-8811) ##
 IndexSearcher now performs max clause count checks on all types of queries (including BooleanQueries).
 This led to a logical move of the clauses count from BooleanQuery to IndexSearcher.
+
+## TopDocs.merge shall no longer allow setting of shard indices ##
+
+TopDocs.merge's API has been changed to stop allowing passing in a parameter to indicate if it should
+set shard indices for hits as they are seen during the merge process. This is done to simplify the API
+to be more dynamic in terms of passing in custom tie breakers
--- a/lucene/MIGRATE.txt
+++ b/lucene/MIGRATE.txt
@@ -187,4 +187,6 @@ This led to a logical move of the clauses count from BooleanQuery to IndexSearch
 
 TopDocs.merge's API has been changed to stop allowing passing in a parameter to indicate if it should
 set shard indices for hits as they are seen during the merge process. This is done to simplify the API
-to be more dynamic in terms of passing in custom tie breakers
+to be more dynamic in terms of passing in custom tie breakers.
+If shard indices are to be used for tie breaking docs with equal scores during TopDocs.merge, then it is
+mandatory that the input ScoreDocs have their shard indices set to valid values prior to calling TopDocs.merge
--- a/lucene/core/src/java/org/apache/lucene/search/TopDocs.java
+++ b/lucene/core/src/java/org/apache/lucene/search/TopDocs.java
@@ -16,6 +16,8 @@
  */
 package org.apache.lucene.search;
 
+import java.util.Comparator;
+
 import org.apache.lucene.util.PriorityQueue;
 
 /** Represents hits returned by {@link
@@ -28,6 +30,25 @@ public class TopDocs {
   /** The top hits for the query. */
   public ScoreDoc[] scoreDocs;
 
+  /** Internal comparator with shardIndex */
+  private static final Comparator<ScoreDoc> shardIndexTieBreaker = Comparator.comparingInt(d -> d.shardIndex);
+
+  /** Internal comparator with docID */
+  private static final Comparator<ScoreDoc> docIDTieBreaker = Comparator.comparingInt(d -> d.doc);
+
+  /** Default comparator */
+  private static final Comparator<ScoreDoc> defaultTieBreaker = new Comparator<ScoreDoc>() {
+    @Override
+    public int compare(ScoreDoc o1, ScoreDoc o2) {
+      int shardResultVal = shardIndexTieBreaker.compare(o1, o2);
+      if (shardResultVal != 0) {
+        return shardResultVal;
+      }
+
+      return docIDTieBreaker.compare(o1, o2);
+    }
+  };
+
   /** Constructs a TopDocs. */
   public TopDocs(TotalHits totalHits, ScoreDoc[] scoreDocs) {
     this.totalHits = totalHits;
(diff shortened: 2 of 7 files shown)


atris:
Does this seem fine? I ran the test suite a couple of times and nothing seems to be breaking. This also fixes the existing bug in TopDocs.merge

lucene/core/src/java/org/apache/lucene/search/TopDocs.java
```diff
@@ -28,6 +30,25 @@
...
   public ScoreDoc[] scoreDocs;
 
+  /** Internal comparator with shardIndex */
+  private static final Comparator<ScoreDoc> shardIndexTieBreaker = Comparator.comparingInt(d -> d.shardIndex);
+
+  /** Internal comparator with docID */
+  private static final Comparator<ScoreDoc> docIDTieBreaker = Comparator.comparingInt(d -> d.doc);
+
+  /** Default comparator */
+  private static final Comparator<ScoreDoc> defaultTieBreaker = new Comparator<ScoreDoc>() {
```
jpountz: this could be `shardIndexTieBreaker.thenComparing(docIDTieBreaker)`

lucene/core/src/java/org/apache/lucene/search/TopDocs.java
```diff
@@ -28,6 +30,25 @@
   /** The top hits for the query. */
   public ScoreDoc[] scoreDocs;
 
+  /** Internal comparator with shardIndex */
+  private static final Comparator<ScoreDoc> shardIndexTieBreaker = Comparator.comparingInt(d -> d.shardIndex);
```
jpountz: can you use upper case for constants?

lucene/core/src/java/org/apache/lucene/search/TopDocs.java
```diff
@@ -39,66 +60,52 @@ public TopDocs(TotalHits totalHits, ScoreDoc[] scoreDocs) {
...
-        return shardIndex;
-      }
-    }
   }
 
   /**
-   * if we need to tie-break since score / sort value are the same we first compare shard index (lower shard wins)
-   * and then iff shard index is the same we use the hit index.
+   * Use the tie breaker if provided. If tie breaker returns 0 signifying equal values, we use shard indices
+   * to tie break intra shard ties
```
jpountz: do you mean hit indices rather than shard indices?

lucene/core/src/java/org/apache/lucene/search/TopDocs.java
```diff
@@ -260,11 +291,12 @@ private static TopDocs mergeAux(Sort sort, int start, int size, TopDocs[] shardH
       }
       if (shard.scoreDocs != null && shard.scoreDocs.length > 0) {
         availHitCount += shard.scoreDocs.length;
-        queue.add(new ShardRef(shardIDX, setShardIndex == false));
+        queue.add(new ShardRef(shardIDX));
       }
     }
 
     final ScoreDoc[] hits;
+    final boolean unsetShardIndex = shardHits[0].scoreDocs[0].shardIndex == -1;
```
jpountz: this is prone to out-of-bounds exceptions, you should find the first TopDocs instance that has a non-empty ScoreDoc[] if any

lucene/core/src/java/org/apache/lucene/search/TopDocs.java
```diff
@@ -276,11 +308,12 @@ private static TopDocs mergeAux(Sort sort, int start, int size, TopDocs[] shardH
...
-        if (setShardIndex) {
-          // caller asked us to record shardIndex (index of the TopDocs array) this hit is coming from:
-          hit.shardIndex = ref.shardIndex;
-        } else if (hit.shardIndex == -1) {
-          throw new IllegalArgumentException("setShardIndex is false but TopDocs[" + ref.shardIndex + "].scoreDocs[" + (ref.hitIndex-1) + "] is not set");
+
+        //Irrespective of whether we use shard indices for tie breaking or not, we check for consistent
+        // order in shard indices to defend against potential bugs
+        if ((unsetShardIndex && hit.shardIndex != -1) ||
+            (unsetShardIndex == false && hit.shardIndex == -1)) {
```
jpountz: we could combine both into `unsetShardIndex != (hit.shardIndex == -1)` ?

lucene/core/src/java/org/apache/lucene/search/TopDocs.java
```diff
@@ -276,11 +308,12 @@ private static TopDocs mergeAux(Sort sort, int start, int size, TopDocs[] shardH
         assert queue.size() > 0;
         ShardRef ref = queue.top();
         final ScoreDoc hit = shardHits[ref.shardIndex].scoreDocs[ref.hitIndex++];
-        if (setShardIndex) {
-          // caller asked us to record shardIndex (index of the TopDocs array) this hit is coming from:
-          hit.shardIndex = ref.shardIndex;
-        } else if (hit.shardIndex == -1) {
-          throw new IllegalArgumentException("setShardIndex is false but TopDocs[" + ref.shardIndex + "].scoreDocs[" + (ref.hitIndex-1) + "] is not set");
+
+        //Irrespective of whether we use shard indices for tie breaking or not, we check for consistent
```
jpountz: can you leave a space between the slashes and the beginning of the comment?

lucene/core/src/test/org/apache/lucene/search/TestMultiSliceMerge.java
```diff
@@ -0,0 +1,148 @@
...
+    dir2.close();
+  }
+
+  public void testMultipleSlicesOfSameIndexSearcher() throws Exception {
+    ExecutorService service1 = new ThreadPoolExecutor(4, 4, 0L, TimeUnit.MILLISECONDS,
+        new LinkedBlockingQueue<Runnable>(),
+        new NamedThreadFactory("TestMultiSliceMerge"));
+    ExecutorService service2 = new ThreadPoolExecutor(4, 4, 0L, TimeUnit.MILLISECONDS,
+        new LinkedBlockingQueue<Runnable>(),
+        new NamedThreadFactory("TestMultiSliceMerge"));
```
jpountz: threading is not really relevant to this test, could we do something like `Executor executor = runnable -> runnable.run();` instead of creating an actual thread pool?

lucene/core/src/test/org/apache/lucene/search/TestMultiSliceMerge.java
```diff
@@ -0,0 +1,148 @@
...
+  IndexReader reader1;
+  IndexReader reader2;
+
+  @Override
+  public void setUp() throws Exception {
+    super.setUp();
+    dir1 = newDirectory();
+    dir2 = newDirectory();
+    Random random = random();
+    RandomIndexWriter iw1 = new RandomIndexWriter(random(), dir1);
```
jpountz: If you rely on the order of segments, you will need to configure a LogMergePolicy (see usage of LuceneTestCase#newLogMergePolicy)

lucene/core/src/test/org/apache/lucene/search/TestMultiSliceMerge.java
```diff
@@ -0,0 +1,148 @@
...
+    for (int i = 0; i < 100; i++) {
+      Document doc = new Document();
+      doc.add(newStringField("field", Integer.toString(i), Field.Store.NO));
+      doc.add(newStringField("field2", Boolean.toString(i % 2 == 0), Field.Store.NO));
+      doc.add(new SortedDocValuesField("field2", new BytesRef(Boolean.toString(i % 2 == 0))));
+      iw1.addDocument(doc);
+
+      if (random.nextBoolean()) {
+        iw1.commit();
```
jpountz: you could do iw1.getReader().close(), which writes a segment as well but saves a fsync

atris:
@jpountz Thanks for your comments, I have fixed the same.

I think we should push this to master as you advised. Concurrently, should I raise a PR to add the old slicing algorithm back and call it under a version check?

s1monw:
> I think we should push this to master as you advised. Concurrently, should I raise a PR to add the old slicing algorithm back and call it under a version check?

We should just revert the slicing change in 8.x IMO for now and keep it in master. I can do that after this is merged. I think we need a changes entry for this one too. LGTM otherwise.

atris:
Thanks @s1monw !

@jpountz Would you want to add the CHANGES entry during merging, or should I push an iteration?

jpountz:
If you can add it now that would be great.

lucene/MIGRATE.txt
```diff
@@ -182,3 +182,9 @@ to concurrent changes.
 ## maxClausesCount moved from BooleanQuery To IndexSearcher (LUCENE-8811) ##
 IndexSearcher now performs max clause count checks on all types of queries (including BooleanQueries).
 This led to a logical move of the clauses count from BooleanQuery to IndexSearcher.
+
+## TopDocs.merge shall no longer allow setting of shard indices ##
+
+TopDocs.merge's API has been changed to stop allowing passing in a parameter to indicate if it should
+set shard indices for hits as they are seen during the merge process. This is done to simplify the API
+to be more dynamic in terms of passing in custom tie breakers
```
jpountz: Maybe say explicitly that if shard indices should be used for tie-breaking then they should be set on the ScoreDoc objects before calling TopDocs#merge?

lucene/core/src/java/org/apache/lucene/search/TopDocs.java
```diff
@@ -39,66 +50,52 @@ public TopDocs(TotalHits totalHits, ScoreDoc[] scoreDocs) {
...
+
+    if (value == 0) {
+      // Equal Values
       // Tie break in same shard: resolve however the
       // shard had resolved it:
       assert first.hitIndex != second.hitIndex;
       return first.hitIndex < second.hitIndex;
     }
+
+    return value < 0 ? true : false;
```
jpountz: should be just `return value < 0;`

atris:
@jpountz I have updated per comments and added CHANGES entry, let me know if this looks ok

jpountz:
I tried to merge but I'm hitting a reproducible test failure: `ant test  -Dtestcase=TestShardSearching -Dtests.method=testSimple -Dtests.seed=A85246DA527C274 -Dtests.slow=true -Dtests.badapples=true -Dtests.locale=ar-LB -Dtests.timezone=America/Mexico_City -Dtests.asserts=true -Dtests.file.encoding=UTF-8`.

lucene/core/src/java/org/apache/lucene/search/TopDocs.java
```diff
@@ -276,12 +298,16 @@ private static TopDocs mergeAux(Sort sort, int start, int size, TopDocs[] shardH
...
+
+        // Irrespective of whether we use shard indices for tie breaking or not, we check for consistent
+        // order in shard indices to defend against potential bugs
+        if (hitUpto > 0) {
+          if (unsetShardIndex != (hit.shardIndex == -1)) {
+            throw new IllegalArgumentException("Inconsistent order of shard indices");
+          }
+      }
+
+      unsetShardIndex |= hit.shardIndex == -1;
```
jpountz: it looks like there is an indentation issue on the above 3 lines?

atris:
@jpountz Thanks for pointing it out. That was due to the test assuming that TopDocs.merge will set the shard indices and depending on it. I have refactored the test to set its own shard indices, and the test passes for me. Could you please check on your end and let me know?

jpountz:
Would you mind running the whole test suite to make sure there are no other tests that fail with this change?

atris:
@jpountz  Ran the same, it passed. Fixed the indent as well.

Could not run ant precommit due to some unneeded files in my directory, but ran ant test.

msokolov:
You really should be running precommit; otherwise you are just asking
someone else to do it for you. Is there some reason you couldn't you remove
the unneeded files to get precommit to run?If not (you have notes you don't
want to delete, etc), you can always create another clean working directory
with which to run tests, precommit, etc.

atris:
@msokolov I agree, ant precommit is a must for PRs (have been doing it for my PRs prior to the last couple). The reason I am not able to run precommit is because of some muddle in my test machine (which is really an EC2 VM). I am imaging another one, just that did not want my hardware issues to hang up the progress of this PR, so called out my inability to run ant precommit.

I should be able to run precommits by end of day tomorrow, so further PRs will again have that base covered. Thanks for pointing it out.

msokolov:
I see - sorry your machine is muddled!

PR #736, opened 2019-06-21
SOLR-13532: Fix for non-recovering cores due to low timeouts

surilshah:
15000 timeout is much better than 1000, since 1 second is too low for most clusters


cpoerschke:
Hello. https://issues.apache.org/jira/browse/SOLR-13532 is closed now, could this PR be closed too then manually? Sometimes pull requests get auto-closed by the bot when it 'sees' the right phrase in the commit messages but looks like in this case the trigger phrase was not used.

surilshah:
Closing this PR as fix for https://issues.apache.org/jira/browse/SOLR-13532 has been submitted by @hossman

PR #742, opened 2019-06-26
LUCENE-8882: Extend QueryVisitor To Maintain Metadata State

atris:
This commit introduces the notion of state to QueryVisitor API, thus
allowing queries to pass metadata across the entire query tree. This
can allow things like upper queries making intelligent decisions
for cases like IndexOrDocValuesQuery, or cases where a sorted
index can benefit from a sorted index query.


atris:
Any thoughts on this? Happy to iterate

PR #744, opened 2019-06-26
LUCENE-8856: Promote intervals queries from sandbox to queries module

romseygeek:


romseygeek:
One thing I haven't done is add a shim from the old module to the new, which would be simple enough but would require a dependency from sandbox to queries, which we've tried to avoid in the past.

jpountz:
@romseygeek Looks like this has been merged already, can we close?

PR #746, opened 2019-06-27
LUCENE-8885: Optimise BKD reader by exploiting cardinality information stored on leaves

iverase:
A `DocIdSetIterator` is added to the BKD reader and use when we have cardinality information.


lucene/core/src/java/org/apache/lucene/util/bkd/BKDReader.java
```diff
@@ -780,4 +783,43 @@ public int getDocCount() {
...
+      if (idx == length - 1) {
+        return DocIdSetIterator.NO_MORE_DOCS;
+      }
+      idx++;
+      return docIds[offset + idx];
+    }
+
+    @Override
+    public int advance(int target) throws IOException {
+      throw new UnsupportedOperationException();
```
jpountz: I think it'd be better to support it, even if slow. Eg. you can return `slowAdvance(target)`.

lucene/core/src/java/org/apache/lucene/util/bkd/BKDReader.java
```diff
@@ -780,4 +783,43 @@ public int getDocCount() {
...
+    @Override
+    public int advance(int target) throws IOException {
+      return slowAdvance(target);
+    }
+
+    @Override
+    public long cost() {
+      return length;
+    }
+  }
```
jpountz: can you add tests for it? e.g. making sure `docID()` actually returns -1 when unpositioned and NO_MORE_DOCS when exhausted (I think the latter is not correct at the moment)

iverase: You are totally right and indeed we should be testing it. I added the test in the verify method. I call randomly the default implementation or check manually the answer from the iterator.

I changed the iterator to hold the current docID so I do not need to check the index of the array all the time.

lucene/core/src/java/org/apache/lucene/util/bkd/BKDReader.java
```diff
@@ -780,4 +779,51 @@ public int getDocCount() {
...
     return nodeID >= leafNodeOffset;
   }
+
+  protected static class BKDReaderDocIDSetIterator extends DocIdSetIterator {
+
+    private int idx;
+    private int length;
+    private int offset;
+    private int docID;
+    int[] docIDs;
```
jpountz: make it final?

lucene/core/src/test/org/apache/lucene/util/bkd/TestBKD.java
```diff
@@ -844,7 +845,28 @@ public void visit(int docID, byte[] packedValue) {
...
             }
 
-            @Override
+          @Override
+          public void visit(DocIdSetIterator iterator, byte[] packedValue) throws IOException {
+              if (random().nextBoolean()) {
+                //check the default method is correct
+                IntersectVisitor.super.visit(iterator, packedValue);
+              } else {
+                assertTrue(iterator.docID() == -1);
```
jpountz: Can you use assertEquals so that we get the current doc ID according to the iterator in case of failure?

lucene/core/src/test/org/apache/lucene/util/bkd/TestBKD.java
```diff
@@ -844,7 +845,28 @@ public void visit(int docID, byte[] packedValue) {
...
+                int cost = (int) iterator.cost();
```
jpountz: being paranoid: can you use Math#toIntExact?

lucene/core/src/test/org/apache/lucene/util/bkd/TestBKD.java
```diff
@@ -844,7 +845,28 @@ public void visit(int docID, byte[] packedValue) {
...
+                int numberOfPoints = 0;
+                int docID;
+                while ((docID = iterator.nextDoc()) != DocIdSetIterator.NO_MORE_DOCS) {
+                  visit(docID, packedValue);
+                  numberOfPoints++;
+                }
+                assertTrue(cost == numberOfPoints);
+                assertTrue(iterator.docID() == DocIdSetIterator.NO_MORE_DOCS);
+                assertTrue(iterator.nextDoc() == DocIdSetIterator.NO_MORE_DOCS);
+                assertTrue(iterator.docID() == DocIdSetIterator.NO_MORE_DOCS);
```
jpountz: can you use assertEquals here as well?

lucene/core/src/test/org/apache/lucene/util/bkd/TestBKD.java
```diff
@@ -844,7 +845,28 @@ public void visit(int docID, byte[] packedValue) {
...
+              if (random().nextBoolean()) {
+                //check the default method is correct
+                IntersectVisitor.super.visit(iterator, packedValue);
+              } else {
+                assertTrue(iterator.docID() == -1);
+                int cost = (int) iterator.cost();
+                int numberOfPoints = 0;
+                int docID;
+                while ((docID = iterator.nextDoc()) != DocIdSetIterator.NO_MORE_DOCS) {
+                  visit(docID, packedValue);
```
jpountz: let's assert on the value of docID() here as well?

PR #747, opened 2019-06-27
LUCENE-8888: Improve distribution of points with data dimension in BKD tree leaves

iverase:
Sorting/selection algorithms on the BKD tree are updated to tie-break on the data dimensions when they exist.


lucene/core/src/java/org/apache/lucene/util/bkd/BKDRadixSelector.java
```diff
@@ -60,14 +60,18 @@
   // prefix for temp files
   private final String tempFileNamePrefix;
 
+  private  int numDatadim, numIndexDim;
```
jpountz: s/numDatadim/numDataDim/

iverase:
Thanks @jpountz, I can indeed refactor PointValue and it seems it makes a lot of sense. It simplifies quite a lot the logic of the offline selector.

iverase:
> Do you think we could go even further and use a similar idea in HeapPointWriter instead of storing doc IDs in a separate int[]? (I haven't checked for potential downsides.)

 I do like the idea but there is an issue because when writing the docIds for a leaf, the method that is writing that info to the index (writeDocIds) is expecting an int[]. We would need to change all that logic in order to implement this.

iverase:
Ok, so finally I modify the HeapPointWriter/Reader to hold all the bytes in one array.

lucene/core/src/java/org/apache/lucene/util/bkd/BKDRadixSelector.java
```diff
@@ -324,23 +337,22 @@ protected int byteAt(int i, int k) {
...
       @Override
       protected Selector getFallbackSelector(int d) {
-        int skypedBytes = d + commonPrefix;
-        final int start = dim * bytesPerDim + skypedBytes;
-        final int end =  dim * bytesPerDim + bytesPerDim;
+        final int skypedBytes = d + commonPrefixLength;
+        final int dimStart = dim * bytesPerDim + skypedBytes;
+        final int dimEnd =  dim * bytesPerDim + bytesPerDim;
+        final int dataOffset = numIndexDims * bytesPerDim;
+        final int dataLength = (numDataDims - numIndexDims) * bytesPerDim + Integer.BYTES;
```
jpountz: I'm assuming the Integer.BYTES is about the doc ID, maybe add a comment to clarify since the variable name doesn't imply that this includes the doc ID?

iverase:
I have run the benchmarks and all and all it seems there is little effect in indexing throughput:

First run:
```
||Approach||Index time (sec)||Force merge time (sec)||Index size (GB)||Reader heap (MB)||
          ||Dev||Base||Diff ||Dev  ||Base  ||diff   ||Dev||Base||Diff||Dev||Base||Diff ||
|points|126.0s|125.6s| 0%|56.2s|60.0s|-6%|0.55|0.55| 0%|1.57|1.57| 0%|
|shapes|239.0s|218.7s| 9%|118.5s|115.6s| 3%|1.29|1.29| 0%|1.61|1.61| 0%|
|geo3d|159.0s|155.5s| 2%|61.3s|70.5s|-13%|0.75|0.75| 0%|1.58|1.58| 0%|
```
second run:
```
|points|131.7s|126.3s| 4%|56.4s|57.7s|-2%|0.55|0.55| 0%|1.57|1.57| 0%|
|shapes|215.3s|226.1s|-5%|116.4s|116.2s| 0%|1.29|1.29| 0%|1.61|1.61| 0%|
|geo3d|157.2s|158.0s|-1%|62.1s|66.5s|-7%|0.75|0.75| 0%|1.58|1.58| 0%|
```

PR #751, opened 2019-06-28
SOLR-13132: single sweep iteration over base, foreground, and background sets for "relatedness" calculation

magibney:
Relatedness essentially calculates facets separately
across three different docSets; this patch increments
facet counts over foreground, background, and base sets
in a single sweep, to leverage the efficiency of
existing faceting over the base set


magibney:
As mentioned in a comment on the Jira issue for [SOLR-13132](https://issues.apache.org/jira/browse/SOLR-13132?focusedCommentId=16754164#comment-16754164), this PR incorporates a cache for array-backed uninverted/docValue-style facet term counts (which are otherwise recalculated, at considerable cost, for every request). In some sense this "facet cache" is the more significant contribution here, and would be useful well beyond the context of "relatedness". (See also [SOLR-13807](https://issues.apache.org/jira/browse/SOLR-13807)).

magibney:
Force push of  bc4b18f separates into two logical commits: one introduces single-sweep collection of term facet counts over multiple domains, the second (which builds on the first) introduces a facet cache, which is more generally useful, but is particularly helpful for performance of relatedness calculation over relatively stable "background" sets.
The original monolithic PR, with some iterative improvements after @hossman's test stub patch, is available [here](https://github.com/magibney/lucene-solr/tree/SOLR-13132-mingled-sweep-and-cache).

magibney:
Per [@hossman's suggestion](https://issues.apache.org/jira/browse/SOLR-13807?focusedCommentId=17059139#comment-17059139), force-pushed 27f2e34 including only sweep collection. The "term facet cache" component of the original monolithic commit will be tracked separately at [SOLR-13807](https://issues.apache.org/jira/browse/SOLR-13807) (PR #1357).

tflobbe:
Can this PR be closed?

magibney:
Yes, it can be closed; sorry for the delayed response. (this was squash-merged as 40e2122b5a5b89f446e51692ef0d72e48c7b71e5)

PR #753, opened 2019-06-30
LUCENE-8895: switch all FST usage to enable array-with-gaps encoding

msokolov:
I'm not sure we want to commit this as-is, mainly because of the change to {{blocktreeords}} codec. With this change, if we encoded some node's outgoing Arcs as an array-with-gaps, we treat it as a list, scanning forward with a linear scan, rather than a binary search. Implementing binary search in this case might be possible, but it's quite complex. Another possibility is to leave the option of disabling the array-with-gaps opto and leaving it off for this one case. I'd like to get a sense how widely used this codec is and whether it's worth having that preserved in the API.


msokolov:
Based on feedback so far, this looks like a reasonable step. Also note latest commit removes unused {{Util.getByOutput()}} and related tests.

By the way, in addition to test and precommit, I also ran {{ant test  -Dtests.postingsformat=FST50}} since that uncovered some issues previously.

I'll push in a couple of days if there are no further comments.

msokolov:
OK, I realized that `Util.getByOutput` is a public method, so removing it in a minor rev. would be kind of rude, even if it is part of an @experimental API, given that it is not really necessary to remove, just cleanup. Besides, it's @dsmiley's favorite (part of the) API! So instead I'll just @Deprecate and we can remove later after some baking period, assuming no exciting use cases pop up.  I'll squash commits and push now

PR #754, opened 2019-07-01
LUCENE-8875: Introduce Optimized Collector For Large Number Of Hits

atris:
Existing top score docs collectors prepopulate the priority queue
with sentinel objects, which can be expensive for large number
of hits requested. This commit introduces a new Collector which
optimizes the said case


atris:
I ran some tests with N requested as 100K but the matching docs being only around 10K. This gives a sizeable improvement in the overall size of the priority queue.

lucene/sandbox/src/java/org/apache/lucene/search/LargeNumHitsTopDocsCollector.java
```diff
@@ -0,0 +1,72 @@
...
+import org.apache.lucene.index.LeafReaderContext;
+
+/**
+ * Optimized collector for large number of hits. This collector does
+ * not prepopulate the priority queue with sentinel values to avoid
+ * undue costs
+ */
+public class LargeNumHitsTopDocsCollector extends TopScoreDocCollector {
+
+  LargeNumHitsTopDocsCollector(int numHits, int totalHitsThreshold) {
```
jpountz: users won't be able to use it if it doesn't have a public constructor
also I don't think we need to pass a totalHitsThreshold: the fact that lots of hits need to be fetched would disable optimizations to the retrieval of top hits anyway

atris:
> Not prepopulating the hit queue is only one part of the problem, we would also need to not allocate `numHits` slots in the priority queue right away?
>
That would require a different PriorityQueue implementation, since the default implementation always allocates numHits slots and sets maxSize to numHits. We would need an implementation that only allocates two slots initially, but has the ability to "resize" as needed. WDYT?

> I'd rather like that we don't do any change to TopDocsCollector and not try to extend it, in my opinion those collectors are solving a very different problem and would need to evolve independently.

My main objective to use TopScoreDocsCollector was to reuse code as much as I could -- but I agree with your point. Will change the implementation and keep TopScoreDocsCollector pristine.

lucene/sandbox/src/java/org/apache/lucene/search/LargeNumHitsTopDocsCollector.java
atris: +1, will fix

jpountz:
> That would require a different PriorityQueue implementation

Actually I don't think we need a growable priority queue. For such large number of hits it'd be probably more efficient to collect hits in an ArrayList first and only turn it into a PQ once there are `numHits` hits?

Pinging @tokee since this is a topic he already spent time thinking about. :)

atris:
> Actually I don't think we need a growable priority queue. For such large number of hits it'd be probably more efficient to collect hits in an ArrayList first and only turn it into a PQ once there are `numHits` hits?

Would that mean collecting all hits in the ArrayList, building a heap out of them iteratively once we exhaust documents and then calling top() numHits times?

tokee:
I don't know if I have much to add to https://sbdevel.wordpress.com/2015/10/05/speeding-up-core-search/ that Atri linked to on the JIRA: Yes, it's definitely possible to do tricks with large result sets.  Especially with the simple "just sort on score"-case where the really large wins in my book is less GC pressure by using a single `long[]` to hold the structure instead of a gazillion small objects.

Here's the but: I haven't pursued it further as we (Royal Danish Library) have little use for it. Being able to handle large result sets in a single shards does not help much with multi-shard setups, where the merging node is likely to blow up. An iterative approach, such as `cursorMark` or Solr's `export`, is less prone to surprises and scales indefinitely. That being said, I won't stand in the way of building a fine foot blowing gun - for some use cases it would be a great win.

atris:
@jpountz @tokee Thanks for your comments.

I am planning to maintain an ArrayList of ScoreDocs and collect numHits, then do a sort on score and return top N as Adrien suggested. This should optimize the performance bottleneck around PQ allocating slots and prepopulating sentinel values.

I am inclined to pursue the idea of not using ScoreDoc and representing score + docID as a separate issue primarily due to the nuances involved with handling shard indices. I do feel that should not be a major blocker anymore given the fact that TopDocs.merge can now tie break on docIDs and ignore shard indices completely during tie breaking.

Thoughts?

atris:
@jpountz I have pushed a new iteration which does as discussed i.e. builds a hits list and populates hits as long as the number of collected hits are lesser than number of hits requested. Once that threshold is reached, a priority queue is built and minimum competitive score is set and used to filter further hits.

Please let me know if it looks fine.

lucene/sandbox/src/java/org/apache/lucene/search/LargeNumHitsTopDocsCollector.java
```diff
@@ -0,0 +1,177 @@
...
+          // Convert the list to a priority queue
+
+          // We should get here only when priority queue
+          // has been built
+          assert pq == null;
+          assert pqTop == null;
+          pq = new HitQueue(numHits, false);
+
+          for (int i = 0; i < hits.size(); i++) {
+            pq.add(hits.get(i));
```
jpountz: use `for (ScoreDoc scoreDoc : hits)`  instead?

lucene/sandbox/src/java/org/apache/lucene/search/LargeNumHitsTopDocsCollector.java
```diff
@@ -0,0 +1,177 @@
...
+      }
+    };
+  }
+
+  protected void updateMinCompetitiveScore(Scorable scorer) throws IOException {
+    if (pqTop != null) {
+      scorer.setMinCompetitiveScore(Math.nextUp(pqTop.score));
+    }
+    totalHitsRelation = TotalHits.Relation.GREATER_THAN_OR_EQUAL_TO;
+  }
```
jpountz: Maybe remove this logic, this is unlikely to help when fetching lots of hits.

lucene/sandbox/src/java/org/apache/lucene/search/LargeNumHitsTopDocsCollector.java
```diff
@@ -0,0 +1,177 @@
...
+        }
+        else if (o1.score > o2.score) {
+          return 1;
+        }
+
+        assert o1.doc != o2.doc;
+
+        return o1.doc > o2.doc ? 1 : -1;
+      }
+    });
```
jpountz: `Comparator.comparing(scoreDoc -> scoreDoc.score).reversed().thenComparing(scoreDoc -> scoreDoc.doc)`?

lucene/sandbox/src/java/org/apache/lucene/search/LargeNumHitsTopDocsCollector.java
```diff
@@ -0,0 +1,177 @@
...
+      return;
+    }
+
+    // Total number of hits collected were less than numHits
+    assert totalHits < numHits;
+    Collections.sort(hits, new Comparator<ScoreDoc>() {
+      @Override
+      public int compare(ScoreDoc o1, ScoreDoc o2) {
+        if (o1.score < o2.score) {
+          return -1;
```
jpountz: This seems to be putting lower scores first, while we'd like the higher scores first?

atris: Removed the comparator

atris:
@jpountz Thanks for the comments, update the PR.

lucene/sandbox/src/test/org/apache/lucene/search/TestLargeNumHitsTopDocsCollector.java
```diff
@@ -0,0 +1,133 @@
...
+
+    assertEquals(firstTopDocs.scoreDocs.length, secondTopDocs.scoreDocs.length);
+
+    for (int i = 0; i < firstTopDocs.scoreDocs.length; i++) {
+      ScoreDoc firstScoreDoc = firstTopDocs.scoreDocs[0];
+      ScoreDoc secondScoreDoc = secondTopDocs.scoreDocs[0];
+
+      assertEquals(firstScoreDoc.doc, secondScoreDoc.doc);
+      assertEquals(firstScoreDoc.score, secondScoreDoc.score, 0);
+    }
```
jpountz: You can replace this look with CheckHits#checkEquals.

lucene/sandbox/src/test/org/apache/lucene/search/TestLargeNumHitsTopDocsCollector.java
```diff
@@ -0,0 +1,133 @@
...
+    searcher.search(q, regularCollector);
+
+    assertEquals(largeCollector.totalHits, regularCollector.totalHits);
+
+    assertNotEquals(largeCollector.pq, null);
+    assertNotEquals(largeCollector.pqTop, null);
+  }
+
+  private void runNumHits(int numHits) throws IOException {
+    Query q = new MatchAllDocsQuery();
```
jpountz: Maybe use a query that doesn't produce constant scores to make sure that the collector orders by score? For instance you could add a term to some documents and then build a disjunction here across a query on this term and a MatchAllDocsQuery?

lucene/sandbox/src/test/org/apache/lucene/search/TestLargeNumHitsTopDocsCollector.java
```diff
@@ -0,0 +1,133 @@
...
+    TopScoreDocCollector regularCollector = TopScoreDocCollector.create(15, null, Integer.MAX_VALUE);
+
+    searcher.search(q, largeCollector);
+    searcher.search(q, regularCollector);
+
+    assertEquals(largeCollector.totalHits, regularCollector.totalHits);
+
+    expectThrows(IllegalArgumentException.class, () -> {
+      largeCollector.topDocs(350_000);
+    });
```
jpountz: can you check the error message?

jpountz:
@atris FYI avoiding force pushes would be helpful to reviewers as we could then look at what exactly changed compared to the previous PR.

lucene/sandbox/src/test/org/apache/lucene/search/TestLargeNumHitsTopDocsCollector.java
atris: +1, updated

atris:
@jpountz  I have updated the PR per your comments. ant precommit passes.

Apologies, this iteration also got force pushed. I have a local daemon which auto squashes and force pushes to my fork each time I create a new commit on a branch (this was meant to help the committer merge the PR without needing to squash). I will disable that in the next runs.

atris:
FYI ant precommit passes -- ran it with latest iteration

lucene/sandbox/src/java/org/apache/lucene/search/LargeNumHitsTopDocsCollector.java
```diff
@@ -0,0 +1,157 @@
...
+ * score.
+ */
+public class LargeNumHitsTopDocsCollector implements Collector {
+  private final int numHits;
+  private List<ScoreDoc> hits = new ArrayList<>();
+  HitQueue pq;
+  ScoreDoc pqTop;
+  int totalHits;
+  /** Whether {@link #totalHits} is exact or a lower bound. */
+  protected TotalHits.Relation totalHitsRelation = TotalHits.Relation.EQUAL_TO;
```
jpountz: No need for `protected` since since class is not designed to be extended? Maybe we should make it final by the way?

lucene/sandbox/src/java/org/apache/lucene/search/LargeNumHitsTopDocsCollector.java
```diff
@@ -0,0 +1,157 @@
...
+ * the requested number of hits. Post that, it builds a Priority Queue
+ * and starts filtering further hits based on the minimum competitive
+ * score.
+ */
+public class LargeNumHitsTopDocsCollector implements Collector {
+  private final int numHits;
+  private List<ScoreDoc> hits = new ArrayList<>();
+  HitQueue pq;
+  ScoreDoc pqTop;
+  int totalHits;
```
jpountz: make these members private?

lucene/sandbox/src/java/org/apache/lucene/search/LargeNumHitsTopDocsCollector.java
```diff
@@ -0,0 +1,163 @@
...
+   * Returns a {@link TopDocs} instance containing the given results. If
+   * <code>results</code> is null it means there are no results to return,
+   * either because there were 0 calls to collect() or because the arguments to
+   * topDocs were invalid.
+   */
+  protected TopDocs newTopDocs(ScoreDoc[] results) {
+    return results == null ? EMPTY_TOPDOCS : new TopDocs(new TotalHits(totalHits, totalHitsRelation), results);
+  }
+
+  public TopDocs topDocs() {
```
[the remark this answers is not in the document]
jpountz: Add javadocs?

lucene/sandbox/src/test/org/apache/lucene/search/TestLargeNumHitsTopDocsCollector.java
```diff
@@ -0,0 +1,141 @@
...
+    super.setUp();
+    dir = newDirectory();
+    RandomIndexWriter writer = new RandomIndexWriter(random(), dir);
+    for (int i = 0; i < 200_000; i++) {
+      if (random().nextBoolean()) {
+        Document doc = new Document();
+        doc.add(newStringField("field", "5", Field.Store.NO));
+        writer.addDocument(doc);
+      }
+      else {
```
jpountz: we usually put the else statement on the same line as the closing bracket of the if statement

lucene/sandbox/src/test/org/apache/lucene/search/TestLargeNumHitsTopDocsCollector.java
```diff
@@ -0,0 +1,141 @@
...
+  }
+
+  public void testNoPQBuild() throws IOException {
+    Query q = new MatchAllDocsQuery();
+    IndexSearcher searcher = newSearcher(reader);
+    LargeNumHitsTopDocsCollector largeCollector = new LargeNumHitsTopDocsCollector(250_000);
+    TopScoreDocCollector regularCollector = TopScoreDocCollector.create(250_000, null, Integer.MAX_VALUE);
+
+    searcher.search(q, largeCollector);
+    searcher.search(q, regularCollector);
```
jpountz: can you use testQuery instead to make sure that the list is sorted before creating top docs?

lucene/sandbox/src/test/org/apache/lucene/search/TestLargeNumHitsTopDocsCollector.java
```diff
@@ -0,0 +1,141 @@
...
+  }
+
+  public void testPQBuild() throws IOException {
+    Query q = new MatchAllDocsQuery();
+    IndexSearcher searcher = newSearcher(reader);
+    LargeNumHitsTopDocsCollector largeCollector = new LargeNumHitsTopDocsCollector(100_000);
+    TopScoreDocCollector regularCollector = TopScoreDocCollector.create(100_000, null, Integer.MAX_VALUE);
+
+    searcher.search(q, largeCollector);
+    searcher.search(q, regularCollector);
```
jpountz: use testQuery here as well?

lucene/sandbox/src/java/org/apache/lucene/search/LargeNumHitsTopDocsCollector.java
atris: That was something I was not sure about -- should we allow subclasses for this one? I agree with your line of thinking, will make it final

lucene/sandbox/src/java/org/apache/lucene/search/LargeNumHitsTopDocsCollector.java
atris: I am letting pq and pqTop be package private to allow tests to verify correctness

jpountz: Ah I see it now. Can you add "// pkg-private for testing" on top? This is usually what we do when we need to do this.

atris:
@jpountz Thanks, I have fixed the comments. The reason I did not use the BooleanQuery for the PQ build and not build tests was because those tests only exercise whether the priority queue was built or not which is dependent on the number of hits collected vs number of hits requested. However, per your suggestion, I have fixed the same.

atris:
Added a dedicated test for checking order of hits when TopDocs are returned from just the hits list and no PQ is built

jpountz:
Have you run tests? `TestLargeNumHitsTopDocsCollector.testLargeNumAndSparseHits` fails all the time for me.

jpountz:
Also tests run in several seconds right now, which is too slow. Can we make them run faster?

lucene/sandbox/src/java/org/apache/lucene/search/LargeNumHitsTopDocsCollector.java
```diff
@@ -0,0 +1,159 @@
...
+ * The collector maintains an ArrayList of hits until it accumulates
+ * the requested number of hits. Post that, it builds a Priority Queue
+ * and starts filtering further hits based on the minimum competitive
+ * score.
+ */
+public final class LargeNumHitsTopDocsCollector implements Collector {
+  private final int numHits;
+  private List<ScoreDoc> hits = new ArrayList<>();
+  /** Whether {@link #totalHits} is exact or a lower bound. */
+  private TotalHits.Relation totalHitsRelation = TotalHits.Relation.EQUAL_TO;
```
mikemccand: Why do we need a member for this?  It's never changed?  Can we just inline `TotalHits.Relation.EQUAL_TO` below?

lucene/sandbox/src/java/org/apache/lucene/search/LargeNumHitsTopDocsCollector.java
```diff
@@ -0,0 +1,159 @@
...
+  // package private for testing
+  HitQueue pq;
+  ScoreDoc pqTop;
+  int totalHits;
+
+  public LargeNumHitsTopDocsCollector(int numHits) {
+    this.numHits = numHits;
```
mikemccand: Can we rename `numHits` to `requestedHitCount` or so?  Otherwise `numHits` and `totalHits` are confusable?

lucene/sandbox/src/java/org/apache/lucene/search/LargeNumHitsTopDocsCollector.java
```diff
@@ -0,0 +1,159 @@
...
+
+        if (totalHits < numHits) {
+          hits.add(new ScoreDoc(doc, score));
+          totalHits++;
+          return;
+        } else if (totalHits == numHits) {
+          // Convert the list to a priority queue
+
+          // We should get here only when priority queue
+          // has been built
```
mikemccand: has `not yet` been built you mean?

lucene/sandbox/src/java/org/apache/lucene/search/LargeNumHitsTopDocsCollector.java
atris: Sure, I kept it here to increase readability -- will inline

lucene/sandbox/src/java/org/apache/lucene/search/LargeNumHitsTopDocsCollector.java
atris: Good catch, thanks :)

atris:
@jpountz @mikemccand Thanks for your comments, fixed the same.

I am surprised that the test is failing : it seems to be consistently passing for me. I ran a couple of times.

I have also reduced the dataset for tests: there is no real functional difference between 100 and 250k documents for testing, since these are not performance tests.

lucene/sandbox/src/java/org/apache/lucene/search/LargeNumHitsTopDocsCollector.java
```diff
@@ -0,0 +1,157 @@
...
+
+      @Override
+      public void collect(int doc) throws IOException {
+        float score = scorer.score();
+
+        // This collector relies on the fact that scorers produce positive values:
+        assert score >= 0; // NOTE: false for NaN
+
+        if (totalHits < requestedHitCount) {
+          hits.add(new ScoreDoc(doc, score));
```
jpountz: OK I dug the test failure, it was due to the fact that you are passing `doc` here instead of `docBase + doc`. I'm curious why you couldn't find it, it happened pretty often for me.

jpountz:
> Too bad we cannot share any (of this complex) code with existing collectors

The initial PR tried to share some logic, but I suggested to avoid doing this as I was expecting this collector to work fairly differently given that it operates on different assumptions. Looking at the current code, it doesn't seem to be like there is a lot to be shared?

> I am surprised that the test is failing : it seems to be consistently passing for me. I ran a couple of times.

I identified the bug, it was due to collecting `doc` instead of `docBase+doc` when collecting into an ArrayList. Maybe there is something to be refactored in the test so that it would have surfaced this issue more consistently?

atris:
@jpountz Thanks for looking through!

I refactored the tests to remove the dependency on random() to populate the string Field.  That triggered the test for me in the first run. Maybe something to do with the seed?

atris:
I just discovered ant beast that seems like the right tool for ensuring that such random tests get reproduced and are not dependent on lucky/unlucky seeds. Further PRs will have ant beast run on the new random tests before I raise a PR

wuyunfeng:
@atris  This PR is just resolve the cost for  prepopulating the priority queue with sentinel objects?   So amazing for the cost for  prepopulating.

atris:
Not just the cost of prepopulating the PQ, but also building it in first place. If total number of hits < number of requested hits and number of requested hits is  a large value, then it makes sense to skip the PQ

wuyunfeng:
@atris  Can you give some result comparison or other evidence?

atris:
> @atris Can you give some result comparison or other evidence?

Please see the blog post listed above.

In addition, the targeted use case for this collector makes the benefits obvious: I ran a TermQuery on a 5 TB dataset manufactured from Elasticsearch's nyc_taxis dataset. A string field was modified to have a specific value only for 50,000 documents, and number of hits requested was 25,00,000 by each of 30 concurrent requests (the same test mentioned in Toke's blogpost). The space savings were in order of multiple GBs, as expected.

atris:
@jpountz JFYI I ran ant beast in 3 batches of 10 times each and it ran clean:

WARNING: All illegal access operations will be denied in a future release
  [beaster] Beast round 1 results: /Users/atris/merge_refactor/lucene-solr/lucene/build/sandbox/test/1
  [beaster] Beast round 2 results: /Users/atris/merge_refactor/lucene-solr/lucene/build/sandbox/test/2
  [beaster] Beast round 3 results: /Users/atris/merge_refactor/lucene-solr/lucene/build/sandbox/test/3
  [beaster] Beast round 4 results: /Users/atris/merge_refactor/lucene-solr/lucene/build/sandbox/test/4
  [beaster] Beast round 5 results: /Users/atris/merge_refactor/lucene-solr/lucene/build/sandbox/test/5
  [beaster] Beast round 6 results: /Users/atris/merge_refactor/lucene-solr/lucene/build/sandbox/test/6
  [beaster] Beast round 7 results: /Users/atris/merge_refactor/lucene-solr/lucene/build/sandbox/test/7
  [beaster] Beast round 8 results: /Users/atris/merge_refactor/lucene-solr/lucene/build/sandbox/test/8
  [beaster] Beast round 9 results: /Users/atris/merge_refactor/lucene-solr/lucene/build/sandbox/test/9
  [beaster] Beast round 10 results: /Users/atris/merge_refactor/lucene-solr/lucene/build/sandbox/test/10
  [beaster] Beasting finished Successfully.

-check-totals:

beast:

BUILD SUCCESSFUL

wuyunfeng:
@atris Hi，atris.  which Lucene release version would use this PR？I hope i can use this.

PR #758, opened 2019-07-02
LUCENE-8833: Add a #load() method to IndexInput to allow preloading file content into physical memory

s1monw:
Today we have an all or nothing option on MMapDirectory to preload the contents
of the file into physical memory. Yet, with recent efforts moving datastructures
to disk if they can be accessed efficiently can benefit from a per IndexInput API
to make better decisions for defaults if the index is mmapped.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/codecs/blocktree/FieldReader.java
+++ b/lucene/core/src/java/org/apache/lucene/codecs/blocktree/FieldReader.java
@@ -25,7 +25,6 @@ import org.apache.lucene.index.IndexOptions;
 import org.apache.lucene.index.Terms;
 import org.apache.lucene.index.TermsEnum;
 import org.apache.lucene.store.ByteArrayDataInput;
-import org.apache.lucene.store.ByteBufferIndexInput;
 import org.apache.lucene.store.IndexInput;
 import org.apache.lucene.util.Accountable;
 import org.apache.lucene.util.Accountables;
@@ -98,7 +97,7 @@ public final class FieldReader extends Terms implements Accountable {
           isFSTOffHeap = ((this.docCount != this.sumDocFreq) || openedFromWriter == false);
           break;
         case AUTO:
-          isFSTOffHeap = ((this.docCount != this.sumDocFreq) || openedFromWriter == false) && indexIn instanceof ByteBufferIndexInput;
+          isFSTOffHeap = ((this.docCount != this.sumDocFreq) || openedFromWriter == false) &&  indexIn.load(); // NOCOMMIT - should we do this should we do it all the time?
           break;
         default:
           throw new IllegalStateException("unknown enum constant: " + fstLoadMode);
--- a/lucene/core/src/java/org/apache/lucene/store/ByteBufferIndexInput.java
+++ b/lucene/core/src/java/org/apache/lucene/store/ByteBufferIndexInput.java
@@ -21,6 +21,7 @@ import java.io.EOFException;
 import java.io.IOException;
 import java.nio.BufferUnderflowException;
 import java.nio.ByteBuffer;
+import java.nio.MappedByteBuffer;
 
 /**
  * Base IndexInput implementation that uses an array
@@ -490,5 +491,15 @@ public abstract class ByteBufferIndexInput extends IndexInput implements RandomA
     protected ByteBufferIndexInput buildSlice(String sliceDescription, long ofs, long length) {
       return super.buildSlice(sliceDescription, this.offset + ofs, length);
     }
+  }
+
+  @Override
+  public boolean load() {
+    for (ByteBuffer buffer : buffers) {
+      if (buffer instanceof MappedByteBuffer) {
+        ((MappedByteBuffer) buffer).load();
+      }
+    }
+    return true;
   }
 }
--- a/lucene/core/src/java/org/apache/lucene/store/MMapDirectory.java
+++ b/lucene/core/src/java/org/apache/lucene/store/MMapDirectory.java
@@ -91,7 +91,6 @@ import org.apache.lucene.util.SuppressForbidden;
  */
 public class MMapDirectory extends FSDirectory {
   private boolean useUnmapHack = UNMAP_SUPPORTED;
-  private boolean preload;
 
   /** 
    * Default max chunk size.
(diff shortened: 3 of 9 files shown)


lucene/core/src/java/org/apache/lucene/store/ByteBufferIndexInput.java
```diff
@@ -491,4 +492,14 @@ protected ByteBufferIndexInput buildSlice(String sliceDescription, long ofs, lon
       return super.buildSlice(sliceDescription, this.offset + ofs, length);
     }
   }
+
+  @Override
+  public boolean load() {
+    for (ByteBuffer buffer : buffers) {
+      if (buffer instanceof MappedByteBuffer) {
+        ((MappedByteBuffer) buffer).load();
```
jpountz: This would load more than necessary in the case of a slice. If you look at ByteBufferIndexInput#buildSlice, the offset of the first ByteBuffer might not be updated, so calling load() this way might load up to 1GB of data in main memory that we don't need. This might be especially problematic in the case of CFS files where all index inputs are slices of a single large file. We should maybe override this is MultiBufferImpl to ignore the first `offset` bytes of the first slice?

s1monw: good point I will look into this. I might need to think more about testing. I am not happy with the status quo.

lucene/core/src/java/org/apache/lucene/codecs/blocktree/FieldReader.java
```diff
@@ -98,7 +97,7 @@
           isFSTOffHeap = ((this.docCount != this.sumDocFreq) || openedFromWriter == false);
           break;
         case AUTO:
-          isFSTOffHeap = ((this.docCount != this.sumDocFreq) || openedFromWriter == false) && indexIn instanceof ByteBufferIndexInput;
+          isFSTOffHeap = ((this.docCount != this.sumDocFreq) || openedFromWriter == false) &&  indexIn.load(); // NOCOMMIT - should we do this should we do it all the time?
```
jpountz: This looks reasonable to me. The extreme minority of expert users who don't want to preload this slice can still use a Directory wrapper whose index inputs don't do anything in load() and return `true`.

solr/core/src/java/org/apache/solr/core/MMapDirectoryFactory.java
```diff
@@ -35,7 +37,8 @@
  * Can set the following parameters:
  * <ul>
  *  <li>unmap -- See {@link MMapDirectory#setUseUnmap(boolean)}</li>
- *  <li>preload -- See {@link MMapDirectory#setPreload(boolean)}</li>
+ *  <li>preload -- Set to {@code true} to ask mapped pages to be loaded into physical memory on init.
+ *  he behavior is best-effort and operating system dependent.</li>
```
mikemccand: `he` -> `The`

lucene/core/src/java/org/apache/lucene/codecs/blocktree/FieldReader.java
mikemccand: +1

Extra space after the `&&`?

lucene/core/src/java/org/apache/lucene/store/MMapDirectory.java
```diff
@@ -202,24 +201,6 @@ public boolean getUseUnmap() {
     return useUnmapHack;
   }
   
-  /**
-   * Set to {@code true} to ask mapped pages to be loaded
-   * into physical memory on init. The behavior is best-effort 
-   * and operating system dependent.
-   * @see MappedByteBuffer#load
-   */
-  public void setPreload(boolean preload) {
```
mikemccand: We lose the ability to always pre-load here?  But one can always make a small wrapper like `MMapDirectoryFactory` does; maybe add that as an example to `MIGRATE.txt`?

s1monw: sure I can do that

lucene/core/src/java/org/apache/lucene/store/ByteBufferIndexInput.java
s1monw: I pushed changes to address this. I still need to look into testing more.

s1monw:
@uschindler I think you should look at this too.

s1monw:
ping @uschindler again

lucene/core/src/java/org/apache/lucene/store/MMapDirectory.java
uschindler: I tend to keep the "always preload method", just deprecate it for now. I know many people that use this globally. It has the side-effect of setting madvise to "WILL_NEED" for the whole buffer (in addition to load it), so it is a bit more sticky as it's marked to the kernel as "i will need it soon".

solr/core/src/java/org/apache/solr/core/MMapDirectoryFactory.java
```diff
@@ -61,13 +64,23 @@ public void init(NamedList args) {
   @Override
   protected Directory create(String path, LockFactory lockFactory, DirContext dirContext) throws IOException {
     // we pass NoLockFactory, because the real lock factory is set later by injectLockFactory:
-    MMapDirectory mapDirectory = new MMapDirectory(new File(path).toPath(), lockFactory, maxChunk);
+    final boolean preloadIndexInput = this.preload;
+    MMapDirectory mapDirectory = new MMapDirectory(new File(path).toPath(), lockFactory, maxChunk) {
```
uschindler: I know you don't care about Solr, but this seems to change the system information to show a strange class name instead of the current MMapDirectory output in the system info page of Solr. I have to investigate this, but as said before, I tend to not remove the setPreload() method.

uschindler:
> I am still in review (have to feed my brain with all details, reading C++ source code of JVM and so on).
> 
> Because of Adrien's comment you changed the slice building to also take the start position into account, so the slice method returns a byte buffer that is really only a view and a call to load() would not read data of the whole CFS file.
> 
> As you changed this method to take care of the start offset/position, please fix the javadocs comment:
> 
> ```
>   /** Returns a sliced view from a set of already-existing buffers: 
>    *  the last buffer's limit() will be correct, but
>    *  you must deal with offset separately (the first buffer will not be adjusted) */
>   private ByteBuffer[] buildSlice(ByteBuffer[] buffers, long offset, long length) {
> ```

I am not fully sure how the comment would look like. Currently it only fixes the start offset for single-buffer indexinputs.

The code is horrible here, I still don't fully understand the code difference for single and multi buffer indexinputs. IMHO, we should change the code and make buildSlice always fix position() and limit() and build a buffer slice for the first and last buffer. How it's done currently is hard to understand. I cannot make any guarantees that the code behaves correctly after this patch and I am not sure if testing of all cases is fine. Maybe only apply to master first and let Jenkins test it.

PR #763, opened 2019-07-05
SOLR-13608: Incremental backup for Solr

CaoManhDat:


tejassawai:
It is still in Open state, which means we cannot have the incremental backup as of now in apache solr, right? Or is there any possibility to do so?

gerlowskija:
This PR has fallen out of date, especially with some of the review feedback drawn out during the SIP-12 discussion process.  I'm going to close this in favor of moving forward with the replacement PR [here](https://github.com/apache/lucene-solr/pull/2250) (which is based on very similar code).

PR #767, opened 2019-07-07
LUCENE-8904: enhance Nori DictionaryBuilder tool

danmuzi:
It is the Nori version of LUCENE-8863.
This patch has two changes.
1) Improve exception handling
2) Enable external dictionary for testing

Please check the following link:
https://issues.apache.org/jira/browse/LUCENE-8904

### DIFF
--- a/lucene/analysis/nori/src/tools/java/org/apache/lucene/analysis/ko/util/BinaryDictionaryWriter.java
+++ b/lucene/analysis/nori/src/tools/java/org/apache/lucene/analysis/ko/util/BinaryDictionaryWriter.java
@@ -143,13 +143,16 @@ public abstract class BinaryDictionaryWriter {
       throw new IllegalArgumentException("leftId >= " + ID_LIMIT + ": " + leftId);
     }
     if (posType.ordinal() >= 4) {
-      throw new IllegalArgumentException("posType.ordinal() >= " + 4 + ": " + posType.ordinal());
+      throw new IllegalArgumentException("posType.ordinal() >= " + 4 + ": " + posType.name());
     }
     buffer.putShort((short)(leftId << 2 | posType.ordinal()));
     buffer.putShort((short) (rightId << 2 | flags));
     buffer.putShort(wordCost);
 
     if (posType == POS.Type.MORPHEME) {
+      if (leftPOS != rightPOS) {
+        throw new IllegalArgumentException("leftPOS != rightPOS: " + leftPOS + " " + rightPOS);
+      }
       if (reading.length() > 0) {
         writeString(reading);
       }
--- a/lucene/analysis/nori/src/tools/java/org/apache/lucene/analysis/ko/util/BinaryDictionaryWriter.java
+++ b/lucene/analysis/nori/src/tools/java/org/apache/lucene/analysis/ko/util/BinaryDictionaryWriter.java
@@ -38,6 +38,8 @@ import org.apache.lucene.util.ArrayUtil;
 import org.apache.lucene.analysis.ko.dict.BinaryDictionary;
 
 public abstract class BinaryDictionaryWriter {
+  private final static int ID_LIMIT = 8192;
+
   protected final Class<? extends BinaryDictionary> implClazz;
   protected ByteBuffer buffer;
   private int targetMapEndOffset = 0, lastWordId = -1, lastSourceId = -1;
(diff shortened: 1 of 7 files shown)


lucene/analysis/nori/src/tools/java/org/apache/lucene/analysis/ko/util/BinaryDictionaryWriter.java
```diff
@@ -137,14 +139,17 @@ public int put(String[] entry) {
       flags |= BinaryDictionary.HAS_READING;
     }
 
-    assert leftId < 8192; // there are still unused bits
-    assert posType.ordinal() < 4;
+    if (leftId >= ID_LIMIT) {
+      throw new IllegalArgumentException("leftId >= " + ID_LIMIT + ": " + leftId);
+    }
+    if (posType.ordinal() >= 4) {
+      throw new IllegalArgumentException("posType.ordinal() >= " + 4 + ": " + posType.ordinal());
```
msokolov: Maybe also log posType itself? If this ever happens, that should help

lucene/analysis/nori/src/tools/java/org/apache/lucene/analysis/ko/util/BinaryDictionaryWriter.java
```diff
@@ -137,14 +139,17 @@ public int put(String[] entry) {
...
+    }
     buffer.putShort((short)(leftId << 2 | posType.ordinal()));
     buffer.putShort((short) (rightId << 2 | flags));
     buffer.putShort(wordCost);
 
     if (posType == POS.Type.MORPHEME) {
-      assert leftPOS == rightPOS;
```
msokolov: I don't understand how this works in Korean - are you sure it's OK to remove?

lucene/analysis/nori/src/tools/java/org/apache/lucene/analysis/ko/util/BinaryDictionaryWriter.java
danmuzi: +1.

Which log style is better?
1) `throw new IllegalArgumentException("posType.ordinal() >= " + 4 + ": " + posType.name());`
2) `throw new IllegalArgumentException("posType should be MORPHOME or COMPOUND or INFLECT or PREANALYSIS" + ": " + posType.name());`

lucene/analysis/nori/src/tools/java/org/apache/lucene/analysis/ko/util/BinaryDictionaryWriter.java
danmuzi: It is an unnecessary condition.
The leftPOS and rightPOS are determined on lines 83-90.
If posType is POS.Type.MORPHEME, leftPOS and rightPOS are always the same.

danmuzi:
Thank you for your review! @msokolov
I left some replies :D

lucene/analysis/nori/src/tools/java/org/apache/lucene/analysis/ko/util/BinaryDictionaryWriter.java
msokolov: Well, assertions are always unnecessary! I think their value is exactly in helping ensure that the correct code path was followed to get here, and confirming what we know to be true by analysis or anyway knowledge, in fact holds in practice (empirically). For example, if I were to somehow mess up the code above that guarantees this, due to my ignorance, this assertion would helpfully let me know. So I think we should keep it?

lucene/analysis/nori/src/tools/java/org/apache/lucene/analysis/ko/util/BinaryDictionaryWriter.java
msokolov: (1), please!

lucene/analysis/nori/src/tools/java/org/apache/lucene/analysis/ko/util/BinaryDictionaryWriter.java
danmuzi: OK :)
I understand what you mean and I'll keep it.

lucene/analysis/nori/src/tools/java/org/apache/lucene/analysis/ko/util/BinaryDictionaryWriter.java
danmuzi: Thanks :)

danmuzi:
Thank you for your review! @msokolov 
I'll merge(Squash and merge) this PR and close.

PR #769, opened 2019-07-08
LUCENE-8905: Better Error Handling For Illegal Arguments

atris:


msokolov:
This seems better to me, but it could break people that rely on the leniency, so it should only go to master, and requires a changes entry to warn people

atris:
Ok, it breaks quite a lot of tests. I investigated around 30 odd failures, and looks like all of those tests are failing on the new exception. Does that mean that our tests are silently passing in malformed arguments, and relying on the inability of TopDocsCollector to complain?

lucene/CHANGES.txt
```diff
@@ -25,6 +25,8 @@ API Changes
 * LUCENE-8811: Deprecated BooleanQuery#setMaxClauseCount() and getMaxClauseCount
   have been removed (Atri Sharma, Adrien Grand, Alan Woodward)
 
+* LUCENE-8905: Better Defence Against Malformed Arguments in TopDocsCollector
```
mikemccand: We don't typically capitalize the whole sentence in `CHANGES.txt`, just the first word.

lucene/MIGRATE.txt
```diff
@@ -190,3 +190,10 @@ set shard indices for hits as they are seen during the merge process. This is do
 to be more dynamic in terms of passing in custom tie breakers.
 If shard indices are to be used for tie breaking docs with equal scores during TopDocs.merge, then it is
 mandatory that the input ScoreDocs have their shard indices set to valid values prior to calling TopDocs.merge
+
+## TopDocsCollector Shall Throw IllegalStateException For Malformed Arguments ##
+
+TopDocsCollector shall no longer return an empty TopDocs for malformed arguments.
+Rather, an IllegalStateException shall be thrown. This is introduced for better
```
mikemccand: `IllegalArgumentException`?

atris:
Thanks @mikemccand and @msokolov 

Unfortunately the number of tests failing is a bit too huge :( I was hoping we could have a gradual effort (maybe take this on as a technical debt and push it to a branch?)

mikemccand:
Can you share one or two of the failures?  How are so many tests relying on this leniency?

atris:
[junit4] Tests with failures [seed: 214CBCDED4C34EFB] (first 10 out of 156):
   [junit4]   - org.apache.lucene.search.spans.TestNearSpansOrdered.testOverlappedOrderedSpan
   [junit4]   - org.apache.lucene.search.TestTopFieldCollectorEarlyTermination.testEarlyTermination
   [junit4]   - org.apache.lucene.search.TestTopFieldCollectorEarlyTermination.testEarlyTerminationWhenPaging
   [junit4]   - org.apache.lucene.search.TestSloppyPhraseQuery2.testIncreasingSloppiness
   [junit4]   - org.apache.lucene.search.TestSloppyPhraseQuery2.testRepetitiveIncreasingSloppinessWithHoles
   [junit4]   - org.apache.lucene.search.TestSloppyPhraseQuery2.testIncreasingSloppiness3WithHoles
   [junit4]   - org.apache.lucene.search.TestSloppyPhraseQuery2.testRepetitiveIncreasingSloppiness3
   [junit4]   - org.apache.lucene.search.TestSloppyPhraseQuery2.testIncreasingSloppiness3
   [junit4]   - org.apache.lucene.search.TestSloppyPhraseQuery2.testIncreasingSloppinessWithHoles
   [junit4]   - org.apache.lucene.search.TestSloppyPhraseQuery2.testRepetitiveIncreasingSloppiness

jpountz:
Can you share the stack trace of the exception? I suspect we could get rid of most test failures by fixing only a couple call sites.

jpountz:
For instance I suspect that most of those are due to the call to `return TopFieldCollector.create(rewrittenSort, cappedNumHits, after, TOTAL_HITS_THRESHOLD);` in IndexSearcher, which we'd just need to replace with `return TopFieldCollector.create(rewrittenSort, cappedNumHits, after, Math.max(cappedNumHits, TOTAL_HITS_THRESHOLD));`?

lucene/core/src/test/org/apache/lucene/search/TestTopDocsCollector.java
```diff
@@ -179,6 +179,22 @@ public void testGetResultsFromStart() throws Exception {
...
   }
+
+  public void testIllegalArguments() throws Exception {
+    final TopDocsCollector<ScoreDoc> tdc = doSearch(15);
+
+    IllegalArgumentException expected = expectThrows(IllegalArgumentException.class, () -> {
+      tdc.topDocs(20);
+    });
+
+    assertEquals(expected.getMessage(), "Illegal arguments when requesting top docs");
```
jpountz: nit: Junit assumes that the expected value is the first argument and the second value is what the test produced. Swapping both would produce better error messages.

lucene/core/src/java/org/apache/lucene/search/TopDocsCollector.java
```diff
@@ -136,12 +136,9 @@ public TopDocs topDocs(int start, int howMany) {
...
     int size = topDocsSize();
 
-    // Don't bother to throw an exception, just return an empty TopDocs in case
-    // the parameters are invalid or out of range.
-    // TODO: shouldn't we throw IAE if apps give bad params here so they dont
-    // have sneaky silent bugs?
+
     if (start < 0 || start >= size || howMany <= 0) {
-      return newTopDocs(null, start);
+      throw new IllegalArgumentException("Illegal arguments when requesting top docs");
```
jpountz: can you add the value of start and howMany to the error message and what the expectation is for their values?

atris:
> For instance I suspect that most of those are due to the call to `return TopFieldCollector.create(rewrittenSort, cappedNumHits, after, TOTAL_HITS_THRESHOLD);` in IndexSearcher, which we'd just need to replace with `return TopFieldCollector.create(rewrittenSort, cappedNumHits, after, Math.max(cappedNumHits, TOTAL_HITS_THRESHOLD));`?

-Dtests.timezone=Etc/GMT+5 -Dtests.asserts=true -Dtests.file.encoding=UTF-8
   [junit4] ERROR   0.01s J0 | TestIndexSearcher.testHugeN <<<

 Throwable #1: java.lang.IllegalArgumentException: Illegal arguments when requesting top docs
   [junit4]    > 	at __randomizedtesting.SeedInfo.seed([5E8112825AFE9F23:231370EC0B1BDF30]:0)
   [junit4]    > 	at org.apache.lucene.search.TopDocsCollector.topDocs(TopDocsCollector.java:141)
   [junit4]    > 	at org.apache.lucene.search.TopDocsCollector.topDocs(TopDocsCollector.java:96)
   [junit4]    > 	at org.apache.lucene.search.IndexSearcher$2.reduce(IndexSearcher.java:480)
   [junit4]    > 	at org.apache.lucene.search.IndexSearcher$2.reduce(IndexSearcher.java:468)
   [junit4]    > 	at org.apache.lucene.search.IndexSearcher.search(IndexSearcher.java:636)
   [junit4]    > 	at org.apache.lucene.search.IndexSearcher.searchAfter(IndexSearcher.java:487)
   [junit4]    > 	at org.apache.lucene.search.TestIndexSearcher.testHugeN(TestIndexSearcher.java:104)
   [junit4]    > 	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
   [junit4]    > 	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
   [junit4]    > 	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)

atris:
@jpountz Another one:

TestBooleanMinShouldMatch.testMinHigherThenNumOptional <<<
   [junit4]    > Throwable #1: java.lang.IllegalArgumentException: Illegal arguments when requesting top docs
   [junit4]    > 	at __randomizedtesting.SeedInfo.seed([E2642EBDFCA3189E:98AE10AFC431B37C]:0)
   [junit4]    > 	at org.apache.lucene.search.TopDocsCollector.topDocs(TopDocsCollector.java:141)
   [junit4]    > 	at org.apache.lucene.search.TopDocsCollector.topDocs(TopDocsCollector.java:96)
   [junit4]    > 	at org.apache.lucene.search.IndexSearcher$2.reduce(IndexSearcher.java:480)
   [junit4]    > 	at org.apache.lucene.search.IndexSearcher$2.reduce(IndexSearcher.java:468)
   [junit4]    > 	at org.apache.lucene.search.IndexSearcher.search(IndexSearcher.java:636)
   [junit4]    > 	at org.apache.lucene.search.IndexSearcher.searchAfter(IndexSearcher.java:487)
   [junit4]    > 	at org.apache.lucene.search.IndexSearcher.search(IndexSearcher.java:498)
   [junit4]    > 	at org.apache.lucene.search.TestBooleanMinShouldMatch.verifyNrHits(TestBooleanMinShouldMatch.java:87)
   [junit4]    > 	at org.apache.lucene.search.TestBooleanMinShouldMatch.testMinHigherThenNumOptional(TestBooleanMinShouldMatch.java:256)
   [junit4]    > 	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
   [junit4]    > 	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
   [junit4]    > 	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
   [junit4]    > 	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
   [junit4]    > 	at java.base/java.lang.Thread.run(Thread.java:834)

atris:
@jpountz I updated the PR per comments.

Given some of the error stack traces, what is your advice on the next step forward?

atris:
@mikemccand Thanks, updated the PR.

lucene/core/src/java/org/apache/lucene/search/TopDocsCollector.java
```diff
@@ -136,12 +136,14 @@ public TopDocs topDocs(int start, int howMany) {
...
-    if (start < 0 || start >= size || howMany <= 0) {
-      return newTopDocs(null, start);
+
+    if (start < 0 || start >= size) {
+      throw new IllegalArgumentException("Expected value of starting position is between 0 and " + size +
+          ", got " + start);
+    }
+
+    if (howMany <= 0) {
+      throw new IllegalArgumentException("Number of hits requested must be greater than 0");
```
jpountz: can you add the value of `howMany` to the error message?

atris:
> @atris I think the check you introduced always fails a query when there are no hits?

@jpountz Thanks for highlighting that. Interesting that TopDocsCollector handles no hits and illegal argument cases in the same check.

atris:
@jpountz Updated the PR. With the current version, all tests for both Lucene and Solo pass (yay!).

atris:
Any further thoughts on this one? Happy to iterate

lucene/core/src/java/org/apache/lucene/search/TopDocsCollector.java
```diff
@@ -136,14 +136,21 @@ public TopDocs topDocs(int start, int howMany) {
     // pq.size() or totalHits.
     int size = topDocsSize();
 
-    // Don't bother to throw an exception, just return an empty TopDocs in case
-    // the parameters are invalid or out of range.
-    // TODO: shouldn't we throw IAE if apps give bad params here so they dont
-    // have sneaky silent bugs?
-    if (start < 0 || start >= size || howMany <= 0) {
+
+    if (start < 0 || start > size) {
```
jpountz: I don't think we should validate `start` against `size` since `size` depends on the number of matches, and callers have no clue about its value at this point? Otherwise we are making this method almost impossible to use?

The thing that we should validate is that `start + howMany` is less than or equal to the `numHits` parameter (which doesn't seem to be passed to TopDocsCollector by its sub classes at the moment, but we could)?

atris: @jpountz That can allow potential abuse by passing in Integer.MAX_VALUE for numHits, and a value of start > size. With this check gone, we would not be able to detect the case.

Should we keep the existing one and add the one you proposed?

mikemccand:
I ran top-level `ant test` and hit this exception on the PR as of yesterday:

```
   [junit4] Suite: org.apache.lucene.search.grouping.TestGrouping
   [junit4]   2> NOTE: reproduce with: ant test  -Dtestcase=TestGrouping -Dtests.method=testRandom -Dtests.seed=F08BE82DC9B3756C -Dtests.badapples=true -Dtests.locale=es-BZ -Dtests.timezone\
=Africa/Gaborone -Dtests.asserts=true -Dtests.file.encoding=ISO-8859-1
   [junit4] ERROR   1.65s J0 | TestGrouping.testRandom <<<
   [junit4]    > Throwable #1: java.lang.IllegalArgumentException: Expected value of starting position is between 0 and 3, got 7
   [junit4]    >        at __randomizedtesting.SeedInfo.seed([F08BE82DC9B3756C:82C7CD2278D3C31F]:0)
   [junit4]    >        at org.apache.lucene.search.TopDocsCollector.topDocs(TopDocsCollector.java:141)
   [junit4]    >        at org.apache.lucene.search.grouping.TopGroupsCollector.getTopGroups(TopGroupsCollector.java:163)
   [junit4]    >        at org.apache.lucene.search.grouping.TestGrouping.getTopGroups(TestGrouping.java:319)
   [junit4]    >        at org.apache.lucene.search.grouping.TestGrouping.testRandom(TestGrouping.java:965)
   [junit4]    >        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
   [junit4]    >        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
   [junit4]    >        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
   [junit4]    >        at java.base/java.lang.reflect.Method.invoke(Method.java:566)
   [junit4]    >        at java.base/java.lang.Thread.run(Thread.java:834)
   [junit4]   2> NOTE: test params are: codec=Asserting(Lucene80): {groupend=TestBloomFilteredLucenePostings(BloomFilteringPostingsFormat(Lucene50(blocksize=128))), sort1=FST50, sort2=TestB\
loomFilteredLucenePostings(BloomFilteringPostingsFormat(Lucene50(blocksize=128))), content=PostingsFormat(name=Direct), group=FST50}, docValues:{author=DocValuesFormat(name=Asserting), sort\
1=DocValuesFormat(name=Lucene80), id=DocValuesFormat(name=Lucene80), sort2=DocValuesFormat(name=Direct), ___soft_deletes=DocValuesFormat(name=Lucene80), group=DocValuesFormat(name=Lucene80)\
}, maxPointsInLeafNode=2036, maxMBSortInHeap=7.045070063116054, sim=Asserting(org.apache.lucene.search.similarities.AssertingSimilarity@19474185), locale=es-BZ, timezone=Africa/Gaborone
   [junit4]   2> NOTE: Linux 4.4.0-96-generic amd64/Oracle Corporation 11.0.2 (64-bit)/cpus=8,threads=1,free=510432528,total=536870912
   [junit4]   2> NOTE: All tests run in this JVM: [TestGrouping]
```

atris:
> I ran top-level `ant test` and hit this exception on the PR as of yesterday:
> 
> ```
>    [junit4] Suite: org.apache.lucene.search.grouping.TestGrouping
>    [junit4]   2> NOTE: reproduce with: ant test  -Dtestcase=TestGrouping -Dtests.method=testRandom -Dtests.seed=F08BE82DC9B3756C -Dtests.badapples=true -Dtests.locale=es-BZ -Dtests.timezone\
> =Africa/Gaborone -Dtests.asserts=true -Dtests.file.encoding=ISO-8859-1
>    [junit4] ERROR   1.65s J0 | TestGrouping.testRandom <<<
>    [junit4]    > Throwable #1: java.lang.IllegalArgumentException: Expected value of starting position is between 0 and 3, got 7
>    [junit4]    >        at __randomizedtesting.SeedInfo.seed([F08BE82DC9B3756C:82C7CD2278D3C31F]:0)
>    [junit4]    >        at org.apache.lucene.search.TopDocsCollector.topDocs(TopDocsCollector.java:141)
>    [junit4]    >        at org.apache.lucene.search.grouping.TopGroupsCollector.getTopGroups(TopGroupsCollector.java:163)
>    [junit4]    >        at org.apache.lucene.search.grouping.TestGrouping.getTopGroups(TestGrouping.java:319)
>    [junit4]    >        at org.apache.lucene.search.grouping.TestGrouping.testRandom(TestGrouping.java:965)
>    [junit4]    >        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
>    [junit4]    >        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
>    [junit4]    >        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
>    [junit4]    >        at java.base/java.lang.reflect.Method.invoke(Method.java:566)
>    [junit4]    >        at java.base/java.lang.Thread.run(Thread.java:834)
>    [junit4]   2> NOTE: test params are: codec=Asserting(Lucene80): {groupend=TestBloomFilteredLucenePostings(BloomFilteringPostingsFormat(Lucene50(blocksize=128))), sort1=FST50, sort2=TestB\
> loomFilteredLucenePostings(BloomFilteringPostingsFormat(Lucene50(blocksize=128))), content=PostingsFormat(name=Direct), group=FST50}, docValues:{author=DocValuesFormat(name=Asserting), sort\
> 1=DocValuesFormat(name=Lucene80), id=DocValuesFormat(name=Lucene80), sort2=DocValuesFormat(name=Direct), ___soft_deletes=DocValuesFormat(name=Lucene80), group=DocValuesFormat(name=Lucene80)\
> }, maxPointsInLeafNode=2036, maxMBSortInHeap=7.045070063116054, sim=Asserting(org.apache.lucene.search.similarities.AssertingSimilarity@19474185), locale=es-BZ, timezone=Africa/Gaborone
>    [junit4]   2> NOTE: Linux 4.4.0-96-generic amd64/Oracle Corporation 11.0.2 (64-bit)/cpus=8,threads=1,free=510432528,total=536870912
>    [junit4]   2> NOTE: All tests run in this JVM: [TestGrouping]
> ```

Interesting, I cannot see this for previous versions of Lucene, but rebasing got me this error. I added the check Adrien suggested and got the following failure:

java.lang.IllegalArgumentException: Requested number of hits greater than original hits limit for the collector. size 4 start 2 howMany 49

I think why that is happening is because today, we fix howMany to be within valid limits in case it is larger than the available number of hits, rather than erroring out.

jpountz:
I'm wondering whether we should consider removing this `topDocs` variant which takes a `howMany` parameter. I can't think of a good reason to pass something else than `numHits-start` there? Is there a use-case I'm missing?

atris:
> I'm wondering whether we should consider removing this `topDocs` variant which takes a `howMany` parameter. I can't think of a good reason to pass something else than `numHits-start` there? Is there a use-case I'm missing?

I think it was intended to allow people to "scroll" through hits and "pick" a small subset to get e.g. get the hits from 4th best to 8th best. I agree that there is no concrete use case that comes to mind though, so +1 to removing it

atris:
> > I'm wondering whether we should consider removing this `topDocs` variant which takes a `howMany` parameter. I can't think of a good reason to pass something else than `numHits-start` there? Is there a use-case I'm missing?
> 
> I think it was intended to allow people to "scroll" through hits and "pick" a small subset to get e.g. get the hits from 4th best to 8th best. I agree that there is no concrete use case that comes to mind though, so +1 to removing it

Another question is -- what is the expected behaviour when start > size? Should we error out, or return a null topdocs (as we do today), or normalize the start value to be 0?

atris:
I looked at the testGrouping failure -- looks like the test assumes that the second pass collector will always collect atleast docsInGroup hits, and thus takes a random offset there to start with. However, that is not always true, and the number of hits collected can be lesser.

I think the ideal way of usage of topDocs is to get the totalHits from the corresponding TopDocsCollector and use that as the fence to set the starting point.

jpountz:
> what is the expected behaviour when start > size? Should we error out, or return a null topdocs (as we do today), or normalize the start value to be 0?

I don't think we should error, because users have no way to know up-front that the query they ran has less than `start` hits. So let's keep returning empty hits?

lucene/core/src/java/org/apache/lucene/search/TopDocsCollector.java
```diff
@@ -136,14 +136,21 @@ public TopDocs topDocs(int start, int howMany) {
...
     int size = topDocsSize();
 
-    // Don't bother to throw an exception, just return an empty TopDocs in case
-    // the parameters are invalid or out of range.
-    // TODO: shouldn't we throw IAE if apps give bad params here so they dont
-    // have sneaky silent bugs?
-    if (start < 0 || start >= size || howMany <= 0) {
+
+    if (start < 0) {
+      throw new IllegalArgumentException("Expected value of starting position is between 0 and " + size +
```
jpountz: The message should say between "0 and $howMany" rather than size I think, since howMany is the user-supplied parameter and size depends on the number of matches, which the user has no control on?

lucene/core/src/java/org/apache/lucene/search/TopDocsCollector.java
```diff
@@ -136,14 +136,21 @@ public TopDocs topDocs(int start, int howMany) {
...
+          ", got " + start);
+    }
+
+    if (start >= size) {
       return newTopDocs(null, start);
     }
 
+    if (howMany <= 0) {
+      throw new IllegalArgumentException("Number of hits requested must be greater than 0 but value was "
+      + howMany);
```
jpountz: can you fix the indentation?

lucene/core/src/java/org/apache/lucene/search/TopDocsCollector.java
```diff
@@ -137,12 +137,12 @@ public TopDocs topDocs(int start, int howMany) {
     int size = topDocsSize();
 
 
-    if (start < 0 || start > size) {
+    if (start < 0) {
```
jpountz: I think we should still fail if start > howMany?

atris:
@jpountz Updated, please see

lucene/core/src/java/org/apache/lucene/search/TopDocsCollector.java
jpountz:
```suggestion
      throw new IllegalArgumentException("Expected value of starting position is between 0 and " + howMany +
```

lucene/core/src/java/org/apache/lucene/search/TopDocsCollector.java
```diff
@@ -136,14 +136,20 @@ public TopDocs topDocs(int start, int howMany) {
...
 
-    // Don't bother to throw an exception, just return an empty TopDocs in case
-    // the parameters are invalid or out of range.
-    // TODO: shouldn't we throw IAE if apps give bad params here so they dont
-    // have sneaky silent bugs?
-    if (start < 0 || start >= size || howMany <= 0) {
+
+    if (start < 0 || start > howMany) {
+      throw new IllegalArgumentException("Expected value of starting position is between 0 and " + size +
+          ", got " + howMany);
```
jpountz:
```suggestion
          ", got " + start);
```

lucene/core/src/java/org/apache/lucene/search/TopDocsCollector.java
```diff
@@ -136,14 +136,20 @@ public TopDocs topDocs(int start, int howMany) {
...
+    }
+
+    if (start >= size) {
       return newTopDocs(null, start);
     }
 
+    if (howMany <= 0) {
+      throw new IllegalArgumentException("Number of hits requested must be greater than 0 but value was " + howMany);
+    }
```
jpountz: can you move this up? We should validate input parameters before we do anything else?

lucene/core/src/java/org/apache/lucene/search/TopDocsCollector.java
atris: Oops, my bad, sorry

jpountz:
Could you confirm all tests pass now, including the grouping ones that failed previously?

atris:
@jpountz Thinking more about this, should we be asserting if howMany is greater than start? Would that still be true for pagination?

jpountz:
Hmm I got confused indeed, we shouldn't check that start is less than howMany since howMany is not the rank of the last document to fetch but the page size. Sorry about that.

atris:
@jpountz Updated the PR. Ran the test suite, came in clean. Also, beasted `TestGrouping` 20 times, ran clean.

jpountz:
I'm seeing test failures such as
```
17:50:42    [junit4] Suite: org.apache.lucene.search.TestDiversifiedTopDocsCollector
17:50:42    [junit4]   2> NOTE: reproduce with: ant test  -Dtestcase=TestDiversifiedTopDocsCollector -Dtests.method=testInvalidArguments -Dtests.seed=7DDC995A69B8EEE3 -Dtests.slow=true -Dtests.badapples=true -Dtests.locale=fr-DZ -Dtests.timezone=Asia/Qatar -Dtests.asserts=true -Dtests.file.encoding=UTF8
17:50:42    [junit4] ERROR   0.33s J0 | TestDiversifiedTopDocsCollector.testInvalidArguments <<<
17:50:42    [junit4]    > Throwable #1: java.lang.IllegalArgumentException: Expected value of starting position is between 0 and 5, got -1
17:50:42    [junit4]    > 	at __randomizedtesting.SeedInfo.seed([7DDC995A69B8EEE3:2C3B93B8CB053023]:0)
17:50:42    [junit4]    > 	at org.apache.lucene.search.TopDocsCollector.topDocs(TopDocsCollector.java:144)
17:50:42    [junit4]    > 	at org.apache.lucene.search.TopDocsCollector.topDocs(TopDocsCollector.java:115)
17:50:42    [junit4]    > 	at org.apache.lucene.search.TestDiversifiedTopDocsCollector.testInvalidArguments(TestDiversifiedTopDocsCollector.java:96)
17:50:42    [junit4]    > 	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
17:50:42    [junit4]    > 	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
17:50:42    [junit4]    > 	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
17:50:42    [junit4]    > 	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
17:50:42    [junit4]    > 	at java.base/java.lang.Thread.run(Thread.java:834)
17:50:42    [junit4]   2> NOTE: test params are: codec=Asserting(Lucene80): {song=PostingsFormat(name=LuceneFixedGap), year=PostingsFormat(name=LuceneFixedGap), artistName=PostingsFormat(name=Direct), id=TestBloomFilteredLucenePostings(BloomFilteringPostingsFormat(Lucene50(blocksize=128)))}, docValues:{song=DocValuesFormat(name=Asserting), weeks=DocValuesFormat(name=Asserting), artist=DocValuesFormat(name=Asserting), year=DocValuesFormat(name=Asserting), weeksAtNumberOne=DocValuesFormat(name=Asserting), artistName=DocValuesFormat(name=Lucene80), id=DocValuesFormat(name=Asserting)}, maxPointsInLeafNode=650, maxMBSortInHeap=7.269775733297731, sim=Asserting(org.apache.lucene.search.similarities.AssertingSimilarity@21758fe2), locale=fr-DZ, timezone=Asia/Qatar
17:50:42    [junit4]   2> NOTE: Linux 4.15.0-1041-gcp amd64/Oracle Corporation 11.0.2 (64-bit)/cpus=16,threads=1,free=285526208,total=536870912
17:50:42    [junit4]   2> NOTE: All tests run in this JVM: [TestInetAddressPoint, TestDocValuesStatsCollector, TestPKIndexSplitter, TestHighFreqTerms, TestDiversifiedTopDocsCollector]
17:50:42    [junit4] Completed [9/16 (1!)] on J0 in 1.54s, 4 tests, 1 error <<< FAILURES!
```

I'll look at it a bit later unless you beat me to it.

atris:
> TestDiversifiedTopDocsCollector

Thanks for taking a look. The test failure is not really a code issue but a test issue in `TestDiversifiedTopDocsCollector#testInvalidArguments` where it needed updating to expect exceptions and error messages. Fixed the same and raised #858 , please take a look

PR #770, opened 2019-07-08
LUCENE-8746: Component2D topology library that works on encoded space

iverase:
With the upcoming of a new Shape type working in cartesian space (#726), I think we need to put some structure in the objects that contain spatial logic. In particular I have tried to remove all the mixed notation between latitude/longitude and x/y as well as defined factory methods to create those shapes from LatLonShape. 

This library chooses to use X/Y notation as it is mainly cartesian, it works on the encoding space and solves problems like the neighbourhood issue (https://discuss.elastic.co/t/neighboring-touching-geo-shapes-not-found/175543) when not encoded query shapes are used against encoded indexed shapes. It potentially can simplify all the query logic as it is only needed a query by Component2D for this case.

Currently it contains factory methods to create Component2D shapes from LatLonShapes, it should be trivial to add a factory class for XYShapes.

@jpountz @nknize @rmuir @dsmiley  let me know what do you think?


iverase:
After merging master with XYShape, it seems clear that this approach won't work for that encoding as it is not linear and shapes in the encoded space have different properties.

iverase:
Closing this PR. I am opening a new one with narrow scope

PR #772, opened 2019-07-10
Payload-stored position length intervals

romseygeek:
An experimental token filter and intervals source that use payloads to encode
token lengths

### DIFF
--- a/lucene/sandbox/src/java/org/apache/lucene/payloads/PayloadLengthTermIntervalsSource.java
+++ b/lucene/sandbox/src/java/org/apache/lucene/payloads/PayloadLengthTermIntervalsSource.java
@@ -1,0 +1,299 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.lucene.payloads;
+
+import java.io.IOException;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Objects;
+
+import org.apache.lucene.index.LeafReaderContext;
+import org.apache.lucene.index.PostingsEnum;
+import org.apache.lucene.index.Term;
+import org.apache.lucene.index.Terms;
+import org.apache.lucene.index.TermsEnum;
+import org.apache.lucene.queries.intervals.IntervalIterator;
+import org.apache.lucene.queries.intervals.IntervalQuery;
+import org.apache.lucene.queries.intervals.IntervalsSource;
+import org.apache.lucene.search.MatchesIterator;
+import org.apache.lucene.search.Query;
+import org.apache.lucene.search.QueryVisitor;
+import org.apache.lucene.util.BytesRef;
+
+/**
+ * An IntervalsSource that returns intervals over single terms, reading token
+ * length from each term's associated payload
+ *
+ * Tokens should be indexed using {@link PayloadTokenLengthFilter} to write
+ * token lengths into payloads
+ */
+public class PayloadLengthTermIntervalsSource extends IntervalsSource {
+
+  private final BytesRef term;
+
+  /**
+   * Create a PayloadLengthTermIntervalsSource for the given term
+   */
+  public PayloadLengthTermIntervalsSource(BytesRef term) {
+    this.term = term;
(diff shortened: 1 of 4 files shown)


lucene/sandbox/src/java/org/apache/lucene/payloads/PayloadLengthTermIntervalsSource.java
```diff
@@ -0,0 +1,299 @@
...
+    }
+    TermsEnum te = terms.iterator();
+    if (te.seekExact(term) == false) {
+      return null;
+    }
+    return matches(te, doc);
+  }
+
+  private static MatchesIterator matches(TermsEnum te, int doc) throws IOException {
+    PostingsEnum pe = te.postings(null, PostingsEnum.ALL);
```
jpountz: could this be PostingsEnum.PAYLOADS, you don't seem to need offsets?

lucene/sandbox/src/java/org/apache/lucene/payloads/PayloadLengthTermIntervalsSource.java
```diff
@@ -0,0 +1,299 @@
...
+        }
+        else {
+          upto = pe.freq();
+          pos = -1;
+        }
+      }
+
+      @Override
+      public String toString() {
+        return term.utf8ToString() + ":" + super.toString();
```
jpountz: Is the term guaranteed to be a UTF8-encoded string?

lucene/sandbox/src/java/org/apache/lucene/payloads/PayloadTokenLengthFilter.java
```diff
@@ -0,0 +1,71 @@
...
+import org.apache.lucene.analysis.tokenattributes.PayloadAttribute;
+import org.apache.lucene.analysis.tokenattributes.PositionLengthAttribute;
+import org.apache.lucene.util.BytesRef;
+
+/**
+ * Encodes a token's position length into its associated payload
+ *
+ * @see PayloadLengthTermIntervalsSource
+ */
+public final class PayloadTokenLengthFilter extends TokenFilter {
```
jpountz: Create a factory for it as well?

jpountz (approved):

magibney:
Would you want to consider reordering tokens in PositionLengthTokenFilter, so that query implementations can expect a reliable ordering of tokens from the index? I think this would affect the order of terms recorded in the index only for "prefix phrases", as might be generated with a synonym such as "toaster"=="toaster oven"; but if we're recording position length in the index, it's good to have a defined ordering that reflects this new structural information.
Reordering would be similar to that done by FlattenGraphFilter; for reference, an example that reorders for position length (and records position length to Payloads) can be found [here](https://github.com/magibney/lucene-solr/blob/LUCENE-7398/master/lucene/core/src/java/org/apache/lucene/analysis/core/PositionLengthOrderTokenFilter.java).

lucene/sandbox/src/java/org/apache/lucene/payloads/PayloadLengthTermIntervalsSource.java
romseygeek: We need offsets for MatchesIterator

lucene/sandbox/src/java/org/apache/lucene/payloads/PayloadLengthTermIntervalsSource.java
romseygeek: It's not, and this applies to `TermIntervalsSource` as well.  TermQuery uses `term.text()` here, I'll change to use that and open a separate issue for the existing sources.

lucene/sandbox/src/java/org/apache/lucene/payloads/PayloadTokenLengthFilter.java
romseygeek: ++

uschindler (approved):
Looks good.
How about adding the length attribute to a future index format. I agree it's a separate issue, but if this works well now, that would be the next logical step.

jpountz:
@uschindler This proved controversial on https://issues.apache.org/jira/browse/LUCENE-4312 so we thought we could explore without adding to the index format by storing lengths in payloads for now.

romseygeek:
Re-ordering raises an interesting problem, which is that we immediately hit LUCENE-7398 again - shared-prefix synonyms start breaking, because a token of length 2 will be minimized away.  So for this to be useable, I think we also need @magibney's space-searching changes; still doable as a sandbox add-on, but more complicated than just these two classes.  Essentially we need to re-implement phrase and ordered-near interval functions to do some read-ahead.

magibney:
Agreed, @romseygeek; just to clarify, re-ordering may _raise_ the problem, but doesn't _introduce_ it. This is indeed a variant of the issue underlying [LUCENE-7398](https://issues.apache.org/jira/browse/LUCENE-7398) (focused on Spans); the same issue also underlies [LUCENE-8477](https://issues.apache.org/jira/browse/LUCENE-8477) (for intervals), except that in the latter case (where variable-length subclauses are the result of a disjunction), variable length behavior can be predicted, and pre-emptively "pulled out" to the top level (as in the patch to address LUCENE-8477). In this case (with position length recorded in the index), variable-length behavior of subclauses would be unknown until positions are actually read from the index.

rgoene:
Hi all, is there any chance this might get merged somewhere soon? It seems to be an innocent merge, right?

itygh:

PR #776, opened 2019-07-11
LUCENE-8912: remove nori/tools dependency on ICU

danmuzi:
It is the Nori version of LUCENE-8866.
Nori doesn't need the ICU library because it uses Normalizer2 only for NFKC normalization like Kuromoji.

Please check the following link:
https://issues.apache.org/jira/browse/LUCENE-8912


danmuzi:
That's okay. @jimczi 
Thank you very much for your review :D
I'll merge this PR and close.

PR #779, opened 2019-07-12
LUCENE-8762: Introduce Specialized Impacts For Doc + Freq

atris:
ant test for both Lucene and Solr pass. ant precommit passes as well.

I did not add new tests since multiple existing tests exercise the new functionality

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/codecs/lucene50/Lucene50PostingsReader.java
+++ b/lucene/core/src/java/org/apache/lucene/codecs/lucene50/Lucene50PostingsReader.java
@@ -234,7 +234,9 @@ public final class Lucene50PostingsReader extends PostingsReaderBase {
     final boolean indexHasOffsets = fieldInfo.getIndexOptions().compareTo(IndexOptions.DOCS_AND_FREQS_AND_POSITIONS_AND_OFFSETS) >= 0;
     final boolean indexHasPayloads = fieldInfo.hasPayloads();
 
-    if (indexHasPositions &&
+   if (indexHasPositions == false || PostingsEnum.featureRequested(flags, PostingsEnum.POSITIONS) == false) {
+      return new BlockImpactsDocsEnum(fieldInfo, (IntBlockTermState) state);
+    } else if (indexHasPositions &&
         PostingsEnum.featureRequested(flags, PostingsEnum.POSITIONS) &&
         (indexHasOffsets == false || PostingsEnum.featureRequested(flags, PostingsEnum.OFFSETS) == false) &&
         (indexHasPayloads == false || PostingsEnum.featureRequested(flags, PostingsEnum.PAYLOADS) == false)) {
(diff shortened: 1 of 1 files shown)


atris:
cc: @jpountz

atris:
Any thoughts on this one? I am hoping this will bring gains to our nightly benchmarks :)

lucene/core/src/java/org/apache/lucene/codecs/lucene50/Lucene50PostingsReader.java
```diff
@@ -234,7 +234,9 @@ public ImpactsEnum impacts(FieldInfo fieldInfo, BlockTermState state, int flags)
     final boolean indexHasOffsets = fieldInfo.getIndexOptions().compareTo(IndexOptions.DOCS_AND_FREQS_AND_POSITIONS_AND_OFFSETS) >= 0;
     final boolean indexHasPayloads = fieldInfo.hasPayloads();
 
-    if (indexHasPositions &&
+   if (indexHasPositions == false || PostingsEnum.featureRequested(flags, PostingsEnum.POSITIONS) == false) {
+      return new BlockImpactsDocsEnum(fieldInfo, (IntBlockTermState) state);
+    } else if (indexHasPositions &&
```
jpountz: indentation looks wrong?

lucene/core/src/java/org/apache/lucene/codecs/lucene50/Lucene50PostingsReader.java
```diff
@@ -1761,6 +1763,198 @@ public long cost() {
...
+      return doc;
+    }
+
+    private void refillDocs() throws IOException {
+      final int left = docFreq - docUpto;
+      assert left > 0;
+
+      if (left >= BLOCK_SIZE) {
+        forUtil.readBlock(docIn, encoded, docDeltaBuffer);
+        if (indexHasFreq) {
```
jpountz: can you read freqs lazily like BlockDocsEnum?

atris:
@jpountz I have fixed your comments, please let me know if they look fine.

luceneutil is running right now with 10k documents, will post results ASAP

atris:
@jpountz Three Luceneutil runs with wikimedium2m:

https://gist.github.com/atris/5057e3372b287873a4e840c5f4e65a0d
https://gist.github.com/atris/76719423071662d5d25eed4bdb7bf4a3
https://gist.github.com/atris/dd7b7336147da138ac340670c81e5452

jpountz:
I don't expect your last commit to help since impacts should always be requested with frequencies. What I had in mind when mentioning loading frequencies lazily was to do something similar to what was done in #595 to only read and decode the block of frequencies if freq() is called on the current block.

atris:
Ah, yes, that makes sense. I did a couple of more runs with 10m docs and saw similar results.

Will update the PR

atris:
@jpountz Updated the PR with lazy loading. Please let me know if it looks fine

atris:
luceneutil run:

https://gist.github.com/atris/1b2c25021ca11138338aa73efde2aa38

wikimedium2m

lucene/core/src/java/org/apache/lucene/codecs/lucene50/Lucene50PostingsReader.java
```diff
@@ -1761,6 +1763,223 @@ public long cost() {
...
+    final boolean indexHasPayloads;
+    final boolean indexHasFreq;
+
+    private int docFreq;                              // number of docs in this posting list
+    private int docUpto;                              // how many docs we've read
+    private int doc;                                  // doc we last read
+    private int accum;                                // accumulator for doc deltas
+    private int freq;                                 // freq we last read
+
+    private boolean needsFreq; // true if the caller actually needs frequencies
```
jpountz: This will always be true, it is illegal to read impacts and not ask for term frequencies (it doesn't make much sense).

jpountz:
I ran luceneutil on this patch on my machine too and I'm getting similar results. It's a bit disappointing, I was expecting some gains though I'm rather happy that we can keep things easier to maintain by not having to sepcialize too much. Sorry for the time you spent on it but I think it's better to keep things the way they are today?

atris:
> I ran luceneutil on this patch on my machine too and I'm getting similar results. It's a bit disappointing, I was expecting some gains though I'm rather happy that we can keep things easier to maintain by not having to sepcialize too much. Sorry for the time you spent on it but I think it's better to keep things the way they are today?

No sweat, I understand. Makes sense to keep things straightforward!

PR #780, opened 2019-07-12
SOLR-11866: Support efficient subset matching in query elevation rules

bruno-roustant:
New SubsetMatchElevationProvider and TrieSubsetMatcher that can elevate documents based on query subset matching very efficiently scalable to a large number of elevation rules.

### DIFF
--- a/solr/core/src/java/org/apache/solr/handler/component/QueryElevationComponent.java
+++ b/solr/core/src/java/org/apache/solr/handler/component/QueryElevationComponent.java
@@ -25,26 +25,33 @@ import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.lang.invoke.MethodHandles;
+import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.NoSuchElementException;
+import java.util.Queue;
 import java.util.Set;
+import java.util.SortedSet;
 import java.util.WeakHashMap;
 
 import com.carrotsearch.hppc.IntIntHashMap;
 import com.carrotsearch.hppc.cursors.IntIntCursor;
 import com.google.common.annotations.VisibleForTesting;
-import com.google.common.base.Throwables;
 import com.google.common.collect.Collections2;
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
-import com.google.common.collect.Maps;
+import com.google.common.collect.ImmutableSortedSet;
+import com.google.common.collect.ObjectArrays;
 import org.apache.lucene.analysis.Analyzer;
 import org.apache.lucene.analysis.TokenStream;
 import org.apache.lucene.analysis.tokenattributes.CharTermAttribute;
@@ -208,7 +215,6 @@ public class QueryElevationComponent extends SearchComponent implements SolrCore
    * @param core The core holding this component.
    * @return The number of elevation rules parsed.
    */
-  @SuppressWarnings("WeakerAccess")
   protected int loadElevationConfiguration(SolrCore core) throws Exception {
     synchronized (elevationProviderCache) {
       elevationProviderCache.clear();
@@ -376,12 +382,10 @@ public class QueryElevationComponent extends SearchComponent implements SolrCore
   /**
    * Loads the {@link ElevationProvider}.
    *
-   * @throws java.io.IOException      If an I/O error occurs while analyzing the triggering queries.
    * @throws RuntimeException If the config does not provide an XML content of the expected format
    *                          (either {@link RuntimeException} or {@link org.apache.solr.common.SolrException}).
    */
-  @SuppressWarnings("WeakerAccess")
-  protected ElevationProvider loadElevationProvider(XmlConfigFile config) throws IOException {
+  protected ElevationProvider loadElevationProvider(XmlConfigFile config) {
     Map<ElevatingQuery, ElevationBuilder> elevationBuilderMap = new LinkedHashMap<>();
     XPath xpath = XPathFactory.newInstance().newXPath();
     NodeList nodes = (NodeList) config.evaluate("elevate/query", XPathConstants.NODESET);
@@ -389,7 +393,7 @@ public class QueryElevationComponent extends SearchComponent implements SolrCore
       Node node = nodes.item(i);
       String queryString = DOMUtil.getAttr(node, "text", "missing query 'text'");
       String matchString = DOMUtil.getAttr(node, "match");
-      ElevatingQuery elevatingQuery = new ElevatingQuery(queryString, parseMatchPolicy(matchString));
+      ElevatingQuery elevatingQuery = new ElevatingQuery(queryString, isSubsetMatchPolicy(matchString));
 
       NodeList children;
       try {
@@ -425,10 +429,10 @@ public class QueryElevationComponent extends SearchComponent implements SolrCore
         previousElevationBuilder.merge(elevationBuilder);
       }
     }
-    return createElevationProvider(queryAnalyzer, elevationBuilderMap);
+    return createElevationProvider(elevationBuilderMap);
   }
 
-  private boolean parseMatchPolicy(String matchString) {
+  protected boolean isSubsetMatchPolicy(String matchString) {
     if (matchString == null) {
       return DEFAULT_SUBSET_MATCH;
     } else if (matchString.equalsIgnoreCase("exact")) {
@@ -591,7 +595,7 @@ public class QueryElevationComponent extends SearchComponent implements SolrCore
       fields.add(currentFields.get(i));
     }
     return modify ?
-            new SortSpec(new Sort(sorts.toArray(new SortField[sorts.size()])),
+            new SortSpec(new Sort(sorts.toArray(new SortField[0])),
                     fields,
                     current.getCount(),
                     current.getOffset())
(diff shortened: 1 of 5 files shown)


solr/core/src/java/org/apache/solr/handler/component/QueryElevationComponent.java
```diff
@@ -834,48 +850,77 @@ public int size() {
...
-      elevationMap = buildElevationMap(elevationBuilderMap);
-    }
-
-    private Map<String, Elevation> buildElevationMap(Map<ElevatingQuery, ElevationBuilder> elevationBuilderMap) {
-      Map<String, Elevation> elevationMap = Maps.newHashMapWithExpectedSize(elevationBuilderMap.size());
+    /**
+     * @param subsetMatcherBuilder The {@link SubsetMatcher.Builder} to build the {@link SubsetMatcher}.
+     * @param elevationBuilderMap The map of elevation rules.
+     */
+    protected SubsetMatchElevationProvider(SubsetMatcher.Builder<String, Elevation> subsetMatcherBuilder,
```
dsmiley: This is the only ElevationProvider that Solr provides.  So I think we should just call it DefaultElevationProvider.   By using the name SubsetMatchElevationProvider it suggests it only does this when it does exact as well.  It's not like there are separate impls.

solr/core/src/java/org/apache/solr/handler/component/QueryElevationComponent.java
```diff
@@ -1131,4 +1227,333 @@ public int compareTop(int doc) {
...
+   * That is, finds all subsets <code>n</code> in <code>N</code> for which <code>s.containsAll(n)</code>
+   * (<code>s</code> contains all the elements of <code>n</code>, in any order).
+   * <p>
+   * Associates a match value of type &lt;M&gt; to each subset and provides it each time the subset matches (i.e. is
+   * contained by the provided set).
+   *
+   * @param <E> Subset element type.
+   * @param <M> Subset match value type.
+   */
+  protected interface SubsetMatcher<E, M> {
```
dsmiley: I don't think there's a point to this interface (with builder) as separate from implementation since there is one and it's highly theoretical to imagine another user provided one.  So lets combine this with the implementation below, keeping the javadocs.  I suppose the name TrieSubsetMatcher is nicest to me but I don't care.

solr/core/src/java/org/apache/solr/handler/component/QueryElevationComponent.java
```diff
@@ -1131,4 +1227,333 @@ public int compareTop(int doc) {
...
+   * <p>
+   * The worst case time complexity of the subset matching is <code>O(2^s)</code>, however a more typical case time
+   * complexity is <code>O(s^3)</code> where s is the size of the set to partially match.
+   * Note it does not depend on <code>N</code>, the size of the collection of subsets, nor on <code>n</code>, the size of
+   * a subset.
+   */
+  protected static class TrieSubsetMatcher<E extends Comparable<? super E>, M> implements SubsetMatcher<E, M> {
+
+      /*
+      Trie structure:
```
dsmiley: I love all these comments!

solr/core/src/java/org/apache/solr/handler/component/QueryElevationComponent.java
```diff
@@ -425,10 +429,10 @@ protected ElevationProvider loadElevationProvider(XmlConfigFile config) throws I
         previousElevationBuilder.merge(elevationBuilder);
       }
     }
-    return createElevationProvider(queryAnalyzer, elevationBuilderMap);
+    return createElevationProvider(elevationBuilderMap);
```
dsmiley: Just checking: you did this because queryAnalyzer is in scope to caller so no need to pass as param?

solr/core/src/java/org/apache/solr/handler/component/QueryElevationComponent.java
```diff
@@ -682,35 +686,48 @@ public String getDescription() {
...
+    return concatenatedTerms.toString();
+  }
+
+  /**
+   * Analyzes the provided query string, tokenizes the terms and add them to either the provided {@link Collection} or {@link Appendable}.
+   *
+   * @param queryTerms The {@link Collection} that receives the terms; or null if none.
+   * @param concatenatedTerms The {@link Appendable} that receives the terms; or null if none.
+   */
+  protected void analyzeQuery(String query, Collection<String> queryTerms, Appendable concatenatedTerms) {
```
dsmiley: This method's contract is a little awkward in that it has dual optional output methods, but it's okay.  Maybe it's be nicer if it took a Consumer<CharSequence> since the set addition and appending are trivial for the caller to do?  Up to you; just an idea.

solr/core/src/java/org/apache/solr/handler/component/QueryElevationComponent.java
```diff
@@ -1020,8 +1066,58 @@ public Elevation(Set<BytesRef> elevatedIds, Set<BytesRef> excludedIds,
...
+     *         is <code>null</code>.
+     */
+    protected Elevation mergeWith(Elevation elevation) {
+      if (elevation == null) {
+        return this;
+      }
+      Set<BytesRef> elevatedIds = ImmutableSet.<BytesRef>builder().addAll(this.elevatedIds).addAll(elevation.elevatedIds).build();
+      boolean overlappingElevatedIds = elevatedIds.size() != (this.elevatedIds.size() + elevation.elevatedIds.size());
+      BooleanQuery.Builder includeQueryBuilder = new BooleanQuery.Builder();
+      Set<BooleanClause> clauseSet = (overlappingElevatedIds ? new HashSet<>() : null);
```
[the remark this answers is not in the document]
bruno-roustant: I'll initialize with Sets.newHashSetWithExpectedSize(elevatedIds.size()).

solr/core/src/java/org/apache/solr/handler/component/QueryElevationComponent.java
bruno-roustant: This is the trace of a previous version that had two SubsetMatcher. I remove it.

solr/core/src/java/org/apache/solr/handler/component/QueryElevationComponent.java
bruno-roustant: Yes

solr/core/src/java/org/apache/solr/handler/component/QueryElevationComponent.java
bruno-roustant: I pondered for a while on this signature, because I wanted it to be clear but also performant where it is used, to avoid the creation of lots of lambdas in loops.
I'll change back to the Consumer. It will create just an additional lambda instance per query to process, given that I'll declare the lambdas out of the loops.

solr/core/src/java/org/apache/solr/handler/component/QueryElevationComponent.java
```diff
@@ -701,28 +703,20 @@ protected ElevationProvider createElevationProvider(Map<ElevatingQuery, Elevatio
...
       tokens.reset();
       CharTermAttribute termAtt = tokens.addAttribute(CharTermAttribute.class);
       while (tokens.incrementToken()) {
-        if (queryTerms != null) {
-          queryTerms.add(termAtt.toString());
-        }
-        if (concatenatedTerms != null) {
-          concatenatedTerms.append(termAtt);
-        }
+        termsConsumer.accept(termAtt.toString());
```
dsmiley: the termAtt.toString() is a bit of a shame since as a CharSequence, the string appending consumer needn't get the String.  But it's not a big deal.

solr/core/src/java/org/apache/solr/handler/component/QueryElevationComponent.java
```diff
@@ -857,31 +851,33 @@ public int size() {
...
      * @param elevationBuilderMap The map of elevation rules.
      */
-    protected SubsetMatchElevationProvider(SubsetMatcher.Builder<String, Elevation> subsetMatcherBuilder,
-                                           Map<ElevatingQuery, ElevationBuilder> elevationBuilderMap) {
+    protected DefaultElevationProvider(TrieSubsetMatcher.Builder<String, Elevation> subsetMatcherBuilder,
+                                       Map<ElevatingQuery, ElevationBuilder> elevationBuilderMap) {
       exactMatchElevationMap = new LinkedHashMap<>();
       Collection<String> queryTerms = new ArrayList<>();
-      StringBuilder concatenatedTerms = new StringBuilder();
+      Consumer<String> termsConsumer = queryTerms::add;
```
dsmiley: Is pulling the method reference outside the loop actually more efficient?  Please point to info showing this.  I'm kinda sad if it is... I like the clean/simplicity of it used where needed.  @uschindler I have a feeling you know

solr/core/src/java/org/apache/solr/handler/component/QueryElevationComponent.java
```diff
@@ -896,13 +892,18 @@ public Elevation getElevationForQuery(String queryString) {
...
         return exactMatchElevationMap.get(analyzeQuery(queryString));
       }
-      StringBuilder concatenatedTerms = hasExactMatchElevationRules ? new StringBuilder() : null;
       Collection<String> queryTerms = new ArrayList<>();
-      analyzeQuery(queryString, queryTerms, concatenatedTerms);
+      Consumer<String> termsConsumer = queryTerms::add;
+      StringBuilder concatTerms = null;
+      if (hasExactMatchElevationRules) {
+        concatTerms = new StringBuilder();
+        termsConsumer = termsConsumer.andThen(concatTerms::append);
```
dsmiley: Ha!  Love it

solr/solr-ref-guide/src/the-query-elevation-component.adoc
```diff
@@ -93,11 +93,19 @@ Elevated query results can be configured in an external XML file specified in th
...
+    <doc id="11" />
+  </query>
 </elevate>
 ----
 
 In this example, the query "foo bar" would first return documents 1, 2 and 3, then whatever normally appears for the same query. For the query "ipod", it would first return "MA147LL/A", and would make sure that "IW-02" is not in the result set.
 
+Notice the `match` parameter with the value "subset" for the third rule. A query "bill bar foo" would trigger this rule because the rule defines a subset of terms to appear in the query, in any order. This query would elevate document 11 on top.
+The `match` parameter accepts either "exact" (by default) or "subset" values.
+Subset matching is scalable, one can add many rules with the match="subset" parameter.
```
dsmiley: the match="subset" needs code style.

solr/core/src/java/org/apache/solr/handler/component/QueryElevationComponent.java
bruno-roustant: I'll change to a Consumer<CharSequence>.

solr/core/src/java/org/apache/solr/handler/component/QueryElevationComponent.java
bruno-roustant: Yes it makes a difference. The lambda instance is created once, compared to inside the loop where a lambda instance is created for each step of the loop because the collection is not static.

PR #784, opened 2019-07-15
LUCENE-8810: Honor MaxClausesCount in BooleanQuery During Flattening

atris:
BooleanQuery will always try to flatten nested clauses during
rewrite. However, this can cause the maximum number of clauses
to be violated by the new query. This commit disables flattening
in the specific case.

### DIFF
--- a/lucene/core/src/test/org/apache/lucene/search/TestBooleanQuery.java
+++ b/lucene/core/src/test/org/apache/lucene/search/TestBooleanQuery.java
@@ -73,6 +73,21 @@ public class TestBooleanQuery extends LuceneTestCase {
     assertEquals(bq1.build(), bq2.build());
   }
 
+  public void testFlattenInnerDisjunctionsWithMoreThan1024Terms() throws IOException {
+    IndexSearcher searcher = newSearcher(new MultiReader());
+
+    BooleanQuery.Builder builder1024 = new BooleanQuery.Builder();
+    for(int i = 0; i < 1024; i++) {
+      builder1024.add(new TermQuery(new Term("foo", "bar-" + i)), Occur.SHOULD);
+    }
+    Query inner = builder1024.build();
+    Query query = new BooleanQuery.Builder()
+        .add(inner, Occur.SHOULD)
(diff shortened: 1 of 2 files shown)


atris:
cc @jpountz 

ant test for both Lucene and Solr pass, however, the existing Max Clause Limits test introduced in LUCENE-8811 fails as expected since there is no exception thrown anymore. However, for branches 8_2 and 8_x, the revert should fix the failure.

lucene/core/src/test/org/apache/lucene/search/TestBooleanQuery.java
```diff
@@ -73,6 +73,21 @@ public void testEquality() throws Exception {
...
+    BooleanQuery.Builder builder1024 = new BooleanQuery.Builder();
+    for(int i = 0; i < 1024; i++) {
+      builder1024.add(new TermQuery(new Term("foo", "bar-" + i)), Occur.SHOULD);
+    }
+    Query inner = builder1024.build();
+    Query query = new BooleanQuery.Builder()
+        .add(inner, Occur.SHOULD)
+        .add(new TermQuery(new Term("foo", "baz")), Occur.SHOULD)
+        .build();
+    searcher.rewrite(query);
```
jpountz: can you make it `assertSame(query, searcher.rewrite(query))`?

lucene/core/src/test/org/apache/lucene/search/TestBooleanQuery.java
```diff
@@ -73,6 +73,21 @@ public void testEquality() throws Exception {
     assertEquals(bq1.build(), bq2.build());
   }
 
+  public void testFlattenInnerDisjunctionsWithMoreThan1024Terms() throws IOException {
```
jpountz: move this test to `TestBooleanRewrites` where other rewrite-related tests are?

jpountz:
> Can you merge master back into your branch so that it sees the revert of LUCENE-8811 as well?

Actually can you re-create this pull request so that it targets branch_8x instead? I don't think we need to have this in master.

atris:
@jpountz Thanks for your comments. I have opened #787 which fixes your comments and is against 8x

PR #785, opened 2019-07-15
LUCENE-8918: Disallow null terms in PhraseQuery

romseygeek:
It is currently possible to create a PhraseQuery consisting of a single `null` term,
which will then cause NullPointerExceptions deep inside postings decoding at
search time.  We should instead catch this up-front, and prevent construction of
the Query.


romseygeek:
Merged

PR #786, opened 2019-07-15
LUCENE-8916: GraphTokenStreamFiniteStrings preserves all incoming attributes

romseygeek:
Instead of cloning terms and position increments, we keep track of all incoming
attributes and replay them from `getFiniteStrings()`.

Also incorporates the changes from LUCENE-8644, while keeping the old
`newSynonymQuery()` method signature for backwards-compatibility.


lucene/core/src/test/org/apache/lucene/util/graph/TestGraphTokenStreamFiniteStrings.java
```diff
@@ -44,13 +48,13 @@ private void assertTokenStream(TokenStream ts, String[] terms, int[] increments)
...
     assertEquals(terms.length, increments.length);
-    BytesTermAttribute termAtt = ts.getAttribute(BytesTermAttribute.class);
+    CharTermAttribute termAtt = ts.getAttribute(CharTermAttribute.class);
     PositionIncrementAttribute incrAtt = ts.getAttribute(PositionIncrementAttribute.class);
     int offset = 0;
     while (ts.incrementToken()) {
       // verify term and increment
       assert offset < terms.length;
-      assertEquals(terms[offset], termAtt.getBytesRef().utf8ToString());
+      assertEquals(terms[offset], termAtt.toString());
```
jpountz: Was this change necessary, it should be equivalent right?

romseygeek: It needs to change from `BytesTermAttribute` to something else - I could change it to a `TermToBytesRefAttribute`, but `CharTermAttribute` seemed simpler and involves less byte-wrangling

jpountz: ok

romseygeek:
I decided to try and keep the API changes minimal, so I've preserved the old `getTerms()` method on `GraphTokenStreamFiniteStrings` and backed out changes to `QueryBuilder` - we can look at those in LUCENE-8644 separately.

PR #787, opened 2019-07-15
LUCENE-8810: Honor MaxClausesCount in BooleanQuery

atris:
During Flattening, BooleanQuery will always try to flatten
nested clauses during rewrite. However, this can cause the
maximum number of clauses to be violated by the new query.
This commit disables flattening in the specific case.

Supersedes #784


atris:
Thanks for merging @jpountz !

PR #788, opened 2019-07-15
LUCENE-8920: disable FST direct-addressing pending size reduction

msokolov:
Minimal change to disable FST direct addressing


jpountz:
You might need to move CHANGES entries as well?

PR #794, opened 2019-07-18
LUCENE-8769: Introduce Range Query Type With Multiple Ranges

atris:
Currently, multiple ranges need to be specified in different
PointRangeQueries, thus leading to performance implications when
the BKD tree is deep, since each range query will need a traversal.

This commit introduces a new range query type which has multiple
ranges logically connected. All ranges are logically connected
by OR operators.


atris:
cc @jpountz @iverase

iverase:
This approach is ok when the number of ranges is low but I wonder if it would be worthy to order the ranges on an interval tree (similar to EdgeTree) so we can skip quickly ranges that do not match the inner nodes.

atris:
> This approach is ok when the number of ranges is low but I wonder if it would be worthy to order the ranges on an interval tree (similar to EdgeTree) so we can skip quickly ranges that do not match the inner nodes.

Do you feel that can be done as a follow up PR, since this PR is large enough in itself, and the ordering can probably be built on top of this one?

iverase:
Ok, that is fine.

I think we should follow the structure hat we have for other queries, e.g. Add in {X}Point fields a new method like:

```
public static Query newMultiRangeQuery(String field, {X}[] lowerValues, {X}[] upperValues) {
    //Call query with packed values
 }
```

In this case we do not have to change the visibility there pack method.

atris:
@iverase I was hesitant to do that since MultiRangeQuery will take multiple points as inputs, so if we were to build the entire query in one go, that would require the user to pass in an array of arrays upfront. The builder approach, much like BooleanQuery, gives the user a friendly option to populate ranges and then build a query. WDYT?

atris:
I ran beaster 20 times on the new tests -- came clean.

@iverase @jpountz Please let me know if there are any thoughts on this.

atris:
Any comments on this one, please? Happy to iterate

jpountz:
It's probably fine for sandbox. Can you add TODOs about handling overlapping ranges at rewrite time, and organizing query ranges in a tree structure similarly to EdgeTree so that we don't have to walk query ranges linearly when running the query?

atris:
@jpountz updated the same.

ant precommit looks fine

atris:
@jpountz Updated the same

atris:
@jpountz I ran recommit on the latest version again and rebased with master -- all came in clean

atris:
Rebased per master.

Could we merge this, if no objections?

atris:
Gentle ping :)

lucene/CHANGES.txt
```diff
@@ -40,6 +40,12 @@ Improvements
   docs on equal scores. Also, remove the ability of TopDocs.merge to set shard
   indices (Atri Sharma, Adrien Grand, Simon Willnauer)
 
+* LUCENE-8874: Show SPI names instead of class names in Luke Analysis tab. (Tomoko Uchida)
```
jimczi: This should be removed ?

lucene/CHANGES.txt
```diff
@@ -40,6 +40,12 @@ Improvements
...
+
+* LUCENE-8894: Add APIs to find SPI names for Tokenizer/CharFilter/TokenFilter factory classes. (Tomoko Uchida)
```
jimczi: Same here ?

atris: @jimczi Thanks for pointing it out, was a conflict merge error. Fixed now

atris:
@jimczi Thanks for approving. Does this look merge ready now?

jimczi:
> Does this look merge ready now?

I think so, yes

atris:
> > Does this look merge ready now?
> 
> I think so, yes

Thanks, mind merging it, please? :)

jimczi:
I merged in master and will now merge in branch 8x. Apache Jira is not accessible for me at the moment so I'll update the issue later. Thanks @atris !

atris:
> I merged in master and will now merge in branch 8x. Apache Jira is not accessible for me at the moment so I'll update the issue later. Thanks @atris !

Thanks @jimczi

PR #796, opened 2019-07-19
LUCENE-8927: Set.copyOf and Set.of instead of Collections.unmodifiabl…

atris:


atris:
cc @jpountz

lucene/benchmark/src/java/org/apache/lucene/benchmark/byTask/feeds/DemoHTMLParser.java
```diff
@@ -147,7 +147,7 @@ public InputSource resolveEntity(String publicId, String systemId) {
     }
     
     private static final Set<String> createElementNameSet(String... names) {
-      return Collections.unmodifiableSet(new HashSet<>(Arrays.asList(names)));
+      return Set.copyOf(Arrays.asList(names));
```
jpountz: `Set.of(names)`?

lucene/core/src/java/org/apache/lucene/index/FieldInfos.java
```diff
@@ -422,6 +422,10 @@ synchronized boolean contains(String fieldName, DocValuesType dvType) {
       }
     }
     
+    synchronized Set<String> getFieldNames() {
+      return Set.copyOf(nameToNumber.keySet());
+    }
```
jpountz: we should not resurrect this method that we just removed

lucene/benchmark/src/java/org/apache/lucene/benchmark/byTask/feeds/DemoHTMLParser.java
jpountz: actually we should probably remove this method and fix call sites to use `Set.of` instead?

atris:
@jpountz Thanks, updated the PR

jpountz:
I think you forgot to remove `createElementNameSet`?

atris:
> I think you forgot to remove `createElementNameSet`?

Oops, sorry. Updated

munendrasn:
@atris  @jpountz Is it possible to extend the scope to remove all *Collections.un\** ??
Also, can we extend this change to the Solr project too?

atris:
Solr changes are coming in a separate PR since I did not want to overlap and amplify things in a single PR.

We can explore further changes in subsequent PRs.

PR #797, opened 2019-07-19
LUCENE-8921: IndexSearcher.termStatistics requires docFreq totalTermFreq instead of TermStates

bruno-roustant:


lucene/core/src/java/org/apache/lucene/search/IndexSearcher.java
```diff
@@ -861,17 +861,17 @@ public String toString() {
...
+   * the term does not exist (if docFreq == 0).
    * 
    * This can be overridden for example, to return a term's statistics
    * across a distributed collection.
    * @lucene.experimental
    */
-  public TermStatistics termStatistics(Term term, TermStates context) throws IOException {
-    if (context.docFreq() == 0) {
+  public TermStatistics termStatistics(Term term, int docFreq, long totalTermFreq) throws IOException {
+    if (docFreq == 0) {
```
romseygeek: Given that the caller will always know whether or not `docFreq == 0`, maybe we should move this check up a level and make this throw an Exception if somebody passes in an empty docFreq?

bruno-roustant: Good point. I'll look into this.

dsmiley: Using git-blame, I see this check was added in https://issues.apache.org/jira/browse/LUCENE-8020   From my reading of comments there, it appears to only occur for SpanQuery classes and TermAutomatonQuery and that there are some tests for this.

dsmiley: I bring this up because I don't think we should simply add this extra work for _all_ callers (as I see you did in the last commit Bruno); it's only unusual cases where this might happen.  You might first remove them all and tell us what tests reveal which queries need this, then put the zero check only for them.

romseygeek: I think all callers are already doing this work, it's just that they're checking if the returned `TermStatistics` object is null after the call rather than doing it up front?  I think I prefer this patch, as it makes the check more explicit.

bruno-roustant: Previously all callers checked if the returned TermStatistics was null before doing something. So I replaced the null check by a check if docFreq is 0.
I tried to remove the 0 check and ran TestShardSearching (which extends ShardSearchingTestBase I modified). It failed with the exception because docFreq was 0.
I think this check is necessary, unless some specific class is sure docFreq will always be > 0. It this case it would not previously check null.

jpountz: Maybe we should do it in two steps, first change the signature and keep the docFreq==0 check, this could probably go in 8.x, and then reject calls when docFreq==0, which we might only want to push to 9.0?

dsmiley: Makes sense @jpountz.  So in this issue we can keep this focused on the original goal, and leave a separate issue for the 9x change.  I think it's clearer if it's two issues.

dsmiley: Or maybe same issue as you/Bruno wish; I now see Bruno updated it.  Whichever.  I do like the latest 9x logic BTW -- avoids a null check.

jpountz: That works for me, but then let's only merge this to master? My concern is that this IndexSearcher method might not only be used internally but also by Lucene users.

dsmiley: I think in 8x we could keep the old method signature, mark it deprecated, and have it call this new one.  It's marked lucene.experimental, after all.

dsmiley: There is still the issue of subclassing but the current one could be marked final.  Again, this is lucene.experimental code.

jpountz: +1 to your idea of deprecating + marking final and introducing the new API

bruno-roustant: +1 also

PR #798, opened 2019-07-19
LUCENE-8906: Lucene50PostingsFormat.IntBlockTermState becomes public

bruno-roustant:


lucene/core/src/java/org/apache/lucene/codecs/lucene50/Lucene50PostingsFormat.java
```diff
@@ -453,15 +453,15 @@ public FieldsProducer fieldsProducer(SegmentReadState state) throws IOException
     }
   }
   
-  final static class IntBlockTermState extends BlockTermState {
-    long docStartFP = 0;
-    long posStartFP = 0;
-    long payStartFP = 0;
-    long skipOffset = -1;
-    long lastPosBlockOffset = -1;
+  public static final class IntBlockTermState extends BlockTermState {
```
dsmiley: I suspect "ant precommit" will not like this as it has no javadocs.  Remember to add the lucene.internal annotation (it extends BlockTermState which is annotated as-such).

bruno-roustant: You're right. I had hard time to fix all the javadoc for the precommit. I imposes to have a constructor doc since we have fields initialized with non 0 value. So I added a constructor.

PR #799, opened 2019-07-19
LUCENE-8920: encapsulate FST.Arc data

msokolov:
# Description

This changes the visibility of FTS.Arc members to private and adds accessor methods for outside consumers of them. It also reduces the visibility of some other FST members.

Commit #2 then does some refactoring so we can eliminate the public Arc setters.

# Tests
No new tests since there is no functional change


msokolov:
This PR got the wrong number (8902 vs 8920). I'm closing and will open a new one

PR #804, opened 2019-07-23
SOLR-13647 default solr.in.sh contains uncommented lines

jnyryan:
# Description

default version of this file should be completely commented

ENABLE_REMOTE_JMX_OPTS had defaults

Issue: https://issues.apache.org/jira/browse/SOLR-13647

# Solution

Comment the line in question

### DIFF
--- a/solr/bin/solr.in.sh
+++ b/solr/bin/solr.in.sh
@@ -82,7 +82,7 @@
 # Set to true to activate the JMX RMI connector to allow remote JMX client applications
 # to monitor the JVM hosting Solr; set to "false" to disable that behavior
 # (false is recommended in production environments)
-ENABLE_REMOTE_JMX_OPTS="false"
+# ENABLE_REMOTE_JMX_OPTS="false"
 
 # The script will use SOLR_PORT+10000 for the RMI_PORT or you can set it here
 # RMI_PORT=18983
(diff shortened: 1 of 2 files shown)


janhoy:
Linking with SOLR-13647

solr/bin/solr.in.sh
```diff
@@ -82,7 +82,7 @@
 # Set to true to activate the JMX RMI connector to allow remote JMX client applications
 # to monitor the JVM hosting Solr; set to "false" to disable that behavior
 # (false is recommended in production environments)
-ENABLE_REMOTE_JMX_OPTS="true"
+# ENABLE_REMOTE_JMX_OPTS="true"
```
janhoy: Actually the commented text should say "false" as it was in 8.0. Commented defaults should reflect the implicit defaults. Can you fix? Please also add a CHANGES.TXT entry about the fix.

jnyryan: I made those changes. review at your leisure

janhoy: Thanks. I meant to *leave* the line commented, but change from true to false..

jnyryan: keeping you on your toes - /facepalm

janhoy:
Merged

PR #809, opened 2019-07-27
LUCENE-8933: Validate JapaneseTokenizer user dictionary entry

mocobeta:
# Description

This adds a format check to Kuromoji user dictionary if the concatenated segment is same as its surface form. See https://issues.apache.org/jira/browse/LUCENE-8933


lucene/analysis/kuromoji/src/java/org/apache/lucene/analysis/ja/dict/UserDictionary.java
```diff
@@ -104,6 +104,8 @@ public int compare(String[] left, String[] right) {
     long ord = 0;
     
     for (String[] values : featureEntries) {
+      String surface = values[0].replaceAll(" ", "");
```
msokolov: maybe replace all white space here (ie including tabs) by using \s?

msokolov: ah never mind my question, I see from the email discussion that is the case: we shouldn't allow this kind of entry

mocobeta: > maybe replace all white space here (ie including tabs) by using \s?

Thanks, fixed.

lucene/analysis/kuromoji/src/test/org/apache/lucene/analysis/ja/dict/UserDictionaryTest.java
```diff
@@ -77,4 +78,19 @@ public void testRead() throws IOException {
     UserDictionary dictionary = TestJapaneseTokenizer.readDict();
     assertNotNull(dictionary);
   }
+
+  @Test(expected = RuntimeException.class)
```
jimczi: Can you use `expectThrows` and check that the message is correct ?

lucene/analysis/kuromoji/src/test/org/apache/lucene/analysis/ja/dict/UserDictionaryTest.java
```diff
@@ -77,4 +78,19 @@ public void testRead() throws IOException {
...
+  public void testReadInvalid1() throws IOException {
+    // the concatenated segment must be the same as the surface form
+    String invalidEntry = "日経新聞,日本 経済 新聞,ニホン ケイザイ シンブン,カスタム名詞";
+    UserDictionary dictionary = UserDictionary.open(new StringReader(invalidEntry));
+  }
+
+  @Test(expected = RuntimeException.class)
```
jimczi: same here

lucene/analysis/kuromoji/src/test/org/apache/lucene/analysis/ja/dict/UserDictionaryTest.java
mocobeta: I think you mean `assertThrows` that was introduced from JUnit 4.13
https://junit.org/junit4/javadoc/latest/org/junit/Assert.html#assertThrows(java.lang.String,%20java.lang.Class,%20org.junit.function.ThrowingRunnable)

Lucene uses junit 4.12, so we cannot use it (Am I missing something?).

mocobeta:
Thanks, I will merge it soon.

lucene/analysis/kuromoji/src/test/org/apache/lucene/analysis/ja/dict/UserDictionaryTest.java
mocobeta: Sorry, I missed LuceneTestCase#expectThrows. Will fix this in another patch.

mocobeta: Fixed in https://github.com/apache/lucene-solr/pull/830

PR #810, opened 2019-07-27
LUCENE-8933: Validate JapaneseTokenizer user dictionary entry (for 8.x)

mocobeta:
# Description

This adds a format check to Kuromoji user dictionary if the concatenated segment is not longer than its surface form. (For backwards compatibility, we check only segmentation length in 8.x and will add more strict equality check as of 9.0.) See https://issues.apache.org/jira/browse/LUCENE-8933


lucene/analysis/kuromoji/src/test/org/apache/lucene/analysis/ja/dict/UserDictionaryTest.java
```diff
@@ -77,4 +78,11 @@ public void testRead() throws IOException {
     UserDictionary dictionary = TestJapaneseTokenizer.readDict();
     assertNotNull(dictionary);
   }
+
+  @Test(expected = RuntimeException.class)
```
jimczi: Can you use `expectThrows` and check that the message is correct ?

mocobeta: Thanks, fixed.

PR #815, opened 2019-07-31
LUCENE-8213: Introduce Asynchronous Caching in LRUQueryCache

atris:


atris:
I ran luceneutil for wikipedia 10M with the concurrent searching and latency calculation patch applied.

https://gist.github.com/atris/e0fa10e79fb5ef62bd571406acf98433

There was no significant degradation to QPS, and the P999 and P100 latencies generally saw an improvement

msokolov:
It should be enough to report the stats after the last iteration - it is cumulative, so the previous ones just add noise? I agree QPS looks pretty noisy, probably no real change. Could you post the latency stats in a more readable table here? It looks as if you have markdown there: I think github will accept that

atris:
> It should be enough to report the stats after the last iteration - it is cumulative, so the previous ones just add noise? I agree QPS looks pretty noisy, probably no real change. 

I dont think thats true, since each run is its own JVM?

atris:
Another set of runs on wikimedium all with concurrent searching enabled:

                  Fuzzy1       47.29      (7.1%)       45.06     (11.1%)   -4.7% ( -21% -   14%)
            OrHighNotMed      405.86      (3.4%)      392.55      (2.2%)   -3.3% (  -8% -    2%)
           OrNotHighHigh      386.16      (4.7%)      373.54      (4.1%)   -3.3% ( -11% -    5%)
           BrowseDayOfYearTaxoFacets     6003.62      (2.6%)     5808.73      (2.2%)   -3.2% (  -7% -    1%)
                 Prefix3      176.87     (10.1%)      172.28      (8.9%)   -2.6% ( -19% -   18%)
           BrowseMonthTaxoFacets     6190.97      (3.8%)     6044.46      (4.9%)   -2.4% ( -10% -    6%)
               MedPhrase       40.97      (5.1%)       40.06      (5.5%)   -2.2% ( -12% -    8%)
            OrNotHighMed      383.00      (3.3%)      374.82      (4.7%)   -2.1% (  -9% -    6%)
              AndHighLow      191.05      (3.4%)      187.88      (3.2%)   -1.7% (  -7% -    5%)
            OrHighNotLow      416.92      (4.4%)      411.50      (4.3%)   -1.3% (  -9% -    7%)
              AndHighMed       39.58      (2.2%)       39.17      (1.9%)   -1.0% (  -5% -    3%)
                Wildcard       24.72      (7.9%)       24.49      (6.6%)   -0.9% ( -14% -   14%)
              HighPhrase       52.11      (5.6%)       51.63      (4.8%)   -0.9% ( -10% -   10%)
               LowPhrase       13.43      (2.7%)       13.33      (2.5%)   -0.8% (  -5% -    4%)
             AndHighHigh       12.68      (3.9%)       12.58      (3.2%)   -0.8% (  -7% -    6%)
                HighTerm      717.09      (4.8%)      712.98      (5.2%)   -0.6% ( -10% -    9%)
                PKLookup       91.70      (2.8%)       91.27      (3.8%)   -0.5% (  -6% -    6%)
                  IntNRQ       21.92     (17.9%)       21.83     (18.0%)   -0.4% ( -30% -   43%)
                 Respell       34.38      (3.3%)       34.24      (2.3%)   -0.4% (  -5% -    5%)
            HighTermDayOfYearSort       27.44      (3.2%)       27.33      (1.6%)   -0.4% (  -5% -    4%)
           OrHighNotHigh      463.40      (5.4%)      461.74      (4.3%)   -0.4% (  -9% -    9%)
           BrowseDateTaxoFacets        0.69      (0.4%)        0.69      (0.5%)   -0.2% (  -1% -    0%)
           BrowseMonthSSDVFacets        2.63      (1.6%)        2.63      (1.1%)   -0.1% (  -2% -    2%)
                 MedTerm      885.64      (5.0%)      885.56      (3.6%)   -0.0% (  -8% -    8%)
                  Fuzzy2       39.47      (7.1%)       39.47      (9.4%)   -0.0% ( -15% -   17%)
           BrowseDayOfYearSSDVFacets        2.41      (0.4%)        2.41      (0.3%)   -0.0% (   0% -    0%)
            HighSpanNear        6.53      (1.0%)        6.53      (1.3%)    0.0% (  -2% -    2%)
            MedSloppyPhrase       31.76      (2.0%)       31.79      (1.6%)    0.1% (  -3% -    3%)
           HighIntervalsOrdered        6.10      (1.6%)        6.11      (2.1%)    0.1% (  -3% -    3%)
               OrHighLow      177.27      (2.2%)      177.50      (2.5%)    0.1% (  -4% -    5%)
         LowSloppyPhrase       30.62      (1.9%)       30.67      (1.7%)    0.2% (  -3% -    3%)
             MedSpanNear        7.63      (1.7%)        7.64      (2.0%)    0.2% (  -3% -    3%)
             LowSpanNear        8.34      (1.3%)        8.37      (1.8%)    0.3% (  -2% -    3%)
            OrNotHighLow      308.14      (2.0%)      309.75      (4.9%)    0.5% (  -6% -    7%)
                 LowTerm      861.94      (4.6%)      870.93      (2.8%)    1.0% (  -6% -    8%)
        HighSloppyPhrase        5.58      (3.2%)        5.64      (2.8%)    1.1% (  -4% -    7%)
               OrHighMed       10.81      (2.4%)       10.95      (2.6%)    1.4% (  -3% -    6%)
              OrHighHigh       10.28      (2.5%)       10.48      (3.2%)    1.9% (  -3% -    7%)


Seems there is no degradation?

atris:
Rebased with master.

Any thoughts on this one? Seems like a useful change with no degradation in the happy path?

mikemccand:
> > It should be enough to report the stats after the last iteration - it is cumulative, so the previous ones just add noise? I agree QPS looks pretty noisy, probably no real change.
> 
> I dont think thats true, since each run is its own JVM?

It is true -- `luceneutil` runs multiple JVMs to try to sample the noise due to hotspot mis-compilation, and then each iteration in reports the cumulative results of all prior iterations so far.  So reporting the last result is good.

mikemccand:
> It should be enough to report the stats after the last iteration - it is cumulative, so the previous ones just add noise? I agree QPS looks pretty noisy, probably no real change. Could you post the latency stats in a more readable table here? It looks as if you have markdown there: I think github will accept that

+1 to inline the latency results in a readable way here.

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
```diff
@@ -367,6 +393,7 @@ public void clearQuery(Query query) {
       if (singleton != null) {
         onEviction(singleton);
       }
+      inFlightAsyncLoadQueries.remove(query);
```
mikemccand: Let's get this under `finally` clause?

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
```diff
@@ -262,6 +283,11 @@ DocIdSet get(Query key, IndexReader.CacheHelper cacheHelper) {
     assert lock.isHeldByCurrentThread();
     assert key instanceof BoostQuery == false;
     assert key instanceof ConstantScoreQuery == false;
+
+    if (inFlightAsyncLoadQueries.contains(key)) {
```
mikemccand: Can you add to javadocs for this method explaining when/why this exception is thrown?

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
```diff
@@ -656,13 +713,20 @@ public long ramBytesUsed() {
     // threads when IndexSearcher is created with threads
     private final AtomicBoolean used;
 
+    private Executor executor;
```
mikemccand: `final`?

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
```diff
@@ -726,14 +790,35 @@ public ScorerSupplier scorerSupplier(LeafReaderContext context) throws IOExcepti
       DocIdSet docIdSet;
       try {
         docIdSet = get(in.getQuery(), cacheHelper);
+      } catch (AsyncQueryLoadInProgressException e) {
```
mikemccand: We could alternatively have this request wait for cache to populate and then return that cached result.  It'd be cheaper CPU overall, especially in degenerate queries where the same query is often executed multiple times concurrently, but I think more complicated here, so let's go forward w/ this approach?

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
atris: I did not add that due to the way the constructors use each other -- for the default case, the executor is null.

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
atris: +1, I agree

atris:
@mikemccand Thanks for the inputs, updated the PR. Please let me know your comments

atris:
Any further thoughts on this one?

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
```diff
@@ -89,12 +92,30 @@
...
   private final long maxRamBytesUsed;
   private final Predicate<LeafReaderContext> leavesToCache;
   // maps queries that are contained in the cache to a singleton so that this
   // cache does not store several copies of the same query
   private final Map<Query, Query> uniqueQueries;
+  // Marks the inflight queries that are being asynchronously loaded into the cache
+  // This is used primarily to ensure that multiple threads do not trigger loading
+  // of the same query in the same cache. We use a set because it is an invariant that
+  // the entries of this data structure be unique.
+  private final Set<Query> inFlightAsyncLoadQueries = Collections.newSetFromMap(new ConcurrentHashMap<Query,Boolean>());
```
mikemccand: It looks like we only ever access this map/set while holding the `lock`?  Can we use an ordinary `HashMap` if so?

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
```diff
@@ -726,14 +794,35 @@ public ScorerSupplier scorerSupplier(LeafReaderContext context) throws IOExcepti
...
+          if (executor != null) {
+            FutureTask<Void> task = new FutureTask<>(() -> {
+              DocIdSet localDocIdSet = cache(context);
+              putIfAbsent(in.getQuery(), localDocIdSet, cacheHelper);
+
+              //remove the key from inflight -- the key is loaded now
+              inFlightAsyncLoadQueries.remove(in.getQuery());
+              return null;
+            });
+            inFlightAsyncLoadQueries.add(in.getQuery());
```
mikemccand: I think we have a concurrency bug here -- if the same query across different threads is checked for caching, then two threads can get to this point, and both can compute the cached result and store it?  Because, this `.add` happens at a different time from the above check (`.contains`)...

Maybe this doesn't matter?  It just means two threads do the work of caching, and one is wasted?

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
```diff
@@ -807,14 +896,35 @@ public BulkScorer bulkScorer(LeafReaderContext context) throws IOException {
...
-          putIfAbsent(in.getQuery(), docIdSet, cacheHelper);
+          // If asynchronous caching is requested, perform the same and return
+          // the uncached iterator
+          if (executor != null) {
+            FutureTask<Void> task = new FutureTask<>(() -> {
+              DocIdSet localDocIdSet = cache(context);
+              putIfAbsent(in.getQuery(), localDocIdSet, cacheHelper);
+
+              //remove the key from inflight -- the key is loaded now
+              inFlightAsyncLoadQueries.remove(in.getQuery());
```
mikemccand: Disappointing that all this code is duplicated between `scorerSupplier` and `bulkScorer` methods ... can we somehow factor to share?

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
atris: Yes, that is precisely the idea. The alternative to that is a more aggressive locking which can potentially have a larger negative effect than a thread's caching effort. Thanks for raising this, I have added docs to explain this now.

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
atris: We add to this map without a lock when triggering a caching operation, hence the need for a concurrent safe set. The idea is that it is a better idea to let two threads request an asynchronous cache load for the same query concurrently than syncing the entire operation under a lock. However, this means that two threads might want to add the same key to `inFlightAsyncLoadQueries` in the same duration, so we need that operation to be idempotent and thread safe.

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
atris: Yeah, that is the idea, since that saves us a bit of pessimism when locking. Thanks for highlighting, I have added docs around the same.

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
atris: Good point -- refactored to a common method.

atris:
@mikemccand Thanks for reviewing -- updated per comments. Please see and let me know your thoughts.

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
```diff
@@ -832,5 +924,20 @@ public BulkScorer bulkScorer(LeafReaderContext context) throws IOException {
       return new DefaultBulkScorer(new ConstantScoreScorer(this, 0f, ScoreMode.COMPLETE_NO_SCORES, disi));
     }
 
+    // Perform a cache load asynchronously
+    // NOTE: Potentially, two threads can trigger a load for the same query concurrently as the check for presence of the query
+    // done upstream and the lock is not he
```
mikemccand: Hmm this sentence abruptly ended?

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
```diff
@@ -262,6 +284,15 @@ DocIdSet get(Query key, IndexReader.CacheHelper cacheHelper) {
     assert lock.isHeldByCurrentThread();
     assert key instanceof BoostQuery == false;
     assert key instanceof ConstantScoreQuery == false;
+
+    /*
+     * If the current query is already being asynchronously cached,
+     * do not trigger another cache operation
+     */
+    if (inFlightAsyncLoadQueries.contains(key)) {
```
mikemccand: Hmm can we fix the concurrency issue here, e.g. just call `inFlightAsyncLoadQueries.add` and if the returned value is `false` (because the query was already in the set) then throw the exception?  Or, move this check down into the `cacheAsynchronously` method where we are doing the `add`?

atris: Good idea, fixed, thanks

atris:
@mikemccand Thanks, fixed. Interestingly, moving the asynchronous load check to `cacheAsynchronously` also removed the need for the new exception. Please see the latest and share your comments.

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
```diff
@@ -95,6 +98,11 @@
   // maps queries that are contained in the cache to a singleton so that this
   // cache does not store several copies of the same query
   private final Map<Query, Query> uniqueQueries;
+  // Marks the inflight queries that are being asynchronously loaded into the cache
+  // This is used primarily to ensure that multiple threads do not trigger loading
```
mikemccand: Remove `primarily`?  This is its sole purpose?

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
mikemccand: Hmm can we duplicate a little constructor code and then make it `final`?

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
```diff
@@ -832,5 +894,25 @@ public BulkScorer bulkScorer(LeafReaderContext context) throws IOException {
...
+
+        //remove the key from inflight -- the key is loaded now
+        inFlightAsyncLoadQueries.remove(in.getQuery());
+        return null;
+      });
+      /*
+       * If the current query is already being asynchronously cached,
+       * do not trigger another cache operation
+       */
+      if (inFlightAsyncLoadQueries.contains(in.getQuery())) {
```
mikemccand: Can you move the `task` creation to after this `if`?

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
```diff
@@ -832,5 +894,25 @@ public BulkScorer bulkScorer(LeafReaderContext context) throws IOException {
...
+        return;
+      }
+      inFlightAsyncLoadQueries.add(in.getQuery());
```
mikemccand: Hmm still a concurrency bug here, allowing two threads to cache the same query -- instead of separate `.contains` and `.add` calls, can you simply call `.add` and look at its return value to decide whether you should in fact cache?

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
```diff
@@ -448,13 +459,42 @@ void assertConsistent() {
     }
   }
 
+  // pkg-private for testing
+  // return the list of queries being loaded asynchronously
+  List<Query> inFlightQueries() {
+    lock.lock();
+    try {
+      return new ArrayList<>(inFlightAsyncLoadQueries);
```
mikemccand: I'm still confused about `lock` -- do we always hold the lock when checking if query is already in the map?  If so, we don't need a `ConcurrentHashMap`?  If not, why do we even have the `lock` since it is a `ConcurrentHashMap`.

atris: Not all places which access `inFlightAsyncLoadQueries` take a lock -- the main being `cacheAsynchronously`

atris: This specific access site does not lead a lock -- thanks for highlighting that, fixed!

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
atris: I was hesitant to do that since that would entail moving out the addition of the entry in the map from the actual triggering of cache load. However, there is minimal risk involved in decoupling these two, and will indeed safeguard us completely from two threads caching the same query. Fixed and updated

atris:
@mikemccand Thanks for your inputs, updated the same

mikemccand:
> @mikemccand Thanks, fixed. Interestingly, moving the asynchronous load check to `cacheAsynchronously` also removed the need for the new exception. Please see the latest and share your comments.

Ahh that is a nice side effect!

atris:
> > @mikemccand Thanks, fixed. Interestingly, moving the asynchronous load check to `cacheAsynchronously` also removed the need for the new exception. Please see the latest and share your comments.
> 
> Ahh that is a nice side effect!

Indeed!

Does the latest iteration look ready? Anything that sticks out?

lucene/CHANGES.txt
```diff
@@ -56,6 +56,9 @@ Improvements
 * LUCENE-8937: Avoid agressive stemming on numbers in the FrenchMinimalStemmer.
   (Adrien Gallou via Tomoko Uchida)
 
+* LUCENE-8213: LRUQueryCache#doCache can now use IndexSearcher's Executor (if present)
```
mikemccand: s/`can now use`/`now uses`?

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
```diff
@@ -832,5 +889,25 @@ public BulkScorer bulkScorer(LeafReaderContext context) throws IOException {
...
+       * If the current query is already being asynchronously cached,
+       * do not trigger another cache operation
+       */
+      if (inFlightAsyncLoadQueries.add(in.getQuery()) == false) {
+        return;
+      }
+
+      FutureTask<Void> task = new FutureTask<>(() -> {
+        DocIdSet localDocIdSet = cache(context);
+        putIfAbsent(in.getQuery(), localDocIdSet, cacheHelper);
```
mikemccand: Can you save the return value from `putIfAbsent` call and then `assert` it's `null`?  I.e. the cache should never already have this query since we checked up-front in thread tight manner?

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
```diff
@@ -732,8 +774,16 @@ public ScorerSupplier scorerSupplier(LeafReaderContext context) throws IOExcepti
...
-          docIdSet = cache(context);
-          putIfAbsent(in.getQuery(), docIdSet, cacheHelper);
+          // If asynchronous caching is requested, perform the same and return
+          // the uncached iterator
+          if (executor != null) {
+            cacheAsynchronously(context, cacheHelper);
+            return in.scorerSupplier(context);
+          }
+          else {
+            docIdSet = cache(context);
```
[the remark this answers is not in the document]
atris: +1, that is a great observation, thanks for highlighting it!

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
```diff
@@ -832,5 +889,25 @@ public BulkScorer bulkScorer(LeafReaderContext context) throws IOException {
...
+
+      FutureTask<Void> task = new FutureTask<>(() -> {
+        DocIdSet localDocIdSet = cache(context);
+        putIfAbsent(in.getQuery(), localDocIdSet, cacheHelper);
+
+        //remove the key from inflight -- the key is loaded now
+        inFlightAsyncLoadQueries.remove(in.getQuery());
+        return null;
+      });
+      executor.execute(task);
```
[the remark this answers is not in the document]
atris: That depends on the implementation of the executor passed in -- I think we do not publish any such contract in `IndexSearcher`, and since this is reusing the same `Executor`, the behaviour is the same.

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
atris: `putIfAbsent` does not return anything today -- changing that would be an orthogonal change IMO. Maybe as a follow up?

mikemccand: Ahh OK let's do that later then.

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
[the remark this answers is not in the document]
mikemccand: Well, the `RejectedExecutionException` is a public part of this API ... I think we need to handle it in some fashion, maybe fallback to sync'd (calling-thread-executes) implementation, or, we skip caching and continue running the query uncached?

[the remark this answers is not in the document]
atris: Fixed. I think that actually is a valid problem even in `IndexSearcher`? I will open another issue to track that.

atris:
@mikemccand Updated the PR, please see and let me know.

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
```diff
@@ -813,8 +881,28 @@ public BulkScorer bulkScorer(LeafReaderContext context) throws IOException {
...
         if (policy.shouldCache(in.getQuery())) {
-          docIdSet = cache(context);
-          putIfAbsent(in.getQuery(), docIdSet, cacheHelper);
+          boolean performSynchronousCaching = !(executor != null);
+          // If asynchronous caching is requested, perform the same and return
+          // the uncached iterator
+          if (!performSynchronousCaching) {
+            try {
+              cacheAsynchronously(context, cacheHelper);
+            } catch (RejectedExecutionException e) {
```
mikemccand: Hmm, can you shrink-wrap this exception handling?  I.e., handle it down in `cacheAsynchronously` where we ask the executor to execute the task, instead of up here?

Maybe then return a `boolean` from `cacheAsynchronously` to know (here) if caller should then cache synchronously ...

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
```diff
@@ -832,5 +920,25 @@ public BulkScorer bulkScorer(LeafReaderContext context) throws IOException {
...
+      if (inFlightAsyncLoadQueries.add(in.getQuery()) == false) {
+        return;
+      }
+
+      FutureTask<Void> task = new FutureTask<>(() -> {
+        DocIdSet localDocIdSet = cache(context);
+        putIfAbsent(in.getQuery(), localDocIdSet, cacheHelper);
+
+        //Remove the key from inflight -- the key is loaded now
+        inFlightAsyncLoadQueries.remove(in.getQuery());
```
mikemccand: Can you `assert` the return value here?  I.e. we should *always* succeed in removing this query from the set.

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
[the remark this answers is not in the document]
mikemccand: Yeah it is, thanks for opening separate issue ...

atris:
@mikemccand Updated, please see

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
```diff
@@ -832,5 +911,36 @@ public BulkScorer bulkScorer(LeafReaderContext context) throws IOException {
...
+        //Remove the key from inflight -- the key is loaded now
+        Object retValue = inFlightAsyncLoadQueries.remove(in.getQuery());
+
+        assert retValue != null;
+
+        return null;
+      });
+      try {
+        executor.execute(task);
+      } catch (RejectedExecutionException e) {
```
mikemccand: Can you updated `IndexSearcher` javadocs indicating that 1) the `Executor` is used for concurrent query caching, and 2) how we handle `RejectedExecutionException` from it (caller runs)?

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
```diff
@@ -867,12 +881,25 @@ public BulkScorer bulkScorer(LeafReaderContext context) throws IOException {
...
           // If asynchronous caching is requested, perform the same and return
           // the uncached iterator
-          if (executor != null) {
-            cacheAsynchronously(context, cacheHelper);
-            return in.bulkScorer(context);
-          } else {
+          if (!performSynchronousCaching) {
+            try {
+              cacheAsynchronously(context, cacheHelper);
+            } catch (RejectedExecutionException e) {
```
msokolov: Do we need to handle these exceptions? Isn't this equivalent to the caller's having specified a "run in caller" RejectedExecutionHandler? The caller may have specified a different policy, but here we are not respecting that

atris: If the caller has specified a different policy -- we should not be even getting here. `IndexSearcher#search` API does not advertise the behaviour for the case if the underlying `Executor` does not have enough capacity, nor does it consider the size of the executor during allocation of leaves. Both of the above facts lead to an implicit API contract where we are expected to gracefully handle the case of rejected execution. In case the user wishes to handle the exception themselves, they are free to override the default methods -- similar to what we advice for other methods in `IndexSearcher`.

atris:
Thanks @mikemccand ! This was an extensive review -- thank you for spending the time on it!

atris:
Ran the Lucene test suite on the latest iteration -- came in clean.

PR #817, opened 2019-08-02
SOLR-13655:Upgrade Collections.unModifiableSet to Set.of and Set.copyOf

atris:


atris:
Any thoughts on this? Seems safe enough to merge?

atris:
Gentle Ping

tflobbe:
Thanks @atris. Changes look good to me, however, I think there are a couple more cases that we could change:
* `ConfigNameConstants`
* `Variable.Type`
* `SolrRequest`
* `ZkStateReader.KNOWN_CLUSTER_PROPS`

atris:
@tflobbe Thanks, for some reason, my grep command missed those ones. Updated the PR with the same.

atris:
@tflobbe Thanks. I was slightly unsure about whether these sections actually need an immutable copy of the original set (by their usage patterns), hence was being conservative by doing copyOf instead of of(). However, post your comment, I looked at the call sites of these methods and they look to be fine with an immutable reference, so changed the calls.

tflobbe:
Thanks @atris! squashed/merged

atris:
Thanks @tflobbe !

PR #819, opened 2019-08-02
SOLR-13676: Reduce log verbosity in TestDistributedGrouping 

diegoceccarelli:
using ignoreException

# Description

SOLR-13404 added a test that expects Solr to fail if grouping is called with group.offset < 0. When the test runs it succeeds but the whole stack trace is printed out in the logs. 
# Solution

This small patch avoid the stack trace by using ignoreException. I also replaced an `assertTrue` with a more specific check. 

# Tests

This patch is improving a test.

# Checklist

- [...] I have added documentation for the [Ref Guide](https://github.com/apache/lucene-solr/tree/master/solr/solr-ref-guide) (for Solr changes only).

### DIFF
--- a/solr/core/src/test/org/apache/solr/TestDistributedGrouping.java
+++ b/solr/core/src/test/org/apache/solr/TestDistributedGrouping.java
@@ -231,6 +231,7 @@ public class TestDistributedGrouping extends BaseDistributedSearchTestCase {
     );
     assertEquals(SolrException.ErrorCode.BAD_REQUEST.code, exception.code());
     Assert.assertThat(exception.getMessage(), containsString("'group.offset' parameter cannot be negative"));
+    resetExceptionIgnores();
 
     query("q", "*:*",
         "group", "true",
--- a/solr/core/src/test/org/apache/solr/TestDistributedGrouping.java
+++ b/solr/core/src/test/org/apache/solr/TestDistributedGrouping.java
@@ -29,7 +29,6 @@ import org.apache.solr.common.params.CommonParams;
 import org.apache.solr.common.params.ModifiableSolrParams;
 import org.apache.solr.common.util.NamedList;
 import org.apache.solr.SolrTestCaseJ4.SuppressPointFields;
-import org.junit.Assert;
 import org.junit.Test;
 
 import static org.hamcrest.CoreMatchers.containsString;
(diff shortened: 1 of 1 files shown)


diegoceccarelli:
Please note: I ran precommit but it failed, I'm not sure it is my change, I'll look into that tomorrow..

solr/core/src/test/org/apache/solr/TestDistributedGrouping.java
```diff
@@ -29,8 +29,11 @@
 import org.apache.solr.common.params.ModifiableSolrParams;
 import org.apache.solr.common.util.NamedList;
 import org.apache.solr.SolrTestCaseJ4.SuppressPointFields;
+import org.junit.Assert;
```
munendrasn: I think we don't need this import. All test cases extend `Assert`

solr/core/src/test/org/apache/solr/TestDistributedGrouping.java
```diff
@@ -221,12 +224,13 @@ public void test() throws Exception {
         "fl", "id", "group.format", "grouped", "group.limit", "-12",
         "sort", i1 + " asc, id asc");
 
+    ignoreException("'group.offset' parameter cannot be negative");
```
munendrasn: If we are doing this, shouldn't we call `resetExceptionIgnores()` after exception verification?

solr/core/src/test/org/apache/solr/TestDistributedGrouping.java
diegoceccarelli: I think `Assert` is still needed  (I use `Assert.assertThat`, code won't compile otherwise).

solr/core/src/test/org/apache/solr/TestDistributedGrouping.java
diegoceccarelli: Good point, added `resetExceptionIgnores`. Thanks!

solr/core/src/test/org/apache/solr/TestDistributedGrouping.java
munendrasn: What I meant was since all tests extend Assert.. you could just use assertThat instead of Assert.assertThat

diegoceccarelli: Ouch sorry, I didn't understand, done! thanks!

PR #824, opened 2019-08-07
LUCENE-8755: QuadPrefixTree robustness

chenkovsky:
Reimplement QuadPrefixTree and PackedQuadPrefixTree's getCell method.

### DIFF
--- a/lucene/spatial-extras/src/java/org/apache/lucene/spatial/prefix/tree/PackedQuadPrefixTree.java
+++ b/lucene/spatial-extras/src/java/org/apache/lucene/spatial/prefix/tree/PackedQuadPrefixTree.java
@@ -22,6 +22,7 @@ import java.util.List;
 import java.util.NoSuchElementException;
 
 import org.apache.lucene.util.BytesRef;
+import org.apache.lucene.util.Version;
 import org.locationtech.spatial4j.context.SpatialContext;
 import org.locationtech.spatial4j.shape.Point;
 import org.locationtech.spatial4j.shape.Rectangle;
@@ -61,7 +62,7 @@ public class PackedQuadPrefixTree extends QuadPrefixTree {
     @Override
     protected SpatialPrefixTree newSPT() {
       PackedQuadPrefixTree tree = new PackedQuadPrefixTree(ctx, maxLevels != null ? maxLevels : MAX_LEVELS_POSSIBLE);
-      tree.robust = robust;
+      tree.robust = getVersion().onOrAfter(Version.LUCENE_8_3_0);
       return tree;
     }
   }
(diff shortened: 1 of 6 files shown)


dsmiley:
Thanks for the PR!  Just curious; did you encounter this problem and thus were motivated to fix it?  

I didn't spend much time on it tonight, but it looks pretty good, and it has a cleaner appearance that reads better to me.  You did remove a comment or two that are still appropriate, like the link to Z-order.  Did you run Lucene spatial-extras tests?  "ant precommit"?

I suspect a tokenization change ought to be toggled by Version, similar to some other behavior changes in Lucene analyzers.  And that means finding a way for the old tokenization and the new-one to co-exist, plus a means to communicate that choice -- probably via `org.apache.lucene.spatial.prefix.tree.SpatialPrefixTreeFactory#init`.  If we don't do this, then some use-cases with existing indexes may break, like exact point lookup of itself should round-trip.

chenkovsky:
yes, I run  spatial-extras tests.

chenkovsky:
@dsmiley for backward compatibility problem, is there something like "migration" in lucene?  When I develop web project, I find that migration idea benefits me a lot when the schema changes. 

Does the indexed data contain luncene version info which created the index? If it contains, We can use old code to search in old indexed data.

chenkovsky:
I find that Lucene has a tool called IndexUpgrader. Maybe updating index is a better solution for backward compatibility?

ErickErickson:
No, IndexUpgraderTool doesn’t really help. It rewrites the index in the current format, but does not (and cannot) make it look just as though the index had been indexed from scratch. From Robert Muir:

“I think the key issue here is Lucene is an index not a database. Because it is a lossy index and does not retain all of the user's data, its not possible to safely migrate some things automagically. In the norms case IndexWriter needs to re-analyze the text ("re-index") and compute stats to get back the value, so it can be re-encoded. The function is y = f(x) and if x is not available its not possible, so lucene can't do it.”

So Lucene will work with X-1, but not version X-2. As of 8.0, it will refuse to even open an index that has ever been touched by Solr 6x or earlier, regardless of running IndexUpgraderTool etc.

Best,
Erick

dsmiley:
> Does the indexed data contain luncene version info which created the index? If it contains, We can use old code to search in old indexed data.

It does contain the version of Lucene, but not the Version used in analysis components (see class Version.java).  In a sense the SpatialStrategy impls play the role of an analysis component & more).  In order for a user of Lucene (e.g. Solr and it's users) to be able to use an existing index when they upgrade Lucene, a user can configure a specific version for analysis components consistent with the version that was used to write the index (will get set on an Analyzer.setVersion(...).  If this is Solr, then this is set via luceneMatchVersion; raw Lucene usage requires explicit control here.

There has yet to be a SpatialStrategy that tweaked it's operation based on the Version; the only change or two that comes to mind were backwards compatible (e.g. fewer needless terms).  So there's no trace of Version usage in spatial-extras as of yet.

I know this is kinda a pain to concern yourself with.

chenkovsky:
I reimplemented it. now user can specify the version that the tree's compatible with.

dsmiley:
Any concerns/questions on the commit I added @chenkovsky ?

BTW this is the first time I used a feature of GitHub that allows me (a member of this repo) to commit to the source branch of someone's PR.  Pretty cool for collaboration.

PR #825, opened 2019-08-07
SOLR-13677 All Metrics Gauges should be unregistered by the objects that registered them

noblepaul:


solr/core/src/java/org/apache/solr/metrics/SolrMetricProducer.java
```diff
@@ -19,17 +19,58 @@
 /**
  * Used by objects that expose metrics through {@link SolrCoreMetricManager}.
  */
-public interface SolrMetricProducer {
+public interface SolrMetricProducer extends AutoCloseable {
+
+  /**
+   * Unique metric name is in the format of A/B/C
```
sigram: This is purely aesthetic, but maybe instead of `/` we should use a `.` as in all other metric-related namespaces?

solr/core/src/java/org/apache/solr/search/FastLRUCache.java
```diff
@@ -227,6 +227,7 @@ public void warm(SolrIndexSearcher searcher, SolrCache old) {
 
   @Override
   public void close() {
+    if(metricsInfo != null) metricsInfo.unregister();
```
sigram: IIRC Apache code style inserts a space between `if` and condition.

solr/core/src/java/org/apache/solr/metrics/SolrMetricProducer.java
```diff
@@ -19,17 +19,58 @@
...
+    final SolrMetricManager metricManager;
+    final String myTag;
+
+    public MetricsInfo(SolrMetricManager metricManager, String registry, String tag) {
+      this.registry = registry;
+      this.metricManager = metricManager;
+      this.myTag = tag;
+    }
+
+    public String getTag(){
```
sigram: Space before curly bracket.

solr/core/src/java/org/apache/solr/metrics/SolrMetricProducer.java
```diff
@@ -19,17 +19,58 @@
...
-   * @param scope scope of the metrics (eg. handler name) to separate metrics of
+   * @param tag      a symbolic tag that represents this instance of the producer,
+   *                 or a group of related instances that have the same life-cycle. This tag is
+   *                 used when managing life-cycle of some metrics and is set when
+   *                 {@link #initializeMetrics(SolrMetricManager, String, String, String)} is called.
+   * @param scope    scope of the metrics (eg. handler name) to separate metrics of
    */
   void initializeMetrics(SolrMetricManager manager, String registry, String tag, String scope);
+
+  class MetricsInfo {
```
sigram: I think the name `MetricsInfo` is too generic. This is more like a decorator for a tag, so maybe `MetricsTag` ?

solr/core/src/java/org/apache/solr/search/LFUCache.java
```diff
@@ -227,6 +227,7 @@ public void close() {
     statsList.get(0).add(cache.getStats());
     statsList.remove(cache.getStats());
     cache.destroy();
+
```
sigram: `metricsInfo.unregister()` ?

solr/core/src/java/org/apache/solr/handler/RequestHandlerBase.java
```diff
@@ -319,6 +325,11 @@ public PluginInfo getPluginInfo(){
     return  pluginInfo;
   }
 
+  @Override
+  public void close()  {
+    if (metricsInfo != null) metricsInfo.unregister();
```
sigram: Again, this is also a repeating pattern that is easy to forget (in fact you forgot about this in one place even in this patch). If it's something that has to be done reliably then I think the framework should take care of that.

See for example how `SolrInfoBean.registerMetricName()` is designed - it uses result of another method to determine whether it's a no-op. Similarly, we could add `SolrMetricProducer.getMetricsInfo()` in addition to `AutoCloseable` and then always call `unregister` in `default void close()`.

solr/core/src/java/org/apache/solr/search/LFUCache.java
```diff
@@ -254,8 +255,13 @@ private static String calcHitRatio(long lookups, long hits) {
     return "0." + hundredths;
   }
 
+
+  private MetricsInfo metricsInfo;
+
   @Override
   public void initializeMetrics(SolrMetricManager manager, String registryName, String tag, String scope) {
+    metricsInfo = new MetricsInfo(manager, registryName, getUniqueMetricTag(tag));
+    tag = metricsInfo.getTag();
```
sigram: Not needed, since the call below uses metricsInfo.getTag().

solr/core/src/java/org/apache/solr/metrics/SolrMetricProducer.java
noblepaul: Well, it contains more information than just a tag. That's why I had to choose a name other than MetricsTag. The problem we have with Metrics framework is that every component that is a MetricsProducer is forced to be aware of all internals of MetricsFramework. We should minimize the no:of Objects one need to keep track of (ideally a single Object)

solr/core/src/java/org/apache/solr/handler/RequestHandlerBase.java
```diff
@@ -145,8 +145,14 @@ public void init(NamedList args) {
 
   }
 
+  protected MetricsInfo metricsInfo;
+
+
   @Override
   public void initializeMetrics(SolrMetricManager manager, String registryName, String tag, final String scope) {
+    metricsInfo = new MetricsInfo(manager, registryName, getUniqueMetricTag(tag)) ;
+    tag = metricsInfo.getTag();
```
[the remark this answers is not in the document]
noblepaul: Yeah, we should avoid passing this zillion different parameters . But the problem is it is too much of a change. However I'll try to do it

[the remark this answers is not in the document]
sigram: We can change it in 9.0 - how about this for 8x, in `SolrMetricProducer`:
```
default void initializeMetrics(MetricsInfo info, scope) {
  initializeMetrics(info.manager, info.registry, info.tag, scope);
}
```
Then we can change the components that we know about to use the new API (and create the tags not in the component itself but in the parent), and still preserve back-compat for third-party components.

solr/core/src/java/org/apache/solr/metrics/SolrMetricProducer.java
sigram: Maybe `MetricsContext`?

noblepaul:
I've opened another PR https://github.com/apache/lucene-solr/pull/836

PR #826, opened 2019-08-09
SOLR-13399: add splitByPrefix configuration to IndexSizeTrigger

megancarey:
# Description

I've added the `splitByPrefix` parameter to the IndexSizeTrigger, so that the trigger is able to execute shard splits by prefix. I also added test cases to validate the split configs baked into this trigger. 

Plus, I made a small change to the splitByPrefix code for splitting by ID, and cleaned up the tests a bit.

# Solution

The solution I took to address this problem models the convention of JIRA-12942, which adds the split configs directly into the trigger config. 

The change I made to the splitByPrefix code doesn't modify functionality at all, but the implementation uses BytesRef.bytesEquals instead of manually comparing byte arrays.

# Tests

I added additional checks to the test IndexSizeTriggerTest.testTrigger, which validates that the default split configs are added to the trigger config. I also added a check to IndexSizeTriggerTest.testSplitConfig that validates the splitByPrefix override. Finally, I added a test case that attempts trigger configuration with invalid configs, and fails.

# Checklist

- [] I have created a Jira issue and added the issue ID to my pull request title.
  - I'm piggybacking on existing JIRA SOLR-13399


solr/core/src/java/org/apache/solr/handler/admin/SplitOp.java
```diff
@@ -385,16 +385,8 @@ static String toSplitString(Collection<DocRouter.Range> splits) throws Exception
...
-        for (; i < currPrefix.length; i++) {
-          if (currPrefix.bytes[i] != term.bytes[term.offset + i]) {
-            break;
-          }
-        }
-
-        if (i == currPrefix.length) {
-          // prefix was the same (common-case fast path)
-          // int count = termsEnum.docFreq();
+        BytesRef termPrefix = new BytesRef(term.bytes, term.offset, currPrefix.length);
```
yonik: Found a better choice here that doesn't involve  object creation: StringHelper.startsWith()

yonik:
OK, I implemented the change to use StringHelper.startsWith(), squashed the commits, added SOLR-13399 at the start of the commit message, and pushed to master & 8x.
Thanks Megan!

PR #832, opened 2019-08-14
LUCENE-8952: Use a sort key instead of true distance in NearestNeighbor.

jtibshirani:
This commit addresses a TODO in `NearestNeighbors` around switching to
`SloppyMath.haversinSortKey`. When comparing candidate hits, we now only compute
a distance sort key. The sort key is converted to a true distance only when returning
the final set of `FieldDocs` and when calculating the bbox for the current search area.


iverase:
FYI: I opened #842 which seems to help a lot in performance as well.

jtibshirani:
Thanks for the review @iverase, I pushed some new changes.

iverase:
Thank you @jtibshirani!

PR #838, opened 2019-08-20
SOLR-13705 Double-checked locking bug is fixed.

kamaci:
# Description

Using double-checked locking for the lazy initialization of any other type of primitive or mutable object risks a second thread using an uninitialized or partially initialized member while the first thread is still creating it, and crashing the program.

# Solution

Double-checked locking bug is fixed to prevent it.

# Tests

No need to write extra tests. Existing tests should not fail after this implementation.

# Checklist

### DIFF
--- a/solr/core/src/java/org/apache/solr/handler/admin/SecurityConfHandler.java
+++ b/solr/core/src/java/org/apache/solr/handler/admin/SecurityConfHandler.java
@@ -260,22 +260,25 @@ public abstract class SecurityConfHandler extends RequestHandlerBase implements 
 
   @Override
   public Collection<Api> getApis() {
-    if (apis == null) {
+    Collection<Api> syncedApis = apis;
+    if (syncedApis == null) {
       synchronized (this) {
-        if (apis == null) {
-          Collection<Api> apis = new ArrayList<>();
+        syncedApis = apis;
+        if (syncedApis == null) {
+          syncedApis = new ArrayList<>();
+          apis = syncedApis;
           final SpecProvider authcCommands = Utils.getSpec("cluster.security.authentication.Commands");
           final SpecProvider authzCommands = Utils.getSpec("cluster.security.authorization.Commands");
-          apis.add(new ReqHandlerToApi(this, Utils.getSpec("cluster.security.authentication")));
-          apis.add(new ReqHandlerToApi(this, Utils.getSpec("cluster.security.authorization")));
+          syncedApis.add(new ReqHandlerToApi(this, Utils.getSpec("cluster.security.authentication")));
+          syncedApis.add(new ReqHandlerToApi(this, Utils.getSpec("cluster.security.authorization")));
           SpecProvider authcSpecProvider = () -> {
             AuthenticationPlugin authcPlugin = cores.getAuthenticationPlugin();
-            return authcPlugin != null && authcPlugin instanceof SpecProvider ?
+            return authcPlugin instanceof SpecProvider ?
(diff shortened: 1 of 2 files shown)


dsmiley:
Thanks for contributing!  Is there a reason you didn't check the authorization checkbox in the template?

Your description mentions a mutable object but I think it's the _immutability_ here that enables us to avoid a volatile.
Speaking of which I looked closer to see if these lazily created objects are immutable. SSLConfigurations maybe has an issue.  It looks okay but it's only field is an SSLCredentialProviderFactory that _should_ be immutable but technically lacks it's only field, a String providerChain, from being declared final.  That is required for "publication safety" I think.
Furthermore I think the immutability of these things should be declared with a comment as justification for the lack of volatile.

Reference: https://wiki.sei.cmu.edu/confluence/display/java/LCK10-J.+Use+a+correct+form+of+the+double-checked+locking+idiom
Reference:  https://www.cs.umd.edu/~pugh/java/memoryModel/DoubleCheckedLocking.html

dsmiley:
+1 to the code you have here.

solr/core/src/java/org/apache/solr/handler/admin/SecurityConfHandler.java
```diff
@@ -260,22 +260,25 @@ public String toString() {
...
-    if (apis == null) {
+    Collection<Api> syncedApis = apis;
+    if (syncedApis == null) {
       synchronized (this) {
-        if (apis == null) {
-          Collection<Api> apis = new ArrayList<>();
+        syncedApis = apis;
+        if (syncedApis == null) {
+          syncedApis = new ArrayList<>();
+          apis = syncedApis;
```
gus-asf: Is this correct? Doesn't this line assigning syncedApis to apis expose the partially configured empty list to the rest of the world as a non-immutable collection while the rest of this method executes (which was the original complaint)? I think without this line the patch is an improvement...(the existing code has this problem too).

solr/core/src/java/org/apache/solr/handler/admin/SecurityConfHandler.java
```diff
@@ -303,11 +306,11 @@ public String toString() {
             }
           });
 
-          this.apis = ImmutableList.copyOf(apis);
+          syncedApis = ImmutableList.copyOf(syncedApis);
```
gus-asf: Shouldn't this be assigning to this.apis so that we don't have to do this work and create a new list every time?

solr/core/src/java/org/apache/solr/handler/admin/SecurityConfHandler.java
```diff
@@ -303,11 +306,11 @@ public String toString() {
...
         }
       }
     }
-    return this.apis;
+    return syncedApis;
```
gus-asf: And here we return the mutable collection???

gus-asf: This probably didn't need too change. If the prev comment is fixed, this will return the mutable version....

solr/core/src/java/org/apache/solr/handler/admin/SecurityConfHandler.java
gus-asf: actually, I guess in the code as it stands the prior assigment will actually allow us to return the prior work, but we haven't assigned the immutable copy anywhere so the next time this is called the return value will be mutable.... If the prior line is removed *then* we repeat the work...

PR #844, opened 2019-08-23
LUCENE-8860: Make more decision on inner nodes in ShapeBoundingBoxQuery

imotov:
Relaxes the bounding box intersection check in bounding box queries on
the inner node level. For the intersection it is not required that the
query bounding box contains the entire cell bounding box. It is
sufficient for the bounding box to include at least one edge of min
bounding box and max bounding. The edge has to be on the same side of
both boxes.


lucene/sandbox/src/java/org/apache/lucene/document/LatLonShapeBoundingBoxQuery.java
```diff
@@ -42,7 +42,8 @@ public LatLonShapeBoundingBoxQuery(String field, QueryRelation queryRelation, do
   @Override
   protected Relation relateRangeBBoxToQuery(int minXOffset, int minYOffset, byte[] minTriangle,
                                             int maxXOffset, int maxYOffset, byte[] maxTriangle) {
-    return rectangle2D.relateRangeBBox(minXOffset, minYOffset, minTriangle, maxXOffset, maxYOffset, maxTriangle);
+    return rectangle2D.relateRangeBBox(minXOffset, minYOffset, minTriangle, maxXOffset, maxYOffset, maxTriangle,
+        queryRelation == QueryRelation.INTERSECTS);
```
iverase: Shouldn't this work as well for Disjoint?

lucene/sandbox/src/java/org/apache/lucene/document/XYShapeBoundingBoxQuery.java
```diff
@@ -41,7 +41,8 @@ public XYShapeBoundingBoxQuery(String field, QueryRelation queryRelation, double
   @Override
   protected PointValues.Relation relateRangeBBoxToQuery(int minXOffset, int minYOffset, byte[] minTriangle,
                                                         int maxXOffset, int maxYOffset, byte[] maxTriangle) {
-    return rectangle2D.relateRangeBBox(minXOffset, minYOffset, minTriangle, maxXOffset, maxYOffset, maxTriangle);
+    return rectangle2D.relateRangeBBox(minXOffset, minYOffset, minTriangle, maxXOffset, maxYOffset, maxTriangle,
+        queryRelation == QueryRelation.INTERSECTS);
```
iverase: Shouldn't this work as well for Disjoint?

lucene/sandbox/src/java/org/apache/lucene/geo/Rectangle2D.java
```diff
@@ -183,6 +191,33 @@ private static Relation compareBBoxToRangeBBox(final byte[] bbox,
         Arrays.compareUnsigned(maxTriangle, maxYOffset, maxYOffset + BYTES, bbox, 2 * BYTES, 3 * BYTES) <= 0) {
       return Relation.CELL_INSIDE_QUERY;
     }
+
```
iverase: This first check can be skipped for intersects? maybe add it in an else statement

iverase:
@imotov Could you add an entry in CHANGES.txt? Then I could merge.

PR #850, opened 2019-08-30
SOLR-13727: Bug fix for V2Request handling in HttpSolrClient

megancarey:
# Description

This change addresses a bug that was identified in the HttpSolrClients. Because the createMethod method just uses string.replace, it's possible that the wrong portion of the host address string is replaced. The intention is to change the start of the path from "/solr" to "/api", but if the hostname begins with "solr", it replaces the beginning of the hostname. 

For example, if the address is `http://solr-host.com:8983/solr/`, instead of changing to `http://solr-host.com:8983/api/` it will become `http:/api-host.com:8983/solr/`.

# Solution

My solution is to create a java.net URL object from the address string, replace the beginning of the path as expected, create a new URL object, and return the string value. This also ensures that the URL is properly formatted.

# Tests

There are no additional tests, but I have verified when running solr locally that this change fixes the bug.

# Checklist


yonik:
Hi Megan, I think this PR accidentally has changes from your previous PR as well.  Could you do one with just the described changes?

megancarey:
@yonik The changed files contain only my changes for this bug fix, but does have other commits attached - would it be easier if I condensed the commit history to a single commit?

yonik:
Hmmm, I do see "files changed 2", but I also see "Commits 7".  I guess I'm getting confused by the merge commits.  I wonder if using the "Squash and merge" from github will keep a linear history?  I think I'll be safe and just cherry-pick the last two commits.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrClient.java
```diff
@@ -330,6 +333,12 @@ protected ModifiableSolrParams calculateQueryParams(Set<String> queryParamNames,
     }
     return queryModParams;
   }
+  
+  private String changeV2RequestEndpoint(String basePath) throws MalformedURLException {
```
noblepaul: this could be made a `public static` method

yonik: This is a widely used client class, so we should probably be slightly more wary of adding public methods.  Perhaps package protected static... until we need it to be public for some reason.  I'll make that change when I commit.

yonik:
Committed manually to master & 8x.  Thanks Megan!

PR #851, opened 2019-09-02
LUCENE-8960: Add LatLonDocValuesPointInPolygonQuery

iverase:
Adds a new query that iterates over LatLonPoint docValues and test if the point is within a provided polygon. In addition, `LatLonPointInPolygonQuery` is rewritten so it can be used efficiently with the new added query using `IndexOrDocValuesQuery`.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/document/LatLonDocValuesPointInPolygonQuery.java
+++ b/lucene/core/src/java/org/apache/lucene/document/LatLonDocValuesPointInPolygonQuery.java
@@ -1,0 +1,146 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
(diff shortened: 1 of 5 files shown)


lucene/core/src/java/org/apache/lucene/document/LatLonDocValuesPointInPolygonQuery.java
```diff
@@ -0,0 +1,146 @@
...
+    if (polygons.length == 0) {
+      throw new IllegalArgumentException("polygons must not be empty");
+    }
+    for (int i = 0; i < polygons.length; i++) {
+      if (polygons[i] == null) {
+        throw new IllegalArgumentException("polygon[" + i + "] must not be null");
+      }
+    }
+    this.field = field;
+    this.polygons =polygons;
```
jpountz: missing space after the equals sign

PR #856, opened 2019-09-04
LUCENE-8965 SometimesConcurrentMergeScheduler

dsmiley:
The initial version here is copied with simple modifications from one I wrote at Salesforce.  I do not propose we actually commit this version, it's more of a straw-man and show & tell.  Instead of the awkward subclassing, it'd be nicer to enhance CMS directly.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/SometimesConcurrentMergeScheduler.java
+++ b/lucene/core/src/java/org/apache/lucene/index/SometimesConcurrentMergeScheduler.java
@@ -1,0 +1,147 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.lucene.index;
+
+import java.io.IOException;
+import java.util.Observable;
+import java.util.Observer;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+
+import org.apache.lucene.util.ThreadInterruptedException;
+
+/**
+ * A variant of CMS: If there are cheap merges, wait for a merge to complete before continuing. This
+ * has the benefit of greatly increasing the odds that an {@link org.apache.lucene.search.IndexSearcher}
+ * will see fewer segments. Normally, CMS does all merges concurrently, and so it won't be until the
+ * next commit that the IndexSearcher benefits from fewer segments. The trade-off is less
+ * concurrency, and there will be some delay on a segment flush for a cheap merge if present.
+ *
+ * @author dsmiley
+ * @since solr.7
+ */
+public class SometimesConcurrentMergeScheduler extends ConcurrentMergeScheduler {
+
+  private long cheapMergeThresholdBytes = 2 * 1024 * 1024; // 2MB
+  private long cheapMaxMergeWaitTimeoutMs = TimeUnit.SECONDS.toMillis(2);
+
+  public long getCheapMergeThresholdBytes() {
+    return cheapMergeThresholdBytes;
+  }
+
+  /** Limit cheap merges to those less than this size in megabytes. */
+  public SometimesConcurrentMergeScheduler setCheapMergeThresholdMB(double v) {
+    v *= 1024 * 1024;
+    cheapMergeThresholdBytes = (v > Long.MAX_VALUE) ? Long.MAX_VALUE : (long) v;
+    return this;
+  }
+
+  /**
+   * If a cheap merge is taking longer than this amount time to finish, give up waiting for it. If
+   * this threshold is reached often, then the system should probably be tuned further.
+   */
+  public SometimesConcurrentMergeScheduler setCheapMaxMergeWaitTimeoutMs(long cheapMaxMergeWaitTimeoutMs) {
+    this.cheapMaxMergeWaitTimeoutMs = cheapMaxMergeWaitTimeoutMs;
+    return this;
+  }
+
+  /** Mechanism to notify when merges complete. */
+  private final Observable observableMerge = new Observable() {
+    @Override
+    public void notifyObservers(Object arg) {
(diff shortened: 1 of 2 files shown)


lucene/core/src/java/org/apache/lucene/index/SometimesConcurrentMergeScheduler.java
```diff
@@ -0,0 +1,147 @@
...
+ * next commit that the IndexSearcher benefits from fewer segments. The trade-off is less
+ * concurrency, and there will be some delay on a segment flush for a cheap merge if present.
+ *
+ * @author dsmiley
+ * @since solr.7
+ */
+public class SometimesConcurrentMergeScheduler extends ConcurrentMergeScheduler {
+
+  private long cheapMergeThresholdBytes = 2 * 1024 * 1024; // 2MB
+  private long cheapMaxMergeWaitTimeoutMs = TimeUnit.SECONDS.toMillis(2);
```
gus-asf: Hard coded timeout...

gus-asf: oops see setter below :)

lucene/core/src/java/org/apache/lucene/index/SometimesConcurrentMergeScheduler.java
```diff
@@ -0,0 +1,147 @@
...
+
+/**
+ * A variant of CMS: If there are cheap merges, wait for a merge to complete before continuing. This
+ * has the benefit of greatly increasing the odds that an {@link org.apache.lucene.search.IndexSearcher}
+ * will see fewer segments. Normally, CMS does all merges concurrently, and so it won't be until the
+ * next commit that the IndexSearcher benefits from fewer segments. The trade-off is less
+ * concurrency, and there will be some delay on a segment flush for a cheap merge if present.
+ *
+ * @author dsmiley
+ * @since solr.7
```
danmuzi: I think lucene,x or common(just 7.x, 8.x, ...) version type is right.
Because this module is lucene-core.

lucene/core/src/java/org/apache/lucene/index/SometimesConcurrentMergeScheduler.java
```diff
@@ -0,0 +1,147 @@
...
+import org.apache.lucene.util.ThreadInterruptedException;
+
+/**
+ * A variant of CMS: If there are cheap merges, wait for a merge to complete before continuing. This
+ * has the benefit of greatly increasing the odds that an {@link org.apache.lucene.search.IndexSearcher}
+ * will see fewer segments. Normally, CMS does all merges concurrently, and so it won't be until the
+ * next commit that the IndexSearcher benefits from fewer segments. The trade-off is less
+ * concurrency, and there will be some delay on a segment flush for a cheap merge if present.
+ *
+ * @author dsmiley
```
danmuzi: As I know, Lucene project doesn't allow @author tag in Javadoc.
(I saw it in check-source-patterns.groovy file)

lucene/core/src/java/org/apache/lucene/index/SometimesConcurrentMergeScheduler.java
```diff
@@ -0,0 +1,147 @@
...
+
+  private long cheapMergeThresholdBytes = 2 * 1024 * 1024; // 2MB
+  private long cheapMaxMergeWaitTimeoutMs = TimeUnit.SECONDS.toMillis(2);
+
+  public long getCheapMergeThresholdBytes() {
+    return cheapMergeThresholdBytes;
+  }
+
+  /** Limit cheap merges to those less than this size in megabytes. */
+  public SometimesConcurrentMergeScheduler setCheapMergeThresholdMB(double v) {
```
danmuzi: Personally, parameter name "mb" seems more straightforward to me.

lucene/core/src/java/org/apache/lucene/index/SometimesConcurrentMergeScheduler.java
```diff
@@ -0,0 +1,147 @@
...
+          // This is what IndexWriter javadocs says we should do; we don't set interrupt state.
+          throw new ThreadInterruptedException(e);
+        }
+      }
+    } finally {
+      observableMerge.deleteObserver(observer);
+    }
+  }
+
+  private boolean hasCheapMerge() throws IOException {
```
danmuzi: We can remove "throws IOException".

lucene/core/src/java/org/apache/lucene/index/SometimesConcurrentMergeScheduler.java
```diff
@@ -0,0 +1,147 @@
...
+    }
+  }
+
+  // called by a CMS MergeThread to do the actual merging
+  @Override
+  protected void doMerge(IndexWriter writer, MergePolicy.OneMerge merge) throws IOException {
+    super.doMerge(writer, merge);
+    observableMerge.notifyObservers(merge);
+  }
+
```
danmuzi: The new line here seems unnecessary.

lucene/core/src/java/org/apache/lucene/index/SometimesConcurrentMergeScheduler.java
```diff
@@ -0,0 +1,147 @@
...
+   * If a cheap merge is taking longer than this amount time to finish, give up waiting for it. If
+   * this threshold is reached often, then the system should probably be tuned further.
+   */
+  public SometimesConcurrentMergeScheduler setCheapMaxMergeWaitTimeoutMs(long cheapMaxMergeWaitTimeoutMs) {
+    this.cheapMaxMergeWaitTimeoutMs = cheapMaxMergeWaitTimeoutMs;
+    return this;
+  }
+
+  /** Mechanism to notify when merges complete. */
+  private final Observable observableMerge = new Observable() {
```
danmuzi: As far as I know, Observable and Observer are deprecated classes since Java 9.
https://cr.openjdk.java.net/~iris/se/11/latestSpec/api/java.base/java/util/Observable.html
https://cr.openjdk.java.net/~iris/se/11/latestSpec/api/java.base/java/util/Observer.html
Is there any other way?

lucene/core/src/java/org/apache/lucene/index/SometimesConcurrentMergeScheduler.java
dsmiley: Whoops; it was a mistake that I left this when copying from internal work.  Different coding practices.

lucene/core/src/java/org/apache/lucene/index/SometimesConcurrentMergeScheduler.java
dsmiley: It could be done manually fairly easy; no magic there.  This was backed out in Lucene/Solr already in a couple spots as part of the Java 11 upgrade, if I recall.

Any way, this use of Observer/Observable in the particular way here is a hack to work around not wanting to modify Lucene from a 3rd party codebase.  By open-sourcing this directly to Lucene/Solr, this opens up options for cleaner solutions, such as making trivial changes to the base class.  That's one path.  Another is to incorporate this non-concurrent mode directly into ConcurrentMergeScheduler.  That's a bigger conversation to be had in the JIRA rather than this by-line aspect, though.

PR #858, opened 2019-09-05
Harden Up TestDiversifiedTopDocsCollector

atris:
TestDiversifiedTopDocsCollector.testInvalidArguments should check for
exceptions and corresponding messages, post LUCENE-8905


jpountz:
Thanks @atris.

PR #862, opened 2019-09-06
LUCENE-8971: Enable constructing JapaneseTokenizer with custom dictio…

msokolov:
…nary

# Description

Extends the API of JapaneseTokenizer so it can accept a dictionary other than the built-in one. The built-in dictionary remains the default, so existing usage is unchanged; this just opens up the possibility of supplying a different dictionary/language model to use when tokenizing Japanese.

### DIFF
--- a/lucene/analysis/kuromoji/src/java/org/apache/lucene/analysis/ja/JapaneseTokenizer.java
+++ b/lucene/analysis/kuromoji/src/java/org/apache/lucene/analysis/ja/JapaneseTokenizer.java
@@ -202,7 +202,7 @@ public final class JapaneseTokenizer extends Tokenizer {
   }
 
   /**
-   * Create a new JapaneseTokenizer.
+   * Create a new JapaneseTokenizer using the system and unknown dictionaries shipped with Lucene.
    *
    * @param factory the AttributeFactory to use
    * @param userDictionary Optional: if non-null, user dictionary.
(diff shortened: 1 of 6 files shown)


msokolov:
Also back-ported to 8x branch - this was a clean cherry-pick, so I didn't post a PR.

jimczi:
Can we wait that we agreed on the full plan before backporting to 8x branch ? I don't see the full picture  of this, will you add a way to build a custom dictionary in the kuromoji module ? Currently this change is not usable since the dictionary is shipped within the jar and there is no public code to build another one ?

msokolov:
Jim, I'm sorry - there was no comment for several days, so I assumed it was OK to push. Let's discuss, and if there's a problem I can revert? DictionaryBuilder was moved into the main jar (no longer part of tools) in LUCENE-8871, so I think the software is all there - we have been using it in an automated build in our system with a patched version of Lucene. Arguably we could/should have documentation around this, that is clearly lacking, but given the complexity of this "feature" I feel it is pretty expert.

msokolov:
But to answer your question a little more fully: the way this works is, you use DictionaryBuilder, in the same way that the build system does it to build the built-in dictionary. Then you put that dictionary on your classpath (or on the filesystem), and pass that path to the JapaneseTokenizer constructor.

jimczi:
> Jim, I'm sorry - there was no comment for several days, so I assumed it was OK to push

No problem. 

> DictionaryBuilder was moved into the main jar

I forgot that the DictionaryBuilder has moved into the main jar. Should it be marked experimental then ? The fact that we ship a single dictionary within the jar also ensures that it is built from the same version but this change breaks this assumption. What kind of compatibility are we expecting here ? Should we require users to rebuild binary dictionary on each minor version ?

> Let's discuss, and if there's a problem I can revert?

Maybe give a few more days for @mocobeta to review as well ?

msokolov:
> Should it be marked experimental then ? The fact that we ship a single dictionary within the jar also ensures that it is built from the same version but this change breaks this assumption. What kind of compatibility are we expecting here ? Should we require users to rebuild binary dictionary on each minor version ?

Yes, these are good questions. I think experimental makes sense for this given that we are not providing detailed documentation and really only experts with knowledge of NLP will ever use this. With expert features there is no compatibility guarantee, so I think that rebuilding with each version would be the recommended policy. I would think users would be well-advised to rebuild whenever they build their software, treating the Kuromoji dictionary as a binary artifact produced from (textual dictionary) source code. Does that make sense?

mocobeta:
As far as LUCENE-8816 ("Decouple Kuromoji's morphological analyser and its dictionary") and related or sub issues, I do not intend to backport them to the 8x branch. I feel like that it's a too big change to be included within minor releases, and exhaustive functional/performance tests will be needed for it.
Sorry for not providing clear whole picture so far, I could not take much time for this but it's of course always in my mind.

I think it's ok to merge minor refactors for Kuromoji to 8x branch, if they are just code clean-ups or good for 8x users. I don't have any opinions about shipping experimental changes to 8x (I have no idea about what "experimental" means on this project).

mocobeta:
Just to be clear, to my understanding, this change has no effects on Solr or Elasticsearch built on Lucene 8.x so only vanilla Lucene users or advanced plugin developers can test or experiment it.
I have questions:
- Do Lucene users/developers benefit from this change if we ship this with 8.x, without LUCENE-8816?
- If so, can we inform the users about this functionality through some documentation, or do you have customers who need it immediately (I mean, are there users who want to test it)?

msokolov:
Yes, we're using this in production today to incorporate a custom dictionary. I think there is enough with this change and LUCENE-8871 that an expert user can build a new Kuromoji FST given the proper input files trained on some corpus using another system.

mocobeta:
Ok, thanks for explaining. Then would it be better that the newly added constructor Javadoc includes that it's experimental and currently provided only for experts?

msokolov:
Yes, that makes sense. I'll post a new CR soon with updated javadocs

PR #864, opened 2019-09-09
SOLR-13101 : Shared storage support in SolrCloud

megancarey:
# Description

This PR is being opened to expose the code for integrating SolrCloud with a shared blobstore. For more details, see the attached JIRA or look at `lucene-solr/solr/README-BLOB.md`.

# Solution

N/A

# Tests

Find tests in the test directory (`solr/core/src/test/org/apache/solr/store/blob/`).

# Checklist


solr/core/src/java/org/apache/solr/store/blob/client/BlobstoreProviderType.java
```diff
@@ -0,0 +1,66 @@
...
+    String provider = System.getenv(BLOB_STORAGE_PROVIDER_ENV);
+
+    if (StringUtils.isEmpty(provider)) {
+      log.error("BlobstoreProviderType: blob storage provider is not defined in environment variable "
+          + BLOB_STORAGE_PROVIDER_ENV + ". Falling back to local file system.");
+      return LOCAL_FILE_SYSTEM;
+    }
+
+    try {
+      log.info("BlobstoreProviderType: Blob storage provider type configured is " + provider);
```
eribeiro:
```suggestion
      log.info("BlobstoreProviderType: Blob storage provider type configured is {}", provider);
```

solr/core/src/java/org/apache/solr/store/blob/metadata/ServerSideMetadata.java
```diff
@@ -0,0 +1,190 @@
...
+     */
+    public boolean isSameDirectoryContent(Directory coreDir) throws NoSuchAlgorithmException, IOException {
+        return directoryHash.equals(getSolrDirectoryHash(coreDir));
+    }
+
+    /**
+     * Computes a hash of a Solr Directory in order to make sure the directory doesn't change as we pull content into it (if we need to
+     * pull content into it)
+     */
+    private String getSolrDirectoryHash(Directory coreDir) throws NoSuchAlgorithmException, IOException {
```
eribeiro: I would suggest to use a [Merkle Tree](https://en.wikipedia.org/wiki/Merkle_tree) because you have flexible and fine grained view (if necessary) of the files that changed. It also can be serialized on disk and stored on blob store if necessary. See the quick-and-dirty implementation I did here: https://gist.github.com/eribeiro/39ff8b73c43d453edd041bf1305425e0 (it is _really_ quick and dirty and slow, even tough much time is dominated by stream().map() call. This snippet outputs a tree as a json snippet that can be copy and pasted here (https://www.sitepoint.com/demos/online-json-tree-viewer/) or here (https://vanya.jp.net/vtree/) for example of usage.

solr/core/src/java/org/apache/solr/store/blob/client/BlobstoreProviderType.java
murblanc: Have you timed the execution of both options? I believe your suggestion is slower than the string concatenation.
Also I find the code harder to read because the values are "far" from where they are used.

solr/core/src/java/org/apache/solr/store/blob/provider/BlobStorageProvider.java
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * Class that provides access to the shared storage client (blob client) and
+ * handles initiation of such client. This class serves as the provider for all
+ * blob store communication channels.
+ */
+public class BlobStorageProvider {
+
+  private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+
+  private CoreStorageClient storageClient;
```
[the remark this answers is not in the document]
murblanc: Good catch.
I think if the first getClient() just does:
return getClient(BlobstoreProviderType.getConfiguredProvider());

Then we're ok. I believe the check of storageClient not null was made as an "optimization" here. We can also synchronize the first getClient(), but I don't like to mix volatile and monitors for access to the same object.

solr/core/src/java/org/apache/solr/store/blob/client/LocalStorageClient.java
```diff
@@ -0,0 +1,259 @@
...
+    
+  /**
+   * On the local FS implementation we'll just delete blob files individually
+   */
+  @Override
+  public void deleteBlobs(Collection<String> paths) throws BlobException {
+    try {
+      for (String blobPath : paths) {
+        final File blobFile = new File(getBlobAbsolutePath(blobPath));
+        blobFile.delete();
```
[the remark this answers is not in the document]
murblanc: I believe it was in case of IOException. If the delete fails locally it's not really an issue, as file names are unique and the local file system is just to help dev/debug/tests, not meant to really be used.

murblanc:
Thanks @eribeiro for having looked at this code and providing comments. We'll definitely take them into account and fix the issues you pointed out.

andyvuong:
Thanks @eribeiro for the feedback. This is very much WIP and engagement from the community is very appreciated as we iterate on this work! Definitely feel free to do an in-depth look at the code :)

solr/core/src/java/org/apache/solr/store/blob/client/LocalStorageClient.java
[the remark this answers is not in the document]
eribeiro: I see. But I can see people deploying NAS or NTFS mounts as shared index locations. As part of the community, I can see myself resorting to this. :) Other possibilities are HDFS and NoSQL like Cassandra, maybe? 

So, if local file system is not mean to be used in production there should be a clear way of stating/preventing this somehow. I feel people will probably (ab)use local file system as a blob storage once this PR is merged. /cc andyvuong

solr/core/src/java/org/apache/solr/store/blob/client/BlobstoreProviderType.java
eribeiro: Good question! :) I'd have to setup a jmh test harness to see.:thinking: But see, interpolation is being used in other files of this PR, so maybe reach uniformity?

yonik:
OK folks, I've created copy of this PR as a branch in the ASF repo for future work:
https://github.com/apache/lucene-solr/tree/jira/SOLR-13101
Feel free to either reuse that JIRA for small stuff or open new subtask JIRA issues as desired.

PR #865, opened 2019-09-10
LUCENE-8973: XYRectangle2D should work on float space

iverase:
New implementation of XYRectangle 2D that works on the float space. Because the encoding is not lineal, working on the encoding space changes the spatial relationship between the bounding box and the other objects.


iverase:
Note: Should XYRectangle holds the values as floats instead of doubles?

lucene/sandbox/src/java/org/apache/lucene/geo/XYRectangle2D.java
```diff
@@ -16,42 +16,156 @@
...
     sb.append(" TO ");
-    sb.append(decode(maxY));
+    sb.append(maxY);
     sb.append(")");
     return sb.toString();
   }
+
+  /** Builds a Rectangle2D from rectangle */
+  public static XYRectangle2D create(XYRectangle rectangle) {
+    return new XYRectangle2D((float)rectangle.minX, (float)rectangle.maxX, (float)rectangle.minY, (float)rectangle.maxY);
```
jpountz: I don't think this is right, float casts will round to the nearest float while we should actually round towards zero so that queries are accurate in the encoded space?

iverase: Not sure if the encoding works as you said. I test the following:

```
  public void testEncoding() {
   double val = ShapeTestUtil.nextDouble();
   double castVal = (double)(float) val;
   double qVal = XYEncodingUtils.decode(XYEncodingUtils.encode(val));
   assertEquals(castVal, qVal, 0.0);
  }
```

This seems to hold true.

iverase:
I have updated the PR so the XYRectangle2D is now a Component2D

iverase:
Fixed in #872

PR #868, opened 2019-09-10
LUCENE-8975: Code Cleanup: Use entryset for map iteration wherever possible.

KoenDG:
https://issues.apache.org/jira/browse/LUCENE-8975

Some simple cleanup. No need to spend company/personal time reviewing this, would be good if someone had some spare time. If you don't, feel free to close this tab.

It's a very simple PR: just changing `keySet()` to `entrySet()` **when possible**. In general just the better choice. Only really a performance gain when going over a TreeMap(as I recall at least), but in general a good idea to consistent with it.

2 exceptions where it gets changed to `values()`.

All changes done automatically by Intellij.

And 1 String construction in a loop that got turned into a Stringbuilder because that's just wasteful.

That's over in solr/core/src/java/org/apache/solr/search/facet/FacetRequest.java

All tests have run locally and succeeded.

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -166,6 +166,8 @@ Other
 
 * LUCENE-8758: QuadPrefixTree: removed levelS and levelN fields which weren't used. (Amish Shah)
 
+* LUCENE-8975: Code Cleanup: Use entryset for map iteration wherever possible.
+
 ======================= Lucene 8.2.0 =======================
 
 API Changes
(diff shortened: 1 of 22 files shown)


lucene/CHANGES.txt
```diff
@@ -65,6 +65,8 @@ Other
 
 * LUCENE-8768: Fix Javadocs build in Java 11. (Namgyu Kim)
 
+* LUCENE-8975: Code Cleanup: Use entryset for map iteration wherever possible.
+
```
jpountz: can you move it under `Lucene 8.3.0` instead of `Lucene 9.0.0`, I plan to backport this to 8.x

KoenDG: Done

PR #872, opened 2019-09-11
LUCENE-8620: LatLonShape contains

iverase:
LatLonShape's implementation for spatial relationship CONTAINS.

superseed #608


nknize:
Dug into this quite a bit and it LGTM. I think we can probably iterate on the API but the overall logic looks good.

jpountz:
> I think we can probably iterate on the API

@nknize I'm curious, what do you have in mind?

PR #874, opened 2019-09-11
LUCENE-8976: Use exact distance between point and bounding rectangle in FloatPointNearestNeighbor

iverase:
The algorithm uses now the exact minimum distance between a point and a bounding rectangle to navigate the tree.

### DIFF
--- a/lucene/sandbox/src/java/org/apache/lucene/document/FloatPointNearestNeighbor.java
+++ b/lucene/sandbox/src/java/org/apache/lucene/document/FloatPointNearestNeighbor.java
@@ -19,7 +19,6 @@ package org.apache.lucene.document;
 
 import java.io.IOException;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.List;
 import java.util.PriorityQueue;
 
@@ -46,7 +45,6 @@ public class FloatPointNearestNeighbor {
     final byte[] minPacked;
     final byte[] maxPacked;
     final BKDReader.IndexTree index;
-
     /** The closest possible distance^2 of all points in this cell */
     final double distanceSquared;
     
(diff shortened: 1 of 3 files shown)


lucene/sandbox/src/java/org/apache/lucene/document/FloatPointNearestNeighbor.java
```diff
@@ -75,132 +73,75 @@ public String toString() {
...
     final PriorityQueue<NearestHit> hitQueue;
     final float[] origin;
-    private int dims;
-    private int updateMinMaxCounter;
-    private float[] min;
-    private float[] max;
-
+    final private int dims;
+    double worstNearestDistanceSquared = Double.POSITIVE_INFINITY;
+    int worstNearestDistanceDoc =Integer.MAX_VALUE;
```
jpountz: missing space after the equal sign

jpountz: also maybe call it "bottom" (as in bottom of the priority queue) instead of "worst", which suggests it is the worst distance across all docs?

lucene/sandbox/src/java/org/apache/lucene/document/FloatPointNearestNeighbor.java
```diff
@@ -306,44 +250,36 @@ public String toString() {
...
-      min[i] = FloatPoint.decodeDimension(minPackedValue, offset);
-      max[i] = FloatPoint.decodeDimension(maxPackedValue, offset);
-      if (insideCell) {
-        if (value[i] < min[i] || value[i] > max[i]) {
-          insideCell = false;
+      double min = FloatPoint.decodeDimension(minPackedValue, offset);
+      if (value[i] < min) {
+        double diff = min - (double)value[i];
+        sumOfSquaredDiffs += diff * diff;
+        if (worstNearestDistanceSquare < sumOfSquaredDiffs) {
```
jpountz: I doubt this shortcut actually helps, does it?

lucene/sandbox/src/java/org/apache/lucene/document/FloatPointNearestNeighbor.java
```diff
@@ -306,44 +250,36 @@ public String toString() {
...
+    //System.out.println(visitor.comp);
     return hits;
   }
 
-  private static double approxBestDistanceSquared(byte[] minPackedValue, byte[] maxPackedValue, float[] value) {
-    boolean insideCell = true;
-    float[] min = new float[value.length];
-    float[] max = new float[value.length];
-    double[] closest = new double[value.length];
+  private static double pointToRectangleDistanceSquared(byte[] minPackedValue, byte[] maxPackedValue, float[] value, double worstNearestDistanceSquare) {
```
jpountz: distanceSquare**d**?

lucene/sandbox/src/java/org/apache/lucene/document/FloatPointNearestNeighbor.java
iverase: Not really, it does not make any difference. I remove it for clarity.

lucene/sandbox/src/java/org/apache/lucene/document/FloatPointNearestNeighbor.java
iverase: renamed

iverase:
In my test this approach looks faster than the previous approach, in particular for low number of dimensions. Still what it seems memory inefficient is the cloning of the index tree every time we recurse down one inner node.

PR #877, opened 2019-09-12
LUCENE-8978: Maximal Bottom Score Based Early Termination

atris:
This commit introduces a mechanism for early termination where,
for indices sorted by relevance, hits are collected in per thread
queue. Full PQs publish their bottom values and the maximum of
them is then used by the corresponding collectors to filter further
hits


atris:
I ran luceneutil with concurrent mode and wikimedium2m -- no degradation to QPS and some tasks show a reduction of around 30% in latencies. Full results at:

https://issues.apache.org/jira/browse/LUCENE-8978?page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel&focusedCommentId=16929030#comment-16929030

lucene/core/src/java/org/apache/lucene/search/BottomValueChecker.java
```diff
@@ -0,0 +1,56 @@
...
+ */
+abstract class BottomValueChecker<T> {
+  /** Maintains global bottom score as the maximum of all bottom scores */
+  private static class MaximumBottomScoreChecker extends BottomValueChecker<Float> {
+    private final AtomicInteger globalBottomValue = new AtomicInteger();
+    private final AtomicBoolean bottomValueAvailable = new AtomicBoolean();
+
+    @Override
+    public void updateThreadLocalBottomValue(Float value) {
+      globalBottomValue.updateAndGet(currentValue -> Float.intBitsToFloat(currentValue) < value ? Float.floatToIntBits(value) : currentValue);
```
jimczi: Instead of using an AtomicInteger you could maybe use a double checked locking with a volatile ? Something like:
```
public void updateThreadLocalBottomValue(float value) {
   if (value <= maxMinScore) {
     return;
   }
   synchronized (this) {
       if (value > maxMinScore) {
          maxMinScore = value;
       }
  }
}
```
This would avoid the float to int conversion and should limit contention ?

lucene/core/src/java/org/apache/lucene/search/TopScoreDocCollector.java
```diff
@@ -156,6 +187,10 @@ public void collect(int doc) throws IOException {
           pqTop.score = score;
           pqTop = pq.updateTop();
           updateMinCompetitiveScore(scorer);
```
jimczi: We should propagate the global minimum score to the underlying scorer (`Scorer#setMinCompetitiveScore`)  too. The optimization above that compares the current doc with the global minimum score is a good one but propagating the global minimum score should be more effective since it will allow skipping hits directly in the scorer.

atris: @jimczi That breaks `testIndexSearcher.testCount` though

lucene/core/src/java/org/apache/lucene/search/BottomValueChecker.java
atris: Fixed, thanks

lucene/core/src/java/org/apache/lucene/search/TopScoreDocCollector.java
jimczi: I don't understand how it could break this test. We should propagate the min score only if the total hit threshold is reached and this test indexes 100 docs only while the default threshold is 10,000. So we should never skip hits in this test, am I missing something ?

atris: Hmm, yeah, not sure what happened. I beasted the tests again -- came in clean

atris:
@jimczi Thanks for the comments -- updated the PR. Please let me know your thoughts

lucene/core/src/java/org/apache/lucene/search/BottomValueChecker.java
```diff
@@ -0,0 +1,63 @@
...
+    public void updateThreadLocalBottomValue(Float value) {
+      if (value <= maxMinScore) {
+        return;
+      }
+      synchronized (this) {
+        if (value > maxMinScore) {
+          maxMinScore = value;
+        }
+      }
+      bottomValueAvailable.compareAndSet(false, true);
```
jimczi: Is this needed ? The default `maxMinScore` is 0 so you can just check that in the collector ?

lucene/core/src/java/org/apache/lucene/search/TopScoreDocCollector.java
```diff
@@ -74,11 +75,25 @@ public void collect(int doc) throws IOException {
...
             if (totalHitsRelation == TotalHits.Relation.EQUAL_TO && hitsThresholdChecker.isThresholdReached()) {
               // we just reached totalHitsThreshold, we can start setting the min
               // competitive score now
               updateMinCompetitiveScore(scorer);
+
+              // Since the queue is prepopulated with sentinel objects, getting here means that the local
+              // priority queue is full
+              if (bottomValueChecker != null) {
+                bottomValueChecker.updateThreadLocalBottomValue(pqTop.score);
+                scorer.setMinCompetitiveScore((float) bottomValueChecker.getBottomValue());
```
jimczi: Can you move this to `updateMinCompetitiveScore` ? If the global minimum score is equals to the local minimum score we can require the next float value since we tie-break on doc id and collect in doc id order of the segment. So we need to reconcile the local minimum score and the global one to call setMinCompetitiveScore only once.

lucene/core/src/java/org/apache/lucene/search/TopScoreDocCollector.java
```diff
@@ -74,11 +75,25 @@ public void collect(int doc) throws IOException {
           totalHits++;
           hitsThresholdChecker.incrementHitCount();
 
-          if (score <= pqTop.score) {
```
jimczi: I find this simpler:
`if (score <= pqTop.score && (bottomValueChecker == null || score < bottomValueChecker.getBottomValue()) {`
This is redundant for the segment that holds the current max minimum score but we can optimize this case later, when the logic is fully implemented.
When comparing the global minimum score we need to use lesser than `<` rather than `<=`. This is important since we tiebreak on document id and the global minimum score can come from a document id that is **after** this  local segment. We can also optimize this part and record the global document id that is associated with the current max score in the BottomValueChecker but that's not necessary in the first iteration IMO.

lucene/core/src/java/org/apache/lucene/search/BottomValueChecker.java
```diff
@@ -0,0 +1,63 @@
...
+
+package org.apache.lucene.search;
+
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * Maintains the bottom value across multiple collectors
+ */
+abstract class BottomValueChecker<T> {
```
jimczi: Why is a generic needed ?

lucene/core/src/java/org/apache/lucene/search/BottomValueChecker.java
```diff
@@ -0,0 +1,63 @@
...
+  /** Maintains global bottom score as the maximum of all bottom scores */
+  private static class MaximumBottomScoreChecker extends BottomValueChecker<Float> {
+    private volatile float maxMinScore;
+    private final AtomicBoolean bottomValueAvailable = new AtomicBoolean();
+
+    @Override
+    public void updateThreadLocalBottomValue(Float value) {
```
jimczi: Just use a simple `float` ?

lucene/core/src/java/org/apache/lucene/search/BottomValueChecker.java
```diff
@@ -0,0 +1,63 @@
...
+      synchronized (this) {
+        if (value > maxMinScore) {
+          maxMinScore = value;
+        }
+      }
+      bottomValueAvailable.compareAndSet(false, true);
+    }
+
+    @Override
+    public Float getBottomValue() {
```
jimczi: Same here, a `float` is enough

lucene/core/src/java/org/apache/lucene/search/TopScoreDocCollector.java
```diff
@@ -89,8 +104,12 @@ public void collect(int doc) throws IOException {
           pqTop.score = score;
           pqTop = pq.updateTop();
           updateMinCompetitiveScore(scorer);
-        }
 
+          if (bottomValueChecker != null && bottomValueChecker.isGlobalBottomValueAvailable()) {
```
jimczi: Same here, this should move `updateMinCompetitiveScore`, as is it doesn't respect the total hits threshold and the  bottom values might be greater than the local bottom score.

lucene/core/src/java/org/apache/lucene/search/TopScoreDocCollector.java
atris: If the global minimum score is equal to the local minimum score, should we not be discarding this document, because, as you said, we collect in docID order?

lucene/core/src/java/org/apache/lucene/search/TopScoreDocCollector.java
atris: Agreed, +1.

lucene/core/src/java/org/apache/lucene/search/BottomValueChecker.java
atris: Since we will need to re use the class for `TopFieldCollector` as well.

lucene/core/src/java/org/apache/lucene/search/BottomValueChecker.java
atris: Removed

atris:
@jimczi Updated, please see.

One thing we could do is cache the bottom docID (doc + docBase) in BottomValueChecker, but as you said, it might not be worth it.

lucene/core/src/java/org/apache/lucene/search/BottomValueChecker.java
jimczi: Let's resolve this later, we're still not sure that we'll use the same mechanism so we should focus on the top score docs collector where we only need a float at the moment ?

lucene/core/src/java/org/apache/lucene/search/TopScoreDocCollector.java
```diff
@@ -155,6 +175,11 @@ public void collect(int doc) throws IOException {
           pqTop.doc = doc + docBase;
           pqTop.score = score;
           pqTop = pq.updateTop();
+
+          if (bottomValueChecker != null && bottomValueChecker.getBottomValue() > 0) {
```
jimczi: The update below should be called only when the total hits threshold is used and the queue is full so it would be easier to set it in `updateMinCompetitiveScore` like we do to propagate the min score in the scorer ?

lucene/core/src/java/org/apache/lucene/search/TopScoreDocCollector.java
```diff
@@ -138,14 +152,20 @@ public void collect(int doc) throws IOException {
           if (score > after.score || (score == after.score && doc <= afterDoc)) {
             // hit was collected on a previous page
             if (totalHitsRelation == TotalHits.Relation.EQUAL_TO && hitsThresholdChecker.isThresholdReached()) {
+              // Since the queue is prepopulated with sentinel objects, getting here means that the local
+              // priority queue is full
+              if (bottomValueChecker != null) {
```
jimczi: It would be easier to put this logic in updateMinCompetitiveScore as well ?

atris: Agreed -- The reason I did not put it in `updateMinCompetitiveScore` is because the function already became significantly heavy by the additions in this PR. Added now

atris:
@jimczi Updated

jimczi:
Can you run another benchmark now that we propagate the global min score to the scorer ?

atris:
@jimczi Here is the link:

https://issues.apache.org/jira/browse/LUCENE-8978?focusedCommentId=16929277&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-16929277

atris:
Hmm, the precommit failure seems to be coming from `JdbcDataSource.java`?

mikemccand:
> I ran luceneutil with concurrent mode and wikimedium2m -- no degradation to QPS and some tasks show a reduction of around 30% in latencies.

We have to be careful interpreting luceneutil results: it does not run a true "red-line" (capacity) test.  When it reports QPS, that's the result of N threads running random queries over and over, not necessarily saturating CPU/IO on the server you are testing on.

In fact, it's odd you didn't see a QPS gain when you saw latency improvements -- those latency gains should have meant the search threads run more queries per second, hmm.

atris:
Well, there was not a very significant increase in the QPS ( I did see
gains of around 7%) in some scenarios but did not really pay much attention
to them — my bigger concern was to check that qps shows no degradation :)
-- 
Regards,

Atri
*l'apprenant*

atris:
@jimczi Any further thoughts on this?

jimczi:
Sorry I won't have a lot of time this week to look but I think we need to find a way to benchmark this properly. No degradation in qps is one thing but if the optimization doesn't prove to be useful there is no point adding the complexity ? How did you run the benchmark, maybe you should try with wikimedium to check with bigger segments ?

atris:
Interestingly, the latency (P99, especially) seems to be getting
significantly better.

I have triggered a wikimedium2m and wikimedium10m with
-doConcurrentSearches — hoping it should generate some more data
-- 
Regards,

Atri
*l'apprenant*

mikemccand:
We should expect to see an improvement to red-line QPS with this change, right?  I.e. it is spending less CPU per query since the thread slices are communicating with one another as they collect about how good a new hit must be to be competitive.

Versus today, where each query thread collects the full top N and only in the end does a partial merge sort to pick the total top N.

We should maybe expect query latencies (at well below red-line QPS) to get a little worse than concurrent search today, because we are adding a bit of thread contention?

atris:
I realised I had some issues with my previous runs -- they were messing up because of a hack that I had put in luceneutil and hadn't removed it (from an experiment done on a separate patch). I did a fresh checkout of luceneutil and ran `wikimedium2m` and `wikimedium10m` with concurrent searches enabled. For *all* of the tasks run, the total number of hits visited by this patch were lesser than that visited by vanilla. The difference ranged from 100 odd hits to 1/6 number of hits e.g. for  OrHighNotHigh: hit count visited by vanilla was 32112 vs this patch visited 5618 hits. No errors in difference in results were reported by luceneutil.

The QPS difference range is the same : For HighPhrase, QPS doubled in the case of `wikimedium10m`.

Full results at:

https://issues.apache.org/jira/browse/LUCENE-8978?focusedCommentId=16934206&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-16934206

atris:
Somehow, upstream merging screwed up the order of commits -- raised a new PR rebased against master:

https://github.com/apache/lucene-solr/pull/897

Closing this one

PR #882, opened 2019-09-15
Add versions.lock entry for "org.apache.yetus:audience-annotations"

anshumg:


anshumg:
Again, not sure if it's just me but adding this certainly fixes the problem. This is happening on macOS.

PR #888, opened 2019-09-17
SOLR-13774 add lucene/solr openjdk compatibility matrix to ref guide.

lw-nick:
# Description

Create a reusable build system to run Lucene/Solr ant test source code suite against different versions of OpenJDK binaries. Generate a table with the results of BUILD SUCCESSFUL or BUILD FAILED and incorporate the output into the Ref Guide here: solr/solr-ref-guide/src/solr-system-requirements.adoc

# Solution

The build system is written in Python and Bash. Several scripts configure the environment, download all the Solr sources for each version, and OpenJDK binaries from jdk.java.net. Java versions are included in the submitted documentation. Python multiprocessing is used to parallelize concurrent Solr+OpenJDK ant test runs. Params for JVM concurrency are passed through scripts to ant tasks as well giving fin-grained concurrency control. Tests that resulted in BUILD SUCCESSFUL are passed as "Y" and those resulting in BUILD FAILED are failed as "N".

# Tests

Test suites run as described in solution. For validating syntax of submitted documentation in this PR Asciidoctor was run to convert to HTML that was inspected visually.

# Checklist


solr/solr-ref-guide/src/solr-system-requirements.adoc
```diff
@@ -93,3 +93,119 @@ The success rate in our automated tests is similar with all the Java versions te
...
 * There were known issues with Kerberos with Java 9+ prior to Solr 8.1. If using 8.0, you should test in your environment.
 * Be sure to test with SSL/TLS and/or authorization enabled in your environment if you require either when using Java 9+.
+
+=== Lucene/Solr OpenJDK Compatibility
+The following compatibility matrix was generated by running an `ant test` command for each version of Solr and OpenJDK. The tests were run in a non-SSL/TLS environment without authorization enabled. A BUILD SUCCESSFUL message resulted in a "Y" and a BUILD FAILED message resulted in a "_N_".
+
+[cols="1,6*^" options="header"]
+|===
+|Lucene/Solr|OpenJDK 8|OpenJDK 9|OpenJDK 10|OpenJDK 11|OpenJDK 12|OpenJDK 13
+|*3.1.0* |_N_     |_N_     |_N_      |_N_      |_N_      |_N_
```
anshumg: I don't think it makes sense to add all versions in here. Also, `ant test` is not necessarily the best way to generate this - however, this seems very promising w.r.t. the successful test % :)

janhoy: Agree. It's cool to see the output but doesn't provide much above what we already document for each version. And the reference Guide is version specific, so to tell in RefGuide for Solr 8.3 that Solr 4.10 won't work with Java 13 is not very interesting?

I'm more interested in running tests against AdoptOpenJDK, Corretto, OracleOpenJDK etc, than against all versions of Oracles non-free options.

ErickErickson: I almost agree, but we _don't_ have this already in past versions. So anyone trying to answer "can I run Solr 6.6 on OpenJDK 9" is out of luck if we just add to the current Solr reference guide.

WDYT about making this a Wiki page and linking to it from the ref guide?

lw-nick: Thank you for providing your feedback. I have added some additional context and a couple clarifying questions below:

I believe this compatibility matrix has a benefit to Solr end users and it will be helpful in the planning of deployments. I am not able to find these specific details in the current reference guide and I think that is the best place for this information.

The primary use case is for an IT environment with many installations of Solr and different JDK versions, where migration and standardization efforts require this level of detailed planning.

The secondary use case is for anyone planning a new deployment of Solr with latest JDK version. For example, when I run ant test for Solr 8.0.0 using OpenJDK 13, I get a compilation error and build failed message. I am sure there is a workaround for this issue but it may create confusion when end users are evaluating Solr for a new project.

@anshumg The ant test suite runs many tests, and if the build is successful, then that is a good indication of compatibility in my view. The majority of the failures resulted from compilation errors. What method would you suggest for generating the compatibility matrix?

@janhoy I am happy to add other versions of commonly deployed open JDKs. The build system is extensible and new JDKs and Solr versions can be added relatively easily. If you could add one other JDK+version what would it be and why?

janhoy:
I think it belongs in confluence and add a link from refguide.

ErickErickson:
Hmmm, I think overall the idea of putting it on a Wiki page then linking to it from the ref guide makes sense. We can put some disclaimers in about testing etc. as well.

PR #889, opened 2019-09-18
LUCENE-8983: Add PhraseWildcardQuery to control multi-terms expansions in a phrase

bruno-roustant:
A generalized version of PhraseQuery, built with one or more MultiTermQuery that provides term expansions for multi-terms (one of the expanded terms must match).
Its main advantage is to control the total number of expansions across all MultiTermQuery and across all segments.
 This query is similar to MultiPhraseQuery, but it handles, controls and optimizes the multi-term expansions.  
This query is equivalent to building an ordered SpanNearQuery with a list of SpanTermQuery and SpanMultiTermQueryWrapper. But it optimizes the multi-term expansions and the segment accesses. It first resolves the single-terms to early stop if some does not match. Then it expands each multi-term sequentially, stopping immediately if one does not match. It detects the segments that do not match to skip them for the next expansions. This often avoid expanding the other multi-terms on some or even all segments. And finally it controls the total number of expansions.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/MultiTermQuery.java
+++ b/lucene/core/src/java/org/apache/lucene/search/MultiTermQuery.java
@@ -292,11 +292,14 @@ public abstract class MultiTermQuery extends Query {
    */
   protected abstract TermsEnum getTermsEnum(Terms terms, AttributeSource atts) throws IOException;
 
-  /** Convenience method, if no attributes are needed:
-   * This simply passes empty attributes and is equal to:
-   * <code>getTermsEnum(terms, new AttributeSource())</code>
+  /**
+   * Constructs an enumeration that expands the pattern term.
+   * This method should only be called if the field exists (ie,
+   * implementations can assume the field does exist).
+   * This method never returns null.
+   * The returned TermsEnum is positioned to the first matching term.
    */
-  protected final TermsEnum getTermsEnum(Terms terms) throws IOException {
+  public final TermsEnum getTermsEnum(Terms terms) throws IOException {
     return getTermsEnum(terms, new AttributeSource());
   }
 
--- a/lucene/core/src/java/org/apache/lucene/search/PhraseQuery.java
+++ b/lucene/core/src/java/org/apache/lucene/search/PhraseQuery.java
@@ -330,6 +330,22 @@ public class PhraseQuery extends Query {
       }
     }
 
+    public PostingsAndFreq(PostingsEnum postings, ImpactsEnum impacts, int position, List<Term> terms) {
+      this.postings = postings;
+      this.impacts = impacts;
+      this.position = position;
+      nTerms = terms == null ? 0 : terms.size();
+      if (nTerms > 0) {
+        Term[] terms2 = terms.toArray(new Term[0]);
+        if (nTerms > 1) {
+          Arrays.sort(terms2);
+        }
+        this.terms = terms2;
+      } else {
+        this.terms = null;
+      }
+    }
+
     @Override
     public int compareTo(PostingsAndFreq other) {
       if (position != other.position) {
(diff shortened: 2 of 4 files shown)


jpountz:
This is a bit too esoteric for lucene/core in my opinion, would it work for you if we had it in lucene/sandbox?

bruno-roustant:
Sandbox is fine for me yes.
I'll push a commit soon to fix the precommit here and I'll move it to sandbox.

uschindler:
The current way how this is done (create the MultiTermQuery termsenum) is violating the API. The method MTQ#getTermsEnum is protected, so it should never ever called from the outside. Java just allows this from the same package, but it's just incorrect. protected methods should only be called from the class itsself and its subclasses.

Expanding terms of a MTQ should be done by passing a RewriteMethod and then rewriting it.

bruno-roustant:
> Expanding terms of a MTQ should be done by passing a RewriteMethod and then rewriting it

I see. The main point of this PhraseWildcardQuery is to not rewrite to the numerous expansions. This is one of the primary goals in its design. It seems to be a blocker here.

Out of curiosity, is there a strong requirement to have the MTQ#getTermsEnum protected and not public?

uschindler:
Hi, the trick is to use a custom rewrite method that rewrites instead to a query to some arraylist or whatever. Once collected enough terms stop it. I did this for many implementations (I think there is also one in Lucene).

Nevertheless, we can still think about making getTermsEnum() public, but then we should clearly define how this method must be called. I'd prefer to add some final public convenience wrapper method in the MTQ class to allow external usage. Maybe this one should take a LeafReader instance.

Another way is to make the one without Attributes final and public, as this one looks like one to be called from the outside.

bruno-roustant:
Yes I clearly like the third option to make the getTermsEnum() without Attributes final and public.

bruno-roustant:
Can I include the MTQ#getTermsEnum(Terms) visibility change in the next commit of this PR?

I'll also make the empty "new AttributeSource()" a constant of AttributeSource, because internally AttributeSource creates 2 LinkedHashMap and an array that we don't have to repeat when calling this MTQ#getTermsEnum(Terms) without attribute.

uschindler:
No, it must be a new `AttributeSource`. For wildcards it does not matter, but some MTQ *modify* the item attributes. The cost is minimal, so please don't overoptimize!

ErickErickson:
Could I ask that a brief summary (just a few sentences) describing the decision coming out of this conversation be attached to the JIRA for posterity? 

Thanks,
Erick

dsmiley:
I'm a bit confused why this query was characterized as _esoteric_.  SpanMultiTermQueryWrapper is in core and is what people must use today, as a building block to get this functionality.

gus-asf:
Would the getTermsEnum(Terms) method **really** need to be final when it goes to public? Maybe it does, I'm not expert in this protion of the code (just happened to stumble into this ticket and read because I was curious), but often final methods cause later pain for someone trying to do something nifty. Maybe javadocs should say what's important about it being final so others don't remove it later when it gets in the way of something (or at least account for the potential problems when they do)? Also the sentiment above that the protected method shouldn't be called even when it's allowed by java should also be in the javadocs for that method. The overall gripe against java for not having a feature to enforce the convention you specify I share (this lack was actually my very first thought the first time I read about the modifiers in a book about java many years ago!), but there's no way for someone else to know that the authors are relying on that convention.

uschindler:
@gus-asf Lucene prefers "composition over inheritance", so we try to make everything final that's not protected or abstract.

gus-asf:
Ok I understand, composition over inheritance is a good principle. Not sure I agree with using final to enforce it, but that is a discussion for elsewhere and some point where i have more time.

jpountz:
> I'm a bit confused why this query was characterized as esoteric

@dsmiley I was mostly referring to how it tries to optimize for segments that don't have a term instead of rewriting to a MultiPhraseQuery. These sorts of optimizations are fragile as e.g. they'd stop working if you force-merged you index.

bruno-roustant:
Merged

PR #892, opened 2019-09-20
LUCENE-8972: Add ICUTransformCharFilter, to support pre-tokenizer ICU text transformation

magibney:


lucene/analysis/icu/src/java/org/apache/lucene/analysis/icu/ICUTransformCharFilter.java
```diff
@@ -0,0 +1,322 @@
...
+
+  /**
+   * Construct new {@link ICUTransformCharFilter} with the specified {@link Transliterator}, backed by
+   * the specified {@link Reader}.
+   * @param in input source
+   * @param transform used to perform transliteration
+   * @param maxRollbackBufferCapacityHint used to control the maximum size to which this
+   * {@link ICUTransformCharFilter} will buffer and rollback partial transliteration of input sequences.
+   * The provided hint will be converted to an enforced limit of "the greatest power of 2 (excluding '1')
+   * less than or equal to the specified value". Specifying a negative value allows the rollback buffer to
```
msokolov: Is there any reason to support negative values, rather than having the caller supply MAX_VALUE? I think elsewhere in this codebase we tend to error on negative values in cases like this.

lucene/analysis/icu/src/java/org/apache/lucene/analysis/icu/ICUTransformCharFilter.java
```diff
@@ -0,0 +1,322 @@
...
+   * @param len
+   *          Maximum number of characters to read
+   * @return The number of characters read, or -1 if the end of the stream has been reached
+   * @throws IOException
+   *           If an I/O error occurs
+   */
+  @Override
+  public int read(char[] cbuf, int off, int len) throws IOException {
+    if (off < 0) throw new IllegalArgumentException("off < 0");
+    if (off >= cbuf.length) throw new IllegalArgumentException("off >= cbuf.length");
```
msokolov: This error message probably won't be meaningful to most callers - how about something like "offset exceeeds buffer size"?

lucene/analysis/icu/src/java/org/apache/lucene/analysis/icu/ICUTransformCharFilter.java
```diff
@@ -0,0 +1,322 @@
...
+      }
+
+      int resLen = readFromIoNormalizeUptoBoundary();
+      if (resLen > 0) {
+        return outputFromResultBuffer(cbuf, off, len);
+      }
+
+      if (!readInputToBuffer() && position.start < buffer.length()) {
+        final int preStart = position.start;
+        final int preLimit = position.contextLimit = position.limit = buffer.length();
```
msokolov: I haven't seen multiple assignments elsewhere in Lucene code. Personally I'm fine with it, but for consistency maybe we should break up into multiple statements?

magibney:
Thank you for the review/feedback @msokolov! In addition to (as you mention) offset accuracy, some form of incremental transliteration with rollback is necessary for baseline support of consistent transliteration in a streaming fashion. The only alternatives I can think of are:
1. to load the entire input in memory and transliterate the entire block at once (which would lose you _all_ offset accuracy and could potentially be problematic memory-wise for large inputs), or
2. transliterate input in arbitrary-sized chunks (which would would get you offset accuracy at the level of granularity of the individual chunks).

The second of these options sounds ok for cases where you don't care about offsets, but a bigger problem with this option is that regardless of whether nominally processed "incrementally", the tail end of each chunk would in some cases undergo partial transformation yielding inconsistent results.

I was curious as well about the performance implications of rollback (as compared to the `ICUTransformTokenFilter`, which has the input pre-split and can thus afford single-pass transliteration of each chunk). I put together a _really_ quick and dirty comparison that just runs inputs with configurably-sized, whitespace-delimited tokens that consist entirely of one character. The results (at least initially, superficially) seem to suggest that the latency scales ~linearly with respect to the average length of character runs that can be completely transliterated (advancing the rollback window) -- or, put another way, the average size to which the rollback buffer content grows before being "committed". See initial performance evaluation code at:

[charFilterPerformanceTest.txt](https://github.com/apache/lucene-solr/files/3672215/charFilterPerformanceTest.txt)

It's a tricky thing to test, and I'd like to see it evaluated against real-world input, but the performance impact of incremental transliteration would also be highly dependent on the particular transliteration type (e.g., top-level `RuleBasedTransliterator` instances should never require any rollback, and should be plenty fast; some `CompoundTransliterator` instances should require little or no rollback, others (particularly ones that bundle trailing NFC transformation, like "Cyrillic-Latin") might require rollback for most non-delimiter character runs.

If we somehow know that offsets are not necessary for a given instance, there would be some optimizations possible (like initially buffering some large-but-not-too-large number of characters for a given input, with the hope that we'll reach the end of the input and can just do block transliteration).

lucene/analysis/icu/src/java/org/apache/lucene/analysis/icu/ICUTransformCharFilter.java
magibney: +1, pushed change that also changes exception type to `IndexOutOfBoundsException`, as prescribed in `Reader` class abstract method javadocs. (I had copied bounds-checking code from [ICUNormalizer2CharFilter.java](https://github.com/apache/lucene-solr/blob/master/lucene/analysis/icu/src/java/org/apache/lucene/analysis/icu/ICUNormalizer2CharFilter.java#L70))

lucene/analysis/icu/src/java/org/apache/lucene/analysis/icu/ICUTransformCharFilter.java
magibney: Thanks for pointing this out; I've pushed a fix that removes multiple assignments (here and in a couple of other places). I personally find the multiple assignments on one line to be clearer, because it's easier for me to see that there's no extra procedural logic. But it makes sense to prioritize consistency with existing code.

lucene/analysis/icu/src/java/org/apache/lucene/analysis/icu/ICUTransformCharFilter.java
magibney: +1. This was here more as a convenience targeted at the external schema config API, so I've preserved that convenience, but moved it out to the `ICUTransformCharFilterFactory` instead (to allow users who essentially want to configure "no limit" to avoid having to explicitly write something like "2147483647" in their config files). Does that sound ok?

I also noticed that there is no power of 2 greater than or equal to hint, for hint greater than `Integer.highestOneBit(Integer.MAX_VALUE)`, so I added special handling for that range of hint. It's a little fiddly, but basically we're protecting against overflow while recognizing that if the user specifies a very large value, it makes sense to resolve to the largest _legal_ value, even though it's smaller than the specified hint.

magibney:
Here's a thought: what if we provided a boolean configuration option like `assumeExternalUnicodeNormalization`. Many of these transforms work on NFD input, and produce NFC output, but they generally are configured defensively (not assuming input to be NFD, and not assuming that output will be externally converted to NFC).

This is understandable, but results in the odd situation (for example) that an analysis component like "ICUTransformFilter(Cyrillic-Latin)" would have NFC output, but _only_ for characters whose input representation matched the top-level Cyrillic-Latin filter (which is pretty restrictive). Input characters that didn't match the top-level filter would be untouched by any component of the underlying CompoundTransliterator. So if you want fully unicode-normalized output (and in the context of an analysis chain, most do), you have to separately apply post-transform NFD normalization anyway.

At best, for this ends up doing some redundant work; but for the performance case we're considering here, there are particular implications. NFC, as a trailing transformation step, is both _very_ common and _very_ active -- active in the sense that it will in many common contexts block output waiting for combining diacritics for literally almost every character. If we know we're externally applying unicode normalization over the entire output, skipping baked-in post-NFC for every transform component avoids redundant work, but more importantly avoids a common case that's virtually guaranteed to result in a substantial amount of partial transliteration, rollback, etc. I think this can be done relatively cleanly using Transliterator getElements(), toRules(false), and createFromRules(...).

I'd be curious to know what you think, @msokolov, and perhaps @rmuir?

magibney:
Tried this out, and the performance gain was indeed significant. Comparing apples to apples here:

[charFilterPerformanceTest2.txt](https://github.com/apache/lucene-solr/files/3682491/charFilterPerformanceTest2.txt)

Results (again, very quick and dirty):

```
   [junit4] Suite: org.apache.lucene.analysis.icu.TestICUTransformCharFilter
   [junit4]   2> tokenCount: 1000000, elapsed: 20244
   [junit4] OK      20.6s | TestICUTransformCharFilter.testFilterPerformanceChar
   [junit4]   2> tokenCount: 1000000, elapsed: 3040
   [junit4] OK      3.05s | TestICUTransformCharFilter.testFilterPerformanceToken
   [junit4]   2> tokenCount: 1000000, elapsed: 4339
   [junit4] OK      4.36s | TestICUTransformCharFilter.testFilterPerformanceModifiedChar

```

lucene/analysis/icu/src/java/org/apache/lucene/analysis/icu/ICUTransformCharFilter.java
```diff
@@ -0,0 +1,384 @@
...
+      if (position.start > outputCursor) {
+        return outputFromResultBuffer(cbuf, off, len);
+      }
+
+      int resLen = readFromIoNormalizeUptoBoundary();
+      if (resLen > 0) {
+        return outputFromResultBuffer(cbuf, off, len);
+      }
+
+      if (!readInputToBuffer()) {
```
msokolov: in what sort of circumstance would we fail to process any chars in readFromIo... (ie resLen == 0)? Does that mean the transliterator was still waiting for more input, but we ran out of space or completely consumed the input?

lucene/analysis/icu/src/java/org/apache/lucene/analysis/icu/ICUTransformCharFilter.java
```diff
@@ -0,0 +1,384 @@
...
+    final int preCharCount = charCount;
+    int preStart = position.start;
+    int preLimit;
+    do {
+      position.limit += nextCharLength;
+      preLimit = position.limit;
+      position.contextLimit = preLimit;
+      transform.filteredTransliterate(replaceable, position, true);
+      boolean rollbackSizeWithinBounds = true;
+      if (rollbackBuffer != null && position.start < position.limit && (rollbackSizeWithinBounds = ensureRollbackBufferCapacity())) {
```
msokolov: Is it possible to move the side effect (assignment to rollbackSizeWIthinBounds) inside this if statement, and not to do it as part of the test expression? I think if you use a `continue` in this block you can do it?

lucene/analysis/icu/src/test/org/apache/lucene/analysis/icu/TestICUTransformCharFilter.java
```diff
@@ -0,0 +1,153 @@
...
+    checkToken(Transliterator.getInstance("Han-Latin"),
+        "中国", "zhōng guó");
+  }
+  
+  public void testRollbackBuffer() throws Exception {
+    checkToken(Transliterator.getInstance("Cyrillic-Latin"),
+        "яяяяя", "âââââ"); // final NFC transform applied
+    checkToken(Transliterator.getInstance("Cyrillic-Latin"), 0,
+        "яяяяя", "a\u0302a\u0302a\u0302a\u0302a\u0302"); // final NFC transform never applied
+    checkToken(Transliterator.getInstance("Cyrillic-Latin"), 2,
```
msokolov: Can you explain why the rollback buffer needs to be so large to correctly handle this transliteration? Isn't it just a chartacter-by-character process?

lucene/analysis/icu/src/test/org/apache/lucene/analysis/icu/TestICUTransformCharFilter.java
```diff
@@ -0,0 +1,153 @@
...
+  }
+
+  private void checkToken(Transliterator transform, int maxRollbackBufferCapacity, String input, String expected) throws IOException {
+    final KeywordTokenizer input1 = new KeywordTokenizer();
+    input1.setReader(new ICUTransformCharFilter(new StringReader(input), transform, maxRollbackBufferCapacity));
+    assertTokenStreamContents(input1, new String[] { expected });
+  }
+  
+  /** blast some random strings through the analyzer */
+  public void testRandomStrings() throws Exception {
```
msokolov: Have you "beasted" this test? ie run it a lot?

lucene/analysis/icu/src/test/org/apache/lucene/analysis/icu/TestICUTransformCharFilter.java
```diff
@@ -0,0 +1,153 @@
...
+    final Transliterator transform = Transliterator.getInstance("Any-Latin");
```
msokolov: What does this transformer do? Does it change the lengths of characters? Do we adequately test the many edge cases in the charfilter? Have you for example looked at code coverage from these tests?

lucene/analysis/icu/src/java/org/apache/lucene/analysis/icu/ICUTransformCharFilter.java
```diff
@@ -0,0 +1,384 @@
...
+   *          Offset at which to start storing characters
+   * @param len
+   *          Maximum number of characters to read
+   * @return The number of characters read, or -1 if the end of the stream has been reached
+   * @throws IOException
+   *           If an I/O error occurs
+   */
+  @Override
+  public int read(char[] cbuf, int off, int len) throws IOException {
+    if (off < 0) throw new IndexOutOfBoundsException("specified negative array offset");
```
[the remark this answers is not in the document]
magibney: +1; I only left this in there because, as I had copied bounds-checking code from [ICUNormalizer2CharFilter.java](https://github.com/apache/lucene-solr/blob/master/lucene/analysis/icu/src/java/org/apache/lucene/analysis/icu/ICUNormalizer2CharFilter.java#L70) and was trying to follow precedent/change as little as possible. I'll take care of this shortly (and any analogous cases).

lucene/analysis/icu/src/java/org/apache/lucene/analysis/icu/ICUTransformCharFilter.java
```diff
@@ -0,0 +1,384 @@
...
+    if (off >= cbuf.length) throw new IndexOutOfBoundsException("specified offset exceeds buffer length");
+    if (len <= 0) throw new IndexOutOfBoundsException("non-positive length specified");
+    if (len > cbuf.length - off) throw new IndexOutOfBoundsException("requested end array index exceeds buffer length");
+
+    // !inputFinished || output remains to be flushed
```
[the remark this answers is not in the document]
magibney: That'd be ok with me. To explain the initial choice, there may be one of two things going on here: "input not finished" _or_ "output remains to be flushed" ... the comment was intended to be directly associated with the "while" condition below. This does feel strange, but the while loop is basically checking the status of the "input buffer" and the "output buffer", with both needing to be "done" before exiting the loop. I could make the comment more verbose, like
`// "expecting more input" || "output remains that has not been flushed"`
If that sounds better, I'm happy to put the comment inside the while loop if you think that'd be clearer.

lucene/analysis/icu/src/java/org/apache/lucene/analysis/icu/ICUTransformCharFilter.java
magibney: Yes, that's weird. This is because `readFromIoNormalizeUptoBoundary()` is the method that does the incremental transliteration, and we don't know _a priori_ how many characters are needed for an arbitrary transliterator to determine that it has an exact match. For example, think a dictionary-based transliterator that has two terms that share a very long common prefix -- we'd have to read input until we get past the shared prefix, which could be arbitrarily long ... and if it's longer than `READ_BUFFER_SIZE`, a single call to `readInputToBuffer()` would not supply enough characters. N.b., this is different from `getMaximumContextLength()`, which says nothing about how much must be provided to the _active_ input buffer. I found this surprising as well!

lucene/analysis/icu/src/java/org/apache/lucene/analysis/icu/ICUTransformCharFilter.java
```diff
@@ -0,0 +1,384 @@
...
+      }
+
+      if (!readInputToBuffer()) {
+        final int preStart = position.start;
+        final int preLimit;
+        final int bufferLength = buffer.length();
+        cursorUpdate:
+        {
+          if (preStart < bufferLength) {
+            // if last char is a lead surrogate, transform won't handle it properly anyway
```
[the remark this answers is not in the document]
magibney: Yes, that's the case. I agree with the approach you outline. I'll double-check, but I think this code doesn't try to _correct_ the error (i.e., doesn't trim the invalid input off the end) -- it just doesn't attempt to transform it. So the approach is basically "I don't care what you gave me, I'll do what I can with it and pass it on" -- making the transliterator totally unopinionated, and not treating it as a validator. Garbage in, garbage out, no judgment :-)
I could see throwing an error instead if you think that'd be preferable.

lucene/analysis/icu/src/java/org/apache/lucene/analysis/icu/ICUTransformCharFilter.java
```diff
@@ -0,0 +1,384 @@
...
+            return 1;
+          } else {
+            return 0; // wait for more input to be available
+          }
+        }
+      }
+    }
+  }
+
+  private int readFromIoNormalizeUptoBoundary() {
```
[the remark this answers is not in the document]
magibney: +1; or as long as we're breaking from the precedent set by ICUNormalizer2CharFilter (a fine idea), perhaps `transliterateBufferContents()` would better convey what's actually happening here?

lucene/analysis/icu/src/java/org/apache/lucene/analysis/icu/ICUTransformCharFilter.java
```diff
@@ -0,0 +1,384 @@
...
+        cursorUpdate:
+        {
+          if (preStart < bufferLength) {
+            // if last char is a lead surrogate, transform won't handle it properly anyway
+            preLimit = UTF16.isLeadSurrogate(buffer.charAt(bufferLength - 1)) ? bufferLength - 1 : bufferLength;
+            position.contextLimit = preLimit;
+            position.limit = preLimit;
+            transform.finishTransliteration(replaceable, position);
+          } else if (offsetDiffAdjust == 0) {
+            break cursorUpdate;
```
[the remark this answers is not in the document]
magibney: Oh yeah! `continue` would definitely work in this case. Will make that change.

lucene/analysis/icu/src/java/org/apache/lucene/analysis/icu/ICUTransformCharFilter.java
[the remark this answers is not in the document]
msokolov: Thanks, expanding a little helps a lot! I'm not too concerned about the placement of the comment. I tend to put them inside if the comment is describing the positive condition that is being tested for, though, "if we got here, then this is true"

lucene/analysis/icu/src/java/org/apache/lucene/analysis/icu/ICUTransformCharFilter.java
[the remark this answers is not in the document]
msokolov: I think we should provide whatever behavior the Transliterator would if you gave *it* malformed input as a String, and ideally we'd do that without any special error-handling code of our own, just for simplicity's sake, and in order to track Transliterator's implementation should it change.

lucene/analysis/icu/src/java/org/apache/lucene/analysis/icu/ICUTransformCharFilter.java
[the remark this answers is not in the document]
msokolov: +1

lucene/analysis/icu/src/java/org/apache/lucene/analysis/icu/ICUTransformCharFilter.java
msokolov: I guess it just seems a bit overengineered to me - we are adding a fair bit of complexity to handle a buffer overflow edge case that we never expect to see in practice (I think?), and it might be reasonable to throw an error if it's expected to be very rare, and the user could configure a larger input buffer. But is this something that you saw while testing?

magibney: This sounds reasonable. Expanding a bit on this: the only potential simplification in this case would be to _always_ call `outputFromResultBuffer(cbuf, off, len)` after `readFromIoNormalizeUptoBoundary()`, which although not _much_ of a simplification, could be done without changing the behavior in _any_ case (`resLen==0` edge case included, in which case `outputFromResultBuffer(cbuf, off, len)` would be effectively a noop, called to no ill effect.
There'd be no errors to worry about even -- the buffer in question here is a `StringBuffer`, so it grows dynamically.

lucene/analysis/icu/src/java/org/apache/lucene/analysis/icu/ICUTransformCharFilter.java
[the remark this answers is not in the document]
magibney: IIRC I actually did encounter this situation, and the ICU Transliterator didn't throw an error itself, it just screwed up the positions in an insidious way that eventually manifested as illegal offsets.
Ultimately though there's nothing in the ICU abstract `Transliterator` class that indicates how `finishTransliterate(...)` behaves on malformed input (whether it throws errors, recovers, whatever); it just delegates to the abstract `handleTransliterate(...)` method, so perhaps the best thing to do would be to proactively throw an error -- that's also the _easiest_ thing to do since the current code goes to the trouble of detecting this situation anyway.

lucene/analysis/icu/src/java/org/apache/lucene/analysis/icu/ICUTransformCharFilter.java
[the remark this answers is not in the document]
magibney: Actually, removing the label and simply using `break;` would work here too.

lucene/analysis/icu/src/java/org/apache/lucene/analysis/icu/ICUTransformCharFilter.java
magibney: I can't think of a straightforward way to do this. Basically, we only need to `ensureRollbackBufferCapacity` if we're going to use the buffer to rollback (i.e., if the first two conditional statements are true). And then in the `else` block, we need to know whether we got there because `position.start == position.limit` (business as usual), or because we _wanted_ to rollback but there was no space in the rollback buffer.
One option would be to `ensureRollbackBufferCapacity()` _every_ time, but for the many Transliterators that will never need to _use_ the rollback buffer, this extra work would be totally unnecessary (and probably not JIT-optimizable in a runtime with several different active Transliterators?).

lucene/analysis/icu/src/java/org/apache/lucene/analysis/icu/ICUTransformCharFilter.java
```diff
@@ -0,0 +1,384 @@
...
+        // the buffer) so that we can fall through to the else clause, clear the rollback
+        // buffer, and proceed.
+        rollback(preStart, preLimit);
+      } else {
+        rollbackBufferSize = 0;
+        if (position.start > preStart) {
+          // cursor advanced
+          cursorAdvanced(preStart, preLimit);
+
+          // N.b.: advancing contextStart precludes support for quantifiers, but is crucial for streaming,
```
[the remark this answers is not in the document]
magibney: I'm not sure. I went digging in the ICU code and couldn't find much at all about quantifiers. There's no obvious API for this, and the only clue I have is that ICU Transliterator currently [doesn't change `contextStart` at all](https://github.com/unicode-org/icu/blob/a075ac9c/icu4j/main/classes/translit/src/com/ibm/icu/text/Transliterator.java#L877), which if we did the same would result in reading the entire input into the `StringBuffer`, which @rmuir [warned to avoid](https://issues.apache.org/jira/browse/LUCENE-8972?focusedCommentId=16926589#comment-16926589). We could probably choose some arbitrary "maximum backup" to pad the `contextStart` and call it a day. This would support streaming processing but also probably satisfy "reasonable" uses of quantifiers. Does that sound ok?

lucene/analysis/icu/src/java/org/apache/lucene/analysis/icu/ICUTransformCharFilter.java
```diff
@@ -0,0 +1,384 @@
...
+            }
+            buffer.getChars(preStart, position.limit, rollbackBuffer, 0);
+          }
+        } else if (preLimit != position.limit) {
+          // cursor hasn't advanced; incoming characters have probably been deleted
+          offsetDiffAdjust += preLimit - position.limit;
+          // edge case of !rollbackSizeWithinBounds needs no special handling here, since input characters *are* being
+          // processed (deleted) -- we *are* progressing through the input stream, although the output stream hasn't changed.
+        } else if (!rollbackSizeWithinBounds) {
+          // cursor hasn't advanced, no incoming characters have been deleted, and the rollback buffer is full.
```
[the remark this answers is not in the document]
magibney: Yeah the default size (8192) as initially designed is quite large. What would you think about making the behavior (strict/lenient) configurable? I'm hesitant to unconfigurably throw an error, because problematic input is not strictly-speaking _invalid_; and the maximum rollback buffer size is configurable, with smaller configurations increasing performance, but also increasing the likelihood of rollback buffer overflow.

lucene/analysis/icu/src/java/org/apache/lucene/analysis/icu/ICUTransformCharFilter.java
```diff
@@ -0,0 +1,384 @@
...
+    } else {
+      // limit change indicates change in length of replacement text; correct offsets accordingly
+      recordOffsetDiff(diff, outputLength);
+    }
+  }
+
+  private void recordOffsetDiff(int diff, int outputLength) {
+    final int cumuDiff = getLastCumulativeDiff();
+    if (diff < 0) {
+      for (int i = 1; i <= -diff; ++i) {
```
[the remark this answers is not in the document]
magibney: This is based on code from [ICUNormalizer2CharFilter](https://github.com/apache/lucene-solr/blob/master/lucene/analysis/icu/src/java/org/apache/lucene/analysis/icu/ICUNormalizer2CharFilter.java#L194-L209). I'll try to wrap my head around this a little better and put some comments in, as you say.

lucene/analysis/icu/src/test/org/apache/lucene/analysis/icu/TestICUTransformCharFilter.java
magibney: This happens because "Cyrillic-Latin" is a compound transliterator, composed of child transliterators "NFD, [main Cyrillic-Latin tranliteration], NFC". After the inner child transliterator changes a single "я" into "â" (decomposed), the final NFC transformation sees the two-character decomposition of "â", but doesn't do anything "yet" because it's not sure whether it will see more combining diacritics. But NFC never sees those input characters again, because on subsequent invocation, they are excluded by the toplevel filters on the main "Cyrillic-Latin" transliterator (which only pays attention to Cyrillic input characters). So the partially-transliterated, decomposed characters are passed through to the output. There's a solid, more general explanation in the code for ICU [Transliterator](https://github.com/unicode-org/icu/blob/a075ac9c/icu4j/main/classes/translit/src/com/ibm/icu/text/Transliterator.java#L1137-L1163).

lucene/analysis/icu/src/test/org/apache/lucene/analysis/icu/TestICUTransformCharFilter.java
magibney: This maps any-character input to Latin output (e.g., "φῄς" => "phḗis" -- length 3 => length 5). So it often increases character length wrt input, but only occasionally (e.g., "ऄॻ" => "ॻ" -- length 2 => length 1) decreases character length wrt input (in this case -- and probably most/all others -- simply by deleting a character, it seems?).
What if we added a test that took a bidirectional CJK filter like "Katakana-Latin", mapped from a Kana dictionary to Latin, then tried to round-trip that transformation (back to Katakana). Both directions would be active transformations, aggressively manipulating the offset map (increasing and decreasing character count, respectively)?

lucene/analysis/icu/src/test/org/apache/lucene/analysis/icu/TestICUTransformCharFilter.java
magibney: Depends on the definition of "a lot" :-) ... I haven't set out specifically to run this many times, but it helped me uncover a number of bugs/edge cases, and has been running consistently clean through repeated testing in the course of development. Because it works off of random input, it already runs a lot of input (1000 iterations); I just tried increasing the "iteration" count to Integer.MAX_VALUE, and it ran fine for as long as I was patient enough to wait for it, fwiw.

lucene/analysis/icu/src/test/org/apache/lucene/analysis/icu/TestICUTransformCharFilter.java
magibney: Got serviceable results by just running another `testRandomStrings()` with "Latin-Katakana". This resulted in plenty of positive diffs (decreased character count) to complement the negative diffs (decreased character counts) that were generated by the "Any-Latin" transform, so I think that'll suffice? (just pushed)

lucene/analysis/icu/src/test/org/apache/lucene/analysis/icu/TestICUTransformCharFilter.java
magibney: Running these tests with the flag `-Dtests.multiplier=n` or something causes the `testRandomStrings*` methods to run over `n` times more random input. Run with n=100, no problems ...

lucene/analysis/icu/src/java/org/apache/lucene/analysis/icu/ICUTransformCharFilter.java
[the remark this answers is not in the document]
magibney: [done](https://github.com/apache/lucene-solr/pull/892/commits/9ccea54dede858380a34f9222082e2ff242ea7e3)

lucene/analysis/icu/src/java/org/apache/lucene/analysis/icu/ICUTransformCharFilter.java
magibney: Sorry, my last comment wasn't accounting for the fact that `outputFromResultBuffer(cbuf, off, len)` is not just _called_, but is the return value. So the potential simplification would be that the top-level entry point `read(char[], int, int)` method wouldn't loop at all, but rather would simply:
1. read input
2. transform as much of it as possible
3. return output to the caller in the passed `char[]`?
I'm not 100% on all this, but my sense is that you'd be vulnerable to buffer underflow on the input side unless you looped in `readInputToBuffer()` until reaching some arbitrary threshold of available chars (how may is enough?); and instead of such an arbitrary loop, you could maybe doing useful work with already-received input, while waiting for more input to become available on the input stream. And/or you'd have an increased incidence of reporting "0 chars read" back to the caller of `read(char[], int, int)`, forcing _them_ to loop (and call us repeatedly) waiting for our output.
I took a stab at a POC reorganization of this, but it ended up being pretty complex itself, and helped clarify what I like about the current version (which I can't claim any credit for, it's based off of [ICUNormalizer2CharFilter](https://github.com/apache/lucene-solr/blob/cb0ff1a/lucene/analysis/icu/src/java/org/apache/lucene/analysis/icu/ICUNormalizer2CharFilter.java#L74-L96)): it doesn't block waiting for arbitrary amounts of data to be read or written -- it reads until it gets something, tries to transform it, and once it has something to give back, it passes it along.

msokolov:
Hi @magibney, thanks for addressing my earlier comments and waiting so long for me to get back. Just a few more things I'd like to see changed before we commit. 

The constructors for the char filter are package private, so this can only be created using the factory. I think this is a kind of unusual pattern, at least I checked a couple of other CharFilter classes and they do seem to be accessible, with public constructors. I know we have consumers of the analysis code that don't use the factory abstraction to create their analysis chains, so I think it would be good to support that here too. What do you think about (1) making both ICUTransformCharFilter constructors public, (2) moving modifyRules to that class so users will think of using it, and perhaps suggesting that it would be a good idea to do so for perf reasons in javadoc of this class, and (3) testing this in the unit tests?

I rebased and ran precommit locally and it passed OK: I'm not sure why we got that failure above.

magibney:
Thanks, @msokolov! Exposing and better documenting the constructors and rule modification/extraction (for use without the factory abstraction) would definitely be better. I will push an update shortly (if not today, then certainly by early next week).

Regarding the test failure on 086697e, I've noticed that sometimes if I push twice in close succession, even with trivial changes the later push sometimes triggers a test failure -- maybe something about leftover build sandbox state? At this point it's happened enough that my working assumption is that the failures are spurious; as a workaround I'll probably try to wait a little longer between pushes.

msokolov:
I started tinkering with a patch, tried something naive where I just refactored the method and then randomly "remove unicode normalization" from Transliterators created in the unit test, but I got some failures: probably some tests rely on the normalization, so this requires a little more thought than mere monkey-patching

magibney:
I'm curious to know more specifically what caused the tests to fail. If you're talking about randomly setting `assumeExternalUnicodeNormalization` to true, then I would indeed expect this to change results/break tests, unless accompanied by a corresponding addition of top-level Normalizer2CharFilters.

msokolov:
Yes, I think it was what you were thinking. Here's patch 
[LUCENE-8972.txt](https://github.com/apache/lucene-solr/files/3934301/LUCENE-8972.txt)
that shows what I was trying. This one actually passes

magibney:
Ah, I see. The refactoring looks good.

Separately (and wrt the test-failures induced by randomization): the `withoutUnicodeNormalization` randomization provides a good opportunity to make some documentation more explicit (and/or build more nuance/complexity into the handling of the `assumeExternalUnicodeNormalization` flag.

The `assumeExternalUnicodeNormalization` arg cannot simply be flipped in isolation without affecting behavior. My conception of this arg is as an "expert" option that can significantly improve performance, but requires users to know (and externally compensate for, with specific ICUNormalizer2CharFilters in the top-level analysis chain) the specific type of input and output unicode normalization applied by a _particular_ transliterator.

It's true that this makes it possible for users to "shoot themselves in the foot". One alternative would be to parse unicode normalization out of the Transliterator and effectively "rewrite" the analysis chain, pulling internal unicode normalization out to the top level as offset-aware ICUNormalizer2CharFilters. Pulling at this thread though leads to other potential considerations, such as:

1. I looked, and couldn't find any precedent for this type of under-the-hood analysis-chain rewriting.
2. Should we prune redundant unicode normalization induced by such rewriting? Particularly if the redundancy is introduced by rewriting adjacent Transliterators?
3. The more we attempt to handle for the user, the more we encourage users to think that this is a one-size-fits-all drop-in replacement. I actually _do_ think it would be possible to design in this way, but I'm not sure I'd be prepared to guarantee it and my sense is that really doing it "right" -- flattening, respecting nested filters, propagating offsets -- could become quite complex.

Might we consider proceeding in two phases? One with `assumeExternalUnicodeNormalization` conceived as an expert option that doesn't attempt anything more than disabling transliterator i/o normailzation, and a  potential future step that would attempt rewriting composite (potentially hierarchical) Transliterators as top-level CharFilters (with all the complexity that might entail)? For the first phase, that could mean leaving tests as they were before the patch introducing the randomized disabling of normalization, and adding additional tests that explicitly disable normalization and compensate with top-level normalization as appropriate for the associated transliterator.

msokolov:
OK, so you are good with the change I posted? If you don't see any issues,
I will go ahead and commit after beasting again to make sure there are no
more test failures due to randomization -- I think I was able to root them
out.

Regarding further automatic performance optimization, I think that is very
ambitious and maybe too much magic. But let's see if it is needed in
practice; if  a common "manual" pattern develops we can encode it as a
convenience later.

-Mike

magibney:
> Regarding further automatic performance optimization, I think that is very ambitious and maybe too much magic.

Agreed.

The only issue I have with the change you recently posted is the randomization of `assumeExternalUnicodeNormalization`. The tests _shouldn't_ generally pass with that arg randomly flipped, unless accompanied by a complementary, transliterator-specific top-level normalization change. If the tests pass with that arg randomized, that's dependent on the specific transliterator being used, and I'm concerned that even if the tests happen pass as written, randomly flipping that arg in tests conveys the impression that that's something one should be able to generally.

Would you object to reverting/removing (from the proposed patch) the randomized form of `getTransliterator(String)` in `TestICUTransformCharFilter`? Could be replaced with additional test variants that _explicitly_ disable bundled normalization (and specifically add corresponding top-level -- i.e., `ICUNormalizer2CharFilter` -- normalization if necessary)?

magibney:
As discussed above, automatically adding external (top-level analysis chain) unicode normalization to compensate for `assumeExternalUnicodeNormalization` seems too heavy-handed, and potentially complex/brittle, to be introduced at this point as user-facing functionality.

But for _tests_, where we know the specific Transliterators being used, it's fine to compensate automatically, esp. given that doing so allows us to benefit from randomized testing of the `assumeExternalUnicodeNormalization` arg.

Commit 3f4fcbb is the patch uploaded above by @msokolov. Commit feae28c exposes package-private method `ICUTransformCharFilter.unicodeNormalizationType(String)` to enable tests to properly compensate for randomly flipping the `assumeExternalUnicodeNormalization` arg, preserving @msokolov's introduction of the randomization of that arg.

In the course of work on commit feae28c, I realized that there are some Transliterators (e.g., from the tests, `Latin-Katakana`) that have leading/trailing sub-transliterators that effectively perform unicode normalization, but are not detected/removed by the `assumeExternalUnicodeNormalization` rule modification. In the case of `Latin-Katakana`, this is because the sub-transliterators in question are themselves composite transliterators). I'm not sure what (if anything) to do about this, but commit 92209c5 calls attention to the issue by adding detection for this condition, and provides a stub that could be used to log a warning to make things easier in the future if this should merit more attention.

magibney:
I didn't mean to derail things with 92209c5. I'd love to have a more robust way to handle that situation, but the problem is that I think it might be almost as much trouble to robustly detect and "properly" (whatever that would mean?) handle nested unicode normalization, as it would be to systematically decompose _all_ composite Transliterators into a tree of ICUTransformCharFilters, even implementing multi-level filters/bypass.

If the latter approach were to work as I think it would, it should be ideal in terms of performance, and in terms of functionality would if anything improve the accuracy/granularity of offsets. _But_ ... I still think the previous intention to move forward with the patch in something close to its current state makes sense, esp. given that it provides currently-absent functionality that's necessary for dictionary-based tokenization (e.g., ICUTokenizer CJ) to work reliably.

magibney:
superseded by: https://github.com/apache/lucene/pull/15

PR #895, opened 2019-09-21
fix LUCENE-8985

chenkovsky:
I have passed all test.

### DIFF
--- a/lucene/analysis/common/src/java/org/apache/lucene/analysis/synonym/SynonymGraphFilter.java
+++ b/lucene/analysis/common/src/java/org/apache/lucene/analysis/synonym/SynonymGraphFilter.java
@@ -272,7 +272,9 @@ public final class SynonymGraphFilter extends TokenFilter {
 
     //System.out.println("    lastNodeOut=" + lastNodeOut);
     //System.out.println("    term=" + termAtt);
-
+    if (lastNodeOut == -1) {
+      lastNodeOut = 0;
+    }
     posIncrAtt.setPositionIncrement(token.startNode - lastNodeOut);
     lastNodeOut = token.startNode;
     posLenAtt.setPositionLength(token.endNode - token.startNode);
@@ -336,6 +338,9 @@ public final class SynonymGraphFilter extends TokenFilter {
           bufferLen = termAtt.length();
           if (matchStartOffset == -1) {
             matchStartOffset = offsetAtt.startOffset();
+          }
+          if (nextNodeOut == -1){
+            nextNodeOut = posIncrAtt.getPositionIncrement();
           }
           inputEndOffset = offsetAtt.endOffset();
 
@@ -573,7 +578,7 @@ public final class SynonymGraphFilter extends TokenFilter {
     lookaheadNextRead = 0;
     captureCount = 0;
     lastNodeOut = -1;
-    nextNodeOut = 0;
+    nextNodeOut = -1;
     matchStartOffset = -1;
     matchEndOffset = -1;
     finished = false;
--- a/lucene/analysis/common/src/test/org/apache/lucene/analysis/synonym/TestSynonymGraphFilter.java
+++ b/lucene/analysis/common/src/test/org/apache/lucene/analysis/synonym/TestSynonymGraphFilter.java
@@ -21,6 +21,7 @@ import java.io.IOException;
 import java.io.StringReader;
 import java.text.ParseException;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Locale;
@@ -28,12 +29,14 @@ import java.util.Set;
 
 import org.apache.lucene.analysis.Analyzer;
 import org.apache.lucene.analysis.BaseTokenStreamTestCase;
+import org.apache.lucene.analysis.CharArraySet;
 import org.apache.lucene.analysis.MockAnalyzer;
 import org.apache.lucene.analysis.MockGraphTokenFilter;
 import org.apache.lucene.analysis.MockTokenizer;
 import org.apache.lucene.analysis.TokenStream;
 import org.apache.lucene.analysis.Tokenizer;
 import org.apache.lucene.analysis.core.FlattenGraphFilter;
+import org.apache.lucene.analysis.core.StopFilter;
 import org.apache.lucene.analysis.tokenattributes.*;
 import org.apache.lucene.document.Document;
 import org.apache.lucene.document.Field;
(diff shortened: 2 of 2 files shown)


chenkovsky:
wait a minute, I add more testcases. It seems my patch is not correct now.

janhoy:
Waiting for you to update the PR with more test cases, is that correct @chenkovsky ?

chenkovsky:
@janhoy  yes.

janhoy:
Would like to get this into 8.3, and release process starts in one week. @chenkovsky, do you have time to wrap it up?

janhoy:
Also, please add an entry in lucene/CHANGES.txt to this PR

janhoy:
@mikemccand or @msokolov are you available for quick review?

chenkovsky:
@janhoy I have fixed the bug. but I cannot pass some test. I think some test cases are inconsistent.

in test testBasicKeepOrigTwoOutputs, 
`
assertAnalyzesTo(a,
                     "c a b d",
                     new String[] {"c", "x", "m", "a", "y", "n", "o", "b", "d"},
                     new int[]    { 0,   2,   2,   2,   2,   2,   2,   4,   6},
                     new int[]    { 1,   5,   5,   3,   5,   5,   5,   5,   7},
                     new String[] {"word", "SYNONYM", "SYNONYM", "word", "SYNONYM", "SYNONYM", "SYNONYM", "word", "word"},
                     new int[]    { 1,   1,   0,   0,   1,   1,   1,   1,   1},
                     new int[]    { 1,   1,   2,   4,   4,   1,   2,   1,   1});
`
the graph looks like
`
c x y _ _ _ d
  m _ n o
  a _ _ _ b
`
the tails of synonyms have no overlaps.

but in testBasic1, 
`
b.add(b, "c d", "dog harness holder extras", true);
// ....
assertAnalyzesTo(a, "c d c d",
                     new String[] {"dog", "c", "harness", "d", "holder", "extras", "dog", "c", "harness", "d", "holder", "extras"},
                     new int[] {1, 0, 1, 0, 1, 1, 1, 0, 1, 0, 1, 1});
`
the graph looks like
`
dog harness holder extras dog harness holder extras
c   d       _      _      c    d      _      _  
`
the tails of synonyms overlaps.

chenkovsky:
@janhoy I have submitted the code, and passed all testcases. A small difference compared to original implementation is that I will capture all tokens(original implementation won't capture token if the first token is not matched. I think it will make code lengthy and more fragile, but get little performance improvement. so I haven't implemented it. If you think it's important, I can change my code to implement it).

janhoy:
Hmm, your commit 867cbeff010ff0febbbd258368a58ddecc770b93 labeled "bug fix" is changing a ton of code and changing this PR from a quite simple low-risk patch into a more complex one.

And you disabled a test, I don't know the consequences of that, are you saying the new patch has a performance hit?

I have not attempted to review the patch (I don't even know the code) but invited two committers to review, so far no response. Until then it would be helpful if you could clean up any whitespace-only edits, and other things that would make it easier to understand your changes.

I'll leave further review to the experts :)

chenkovsky:
I have changed the code, and enabled the capture num test. now the capture behavior is same as the original implementation. actually no performance impact.

lucene/analysis/common/src/java/org/apache/lucene/analysis/synonym/SynonymGraphFilter.java
```diff
@@ -98,67 +100,44 @@
...
-  // Start/end offset of the current match:
-  private int matchStartOffset;
-  private int matchEndOffset;
-
-  // True once the input TokenStream is exhausted:
-  private boolean finished;
 
-  private int lookaheadNextRead;
-  private int lookaheadNextWrite;
+  static class BufferedToken{
```
mikemccand: Space before `{`?

lucene/analysis/common/src/java/org/apache/lucene/analysis/synonym/SynonymGraphFilter.java
```diff
@@ -178,201 +157,136 @@ public SynonymGraphFilter(TokenStream input, SynonymMap synonyms, boolean ignore
...
-    }
-
-    if (lookaheadNextRead == lookaheadNextWrite) {
-
-      // Fast path: parse pulled one token, but it didn't match
-      // the start for any synonym, so we now return it "live" w/o having
-      // cloned all of its atts:
-      if (finished) {
-        //System.out.println("  syn: ret END");
+      if (!tokenQueue.isEmpty()){
```
mikemccand: Can you use `== false` instead, for better future readability?  And also add a space before `{`?

lucene/analysis/common/src/java/org/apache/lucene/analysis/synonym/SynonymGraphFilter.java
```diff
@@ -178,201 +157,136 @@ public SynonymGraphFilter(TokenStream input, SynonymMap synonyms, boolean ignore
...
-
-        // We used up our lookahead buffer of input tokens
-        // -- pull next real input token:
-
-        assert finished || liveToken == false;
-
-        if (finished) {
-          //System.out.println("    break: finished");
+        pendingOutput = termMatch(token.term, pendingOutput, token.term.length);
+      }else{
```
mikemccand: Spaces around `}` and `{`.

lucene/analysis/common/src/java/org/apache/lucene/analysis/synonym/SynonymGraphFilter.java
```diff
@@ -178,201 +157,136 @@ public SynonymGraphFilter(TokenStream input, SynonymMap synonyms, boolean ignore
...
+        if (inputFinished){
```
mikemccand: Space before `{`?

lucene/analysis/common/src/java/org/apache/lucene/analysis/synonym/SynonymGraphFilter.java
```diff
@@ -178,201 +157,136 @@ public SynonymGraphFilter(TokenStream input, SynonymMap synonyms, boolean ignore
...
-          // No more input tokens
-          finished = true;
-          //System.out.println("    break: now set finished");
+        }
+        if (!restored && lastInputToken != null) {
+          // restore state before increase
+          restoreState(lastInputToken.state);
+        }
+        restored = true;
+        if(!input.incrementToken()){
```
mikemccand: Spaces here too, and `== false`.

chenkovsky:
@mikemccand I agree that separating refactoring code into another pull request is better. actually I tried it at the beginning. But I find that I have to pay much more effort to reach this aim. I find that the old code is trying to maintain the state of looking ahead tokens in a complex way. the bug hides in this complex way(a lot of liveToken if else statement), I cannot tell if I fixed the bug, are there any more bugs. So I simplify the implementation.

chenkovsky:
> is SynonymGraphFilter also broken if a synonym contains a stopword? I think (not sure) the old SynonymFilter could handle this case?
@mikemccand  I tested some cases, the new code behaves same to old code. could you please give me an example, I can add it to testcase.

janhoy:
To move this forward, perhaps adding more unit test examples to the PR is a good way. I think what Mike meant was a test that have a synonym with a stopword, e.g. `RTFM => read the fine manual` where "the" is a stopword and should leave a hole.

janhoy:
@chenkovsky I'd like to help move this forward, and I think the next step is to add 5-10 more unit tests that prove that behavior is not broken in unpredictable ways.

chenkovsky:
@janhoy I added more tests. I can add more, if you think it's necessary. I find that FlattenGraphFilter also removes holes. for example, the sentence 'the the usa', we filter 'the'. for word 'usa', the posIncr of SynonymGraphFilter is 3. but posIncr of FlattenGraphFilter is 2. does it match expectation? if not, I think we should also fix FlattenGraphFilter.

PR #898, opened 2019-09-25
SOLR-13661: A package management system for Solr

noblepaul:


chatman:
* There are merge conflicts
* The branch is SOLR-13722, but title is SOLR-13661. Is this raised from the right branch?
* Seems like blob store changes are in this PR that is package manager related; shouldn't they be separated in different issues?
* The commit messages seem very temporary, we should squash merge this branch (if we decide to do so).
* One of the commits suggests "repossitory CRUD", whereas it shouldn't be here. Maybe it was removed in a subsequent commit, but the commit messages don't indicate there are any.

PR #899, opened 2019-09-25
LUCENE-8989: Allow IndexSearcher To Handle Rejected Execution

atris:
When executing queries using Executors, we should gracefully handle the case when Executor rejects a task and run the task on the caller thread.


atris:
Any thoughts on this one? Seems safe enough to merge?

I plan to merge it in another 12 hours from now -- unless any objections

atris:
Merging now -- assuming lazy consensus

jpountz:
@atris I don't have objections to that change, however 12 hours sounds too short to me for a change that could have been controversial, for instance some other committers could have assumed that a rejection being thrown is actually misconfiguration of the rejection policy of the threadpool.

atris:
+1, will increase it to 48 :)

jpountz:
Thanks.

PR #902, opened 2019-09-26
SOLR-13795: Reload solr core after schema is persisted.

thomaswoeckinger:
# Description

# Solution

# Tests

# Checklist

### DIFF
--- a/solr/core/src/java/org/apache/solr/schema/IndexSchema.java
+++ b/solr/core/src/java/org/apache/solr/schema/IndexSchema.java
@@ -138,7 +138,7 @@ public class IndexSchema {
 
   protected List<SchemaField> fieldsWithDefaultValue = new ArrayList<>();
   protected Collection<SchemaField> requiredFields = new HashSet<>();
-  protected volatile DynamicField[] dynamicFields;
+  protected DynamicField[] dynamicFields = new DynamicField[] {};
   public DynamicField[] getDynamicFields() { return dynamicFields; }
 
   protected Map<String, SchemaField> dynamicFieldCache = new ConcurrentHashMap<>();
@@ -151,7 +151,7 @@ public class IndexSchema {
   protected Map<String, List<CopyField>> copyFieldsMap = new HashMap<>();
   public Map<String,List<CopyField>> getCopyFieldsMap() { return Collections.unmodifiableMap(copyFieldsMap); }
 
-  protected DynamicCopy[] dynamicCopyFields;
+  protected DynamicCopy[] dynamicCopyFields = new DynamicCopy[] {};
   public DynamicCopy[] getDynamicCopyFields() { return dynamicCopyFields; }
 
   private Map<FieldType, PayloadDecoder> decoders = new HashMap<>();  // cache to avoid scanning token filters repeatedly, unnecessarily
@@ -962,17 +962,12 @@ public class IndexSchema {
   private void incrementCopyFieldTargetCount(SchemaField dest) {
     copyFieldTargetCounts.put(dest, copyFieldTargetCounts.containsKey(dest) ? copyFieldTargetCounts.get(dest) + 1 : 1);
   }
-  
-  private void registerDynamicCopyField( DynamicCopy dcopy ) {
-    if( dynamicCopyFields == null ) {
-      dynamicCopyFields = new DynamicCopy[] {dcopy};
-    }
-    else {
-      DynamicCopy[] temp = new DynamicCopy[dynamicCopyFields.length+1];
-      System.arraycopy(dynamicCopyFields,0,temp,0,dynamicCopyFields.length);
-      temp[temp.length -1] = dcopy;
-      dynamicCopyFields = temp;
-    }
+
+  private void registerDynamicCopyField(DynamicCopy dcopy) {
+    DynamicCopy[] temp = new DynamicCopy[dynamicCopyFields.length + 1];
+    System.arraycopy(dynamicCopyFields, 0, temp, 0, dynamicCopyFields.length);
+    temp[temp.length - 1] = dcopy;
+    dynamicCopyFields = temp;
     log.trace("Dynamic Copy Field:" + dcopy);
   }
 
@@ -1337,11 +1332,9 @@ public class IndexSchema {
         }
       }
     }
-    if (null != dynamicCopyFields) {
-      for (DynamicCopy dynamicCopy : dynamicCopyFields) {
-        if (dynamicCopy.getDestFieldName().equals(destField)) {
-          fieldNames.add(dynamicCopy.getRegex());
-        }
+    for (DynamicCopy dynamicCopy : dynamicCopyFields) {
+      if (dynamicCopy.getDestFieldName().equals(destField)) {
+        fieldNames.add(dynamicCopy.getRegex());
       }
     }
     return fieldNames;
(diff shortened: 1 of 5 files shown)


thomaswoeckinger:
@dsmiley: May you have time to review!

dsmiley:
quick comment:  the PR description is a template that you did not fill in.  Either cut out or fill in matching JIRA.

Is there an existing test suite you might add this to that is suitable, rather than creating a new one?  If not then okay but if so I'd rather reduce test suite classes when not needed (more test overhead).

thomaswoeckinger:
Puh, quick search did not show up one, i have to look closer

David Smiley <notifications@github.com> schrieb am Fr., 27. Sep. 2019,
13:59:

solr/core/src/java/org/apache/solr/schema/SchemaManager.java
```diff
@@ -138,6 +138,7 @@ private List doOperations(List<CommandOperation> operations) throws InterruptedE
             //only for non cloud stuff
             managedIndexSchema.persistManagedSchema(false);
             core.setLatestSchema(managedIndexSchema);
+            core.getCoreContainer().reload(core.getName());
```
dsmiley: I suppose this does look that easy.  Please run all tests to insure we've not broken something.

thomaswoeckinger: I fave 2 test failing, which behave different when running with ant or in eclipse:

org.apache.solr.rest.schema.TestBulkSchemaAPI.testAnalyzerClass()
org.apache.solr.client.solrj.request.SchemaTest.addFieldTypeWithAnalyzerClassAccuracy()

Both are failing because luceneMatchVersion attribute of the analyzer configuration does not match, when running with ant it is always the latest version, in eclipse null is returned.

I am not sure what is the expected result when setting luceneMatchVersion on an analyzer.

Found additional NullPointer issues in some Schema test which are not related to the change but to the execution order of tests, fixed them to.

I pushed the changes, waiting for your feedback.

thomaswoeckinger:
Searched deeper for a existing test suite, did not found one where it possible to call query after schema change.  At SchemaTest and TestBulkSchemaAPI it is not possible to use getSolrClient().query, results in 404 .../select not found.

If you have any suggestions...

solr/core/src/java/org/apache/solr/schema/SchemaManager.java
thomaswoeckinger: After some code analysis it is 100% clear that the only analyzer property which get written to the schema is the class property look at method org.apache.solr.response.SchemaXmlWriter.writeAnalyzer(SimpleOrderedMap<Object>, String) at the lines 210 to 222, i will push the modification again

thomaswoeckinger:
All tests passed

dsmiley:
So you seem to have uncovered a bug... SchemaXmlWriter.writeAnalyzer isn't writing the luceneMatchVersion (a rather rare setting at this level but still).  Please file an issue and reference the issue in the test assertions you removed.  Leave them commented so they can be reinstated in the future.

Those null checks for the array of dynamic fields is sad... I'd rather see that they are never null (use empty arrays).  I'm missing the context right now but you think that could be fixed?  And if we are going to modify these lines of code then perhaps we could update them to use Arrays.copyOf?

thomaswoeckinger:
> So you seem to have uncovered a bug... SchemaXmlWriter.writeAnalyzer isn't writing the luceneMatchVersion (a rather rare setting at this level but still). Please file an issue and reference the issue in the test assertions you removed. Leave them commented so they can be reinstated in the future.

So is the property luceneMatchVersion the only one, or should it be fixed in a more general way to write any property to the schema?

> 
> Those null checks for the array of dynamic fields is sad... I'd rather see that they are never null (use empty arrays).

I think the same, but the the nulling out in method org.apache.solr.schema.ManagedIndexSchema.deleteCopyField(String, String) in line number 880 let me think different. Also the initialization of dynamicFields and dynamicCopyFields is different from the other members of IndexSchema, dynamicFields is volatile too. 

> I'm missing the context right now but you think that could be fixed? And if we are going to modify these lines of code then perhaps we could update them to use Arrays.copyOf?

I think it could be fixed, not that difficult, but before refactoring i think we must know why this dynamicFields field is nulled out. So did you know a somebody who is responsible for that part?

thomaswoeckinger:
Pushed again; @dsmiley  Did you already had time to talk to somebody about the nulled out arrays dynamicFields?

dsmiley:
> Did you already had time to talk to somebody about the nulled out arrays dynamicFields?

No.  On such matters you can try and look at code history and associated JIRA issues.  I *suspect* this will lead you to Steve Rowe but I won't at-mention as I'd rather you figure this out.  Maybe it's him; I don't know.

thomaswoeckinger:
Pushed again, @dsmiley plz review, have done research on this null behaviour, it was introduced 5 years ago, did not find any reason why not initialize it, so it is initialized and all null checks are gone.

solr/core/src/java/org/apache/solr/schema/ManagedIndexSchema.java
```diff
@@ -81,7 +81,7 @@
 /** Solr-managed schema - non-user-editable, but can be mutable via internal and external REST API requests. */
 public final class ManagedIndexSchema extends IndexSchema {
 
-  private boolean isMutable = false;
+  private final boolean isMutable;
```
dsmiley: nice

solr/core/src/test/org/apache/solr/rest/schema/TestBulkSchemaAPI.java
```diff
@@ -184,8 +188,6 @@ public void testAnalyzerClass() throws Exception {
     response = restTestHarness.post("/schema", json(addFieldTypeAnalyzerWithClass + suffix));
     map = (Map) fromJSONString(response);
     assertNull(response, map.get("error"));
-    
-    restTestHarness.checkAdminResponseStatus("/admin/cores?wt=xml&action=RELOAD&core=" + coreName, "0");
```
dsmiley: I'm glad you remembered

solr/core/src/java/org/apache/solr/schema/IndexSchema.java
```diff
@@ -1556,48 +1547,46 @@ SimpleOrderedMap getProperties(SchemaField sf) {
         }
       }
     }
-    if (null != dynamicCopyFields) {
```
dsmiley: Is this big diff just a re-indent after removing the wrapping null check?

solr/core/src/java/org/apache/solr/schema/IndexSchema.java
```diff
@@ -138,7 +138,7 @@
 
   protected List<SchemaField> fieldsWithDefaultValue = new ArrayList<>();
   protected Collection<SchemaField> requiredFields = new HashSet<>();
-  protected volatile DynamicField[] dynamicFields;
```
dsmiley: @sarowe why was this volatile?  It's fishy to see this as the only volatile field.

solr/core/src/java/org/apache/solr/schema/IndexSchema.java
thomaswoeckinger: > Is this big diff just a re-indent after removing the wrapping null check?

Yes just removed the null check.

solr/core/src/java/org/apache/solr/schema/IndexSchema.java
thomaswoeckinger: > @sarowe why was this volatile? It's fishy to see this as the only volatile field.

I was looking for any double check pattern, or lazy init, found nothing, as you mentioned it is the only one, so i removed it, makes no sense from my point of view

PR #904, opened 2019-09-27
LUCENE-8992: Share minimum score across segment in concurrent search

jimczi:
This is a follow up of LUCENE-8978 that introduces shared minimum score across segment
in concurrent search for top field collectors that sort by relevance first.


lucene/core/src/java/org/apache/lucene/search/TopFieldCollector.java
```diff
@@ -389,14 +406,14 @@ public static TopFieldCollector create(Sort sort, int numHits, FieldDoc after, i
       throw new IllegalArgumentException("totalHitsThreshold must be >= 0, got " + totalHitsThreshold);
     }
 
-    return create(sort, numHits, after, HitsThresholdChecker.create(totalHitsThreshold));
+    return create(sort, numHits, after, HitsThresholdChecker.create(totalHitsThreshold), null);
```
atris: Nit: add a comment explaining the null? like `, null /* bottomValueChecker */`

atris:
Thanks @jimczi !

The changes look good -- I left some comments.

lucene/core/src/java/org/apache/lucene/search/TopFieldCollector.java
```diff
@@ -423,22 +440,24 @@ static TopFieldCollector create(Sort sort, int numHits, FieldDoc after,
...
   }
 
   /**
    * Create a CollectorManager which uses a shared hit counter to maintain number of hits
+   * and a shared bottom value checker to propagate the minimum score accross segments if
+   * the primary sort is by relevancy.
    */
-  public static CollectorManager<TopFieldCollector, TopFieldDocs> createSharedManager(Sort sort, int numHits, FieldDoc after,
-                                                                                 int totalHitsThreshold) {
+  public static CollectorManager<TopFieldCollector, TopFieldDocs> createSharedManager(Sort sort, int numHits, FieldDoc after, int totalHitsThreshold) {
```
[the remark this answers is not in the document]
jimczi: good catch, this shouldn't be changed

jimczi:
Thanks for reviewing @atris . I pushed some changes to address your comments and add unit tests for the bottom value checker, can you take another look ?

lucene/core/src/java/org/apache/lucene/search/TopFieldCollector.java
```diff
@@ -318,18 +327,35 @@ private TopFieldCollector(FieldValueHitQueue<Entry> pq, int numHits,
...
+          && (queueFull || (bottomValueChecker != null && bottomValueChecker.getBottomValue() > 0f))) {
+      float maxMinScore = Float.NEGATIVE_INFINITY;
+      if (queueFull) {
+        assert bottom != null && firstComparator != null;
+        maxMinScore = firstComparator.value(bottom.slot);
+        if (bottomValueChecker != null) {
+          bottomValueChecker.updateThreadLocalBottomValue(maxMinScore);
+        }
+      }
+      if (bottomValueChecker != null) {
```
atris: Do we need these checks again, since we could have gotten here if `shouldUpdateMinScore` returned true, which would signal that the global score is better than local minimum?

lucene/core/src/java/org/apache/lucene/search/TopScoreDocCollector.java
```diff
@@ -139,7 +140,8 @@ public void collect(int doc) throws IOException {
 
           if (score > after.score || (score == after.score && doc <= afterDoc)) {
             // hit was collected on a previous page
-            if (totalHitsRelation == TotalHits.Relation.EQUAL_TO && hitsThresholdChecker.isThresholdReached()) {
+            if ((totalHitsRelation == TotalHits.Relation.EQUAL_TO || shouldUpdateMinScore())
```
atris: I am not really sure if we need `shouldUpdateMinScore` here -- it would be called anyway if the global threshold is reached. Is the objective to reduce the number of redundant calls to `updateMinCompetitiveScore` i.e. only call when we know that we have a better global score?

lucene/core/src/java/org/apache/lucene/search/TopFieldCollector.java
```diff
@@ -318,18 +327,35 @@ private TopFieldCollector(FieldValueHitQueue<Entry> pq, int numHits,
...
     }
+    this.bottomValueChecker = bottomValueChecker;
   }
 
   @Override
   public ScoreMode scoreMode() {
     return scoreMode;
   }
 
+  protected boolean shouldUpdateMinScore() {
```
atris: Should we call it `betterGlobalScoreAvailable`, since the minimum score could still be left unchanged (if the threshold is not reached, for eg?) Another idea is to refactor `updateMinCompetitiveScore` to be split into two discrete methods, so that `shouldUpdateMinScore` can be extended to also update the score?

lucene/core/src/java/org/apache/lucene/search/TopScoreDocCollector.java
jimczi: The idea here is that currently we call updateMinCompetitiveScore only once when the threshold is reached, we also update the minimum score when the queue is full and a new competitive hit is inserted. However now that the minimum score can be updated by other collectors it is possible that the global minimum score changed while the local one remains unchanged (because we see only non-competitive hits). This change ensures that we update the current min score to take into account the new global minimum score if it is greater than the local one. I think it's important because otherwise we'll miss a lot of updates to the global minimum score (e.g. if the collector only see non-competitive hits).

lucene/core/src/java/org/apache/lucene/search/TopFieldCollector.java
jimczi: +1 for renaming, I am not sure about the splitting of `updateMinCompetitiveScore` since `shouldUpdateMinScore` is a special thing that is only needed when we collect a non-competitive hits.

lucene/core/src/test/org/apache/lucene/search/TestTopDocsCollector.java
```diff
@@ -404,44 +424,288 @@ public void testTotalHits() throws Exception {
 
   public void testGlobalScore() throws Exception {
     Directory dir = newDirectory();
-    RandomIndexWriter writer = new RandomIndexWriter(random(), dir);
-    try (LineFileDocs docs = new LineFileDocs(random())) {
```
atris: Hmm, could we also let the `LineFileDocs` based test remain, in addition to the new tests? I find it to be a good way to do random tests on realistic(ish) data sets.

jimczi: I restored it in https://github.com/apache/lucene-solr/pull/904/commits/f55633c8812d6e33ffab309f1f6d8af53e56b774

jimczi:
Thanks @atris, I refactored the logic to update the minimum competitive score in https://github.com/apache/lucene-solr/pull/904/commits/7eab6ebb2752c7a3ab61a66aaf531e07838b36a9 to make it more readable. Now I'll run some benchmark with and without concurrent search to evaluate the impact.

lucene/core/src/java/org/apache/lucene/search/TopScoreDocCollector.java
atris: I am not sure if I understand. If a collector is only seeing non competitive hits, it would not be able to contribute to the global minimum score?

On the other hand, if the local minimum score is enough to filter out non competitive hits but not good enough to beat the current global minimum, we could run with the local score and let it be the sentinel of the collector's bottom score.

jimczi: Currently if the local queue if full and the collector only sees non-competitive hits we don't update the minimum competitive score. This is normal since the local minimum competitive score cannot changed on non-competitive hits, however it is possible that the global minimum score changed in the meantime and that the collector is missing a big update that propagated to the local scorer could make the skipping of hits more efficient. That's what the test here is doing, checking that we can update the minimum score from the global one.
I refactored this part a bit in my last commit but the idea is the same.

jimczi:
So it seems that my idea to always check the global minimum score when a non-competitive hits is collected makes wildcard query 20% slower. I removed this path in the last commit and re-ran the benchmark, here's the result:
`````
                    TaskQPS baseline      StdDev   QPS patch      StdDev                Pct diff
              AndHighLow     1018.59      (0.0%)      969.69      (0.0%)   -4.8% (  -4% -   -4%)
                 Prefix3        8.76      (0.0%)        8.63      (0.0%)   -1.4% (  -1% -   -1%)
                 Respell       38.17      (0.0%)       37.66      (0.0%)   -1.3% (  -1% -   -1%)
             MedSpanNear       25.13      (0.0%)       25.04      (0.0%)   -0.4% (   0% -    0%)
             LowSpanNear       28.09      (0.0%)       28.01      (0.0%)   -0.3% (   0% -    0%)
    HighIntervalsOrdered        5.77      (0.0%)        5.76      (0.0%)   -0.2% (   0% -    0%)
                  IntNRQ        0.95      (0.0%)        0.95      (0.0%)    0.1% (   0% -    0%)
                Wildcard        9.69      (0.0%)        9.72      (0.0%)    0.3% (   0% -    0%)
            HighSpanNear        5.57      (0.0%)        5.61      (0.0%)    0.6% (   0% -    0%)
              AndHighMed      126.80      (0.0%)      127.69      (0.0%)    0.7% (   0% -    0%)
                  Fuzzy2       54.33      (0.0%)       54.74      (0.0%)    0.7% (   0% -    0%)
                  Fuzzy1       68.71      (0.0%)       70.13      (0.0%)    2.1% (   2% -    2%)
               OrHighMed       86.45      (0.0%)       89.62      (0.0%)    3.7% (   3% -    3%)
        HighSloppyPhrase       11.15      (0.0%)       11.64      (0.0%)    4.4% (   4% -    4%)
              OrHighHigh       25.15      (0.0%)       26.25      (0.0%)    4.4% (   4% -    4%)
         MedSloppyPhrase       20.45      (0.0%)       21.56      (0.0%)    5.4% (   5% -    5%)
         LowSloppyPhrase       28.62      (0.0%)       30.71      (0.0%)    7.3% (   7% -    7%)
            OrNotHighLow      982.15      (0.0%)     1075.26      (0.0%)    9.5% (   9% -    9%)
             AndHighHigh       33.32      (0.0%)       37.07      (0.0%)   11.2% (  11% -   11%)
               LowPhrase       54.12      (0.0%)       65.54      (0.0%)   21.1% (  21% -   21%)
               MedPhrase       51.57      (0.0%)       64.64      (0.0%)   25.3% (  25% -   25%)
                 LowTerm      691.12      (0.0%)     1076.32      (0.0%)   55.7% (  55% -   55%)
              HighPhrase       57.17      (0.0%)       92.55      (0.0%)   61.9% (  61% -   61%)
            OrNotHighMed      415.02      (0.0%)      808.04      (0.0%)   94.7% (  94% -   94%)
                 MedTerm      409.34      (0.0%)      903.03      (0.0%)  120.6% ( 120% -  120%)
           OrNotHighHigh      330.80      (0.0%)      746.05      (0.0%)  125.5% ( 125% -  125%)
            OrHighNotLow      297.12      (0.0%)      697.12      (0.0%)  134.6% ( 134% -  134%)
           OrHighNotHigh      281.15      (0.0%)      668.95      (0.0%)  137.9% ( 137% -  137%)
            OrHighNotMed      315.19      (0.0%)      760.68      (0.0%)  141.3% ( 141% -  141%)
               OrHighLow      177.73      (0.0%)      467.47      (0.0%)  163.0% ( 163% -  163%)
                HighTerm      341.85      (0.0%)      914.70      (0.0%)  167.6% ( 167% -  167%)
`````
The performance seems stable and some important queries (highterm, highphrase, orhighlow, ...) get a nice boost so I think it's ready.
To run the benchmark on TopFieldCollector I patched luceneutil to always use the following sort:
 `new Sort(SortField.FIELD_SCORE, SortField.FIELD_DOC));`.
I also had to deactivate some checks that ensure that multiple runs of the same query return the same number of total hits. With this change and the one in LUCENE-8978 this is not guaranteed if you use concurrent searches. I don't think it's an issue but would be good to mention in javadocs and changes.

File: lucene/core/src/java/org/apache/lucene/search/TopScoreDocCollector.java (hunk 4 above)
Comment (atris) [reply lucene/core/src/java/org/apache/lucene/search/TopScoreDocCollector.java]: Ok, I think you reverted this change in the latest iteration?

Comment (jimczi) [reply lucene/core/src/java/org/apache/lucene/search/TopScoreDocCollector.java]: Yep sorry, after running some benchmarks it seems that checking the global score on every non-competitive hits hurts the performance of constant score queries.
The gains are already significative without it so I restored the previous version.

File: lucene/core/src/java/org/apache/lucene/search/TopScoreDocCollector.java [hunk 7]
```diff
@@ -277,31 +288,44 @@ public ScoreMode scoreMode() {
...
+    if (hitsThresholdChecker.isThresholdReached()) {
+      boolean hasChanged = false;
+      if (isQueueFull()) {
+        // since we tie-break on doc id and collect in doc id order, we can require
+        // the next float
+        float localMinScore = Math.nextUp(pqTop.score);
+        if (localMinScore > minCompetitiveScore) {
+          hasChanged = true;
+          minCompetitiveScore = localMinScore;
+          if (bottomValueChecker != null) {
```
atris: I am not sure about the benefit of this refactoring : seems like it is still checking if the local scorer's minimum score can be improved by either pqTop's score or the global minimum score, which is what was being done earlier? Could you describe the idea behind this refactoring, please?

jimczi: I agree, this refactoring was made because I wanted to call `updateMinCompetitiveScore` eagerly on non-competitive hits so I added some checks to avoid propagating the same minimum score multiple times. It is no longer needed since I removed this pre-optimization in the last iteration. I pushed a new [commit](https://github.com/apache/lucene-solr/pull/904/commits/5895797faed1f32b1fd2b733de66ea9592bfac76) to restore the original one.

jpountz:
> I also had to deactivate some checks that ensure that multiple runs of the same query return the same number of total hits.

FYI luceneutil has an option for this: `verifyCounts=False`, e.g. I have this in my benchmark configuration when I test changes that don't provide reproducible counts:

```python
comp = competition.Competition(verifyScores=True, verifyCounts=False)
```

One thought I had when reviewing the patch: currently we check the minimum competitive score of other collectors when the minimum competitive score of the current collector needs updating. Depending on the query, this might be very frequent, or infrequent. Maybe another way to do this that would give us more control over this would be to introduce checkpoints via bulk scorers. IndexSearcher could do something like that:

```java
      BulkScorer scorer = weight.bulkScorer(ctx);
      if (scorer != null) {
        try {
          Bits liveDocs = ctx.reader().getLiveDocs();
          if (multipleThreads && sortByScore) {
            // Split the doc ID space to introduce checkpoints when collectors exchange information about their minimum competitive scores
            int windowSize = 1 << 20; // arbitrary
            for (int doc = 0; doc < maxDoc; ) {
              int start = doc;
              int end = (int) Math.min(maxDoc, (long) doc + windowSize);
              float maxOfMinCompetitiveScores = ...; // can we get it here?
              scorer.setMinCompetitiveScore(maxOfMinCompetitiveScores); // this method would have to be added, and implemented by propagating to the underlying scorer
              doc = scorer.score(leafCollector, liveDocs, start, end);
            }
          } else {
            scorer.score(leafCollector, liveDocs);
          }
        } catch (CollectionTerminatedException e) {
          // collection was terminated prematurely
          // continue with the following leaf
        }
      }
```

jimczi:
> FYI luceneutil has an option for this: verifyCounts=False, e.g. I have this in my benchmark configuration when I test changes that don't provide reproducible counts:

Yes I am aware of that but the problem is different, with this change a query that runs on the same searcher with concurrent execution can report different total hits depending on which segments execute faster since we share the global min score. In luceneutils we expect that a query that runs on a searcher always return the same number of hits so we always check that it is the case (taskRepeatCount is greater than 1): see https://github.com/mikemccand/luceneutil/blob/master/src/python/benchUtil.py#L682.
It can be surprising for users which is why I think it should be added in the javadocs ?

> One thought I had when reviewing the patch: currently we check the minimum competitive score of other collectors when the minimum competitive score of the current collector needs updating. Depending on the query, this might be very frequent, or infrequent. Maybe another way to do this that would give us more control over this would be to introduce checkpoints via bulk scorers. IndexSearcher could do something like that:

That's an interesting idea, I'll try to modify this pr to see how it could work.

jpountz:
Oops, then sorry for the noise about verifyCounts!

lucene/core/src/java/org/apache/lucene/search/TopFieldCollector.java
```diff
@@ -318,19 +350,41 @@ private TopFieldCollector(FieldValueHitQueue<Entry> pq, int numHits,
...
+        totalHitsRelation = TotalHits.Relation.GREATER_THAN_OR_EQUAL_TO;
+      }
+    }
+  }
+
   protected void updateMinCompetitiveScore(Scorable scorer) throws IOException {
-    if (canSetMinScore && hitsThresholdChecker.isThresholdReached() && queueFull) {
+    if (canSetMinScore
+          && hitsThresholdChecker.isThresholdReached()
+          && queueFull) {
```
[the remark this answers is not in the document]
atris: +1, agreed. We should not be publishing minimum scores when the PQ is not full, but there is no reason why we shouldnt use the value if it is present already.

lucene/core/src/java/org/apache/lucene/search/MaxScoreAccumulator.java
```diff
@@ -0,0 +1,89 @@
...
+  // non-final and visible for tests
+  long modInterval;
+
+  MaxScoreAccumulator() {
+    this.modInterval = DEFAULT_INTERVAL;
+  }
+
+  void accumulate(int docID, float score) {
+    assert docID >= 0 && score >= 0;
+    long encode = (((long) Float.floatToIntBits(score)) << 32) | (docID & 0xffffffffL);
```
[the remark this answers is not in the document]
jimczi: ++, resolved in https://github.com/apache/lucene-solr/pull/904/commits/b623f6363d818982aa280989d2ae756944375a17

lucene/core/src/java/org/apache/lucene/search/MaxScoreAccumulator.java
```diff
@@ -0,0 +1,89 @@
...
+    long value = acc.get();
+    if (value == Long.MIN_VALUE) {
+      return null;
+    }
+    float score = Float.intBitsToFloat((int) (value >> 32));
+    int docID = (int) value;
+    return new Result(docID, score);
+  }
+
+  static class Result implements Comparable<Result> {
```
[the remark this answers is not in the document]
jimczi: ++, https://github.com/apache/lucene-solr/pull/904/commits/b623f6363d818982aa280989d2ae756944375a17

lucene/core/src/java/org/apache/lucene/search/TopFieldCollector.java
```diff
@@ -237,7 +255,9 @@ public void collect(int doc) throws IOException {
                 } else {
                   collectedAllCompetitiveHits = true;
                 }
-              } else if (totalHitsRelation == Relation.GREATER_THAN_OR_EQUAL_TO) {
+              } else if (totalHitsRelation == Relation.EQUAL_TO) {
```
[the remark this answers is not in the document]
jimczi: Yes the comments above explain why. We want to update the minimum competitive score if the threshold is reached for the first time here. That's the only case where it matters to call updateMinScore on a non-competitive hits. Otherwise it is expected that the minimum score cannot change.

lucene/core/src/java/org/apache/lucene/search/TopFieldCollector.java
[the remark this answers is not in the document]
jimczi: We do set the min competitive score from other threads if the queue is not full but only in `updateGlobalMinCompetitiveScore` where we only check that the threshold is reached. We want to control the rate at which we check the global minimum score so we don't try to access it here.

lucene/core/src/java/org/apache/lucene/search/TopFieldCollector.java
```diff
@@ -123,13 +124,22 @@ public LeafCollector getLeafCollector(LeafReaderContext context) throws IOExcept
...
+            updateGlobalMinCompetitiveScore(scorer);
+          }
         }
 
         @Override
         public void collect(int doc) throws IOException {
           ++totalHits;
           hitsThresholdChecker.incrementHitCount();
+
+          if (minScoreAcc != null && totalHits % minScoreAcc.modInterval == 0) {
```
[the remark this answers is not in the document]
jimczi: yep no idea, I do it like that in order to be able to change the interval in tests but if it hurt performance we might want to have the optimized version. I'll run a benchmark to see if it makes a difference  since this call is for every doc in the concurrent case

jimczi:
I pushed another commit to replace the modulo with a bitwise operation as suggested by @jpountz . That seemed to help a bit and since there are no regressions and some nice boosts I think it is ready for another review.

lucene/core/src/java/org/apache/lucene/search/TopFieldCollector.java
```diff
@@ -223,6 +238,10 @@ public void collect(int doc) throws IOException {
           totalHits++;
           hitsThresholdChecker.incrementHitCount();
 
+          if (minScoreAcc != null && totalHits % minScoreAcc.modInterval == 0) {
```
jpountz:
```suggestion
          if (minScoreAcc != null && (totalHits & minScoreAcc.modInterval) == 0) {
```

lucene/core/src/java/org/apache/lucene/search/TopScoreDocCollector.java
```diff
@@ -137,9 +147,13 @@ public void collect(int doc) throws IOException {
           totalHits++;
           hitsThresholdChecker.incrementHitCount();
 
+          if (minScoreAcc != null && totalHits % minScoreAcc.modInterval == 0) {
```
jpountz:
```suggestion
          if (minScoreAcc != null && (totalHits & minScoreAcc.modInterval) == 0) {
```

dsmiley:
Jim, next time please squash-merge.  That gives you the chance to edit the commit message so that it can conform to our standards, and it also makes history browsing easier.

jimczi:
@dsmiley yes sorry about that, I am used to Elasticsearch repo where squash-merge is the default option so I clicked too quickly. Maybe we can make it the default option for Lucene too though ?

PR #912, opened 2019-10-01
Make testLUCENE8736 Check For Async Caching

atris:
Allow testLUCENE8736 to wait for async caching query loads


atris:
Ran `ant beast -Dbeast.iters=20 -Dtests.dups=6  -Dtestcase=TestLatLonShape -Dtests.method=testLUCENE8736* -Dtests.seed=D784C9A46B62FB84 -Dtests.slow=true -Dtests.badapples=true -Dtests.locale=brx -Dtests.timezone=America/Guatemala -Dtests.asserts=true`

Came in clean

atris:
precommit fails due to errors from `MockInitialContextFactory.java`

iverase:
I don't think we should do that, caching the query should not affect the result of the query.

iverase:
And there are other test failing for the same reason, see:

https://elasticsearch-ci.elastic.co/job/apache+lucene-solr+master/5609/console

atris:
I see, I will investigate the other test and reopen the PR

PR #913, opened 2019-10-01
LUCENE-8995: TopSuggestDocsCollector#collect should be able to signal…

cbuescher:
… rejection

Currently the suggestion collectors collect method has no way of signaling back
to the caller whether the matched completion it received was indeed collected.
While this is currenty always the case for the default TopSuggestDocsCollector,
there are implementations that overwrite this with custom collection logic that
also deduplicates based on e.g. docID and/or context. Currently if those
completions are rejected, the calling TopNSearcher has no way of noting these
rejections and might therefore terminate early, missing on completions that
should be returned.

# Description

# Solution

# Tests

# Checklist

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/util/fst/Util.java
+++ b/lucene/core/src/java/org/apache/lucene/util/fst/Util.java
@@ -460,11 +460,6 @@ public final class Util {
           continue;
         }
 
-        if (results.size() == topN-1 && maxQueueDepth == topN) {
-          // Last path -- don't bother w/ queue anymore:
-          queue = null;
-        }
-
         // We take path and find its "0 output completion",
         // ie, just keep traversing the first arc with
         // NO_OUTPUT that we can find, since this must lead
--- a/lucene/suggest/src/java/org/apache/lucene/search/suggest/document/NRTSuggester.java
+++ b/lucene/suggest/src/java/org/apache/lucene/search/suggest/document/NRTSuggester.java
@@ -39,6 +39,7 @@ import org.apache.lucene.util.fst.PairOutputs.Pair;
 import org.apache.lucene.util.fst.PairOutputs;
 import org.apache.lucene.util.fst.PositiveIntOutputs;
 import org.apache.lucene.util.fst.Util;
+import org.apache.lucene.util.fst.Util.TopResults;
 
 import static org.apache.lucene.search.suggest.document.NRTSuggester.PayLoadProcessor.parseSurfaceForm;
 
@@ -200,23 +201,30 @@ public final class NRTSuggester implements Accountable {
         if (!scorer.accept(docID, acceptDocs)) {
           return false;
         }
+        boolean duplicateSurfaceForm = false;
+        boolean collected = false;
         if (collector.doSkipDuplicates()) {
           // now record that we've seen this surface form:
           char[] key = new char[spare.length()];
           System.arraycopy(spare.chars(), 0, key, 0, spare.length());
           if (collector.seenSurfaceForms.contains(key)) {
             // we already collected a higher scoring document with this key, in this segment:
-            return false;
+            duplicateSurfaceForm = true;
+          } else {
+            collector.seenSurfaceForms.add(key);
           }
-          collector.seenSurfaceForms.add(key);
         }
-        try {
-          float score = scorer.score(decode(path.output.output1), path.boost);
-          collector.collect(docID, spare.toCharsRef(), path.context, score);
-          return true;
-        } catch (IOException e) {
-          throw new RuntimeException(e);
+
+        // only try collecting if we didn't already detect a surface form duplicate and collector.doSkipDuplicates() == true
+        if (duplicateSurfaceForm == false) {
+          try {
+            float score = scorer.score(decode(path.output.output1), path.boost);
+            collected = collector.collect(docID, spare.toCharsRef(), path.context, score);
+          } catch (IOException e) {
+            throw new RuntimeException(e);
+          }
         }
+        return collected;
       }
     };
 
@@ -239,7 +247,11 @@ public final class NRTSuggester implements Accountable {
     }
     // hits are also returned by search()
     // we do not use it, instead collect at acceptResult
-    searcher.search();
+    TopResults<Pair<Long,BytesRef>> results = searcher.search();
+    if (results.isComplete == false) {
+      collector.notComplete();
+    }
+    return;
     // search admissibility is not guaranteed
     // see comment on getMaxTopNSearcherQueueSize
     // assert  search.isComplete;
--- a/lucene/suggest/src/java/org/apache/lucene/search/suggest/document/TopSuggestDocsCollector.java
+++ b/lucene/suggest/src/java/org/apache/lucene/search/suggest/document/TopSuggestDocsCollector.java
@@ -63,6 +63,8 @@ public class TopSuggestDocsCollector extends SimpleCollector {
   /** Document base offset for the current Leaf */
   protected int docBase;
 
+  private boolean isComplete = true;
+
   /**
    * Sole constructor
    *
@@ -116,7 +118,7 @@ public class TopSuggestDocsCollector extends SimpleCollector {
    * NOTE: collection at the leaf level is guaranteed to be in
    * descending order of score
    */
-  public void collect(int docID, CharSequence key, CharSequence context, float score) throws IOException {
+  public boolean collect(int docID, CharSequence key, CharSequence context, float score) throws IOException {
     SuggestScoreDoc current = new SuggestScoreDoc(docBase + docID, key, context, score);
     if (current == priorityQueue.insertWithOverflow(current)) {
       // if the current SuggestScoreDoc has overflown from pq,
@@ -125,6 +127,7 @@ public class TopSuggestDocsCollector extends SimpleCollector {
       // TODO: reuse the overflow instance?
       throw new CollectionTerminatedException();
     }
+    return true;
   }
 
   /**
@@ -200,5 +203,20 @@ public class TopSuggestDocsCollector extends SimpleCollector {
   @Override
   public ScoreMode scoreMode() {
     return ScoreMode.COMPLETE;
+  }
+
+  /**
+   * returns true if the collector clearly exhausted all possibilities to collect results
+   */
+  public boolean isComplete() {
+    return this.isComplete ;
+  }
+
+  /**
+   * call to signal that during collection at least one segment might have returned incomplete results, e.g. because
+   * of too many rejections
+   */
+  void notComplete() {
+    this.isComplete = false;
   }
 }
--- a/lucene/suggest/src/test/org/apache/lucene/search/suggest/document/TestPrefixCompletionQuery.java
+++ b/lucene/suggest/src/test/org/apache/lucene/search/suggest/document/TestPrefixCompletionQuery.java
@@ -17,10 +17,15 @@
 package org.apache.lucene.search.suggest.document;
 
 import java.io.IOException;
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
 import java.util.Objects;
+import java.util.Set;
 
 import org.apache.lucene.analysis.Analyzer;
 import org.apache.lucene.analysis.MockAnalyzer;
+import org.apache.lucene.analysis.MockSynonymAnalyzer;
 import org.apache.lucene.analysis.MockTokenFilter;
 import org.apache.lucene.analysis.MockTokenizer;
 import org.apache.lucene.document.Document;
@@ -248,6 +253,61 @@ public class TestPrefixCompletionQuery extends LuceneTestCase {
     query = new PrefixCompletionQuery(analyzer, new Term("suggest_field", "app"), filter);
     suggest = indexSearcher.suggest(query, 3, false);
     assertSuggestions(suggest, new Entry("applle", 4), new Entry("apples", 3));
+
+    reader.close();
+    iw.close();
+  }
+
+  /**
+   * Test that the correct amount of documents are collected if using a collector that also rejects documents.
+   */
+  public void testCollectorThatRejects() throws Exception {
+    // use synonym analyzer to have multiple paths to same suggested document. This mock adds "dog" as synonym for "dogs"
+    Analyzer analyzer = new MockSynonymAnalyzer();
+    RandomIndexWriter iw = new RandomIndexWriter(random(), dir, iwcWithSuggestField(analyzer, "suggest_field"));
+    List<Entry> expectedResults = new ArrayList<Entry>();
+
+    for (int docCount = 10; docCount > 0; docCount--) {
+      Document document = new Document();
+      String value = "ab" + docCount + " dogs";
+      document.add(new SuggestField("suggest_field", value, docCount));
+      expectedResults.add(new Entry(value, docCount));
+      iw.addDocument(document);
+    }
+
+    if (rarely()) {
+      iw.commit();
+    }
+
+    DirectoryReader reader = iw.getReader();
+    SuggestIndexSearcher indexSearcher = new SuggestIndexSearcher(reader);
+
+    PrefixCompletionQuery query = new PrefixCompletionQuery(analyzer, new Term("suggest_field", "ab"));
+    int topN = 5;
+
+    // use a TopSuggestDocsCollector that rejects results with duplicate docIds
+    TopSuggestDocsCollector collector = new TopSuggestDocsCollector(topN, false) {
+
+      private Set<Integer> seenDocIds = new HashSet<>();
+
+      @Override
+      public boolean collect(int docID, CharSequence key, CharSequence context, float score) throws IOException {
+          int globalDocId = docID + docBase;
+          boolean collected = false;
+          if (seenDocIds.contains(globalDocId) == false) {
+              super.collect(docID, key, context, score);
+              seenDocIds.add(globalDocId);
+              collected = true;
+          }
+          return collected;
+      }
+    };
+
+    indexSearcher.suggest(query, collector);
+    assertSuggestions(collector.get(), expectedResults.subList(0, topN).toArray(new Entry[0]));
+
+    // TODO expecting true here, why false?
+    assertFalse(collector.isComplete());
 
     reader.close();
     iw.close();


lucene/suggest/src/java/org/apache/lucene/search/suggest/document/NRTSuggester.java
```diff
@@ -200,23 +201,30 @@ protected boolean acceptResult(Util.FSTPath<Pair<Long, BytesRef>> path) {
...
         if (collector.doSkipDuplicates()) {
           // now record that we've seen this surface form:
           char[] key = new char[spare.length()];
           System.arraycopy(spare.chars(), 0, key, 0, spare.length());
           if (collector.seenSurfaceForms.contains(key)) {
             // we already collected a higher scoring document with this key, in this segment:
-            return false;
+            duplicateSurfaceForm = true;
+          } else {
+            collector.seenSurfaceForms.add(key);
```
jimczi: We should not add the key if we decide to ignore the document below ?

lucene/suggest/src/java/org/apache/lucene/search/suggest/document/TopSuggestDocsCollector.java
```diff
@@ -201,4 +204,19 @@ public void collect(int doc) throws IOException {
   public ScoreMode scoreMode() {
     return ScoreMode.COMPLETE;
   }
+
+  /**
+   * returns true if the collector clearly exhausted all possibilities to collect results
+   */
+  public boolean isComplete() {
```
jimczi: Can we make this package protected and add a boolean in `TopSuggestDocs` that indicates if the search is complete or not ?

lucene/suggest/src/test/org/apache/lucene/search/suggest/document/TestPrefixCompletionQuery.java
```diff
@@ -253,6 +258,61 @@ public void testDocFiltering() throws Exception {
...
+              collected = true;
+          }
+          return collected;
+      }
+    };
+
+    indexSearcher.suggest(query, collector);
+    assertSuggestions(collector.get(), expectedResults.subList(0, topN).toArray(new Entry[0]));
+
+    // TODO expecting true here, why false?
```
jimczi: it seems that NRTSuggesterBuilder#maxAnalyzedPathsPerOutput is not computed correctly. From what I understand it records the number of suggestions with the same analyzed form but the comment says that it should be the highest number of analyzed paths we saw for any input surface form. So imo this is a bug, it's exactly related to this change so we should probably open a new issue for this.

lucene/suggest/src/test/org/apache/lucene/search/suggest/document/TestPrefixCompletionQuery.java
```diff
@@ -253,6 +263,126 @@ public void testDocFiltering() throws Exception {
...
+    Analyzer analyzer = new MockAnalyzer(random());
+    RandomIndexWriter iw = new RandomIndexWriter(random(), dir, iwcWithSuggestField(analyzer, "suggest_field"));
+    Set<Integer> acceptedDocs = new HashSet<>();
+    List<Entry> expectedResults = new ArrayList<Entry>();
+
+    for (int docCount = 0; docCount < 10000; docCount++) {
+      Document document = new Document();
+      String value = "ab" + RandomStrings.randomAsciiAlphanumOfLength(random(), 10) +"_" + docCount;
+      document.add(new SuggestField("suggest_field", value, docCount));
+      if (random().nextDouble() > 0.75) {
```
jimczi: the maximum queue size is `5000` so we should ensure that we don't reject more than this number if we want to ensure that the search is complete. If you change the live docs to contain at least `5000` docs, this test should work fine.

lucene/suggest/src/java/org/apache/lucene/search/suggest/document/NRTSuggester.java
```diff
@@ -283,17 +299,25 @@ public int compare(Pair<Long, BytesRef> o1, Pair<Long, BytesRef> o2) {
...
    * half the number of live documents.
+   *
+   * If the collector can reject documents upon collecting, the queue size is
+   * increased by half the number of live documents again.
+   *
    * <p>
    * The maximum queue size is {@link #MAX_TOP_N_QUEUE_SIZE}
    */
-  private int getMaxTopNSearcherQueueSize(int topN, int numDocs, double liveDocsRatio, boolean filterEnabled) {
+  private int getMaxTopNSearcherQueueSize(int topN, int numDocs, double liveDocsRatio, boolean filterEnabled,
```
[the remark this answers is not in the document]
cbuescher: I considered this as well first, but then moved to the two independent flags. Will revert that part.

lucene/suggest/src/java/org/apache/lucene/search/suggest/document/TopSuggestDocsCollector.java
```diff
@@ -201,4 +205,27 @@ public void collect(int doc) throws IOException {
   public ScoreMode scoreMode() {
     return ScoreMode.COMPLETE;
   }
+
+  /**
+   * returns true if the collector clearly exhausted all possibilities to collect results
+   */
+  boolean isComplete() {
```
jimczi: Is this needed now that we provide the information in the `TopSuggestDocs` ?

lucene/suggest/src/java/org/apache/lucene/search/suggest/document/NRTSuggester.java
```diff
@@ -283,6 +299,7 @@ public int compare(Pair<Long, BytesRef> o1, Pair<Long, BytesRef> o2) {
    * <p>
    * If a <code>filter</code> is applied, the queue size is increased by
    * half the number of live documents.
+   *
```
jimczi: nit: restore the formatting

lucene/suggest/src/java/org/apache/lucene/search/suggest/document/TopSuggestDocs.java
```diff
@@ -116,19 +123,28 @@ public TopSuggestDocs(TotalHits totalHits, SuggestScoreDoc[] scoreDocs) {
...
   public static TopSuggestDocs merge(int topN, TopSuggestDocs[] shardHits) {
     SuggestScoreDocPriorityQueue priorityQueue = new SuggestScoreDocPriorityQueue(topN);
+    boolean allComplete = true;
     for (TopSuggestDocs shardHit : shardHits) {
       for (SuggestScoreDoc scoreDoc : shardHit.scoreLookupDocs()) {
         if (scoreDoc == priorityQueue.insertWithOverflow(scoreDoc)) {
           break;
         }
       }
+      allComplete = allComplete && shardHit.isComplete;
```
jimczi: nit: you can do `allComplete &= shardHit.isComplete` ?

lucene/suggest/src/java/org/apache/lucene/search/suggest/document/TopSuggestDocs.java
```diff
@@ -116,19 +123,28 @@ public TopSuggestDocs(TotalHits totalHits, SuggestScoreDoc[] scoreDocs) {
...
-      return new TopSuggestDocs(new TotalHits(topNResults.length, TotalHits.Relation.EQUAL_TO), topNResults);
+      return new TopSuggestDocs(new TotalHits(topNResults.length, TotalHits.Relation.EQUAL_TO), topNResults,
+          allComplete);
     } else {
       return TopSuggestDocs.EMPTY;
     }
   }
 
+  /**
+   * returns true if we exhausted all possibilities to collect results
```
jimczi: nit: the flag indicates that the list of results is complete but we don't need to exhaust all possibilities to achieve this.  Maybe something like: `indicate if the list of results is complete or not, this might be <code>false</code> if the {@link TopNSearcher} rejected too many results.` ?

lucene/suggest/src/java/org/apache/lucene/search/suggest/document/TopSuggestDocsCollector.java
cbuescher: Not necessary, but since we need the underlying flag and its currently private, I thought its okay to have at least a package private getter. Or does it bloat the class too much? Happy to remove either way, wdyt?

lucene/suggest/src/java/org/apache/lucene/search/suggest/document/TopSuggestDocs.java
cbuescher: Sounds good.

lucene/suggest/src/java/org/apache/lucene/search/suggest/document/TopSuggestDocs.java
```diff
@@ -105,6 +105,10 @@ public TopSuggestDocs(TotalHits totalHits, SuggestScoreDoc[] scoreDocs, boolean
     this.isComplete = isComplete;
   }
 
+  public TopSuggestDocs(TotalHits totalHits, SuggestScoreDoc[] scoreDocs) {
```
cbuescher: I'm not sure about how bwc is handled in Lucene, especially if we intend to backport this to maybe 8.x, so I don't know if we need a ctor without the isComplete flag?

cbuescher:
@jimczi thanks for the review, pushed a commit adressing most of the nits and left two questions. Happy to go either way with both of them, just want to hear your opinion.

lucene/core/src/java/org/apache/lucene/util/fst/Util.java
```diff
@@ -460,11 +460,6 @@ public void addStartPaths(FST.Arc<T> node, T startOutput, boolean allowEmptyStri
           continue;
         }
 
-        if (results.size() == topN-1 && maxQueueDepth == topN) {
-          // Last path -- don't bother w/ queue anymore:
-          queue = null;
```
mikemccand: Whoa, was this opto breaking something?  I guess if this final path is filtered out, we still need the queue?  Have you run the suggest benchmarks to see if removing this opto hurt performance?

lucene/suggest/src/java/org/apache/lucene/search/suggest/document/TopSuggestDocsCollector.java
```diff
@@ -201,4 +205,27 @@ public void collect(int doc) throws IOException {
...
+    return this.isComplete ;
+  }
+
+  /**
+   * call to signal that during collection at least one segment might have returned incomplete results, e.g. because
+   * of too many rejections
+   */
+  void notComplete() {
```
mikemccand: Could we rename to `setNotComplete`?

lucene/suggest/src/java/org/apache/lucene/search/suggest/document/TopSuggestDocs.java
mikemccand: Since the javadocs for this class are marked with the `@lucene.experimental` tag, we are allowed to change the API even in the next dot release (8.x), so it is not necessary to add telescoping constructors here, so we can keep only the constructor above that takes the new `boolean isComplete`.

lucene/suggest/src/java/org/apache/lucene/search/suggest/document/TopSuggestDocs.java
```diff
@@ -116,19 +133,29 @@ public TopSuggestDocs(TotalHits totalHits, SuggestScoreDoc[] scoreDocs) {
...
+      return new TopSuggestDocs(new TotalHits(topNResults.length, TotalHits.Relation.EQUAL_TO), topNResults,
+          allComplete);
     } else {
       return TopSuggestDocs.EMPTY;
     }
   }
 
+  /**
+   * Indicates if the list of results is complete or not. Might be <code>false</code> if the {@link TopNSearcher} rejected
+   * too many of the queued results.
```
mikemccand: Isn't it sometimes `false` if the `TopNSearcher`'s queue filled up and it chose not to pursue some low-scoring partial paths?

lucene/suggest/src/java/org/apache/lucene/search/suggest/document/TopSuggestDocsCollector.java
```diff
@@ -116,7 +118,7 @@ protected void doSetNextReader(LeafReaderContext context) throws IOException {
    * NOTE: collection at the leaf level is guaranteed to be in
    * descending order of score
    */
```
mikemccand: Can you explain the new `boolean` return value in the javadocs here?

lucene/suggest/src/test/org/apache/lucene/search/suggest/document/TestPrefixCompletionQuery.java
```diff
@@ -253,6 +263,123 @@ public void testDocFiltering() throws Exception {
...
+    // use a TopSuggestDocsCollector that rejects results with duplicate docIds
+    TopSuggestDocsCollector collector = new TopSuggestDocsCollector(topN, false) {
+
+      private Set<Integer> seenDocIds = new HashSet<>();
+
+      @Override
+      public boolean collect(int docID, CharSequence key, CharSequence context, float score) throws IOException {
+          int globalDocId = docID + docBase;
+          boolean collected = false;
+          if (seenDocIds.contains(globalDocId) == false) {
```
mikemccand: Hmm why would the suggester even send the same `docID` multiple times?  There is only one suggestion (`SuggestField`) per indexed document in this test ... maybe improve the test to assert that sometimes we reject and sometimes we don't?

lucene/suggest/src/test/org/apache/lucene/search/suggest/document/TestTopSuggestDocs.java
```diff
@@ -0,0 +1,55 @@
...
+
+import org.apache.lucene.search.TotalHits;
+import org.apache.lucene.search.suggest.document.TopSuggestDocs.SuggestScoreDoc;
+import org.apache.lucene.util.LuceneTestCase;
+import org.apache.lucene.util.TestUtil;
+import org.junit.Test;
+
+public class TestTopSuggestDocs extends LuceneTestCase {
+
+  @Test
```
mikemccand: We don't use `@Test` annotations for Lucene's tests ... the test runner/infra infers this by the name of the methods.

lucene/suggest/src/test/org/apache/lucene/search/suggest/document/TestPrefixCompletionQuery.java
```diff
@@ -253,6 +263,123 @@ public void testDocFiltering() throws Exception {
...
+      @Override
+      protected boolean canReject() {
+        return true;
+      }
+    };
+
+    indexSearcher.suggest(query, collector);
+    TopSuggestDocs suggestions = collector.get();
+    assertSuggestions(suggestions, expectedResults.subList(0, topN).toArray(new Entry[0]));
+    assertTrue(suggestions.isComplete());
```
mikemccand: Is it possible to concoct a full test case (invoking suggester) where `isComplete()` returns false?

lucene/suggest/src/test/org/apache/lucene/search/suggest/document/TestPrefixCompletionQuery.java
cbuescher: The collector is called multiple times with the same docID because of the MockSynonymAnalyzer used in the test setup which adds "dog" for "dogs", so each document has two completion paths. This collector is meant to de-duplicate this. I added a note explaining this. This is a simplified version of the behaviour we observe in https://github.com/elastic/elasticsearch/issues/46445.

lucene/core/src/java/org/apache/lucene/util/fst/Util.java
cbuescher: As far as I understand this optimization assumes we surely accept (and collect) the path later in L516s acceptResult(), which always seems to be the case for collectors that don't reject, but if the collector that is eventually called via NRTSuggesters acceptResult() chooses to reject this option, we were losing expected results. This surfaced in the prefix completion tests I added. @jimczi might be able to explain this a bit better than me.

> Have you run the suggest benchmarks to see if removing this opto hurt performance?
No, where are they and how can I run them?

lucene/suggest/src/java/org/apache/lucene/search/suggest/document/TopSuggestDocs.java
jimczi: The admissibility of the search is computed from the reject count so a value of `false` means that we exhausted all the paths but we had to reject all of them so the topN is truncated. It's hard to follow the full logic but it should be ok as long as it is ok to return less than the topN when there are more rejections than the queue size can handle ?

lucene/suggest/src/test/org/apache/lucene/search/suggest/document/TestPrefixCompletionQuery.java
jimczi: I'll open an issue. I also wonder if we shouldn't rely on the fact that the top suggest collector will also early terminate so whenever we expect rejection (because of deleted docs or because we deduplicate on suggestions/doc) we could set the queue size to its maximum value (5000). Currently we have different heuristics that tries to pick a sensitive value automatically but there is no guarantee of admissibility. For instance if we want to deduplicate by document id we should ensure that the queue size is greater than `topN*maxAnalyzedValuesPerDoc` and we'd need to compute this value at index time.
I may be completely off but it would be interesting to see the effects  of setting the queue size to its maximum value on all search. This way the admissibility is easier to reason about and we don't need to correlate it with the choice made by the heuristic.

lucene/suggest/src/test/org/apache/lucene/search/suggest/document/TestPrefixCompletionQuery.java
cbuescher: This will happen if the estimated queue size in `NRTSuggester#lookup` is not large enough to accout for all rejected documents, e.g. when in this particular test we try to get the top 5 of only 5 documents. In that case the queue size heuristic in `NRTSuggester#getMaxTopNSearcherQueueSize` will only size to queue to 7 (topN + numDocs/2), which is less than the number of topN + rejections, so the TopResults returned will have the `isComplete` flag set.
I can add that case to the existing test if this helps.

PR #916, opened 2019-10-02
LUCENE-8213: Asynchronous Caching in LRUQueryCache

atris:


atris:
I still have not figured out why testLUCENE8736 sometimes fails. The test is reproducible only with a seed, the command that reproduces it is:

ant beast -Dbeast.iters=20 -Dtests.dups=6  -Dtestcase=TestLatLonShape -Dtests.method=testLUCENE8736* -Dtests.seed=D784C9A46B62FB84 -Dtests.slow=true -Dtests.badapples=true -Dtests.locale=brx -Dtests.timezone=America/Guatemala -Dtests.asserts=true

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
```diff
@@ -448,13 +491,48 @@ void assertConsistent() {
     }
   }
 
+  // pkg-private for testing
+  void setCountDownLatch(CountDownLatch latch) {
```
jpountz: do you think we could avoid setting a latch here, and maybe instead calling countDown from a subclass' onDocIdSetCache?

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
```diff
@@ -88,13 +93,36 @@
...
   private final long maxRamBytesUsed;
   private final Predicate<LeafReaderContext> leavesToCache;
   // maps queries that are contained in the cache to a singleton so that this
   // cache does not store several copies of the same query
   private final Map<Query, Query> uniqueQueries;
+  // Marks the inflight queries that are being asynchronously loaded into the cache
+  // This is used to ensure that multiple threads do not trigger loading
+  // of the same query in the same cache. We use a set because it is an invariant that
+  // the entries of this data structure be unique.
+  private final Map<MapKey, IndexReader.CacheKey> inFlightAsyncLoadQueries = new HashMap<>();
```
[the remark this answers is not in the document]
atris: I originally used a Set, but moved it to a map specifically to enable tests for the double caching case.

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
```diff
@@ -88,13 +93,36 @@
...
+    }
+
+    public Query getQuery() {
+      return query;
+    }
+
+    public IndexReader.CacheKey getCacheKey() {
+      return cacheKey;
+    }
+  }
```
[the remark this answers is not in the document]
atris: Oops, dont know how it did not make it in this commit, let me check that right away

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
```diff
@@ -813,8 +918,23 @@ public BulkScorer bulkScorer(LeafReaderContext context) throws IOException {
...
-          putIfAbsent(in.getQuery(), docIdSet, cacheHelper);
+          boolean cacheSynchronously = executor == null;
+          // If asynchronous caching is requested, perform the same and return
+          // the uncached iterator
+          if (cacheSynchronously == false) {
+            cacheSynchronously = cacheAsynchronously(context, cacheHelper);
+
+            // If async caching failed, we will perform synchronous caching
+            // hence do not return the uncached value here
+            if (cacheSynchronously == false) {
```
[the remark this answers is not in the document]
atris: Not necessarily -- cacheAsynchronously() might have failed, in which case it will return a true and this code path will trigger a synchronous caching

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
```diff
@@ -368,10 +400,20 @@ public void clearQuery(Query query) {
         onEviction(singleton);
       }
     } finally {
+      removeQuery(query);
```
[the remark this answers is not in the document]
atris: Removed, thanks

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
```diff
@@ -389,6 +431,7 @@ public void clear() {
       cache.clear();
       // Note that this also clears the uniqueQueries map since mostRecentlyUsedQueries is the uniqueQueries.keySet view:
       mostRecentlyUsedQueries.clear();
+      inFlightAsyncLoadQueries.clear();
```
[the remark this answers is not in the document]
atris: Removed

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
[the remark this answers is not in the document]
atris: Moved to set

lucene/core/src/java/org/apache/lucene/search/QueryCache.java
```diff
@@ -33,4 +35,10 @@
    */
   Weight doCache(Weight weight, QueryCachingPolicy policy);
 
+  /**
+   * Same as above, but allows passing in an Executor to perform caching
+   * asynchronously
+   */
+  Weight doCache(Weight weight, QueryCachingPolicy policy, Executor executor);
```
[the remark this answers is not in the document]
atris: We cannot remove this constructor since it is defined in `QueryCache` -- made it directly delegate to the new constructor

atris:
@jpountz Updated, please see and let me know your thoughts

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
```diff
@@ -88,13 +92,47 @@
...
+    public Query getQuery() {
+      return query;
+    }
+
+    public IndexReader.CacheKey getCacheKey() {
+      return cacheKey;
+    }
+
+    @Override
+    public int hashCode() { return query.hashCode() ^ cacheKey.hashCode(); }
```
jpountz: Can you use the usual polynomial formula `h0 + 31 * (h1 + 31 * ( ... ))`. The use of `^` is fine here, but it is a source of performance bugs in some scenarios. For instance Java's AbstractMap.Entry does this, which means that whenever you map a key to itself, the hashcode is 0 (because both hashcodes are equal). This is a lot of collisions if you have lots of entries that have the same key and value. So I prefer avoiding this pattern.

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
```diff
@@ -656,10 +713,21 @@ public long ramBytesUsed() {
...
+      used = new AtomicBoolean(false);
+    }
+
+    CachingWrapperWeight(Weight in, QueryCachingPolicy policy, Executor executor) {
+      super(in.getQuery(), 1f);
+      this.in = in;
+      this.policy = policy;
+      this.executor = executor;
       used = new AtomicBoolean(false);
     }
```
jpountz: nit: can we have a single constructor? The first one looks unused?

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
```diff
@@ -732,8 +821,24 @@ public ScorerSupplier scorerSupplier(LeafReaderContext context) throws IOExcepti
...
+            cacheSynchronously = cacheAsynchronously(context, cacheHelper);
+
+            // If async caching failed, synchronous caching will
+            // be performed, hence do not return the uncached value
+            if (cacheSynchronously == false) {
+              return in.scorerSupplier(context);
+            }
+          }
+
+          if (cacheSynchronously) {
```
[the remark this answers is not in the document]
jpountz: Thanks I had missed it.

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
```diff
@@ -832,5 +931,43 @@ public BulkScorer bulkScorer(LeafReaderContext context) throws IOException {
...
+        return false;
+      }
+
+      FutureTask<Void> task = new FutureTask<>(() -> {
+        try {
+          DocIdSet localDocIdSet = cache(context);
+          putIfAbsent(in.getQuery(), localDocIdSet, cacheHelper);
+        } finally {
+          // Remove the key from inflight
+          Object retValue = inFlightAsyncLoadQueries.remove(new MapKey(in.getQuery(), cacheHelper.getKey()));
```
jpountz: this call is protected by no lock while `inFlightAsyncLoadQueries` is not concurrent?

atris: Removed the map

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
atris: Removed the key for now since the parent set is now removed

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
atris: Good catch, thanks.

atris:
@jpountz  Updated -- removed the in flight map. Please see and let me know your comments.

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
```diff
@@ -450,11 +453,15 @@ void assertConsistent() {
...
-    while (weight instanceof CachingWrapperWeight) {
-      weight = ((CachingWrapperWeight) weight).in;
-    }
+    return doCache(weight, policy, null /* executor */);
+  }
 
-    return new CachingWrapperWeight(weight, policy);
+  // Should be used only when the user wishes to trade throughput for latency
+  // This method was not merged in the method above as to not break the existing contract
+  // advertised by QueryCache
```
jpountz: I think breaking is fine since QueryCache is `@lucene.experimental`.

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
```diff
@@ -832,5 +873,23 @@ public BulkScorer bulkScorer(LeafReaderContext context) throws IOException {
       return new DefaultBulkScorer(new ConstantScoreScorer(this, 0f, ScoreMode.COMPLETE_NO_SCORES, disi));
     }
 
+    // Perform a cache load asynchronously
+    // @return true if synchronous caching is needed, false otherwise
```
jpountz: I think this contract is a bit confusing. Usually `true` would indicate success, like `Lock#tryLock`. Maybe it would be less confusing if `true` meant that we successfully triggered an async cache operation?

lucene/core/src/test/org/apache/lucene/search/TestLRUQueryCache.java
```diff
@@ -1229,17 +1476,53 @@ public void testMinSegmentSizePredicate() throws IOException {
...
+
     searcher.setQueryCache(cache);
+
     searcher.count(new DummyQuery());
-    assertEquals(1, cache.getCacheCount());
+    if (cache.getCacheCount() != 1) {
+      try {
+        latch[0].await(200, TimeUnit.MILLISECONDS);
+      } catch (InterruptedException e) {
+        throw new RuntimeException(e.getMessage());
```
jpountz: Can we avoid swallowing the exception?

atris:
@jpountz Updated, please see

lucene/core/src/java/org/apache/lucene/search/IndexSearcher.java
```diff
@@ -181,7 +181,16 @@ public IndexSearcher(IndexReader r) {
...
-   *  provided Executor. NOTE:
+   *  provided Executor. The passed in Executor will also be
+   *  used by LRUQueryCache (if enabled) to perform asynchronous
+   *  query caching.
+   *  If a task is rejected by the host Executor, the failed task
+   *  will then be executed on the caller thread. This is done to
+   *  ensure that a query succeeds, albeit with a higher latency.
+   *  If a user wishes to modify the said behaviour, they can either
+   *  handle the exception in the provided Executor, or override
+   *  the said method in a custom extension of IndexSearcher.
```
jpountz: it's unclear to me which method you are recommending overriding?

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
```diff
@@ -732,8 +734,25 @@ public ScorerSupplier scorerSupplier(LeafReaderContext context) throws IOExcepti
...
+            asyncCachingSucceeded = cacheAsynchronously(context, cacheHelper);
+
+            // If async caching failed, synchronous caching will
+            // be performed, hence do not return the uncached value
+            if (asyncCachingSucceeded) {
+              return in.scorerSupplier(context);
+            }
+          }
+
+          if (cacheSynchronously || !asyncCachingSucceeded) {
```
jpountz: can you do == false for consistency?

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
```diff
@@ -832,5 +866,23 @@ public BulkScorer bulkScorer(LeafReaderContext context) throws IOException {
       return new DefaultBulkScorer(new ConstantScoreScorer(this, 0f, ScoreMode.COMPLETE_NO_SCORES, disi));
     }
 
+    // Perform a cache load asynchronously
+    // @return true if asynchronous caching succeeded, false otherwise
+    private boolean cacheAsynchronously(LeafReaderContext context, IndexReader.CacheHelper cacheHelper) {
+      FutureTask<Void> task = new FutureTask<>(() -> {
+        DocIdSet localDocIdSet = cache(context);
```
jpountz: Since we are caching asynchronously, we might me hitting a closed segment. We should use `LeafReader#tryIncRef` to make sure the reader is still open and add tests for that case.

jgq2008303393:
Hi, @atris, is this optimization continuing to follow up? Should we skip the cache process directly when IndexOrDocValuesQuery appears with a selective lead iterator?

jpountz:
@jgq2008303393 I agree there might be interesting decisions to make depending on how the cost of the query that is being cached compares with the cost of the overall query. This PR focuses on enabling asynchronous caching, maybe open a JIRA with your ideas?

atris:
> I agree there might be interesting decisions to make depending on how the cost of the query that is being cached compares with the cost of the overall query. This PR focuses on enabling asynchronous caching, maybe open a JIRA with your ideas?

I believe this is the same idea that was the original motivation for the JIRA that led to this PR?

@jpountz I was trying to track your change where you introduced a prototype for a max cost factor when deciding whether to cache a query or not -- should that not suffice for this case?

lucene/core/src/java/org/apache/lucene/search/IndexSearcher.java
atris: Removed

jpountz:
I believe you are referring to https://issues.apache.org/jira/browse/LUCENE-8027. That might work indeed. I wonder how this applies to asynchronous caching: should we also ignore this heuristic when async caching is used because then caching doesn't affect the latency of the current query, or should we still apply it because caching costly queries might further delay upcoming queries in the queue?

jgq2008303393:
# Problem scene
[LUCENE-8027](https://issues.apache.org/jira/browse/LUCENE-8027) is very similar to our idea of solving the problem in our ES clusters(based on Lucene). We have dozens of clusters for metric scenarios. As most of the queries are similar to the following, we frequently encounter some absurdly slow queries.
```
GET host_monitor/_search
{
  "size": 10000, 
  "query": {
    "bool": {
      "filter": [
        {
          "term": {
            "host_ip": "xxx"
          }
        },
        {
          "range": {
            "timestamp": {
              "gte": "now-5d/d",
              "lt": "now/d"
            }
          }
        }
      ]
    }
  },
  "docvalue_fields": ["cpu_usage"]
}
```
After found out that it's a caching issue, we skip the caching process directly when IndexOrDocValuesQuery appears with a selective lead iterator. And the following is the test result. 

Query | 1th | 2th | 3th | 4th | 5th | 6th
-- | -- | -- | -- | -- | -- | --
Before Optimization | 47ms | 42ms | 48ms | 49ms | 742ms | 25ms
After Optimization| 44ms | 45ms | 46ms | 52ms | 43ms | 47ms

As the above result shows:
- Before optimization: after the 5th query consumes 742ms for caching, the next query latency drops to 25ms.
- After optimization: query latency is stable and keeps at 45ms~.

# Our proposal
In general, if there are enough queries to hit the cache, then the asynchronous caching is valuable. However, for a long time range query(e.g. 5 days), each range query will consume tens of megabytes of memory and spend hundreds of milliseconds, but the benefits are not obvious. And those large cache entries will cause frequent cache eviction.

So in my opinion, although asynchronous caching would benefit for most queries, it's still necessary to apply heuristic caching when the range query is too large. 

If you also think that heuristic caching is necessary, I am pleasure to open a new PR and provide our patch refer to [LUCENE-8027](https://issues.apache.org/jira/browse/LUCENE-8027). Looking forward to more discussions. @jpountz @atris

atris:
> I believe you are referring to https://issues.apache.org/jira/browse/LUCENE-8027. That might work indeed. I wonder how this applies to asynchronous caching: should we also ignore this heuristic when async caching is used because then caching doesn't affect the latency of the current query, or should we still apply it because caching costly queries might further delay upcoming queries in the queue?

Yes, that is the one.

I think that is a great solution -- re: asynchronous caching, we can have different cost metrics for caching vs not caching when async caching is enabled/disabled. WDYT?

atris:
@jpountz Updated, please see. I have not added a test for the closed reader case because I am unable to formulate one intuitively , but will add it in further iterations.

lucene/core/src/test/org/apache/lucene/search/TestLRUQueryCache.java
```diff
@@ -244,6 +272,225 @@ public void testLRUEviction() throws Exception {
...
+    searcher.setQueryCachingPolicy(ALWAYS_CACHE);
+
+    // First read should miss
+    searcher.search(new ConstantScoreQuery(red), 1);
+
+    if (!(queryCache.cachedQueries().equals(Collections.emptyList()))) {
+      searcher.search(new ConstantScoreQuery(red), 1);
+    } else {
+      // Let the cache load be completed
+      latch[0].await(200, TimeUnit.MILLISECONDS);
```
jpountz: I'd probably call `await()` without a timeout. The first reason is that CI environments are sometimes very busy and even some simple operations can take very long. The second is that if we get a timeout, we'll have no idea whether the query took too long to cache or whether it was something else. On the other hand, if we wait indefinitely, then the test runner will hit a timeout, and when it does so it emits a thread dump in order to understand what the threads were doing at the time of the timeout. So we'd have more information to understand what happened. I'd suggest removing all timeouts in the below await() calls.

lucene/core/src/test/org/apache/lucene/search/TestLRUQueryCache.java
```diff
@@ -244,6 +272,225 @@ public void testLRUEviction() throws Exception {
...
+    assertEquals(Collections.emptyList(), queryCache.cachedQueries());
+
+    searcher.setQueryCachingPolicy(ALWAYS_CACHE);
+
+    // First read should miss
+    searcher.search(new ConstantScoreQuery(red), 1);
+
+    if (!(queryCache.cachedQueries().equals(Collections.emptyList()))) {
+      searcher.search(new ConstantScoreQuery(red), 1);
+    } else {
```
jpountz: what is the motivation for running code under the `if` statement instead of always running the `else` block?

lucene/core/src/test/org/apache/lucene/search/TestLRUQueryCache.java
```diff
@@ -1229,17 +1476,49 @@ public void testMinSegmentSizePredicate() throws IOException {
...
 
-    cache = new LRUQueryCache(2, 10000, new LRUQueryCache.MinSegmentSizePredicate(0, .6f));
+    if (cache.getCacheCount() != 1) {
+      try {
+        latch[0].await(200, TimeUnit.MILLISECONDS);
+      } catch (InterruptedException e) {
+        throw new RuntimeException(e.getMessage());
+      }
+      assertEquals(1, cache.getCacheCount());
+    }
```
jpountz: should we always wait on the latch instead of first checking whether the cache count is 1?

lucene/core/src/test/org/apache/lucene/search/TestLRUQueryCache.java
```diff
@@ -1229,17 +1476,49 @@ public void testMinSegmentSizePredicate() throws IOException {
...
+        super.onDocIdSetCache(readerCoreKey, ramBytesUsed);
+        latch[0].countDown();
+      }
+    };
+
     searcher.setQueryCache(cache);
+
     searcher.count(new DummyQuery());
-    assertEquals(1, cache.getCacheCount());
+    if (cache.getCacheCount() != 1) {
```
jpountz: same question here

jpountz:
> I have not added a test for the closed reader case because I am unable to formulate one intuitively

Maybe we could do this by creating a fake Executor that waits on a latch before executing any task. This way we might be able to run the query first, then close the reader, and finally call `countDown` on the latch to run the caching operation?

jgq2008303393:
When async caching is enabled, cache operation does not slow down queries, so we should cache as much as possible, which is beneficial to subsequent queries. When async caching is disabled, strict cache conditions are required because cache operation will slow down some queries.

I agree with this idea. BTW, if we can do all cache operation asynchronously in the future, it will be simpler and clearer.

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
```diff
@@ -732,8 +734,25 @@ public ScorerSupplier scorerSupplier(LeafReaderContext context) throws IOExcepti
...
+            asyncCachingSucceeded = cacheAsynchronously(context, cacheHelper);
+
+            // If async caching failed, synchronous caching will
+            // be performed, hence do not return the uncached value
+            if (asyncCachingSucceeded) {
+              return in.scorerSupplier(context);
+            }
+          }
+
+          if (cacheSynchronously || asyncCachingSucceeded == false) {
```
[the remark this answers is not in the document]
atris: Agreed, we do not even need to check the cacheSynchronously flag here since we could get here only if that is true. Removed

lucene/core/src/test/org/apache/lucene/search/TestLRUQueryCache.java
atris: Removed

atris:
@jpountz Updated, please see and let me know your comments.

lucene/core/src/test/org/apache/lucene/search/TestLRUQueryCache.java
```diff
@@ -244,6 +275,213 @@ public void testLRUEviction() throws Exception {
...
+    searcher.search(new ConstantScoreQuery(red), 1);
+
+
+    // Let the cache load be completed
+    latch[0].await();
+    searcher.search(new ConstantScoreQuery(red), 1);
+
+    // Second read should hit
+    searcher.search(new ConstantScoreQuery(red), 1);
+    assertEquals(Collections.singletonList(red), queryCache.cachedQueries());
```
jpountz: shouldn't we be able to assert on this directly after the call to `latch[0].await();` returns?

lucene/core/src/test/org/apache/lucene/search/TestLRUQueryCache.java
```diff
@@ -244,6 +275,213 @@ public void testLRUEviction() throws Exception {
...
+
+    searcher.setQueryCachingPolicy(ALWAYS_CACHE);
+
+    // First read should miss
+    searcher.search(new ConstantScoreQuery(red), 1);
+
+
+    // Let the cache load be completed
+    latch[0].await();
+    searcher.search(new ConstantScoreQuery(red), 1);
```
jpountz: I think we should assert that the hit count incremented, in addition to searching again?

lucene/core/src/test/org/apache/lucene/search/TestLRUQueryCache.java
```diff
@@ -244,6 +275,213 @@ public void testLRUEviction() throws Exception {
...
+
+    latch[0] = new CountDownLatch(1);
+    searcher.search(new ConstantScoreQuery(green), 1);
+
+    // Let the cache load be completed
+    latch[0].await();
+    assertEquals(Arrays.asList(red, green), queryCache.cachedQueries());
+
+    searcher.search(new ConstantScoreQuery(red), 1);
+    assertEquals(Arrays.asList(green, red), queryCache.cachedQueries());
```
jpountz: Check that the hit count incremented?

lucene/core/src/test/org/apache/lucene/search/TestLRUQueryCache.java
```diff
@@ -244,6 +275,213 @@ public void testLRUEviction() throws Exception {
...
+    latch[0] = new CountDownLatch(1);
+
+    searcher.search(new ConstantScoreQuery(blue), 1);
+
+    // Let the cache load be completed
+    latch[0].await();
+    assertEquals(Arrays.asList(red, blue), queryCache.cachedQueries());
+
+    searcher.search(new ConstantScoreQuery(blue), 1);
+    assertEquals(Arrays.asList(red, blue), queryCache.cachedQueries());
```
jpountz: check that the hit count incremented?

lucene/core/src/test/org/apache/lucene/search/TestLRUQueryCache.java
```diff
@@ -244,6 +275,213 @@ public void testLRUEviction() throws Exception {
...
+
+    searcher.search(new ConstantScoreQuery(green), 1);
+
+    // Let the cache load be completed
+    latch[0].await();
+    assertEquals(Arrays.asList(blue, green), queryCache.cachedQueries());
+
+    searcher.setQueryCachingPolicy(NEVER_CACHE);
+    searcher.search(new ConstantScoreQuery(red), 1);
+    assertEquals(Arrays.asList(blue, green), queryCache.cachedQueries());
```
jpountz: maybe move the call to service.shutdown() above this line and also call `awaitTermination` to make sure that any ongoing cache operation are done so that the assertion doesn't succeed only because we are lucky with timing?

lucene/core/src/test/org/apache/lucene/search/TestLRUQueryCache.java
```diff
@@ -244,6 +275,213 @@ public void testLRUEviction() throws Exception {
...
+
+    final Query green = new TermQuery(new Term("color", "green"));
+
+    assertEquals(Collections.emptyList(), queryCache.cachedQueries());
+
+    searcher.setQueryCache(queryCache);
+    searcher.setQueryCachingPolicy(ALWAYS_CACHE);
+
+    FutureTask<Void> task = new FutureTask<>(() -> {
+      searcher.search(new ConstantScoreQuery(green), 1);
```
jpountz: should we make them wait on a latch before starting to make sure they all start at the same time, since the purpose is to test concurrent loading?

lucene/core/src/test/org/apache/lucene/search/TestLRUQueryCache.java
```diff
@@ -1691,4 +1954,180 @@ public void testBulkScorerLocking() throws Exception {
...
+    final Query red = new TermQuery(new Term("color", "red"));
+
+    IndexSearcher searcher = new IndexSearcher(reader, service) {
+      @Override
+      protected LeafSlice[] slices(List<LeafReaderContext> leaves) {
+        ArrayList<LeafSlice> slices = new ArrayList<>();
+        for (LeafReaderContext ctx : leaves) {
+          slices.add(new LeafSlice(Arrays.asList(ctx)));
+        }
+        return slices.toArray(new LeafSlice[0]);
```
jpountz: nit: with recent versions of Java I like `slices.toArray(LeafSlice[]::new);` better

lucene/core/src/test/org/apache/lucene/search/TestLRUQueryCache.java
```diff
@@ -1691,4 +1954,180 @@ public void testBulkScorerLocking() throws Exception {
...
+        }
+
+        latch.countDown();
+
+      }
+    });
+
+    searcher.search(new ConstantScoreQuery(red), 1);
+
+    assertEquals(Collections.singletonList(red), queryCache.cachedQueries());
```
jpountz: This assertion is actually proving that the test is not working? We would except that nothing gets cached since the reader is already closed by the time that the executor needs to cache the query?

lucene/core/src/test/org/apache/lucene/search/TestLRUQueryCache.java
atris: The second search is there to test that once the value is loaded asynchronously -- it exists and does not trigger another load (hence the lack of a wait there). Removed the extra search. thanks

lucene/core/src/test/org/apache/lucene/search/TestLRUQueryCache.java
atris: Hmm, yeah, it is kind of strange, since the reader definitely gets closed before LRUQueryCache tries to cache the value -- but the SegmentReader still seems to be open when the caching is attempted. (I attached a debugger and jumped around).

Do we need to go over all LeafReaderContext instances in the associated searcher and manually close them for this to work the way we expect?

atris:
@jpountz Updated, please see and let me know your thoughts

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
```diff
@@ -813,6 +840,19 @@ public BulkScorer bulkScorer(LeafReaderContext context) throws IOException {
...
         if (policy.shouldCache(in.getQuery())) {
+          boolean cacheSynchronously = executor == null;
+          // If asynchronous caching is requested, perform the same and return
+          // the uncached iterator
+          if (cacheSynchronously == false) {
+            boolean asyncCachingSucceeded = cacheAsynchronously(context, cacheHelper);
+
+            // If async caching failed, we will perform synchronous caching
+            // hence do not return the uncached value here
+            if (asyncCachingSucceeded == false) {
```
jpountz:
```suggestion
            if (asyncCachingSucceeded) {
```

lucene/core/src/test/org/apache/lucene/search/TestLRUQueryCache.java
```diff
@@ -1691,4 +1964,184 @@ public void testBulkScorerLocking() throws Exception {
...
+
+    ExecutorService tempService = new ThreadPoolExecutor(2, 2, 0L, TimeUnit.MILLISECONDS,
+        new LinkedBlockingQueue<Runnable>(),
+        new NamedThreadFactory("TestLRUQueryCache"));
+
+    tempService.submit(new Runnable() {
+      @Override
+      public void run() {
+        try {
+          Thread.sleep(100);
```
jpountz: Why do we need to sleep? In general we aim at never sleeping in tests. The only exception you would find use sleep because it might find test failures that couldn't get exposed otherwise.

lucene/core/src/test/org/apache/lucene/search/TestLRUQueryCache.java
```diff
@@ -1691,4 +1964,184 @@ public void testBulkScorerLocking() throws Exception {
...
+          List<LeafReaderContext> leaves = searcher.leafContexts;
+
+          for (LeafReaderContext leafReaderContext : leaves) {
+            leafReaderContext.reader().close();
+          }
```
jpountz: I'd rather do `searcher.getIndexReader().close()` than close at the leaf level directly

lucene/core/src/test/org/apache/lucene/search/TestLRUQueryCache.java
```diff
@@ -1691,4 +1964,184 @@ public void testBulkScorerLocking() throws Exception {
...
+        } catch (Exception e) {
+          throw new RuntimeException(e.getMessage());
+        }
+
+        latch.countDown();
+
+      }
+    });
+
+    expectThrows(AlreadyClosedException.class, () -> searcher.search(new ConstantScoreQuery(red), 1));
```
jpountz: Actually the interesting case is that the search succeeds but the caching fails? Can we simulate this case? Maybe we could create a searcher with a single leaf to force IndexSearcher to run search on the current thread, so that we would know that the only thing that runs in the threadpool is the caching? Or maybe we could wrap the cache in order to force it to use a special BlockedMockExecutor instead of the Executor of the IndexSearcher?

lucene/core/src/test/org/apache/lucene/search/TestLRUQueryCache.java
atris: This sleep was introduced to let the query caching get initiated before the readers are closed -- but I dont think that is actually required. Removed.

lucene/core/src/test/org/apache/lucene/search/TestLRUQueryCache.java
atris: That does not seem to be triggering the error during increment of the reference count.

lucene/core/src/test/org/apache/lucene/search/TestLRUQueryCache.java
atris: That is a good idea. Would just using `RejectingMockExecutor` suffice for that case?

atris:
@jpountz Updated, please see and let me know

lucene/core/src/test/org/apache/lucene/search/TestLRUQueryCache.java
atris: Added a test using `RejectedExecution` to simulate the case where the search succeeds, but the caching fails. Expected behaviour is that the caching should then be done on the caller thread and should still show cached values.

atris:
@jpountz Please let me know your thoughts on this

jpountz:
@atris I tried to build a test case for the above case:

```java
  public void testAsyncCachingHitsClosedReader() throws Exception {
    final RandomIndexWriter w = new RandomIndexWriter(random(), dir);
    Document doc = new Document();
    doc.add(new StringField("color", "red", Store.NO));
    w.addDocument(doc);
    IndexReader reader = w.getReader();
    w.close();

    // Because our index has a single document, it also has a single segment. In that
    // case, IndexSearcher will use the current thread for searching, so the only thing
    // that runs via the executor is the caching of queries.
    final CountDownLatch[] awaitCaching = new CountDownLatch[1];
    awaitCaching[0] = new CountDownLatch(1);
    final CountDownLatch[] cachingRan = new CountDownLatch[1];
    cachingRan[0] = new CountDownLatch(1);
    final AtomicBoolean success = new AtomicBoolean(false);
    Executor executor = runnable -> {
      new Thread(() -> {
        try {
          awaitCaching[0].await();
          runnable.run();
          success.set(true);
        } catch (InterruptedException e) {
          Thread.currentThread().interrupt();
        } finally {
          cachingRan[0].countDown();
        }
      }).start();      
    };

    final IndexSearcher searcher = new IndexSearcher(reader, executor);
    final LRUQueryCache queryCache = new LRUQueryCache(2, 100000, context -> true);
    searcher.setQueryCache(queryCache);
    searcher.setQueryCachingPolicy(ALWAYS_CACHE);

    searcher.search(new ConstantScoreQuery(new TermQuery(new Term("color", "red"))), 1);
    assertEquals(Collections.emptyList(), queryCache.cachedQueries());
    awaitCaching[0].countDown();
    cachingRan[0].await();
    assertTrue(success.get());
    assertEquals(Collections.singletonList(new TermQuery(new Term("color", "red"))), queryCache.cachedQueries());

    awaitCaching[0] = new CountDownLatch(1);
    cachingRan[0] = new CountDownLatch(1);
    success.set(false);
    queryCache.clear();

    searcher.search(new ConstantScoreQuery(new TermQuery(new Term("color", "red"))), 1);
    assertEquals(Collections.emptyList(), queryCache.cachedQueries());
    reader.close();
    awaitCaching[0].countDown();
    cachingRan[0].await();
    assertTrue(success.get());
    assertEquals(Collections.emptyList(), queryCache.cachedQueries());

    dir.close();
  }
```

atris:
@jpountz Thanks for adding the test. I have updated the PR with same and resolved conflicts. Please let me know if it seems fine now

atris:
Update -- Changed tests. Test suite came in clean

atris:
Ran test suite again, and beasted LRUQueryCache tests 20 odd times -- no failure.

Thanks @jpountz !

PR #919, opened 2019-10-02
LUCENE-8994: Code Cleanup - Pass values to list constructor instead of empty constructor followed by addAll().

KoenDG:
https://issues.apache.org/jira/projects/LUCENE/issues/LUCENE-8994

If you have actual serious issues to attend, no need to bother with this PR, it is code cleanup, not features or fixes.

A small and unimportant PR. Some code cleanup. And perhaps in some cases a small performance gain.

I would understand if issue was taken concerning readability in some cases.

I could change those cases to look something like this, if it made if more readable:

```
new ArrayList<>(
    nameOfCollection.getMethod(someExtraVars)
);
```

Frankly, it already was equally unreadable in such cases as:

```
resources.addAll(Accountables.namedAccountables("field", fields));
```

Depends on what the reviewer wants, I suppose.


lucene/CHANGES.txt
```diff
@@ -180,6 +180,8 @@ Other
 * LUCENE-8993, LUCENE-8807: Changed all repository and download references in build files
   to HTTPS. (Uwe Schindler)
 
+* LUCENE-8994: Code Cleanup - Pass values to list constructor instead of empty constructor followed by addAll().
```
jpountz: can you add your name in parentheses?

KoenDG:
Updated as requested

PR #922, opened 2019-10-04
SOLR-13793: limit number of remote queries to number of shard replicas

kesharee:
- Limit number of remote queries to number of shard replicas when considering inactive replicas
- Unchecked remote queries from HttpSolrCall lead to mutual recursive calls to remote node


munendrasn:
This is released with 8.3

PR #923, opened 2019-10-04
LUCENE-8988: Introduce Global Feature Based Early Termination For Sorted Fields

atris:
This commit introduces maximum-of-minimum based early termination for TopFieldCollector. LUCENE-8949 is also a part of this commit since it is a prerequisite for this change.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/FieldValueChecker.java
+++ b/lucene/core/src/java/org/apache/lucene/search/FieldValueChecker.java
@@ -25,6 +25,7 @@ public class FieldValueChecker {
   public volatile int minimumDoc;
   private FieldComparator[] fieldComparators;
   private int[] reverseMul;
+  private boolean isEnabled = false;
 
   public FieldValueChecker(final FieldComparator[] fieldComparators, final int[] reverseMul) {
     assert fieldComparators.length == reverseMul.length;
@@ -33,7 +34,15 @@ public class FieldValueChecker {
     this.reverseMul = reverseMul;
   }
 
+  void disableGlobalHitCheck() {
+    isEnabled = false;
+  }
+
   boolean isBottomValuePresent() {
+    if (!isEnabled) {
+      return false;
+    }
+
     return this.value != null;
   }
 
--- a/lucene/core/src/java/org/apache/lucene/search/FieldValueChecker.java
+++ b/lucene/core/src/java/org/apache/lucene/search/FieldValueChecker.java
@@ -21,7 +21,7 @@ package org.apache.lucene.search;
  * Maintains bottom feature value across multiple collectors
  */
 public class FieldValueChecker {
-  private volatile Object value;
+  public volatile Object value;
   private FieldComparator[] fieldComparators;
   private int[] reverseMul;
 
@@ -51,9 +51,9 @@ public class FieldValueChecker {
   }
 
   boolean isValueCompetitive(Object value, int doc) {
-    if (value == null || this.value == null) {
+    if (this.value == null) {
       //System.out.println("Null case false " + doc);
-      return false;
+      return true;
     }
 
     if (value instanceof Object[]) {
--- a/lucene/core/src/java/org/apache/lucene/search/FieldValueChecker.java
+++ b/lucene/core/src/java/org/apache/lucene/search/FieldValueChecker.java
@@ -47,23 +47,31 @@ public class FieldValueChecker {
   }
 
   void checkAndUpdateBottomValue(Object value, int doc) {
+    if (!isEnabled) {
+      return;
+    }
+
     synchronized (this) {
       if (this.value == null) {
-        //System.out.println("Setting value " + value);
         this.value = value;
         this.minimumDoc = doc;
         return;
       }
+    }
 
-      if (isValueCompetitive(value, doc)== true) {
-        //System.out.println("Updating value " + this.value + " new value " + value);
-        this.value = value;
-        this.minimumDoc = doc;
-      }
+    if (isValueCompetitive(value, doc)== true) {
+      this.value = value;
+      this.minimumDoc = doc;
     }
   }
 
-  boolean isValueCompetitive(Object value, int doc) {
+  /**
+   * Check if the input value is globally competitive
+   * @param value value to be checked
+   * @param doc docID which generated the value
+   * @return true if competitive, false otherwise
+   */
+  synchronized boolean isValueCompetitive(Object value, int doc) {
     if (this.value == null) {
       return true;
     }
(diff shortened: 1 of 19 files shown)


atris:
> Hmm this is sort of smelly -- why is it OK for the comparator to not implement this method?

A comparator can choose not to implement this API -- it essentially disables the global value checking. This can be done if getting the value is tough, or if it is expensive

atris:
Any thoughts on this one? Seems useful enough?

lucene/core/src/java/org/apache/lucene/search/LeafFieldComparator.java
```diff
@@ -116,4 +116,12 @@
...
   void setScorer(Scorable scorer) throws IOException;
 
+  /** Publishes feature values for the given docID
+   *  The passed in docID values should only move ahead --
+   *  DocValues do not support reverse iterations.
+   *  WARNING: Can be a costly API, especially for sorted fields
+   *
+   *  @param doc docID relative to current reader
+   */
+  T getDocValue(int doc) throws IOException;
```
jimczi: I don't understand why this is needed, if you know the bottom slot you can simply call `FieldComparator::value(int)` ?

lucene/core/src/java/org/apache/lucene/search/TopFieldCollector.java
```diff
@@ -130,8 +130,25 @@ public void setScorer(Scorable scorer) throws IOException {
...
           ++totalHits;
           hitsThresholdChecker.incrementHitCount();
-          if (queueFull) {
-            if (collectedAllCompetitiveHits || reverseMul * comparator.compareBottom(doc) <= 0) {
+
+          Object currentValue = null;
+
+          if (fieldValueChecker != null) {
+            try {
+              currentValue = comparator.getDocValue(doc);
```
jimczi: We should consider this operation as very costly. Getting the actual value of a string field means resolving the ordinal and we have field comparator that can avoid this lookup efficiently.

lucene/core/src/java/org/apache/lucene/search/FieldValueChecker.java
```diff
@@ -0,0 +1,114 @@
...
+
+    synchronized (this) {
+      if (this.value == null) {
+        this.value = value;
+        this.minimumDoc = doc;
+        return;
+      }
+    }
+
+    if (isValueCompetitive(value, doc)== true) {
```
jimczi: There is no synchronization here, so you can end up comparing a minimum doc that is not associated to the current value ?

lucene/core/src/java/org/apache/lucene/search/TopFieldCollector.java
atris: I agree and disagree -- some comparators can be costly to get the leaf values from, as the comment highlights. However, for the remaining comparators (Int, Float, Double, Long), this is a fairly cheap operation

lucene/core/src/java/org/apache/lucene/search/FieldValueChecker.java
atris: I am not too sure here -- the checks before this assignment are synchronized. However, I do agree that being safe is key here -- will update to include this in the synchronized block, thanks

lucene/core/src/java/org/apache/lucene/search/LeafFieldComparator.java
atris: `comparator` here is a `LeafFieldComparator`, and we need the feature value for a given docID, which might not necessarily be the bottom value.

atris:
Made the tests more extensive and added additional updates of global scores. Note that the update of global score can be done as a checkpoint as suggested by @jpountz in a separate thread, but I believe that can be done as a follow up change.

jimczi:
> The minimum score is only applicable for indices that are sorted by relevance, whereas this change enables early terminating segments that are non competitive for all sort keys (irrespective of whether they are sorted by relevance or not) ?

That's the part that I don't really understand. For sorted indices, early termination will require to visit only a few documents per segment so I wonder if the extra checks will make a difference or if they will simply slow down everything due to contention. As said above, we have a way to skip group of documents efficiently when sorted by relevancy but for other sort we can just bypass some checks in the priority queue. I am not sure that we'll gain anything with  a synchronized call on every hit. I could be convinced otherwise but we'll need some benchmark to verify that the change really brings something.

> Note that the update of global score can be done as a checkpoint as suggested by @jpountz in a separate thread, but I believe that can be done as a follow up change.

I'd prefer that we find the best way to do this and then push to master. As it is there is no proof that the change brings anything and I'd like to commit #904 first so that you can build on top of it.

atris:
> That's the part that I don't really understand. For sorted indices, early termination will require to visit only a few documents per segment

That is true *if* the segment is not composed of many competitive hits relative to the thread local PQ’s bottom. It is possible that a large segment might be deemed as competitive by the local bottom, but the global bottom thinks otherwise and leads to early termination of the segment.

> we have a way to skip group of documents efficiently when sorted by relevancy but for other sort we can just bypass some checks in the priority queue. 

I am not sure if I understand how relaxing priority queue checks will allow global early termination?


> I am not sure that we'll gain anything with a synchronized call on every hit. I could be convinced otherwise but we'll need some benchmark to verify that the change really brings something.

That’s what benchmarks are for :) unfortunately luceneutil does not support sorted indices during querying — I know you had a local hack for the relevance skipping patch. Mind sharing details of the same, please?

> I'd prefer that we find the best way to do this and then push to master. As it is there is no proof that the change brings anything and I'd like to commit #904 first so that you can build on top of it.

I do not see how this PR blocks #904? The mentioned PR focuses on skipping documents when sorted by relevance and this PR focuses on early terminating segments which might be deemed competitive at the thread level but are not at a global level

jimczi:
> I am not sure if I understand how relaxing priority queue checks will allow global early termination?

My point here is that I don't see a lot of room for improvement for early termination on sorted indices. I might be wrong but this is why I am asking for benchmarks.

> I do not see how this PR blocks #904? The mentioned PR focuses on skipping documents when sorted by relevance and this PR focuses on early terminating segments which might be deemed competitive at the thread level but are not at a global level

We're changing the same class (`TopFieldCollector`) ;) so I'd prefer that we do one change at a time, possibly reusing the changes of #904 to implement the ideas listed here ?

lucene/core/src/java/org/apache/lucene/search/FieldValueChecker.java
```diff
@@ -0,0 +1,114 @@
...
+package org.apache.lucene.search;
+
+/**
+ * Maintains bottom feature value across multiple collectors
+ */
+public class FieldValueChecker {
+  public volatile Object value;
+  public volatile int minimumDoc;
+  private FieldComparator[] fieldComparators;
+  private int[] reverseMul;
```
mikemccand: `final`?

lucene/core/src/test/org/apache/lucene/search/TestGetDocValuesLeafFieldComparators.java
```diff
@@ -0,0 +1,105 @@
...
+
+              if (value instanceof Long) {
+                long castedValue = (long) value;
+                assertNotNull(set.contains(castedValue));
+                ++count;
+              } else if (value instanceof Float) {
+                float castedValue = (float) value;
+                assertNotNull(set.contains(castedValue));
+                ++count;
+              }
```
mikemccand: `else` throw exception to fail the test?

lucene/core/src/java/org/apache/lucene/search/FieldValueChecker.java
```diff
@@ -0,0 +1,114 @@
...
+    this.fieldComparators = fieldComparators;
+    this.reverseMul = reverseMul;
+  }
+
+  void disableGlobalHitCheck() {
+    isEnabled = false;
+  }
+
+  boolean isBottomValuePresent() {
+    if (!isEnabled) {
```
mikemccand: `isEnabled == false`?

jpountz:
+1 to run benchmarks. I have a similar intuition to Jim that early termination for sorted indices is so efficient that adding overhead to add smarter exit conditions might make things slower overall.

PR #927, opened 2019-10-05
LUCENE-8997: : Add type of triangle info to ShapeField encoding

iverase:
# Description

# Solution

# Tests

# Checklist

### DIFF
--- a/lucene/sandbox/src/java/org/apache/lucene/document/ShapeField.java
+++ b/lucene/sandbox/src/java/org/apache/lucene/document/ShapeField.java
@@ -59,18 +59,60 @@ public final class ShapeField {
    */
   public static class Triangle extends Field {
 
-    // constructor for points and lines
-    Triangle(String name, int aXencoded, int aYencoded, int bXencoded, int bYencoded, int cXencoded, int cYencoded) {
+    // constructor for points
+    Triangle(String name, int aXencoded, int aYencoded) {
       super(name, TYPE);
-      setTriangleValue(aXencoded, aYencoded, true, bXencoded, bYencoded, true, cXencoded, cYencoded, true);
+      setTriangleValue(aXencoded, aYencoded);
     }
 
+    // constructor for points and lines
+    Triangle(String name, int aXencoded, int aYencoded, int bXencoded, int bYencoded) {
+      super(name, TYPE);
+      if (aXencoded == bXencoded && aYencoded == bYencoded) {
+        setTriangleValue(aXencoded, aYencoded);
+      } else {
+        setTriangleValue(aXencoded, aYencoded,  bXencoded, bYencoded);
+      }
+    }
 
+    // generic constructor
     Triangle(String name, Tessellator.Triangle t) {
       super(name, TYPE);
-      setTriangleValue(t.getEncodedX(0), t.getEncodedY(0), t.isEdgefromPolygon(0),
-                       t.getEncodedX(1), t.getEncodedY(1), t.isEdgefromPolygon(1),
-                       t.getEncodedX(2), t.getEncodedY(2), t.isEdgefromPolygon(2));
+      int aX = t.getEncodedX(0);
+      int aY = t.getEncodedY(0);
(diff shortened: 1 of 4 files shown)


lucene/sandbox/src/java/org/apache/lucene/document/ShapeField.java
```diff
@@ -101,21 +143,95 @@ protected void setTriangleValue(int aX, int aY, boolean abFromShape, int bX, int
   private static final int MAXY_MINX_MINY_X_Y_MAXX = 6;
   private static final int MINY_MINX_Y_MAXX_MAXY_X = 7;
```
jpountz: What would be the pros/cons vs. adding new triangle types, e.g. as below?

```
private static final int POINT_Y_X = 8;
private static final int LINE_MAXY_MINX_MINY_MAXX = 9;
private static final int LINE_MINY_MINX_MAXY_MAXX = 10;
```

iverase: I missed this update. The problem here is that we are using the first three bits to encode the type of triangle so this will make the change not backwards compatible although the benefit is that you only use one bit for adding the extra information.

iverase: I guess we can get the version of the segment and act accordingly.

iverase:
I added new triangle types but no sure if this can work. Does Lucene merge segments from different Lucene versions? I would expect so, therefore this won't work as you can have different encodings in the same segment?

lucene/sandbox/src/java/org/apache/lucene/document/ShapeField.java
jpountz: Indeed this change needs to be backward compatible, but is it really an issue here? My understanding is that with all segments you would get points reported as triangles, but in practice that wouldn't be an issue? And new indices with this change should never get queried with old versions of Lucene that don't know about these constants (Lucene doesn't guarantee any forward compatibility) so that wouldn't be an issue either?

jpountz:
@iverase Can you expand on why this would be an issue for backward compatibility?

iverase:
@jpountz of course, this is my logic.

Considering two versions of Lucene X.Y and X.Z where Z > Y and with the current ShapeField formats:

X.Y: Three first bits are the triangle type, next three bits are the edges features (current format)
X.Z: Four first bits are the triangle type (including point and lines) and the next three bits are the edges features (new format).

I consider it is possible to have an index created with X.Y and at some point in time this index is starting getting written by X.Z (user has upgraded the Lucene version). At some point merging might happen so we might end up with versions with points in different encoding in the same segment.

In this case, the encoding needs to tells us which format it contains, does it make sense?

lucene/sandbox/src/java/org/apache/lucene/document/ShapeField.java
```diff
@@ -260,111 +379,249 @@ public static void encodeTriangle(byte[] bytes, int aLat, int aLon, boolean abFr
...
+        decodePoint(t,  triangle);
+      }
+    } else if (bit7) {
+      triangle.type = DecodedTriangle.TYPE.LINE;
+      decodeLine(t, triangle, bits);
+    } else {
+      // for backwards compatibility
+      triangle.type = DecodedTriangle.TYPE.UNKNOWN;
+      decodeTriangle(t, triangle, bits);
+    }
```
jpountz: Let's read the two bits at once? E.g.

```java
final int typeBits = (bits >>> 6) & 0x03;
switch (typeBits) {
  case 0:
    triangle.type = DecodedTriangle.TYPE.UNKNOWN;
    decodeTriangle(t, triangle, bits);
    break; 
  [...]
  case 3:
    triangle.type = DecodedTriangle.TYPE.TRIANGLE;
    decodeTriangle(t, triangle, bits);
    break;
}
```

iverase:
Not any meaningful speed up but I hope to get a bit more when using this information to specialise the relate methods on the queries.

PR #933, opened 2019-10-09
SOLR-13209 NullPointerException from call in org.apache.solr.search.SolrIndexSearcher.getDocSet

CaderHancock:
# Description

The root of this issue is the absence of error handling when a null query is passed to SolrIndexSearcher or Grouping$CommandQuery.

# Solution
In the first case, this is solved by returning DocSet.EMPTY before any calls are made on the null reference (absQ).

In the second case, this is solved by a local variable assigned to a ternary statement that returns an empty String literal when the query is null. So that no method calls are made on the null reference.

While the second issue is not in the JIRA, it was discovered after applying the fix for the first case.

The new behavior is that an empty DocList is returned to the user instead of a 500 error and NPE.

# Checklist

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -64,6 +64,32 @@ Other
 
 * LUCENE-8768: Fix Javadocs build in Java 11. (Namgyu Kim)
 
+======================= Lucene 8.4.0 =======================
+
+API Changes
+---------------------
+(No changes)
+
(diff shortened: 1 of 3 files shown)


CaderHancock:
After working on SOLR-13205, I realized that I should throw a SolrException instead of returning DocSet.EMPTY. Will fix.

PR #934, opened 2019-10-09
Fix typo

harry-wood:
I _think_ there is a missing word here. "one **which** has too many states"


janhoy:
Merged for master and 8.4. Thanks!

PR #939, opened 2019-10-10
LUCENE-9003: Compute numDocs() lazily.

jpountz:
See [LUCENE-9003](https://issues.apache.org/jira/browse/LUCENE-9003) for background.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/BaseCompositeReader.java
+++ b/lucene/core/src/java/org/apache/lucene/index/BaseCompositeReader.java
@@ -21,6 +21,7 @@ import java.io.IOException;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
+import java.util.concurrent.atomic.AtomicInteger;
 
 /** Base class for implementing {@link CompositeReader}s based on an array
  * of sub-readers. The implementing class has to add code for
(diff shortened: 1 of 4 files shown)


lucene/core/src/java/org/apache/lucene/index/BaseCompositeReader.java
```diff
@@ -89,7 +88,6 @@ protected BaseCompositeReader(R[] subReaders) throws IOException {
 
     this.maxDoc = Math.toIntExact(maxDoc);
     starts[subReaders.length] = this.maxDoc;
-    this.numDocs = Math.toIntExact(numDocs);
```
jimczi: I guess it was done because we didn't always check the maximum number of documents allowed in an index but now we have better checks so we can safely assume that the sum will not overflow ?

jpountz: I think it's not necessary on numDocs. By the time we compute numDocs(), we already checked that maxDoc values summed up to a number that doesn't overflow. Since numDocs is less than or equal to maxDoc, the sum of numDocs shouldn't overflow either?

lucene/core/src/java/org/apache/lucene/index/BaseCompositeReader.java
```diff
@@ -102,6 +101,23 @@ public final Fields getTermVectors(int docID) throws IOException {
...
+    // hence the opaque read.
+    // http://gee.cs.oswego.edu/dl/html/j9mm.html#opaquesec.
+    int numDocs = this.numDocs.getOpaque();
+    if (numDocs == -1) {
+      numDocs = 0;
+      for (IndexReader r : subReaders) {
+        numDocs += r.numDocs();
+      }
+      assert numDocs >= 0;
+      this.numDocs.set(numDocs);
```
danmuzi: Isn't
`this.numDocs.setOpaque(numDocs)`
right?

uschindler:
Set is harder than setOpaque.

danmuzi:
Oh, thanks. @uschindler

PR #940, opened 2019-10-11
LUCENE-9002: Query caching leads to absurdly slow queries

jgq2008303393:
# Description
We have dozens of ES clusters(based on Lucene) for metric scenarios. Most of the queries are like this: host_ip:10.10.10.10 AND timestamp:[2019-10-01 00:00:00 TO 2019-10-05 23:59:59]. And we frequently encounter some absurdly slow queries.

# Solution
For a long time range query(e.g. 5 days), each range query will consume tens of megabytes of memory and spend hundreds of milliseconds to cache, but the benefits are not obvious. And those large cache entries will cause frequent cache eviction. So it's better to skip the caching action directly when large range query appears with a selective lead iterator.

Related to [LUCENE-9002](https://issues.apache.org/jira/projects/LUCENE/issues/LUCENE-9002)

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
+++ b/lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
@@ -101,7 +101,7 @@ public class LRUQueryCache implements QueryCache, Accountable {
   private final Set<Query> mostRecentlyUsedQueries;
   private final Map<IndexReader.CacheKey, LeafCache> cache;
   private final ReentrantLock lock;
-  private final long skipCacheFactor;
+  private final float skipCacheFactor;
 
   // these variables are volatile so that we do not need to sync reads
   // but increments need to be performed under the lock
@@ -120,11 +120,15 @@ public class LRUQueryCache implements QueryCache, Accountable {
    * will not be cached in order to not slow down queries too much.
    */
   public LRUQueryCache(int maxSize, long maxRamBytesUsed,
-                       Predicate<LeafReaderContext> leavesToCache, long skipCacheFactor) {
+                       Predicate<LeafReaderContext> leavesToCache, float skipCacheFactor) {
     this.maxSize = maxSize;
     this.maxRamBytesUsed = maxRamBytesUsed;
     this.leavesToCache = leavesToCache;
+    if (skipCacheFactor < 1) {
+      throw new IllegalArgumentException("skipCacheFactor must be no less than 1, get " + skipCacheFactor);
+    }
     this.skipCacheFactor = skipCacheFactor;
+
     uniqueQueries = new LinkedHashMap<>(16, 0.75f, true);
     mostRecentlyUsedQueries = uniqueQueries.keySet();
     cache = new IdentityHashMap<>();
@@ -748,7 +752,7 @@ public class LRUQueryCache implements QueryCache, Accountable {
             @Override
             public Scorer get(long leadCost) throws IOException {
               // skip cache operation which would slow query down too much
-              if (cost > leadCost * skipCacheFactor) {
+              if (cost / skipCacheFactor > leadCost) {
                 return supplier.get(leadCost);
               }
 
--- a/lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
+++ b/lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
@@ -101,6 +101,8 @@ public class LRUQueryCache implements QueryCache, Accountable {
   private final Set<Query> mostRecentlyUsedQueries;
   private final Map<IndexReader.CacheKey, LeafCache> cache;
   private final ReentrantLock lock;
+  private final long skipCacheFactor;
+  private final long skipCacheCost;
 
   // these variables are volatile so that we do not need to sync reads
   // but increments need to be performed under the lock
@@ -114,12 +116,19 @@ public class LRUQueryCache implements QueryCache, Accountable {
    * Expert: Create a new instance that will cache at most <code>maxSize</code>
    * queries with at most <code>maxRamBytesUsed</code> bytes of memory, only on
    * leaves that satisfy {@code leavesToCache}.
+   *
+   * Also, clauses whose cost is
+   * over than {@code skipCacheCost} and more than {@code skipCacheFactor} times the cost of the top-level query
+   * will not be cached in order to not slow down queries too much due to caching.
    */
   public LRUQueryCache(int maxSize, long maxRamBytesUsed,
-      Predicate<LeafReaderContext> leavesToCache) {
+                       Predicate<LeafReaderContext> leavesToCache,
+                       long skipCacheCost, long skipCacheFactor) {
     this.maxSize = maxSize;
     this.maxRamBytesUsed = maxRamBytesUsed;
     this.leavesToCache = leavesToCache;
+    this.skipCacheCost = skipCacheCost;
+    this.skipCacheFactor = skipCacheFactor;
     uniqueQueries = new LinkedHashMap<>(16, 0.75f, true);
     mostRecentlyUsedQueries = uniqueQueries.keySet();
     cache = new IdentityHashMap<>();
@@ -141,7 +150,7 @@ public class LRUQueryCache implements QueryCache, Accountable {
    * be cached in order to not hurt latency too much because of caching.
    */
   public LRUQueryCache(int maxSize, long maxRamBytesUsed) {
-    this(maxSize, maxRamBytesUsed, new MinSegmentSizePredicate(10000, .03f));
+    this(maxSize, maxRamBytesUsed, new MinSegmentSizePredicate(10000, .03f), 250, 5000000);
   }
 
   // pkg-private for testing
@@ -732,8 +741,39 @@ public class LRUQueryCache implements QueryCache, Accountable {
 
       if (docIdSet == null) {
         if (policy.shouldCache(in.getQuery())) {
-          docIdSet = cache(context);
-          putIfAbsent(in.getQuery(), docIdSet, cacheHelper);
+          final ScorerSupplier supplier = in.scorerSupplier(context);
+          if (supplier == null) {
+            putIfAbsent(in.getQuery(), DocIdSet.EMPTY, cacheHelper);
+            return null;
+          }
+
+          final long cost = supplier.cost();
+          return new ScorerSupplier() {
+            @Override
+            public Scorer get(long leadCost) throws IOException {
+              // skip cache operation which would slow query down too much
+              if (cost > skipCacheCost && cost > leadCost * skipCacheFactor
+                  && in.getQuery() instanceof IndexOrDocValuesQuery) {
+                return supplier.get(leadCost);
+              }
+
+              Scorer scorer = supplier.get(Long.MAX_VALUE);
+              DocIdSet docIdSet = cacheImpl(new DefaultBulkScorer(scorer), context.reader().maxDoc());
+              putIfAbsent(in.getQuery(), docIdSet, cacheHelper);
+              DocIdSetIterator disi = docIdSet.iterator();
+              if (disi == null) {
+                // docIdSet.iterator() is allowed to return null when empty but we want a non-null iterator here
+                disi = DocIdSetIterator.empty();
+              }
+
+              return new ConstantScoreScorer(CachingWrapperWeight.this, 0f, ScoreMode.COMPLETE_NO_SCORES, disi);
+            }
+
+            @Override
+            public long cost() {
+              return cost;
+            }
+          };
         } else {
           return in.scorerSupplier(context);
         }
@@ -753,7 +793,7 @@ public class LRUQueryCache implements QueryCache, Accountable {
         public Scorer get(long LeadCost) throws IOException {
           return new ConstantScoreScorer(CachingWrapperWeight.this, 0f, ScoreMode.COMPLETE_NO_SCORES, disi);
         }
-        
+
         @Override
         public long cost() {
           return disi.cost();
--- a/lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
+++ b/lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
@@ -752,7 +752,7 @@ public class LRUQueryCache implements QueryCache, Accountable {
             @Override
             public Scorer get(long leadCost) throws IOException {
               // skip cache operation which would slow query down too much
-              if (cost > skipCacheCost && cost > leadCost * skipCacheFactor
+              if ((cost > skipCacheCost || cost > leadCost * skipCacheFactor)
                   && in.getQuery() instanceof IndexOrDocValuesQuery) {
                 return supplier.get(leadCost);
               }
--- a/lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
+++ b/lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
@@ -124,7 +124,7 @@ public class LRUQueryCache implements QueryCache, Accountable {
     this.maxSize = maxSize;
     this.maxRamBytesUsed = maxRamBytesUsed;
     this.leavesToCache = leavesToCache;
-    if (skipCacheFactor < 1) {
+    if (skipCacheFactor >= 1 == false) { // NaN >= 1 evaluates true
       throw new IllegalArgumentException("skipCacheFactor must be no less than 1, get " + skipCacheFactor);
     }
     this.skipCacheFactor = skipCacheFactor;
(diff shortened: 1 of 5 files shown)


lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
```diff
@@ -732,8 +741,39 @@ public ScorerSupplier scorerSupplier(LeafReaderContext context) throws IOExcepti
...
+            return null;
+          }
+
+          final long cost = supplier.cost();
+          return new ScorerSupplier() {
+            @Override
+            public Scorer get(long leadCost) throws IOException {
+              // skip cache operation which would slow query down too much
+              if (cost > skipCacheCost && cost > leadCost * skipCacheFactor
+                  && in.getQuery() instanceof IndexOrDocValuesQuery) {
```
jgq2008303393: This PR is mainly for IndexOrDocValuesQuery now. 

As discussed earlier, the reason why IndexOrDocValuesQuery slow down is that a large amount of data will be read during caching action, while only a small amount of data will be read from doc values when not caching. I don't find any other type of query that reads much more data for caching than it really needs. 

Looking forward to more discussions.

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
```diff
@@ -732,8 +741,39 @@ public ScorerSupplier scorerSupplier(LeafReaderContext context) throws IOExcepti
...
+            putIfAbsent(in.getQuery(), DocIdSet.EMPTY, cacheHelper);
+            return null;
+          }
+
+          final long cost = supplier.cost();
+          return new ScorerSupplier() {
+            @Override
+            public Scorer get(long leadCost) throws IOException {
+              // skip cache operation which would slow query down too much
+              if (cost > skipCacheCost && cost > leadCost * skipCacheFactor
```
jgq2008303393: We have tested different scenarios to observe the query latency with/without cacheing in an online metric ES cluster. The result is as follows:

| query      | latencyWithoutCaching  | latencyWithCaching | leadCost | rangeQueryCost  | skipCacheFactor |
| ---------- | :-----------:  | :-----------: | :-----------:  | :-----------: | :-----------:  |
| ip:xxx AND time:[t-1h, t] | 10ms | 36ms(+260%) | 20528 | 878979 | 42 |
| ip:xxx AND time:[t-4h, t] | 10ms | 100ms(+900%) | 20528 | 4365870 | 212 |
| ip:xxx AND time:[t-8h, t] | 11ms | 200ms(+1700%) | 20528 | 8724483 | 425 |
| ip:xxx AND time:[t-12h, t] | 12ms | 300ms(+2400%) | 20528 | 13083096 | 637 |
| ip:xxx AND time:[t-24h, t] | 16ms | 500ms(+3000%) | 20528 | 26158936 | 1274 |
| ip:xxx AND time:[t-48h, t] | 30ms | 1200ms(3900%) | 20528 | 52310616 | 2548 |

As the table show, query latency without caching is low and related with the final result set, and query latency with caching is much high and mainly related with _rangeQueryCost_. 

We set the default value of _skipCacheFactor_ to 250, which make the query slower by no more than 10 times.

In addition to _skipCacheFactor_ which is similar to _maxCostFactor_ in LUCENE-8027, we have added a new parameter _skipCacheCost_. The mainly reasons are:
- control the time used for caching as the caching time is related to the cost of range query.
- skip caching too large range queries which will consume too much memory and evict cache frequently.

How do you think? Looking forward to your ideas. @jpountz

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
```diff
@@ -732,8 +741,39 @@ public ScorerSupplier scorerSupplier(LeafReaderContext context) throws IOExcepti
...
+            putIfAbsent(in.getQuery(), DocIdSet.EMPTY, cacheHelper);
+            return null;
+          }
+
+          final long cost = supplier.cost();
+          return new ScorerSupplier() {
+            @Override
+            public Scorer get(long leadCost) throws IOException {
+              // skip cache operation which would slow query down too much
+              if ((cost > skipCacheCost || cost > leadCost * skipCacheFactor)
```
jgq2008303393: We have tested different scenarios to observe the query latency with/without cacheing in an online ES cluster. Here is the result:

| queryPattern      | latencyWithoutCaching  | latencyWithCaching | leadCost | rangeQueryCost  | skipCacheFactor |
| ---------- | :-----------:  | :-----------: | :-----------:  | :-----------: | :-----------:  |
| ip:xxx AND time:[t-1h, t] | 10ms | 36ms(+260%) | 20528 | 878979 | 42 |
| ip:xxx AND time:[t-4h, t] | 10ms | 100ms(+900%) | 20528 | 4365870 | 212 |
| ip:xxx AND time:[t-8h, t] | 11ms | 200ms(+1700%) | 20528 | 8724483 | 425 |
| ip:xxx AND time:[t-12h, t] | 12ms | 300ms(+2400%) | 20528 | 13083096 | 637 |
| ip:xxx AND time:[t-24h, t] | 16ms | 500ms(+3000%) | 20528 | 26158936 | 1274 |
| ip:xxx AND time:[t-48h, t] | 30ms | 1200ms(3900%) | 20528 | 52310616 | 2548 |

As the table shows, query latency without caching is low and it's related with the final result set. Query latency with caching is much high and it's mainly related with _rangeQueryCost_. According to the above test, we set the default value of _skipCacheFactor_ to 250, which make the query slower by no more than 10 times.

In addition to _skipCacheFactor_ which is similar to _maxCostFactor_ in LUCENE-8027, we add a new parameter _skipCacheCost_. The mainly reasons are:
- control the time used for caching as the caching time is related to the cost of range query.
- skip caching too large range queries which will consume too much memory and evict cache frequently.

How do you think? Looking forward to your ideas. @jpountz

jpountz: > control the time used for caching as the caching time is related to the cost of range query

Understood, but if the query has a 250x higher cost already, then it's probably fine right?

> skip caching too large range queries which will consume too much memory and evict cache entries frequently

Maybe a more intuitive way to do that would be to call this parameter `maxRamBytesUsedPerEntry` and skip caching if `min(maxDoc/Byte.SIZE /* assuming bitset */, cost * Short.BYTES /* assuming short[] */) > maxRamBytesUsedPerEntry`?

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
```diff
@@ -732,8 +741,39 @@ public ScorerSupplier scorerSupplier(LeafReaderContext context) throws IOExcepti
...
+                  && in.getQuery() instanceof IndexOrDocValuesQuery) {
```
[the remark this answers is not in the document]
jpountz: >  I don't find any other type of query that reads much more data for caching than it really needs.

Actually I think it applies to other queries as well, maybe just not as much an IndexOrDocValuesQuery. For instance term queries skip blocks when none of the 128 documents in the block match, so your 250x factor would probably help there as well. (We don't cache term queries directly, but we do cache combinations of term queries through BooleanQuery.)

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
jgq2008303393: Yes. With a 250x higher cost, it’s good enough now without _skipCacheCost_.

Also we find that _skipCacheCost_ could be done through the following code. Although we think that the allowed cache entry size is a little large, we are not going to modify it now. If we find that the demand is very strong, we will open a new PR.
https://github.com/apache/lucene-solr/blob/61a00822e78a02f14ac060dabf66c2b5053c12d7/lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java#L679-L691

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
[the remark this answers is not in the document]
jgq2008303393: @jpountz Thanks for reminding. I found the following ES query has similar problem. 
```java
GET host_monitor/_search
{
  "size": 100,
  "query": {
    "bool": {
      "filter": [
        {
          // term query cost is 1.8w
          "term": {
            "timestamp": 1570291200
          }
        },
        {
          // terms query cost is 377.9w
          "terms": {
            "cpu_name": ["cpu0", "cpu1", "cpu2", "cpu3"]
          }
        }
      ]
    }
  }
}
```
The test result of this query is as follows:

Query | 1th | 2th | 3th | 4th | 5th | 6th
-- | -- | -- | -- | -- | -- | --
Before Optimization | 24ms | 22ms | 22ms | 307ms | 16ms | 16ms
After Optimization| 24ms | 23ms | 25ms | 23ms | 23ms | 24ms

| queryPattern      | latencyWithoutCaching  | latencyWithCaching | leadCost | rangeQueryCost  | skipCacheFactor |
| ---------- | :-----------:  | :-----------: | :-----------:  | :-----------: | :-----------:  |
| cpu_name with 2 vlaues | 22ms | 165ms(+650%) | 18432 | 1911910 | 103 |
| cpu_name with 4 vlaues | 24ms | 307ms(+1180%) | 18432 | 3778721| 205 |
| cpu_name with 8 vlaues | 27ms | 616ms(+2270%) | 18432 | 7590457 | 411 |


According to above test result, I will apply this PR to all query types in next patch.

jgq2008303393:
@jpountz Please help to review the latest patch. 

By the way, once the [asynchronous caching](https://github.com/apache/lucene-solr/pull/916) is submitted, I will open a new PR to support different skipCacheFactor for synchronous and asynchronous caching.

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
```diff
@@ -114,12 +115,16 @@
...
+   * Also, clauses whose cost is {@code skipCacheFactor} times more than the cost of the top-level query
+   * will not be cached in order to not slow down queries too much.
    */
   public LRUQueryCache(int maxSize, long maxRamBytesUsed,
-      Predicate<LeafReaderContext> leavesToCache) {
+                       Predicate<LeafReaderContext> leavesToCache, long skipCacheFactor) {
     this.maxSize = maxSize;
     this.maxRamBytesUsed = maxRamBytesUsed;
     this.leavesToCache = leavesToCache;
+    this.skipCacheFactor = skipCacheFactor;
```
jpountz: let's check it's greater than 0? I also wonder whether we should make it a float?

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
```diff
@@ -732,8 +737,38 @@ public ScorerSupplier scorerSupplier(LeafReaderContext context) throws IOExcepti
...
+            putIfAbsent(in.getQuery(), DocIdSet.EMPTY, cacheHelper);
+            return null;
+          }
+
+          final long cost = supplier.cost();
+          return new ScorerSupplier() {
+            @Override
+            public Scorer get(long leadCost) throws IOException {
+              // skip cache operation which would slow query down too much
+              if (cost > leadCost * skipCacheFactor) {
```
jpountz: Reading your tests made me wonder that the multiplication would overflow here when `Long.MAX_VALUE` is used for `skipCacheFactor`.

jgq2008303393: Yes, the multiplication may overflow. Considering _leadCost_ may be zero, I will use `cost / skipCacheFactor > leadCost` to replace the multiplication.

BTW, I found out that why my tests success is that _leadCost_ was always 1. I will strengthen my tests a little.

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
jgq2008303393: Considering the multiplication overflow problem, I will update the code here : )

wuyunfeng:
@jpountz  If just only cache the top level query, whether can resolve this problem? Today, such as `BooleanQuery->BooleanWeight` invoke the  `IndexSearcher#createWeight(Query query, ScoreMode scoreMode, float boost)` ,  then `IndexSearcher` cache its sub clause. if just `cache` the `BooleanWeight` maybe avoid some time-consuming cache behavior?

whether it is difficult or perplexed for user to decide `skipCacheFactor ` should be?

jgq2008303393:
@jpountz Please help to merge this PR : )

jgq2008303393:
@wuyunfeng LRUQueryCache will cache compound queries a bit earlier as it explains here:
https://github.com/apache/lucene-solr/blob/61a00822e78a02f14ac060dabf66c2b5053c12d7/lucene/core/src/java/org/apache/lucene/search/UsageTrackingQueryCachingPolicy.java#L113-L137

We carefully choose a default value for skipCacheFactor so that it can fit most scenarios. And only expert users need to adjust it. When LRUQueryCache skips caching, the query latency becomes very smooth. We think that it does not have much side effect for most users :）

wuyunfeng:
@jgq2008303393  I guess you don's get my point, compound query cache sub-clause just because the compound query call the `IndexSearcher#createWeight`, then  the default `createWeight` cache the sub-clause. 

```
  public Weight createWeight(Query query, ScoreMode scoreMode, float boost) throws IOException {
    final QueryCache queryCache = this.queryCache;
    Weight weight = query.createWeight(this, scoreMode, boost);
    if (scoreMode.needsScores() == false && queryCache != null) {
      weight = queryCache.doCache(weight, queryCachingPolicy);
    }
    return weight;
  }
```
 If the  compound query  just `createWeight` and do not cache the sub-clause, this problem maybe not happend?

jgq2008303393:
@wuyunfeng Whether to cache a composite clause or a single clause needs to be determined according to the scenario. For queries are like this: host_ip:xxx AND timestamp:[start, end], if host_ip clause and timestamp clause always have the same parameter value, only the composite clause should be cached. However if host_ip clause has different parameter values in each query, which is the real situation, it makes more sense to only cache timestamp clause. That's why Lucene cache clause according to frequency.

jgq2008303393:
@jpountz Please help to merge : )

jgq2008303393:
@jpountz Please help to merge : )

jgq2008303393:
@jpountz I have resolved the CHANGES.txt conflict, please help to merge : )

lucene/core/src/test/org/apache/lucene/search/TestLRUQueryCache.java
```diff
@@ -56,6 +59,7 @@
 import org.apache.lucene.index.Term;
 import org.apache.lucene.search.BooleanClause.Occur;
 import org.apache.lucene.store.Directory;
+import org.apache.lucene.util.BytesRef;
```
jpountz: Have you seen the `ant precommit` failure? It complains that this import is unused.

jgq2008303393: @jpountz I am very sorry. I have fixed the above problem and now the PR has passed ant precommit.

jgq2008303393:
@jpountz Please take a look. Thanks very much ：）

jgq2008303393:
@jpountz I have resolved the CHANGES.txt conflict again. And I'm sorry for forcing merge. 

Please help to merge : )

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
```diff
@@ -114,12 +115,20 @@
...
+   * Also, clauses whose cost is {@code skipCacheFactor} times more than the cost of the top-level query
+   * will not be cached in order to not slow down queries too much.
    */
   public LRUQueryCache(int maxSize, long maxRamBytesUsed,
-      Predicate<LeafReaderContext> leavesToCache) {
+                       Predicate<LeafReaderContext> leavesToCache, float skipCacheFactor) {
     this.maxSize = maxSize;
     this.maxRamBytesUsed = maxRamBytesUsed;
     this.leavesToCache = leavesToCache;
+    if (skipCacheFactor < 1) {
```
jpountz:
```suggestion
    if (skipCacheFactor >= 1 == false) { // NaN >= 1 evaluates true
```

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
```diff
@@ -114,12 +115,20 @@
...
+   * Also, clauses whose cost is {@code skipCacheFactor} times more than the cost of the top-level query
+   * will not be cached in order to not slow down queries too much.
    */
   public LRUQueryCache(int maxSize, long maxRamBytesUsed,
-      Predicate<LeafReaderContext> leavesToCache) {
+                       Predicate<LeafReaderContext> leavesToCache, float skipCacheFactor) {
     this.maxSize = maxSize;
     this.maxRamBytesUsed = maxRamBytesUsed;
     this.leavesToCache = leavesToCache;
+    if (skipCacheFactor >= 1 == false) { // NaN >= 1 evaluates true
```
jpountz:
```suggestion
    if (skipCacheFactor >= 1 == false) { // NaN >= 1 evaluates false
```

I got it backward, sorry about that.

PR #950, opened 2019-10-14
LUCENE-8920: Disable direct addressing of arcs.

jpountz:
@msokolov Would you mind having a look? Reverts weren't clean so I ended up manually undoing the change by removing the special flag that you introduced for "array with gaps". I'd appreciate if you could take a look to make sure I didn't get anything wrong.


msokolov:
It looks like this will effectively disable the array-with-arcs optimization, yes, and remove any related code.

jpountz:
Thanks @msokolov.

PR #953, opened 2019-10-15
LUCENE-9006: WDGF catenateAll should come before parts

dsmiley:
See nocommits and parent issue.

The BaseTokenStreamTestCase edit was a trivial thing to ensure the primary assertTokenStreamContents gets called first.  This helps during dev when there's a failure, it's more relevant to see a failure there.

### DIFF
--- a/lucene/analysis/common/src/java/org/apache/lucene/analysis/miscellaneous/WordDelimiterGraphFilter.java
+++ b/lucene/analysis/common/src/java/org/apache/lucene/analysis/miscellaneous/WordDelimiterGraphFilter.java
@@ -447,7 +447,7 @@ public final class WordDelimiterGraphFilter extends TokenFilter {
   private class PositionSorter extends InPlaceMergeSorter {
     @Override
     protected int compare(int i, int j) {
-      // sort by smaller start position
+      // smaller start position
       int iPosStart = bufferedParts[4*i];
       int jPosStart = bufferedParts[4*j];
       int cmp = Integer.compare(iPosStart, jPosStart);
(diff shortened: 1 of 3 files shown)


dsmiley:
Thanks for the approval.

Is there a simple test somewhere that illustrates that the graph is incorrect?  Or if you were to write one that does, maybe `org.apache.lucene.analysis.BaseTokenStreamTestCase#getGraphStrings(org.apache.lucene.analysis.TokenStream)` would show the problem for "8-input"?    I edited testCatenateAllEmittedBeforeParts with this code:
```Set<String> graphStrings = getGraphStrings(a, "8-input");
    System.out.println(graphStrings.toString());```
which produced: [input, 8-input, 8input]
which looks correct; no?

jimczi:
> which produced: [input, 8-input, 8input]
which looks correct; no?

It depends what you consider correct ;). With GENERATE_NUMBER_PARTS the graph looks like:
````
[8 input, 8-input, 8input]
  8 other
  8-other
  8other
````
so `other` is one position after `8other` and `8-other` which both spans two positions.
When GENERATE_NUMBER_PARTS is not set `other` appears at the same position than `8other` and `8-other` which now span only one position. With Alan's proposal we'd be able to create the "correct" graph by delaying the removal of `8` in the stream in order to ensure that all nodes are reachable. Not saying that it's the best solution but it would be more consistent since it feels weird to me that `8other` and `other` are considered at the same position when numbers are removed.

dsmiley:
I guess we disagree on what is "correct" and thus correctness is debatable.  If GENERATE_NUMBER_PARTS=0, it seems reasonable that "8other" and "other" occupy the same position.  The numbers become ignored noise like symbols.  You said  ... "in order to ensure that all nodes are reachable."  Reading that implies that we have to do something to prevent an unreachable node... but at present it's _not_ apparent to me that we have such a problem to fix.  *If* the effect of not generating a part should yield a position gap, and if the first part is non-generated, then I can see there would be a problem.

jimczi:
I agree that's debatable and the current output after your fix can be considered as "correct" too. I think I am looking for cases that would emphasize the need for a solution like https://issues.apache.org/jira/browse/LUCENE-8717 but this one is different.

PR #957, opened 2019-10-16
SOLR-13822: Isolated Classloading from packages

noblepaul:
# Description

 load plugins from packages 
main features:

- A new file for packages definition (/packages.json) in ZK
- Public APIs to edit/read the file
- The APIs are registered at /api/cluster/package
- Classes can be loaded from the package classloader using the <package_name>:<fully.qualified.ClassName> syntax

# Checklist


solr/core/src/java/org/apache/solr/core/PluginBag.java
```diff
@@ -138,14 +140,27 @@ static void initInstance(Object inst, PluginInfo info) {
       log.debug("{} : '{}' created with startup=lazy ", meta.getCleanTag(), info.name);
       return new LazyPluginHolder<T>(meta, info, core, core.getResourceLoader(), false);
     } else {
-      T inst = core.createInstance(info.className, (Class<T>) meta.clazz, meta.getCleanTag(), null, core.getResourceLoader());
-      initInstance(inst, info);
-      return new PluginHolder<>(info, inst);
+      if (info.pkgName != null) {
```
dsmiley: minor: please tack on another "else if" from above rather than indenting further needlessly.  Also for consistency it's debatable if getPackagePluginHolder needs its own method vs. inlining.   There is no getRuntimeLibPluginHolder after all.

solr/core/src/java/org/apache/solr/update/processor/UpdateRequestProcessorChain.java
```diff
@@ -339,27 +359,20 @@ public UpdateRequestProcessorFactory get() {
       return lazyFactory;
     }
 
-    public class LazyUpdateRequestProcessorFactory extends UpdateRequestProcessorFactory {
-      private final PluginBag.LazyPluginHolder holder;
-      UpdateRequestProcessorFactory delegate;
+    public static class LazyUpdateRequestProcessorFactory extends UpdateRequestProcessorFactory {
+      private final PluginBag.PluginHolder<UpdateRequestProcessorFactory> holder;
```
dsmiley: I think it might be nicer if we declared this as simply Supplier<UpdateRequestProcessorFactory> and then made that PluginHolder thing implement a Supplier<T>?

solr/core/src/java/org/apache/solr/core/SolrCore.java
```diff
@@ -274,6 +282,15 @@ public SolrResourceLoader getResourceLoader() {
     return resourceLoader;
   }
 
+  public SolrResourceLoader getResourceLoader(String pkg) {
```
dsmiley: String is generic and so the method signature is non-obvious.  Rename to getResourceLoaderForPackage  ?

solr/core/src/java/org/apache/solr/core/SolrResourceLoader.java
```diff
@@ -81,11 +82,11 @@
...
 /**
  * @since solr 1.3
- */ 
-public class SolrResourceLoader implements ResourceLoader,Closeable
-{
+ */
+public class SolrResourceLoader implements ResourceLoader, Closeable {
   private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
 
+  private String name = "";
```
dsmiley: Please order this new instance field with the other instance fields, and not with the statics.  Also, shouldn't it be final?  And I wonder if an empty string default is something to improve... it seems to me we could default it to something useful or null.  I prefer to avoid empty strings.

solr/core/src/java/org/apache/solr/core/PluginInfo.java
```diff
@@ -35,27 +41,46 @@
...
-    this.className = attrs.get(CLASS_NAME);
+    Pair<String, String> parsed = parseClassName(attrs.get(CLASS_NAME));
+    this.className = parsed.second();
+    this.pkgName = parsed.first();
     this.initArgs = initArgs;
     attributes = unmodifiableMap(attrs);
     this.children = children == null ? Collections.<PluginInfo>emptyList(): unmodifiableList(children);
     isFromSolrConfig = false;
   }
+  static Pair<String,String > parseClassName(String name) {
```
dsmiley: A comment on a "parse" method is helpful.  Also a use of "Pair" needs docs on what the components mean.  That could be done by renaming to "parsePackageAndClass()"

solr/core/src/java/org/apache/solr/handler/SolrConfigHandler.java
```diff
@@ -245,10 +246,23 @@ private void handleGET() {
             if (componentName != null) {
               Map map = (Map) val.get(parts.get(1));
               if (map != null) {
-                val.put(parts.get(1), makeMap(componentName, map.get(componentName)));
+                Object o = map.get(componentName);
```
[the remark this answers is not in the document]
noblepaul: This was just format change introduced by IDE

solr/core/src/java/org/apache/solr/core/PluginBag.java
noblepaul: inlined

solr/core/src/java/org/apache/solr/update/processor/UpdateRequestProcessorChain.java
noblepaul: These classes existed before . I didn't want to change too much. We can probably have . a refactoring later

solr/solrj/ivy.xml
```diff
@@ -52,6 +52,8 @@
     <dependency org="org.eclipse.jetty" name="jetty-alpn-java-client" rev="${/org.eclipse.jetty/jetty-alpn-java-client}" conf="compile"/>
     <dependency org="org.eclipse.jetty" name="jetty-alpn-client" rev="${/org.eclipse.jetty/jetty-alpn-client}" conf="compile"/>
 
+    <dependency org="com.fasterxml.jackson.core" name="jackson-annotations" rev="${/com.fasterxml.jackson.core/jackson-annotations}" conf="compile"/>
```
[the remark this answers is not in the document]
noblepaul: This was added in another ticket. Let's have this discussion there

PR #959, opened 2019-10-17
SOLR-13677 fix & cleanup

sigram:
# Description

This PR is based on Noble's patch plus cleanups & refactoring identified in the Jira review.

# Solution

# Tests
`MetricsHandlerTest.testMetricsUnload()` is probably insufficient to test all corner cases. We still need to add tests for hierarchical metric tags and back-compat of the API.

# Checklist


solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -431,12 +432,13 @@ public void register() {
       cache.setState(SolrCache.State.LIVE);
       infoRegistry.put(cache.name(), cache);
     }
-    metricManager = core.getCoreContainer().getMetricManager();
-    registryName = core.getCoreMetricManager().getRegistryName();
+    this.solrMetricsContext = core.getSolrMetricsContext().getChildContext(this);
```
sigram: `SolrIndexSearcher` is not a `SolrMetricProducer` but it registers some gauges so I think it needs a child context - especially since its life-cycle is different from that of SolrCore.

solr/core/src/java/org/apache/solr/servlet/SolrDispatchFilter.java
```diff
@@ -108,7 +108,7 @@
   
   private boolean isV2Enabled = !"true".equals(System.getProperty("disable.v2.api", "false"));
 
-  private final String metricTag = Integer.toHexString(hashCode());
+  private final String metricTag = getClass().getSimpleName() + "@" + Integer.toHexString(hashCode());
```
sigram: Should use the new method from `SolrMetricProducer`.

solr/core/src/java/org/apache/solr/update/SolrIndexWriter.java
```diff
@@ -135,8 +135,7 @@ private SolrIndexWriter(SolrCore core, String name, String path, Directory direc
     infoStream = getConfig().getInfoStream();
     this.directory = directory;
     numOpens.incrementAndGet();
-    metricManager = core.getCoreContainer().getMetricManager();
-    registryName = core.getCoreMetricManager().getRegistryName();
+    solrMetricsContext = core.getSolrMetricsContext().getChildContext(this);
```
sigram: Similar situation to `SolrIndexSearcher`, the life-cycle here is different from SolrCore.

solr/core/src/java/org/apache/solr/update/SolrIndexWriter.java
```diff
@@ -345,6 +344,9 @@ private void cleanup() throws IOException {
       if (directoryFactory != null) {
         directoryFactory.release(directory);
       }
+      if (solrMetricsContext != null) {
```
sigram: Not a `SolrMetricProducer` so we need to call this explicitly.

solr/core/src/test/org/apache/solr/handler/admin/MetricsHandlerTest.java
```diff
@@ -325,8 +335,123 @@ public void testKeyMetrics() throws Exception {
...
         MetricsHandler.KEY_PARAM, "solr.jetty:unknown:baz"), resp);
     values = resp.getValues();
-    metrics = (NamedList)values.get("metrics");
+    metrics = (NamedList) values.get("metrics");
     assertEquals(0, metrics.size());
     assertNotNull(values.findRecursive("errors", "solr.jetty:unknown:baz"));
   }
+
+  @Test
+  public void testMetricsUnload() throws Exception {
```
sigram: We need another test that tests the following scenario:
* instance 1 registers its gauges
* instance 2 is created and registers its gauges
* instance 1 is closed and should unregister its gauges

together with dependent child instances

solr/core/src/java/org/apache/solr/security/AuthenticationPlugin.java
```diff
@@ -184,7 +181,7 @@ public Category getCategory() {
 
   @Override
   public MetricRegistry getMetricRegistry() {
-    return registry;
+    return metrics == null ? null : metrics.getMetricRegistry();
```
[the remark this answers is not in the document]
sigram: Good point, we can.

sigram:
Regarding scope - in many places the SolrMetricsContext instance can be created once in the parent and then used for many components that differ only in their scope. Passing one common instance + component-specific scope seemed more efficient than creating new component-specific context for each component just because the scope was different ... but I'm not strongly attached to this change if you have strong arguments otherwise.

solr/core/src/java/org/apache/solr/security/AuthenticationPlugin.java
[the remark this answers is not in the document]
sigram: Actually, we can't because we already have a default with the same signature in `SolrInfoBean`, but I can improve the implementation there to avoid this repetition.

solr/core/src/java/org/apache/solr/metrics/SolrMetricProducer.java
```diff
@@ -19,17 +19,72 @@
...
+
+  @Override
+  default void close() throws Exception {
+    SolrMetricsContext context = getSolrMetricsContext();
+    if (context == null) {
+      return;
+    } else {
+      context.unregister();
+    }
+    // ??? (ab) no idea what this was supposed to avoid
```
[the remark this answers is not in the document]
sigram: See my reply below. Still, regarding this particular snippet, I don't think it's needed anymore,, maybe it's a leftover from an earlier refactoring?

sigram:
Committed.

PR #964, opened 2019-10-22
LUCENE-9023: GlobalOrdinalsWithScore should not compute occurrences when the provided min is 1

jimczi:
This is a continuation of https://issues.apache.org/jira/browse/LUCENE-9022

Today the GlobalOrdinalsWithScore collector and query checks the number of matching docs per parent if the provided min is greater than 0. However we should also not compute the occurrences of children when min is equals to 1 since this is the minimum requirement for a document to match.


lucene/join/src/java/org/apache/lucene/search/join/GlobalOrdinalsWithScoreCollector.java
```diff
@@ -47,7 +47,7 @@
       throw new IllegalStateException("Can't collect more than [" + Integer.MAX_VALUE + "] ids");
     }
     this.field = field;
-    this.doMinMax = !(min <= 0 && max == Integer.MAX_VALUE);
+    this.doMinMax = min > 1 && max < Integer.MAX_VALUE;
```
jpountz: Should it be an or?
```suggestion
    this.doMinMax = min > 1 || max < Integer.MAX_VALUE;
```

lucene/CHANGES.txt
```diff
@@ -52,6 +52,8 @@ Improvements
 
 * LUCENE-8937: Avoid agressive stemming on numbers in the FrenchMinimalStemmer.
   (Adrien Gallou via Tomoko Uchida)
+
+* LUCENE-9023: GlobalOrdinalsWithScore should not compute occurrences when the provided min is 1
```
jpountz:
```suggestion
* LUCENE-9023: GlobalOrdinalsWithScore should not compute occurrences when the
  provided min is 1. (Jim Ferenczi)
```

PR #966, opened 2019-10-23
LUCENE-9024 Optimize IntroSelector to use median of medians

pcsanwald:
# Description

Today, `IntroSelector.slowSelect` falls back to a heap sort; This PR moves that implementation to use a median of medians approach, suggested by a TODO in the code.
 
# Solution

Implemented a median of medians algorithm.

# Tests

I didn't add any new tests for this as I believe the current test cases, `TestIntroSelector`, should cover the basic functionality. However, one thing I'd like input on: Today, `select(int from, int to, int k)`, the `to` parameter is _exclusive_. Since it's also a specified parameter in the test, the test never checks the case where the full array should be used. `doTestSelect` has a call to `Arrays.sort(expected, from, to);`, but `to` will never be the last index in the array because the size of the array is allocated to `from + to + random().nextInt(5)`, so the actual size of the array will be a minimum of `from + to`.

I'd be happy to address this problem in this PR, or in a subsequent PR if that's preferred. 

# Checklist

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/util/IntroSelector.java
+++ b/lucene/core/src/java/org/apache/lucene/util/IntroSelector.java
@@ -36,29 +36,30 @@ public abstract class IntroSelector extends Selector {
     return medianOfMediansSelect(from, to-1, k);
   }
 
-  int medianOfMediansSelect(int from, int to, int k) {
+  int medianOfMediansSelect(int left, int right, int k) {
     int pivotIndex;
     do {
-      if (from == to) {
-        return from;
+      // Defensive check, this is also checked in the calling
+      // method. Including here so this method can be used
+      // as a self contained quickSelect implementation.
+      if (left == right) {
+        return left;
       }
-      pivotIndex = pivot(from, to);
-      setPivot(pivotIndex);
-      pivotIndex = partition(from, to, k, pivotIndex);
-      setPivot(pivotIndex);
+      pivotIndex = pivot(left, right);
+      pivotIndex = partition(left, right, k, pivotIndex);
       if (k == pivotIndex) {
         return k;
       } else if (k < pivotIndex) {
-        to = pivotIndex-1;
+        right = pivotIndex-1;
       } else {
-        from = pivotIndex+1;
+        left = pivotIndex+1;
       }
-    } while (from != to);
-    setPivot(pivotIndex);
+    } while (left != right);
     return pivotIndex;
   }
 
-  private int partition(int left, int right, int n, int pivotIndex) {
+  private int partition(int left, int right, int k, int pivotIndex) {
+    setPivot(pivotIndex);
     swap(pivotIndex, right);
     int storeIndex = left;
     for (int i = left; i < right; i++) {
@@ -75,10 +76,10 @@ public abstract class IntroSelector extends Selector {
       }
     }
     swap(right, storeIndexEq);
-    if (n < storeIndex) {
+    if (k < storeIndex) {
       return storeIndex;
-    } else if (n <= storeIndexEq) {
-      return n;
+    } else if (k <= storeIndexEq) {
+      return k;
     }
     return storeIndexEq;
   }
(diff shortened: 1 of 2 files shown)


lucene/core/src/java/org/apache/lucene/util/IntroSelector.java
```diff
@@ -33,26 +32,87 @@ public final void select(int from, int to, int k) {
...
-  // heap sort
-  // TODO: use median of median instead to have linear worst-case rather than
-  // n*log(n)
-  void slowSelect(int from, int to, int k) {
-    new Sorter() {
+  int slowSelect(int from, int to, int k) {
+    return medianOfMediansSelect(from, to-1, k);
+  }
+
+  int medianOfMediansSelect(int from, int to, int k) {
```
jpountz: It took me a while to figure out that `to` is inclusive here while it is exclusive elsewhere in this file (e.g. in `select` and `quickselecet`). Maybe rename it to `toInclusive` or something along those lines to make it clear? Or maybe call them `left` and `right` like you did in some other methods?

lucene/core/src/java/org/apache/lucene/util/IntroSelector.java
```diff
@@ -33,26 +32,87 @@ public final void select(int from, int to, int k) {
...
+    int pivotIndex;
+    do {
+      if (from == to) {
```
jpountz: I believe this condition can never happen when `medianOfMediansSelect` is called from `quickSelect` since it first checks whether `to-from==1` before delegating to `medianOfMediansSelect`? It is probably a good idea to be defensive, but maybe leave a comment about it?

lucene/core/src/java/org/apache/lucene/util/IntroSelector.java
```diff
@@ -33,26 +32,87 @@ public final void select(int from, int to, int k) {
...
+      setPivot(pivotIndex);
+      if (k == pivotIndex) {
+        return k;
+      } else if (k < pivotIndex) {
+        to = pivotIndex-1;
+      } else {
+        from = pivotIndex+1;
+      }
+    } while (from != to);
+    setPivot(pivotIndex);
```
jpountz: Setting the pivot here doesn't seem necessary?

lucene/core/src/java/org/apache/lucene/util/IntroSelector.java
```diff
@@ -33,26 +32,87 @@ public final void select(int from, int to, int k) {
...
+    return pivotIndex;
+  }
 
-      @Override
-      protected void swap(int i, int j) {
-        IntroSelector.this.swap(i, j);
+  private int partition(int left, int right, int n, int pivotIndex) {
```
jpountz: can we rename `n` to `k`, I think it's confusing that the same thing gets called different ways?

lucene/core/src/java/org/apache/lucene/util/IntroSelector.java
```diff
@@ -33,26 +32,87 @@ public final void select(int from, int to, int k) {
...
+      swap(median5, left + ((i-left)/5));
+    }
+    int mid = ((right - left) / 10) + left + 1;
+    int to = left + ((right - left)/5);
+    return medianOfMediansSelect(left, to, mid);
+  }
 
-      public void sort(int from, int to) {
-        heapSort(from, to);
+  private int partition5(int left, int right) {
```
jpountz: Is this a bubble sort? If so it might make things clearer to say it in a comment, and maybe also be explicit that it doesn't hurt the runtime complexity since we only apply it to small arrays?

pcsanwald:
@jpountz thanks for the detailed review! I will be addressing your comments today. One comment:

> it should be `to+random.nextInt(5)`, is it what you are proposing?

That's part of it. but even with a right parameter of `to+0`, the code is never exercised over the whole array, as it's exclusive. So, I am not 100% sure I am right, but, it seems to me that there is no way to quickSelect on an entire array, as `to = arr.len` is still exclusive. I guess this has not been a problem in practice, but, I wanted to raise it.

PR #970, opened 2019-10-23
Enable back TestTlogReplica in 8x

tflobbe:
This is a cherry-pick of https://github.com/apache/lucene-solr/commit/98cdac82a18d4d6594804c30e1687bfedbd68838 (note this PR is against branch_8x), however `testAddDocs` fails, and seems to be related to recent metrics changes. @sigram, mind taking a look at the last commit in this PR?
The reason the test fails is that the stats for `updateHandler` come back empty, and I was able to track that to the fact that `getSolrMetricsContext()` returns `null` (from the superclass)


tflobbe:
The change in https://github.com/apache/lucene-solr/pull/970/commits/e8b28895be0177df62f15b85c5d0795443032e66 was committed separately. I'll go ahead and cherry-pick SOLR-13860 commit without a PR

PR #976, opened 2019-10-24
SOLR-13749: Implement support for joining across collections with multiple shards

danmfox:
# Description

Implements support for joining across collections with multiple shards.

# Solution

This PR includes two query parsers.  The XCJF query parser queries a remote collection to obtain a set of join keys to use as a filter against the local collection.  It retrieves the set of join keys from the remote collection using a streaming expression.

The Hash Range query parser takes a field name and a range, and matches documents whose hash for that field falls within the range.  If the local collection uses the join key field as its routing prefix with the compositeId router, applying this filter allows the XCJF query on each shard to request only the join keys that could match against that shard's documents.

# Tests

There is a test class XCJFQueryTest.

# Checklist

### DIFF
--- a/solr/core/src/java/org/apache/solr/search/join/XCJFQParser.java
+++ b/solr/core/src/java/org/apache/solr/search/join/XCJFQParser.java
@@ -46,10 +46,13 @@ public class XCJFQParser extends QParser {
           QueryParsing.TYPE, QueryParsing.V, ZK_HOST, SOLR_URL, COLLECTION, FROM, TO, ROUTED_BY_JOIN_KEY, TTL));
 
   private final String routerField;
+  private final Set<String> solrUrlWhitelist;
 
-  public XCJFQParser(String qstr, SolrParams localParams, SolrParams params, SolrQueryRequest req, String routerField) {
+  public XCJFQParser(String qstr, SolrParams localParams, SolrParams params, SolrQueryRequest req, String routerField, Set<String> solrUrlWhiteList) {
     super(qstr, localParams, params, req);
     this.routerField = routerField;
+    // If specified in the config, this will limit which solr url's the parser can connect to.
+    this.solrUrlWhitelist = solrUrlWhiteList;
   }
 
   @Override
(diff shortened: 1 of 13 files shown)


solr/core/src/java/org/apache/solr/search/join/XCJFQParser.java
```diff
@@ -46,17 +46,25 @@
...
+    // If specified in the config, this will limit which solr url's the parser can connect to.
+    this.solrUrlWhitelist = solrUrlWhiteList;
   }
 
   @Override
   public Query parse() throws SyntaxError {
     String query = localParams.get(QueryParsing.V);
     String zkHost = localParams.get(ZK_HOST);
     String solrUrl = localParams.get(SOLR_URL);
+    // Test if this is a valid solr url.
```
nsoft: This looks good, but I think we need to make it secure by default. Security flaws have recently become a big pain point in the community, having had 5 CVE's in 2019.... Do you mind if I add another if that fails with advice that a white list is required, when none is configured? For expedience I can make this change and we can punt an "insecure, talk to anyone, mode" to a later ticket if you like.

gus-asf: This looks good, but I think we need to make it secure by default. Security flaws have recently become a big pain point in the community, having had 5 CVE's in 2019.... Do you mind if I add another if that fails with advice that a white list is required, when none is configured? For expedience I can make this change and we can punt an "insecure, talk to anyone, mode" to a later ticket if you like.

kwatters: I'm ok with that.  Thanks gus!

gus-asf:
merged manually

solr/core/src/java/org/apache/solr/search/join/XCJFQuery.java
```diff
@@ -0,0 +1,380 @@
...
+        ClusterState clusterState = searcher.getCore().getCoreContainer().getZkController().getClusterState();
+        CloudDescriptor desc = searcher.getCore().getCoreDescriptor().getCloudDescriptor();
+        Slice slice = clusterState.getCollection(desc.getCollectionName()).getSlicesMap().get(desc.getShardId());
+        DocRouter.Range range = slice.getRange();
+
+        // In CompositeIdRouter, the routing prefix only affects the top 16 bits
+        int min = range.min & 0xffff0000;
+        int max = range.max | 0x0000ffff;
+
+        return String.format(Locale.ROOT, "{!hash_range f=%s l=%d u=%d}", fromField, min, max);
```
zkendall, line 199: How come you went with this hash_range filter on the remote handler instead of adding a _route_parameter on the query itself?

I'm new to Solr code so not sure if _route_ parameter is feasible here; or if you just had other purposes/constraints.

epugh: @zkendall This repo is the old 8.x line of Solr, the latest work is in github.com/apache/solr.    Also, you may get more traction on your question on the solr dev mailing list...

kwatters: @zkendall  The route parameter would route the query to one specific shard.  In this scenario, we want the query to execute on every shard, however, only return the joinkeys from those shards within the given hash range.

itygh:

PR #977, opened 2019-10-25
SOLR-10751: Don't replicate index when the master version is 0

tflobbe:
This also fixes SOLR-11094


tflobbe:
@hossman, @CaoManhDat, we discussed this some time back in https://issues.apache.org/jira/browse/SOLR-10751. I'm beasting tests (TestTlogReplicas/TestPullReplicas/ChaosMonkeyNothingIsSafeWithPullReplicasTest) and I've seen no failures so far

PR #980, opened 2019-10-27
LUCENE-8920: Reduce the memory used by direct addressing of arcs

bruno-roustant:
Reduce the memory by using "arc presence" bits, one bit per arc. So the node only stores the present arcs (no "missing arc" anymore). The offset to a given arc is computed by getting the number of "presence bits" set up to the corresponding bit for the arc (and multiplied by the number of bytes per arc).


lucene/core/src/java/org/apache/lucene/util/fst/FST.java
```diff
@@ -864,6 +1012,68 @@ private long readUnpackedNodeTarget(BytesReader in) throws IOException {
...
+    int bitCount = 0;
+    int lastLong = bitIndex >>> 6; // index / 64
+    for (int i = 0; i < lastLong; i++) {
+      bitCount += Long.bitCount(bits[i]);
+    }
+    int indexInLastLong = bitIndex & 63; // index % 64
+    if (indexInLastLong != 0) {
+      long mask = 0xFFFFFFFFFFFFFFFFL >>> (Long.SIZE - indexInLastLong);
+      bitCount += Long.bitCount(bits[lastLong] & mask);
+    }
```
jpountz: we should be able to remove the conditional? e.g. something like this:

```
long mask = (1L << bitIndex) - 1L; // shifts are mode 64
bitCount += Long.bitCount(bits[lastLong] & mask);
```

lucene/core/src/java/org/apache/lucene/util/fst/FST.java
```diff
@@ -864,6 +1012,68 @@ private long readUnpackedNodeTarget(BytesReader in) throws IOException {
...
+  /**
+   * Reads the presence bits of a direct-addressing node, store them in the provided arc {@link Arc#bitTable()} and returns the number of presence bytes.
+   */
+  private int readPresenceBytes(Arc<T> arc, BytesReader in) throws IOException {
+    int numPresenceBytes = getNumPresenceBytes(arc.numArcs());
+    long[] presenceBits = new long[numPresenceBytes + 7 >>> 3];
+    int longIndex = 0;
+    int byteIndex = 0;
+    for (int i = 0; i < numPresenceBytes; i++) {
+      presenceBits[longIndex] |= (long) (in.readByte() & 0xFF) << (i << 3);
```
jpountz:
```suggestion
      presenceBits[longIndex] |= (in.readByte() & 0xFFL) << (i << 3);
```

jpountz: maybe we should even remove longIndex/byteIndex for clarity and do

```
presenceBits[i >>>3] |= (in.readByte() & 0xFFL) << (i << 3);
```

lucene/core/src/java/org/apache/lucene/util/fst/FST.java
```diff
@@ -864,6 +1012,68 @@ private long readUnpackedNodeTarget(BytesReader in) throws IOException {
...
+    }
+    arc.bitTable = presenceBits;
+    assert checkPresenceBytesAreValid(arc);
+    return numPresenceBytes;
+  }
+
+  static boolean checkPresenceBytesAreValid(Arc arc) {
+    assert (arc.bitTable()[0] & 1L) != 0; // First bit must be set.
+    assert (arc.bitTable()[arc.bitTable().length - 1] & (1L << (arc.numArcs() - 1))) != 0; // Last bit must be set.
+    assert countBits(arc.bitTable()) <= arc.numArcs(); // Total bit set must be <= label range.
```
jpountz: the comment says the total bit set must be <= the label range, but it actually tests the total number of bits that are set against the number of arcs, not the range?

lucene/core/src/java/org/apache/lucene/util/fst/FST.java
```diff
@@ -864,6 +1012,68 @@ private long readUnpackedNodeTarget(BytesReader in) throws IOException {
...
+      bitCount += Long.bitCount(l);
+    }
+    return bitCount;
+  }
+
+  /**
+   * Returns whether the bit at given index is set.
+   */
+  static boolean isBitSet(long[] bits, int bitIndex) {
+    return (bits[bitIndex >> 6] & (1L << (bitIndex & 63))) != 0;
```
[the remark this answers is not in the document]
jpountz: I kept the arithmetic shift, but I think it's a bit confusing that your change sometimes uses an arithmetic shift and sometimes a logical shift, can we always use the same type of shift (I have a slight preference for logical shifts, but would be ok to go with arithmetic shifts like FixedBitSet).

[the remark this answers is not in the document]
bruno-roustant: Thanks for the tip about the removal of the mask.

About the arithmetic shift, my idea was to keep the idea of "division by Long.SIZE", but I agree this is not clear. For clarity I finally prefer simply "bits[bitIndex / Long.SIZE]".

lucene/core/src/java/org/apache/lucene/util/fst/FST.java
bruno-roustant: Nice!

lucene/core/src/java/org/apache/lucene/util/fst/FST.java
bruno-roustant: What about
presenceBits[i / Byte.SIZE] |= (in.readByte() & 0xFFL) << (i * Byte.SIZE);

lucene/core/src/java/org/apache/lucene/util/fst/FST.java
bruno-roustant: Actually arc.numArcs() stores the label range for direct addressing. The real number of arcs is the total number of bits set.
I'll change the comment to:
"Total bit set (real num arcs) must be <= label range (stored in arc.numArcs())."

lucene/core/src/java/org/apache/lucene/util/fst/FST.java
```diff
@@ -864,6 +958,64 @@ private long readUnpackedNodeTarget(BytesReader in) throws IOException {
...
+    assert (arc.bitTable()[0] & 1L) != 0; // First bit must be set.
+    assert (arc.bitTable()[arc.bitTable().length - 1] & (1L << (arc.numArcs() - 1))) != 0; // Last bit must be set.
+    assert countBits(arc.bitTable()) <= arc.numArcs(); // Total bit set (real num arcs) must be <= label range (stored in arc.numArcs()).
+    return true;
+  }
+
+  /**
+   * Counts all bits set in the provided longs.
+   */
+  static int countBits(long[] bits) {
```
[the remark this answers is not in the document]
bruno-roustant: Ok

lucene/core/src/java/org/apache/lucene/util/fst/FST.java
```diff
@@ -676,26 +697,38 @@ long addNode(Builder<T> builder, Builder.UnCompiledNode<T> nodeIn) throws IOExce
     */
 
     if (doFixedArray) {
-      final int MAX_HEADER_SIZE = 11; // header(byte) + numArcs(vint) + numBytes(vint)
       assert maxBytesPerArc > 0;
       // 2nd pass just "expands" all arcs to take up a fixed byte size
 
+      // If more than (1 / DIRECT_ARC_LOAD_FACTOR) of the "slots" would be occupied, write an arc
+      // array that may have holes in it so that we can address the arcs directly by label without
```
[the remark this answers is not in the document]
bruno-roustant: Yes

lucene/core/src/java/org/apache/lucene/util/fst/FST.java
```diff
@@ -726,6 +759,57 @@ private void writeArrayPacked(Builder<T> builder, Builder.UnCompiledNode<T> node
...
+
+  private void writePresenceBits(Builder<T> builder, Builder.UnCompiledNode<T> nodeIn, int labelRange, long dest) {
+    long bytePos = dest;
+    byte presenceBits = 1; // The first arc is always present.
+    int presenceIndex = 0;
+    int previousLabel = nodeIn.arcs[0].label;
+    for (int arcIdx = 1; arcIdx < nodeIn.numArcs; arcIdx++) {
+      int label = nodeIn.arcs[arcIdx].label;
+      presenceIndex += label - previousLabel;
+      while (presenceIndex >= 8) {
```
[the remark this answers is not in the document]
bruno-roustant: Ok

lucene/core/src/test/org/apache/lucene/util/TestBitUtil.java
```diff
@@ -0,0 +1,79 @@
...
+              expectedPreviousBitCount, BitUtil.countBitsUpTo(bits, numLong, previousIndex + 1));
+        }
+      }
+    }
+  }
+
+  private long[] buildRandomBits() {
+    long[] bits = new long[random().nextInt(3) + 2];
+    for (int j = 0; j < bits.length; j++) {
+      bits[j] = random().nextLong();
```
jpountz: nit: this way of generating longs makes it very unlikely to have zeros, maybe bias it towards zeros with e.g.

```suggestion
      bits[j] = random().nextBoolean() ? 0L : random().nextLong(); // bias towards zeros which require special logic
```

lucene/core/src/java/org/apache/lucene/util/fst/FST.java
```diff
@@ -249,38 +261,127 @@ public T nextFinalOutput() {
...
+
+      /** Sets the number of longs in the internal long array.
+       * Enlarges it if needed. Always clears the array. */
+      BitTable setNumLongs(int numLongs) {
+        assert numLongs >= 0;
+        this.numLongs = numLongs;
+        if (bits == null || bits.length < numLongs) {
+          bits = new long[ArrayUtil.oversize(numLongs, Long.BYTES)];
+        } else {
+          // Avoid range check in Arrays.fill().
```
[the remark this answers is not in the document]
bruno-roustant: I don't want to pay the range check (3 conditions verified) in Arrays.fill(long[], int, int, long) because they are already verified. I simply remove the comment for clarity.

lucene/core/src/test/org/apache/lucene/util/TestBitUtil.java
bruno-roustant: Good point.

PR #981, opened 2019-10-28
LUCENE-9030: Fix different Solr- and WordnetSynonymParser behaviour

cbuescher:
This fixes an issue where sets of equivalent synonyms in the Wordnet format are
parsed and added to the SynonymMap in a way that leads to the original input
token not being typed as "word" but as SYNONYM instead. Also the original token
doesn't appear first in the token stream output, which is the case for
equivalent solr formatted synonym files.

Currently the WordnetSynonymParser adds all combinations of input/output pairs
of a synset entry into the synonym map, while the SolrSynonymParser excludes
those where input and output term are the same. This change adds the same
behaviour to WordnetSynonymParser and adds tests that show the two formats are
outputting the same token order and types now.

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -100,6 +100,9 @@ Bug Fixes
 
 * LUCENE-9001: Fix race condition in SetOnce. (Przemko Robakowski)
 
+* LUCENE-9030: Fix WordnetSynonymParser behaviour so it behaves similar to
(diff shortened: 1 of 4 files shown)


romseygeek:
Could you add a CHANGES.txt entry?  All looks good other than that, and precommit passes locally for me.

cbuescher:
@romseygeek thanks for the review, I updated the CHANGES.txt.

PR #984, opened 2019-10-30
SOLR-12217: Support shards.preference for individual shard requests

HoustonPutman:
# Description

Use shard preference rules when choosing replicas to send individual shard requests. This includes requests from SolrJ as well as requests sent via streaming expressions.

# Solution
How routing preferences are determined:

For Streaming Expressions, it will use the preferences from the following sources, in order of preference (defaults last):
1. The params given in the stream source
1. The URL params sent in the streaming request
1. The `DEFAULT_SHARD_PREFERENCES` set in cluster properties

For SolrJ, it will use the preferences from the params of the request it is sending.

# Tests

Current tests pass.

**TODO** Add tests for routing functionality

# Checklist


solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseCloudSolrClient.java
```diff
@@ -1109,11 +1125,12 @@ public RouteException(ErrorCode errorCode, NamedList<Throwable> throwables, Map<
         }
       }
 
-      // Shuffle the leaders, if any    (none if !sendToLeaders)
+      // Sort the leader replicas, if any, according to the request preferences    (none if !sendToLeaders)
+      replicaListTransformer.transform(replicas);
```
tflobbe: `theUrlList`?

solr/solrj/src/java/org/apache/solr/client/solrj/io/stream/SignificantTermsStream.java
```diff
@@ -282,6 +281,7 @@ public Explanation toExplanation(StreamFactory factory) throws IOException {
   }
 
   public Tuple read() throws IOException {
+
```
tflobbe: Maybe remove this file from the PR

HoustonPutman:
Want to hear your opinion @tflobbe .

So I changed the logic for [non-update/admin/v2 request routing](https://github.com/apache/lucene-solr/pull/984/files#diff-e188c2b5189b7a912c7c77b50beddf69L1102) to use core URLs instead of collection URLs, unless the request is querying multiple collections. In this way, if there are certain replicas on a host that you want to target you can make that distinction. However that does change the default behavior in a non-trivial way.

The only place where I see this breaks any test is weirdly enough in the [Hadoop auth tests for sending a commit request](https://github.com/apache/lucene-solr/blob/master/solr/core/src/test/org/apache/solr/security/hadoop/TestSolrCloudWithHadoopAuthPlugin.java#L127). It seems that since the commit request is sent directly to a core, instead of the collection url, it is missing an additional count of unauthenticated requests.

HoustonPutman:
I've decided to be conservative and revert that logic. So for non-update requests, SolrJ will send the request to the collection URL, not the replica URL. All tests pass on my end now.

PR #986, opened 2019-10-30
SOLR-13880: Add test that creates/reloads/deletes collections

tflobbe:


risdenk:
https://github.com/apache/solr/pull/1105 is tracking the update of this test.

PR #987, opened 2019-10-30
SOLR-13884: add ConcurrentCreateCollectionTest test

yonik:


dsmiley:
Your test does not use a collection autoscaling policy.  Ideally it would do so, I think, to show that the policy can result in a violation that is avoidable (if it balanced in the first place).  

I was wondering if the org.apache.solr.cloud.api.collections.Assign.AssignStrategyFactory chooses the "LEGACY" strategy or "POLICY" in a default scenario as your test currently would do.   It chooses LEGACY, which I verified in a debugger.  The Policy.Session stuff that supposedly fixes this scenario is, I believe, only used by the collections that have a policy.

yonik:
Yeah, I was trying to start as simple as possible.... and AFAIK, defaults should chose notes with fewer cores.

sigram:
@yonik this has been merged manually, can you close the PR? Thanks!

PR #990, opened 2019-11-01
[SOLR-13885] Typo corrections.

KoenDG:
https://issues.apache.org/jira/browse/SOLR-13885

I noticed this while reading the documenation.

Try pronouncing `it's` as `it is` and you'll head the lines no longer making sense.

All fixes should be correct, though perhaps not complete across the documentation. I only checked the `adoc` files.

### DIFF
--- a/solr/solr-ref-guide/src/spatial-search.adoc
+++ b/solr/solr-ref-guide/src/spatial-search.adoc
@@ -415,7 +415,7 @@ The format, either `ints2D` (default) or `png`.
 
 [TIP]
 ====
-You'll experiment with different `distErrPct` values (probably 0.10 - 0.20) with various input geometries till the default size is what you're looking for. The specific details of how it's computed isn't important. For high-detail grids used in point-plotting (loosely one cell per pixel), set `distErr` to be the number of decimal-degrees of several pixels or so of the map being displayed. Also, you probably don't want to use a geohash-based grid because the cell orientation between grid levels flip-flops between being square and rectangle. Quad is consistent and has more levels, albeit at the expense of a larger index.
+You'll experiment with different `distErrPct` values (probably 0.10 - 0.20) with various input geometries till the default size is what you're looking for. The specific details of how it's computed aren't important. For high-detail grids used in point-plotting (loosely one cell per pixel), set `distErr` to be the number of decimal-degrees of several pixels or so of the map being displayed. Also, you probably don't want to use a geohash-based grid because the cell orientation between grid levels flip-flops between being square and rectangle. Quad is consistent and has more levels, albeit at the expense of a larger index.
 ====
 
 Here's some sample output in JSON (with "..." inserted for brevity):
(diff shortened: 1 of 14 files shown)


solr/solr-ref-guide/src/spatial-search.adoc
```diff
@@ -415,7 +415,7 @@ The format, either `ints2D` (default) or `png`.
...
+You'll experiment with different `distErrPct` values (probably 0.10 - 0.20) with various input geometries till the default size is what you're looking for. The specific details of how it's computed aren't important. For high-detail grids used in point-plotting (loosely one cell per pixel), set `distErr` to be the number of decimal-degrees of several pixels or so of the map being displayed. Also, you probably don't want to use a geohash-based grid because the cell orientation between grid levels flip-flops between being square and rectangle. Quad is consistent and has more levels, albeit at the expense of a larger index.
```
KoenDG: isn't -> aren't

ctargett:
I looked at this PR with an eye to commit it, but since the precommit check failed here I thought I'd just make sure on my local machine. When I download this PR as a patch to my machine, precommit fails for me locally in the same spot the GitHub check failed.

I checked the changes and can't see any reason why it would fail precommit. I checked for errant EOL spaces, etc., and don't see any. Building the Ref Guide with these changes also works fine.

I ran precommit on master after this precommit had failed, and it was fine so it's definitely something in this patch.

Unfortunately, we can't commit this until we can get precommit to pass.

KoenDG:
How very weird.

From the logs, it seems this is the only indication of what went wrong:

```
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by org.apache.ivy.util.url.IvyAuthenticator (file:/home/runner/.ant/lib/ivy-2.4.0.jar) to field java.net.Authenticator.theAuthenticator
WARNING: Please consider reporting this to the maintainers of org.apache.ivy.util.url.IvyAuthenticator
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
```

I don't see anything else in the logs indicating why it is that the precommit failed.

And it's literally only text files. Nothing that should be compiled...

Is there a place where we can read the steps of this precommit check?

KoenDG:
@ctargett And now it's suddenly passing, after rebasing a few times over the past few days.

KoenDG:
I've been updating this a bit because the changes file keeps changing. Makes me wonder: does this go under 8.4 or 9.0 ?

ctargett:
Sorry, I've been a bit busy with other things, but reviewing it again now. When it's committed, it will go under 9.0 and be backported to 8.4.

ctargett:
I made a small local change before committing, to give you attribution on the changes. I used your Github username "KoenDG", but could change it to your full name if you prefer. Thanks for your contribution!

KoenDG:
No worries at all. If anything, sorry for taking your time on this non-functionality-improving PR.

The name can stay like this. I've been using this abbreviation for a few years now.

Thanks for your time, have a good day.

PR #992, opened 2019-11-02
SOLR-13841: removed jackson dependencies from SolrJ and provided a mapping to our annotation

noblepaul:


solr/core/src/java/org/apache/solr/util/SolrPropertyAnnotationInspector.java
```diff
@@ -0,0 +1,83 @@
...
+import com.fasterxml.jackson.databind.AnnotationIntrospector;
+import com.fasterxml.jackson.databind.PropertyName;
+import com.fasterxml.jackson.databind.introspect.Annotated;
+import com.fasterxml.jackson.databind.introspect.AnnotatedField;
+import com.fasterxml.jackson.databind.introspect.AnnotatedMember;
+import com.fasterxml.jackson.databind.introspect.AnnotatedMethod;
+import com.fasterxml.jackson.databind.util.BeanUtil;
+import org.apache.solr.common.annotation.Property;
+import org.apache.solr.search.SolrCache;
+//this class provides a maping between jackson's JsonProperty Annotation to Solr's Property annotation
```
dsmiley: Use actual Javadoc comments please.  You should say why we're even doing this without requiring the reader to read SOLR-13841 (though referring to the issue is fine for another level of detail).  For example I think it's important to say here that we don't want SolrJ to depend on Jackson-annotations, and thus we duplicate the functionality for one annotation we've replicated.  A one-liner reference for how to install this to an ObjectMapper is helpful.

The current name isn't bad but I suggest naming this with Jackson in the name, like "SolrJacksonAnnotationInspector" to very clearly show this is Jackson related.

solr/core/src/test/org/apache/solr/util/TestSolrJacksonAnnotation.java
```diff
@@ -0,0 +1,82 @@
...
+    assertEquals("v1",  m.get("field"));
+    assertEquals("v2",  m.get("friendlyName"));
+    assertEquals("1234",  String.valueOf(m.get("friendlyIntFld")));
+    TestObj o1 = mapper.readValue(json, TestObj.class);
+
+    assertEquals("v1", o1.field);
+    assertEquals("v2", o1.f2);
+    assertEquals(1234, o1.ifld);
+
+    Map<String, Object> schema = JsonSchemaCreator.getSchema(TestObj.class);
```
dsmiley: Overall, this is fine but I thought I'd share that my preference for testing objects is simply a toString and equals comparison against a string literal.  This is way easier to write and understand (IMO), less code, and is easy to see as more _complete_ than by-field assertEquals since you don't need to ask yourself if mysteriously other stuff is in there that is not tested.  But I understand this is a matter of taste and you are perfectly welcome to go with the approach here which is common.

solr/solrj/src/java/org/apache/solr/common/annotation/Property.java
```diff
@@ -0,0 +1,36 @@
...
+package org.apache.solr.common.annotation;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Target({ElementType.FIELD, ElementType.METHOD})
+@Retention(RetentionPolicy.RUNTIME)
+public @interface Property {
```
dsmiley: I insist all top level non-test classes require documentation.  Consider doing this as a habit at the very time you write/name the class at first.  I find that practice helps me.   For this class's javadocs, mention the annotation processor that *uses* it.

I think we should name this the same as Jackson's JsonProperty because it's semantically identical, people know what that means.  It's also a better name IMO -- "Property" alone doesn't mention JSON but isn't that pertinent?

solr/solrj/src/java/org/apache/solr/common/util/JsonSchemaCreator.java
```diff
@@ -0,0 +1,80 @@
...
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+import org.apache.solr.common.annotation.Property;
+
+public class JsonSchemaCreator {
```
dsmiley: Again; lets see some javadocs

solr/core/src/test/org/apache/solr/util/TestSolrJacksonAnnotation.java
noblepaul: I thought of doing String comparisons. But, it then depends on the whims of the json serializing library. If it fails sometime in the future nobody will know why

PR #994, opened 2019-11-04
SOLR-13662: Package Manager (CLI)

chatman:
Adding CLI support for package manager. PackageManagerCLITest is added for test of common operations.

Introduces new dependency on semver library.

Introduces new ZK file, /repositories.json.


solr/core/ivy.xml
```diff
@@ -138,6 +138,10 @@
     <dependency org="com.google.protobuf" name="protobuf-java" rev="${/com.google.protobuf/protobuf-java}" conf="compile"/>
     <dependency org="com.jayway.jsonpath" name="json-path" rev="${/com.jayway.jsonpath/json-path}" conf="compile"/>
 
+    <dependency org="com.github.zafarkhaja" name="java-semver" rev="0.9.0" conf="compile"/>
+    <dependency org="net.minidev" name="json-smart" rev="2.2.1" conf="compile"/>
+    <dependency org="net.minidev" name="asm" rev="1.0.2" conf="compile"/>
```
chatman: There's a problem including asm here. This is needed for semver to work. Will look at some solution to fix this.

solr/core/src/java/org/apache/solr/packagemanager/SolrPackageRepository.java
```diff
@@ -0,0 +1,130 @@
...
+      log.debug("Read plugins of '{}' repository from '{}'", id, pluginsUrl);
+      pluginsJsonReader = new InputStreamReader(pluginsUrl.openStream());
+    } catch (Exception e) {
+      log.error(e.getMessage(), e);
+      packages = Collections.emptyMap();
+      return;
+    }
+
+    ObjectMapper mapper = new ObjectMapper();
+    SolrPackage items[];
```
dsmiley: move the brackets

solr/core/src/java/org/apache/solr/packagemanager/SolrUpdateManager.java
```diff
@@ -0,0 +1,218 @@
...
+import org.apache.solr.client.solrj.response.V2Response;
+import org.apache.solr.common.SolrException;
+import org.apache.solr.common.SolrException.ErrorCode;
+import org.apache.solr.packagemanager.SolrPackage.SolrPackageRelease;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+public class SolrUpdateManager {
```
dsmiley: javadocs

solr/core/src/java/org/apache/solr/packagemanager/SolrUpdateManager.java
```diff
@@ -0,0 +1,218 @@
...
+
+  final private SolrPackageManager packageManager;
```
dsmiley: final goes after scope

solr/core/src/java/org/apache/solr/packagemanager/SolrUpdateManager.java
```diff
@@ -0,0 +1,218 @@
...
+      add.manifest = "/package/" + packageName + "/" + version + "/solr-manifest.json";
+      add.manifestSHA512 = "MY_MANIFEST_SHA512"; // nocommit: hardcoded sha512 of manifest must go away
+
+      V2Request req = new V2Request.Builder("/api/cluster/package")
+          .forceV2(true)
+          .withMethod(SolrRequest.METHOD.POST)
+          .withPayload(Collections.singletonMap("add", add))
+          .build();
+
+      try {
```
dsmiley: this try-catch can be removed as redundant with outer nesting

solr/core/src/java/org/apache/solr/util/PackageTool.java
```diff
@@ -0,0 +1,351 @@
...
+import org.apache.solr.util.SolrCLI.StatusTool;
+import org.apache.zookeeper.CreateMode;
+import org.apache.zookeeper.KeeperException;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.jayway.jsonpath.JsonPath;
+import com.jayway.jsonpath.PathNotFoundException;
+
+
+public class PackageTool extends SolrCLI.ToolBase {
```
dsmiley: javadocs

chatman:
Thanks David. There's a lot of cleanup and logging improvements to be done here. Will delegate some of the http calls to the SolrClient so that SSL can be handled without trusting all connections blindly. This is WIP work, so i hope in the next iteration it will be in a shape where careful review will be helpful. For now, high level flow should be understood from the code here.

solr/core/src/java/org/apache/solr/packagemanager/SolrPackage.java
```diff
@@ -0,0 +1,107 @@
...
+    public String version;
+    public Date date;
+    public String url;
+
+    public String sha512sum;
+    public String sig;
+
+    public Metadata metadata;
+    @Override
+    public String toString() {
```
noblepaul: implement MapWriter, and toString can be using  `jsonStr()`

or even better , implement `ReflectMapWriter` and annotate your fields using `@JsonProperty`

solr/core/src/java/org/apache/solr/packagemanager/SolrPackage.java
```diff
@@ -0,0 +1,107 @@
...
+    public String updateCommand;
+
+    @JsonProperty("uninstall-command")
+    public String uninstallCommand;
+
+    @JsonProperty("verify-command")
+    public Command verifyCommand;
+
+    @Override
+    public String toString() {
```
noblepaul: again , just use the `jsonStr()` method here

solr/core/src/java/org/apache/solr/packagemanager/SolrPackage.java
```diff
@@ -0,0 +1,107 @@
...
+
+  public static class Command {
+    public String path;
+    public String method;
+    public Map payload;
+    public String condition;
+    public String expected;
+    
+    @Override
+      public String toString() {
```
noblepaul: implement `ReflectMapWriter` and use `jsonStr()`

solr/core/src/java/org/apache/solr/packagemanager/SolrPackageManager.java
```diff
@@ -0,0 +1,259 @@
...
+
+  public List<SolrPackageInstance> getPackages() throws SolrException {
+    System.out.println("Getting packages from clusterprops...");
+    List<SolrPackageInstance> ret = new ArrayList<SolrPackageInstance>();
+    packages = new HashMap<String, List<SolrPackageInstance>>();
+    try {
+      String clusterPropsJson = null;
+      Map packagesJson = null;
+
+      if (zkClient.exists("/packages.json", true) == true) {
```
noblepaul: do not use ZooKeeper directly, use the http API

solr/core/src/java/org/apache/solr/packagemanager/SolrPackageRepository.java
```diff
@@ -0,0 +1,130 @@
...
+
+import org.apache.commons.io.FileUtils;
+import org.apache.solr.common.SolrException;
+import org.apache.solr.common.SolrException.ErrorCode;
+import org.apache.solr.packagemanager.SolrPackage.SolrPackageRelease;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonProperty;
```
noblepaul: do no use `com.fasterxml.jackson.annotation.JsonProperty`

use the Solr annotation `org.apache.solr.common.annotation.JsonProperty` instead everywhere

solr/core/src/java/org/apache/solr/packagemanager/SolrPackage.java
```diff
@@ -0,0 +1,107 @@
+package org.apache.solr.packagemanager;
+
+
+import java.io.Serializable;
+import java.util.Date;
+import java.util.List;
+import java.util.Map;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
```
noblepaul: use `org.apache.solr.common.annotation.JsonProperty`

solr/core/ivy.xml
```diff
@@ -138,6 +138,9 @@
     <dependency org="com.google.protobuf" name="protobuf-java" rev="${/com.google.protobuf/protobuf-java}" conf="compile"/>
     <dependency org="com.jayway.jsonpath" name="json-path" rev="${/com.jayway.jsonpath/json-path}" conf="compile"/>
 
+    <!-- Package manager -->
```
noblepaul: what is this for?

solr/core/src/java/org/apache/solr/packagemanager/PackageUtils.java
```diff
@@ -0,0 +1,141 @@
...
+    ByteBuffer jar;
+    try (FileInputStream fis = new FileInputStream(file)) {
+      byte[] buf = new byte[fis.available()];
+      fis.read(buf);
+      jar = ByteBuffer.wrap(buf);
+    }
+    return jar;
+  }
+
+  public static void postFile(SolrClient client, ByteBuffer buffer, String name, String sig)
```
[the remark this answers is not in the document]
chatman: I think the package store / file store should deal with those kinds of input validation.

solr/core/src/java/org/apache/solr/packagemanager/SolrPackage.java
chatman: I wasn't able to. None of the serialization/deserialization stuff worked.

solr/core/src/java/org/apache/solr/packagemanager/SolrPackage.java
chatman: Done, thanks!

solr/core/src/java/org/apache/solr/packagemanager/SolrPackage.java
chatman: Done, thanks!

solr/core/src/java/org/apache/solr/packagemanager/SolrPackage.java
chatman: Done, thanks!

solr/core/ivy.xml
chatman: This is for semantic versions (https://semver.org/). This library gives us comparison and range validations of versions.

solr/core/src/java/org/apache/solr/util/PackageTool.java
chatman: Added wherever applicable. Thanks.

solr/core/src/java/org/apache/solr/packagemanager/PackageUtils.java
```diff
@@ -0,0 +1,141 @@
...
+  public static Configuration jsonPathConfiguration() {
+    MappingProvider provider = new JacksonMappingProvider();
+    JsonProvider jsonProvider = new JacksonJsonProvider();
+    Configuration c = Configuration.builder().jsonProvider(jsonProvider).mappingProvider(provider).options(com.jayway.jsonpath.Option.REQUIRE_PROPERTIES).build();
+    return c;
+  }
+  
+  public static ByteBuffer getFileContent(File file) throws IOException {
+    ByteBuffer jar;
+    try (FileInputStream fis = new FileInputStream(file)) {
```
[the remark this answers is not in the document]
chatman: Removed this in favour of IOUtils.

solr/core/src/java/org/apache/solr/packagemanager/PackageUtils.java
```diff
@@ -0,0 +1,141 @@
...
+public class PackageUtils {
+  
+  public static Configuration jsonPathConfiguration() {
+    MappingProvider provider = new JacksonMappingProvider();
+    JsonProvider jsonProvider = new JacksonJsonProvider();
+    Configuration c = Configuration.builder().jsonProvider(jsonProvider).mappingProvider(provider).options(com.jayway.jsonpath.Option.REQUIRE_PROPERTIES).build();
+    return c;
+  }
+  
+  public static ByteBuffer getFileContent(File file) throws IOException {
```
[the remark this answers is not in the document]
chatman: Removed this in favour of IOUtils.

solr/core/src/java/org/apache/solr/packagemanager/PackageUtils.java
```diff
@@ -0,0 +1,141 @@
...
+        .withParams(params)
+        .build()
+        .process(client);
+    if (!name.equals(rsp.getResponse().get(CommonParams.FILE))) {
+      throw new SolrException(ErrorCode.BAD_REQUEST, "Mismatch in file uploaded. Uploaded: " +
+          rsp.getResponse().get(CommonParams.FILE)+", Original: "+name);
+    }
+  }
+
+  public static String getStringFromStream(String url) {
```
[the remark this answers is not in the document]
chatman: Removed this in favour of IOUtils.

solr/core/src/java/org/apache/solr/packagemanager/PackageUtils.java
```diff
@@ -0,0 +1,141 @@
...
+    return get(url);
+  }
+
+  public static String get(String url) {
```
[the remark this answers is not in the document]
chatman: Removed this in favour of IOUtils.

solr/core/src/java/org/apache/solr/packagemanager/SolrUpdateManager.java
chatman: Done, thanks!

chatman: BTW, this class is now called RepositoryManager.

solr/core/src/java/org/apache/solr/packagemanager/PackageRepository.java
```diff
@@ -0,0 +1,54 @@
...
+import org.apache.solr.common.SolrException;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+/**
+ * Abstract class for a repository, holding {@link SolrPackage} items.
+ */
+public abstract class PackageRepository {
+
+  @JsonProperty("name")
```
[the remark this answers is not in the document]
chatman: Done, thanks.

solr/core/src/java/org/apache/solr/packagemanager/PackageUtils.java
```diff
@@ -0,0 +1,212 @@
...
+      throws SolrServerException, IOException {
+    String resource = "/api/cluster/files" + name;
+    ModifiableSolrParams params = new ModifiableSolrParams();
+    if (sig != null) {
+      params.add("sig", sig);
+    }
+    V2Response rsp = new V2Request.Builder(resource)
+        .withMethod(SolrRequest.METHOD.PUT)
+        .withPayload(buffer)
+        .forceV2(true)
```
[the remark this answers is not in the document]
chatman: I tried removing this, but tests failed (test doesn't directly called this; it is CLI calling Solr). Is there any downside to keeping it in?

solr/core/src/java/org/apache/solr/packagemanager/RepositoryManager.java
```diff
@@ -0,0 +1,328 @@
...
+      // Call Package API to add this version of the package
+      Package.AddVersion add = new Package.AddVersion();
+      add.version = version;
+      add.pkg = packageName;
+      add.files = downloaded.stream().map(file -> "/package/" + packageName + "/" + version + "/" + file.getFileName().toString()).collect(Collectors.toList());  
+      add.manifest = "/package/" + packageName + "/" + version + "/manifest.json";
+      add.manifestSHA512 = manifestSHA512;
+
+      V2Request req = new V2Request.Builder("/api/cluster/package")
+          .forceV2(true)
```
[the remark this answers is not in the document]
chatman: See comment above; test failed without test.

solr/core/src/java/org/apache/solr/packagemanager/RepositoryManager.java
```diff
@@ -0,0 +1,328 @@
...
+    String latestVersion = getLastPackageRelease(packageName).version;
+
+    Map<String, String> collectionsDeployedIn = packageManager.getDeployedCollections(packageName);
+    List<String> peggedToLatest = collectionsDeployedIn.keySet().stream().
+        filter(collection -> collectionsDeployedIn.get(collection).equals("$LATEST")).collect(Collectors.toList());
+    if (!peggedToLatest.isEmpty()) {
+      PackageUtils.printGreen("Collections that will be affected (since they are configured to use $LATEST): "+peggedToLatest);
+    }
+
+    if (version == null || version.equals("latest")) {
```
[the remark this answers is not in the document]
chatman: If user wants to install mypackage or mypackage:latest, it has the same effect here. It will find the most recent package in repository and install it.

solr/core/src/java/org/apache/solr/packagemanager/SolrPackageRepository.java
dsmiley: This mistake is extremely easy to make.  Perhaps your new json annotation processory thingy should detect `com.fasterxml.jackson.annotation.JsonProperty` and fail?

dsmiley: Or configure forbiddenAPI

chatman: +1 to forbidden API. I had suggested (privately) to Noble to put jackson's JsonProperty to forbidden APIs and he agreed. Also, I think JsonIgnore should be added to Solr's annotations, instead of my trying to use jackson's version.

solr/core/src/java/org/apache/solr/packagemanager/PackageUtils.java
```diff
@@ -90,52 +88,123 @@ public static void postFile(SolrClient client, ByteBuffer buffer, String name, S
...
-          return(results);
-        }
-      } catch (IOException e) {
-        e.printStackTrace();
+  /**
+   * Search through the list of jar files for a given file. Returns string of
+   * the file contents or null if file wasn't found. This is suitable for looking
+   * for manifest or property files within pre-downloaded jar files.
+   */
+  public static String getFileFromArtifacts(List<Path> jars, String filename) {
```
dsmiley: I suggest appending "AsString" to the name.  Maybe replace "Artifacts" with "Jars" at your discretion.

solr/core/src/java/org/apache/solr/packagemanager/PackageUtils.java
```diff
@@ -90,52 +88,123 @@ public static void postFile(SolrClient client, ByteBuffer buffer, String name, S
...
-  public static CloseableHttpClient createTrustAllHttpClientBuilder() throws Exception {
-    SSLContextBuilder builder = new SSLContextBuilder();
-    builder.loadTrustMaterial(null, (chain, authType) -> true);           
-    SSLConnectionSocketFactory sslsf = new 
-        SSLConnectionSocketFactory(builder.build(), NoopHostnameVerifier.INSTANCE);
-    return HttpClients.custom().setSSLSocketFactory(sslsf).build();
+  /**
+   * Returns JSON string from a given URL
+   */
+  public static String getJson(HttpClient client, String url) {
```
dsmiley: Rename: getJsonStringFromUrl

solr/core/src/java/org/apache/solr/packagemanager/PackageManager.java
```diff
@@ -0,0 +1,406 @@
...
+            throw new SolrException(ErrorCode.BAD_REQUEST, "Non-POST method not supported for uninstall commands");
+          }
+        } else {
+          PackageUtils.printRed("There is no uninstall command to execute for plugin: " + plugin.name);
+        }
+      }
+
+      // Set the package version in the collection's parameters
+      try {
+        SolrCLI.postJsonToSolr(solrClient, "/api/collections/" + collection + "/config/params", "{set: {PKG_VERSIONS: {"+packageName+": null}}}");
```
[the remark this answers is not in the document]
chatman: Solr's params API only sets the specified package to null. I think we're good :-)

solr/core/src/java/org/apache/solr/packagemanager/PackageUtils.java
chatman: Sounds good, I'll make that change!

solr/core/src/java/org/apache/solr/packagemanager/PackageUtils.java
chatman: Sounds good, I'll make that change!

solr/core/src/java/org/apache/solr/packagemanager/SolrPackage.java
```diff
@@ -0,0 +1,143 @@
...
+    @JsonProperty("version")
+    public String version;
+
+    @JsonProperty("date")
+    public Date date;
+
+    @JsonProperty("artifacts")
+    public List<Artifact> artifacts;
+
+    @JsonProperty("manifest")
```
janhoy: Won't this class represent the content of `manifest.json`? In that case it could be misleading to have a property `manifest` inside? Perhaps I misunderstand?

solr/core/src/java/org/apache/solr/util/PackageTool.java
```diff
@@ -0,0 +1,255 @@
...
+          packageManager = new PackageManager(solrClient, solrBaseUrl, zkHost); 
+          try {
+            repositoryManager = new RepositoryManager(solrClient, packageManager);
+
+            switch (cmd) {
+              case "add-repo":
+                repositoryManager.addRepository(cli.getArgs()[1], cli.getArgs()[2]);
+                break;
+              case "list-installed":
+                packageManager.listInstalled();
```
janhoy: Same as above - should the CLI formatting and printGreen stuff happen in this class and not in the PackageManager?

solr/core/src/java/org/apache/solr/util/PackageTool.java
```diff
@@ -0,0 +1,255 @@
...
+                PackageUtils.printGreen(repositoryManager.toString() + " installed.");
+                break;
+              }
+              case "deploy":
+              {
+                String packageName = parsePackageVersion(cli.getArgList().get(1).toString())[0];
+                String version = parsePackageVersion(cli.getArgList().get(1).toString())[1];
+                boolean noprompt = cli.hasOption('y');
+                boolean isUpdate = cli.hasOption("update") || cli.hasOption('u');
+                packageManager.deploy(packageName, version, cli.getOptionValues("collections"), cli.getOptionValues("param"), isUpdate, noprompt);
```
janhoy: Should we sanitize collection names to validate they do not contain `/` which would break the construction of URL later?

solr/core/src/java/org/apache/solr/util/PackageTool.java
```diff
@@ -0,0 +1,255 @@
...
+              }
+              case "undeploy":
+              {
+                String packageName = parsePackageVersion(cli.getArgList().get(1).toString())[0];
+                packageManager.undeploy(packageName, cli.getOptionValues("collections"));
+                break;
+              }
+              case "help":
+              case "usage":
+                System.out.println("./solr package add-repo <repository-name> <repository-url>");
```
janhoy: Perhaps create a method `out(String msg)` to avoid all these?

solr/core/src/java/org/apache/solr/util/PackageTool.java
```diff
@@ -0,0 +1,255 @@
...
+    }
+  }
+
+  /**
+   * Parses package name and version in the format "name:version" or "name"
+   * @param arg User supplied argument
+   * @return Array of two elements, first the package name, second version (or null if not present)
+   */
+  private String[] parsePackageVersion(String arg) {
+    String packageName = arg.split(":")[0];
```
janhoy: Check for extra `:` as well, i.e. `foo:1.2.3:bar` is not allowed? Could just as well validate the version part as a legal version too.

I'm not very fond of return type `String[]` where the caller must know some magic. Consider `Pair<String,com.github.zafarkhaja.semver.Version>` as a more clear return type.

solr/core/src/test-files/solr/question-answer-repository/repository.json
```diff
@@ -0,0 +1,57 @@
...
+        "date": "2019-01-01",
+        "artifacts": [
+          {
+            "url": "question-answer-request-handler-1.0.jar",
+            "sig": "C9UWKkucmY3UNzqn0VLneVMe9kCbJjw7Urc76vGenoRwp32xvNn5ZIGZ7G34xZP7cVjqn/ltDlLWBZ/C3eAtuw=="
+          }
+        ],
+        "manifest": {
+          "min-solr-version": "8.0",
+          "max-solr-version": "9.99",
```
janhoy: 9.99 isn't very compelling example. If we switch to semVer syntax we don't need these hacks, The min-max combination you try to express here could instead be

    "require-solr-version": "8 - 9"

solr/core/src/test-files/solr/question-answer-repository/repository.json
```diff
@@ -0,0 +1,57 @@
...
+          "plugins": [
+            {
+              "name": "request-handler",
+              "setup-command": {
+                "path": "/api/collections/${collection}/config",
```
janhoy: Will you allow a package to run any HTTP command against the cluster? What privilege does the pkg manager run under? Imagine someone posts a malicious package "superduper-solr-package" somewhere that attempts to run a setup-command  `/api/authentication/disable` :-) or `/api/c/.system/update/foo` or some other arbitrary command?

That's one reason I think we should turn this whole deploy upside down and let Solr plugin points fetch info from the plugin instead of the plugin author hardcoding some literal api commands against the cluster. ´

solr/core/src/java/org/apache/solr/packagemanager/RepositoryManager.java
```diff
@@ -0,0 +1,328 @@
...
+        release.manifest = getMapper().readValue(manifestJson, SolrPackage.Manifest.class);
+      }
+      String manifestJson = getMapper().writeValueAsString(release.manifest);
+      String manifestSHA512 = BlobRepository.sha512Digest(ByteBuffer.wrap(manifestJson.getBytes("UTF-8")));
+      PackageUtils.postFile(solrClient, ByteBuffer.wrap(manifestJson.getBytes("UTF-8")),
+          "/package/" + packageName + "/" + version + "/manifest.json", null);
+
+      // post the artifacts
+      PackageUtils.printGreen("Posting artifacts");
+      for (int i=0; i<release.artifacts.size(); i++) {
```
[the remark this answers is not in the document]
chatman: Wasn't able to, since there are two lists that are index aligned: downloaded and release.artifacts

solr/core/src/java/org/apache/solr/packagemanager/SolrPackage.java
chatman: This represents a version / release. The manifest is inside (either explicitly specified in the repository.json or contained within the jar). Here's an example: https://github.com/apache/lucene-solr/pull/994/files#diff-ab1e50fd5605ee1a39764390b9d4faeeR6-R42

solr/core/src/test-files/solr/question-answer-repository/repository.json
chatman: Package manager (CLI) needs write access to ZK for "add-repo", since it needs to publish the trusted public key. A user must only "add-repo" for a trusted publisher. Only then should the user expect non-malicious packages.
Once a package is already in the system, it doesn't matter if the deploy step (which is a CLI action) executes a HTTP command or does something malicious from inside the jar file.

solr/bin/solr
```diff
@@ -764,6 +764,56 @@ function get_info() {
...
+          CODE=$?
+          echo ""
+        else
+          echo -e "\nSolr process $ID from $PIDF not found."
+          CODE=1
+        fi
+    done < <(find "$SOLR_PID_DIR" -name "solr-*.pid" -type f)
+  else
+    # no pid files but check using ps just to be sure
+    numSolrs=`ps auxww | grep start\.jar | grep solr\.solr\.home | grep -v grep | wc -l | sed -e 's/^[ \t]*//'`
```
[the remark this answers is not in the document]
chatman: Sure, I'll add this later (don't have access to a Windows machine at the moment).

chatman:
> Cannot find RefGuide docs in this PR? Better to include it now than after merge to master.
I'll add it in a separate PR so that it is easier for Cassandra to review.

> Question on signatures. Can we have a pure Java way of signing packages or do we force use of openssl?
Wasn't aware of any such way. I actually want to use GPG (since Apache commiters all use trusted GPG keys). I just didn't feel like tackling the bouncy castle dependency at the moment. Maybe, we can add that later.

solr/bin/solr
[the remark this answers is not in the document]
chatman: Added a script that doesn't have the magic of auto-supplying the -solrUrl parameter (based on a running Solr instance). User would have to supply that himself/herself while using the package manager. I'll open a separate JIRA to remove this annoyance that I can't tackle myself at the moment.

solr/core/src/java/org/apache/solr/packagemanager/PackageManager.java
```diff
@@ -0,0 +1,415 @@
...
+      }
+    } catch (Exception e) {
+      throw new SolrException(ErrorCode.BAD_REQUEST, e);
+    }
+    log.info("Got packages: "+ret);
+    return ret;
+  }
+
+  public Map<String, SolrPackageInstance> getPackagesDeployed(String collection) {
+    String paramsJson = PackageUtils.getJsonStringFromUrl(solrClient.getHttpClient(), solrBaseUrl+"/api/collections/"+collection+"/config/params/PKG_VERSIONS?omitHeader=true");
```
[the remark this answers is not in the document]
chatman: Makes the code look ugly, but done this anyway.

chatman:
> do you limit the calls to /api/collections/${collection}/config* /api/collections/${collection}/schema* ?

No, any http request to Solr is supported. If, in future, there's a configset API that accepts commands for registering components, then those can be used as well. I don't think a whitelisting of paths is a good idea.

chatman:
Merged, thanks. https://issues.apache.org/jira/browse/SOLR-13662?focusedCommentId=16974218&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-16974218

Thanks for all your reviews!

janhoy:
>I just didn't feel like tackling the bouncy castle dependency at the moment
Java 11 has stong crypto included ootb so should be easier. We can add that in master and not backport to 8.

solr/core/src/java/org/apache/solr/packagemanager/RepositoryManager.java
```diff
@@ -0,0 +1,328 @@
...
+    // Currently, not a big problem since signature based checking happens anyway
+
+    try {
+      // post the manifest
+      PackageUtils.printGreen("Posting manifest");
+
+      if (release.manifest == null) {
+        String manifestJson = PackageUtils.getFileFromJarsAsString(downloaded, "manifest.json");
+        if (manifestJson == null) {
+          throw new SolrException(ErrorCode.BAD_REQUEST, "No manifest found for package: " + packageName + ", version: " + version);
```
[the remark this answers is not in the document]
janhoy: You resolved my review comment without a reply or resolution?

solr/core/src/java/org/apache/solr/packagemanager/RepositoryManager.java
```diff
@@ -0,0 +1,328 @@
...
+      for (PackageRepository repo: getRepositories()) {
+        if (repo.hasPackage(packageName)) {
+          for (Artifact art: release.artifacts) {
+            downloadedPaths.add(repo.download(art.url));
+          }
+          return downloadedPaths;
+        }
+      }
+    } catch (IOException e) {
+      throw new SolrException(ErrorCode.BAD_REQUEST, "Error during download of package " + packageName, e);
```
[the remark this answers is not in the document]
janhoy: You resolved my review comment without a reply or resolution?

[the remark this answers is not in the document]
chatman: I've fixed it in a subsequent commit.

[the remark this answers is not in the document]
chatman: I missed the point about retry logic. I think it is a good idea and that can be taken up as an enhancement.

solr/core/src/java/org/apache/solr/util/PackageTool.java
chatman: Done, I've fixed it in a subsequent commit.

solr/core/src/java/org/apache/solr/util/PackageTool.java
chatman: I've fixed it in a subsequent commit.

chatman:
Sounds good. If we get GPG support in JDK or some decent library, it will be great.

solr/core/src/java/org/apache/solr/packagemanager/RepositoryManager.java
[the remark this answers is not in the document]
janhoy: Yea, see it in master now. Better to complete the review in PR to have everything contained and answer all questions in one place!

PR #1002, opened 2019-11-10
[docs]: fixed two small errors in JavaDoc

0xflotus:
I only fixed two small typo errors in the JavaDocs

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/DisjunctionMaxQuery.java
+++ b/lucene/core/src/java/org/apache/lucene/search/DisjunctionMaxQuery.java
@@ -262,9 +262,9 @@ public final class DisjunctionMaxQuery extends Query implements Iterable<Query> 
     return buffer.toString();
   }
 
-  /** Return true iff we represent the same query as o
+  /** Return true if we represent the same query as other
    * @param other another object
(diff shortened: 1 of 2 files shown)


lucene/core/src/java/org/apache/lucene/search/DisjunctionMaxQuery.java
```diff
@@ -262,9 +262,9 @@ public String toString(String field) {
     return buffer.toString();
   }
 
-  /** Return true iff we represent the same query as o
```
gerlowskija: I've seen "iff" used before as shorthand for "if and only if".  So maybe that's what the original author here was going for?  But the docs also work without it, so it's hard to say for sure.

gerlowskija:
These changes look fine to me.  One "typo" was probably intentional ("iff" vs "if"), but the docs work equally well either way (If you only describe one possibility for how a method might return "true", then "if and only if" is already implicit).  So I'm not going to bother trying to read tea-leaves on whether the author intended iff.

I'm re-running the precommit check- it's unclear why it failed earlier.  But if all works out I'll merge this in a few minutes.

Thanks for the docs @0xflotus, sorry it took months to get eyes on it.

PR #1020, opened 2019-11-20
LUCENE-9055: Fixes the detection of lines crossing triangles through edge points

iverase:
When a line crosses a triangle through an edge point, the crossing is not detected.


iverase:
@nknize do you think we can make this fix for 8.4?

PR #1024, opened 2019-11-21
LUCENE-9050: MultiTermIntervalsSource should call visitLeaf() in visit

romseygeek:


lucene/queries/src/java/org/apache/lucene/queries/intervals/MultiTermIntervalsSource.java
```diff
@@ -96,7 +96,7 @@ public MatchesIterator matches(String field, LeafReaderContext ctx, int doc) thr
 
   @Override
   public void visit(String field, QueryVisitor visitor) {
-
+    visitor.visitLeaf(new IntervalQuery(field, this));
```
mkhludnev: Isn't it better to stub AtomatonQuery and yield it here. IIRC it resolves simplest MTQ highlighting cases as nobrainer.

lucene/queries/src/test/org/apache/lucene/queries/intervals/TestIntervals.java
```diff
@@ -770,6 +829,8 @@ public void testPrefix() throws IOException {
       });
       assertEquals("Automaton [p*] expanded to too many terms (limit 1)", e.getMessage());
     }
+
+    checkVisits(Intervals.prefix(new BytesRef("p")), 1);
```
jimczi: Can you add the same for the `testWildcard` below ?

lucene/queries/src/java/org/apache/lucene/queries/intervals/MultiTermIntervalsSource.java
romseygeek: I don't really like the idea of reporting back an entirely different query here.  I think for highlighting multi-terms we need to extend QueryVisitor to add a `consumeTermsMatching(CompiledAutomaton ca)` method, or similar.  I'll open a follow-up to discuss - there are tricky things involved with char-vs-byte automatons etc.

PR #1028, opened 2019-11-21
SOLR-13101:These are pretty much non functional changes, related to S…

mbwaheed:
…HARED replica concurrent updates
https://github.com/apache/lucene-solr/pull/983/commits/581f468f9914ce2488201efbed42fd43dc4b481

- Pull read lock is only required before the local indexing starts. If indexing has to pull it can acquire and release write lock without anything to do with read lock first. Removing the unneeded lock upgrade/downgrade logic to simplify things.
- Added one pager explanation in the start of SharedCoreConcurrencyController around overall concurrency design.
- Added mores comments around BlobCoreMetadata#generation number and its usage.


yonik:
Thanks Bilal, great detail!  Also great timing since at the recent committers meeting, a number of people expressed the desire to have more comments actually in the code around design.

PR #1031, opened 2019-11-22
LUCENE-9059: Reduce garbage created by ByteBuffersDataOutput.

jpountz:


lucene/core/src/java/org/apache/lucene/store/ByteBuffersDataOutput.java
```diff
@@ -412,7 +413,11 @@ public long ramBytesUsed() {
    * lead to hard-to-debug issues, use with great care.
    */
   public void reset() {
-    blocks.stream().forEach(blockReuse);
+    if (blockReuse != NO_REUSE) {
```
jpountz: This check isn't related to what I saw in the profile, though it looks like it could be an easy win in some cases. The important change is the move from `forEach` to a `for` loop.

lucene/core/src/java/org/apache/lucene/store/ByteBuffersDataOutput.java
```diff
@@ -279,11 +279,12 @@ public ByteBuffersDataInput toDataInput() {
    * Copy the current content of this object into another {@link DataOutput}.
    */
   public void copyTo(DataOutput output) throws IOException {
-    for (ByteBuffer bb : toBufferList()) {
+    for (ByteBuffer bb : blocks) {
       if (bb.hasArray()) {
```
jpountz: this condition was always `false` before because toBufferList calls `asReadOnlyBuffer` on the buffers

jpountz:
@dweiss Would you like to have a look?

lucene/core/src/java/org/apache/lucene/store/ByteBuffersDataOutput.java
dweiss: Good catch!

lucene/core/src/java/org/apache/lucene/store/ByteBuffersDataOutput.java
dweiss: That .stream() call is actually not necessary there -- you could just pass a closure to blocks.forEach... This should be pretty much equivalent to having a for loop. I don't mind switching to more verbose format, I just wanted be hip and groovy with streams...

dweiss:
Oops, sorry Adrien - I must have missed your initial request. LGTM, thank you!

lucene/core/src/java/org/apache/lucene/store/ByteBuffersDataOutput.java
jpountz: Good point, I'll change it to `blocks.forEach(blockReuse)`.

PR #1033, opened 2019-11-24
SOLR-13965: Use Plugin to add new expressions to GraphHandler

epugh:
# Description

Adding new streaming expression functions is handled differently in StreamHandler from GraphHandler, and I believe they should be handled the same way.

# Solution

I removed the old `<streamFunctions/>` configuration in favor of using plugins in GraphHandler, which was how StreamHandler did it.

# Tests
No test changes, as there were none around this previously.

# Checklist

- [ X] I have reviewed the guidelines for [How to Contribute](https://wiki.apache.org/solr/HowToContribute) and my code conforms to the standards described there to the best of my ability.
- [ X] I have created a Jira issue and added the issue ID to my pull request title.
- [ X] I am authorized to contribute this code to the ASF and have removed any code I do not have a license to distribute.
- [ X] I have given Solr maintainers [access](https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork) to contribute to my PR branch. (optional but recommended)
- [ X] I have developed this patch against the `master` branch.
- [ X] I have run `ant precommit` and the appropriate test suite.
- [ X] I have added documentation for the [Ref Guide](https://github.com/apache/lucene-solr/tree/master/solr/solr-ref-guide) (for Solr changes only).


solr/core/src/java/org/apache/solr/handler/GraphHandler.java
```diff
@@ -87,14 +88,10 @@ public void inform(SolrCore core) {
     }
 
     // This pulls all the overrides and additions from the config
-    Object functionMappingsObj = initArgs.get("streamFunctions");
```
cpoerschke: I wonder with backwards compatibility in mind, if for a period of time `streamFunctions` might be deprecated but still supported alongside the new `expressible` equivalent. Users would see the 'deprecated' warnings being logged and could transition callers ahead of `streamFunctions` support being taken away in a later version. What do you think?

epugh: I wondered the same thing.   I don't think anyone actually customizes this handler, but that is just a guess.   I can add a deprecated warning, do you have a good example that I could copy of how we normally do that?

epugh: @cpoerschke I am pushing up a fix to do the deprecations...

solr/core/src/java/org/apache/solr/handler/GraphHandler.java
```diff
@@ -92,24 +104,29 @@ public void inform(SolrCore core) {
...
       }
+    }
 
+    for (PluginInfo pluginInfo : pluginInfos) {
+      if (pluginInfo.pkgName != null) {
+        ExpressibleHolder holder = new ExpressibleHolder(pluginInfo, core, SolrConfig.classVsSolrPluginInfo.get(Expressible.class));
+        streamFactory.withFunctionName(pluginInfo.name,
+            () -> holder.getClazz());
+      } else {
+        Class<? extends Expressible> clazz = core.getMemClassLoader().findClass(pluginInfo.className, Expressible.class);
```
cpoerschke: Thanks @epugh for resolving the merge conflicts! The previous `getResourceLoader` is now replaced with `getMemClassLoader` here, just checking if that's intended or accidental perhaps, not yet looked what the difference (if any) between the two methods' behaviour is.

solr/core/src/java/org/apache/solr/handler/GraphHandler.java
```diff
@@ -92,24 +104,29 @@ public void inform(SolrCore core) {
...
-        if (pluginInfo.pkgName == null) {
-          Class<? extends Expressible> clazz = core.getResourceLoader().findClass((String) functionMapping.getValue(),
-              Expressible.class);
-          streamFactory.withFunctionName(key, clazz);
-        } else {
-          StreamHandler.ExpressibleHolder holder = new StreamHandler.ExpressibleHolder(pluginInfo, core, SolrConfig.classVsSolrPluginInfo.get(Expressible.class));
-          streamFactory.withFunctionName(key, () -> holder.getClazz());
-        }
-
+        pluginInfos.add(pluginInfo);
```
cpoerschke: The `pluginInfos.add` here raises an interesting question w.r.t. both `streamFunctions` and `expressible` being present and what should happen if both were to configure the same 'key' ...

My first thought was that simply disallowing the combination is easiest but actually that's not a good idea since `streamFunctions` is on a handler level whereas `expressible` is on the config level and so the presence of existing `expressible` elements would result in the also already present `streamFunctions` no longer working, which would not be very user backwards compatibility friendly.

If use of both elements is supported though then the deprecated variant being appended to the list could (at least theoretically) result in the deprecated `streamFunctions` configuration for name 'foo' using 'FooClass' overriding (for the GraphHandler only) any `expressible` mapping from name 'foo' to 'BarClass'. But then again, adding validation e.g. to reject any `streamFunctions` elements that are already present-and-different in `expressible` would add code complexity.

@joel-bernstein would you have any thoughts on this? Beyond the scope of this change here the question essentially is about multiple [StreamFactory.withFunctionName](https://github.com/apache/lucene-solr/blob/releases/lucene-solr/8.4.0/solr/solrj/src/java/org/apache/solr/client/solrj/io/stream/expr/StreamFactory.java#L85-L88) calls for the same function name.

solr/core/src/java/org/apache/solr/handler/GraphHandler.java
epugh: I'll be honest @cpoerschke I am mostly copying what is in StreamHandler, as this includes the new plugin framework support.   StreamHandler uses the `core.getMemClassLoader`, which when I click into seems to be all about plugins ;-).   My whole goal in this PR was to get StreamHandler and GraphHandler to be much more similar in how they work.

madrob: Since this code is duplicated between Stream & Graph, can we factor it out into a common method?

solr/core/src/java/org/apache/solr/handler/GraphHandler.java
```diff
@@ -92,24 +104,29 @@ public void inform(SolrCore core) {
...
       }
+    }
 
+    for (PluginInfo pluginInfo : pluginInfos) {
+      if (pluginInfo.pkgName != null) {
+        ExpressibleHolder holder = new ExpressibleHolder(pluginInfo, core, SolrConfig.classVsSolrPluginInfo.get(Expressible.class));
```
madrob: This line is likely hiding a bug (and also a bug in StreamHandler) because `SolrConfig.classVsSolrPluginInfo.get(Expressible.class)` should always return null due to type mismatch on the map. Need a String key, but we also don't have any tests to prove this out. If you don't want to address that in this issue, feel free to file a new JIRA and tag me on it. :)

epugh: @madrob can you push up a fix to this PR?

solr/core/src/java/org/apache/solr/handler/GraphHandler.java
cpoerschke: > Since this code is duplicated between Stream & Graph, can we factor it out into a common method?

Attached SOLR-13965.01.patch to the JIRA ticket.

solr/core/src/java/org/apache/solr/handler/GraphHandler.java
cpoerschke: > ... feel free to file a new JIRA ...

Opened https://issues.apache.org/jira/browse/SOLR-14212 for this.

solr/core/src/java/org/apache/solr/handler/GraphHandler.java
cpoerschke: https://github.com/epugh/lucene-solr/pull/1 offers to merge in the current `origin/master` into this PR and then adds usage of the factored out `StreamHandler.addExpressiblePlugins` method from SOLR-13965.01.patch file.

epugh:
Okay, @madrob or @cpoerschke how does this look?

epugh:
Okay, after digging around, I think this was committed to Master, so this PR can be closed?  Sorry about the noise @cpoerschke and @madrob

epugh:
Closing as this was actually committed.

PR #1034, opened 2019-11-24
SOLR-13863: payload query function now handles string encoded payload field (delimited_payloads_string)

freedev:
# Description

I started to use payloads because I had the classical per-store pricing problem.
But standard payload function handles only numbers and, for example, stores can also be in different countries, would be useful also have the currency and other attributes related to the store.

This pull request modify payload function query, this version is now able to read and sort payload field delimited_payloads_string.

Example document:

```
  "id":"my sample doc",
  "currencyPayload":[
    "store1|EUR",
    "store2|USD",
    "store3|GBP"
  ]
```

Querying Solr with

```
fl=payload(currencyPayload,store3)
```

would generate a response like the following:

```
  "response":{
      "docs":[{
          "id":"my sample doc", 
          "payload(currencyPayload,store3)":"GBP"
      }]
   }
```

And executing `payload(payloadCurrency,store2)` returns `EUR`, and so on.

You can use `payload` even as sorting function.

```
sort=payload(payloadField,value) asc
```

# Solution

I was looking at the original payload function is defined into the ValueSourceParser, this function uses a FloatPayloadValueSource to return the value found.

As said I fixed the current version of payload function, adding a part that handles strings. It is able to extract the string value from the payload.

Given the former example where I have a multivalue field payloadCurrency

```
payloadCurrency: [
"store1|USD",
"store2|EUR",
"store3|GBP"
]
```

executing payload(payloadCurrency,store2) returns "EUR", and so on for the remaining key/value in the field.

To implement the payload function, I've modified the ValueSourceParser where is defined the function payload and now returns a StringPayloadValueSource with the value inside (does the same thing of former FloatPayloadValueSource).

# Tests

I've created a sample solr collection where test the custom function
https://github.com/freedev/solr-payload-string-function-query 

# Checklist

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/PostingsEnum.java
+++ b/lucene/core/src/java/org/apache/lucene/index/PostingsEnum.java
@@ -68,9 +68,9 @@ public abstract class PostingsEnum extends DocIdSetIterator {
   }
 
   /**
-   * Returns an empty PostingsEnum object.
+   * Returns an dummy PostingsEnum object.
    */
-  public static PostingsEnum getEmptyInstance() {
+  public static PostingsEnum getDummyInstance() {
     return new PostingsEnum() {
       @Override
       public int freq() {
--- a/lucene/core/src/java/org/apache/lucene/index/PostingsEnum.java
+++ b/lucene/core/src/java/org/apache/lucene/index/PostingsEnum.java
@@ -68,10 +68,9 @@ public abstract class PostingsEnum extends DocIdSetIterator {
   }
 
   /**
-   * Returns an dummy PostingsEnum object.
+   * Dummy PostingsEnum object.
    */
-  public static PostingsEnum getDummyInstance() {
-    return new PostingsEnum() {
+  public static PostingsEnum DUMMY_INSTANCE = new PostingsEnum() {
       @Override
       public int freq() {
         return 0;
@@ -117,7 +116,6 @@ public abstract class PostingsEnum extends DocIdSetIterator {
         return 0;
       }
     };
-  }
 
   /**
    * Returns term frequency in the current document, or 1 if the field was
--- a/lucene/core/src/java/org/apache/lucene/index/PostingsEnum.java
+++ b/lucene/core/src/java/org/apache/lucene/index/PostingsEnum.java
@@ -68,6 +68,58 @@ public abstract class PostingsEnum extends DocIdSetIterator {
   }
 
   /**
+   * Returns an empty PostingsEnum object.
+   */
+  public static PostingsEnum getEmptyInstance() {
+    return new PostingsEnum() {
+      @Override
+      public int freq() {
+        return 0;
+      }
+
+      @Override
+      public int nextPosition() throws IOException {
+        return -1;
+      }
+
+      @Override
+      public int startOffset() throws IOException {
+        return -1;
+      }
+
+      @Override
+      public int endOffset() throws IOException {
+        return -1;
+      }
+
+      @Override
+      public BytesRef getPayload() throws IOException {
+        return null;
+      }
+
+      @Override
+      public int docID() {
+        return DocIdSetIterator.NO_MORE_DOCS;
+      }
+
+      @Override
+      public int nextDoc() {
+        return DocIdSetIterator.NO_MORE_DOCS;
+      }
+
+      @Override
+      public int advance(int target) {
+        return DocIdSetIterator.NO_MORE_DOCS;
+      }
+
+      @Override
+      public long cost() {
+        return 0;
+      }
+    };
+  }
+
+  /**
    * Returns term frequency in the current document, or 1 if the field was
    * indexed with {@link IndexOptions#DOCS}. Do not call this before
    * {@link #nextDoc} is first called, nor after {@link #nextDoc} returns
--- a/solr/core/src/java/org/apache/solr/search/PayloadValueSource.java
+++ b/solr/core/src/java/org/apache/solr/search/PayloadValueSource.java
@@ -18,7 +18,6 @@
 package org.apache.solr.search;
 
 import org.apache.lucene.queries.function.ValueSource;
-import org.apache.lucene.search.SortField;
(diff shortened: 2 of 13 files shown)


solr/core/src/java/org/apache/solr/search/StringPayloadValueSource.java
```diff
@@ -0,0 +1,306 @@
...
+          docValue = defaultValues.strVal(doc);
+          return docValue;
+        } catch (IOException e) {
+          throw new RuntimeException("caught exception in function " + description() + " : doc=" + doc, e);
+        }
+      }
+    };
+  }
+
+  private String bytesRefToString(BytesRef payload) {
```
mkhludnev: I suppose it should be just bytesRef.utf8ToString()

solr/core/src/java/org/apache/solr/search/StringPayloadValueSource.java
```diff
@@ -0,0 +1,306 @@
...
+
+  @Override
+  public int hashCode() {
+    int result = indexedField.hashCode();
+    result = 31 * result + (indexedBytes != null ? indexedBytes.hashCode() : 0);
+    result = 31 * result + defaultValueSource.hashCode();
+    return result;
+  }
+
+  class StringPayloadValueSourceSortField extends SortField {
```
mkhludnev: It doesn't seem like something unique, can't it be reused from other string ValsSourc?

solr/core/src/java/org/apache/solr/search/StringPayloadValueSource.java
```diff
@@ -0,0 +1,306 @@
...
+      return new ValueSourceComparator(context, numHits);
+    }
+  }
+
+  /**
+   * Implement a {@link org.apache.lucene.search.FieldComparator} that works
+   * off of the {@link FunctionValues} for a ValueSource
+   * instead of the normal Lucene FieldComparator that works off of a FieldCache.
+   */
+  class ValueSourceComparator extends SimpleFieldComparator<String> {
```
mkhludnev: again. Can't it be reused from somewhere?

solr/core/src/java/org/apache/solr/search/StringPayloadValueSource.java
```diff
@@ -0,0 +1,306 @@
...
+      PostingsEnum docs;
+      int atDoc;
+      int lastDocRequested = -1;
+      String docValue = null;
+
+      {
+        reset();
+      }
+
+      public void reset() throws IOException {
```
mkhludnev: As you said, there are numeric payload function. Can't these dragging functions unified and reused across them both?

solr/core/src/java/org/apache/solr/search/StringPayloadValueSource.java
freedev: right, removed.

lucene/core/src/java/org/apache/lucene/index/PostingsEnum.java
```diff
@@ -67,6 +67,56 @@ public static boolean featureRequested(int flags, short feature) {
   protected PostingsEnum() {
   }
 
+  /**
+   * Dummy PostingsEnum object.
+   */
+  public static PostingsEnum DUMMY_INSTANCE = new PostingsEnum() {
```
mkhludnev: Please look at org.apache.lucene.search.DocIdSetIterator.empty() 
There's subtle exhausting logic which needs to be supported and you can follow naming as well.

solr/core/src/java/org/apache/solr/search/PayloadValueSource.java
```diff
@@ -0,0 +1,66 @@
...
+    PayloadValueSource that = (PayloadValueSource) o;
+
+    if (!indexedField.equals(that.indexedField)) return false;
+    if (indexedBytes != null ? !indexedBytes.equals(that.indexedBytes) : that.indexedBytes != null) return false;
+    return defaultValueSource.equals(that.defaultValueSource);
+
+  }
+
+  @Override
+  public int hashCode() {
```
mkhludnev: How it will be supported by descendants?

solr/core/src/java/org/apache/solr/search/ValueSourceParser.java
```diff
@@ -722,13 +722,29 @@ public ValueSource parse(FunctionQParser fp) throws SyntaxError {
...
+        final FieldType fieldType = schema.getFieldType(tinfo.field);
+
+        final String payloadEncoder = PayloadUtils.getPayloadEncoder(fieldType);
+        if (payloadEncoder.equals("identity")) {
+
+          ValueSource defaultValueSource = (fp.hasMoreArguments()) ? fp.parseValueSource() : new LiteralValueSource("");
+
+          if (fp.hasMoreArguments()) {
+              // functions are not supported with strings
+              throw new SolrException(SolrException.ErrorCode.BAD_REQUEST, "Invalid payload function: " + fp.parseArg());
```
mkhludnev: Please explain why it;s invalid in exception message. It seems like in/mas isn't supported so far.

solr/core/src/test/org/apache/solr/search/function/TestFunctionQuery.java
```diff
@@ -569,6 +570,57 @@ public void testPayloadFunction() {
...
+        "rows", "1",
+        "sort", "payload(vals_dpss,SORT) asc",
+        "fl", "*,score,field_test:payload(vals_dpss,SORT)"), "//str[@name='field_test']=''");
+
+    assertQ(req("q", "*:*",
+        "rows", "1",
+        "sort", "payload(vals_dpss,SORT) desc",
+        "fl", "*,score,field_test:payload(vals_dpss,SORT)"), "//str[@name='field_test']='C'");
+
+//    assertQ(req("q", "{!func}payload(vals_dpss,A):USD",
```
mkhludnev: There should be tests for 

1. default, for 
1. absent vals as well,
2. also adding fqs let to cover advance() codes

PR #1037, opened 2019-11-25
LUCENE-9062: QueryVisitor.consumeTermsMatching

romseygeek:
This commit adds a `consumeTermsMatching()` method to QueryVisitor, allowing
queries that match against a class of terms to report this back to the visitor.  It also
changes highlighting code to use this new method, replacing the current implementation
via `instanceof` checks.


romseygeek:
I've gone back and forth on the best implementation here, particularly in terms of what we should actually be passing back to `consumeTermsMatching`.  I ended up going with a `BytesRunAutomaton` as it most closely goes with `consumeTerms` which deals with byte-oriented terms.  However, as highlighting is currently the only consumer of this API and it needs char-oriented matchers I can see the arguments the other way.

There's also a nocommit around what we should pass back in the case of a `AUTOMATON_TYPE.ALL` matcher - at the moment I'm just calling `visitLeaf` but we perhaps need another method, something like `consumeAnyTerm`?

I we do go with something like this, we'll probably want to re-work the unified highlighter to merge the visitor that extracts terms and the one that extracts automata, as there is some work that gets done twice currently (particularly for FuzzyQuery)

lucene/core/src/java/org/apache/lucene/search/FuzzyQuery.java
```diff
@@ -156,9 +159,14 @@ public Automaton toAutomaton() {
...
   @Override
   public void visit(QueryVisitor visitor) {
-    // TODO find some way of consuming Automata
-    if (visitor.acceptField(term.field())) {
-      visitor.visitLeaf(this);
+    if (visitor.acceptField(field)) {
+      if (maxEdits == 0 || prefixLength >= term.text().length()) {
+        visitor.consumeTerms(this, term);
+      } else {
+        CompiledAutomaton a = new CompiledAutomaton(toAutomaton(), null, true,
```
mkhludnev: May this instating be a little bit expensive?

romseygeek: It is, which is why I'm still not convinced this is the correct API.  Our whole setup around automata and utf8 vs utf32 is confusing and hard to use unfortunately

lucene/highlighter/src/java/org/apache/lucene/search/uhighlight/MemoryIndexOffsetStrategy.java
```diff
@@ -72,9 +73,6 @@ private static CharacterRunAutomaton buildCombinedAutomaton(UHComponents compone
           MultiTermHighlighting.extractAutomata(spanQuery, components.getFieldMatcher(), true));//true==lookInSpan
     }
 
-    if (allAutomata.size() == 1) {
```
dsmiley: Can't you make this method, buildCombinedAutomaton, return a CharArrayMatcher now?  Then you could retain this optimization for a single automata, and also simplify the last anonymous implementation to *not* be a CharacterRunAutomaton.

lucene/highlighter/src/java/org/apache/lucene/search/uhighlight/MultiTermHighlighting.java
```diff
@@ -86,85 +87,10 @@ public QueryVisitor getSubVisitor(BooleanClause.Occur occur, Query parent) {
...
-      }
+    public void consumeTermsMatching(Query query, String field, ByteRunAutomaton automaton) {
+      runAutomata.add(automaton.asCharArrayMatcher(query.toString()));
     }
 
   }
 
-  private static CharacterRunAutomaton binaryToCharRunAutomaton(Automaton binaryAutomaton, String description) {
-    return new CharacterRunAutomaton(Automata.makeEmpty()) { // empty here is bogus just to satisfy API
-      //   TODO can we get access to the aq.compiledAutomaton.runAutomaton ?
```
dsmiley: From what I see, and I'm not certain, you've succeeded in re-using the underlying AutomatonQuery's CompiledAutomaton.runAutomaton :-)

lucene/core/src/java/org/apache/lucene/util/automaton/CharArrayMatcher.java
```diff
@@ -0,0 +1,24 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.lucene.util.automaton;
+
+public interface CharArrayMatcher {
```
dsmiley: Lets also define getLabel() that we should call instead of toString()?  That way we can make it less hacky to rely on toString() being what we want.  We could offer a default implementation of toString() that calls getLabel()

lucene/core/src/java/org/apache/lucene/search/FuzzyQuery.java
```diff
@@ -156,9 +160,13 @@ public Automaton toAutomaton() {
...
   public void visit(QueryVisitor visitor) {
-    // TODO find some way of consuming Automata
-    if (visitor.acceptField(term.field())) {
-      visitor.visitLeaf(this);
+    if (visitor.acceptField(field)) {
+      if (maxEdits == 0 || prefixLength >= term.text().length()) {
+        visitor.consumeTerms(this, term);
+      } else {
+        visitor.consumeTermsMatching(this, field,
+            new ByteRunAutomaton(toAutomaton(), false, Operations.DEFAULT_MAX_DETERMINIZED_STATES));
```
dsmiley: At least we should acknowledge in a comment that this is expensive :-/
At least I could imagine a user solving this by subclassing FuzzyQuery to override toAutomaton and using a cache.

lucene/highlighter/src/java/org/apache/lucene/search/uhighlight/FieldOffsetStrategy.java
```diff
@@ -192,7 +193,7 @@ protected void createOffsetsEnumsForAutomata(Terms termsIndex, int doc, List<Off
     }
 
     for (int i = 0; i < automata.length; i++) {
-      CharacterRunAutomaton automaton = automata[i];
+      CharArrayMatcher automaton = automata[i];
```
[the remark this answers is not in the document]
romseygeek: I added a new highlighter class, `LabelledCharArrayMatcher`, with a static `wrap` method taking a plain matcher, and cut things over to use that.  So far it's only the unified highlighter that actually needs the label, so I think this puts concerns in the right place.

lucene/core/src/java/org/apache/lucene/util/automaton/CharArrayMatcher.java
```diff
@@ -0,0 +1,24 @@
...
+
+  boolean run(char[] s, int offset, int length);
```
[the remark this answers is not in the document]
romseygeek: ++ I added the CharsRef variant as a default method, although nothing's actually using it at the moment.

lucene/core/src/java/org/apache/lucene/util/automaton/ByteRunAutomaton.java
```diff
@@ -44,4 +46,45 @@ public boolean run(byte[] s, int offset, int length) {
     }
     return accept[p];
   }
+
+  public CharArrayMatcher asCharArrayMatcher(String label) {
```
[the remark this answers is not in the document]
romseygeek: I added javadoc

lucene/core/src/java/org/apache/lucene/search/QueryVisitor.java
```diff
@@ -37,8 +39,9 @@
    */
   public void consumeTerms(Query query, Term... terms) { }
 
-  // TODO it would be nice to have a way to consume 'classes' of Terms from
-  // things like AutomatonQuery
+  public void consumeTermsMatching(Query query, String field, ByteRunAutomaton automaton) {
+    visitLeaf(query); // default impl for BWC
```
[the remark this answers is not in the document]
romseygeek: done

lucene/highlighter/src/java/org/apache/lucene/search/uhighlight/MultiTermHighlighting.java
```diff
@@ -86,85 +87,10 @@ public QueryVisitor getSubVisitor(BooleanClause.Occur occur, Query parent) {
     }
 
     @Override
-    public void visitLeaf(Query query) {
```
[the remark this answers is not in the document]
romseygeek: :)

lucene/highlighter/src/java/org/apache/lucene/search/uhighlight/MemoryIndexOffsetStrategy.java
romseygeek: done

lucene/core/src/java/org/apache/lucene/util/automaton/CharArrayMatcher.java
romseygeek: This is now encapsulated in `LabelledCharArrayMatcher`

lucene/core/src/java/org/apache/lucene/search/FuzzyQuery.java
romseygeek: Reading back through this, it looks as though FuzzyQuery rebuilds its automata on every rewrite, which is pretty wasteful - I'll open a new issue to follow this up.

lucene/core/src/java/org/apache/lucene/util/automaton/CharArrayMatcher.java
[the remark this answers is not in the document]
dsmiley: If we're not going to call this thing an automata anymore and instead it's a "matcher" (a predicate), maybe the method name here should be test() to be inline with Predicate?  Or matches() to be in line with regex Matcher?

lucene/core/src/java/org/apache/lucene/util/automaton/ByteRunAutomaton.java
```diff
@@ -47,44 +47,39 @@ public boolean run(byte[] s, int offset, int length) {
...
+        } else if (code < 0x800) {
+          b = (0xC0 | (code >> 6));
+          state = step(state, b);
+          if (state == -1) return false;
+          b = (0x80 | (code & 0x3F));
+          state = step(state, b);
+          if (state == -1) return false;
+        } else {
+          // more complex
+          byte[] utf8Bytes = new byte[4 * (maxIdx - i)];
```
dsmiley: We should probably create/use this once instead of for each such char?  I know it was like this before; you needn't touch it.

lucene/highlighter/src/java/org/apache/lucene/search/uhighlight/FieldOffsetStrategy.java
```diff
@@ -193,13 +193,13 @@ protected void createOffsetsEnumsForAutomata(Terms termsIndex, int doc, List<Off
...
     for (int i = 0; i < automata.length; i++) {
-      CharArrayMatcher automaton = automata[i];
+      LabelledCharArrayMatcher automaton = automata[i];
       List<PostingsEnum> postingsEnums = automataPostings.get(i);
       if (postingsEnums.isEmpty()) {
         continue;
       }
-      // Build one OffsetsEnum exposing the automata.toString as the term, and the sum of freq
-      BytesRef wildcardTerm = new BytesRef(automaton.toString());
+      // Build one OffsetsEnum exposing the automaton label as the term, and the sum of freq
```
dsmiley: I wonder if we should stop calling these things "automata" and now call them "matcher"?  even if just in some method like perhaps this one.  Automata is a bit of an implementation detail.

lucene/core/src/java/org/apache/lucene/util/automaton/CharArrayMatcher.java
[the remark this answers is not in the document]
dsmiley: Shouldn't the CharsRef overloaded version be used at least in one place -- FieldOffsetStrategy.createOffsetsEnumsForAutomata ?

lucene/highlighter/src/java/org/apache/lucene/search/uhighlight/LabelledCharArrayMatcher.java
```diff
@@ -19,10 +19,19 @@
 
 import org.apache.lucene.util.automaton.CharArrayMatcher;
 
+/**
+ * Associates a label with a CharArrayMatcher to distinguish different sources for terms in highlighting
+ */
```
dsmiley: @lucene.internal

lucene/core/src/java/org/apache/lucene/search/FuzzyQuery.java
romseygeek: I opened https://issues.apache.org/jira/browse/LUCENE-9068

lucene/core/src/java/org/apache/lucene/util/automaton/ByteRunAutomaton.java
romseygeek: ++ let's do that in a followup

dsmiley:
With this change, it's interesting to see CharacterRunAutomaton will be even more niche... relegated to use with RegExp in non-test uses; no longer in the highlighter.  Maybe CharacterRunAutomaton could use a bit more clarification that the Automata it operates on is a UTF32 code point Automaton, not some other type (not bytes/256).  Furthermore Regexp.toAutomaton could document that.  Our Automata is sadly untyped but they are not equivalent!

romseygeek:
I've pushed a change re-organising things again.  `CharArrayMatcher` is now a highlighter class, and `CharacterRunAutomaton` no longer implements it, instead we pass around method references.  I've changed the relevant method to be `match` instead of `run` as well.

lucene/highlighter/src/java/org/apache/lucene/search/uhighlight/CharArrayMatcher.java
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * Matches a character array
+ */
+public interface CharArrayMatcher {
+
+  /**
+   * Return {@code true} if the passed-in character array matches
+   */
+  boolean match(char[] s, int offset, int length);
```
dsmiley: Shouldn't the method be named "matches" ?  Not important any way; at least it's more meaningful than "run".

lucene/highlighter/src/java/org/apache/lucene/search/uhighlight/CharArrayMatcher.java
```diff
@@ -0,0 +1,49 @@
...
+import java.util.List;
+
+import org.apache.lucene.util.BytesRef;
+import org.apache.lucene.util.CharsRef;
+import org.apache.lucene.util.automaton.Automata;
+import org.apache.lucene.util.automaton.CharacterRunAutomaton;
+
+/**
+ * Matches a character array
+ */
```
dsmiley: @lucene.internal

lucene/highlighter/src/java/org/apache/lucene/search/uhighlight/LabelledCharArrayMatcher.java
```diff
@@ -0,0 +1,88 @@
...
+
+import org.apache.lucene.util.UnicodeUtil;
+import org.apache.lucene.util.automaton.ByteRunAutomaton;
+
+/**
+ * Associates a label with a CharArrayMatcher to distinguish different sources for terms in highlighting
+ *
+ * @lucene.internal
+ */
+public interface LabelledCharArrayMatcher extends CharArrayMatcher {
```
dsmiley: If CharArrayMatcher is now specific to the UnifiedHighlighter, then why have a "LabelledCharArrayMatcher" separate?  Lets just have CharArrayMatcher that has a Label?

lucene/queries/src/java/org/apache/lucene/queries/intervals/MultiTermIntervalsSource.java
```diff
@@ -96,7 +96,7 @@ public MatchesIterator matches(String field, LeafReaderContext ctx, int doc) thr
 
   @Override
   public void visit(String field, QueryVisitor visitor) {
-    visitor.visitLeaf(new IntervalQuery(field, this));
+    automaton.visit(visitor, new IntervalQuery(field, this), field);
```
dsmiley: eh; what's this about?

lucene/highlighter/src/java/org/apache/lucene/search/uhighlight/LabelledCharArrayMatcher.java
romseygeek: The advantage of having both is that `CharArrayMatcher` is a function interface, so we can pass `CharacterRunAutomaton::run` as one

lucene/queries/src/java/org/apache/lucene/queries/intervals/MultiTermIntervalsSource.java
romseygeek: It's the intervals equivalent of AutomatonQuery - it means that Intervals.multiterm() should start working with the highlighter

lucene/highlighter/src/java/org/apache/lucene/search/uhighlight/LabelledCharArrayMatcher.java
dsmiley: An extra class just so that we have a functional interface in one spot internally?  I wouldn't make that trade-off but I leave this to you.

romseygeek: There's also the fact that not every matcher requires a label - ones created from a set of terms, for example, in the MemoryIndexOffsetStrategy.  We can probably clear this up later, but I'd rather keep things as they are for now and get this in.

lucene/highlighter/src/java/org/apache/lucene/search/uhighlight/CharArrayMatcher.java
romseygeek: done

PR #1040, opened 2019-11-26
LUCENE-9067: Polygon2D#contains is now thread safe

iverase:
This PR adds an early termination exception for the case a point lies on the boundary of an edge tree.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/geo/EdgeTree.java
+++ b/lucene/core/src/java/org/apache/lucene/geo/EdgeTree.java
@@ -76,7 +76,7 @@ public class EdgeTree {
 
   /**
    * Returns true if the point crosses this edge subtree an odd number of times. It
-   * throws a {@link @ContainBoundaryException} if the point is on an edge.
+   * throws a {@link ContainsBoundaryException} if the point is on an edge.
    * <p>
    * See <a href="https://www.ecse.rpi.edu/~wrf/Research/Short_Notes/pnpoly.html">
    * https://www.ecse.rpi.edu/~wrf/Research/Short_Notes/pnpoly.html</a> for more information.
--- a/lucene/core/src/java/org/apache/lucene/geo/EdgeTree.java
+++ b/lucene/core/src/java/org/apache/lucene/geo/EdgeTree.java
@@ -317,7 +317,7 @@ public class EdgeTree {
   private static class ContainsBoundaryException extends RuntimeException {
 
     /** Sole constructor. */
-  public ContainsBoundaryException() {
+    ContainsBoundaryException() {
       super();
     }
   }
--- a/lucene/core/src/java/org/apache/lucene/geo/EdgeTree.java
+++ b/lucene/core/src/java/org/apache/lucene/geo/EdgeTree.java
@@ -17,7 +17,6 @@
 package org.apache.lucene.geo;
 
 import java.util.Arrays;
-import java.util.concurrent.atomic.AtomicBoolean;
 
 import static org.apache.lucene.geo.GeoUtils.lineCrossesLine;
 import static org.apache.lucene.geo.GeoUtils.lineCrossesLineWithBoundary;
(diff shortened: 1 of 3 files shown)


iverase:
@jpountz, we can try to use a bye as result of `containsPnPoly` and hold in bits the information.

I updated the PR with this approach.

lucene/core/src/java/org/apache/lucene/geo/EdgeTree.java
```diff
@@ -90,30 +102,34 @@
...
+          return ON_EDGE;
         } else if (this.y1 > y != this.y2 > y) {
-          res = x < (this.x2 - this.x1) * (y - this.y1) / (this.y2 - this.y1) + this.x1;
+          res = x < (this.x2 - this.x1) * (y - this.y1) / (this.y2 - this.y1) + this.x1 ? TRUE : FALSE;
         }
       }
       if (this.left != null) {
-        res ^= left.contains(x, y, isOnEdge);
+        res ^= left.containsPnPoly(x, y);
+        if (((res >> 1) & 1) == 1) {
```
jpountz:
```suggestion
        if ((res & 0x02) == 0x02) {
```

iverase:
I am unsure how to apply bitwise logic using a java enum. Maybe I am missing something but creating an enum and adding my own bitwise logic seems inefficient:

```
||Approach||Shape||M hits/sec      ||QPS            ||Hit count      ||
|points|polyRussia|12.67|12.76|-1%|3.61|3.64|-1%|3508846|3508846| 0%|
|points|poly 10|58.95|57.79| 2%|37.28|36.54| 2%|355809475|355809475| 0%|
|points|polyMedium|7.55|7.91|-5%|92.54|96.96|-5%|2693559|2693559| 0%|
|shapes|polyRussia|6.78|7.00|-3%|1.93|2.00|-3%|3508846|3508846| 0%|
|shapes|poly 10|30.97|32.91|-6%|19.58|20.81|-6%|355809475|355809475| 0%|
|shapes|polyMedium|2.63|2.74|-4%|32.20|33.60|-4%|2693559|2693559| 0%|
```

On the other hand using a byte seems to speed up the computations:

```
||Approach||Shape||M hits/sec      ||QPS            ||Hit count      ||
|points|polyRussia|14.68|13.11|12%|4.18|3.74|12%|3508846|3508846| 0%|
|points|poly 10|60.77|59.54| 2%|38.43|37.65| 2%|355809475|355809475| 0%|
|points|polyMedium|8.35|8.19| 2%|102.29|100.29| 2%|2693559|2693559| 0%|
|shapes|polyRussia|7.33|5.36|37%|2.09|1.53|37%|3508846|3508846| 0%|
|shapes|poly 10|32.80|33.14|-1%|20.74|20.95|-1%|355809475|355809475| 0%|
|shapes|polyMedium|3.15|2.87|10%|38.62|35.14|10%|2693559|2693559| 0%|
```

iverase:
I have rerun the performance tests for sanity check and using an enum instead of a byte slows down things. 

I would stick to this approach for performance reasons even though it makes the code a bit more difficult to read. Still the logic is inside of the edge tree and not exposed to other objects.

lucene/core/src/java/org/apache/lucene/geo/EdgeTree.java
```diff
@@ -50,6 +49,10 @@
     EdgeTree left;
     /** right child edge, or null */
     EdgeTree right;
+    /** helper bytes to signal if a point is on an edge, it is within the edge tree or disjoint */
+    final private byte FALSE = 0x00;
```
jpountz: can you make them static?

lucene/core/src/java/org/apache/lucene/geo/EdgeTree.java
```diff
@@ -90,30 +103,34 @@
...
       if (this.right != null && y >= this.low) {
-        res ^= right.contains(x, y, isOnEdge);
+        res ^= right.containsPnPoly(x, y);
+        if ((res & 0x02) == 0x02) {
+          return ON_EDGE;
+        }
       }
     }
-    return isOnEdge.get() || res;
+    return res;
```
jpountz: let's assert that the return value is in the expected range for sanity?
```suggestion
    assert res >= FALSE && res <= EN_EDGE;
    return res;
```

PR #1042, opened 2019-11-26
LUCENE-9068: Build FuzzyQuery automata up-front

romseygeek:
FuzzyTermsEnum can now either take an array of compiled automata, and 
an AttributeSource, to be used across multiple segments (eg during 
FuzzyQuery rewrite); or it can take a term, edit distance, prefix and transition
boolean and build the automata itself if only being used once (eg for fuzzy
nearest neighbour calculations).

Rather than interact via attribute sources and specialized attributes, users of
FuzzyTermsEnum can get the boost and set minimum competitive boosts
directly on the enum.


lucene/core/src/java/org/apache/lucene/search/FuzzyTermsEnum.java
```diff
@@ -58,30 +60,42 @@
...
+   * length <code>prefixLength</code> with <code>term</code> and which have at most {@code maxEdits} edits.
+   * <p>
+   * After calling the constructor the enumeration is already pointing to the first
+   * valid term if such a term exists.
+   *
+   * @param terms Delivers terms.
+   * @param term Pattern term.
+   * @param maxEdits Maximum edit distance.
+   * @param prefixLength the length of the required common prefix
+   * @param transitions whether transitions should count as a single edit
```
madrob: nit: why was this renamed from transpositions?

lucene/core/src/java/org/apache/lucene/search/FuzzyTermsEnum.java
```diff
@@ -344,67 +363,4 @@ public BytesRef term() throws IOException {
...
-      }
-    }
-
-    @Override
-    public boolean equals(Object other) {
-      if (this == other)
-        return true;
-      if (!(other instanceof LevenshteinAutomataAttributeImpl))
-        return false;
-      return Arrays.equals(automata, ((LevenshteinAutomataAttributeImpl) other).automata);
```
madrob: nit: we have some unused imports after this deletion.

lucene/core/src/java/org/apache/lucene/search/FuzzyTermsEnum.java
```diff
@@ -91,58 +105,49 @@
...
+      final int maxEdits, CompiledAutomaton[] automata) throws IOException {
+
     this.maxEdits = maxEdits;
     this.terms = terms;
     this.term = term;
+    this.atts = atts;
     
     // convert the string into a utf32 int[] representation for fast comparisons
-    this.termText = stringToUTF32(term.text());
+    int[] termText = stringToUTF32(term.text());
```
madrob: This seems like a potentially expensive conversion to have to do just to get the length when we already have the automata (and we might have already done it once, if this constructor was called via the other one on L96.)

lucene/core/src/java/org/apache/lucene/search/FuzzyTermsEnum.java
romseygeek: Because I can't read or type, apparently... have fixed

lucene/core/src/java/org/apache/lucene/search/FuzzyTermsEnum.java
romseygeek: fixed

lucene/core/src/java/org/apache/lucene/search/FuzzyTermsEnum.java
romseygeek: I've rearranged things a bit now so that we don't call this twice

romseygeek:
Thanks for the review @madrob, I've pushed an update based on your feedback and merged in the exception handling changes.  I've also done a bit more cleanup - `visit` for example now just delegates to the pre-built automata rather than trying to rebuild things.

lucene/core/src/java/org/apache/lucene/search/FuzzyTermsEnum.java
```diff
@@ -92,76 +105,62 @@
...
-   * Builds a binary Automaton to match a fuzzy term
-   * @param text            the term to match
-   * @param prefixLength    length of a required common prefix
-   * @param transpositions  {@code true} if transpositions should count as a single edit
-   * @param maxEdits        the maximum edit distance of matching terms
+   * Sets the maximum non-competitive boost, which may allow switching to a
+   * lower max-edit automaton at run time
+   */
+  public void setMaxNonCompetitiveBoost(float boost) {
+    this.maxBoostAtt.setMaxNonCompetitiveBoost(boost);
```
madrob: Does this need to call `bottomChanged`?

lucene/core/src/test/org/apache/lucene/util/TestRamUsageEstimator.java
```diff
@@ -32,6 +32,7 @@
 import org.apache.lucene.search.BooleanQuery;
 import org.apache.lucene.search.DisjunctionMaxQuery;
 import org.apache.lucene.search.FuzzyQuery;
```
madrob: This is an unused import failing recommit still.

lucene/core/src/java/org/apache/lucene/search/FuzzyTermsEnum.java
madrob: Thinking on this further... would it make sense to somehow check the terms before building all of the automata, and then possibly skip building the ed=2 case to begin with? Maybe checking the terms is more expensive than building the automaton, so we only care about saving time on how much we use the automaton at runtime? I'm not sure here and haven't done the profiling necessary to be confident.

romseygeek: > Does this need to call bottomChanged?

No, that's an internal call only, and it's delayed anyway so as to preserve term stats after calling `next()`.

> Maybe checking the terms is more expensive than building the automaton

I'd be surprised, unless the input term for the automaton is very long, but it's probably worth profiling and checking this in a followup.

madrob:
LGTM after precommit comes back happy.

PR #1043, opened 2019-11-28
LUCENE-9071: Speed up BM25 scores.

jpountz:
The approach of this change is to try to detect when computing the tf component
of the BM25 score would return the same result in float and double space. In
order to make this condition likely, it trades little precision from the norm
by rounding to the nearest float that has 10 trailing zeros.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/similarities/BM25Similarity.java
+++ b/lucene/core/src/java/org/apache/lucene/search/similarities/BM25Similarity.java
@@ -221,8 +221,15 @@ public class BM25Similarity extends Similarity {
 
     @Override
     public float score(float freq, long encodedNorm) {
-      // In order to guarantee monotonicity with both freq and norm, we rewrite
-      // freq / (freq + norm) to 1 - 1 / (1 + freq / norm).
+      // In order to guarantee monotonicity with both freq and norm without
+      // promoting to doubles, we rewrite freq / (freq + norm) to
+      // 1 - 1 / (1 + freq * 1/norm).
+      // freq * 1/norm is guaranteed to be monotonic for both freq and norm due
+      // to the fact that multiplication and division round to the nearest
+      // float. And then monotonicity is preserved through composition via
+      // x -> 1 + x and x -> 1 - 1/x.
+      // Finally we expand weight * (1 - 1 / (1 + freq * 1/norm)) to
+      // weight - weight / (1 + freq * 1/norm), which runs slightly faster.
       float normInverse = cache[((byte) encodedNorm) & 0xFF];
       return weight - weight / (1f + freq * normInverse);
     }
--- a/lucene/core/src/java/org/apache/lucene/search/similarities/BM25Similarity.java
+++ b/lucene/core/src/java/org/apache/lucene/search/similarities/BM25Similarity.java
@@ -26,6 +26,7 @@ import org.apache.lucene.search.CollectionStatistics;
 import org.apache.lucene.search.Explanation;
 import org.apache.lucene.search.TermStatistics;
 import org.apache.lucene.util.BytesRef;
+import org.apache.lucene.util.MathUtil;
 import org.apache.lucene.util.SmallFloat;
 
 /**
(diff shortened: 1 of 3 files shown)


lucene/core/src/java/org/apache/lucene/search/similarities/BM25Similarity.java
```diff
@@ -221,8 +251,8 @@ public final SimScorer scorer(float boost, CollectionStatistics collectionStats,
 
     @Override
     public float score(float freq, long encodedNorm) {
-      double norm = cache[((byte) encodedNorm) & 0xFF];
-      return weight * (float) (freq / (freq + norm));
+      float norm = cache[((byte) encodedNorm) & 0xFF];
+      return weight * tf(freq, norm);
```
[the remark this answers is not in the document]
rmuir: the crazy casting is not an optimization but instead to ensure that scores dont go backwards with increasing tf, decreasing norm, etc. it has to behave "semimonotonic" (like described in java Math api docs) so that wand type optimizations are correct. so if you changed the api to be all doubles, it would mean we'd have to use BigDecimal which would br slower.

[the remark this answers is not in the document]
bruno-roustant: Thanks, very good point.

lucene/core/src/java/org/apache/lucene/util/MathUtil.java
```diff
@@ -165,4 +165,32 @@ public static double sumRelativeErrorBound(int numValues) {
...
+    // that has the maximum exponent plus one as an exponent, then the sum is
+    // accurate.
+    int expDelta = maxBiasedExp - minBiasedExp;
+    final boolean accurate = (maxBits & 0x01) == 0 &&
+        // beware of overflows, a biased exponent of 254 is the greatest exponent that is not Infinity or NaN
+        maxBiasedExp < 254 &&
+        expDelta <= 22 &&
+        Integer.numberOfTrailingZeros(minBits) > expDelta;
+    return accurate;
+  }
```
rmuir: i have a tough time believing its worth it to do all this stuff to save a cast to float. Are you sure benchmarks are working correctly? I think we should also consider complexity as well...

bruno-roustant: +1, even if I appreciate the technique to determine if a float sum is accurate.

jpountz:
I understand it's surprising that this approach worked, but I've been able to reproduce it consistently with luceneutil, including by swapping the base and patch in order to avoid any bias due to which competitor runs first. I also confirmed it via microbenchmarks. I wonder whether there is something special about my CPU.

I've been exploring other ideas and found an approach that yields almost exactly the same speedups but is simpler. The way it works is by rewriting the BM25 formula from `weight * (freq / (freq + norm))` to `weight - weight / (1 + freq * (1/norm))`. This formula doesn't need double promotion in order to guarantee monotonicity on `freq` or `norm`.

bruno-roustant:
For my personal knowledge, the second formula doesn't need double promotion because it uses float multiplication instead of addition? There might be a loss in precision but the monotonicity is guaranteed? Would you have some doc about this subject to share?

lucene/core/src/java/org/apache/lucene/search/similarities/BM25Similarity.java
```diff
@@ -221,17 +221,22 @@ public final SimScorer scorer(float boost, CollectionStatistics collectionStats,
 
     @Override
     public float score(float freq, long encodedNorm) {
-      double norm = cache[((byte) encodedNorm) & 0xFF];
-      return weight * (float) (freq / (freq + norm));
+      // In order to guarantee monotonicity with both freq and norm, we rewrite
```
bruno-roustant: It is worth mentioning also "without promotion to double".

jpountz:
@bruno-roustant  It's not about multiplication vs. addition but more about preventing rounding errors from being in the way of monotonicity. The problem with `freq / (freq + norm)` is when `freq+norm` rounds down for a given value of `freq` and up for the next value. Then the score may decrease because the denominator suddenly grows slower than the numerator. Rewriting to `1-1/(1+freq*1/norm)` helps notably because `freq` only occurs once in the formula. `freq * 1/norm` is guaranteed to be non-decreasing when freq increases or when norm decreases given that multiplication and division round to the nearest float, and this property is maintained through composition via `x->1+x` and then `x ->1-1/x`. Rounding errors can cause two consecutive frequencies to return the same score, but they can't break monotonicity in that case. `1/(1+norm/freq)` would work too but it performed slower in my testing.

jpountz:
Thanks @bruno-roustant and @rmuir.

PR #1045, opened 2019-11-29
LUCENE-9072: Find matching terms from Matches API

romseygeek:
The Matches API tells us which queries have matched, and what positions they have 
matched at, but not necessarily which actual terms matched. This commit adds
a `getMatchingTerms` method to `Matches`, allowing users to build results pages that 
indicate which specific query terms are present or absent in each top-k hit.


mkhludnev:
notice precommit failure

lucene/core/src/java/org/apache/lucene/search/Matches.java
```diff
@@ -44,4 +47,9 @@
    */
   Collection<Matches> getSubMatches();
 
+  /**
+   * Find all matching terms
+   */
+  void getMatchingTerms(Consumer<Term> termsConsumer) throws IOException;
+
```
pipcet: Wouldn't this make more sense as a method of `MatchesIterator`?

In conjunction with some changes to `PhraseWeight` and `SloppyPhraseMatcher`, that would allow us to accurately capture reordered phrases, which I think `getMatchingTerms` would be a good API for.

My proposed use case is to modify `AnalyzingInfixSuggester` to use a `PhraseQuery` (with slop) and expose this data in order to suggest a reordered phrase to the user without including the entire document in the response.

romseygeek: Can you not already do this with the plain `MatchesIterator`?  It will give you the start offset and end offset of matching sloppy phrases, and you can use those to pull phrases from the source.

PR #1049, opened 2019-12-02
LUCENE-9074: Slice Allocation Circuit Breakers in IndexSearcher

atris:
This commit introduces accounting for the queue length of the ExecutorService
being used to perform concurrent search when allocating slices for an
IndexSearcher. This commit also introduces an abstraction to define
custom parameters for sealing bulkheads under heavy node stress
to allow better predictable behaviour of latencies under varying stress


lucene/core/src/java/org/apache/lucene/search/IndexSearcher.java
```diff
@@ -193,6 +196,13 @@ public IndexSearcher(IndexReader r, Executor executor) {
     this(r.getContext(), executor);
   }
 
+  /** Same as above with additional Circuit Breaker passed in
+   *
+   * @lucene.experimental */
+  public IndexSearcher(IndexReader r, Executor executor, SliceAllocationCircuitBreaker sliceAllocationCircuitBreaker) {
```
mikemccand: I think it'd be cleaner if we combined the `Executor` and this new `SliceAllocationCircuitBreaker` into a single new interface that could both provide the `Executor` and the red-line logic?

lucene/core/src/java/org/apache/lucene/search/QueueSizeBasedCircuitBreaker.java
```diff
@@ -0,0 +1,46 @@
...
+  private final Executor executor;
+  private static final int NUMBER_OF_PROCESSORS = Runtime.getRuntime().availableProcessors();
+  private static final double LIMITING_FACTOR = 1.5;
+
+  public QueueSizeBasedCircuitBreaker(Executor executor) {
+    this.executor = executor;
+  }
+
+  @Override
+  public boolean shouldProceed() {
```
mikemccand: Maybe `isTooBusy` instead?

lucene/core/src/java/org/apache/lucene/search/IndexSearcher.java
```diff
@@ -316,8 +342,8 @@ public QueryCachingPolicy getQueryCachingPolicy() {
     long docSum = 0;
     List<LeafReaderContext> group = null;
     for (LeafReaderContext ctx : sortedLeaves) {
-      if (ctx.reader().maxDoc() > maxDocsPerSlice) {
-        assert group == null;
+      if (ctx.reader().maxDoc() > maxDocsPerSlice &&
+          (isSliceAllocationCBNull || sliceAllocationCircuitBreaker.shouldProceed())) {
```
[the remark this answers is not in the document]
atris: I felt that this was more readable since the actual check is a mouthful but do not have a strong preference. Fixed, thanks

lucene/core/src/java/org/apache/lucene/search/QueueSizeBasedCircuitBreaker.java
atris: Since this is modelling around the concept of Circuit Breakers, I wonder if `hasCircuitBreakerTriggered` is a better name?

atris:
Raised another iteration -- please see and let me know your comments

lucene/core/src/java/org/apache/lucene/search/SliceExecutionControlPlane.java
```diff
@@ -0,0 +1,54 @@
...
+ * along with a circuit breaker condition which will control whether new threads will be
+ * created or not.
+ */
+public interface SliceExecutionControlPlane {
+
+  /**
+   * Return true if the circuit breaker condition has triggered,
+   * false otherwise
+   */
+  boolean hasCircuitBreakerTriggered();
```
jpountz: Maybe we should move the logic of assigning tasks to threads to this class to give it more flexibility instead of just exposing whether the pool is running over capacity. I'm thinking of something that could look like `void invokeAll(Collection<Runnable> tasks)` (similar to ForkJoinPool), which could then decide to merge some runnables together, run some of them on the current thread, etc.

atris: Would that also include creating the actual slices? We would need this class to be involved in that process as well since it can control the number of slices being created. Or maybe we create the slices as we do today, but can run multiple slices on caller thread when things are hot?

jpountz: I was thinking the latter indeed, having more and more slices in a single task as the number of entries in the queue increases.

atris:
Superseded by #1214

PR #1050, opened 2019-12-02
SOLR-13990: Switch out woodstox-core-asl with aalto-xml and upgrade woodstox stax-2 API

anshumg:
aalto-xml is an ultra-high performance next generation Stax XML processor implementation, implementing both basic Stax API (javax.xml.stream) and Stax2 API extension (org.codehaus.woodstox.stax2). In addition, it also implements SAX2 API.


solr/core/src/java/org/apache/solr/handler/tagger/XmlOffsetCorrector.java
```diff
@@ -47,19 +45,13 @@
...
     // note: similar code in Solr's EmptyEntityResolver
-    XML_INPUT_FACTORY = new WstxInputFactory();
-    XML_INPUT_FACTORY.setXMLResolver(new XMLResolver() {
-      @Override
-      public InputStream resolveEntity(String publicId, String systemId, String baseURI, String namespace) {
-        return ClosedInputStream.CLOSED_INPUT_STREAM;
-      }
-    });
+    XML_INPUT_FACTORY = XMLInputFactory2.newFactory();
+    XML_INPUT_FACTORY.setXMLResolver((publicId, systemId, baseURI, namespace) -> ClosedInputStream.CLOSED_INPUT_STREAM);
```
dsmiley: This is a slight bit of tech debt; just use Solr's EmptyEntityResolver and then also remove the comment.  This code pre-dated inclusion of Solr; it used to be a side project.

lucene/ivy-versions.properties
```diff
@@ -257,9 +259,7 @@ org.codehaus.janino.version = 3.0.9
 /org.codehaus.janino/commons-compiler = ${org.codehaus.janino.version}
 /org.codehaus.janino/janino = ${org.codehaus.janino.version}
 
-/org.codehaus.woodstox/stax2-api = 3.1.4
-/org.codehaus.woodstox/woodstox-core-asl = 4.4.1
-
+/org.codehaus.woodstox/stax2-api = 4.2
```
dsmiley: Weird that the API comes from woodstox but now our impl from "aalto-xml"?  Just want to confirm; it's odd.

anshumg: Yes, it's odd but intended/expected. I tried digging into this more too as I inherited this from Mark.

anshumg:
I just found this : [https://issues.apache.org/jira/browse/HADOOP-14501](https://issues.apache.org/jira/browse/HADOOP-14501)

I assume things have changed since 2017 so we are all good.

solr/core/src/java/org/apache/solr/handler/tagger/XmlOffsetCorrector.java
```diff
@@ -45,21 +43,12 @@
...
   static {
-    // note: similar code in Solr's EmptyEntityResolver
-    XML_INPUT_FACTORY = new WstxInputFactory();
-    XML_INPUT_FACTORY.setXMLResolver(new XMLResolver() {
-      @Override
-      public InputStream resolveEntity(String publicId, String systemId, String baseURI, String namespace) {
-        return ClosedInputStream.CLOSED_INPUT_STREAM;
-      }
-    });
-    // TODO disable DTD?
```
madrob: Is this TODO still applicable? Is this now disabled by default? If not, leave the TODO comment in.

PR #1051, opened 2019-12-02
SOLR-13992: Refactor code to have collection, shard name in Replica,Slice

noblepaul:


sigram:
Finally! Big +1.

(However, please check your IDE formatting rules - every time you add a method the opening `{` is glued to the preceding `)` - Apache style mandates a single space here. Similarly before the commas in argument lists).

solr/solrj/src/java/org/apache/solr/common/cloud/ZkStateReader.java
```diff
@@ -944,6 +944,11 @@ public Replica getLeader(String collection, String shard) {
     return null;
   }
 
+  public boolean isNodeLive(String node) {
```
nsoft: While this is potentially convenient it seems off topic for the PR/Issue. Also if it is kept, in the realm of taste/style I tend to not use get/set/is for things that are not properties of the object. maybe hasLiveNode(String node) thus someone using it might write `if (zkReader.hasLiveNode("foo")) ...` which reads quite nicely.

PR #1055, opened 2019-12-03
SOLR-13932 Review directory locking and Blob interactions

murblanc:
SOLR-13932: Review directory locking and Blob interactions

Take into account all local files in a local index to avoid conflict with files pulled from Blob.
Do not make local copy of directory for pushing to Blob.
Minor changes.

# Description

# Solution

# Tests

# Checklist

### DIFF
--- a/solr/core/src/java/org/apache/solr/store/blob/metadata/CorePushPull.java
+++ b/solr/core/src/java/org/apache/solr/store/blob/metadata/CorePushPull.java
@@ -129,20 +129,13 @@ public class CorePushPull {
           try {
 
             /*
-             * Removing from the core metadata the files that should no longer be there.
-             * 
-             * TODO
-             * This is a little confusing: This is equivalent to what we were doing in first-party 
-             * where the files to delete for a push were just the files that we determined were 
-             * missing locally but on blob (filesToPull) for a push. This operates on the assumption
-             * that the core locally was refreshed with what was in blob before this update (both in
-             * first party and Solr Cloud). 
-             * 
-             * SharedMetadataResolutionResult makes no distinction between what action is being taken 
-             * (push or pull) hence the confusing method naming but leaving this for now while we reach
-             * blob feature parity.
-             * 
-             * The deletion logic will move out of this class in the future and make this less confusing. 
+             * Removing from the core metadata the files that are stored on the blob store but no longer needed.
+             *
+             * When this method is executed, the content of the index on Blob is to be replaced with the local content.
+             * The assumption (or normal flow) is for local to refresh from Blob, update locally then push the
+             * changes to Blob.
+             * When merges happen locally the update has to mark old segment files for delete (and for example the
+             * previous "segments_N" is to be deleted as a new higher generation segment has been created).
              */
             for (BlobCoreMetadata.BlobFile d : resolvedMetadataResult.getFilesToDelete()) {
                 bcmBuilder.removeFile(d);
@@ -166,7 +159,7 @@ public class CorePushPull {
             }
             
             // Directory's javadoc says: "Java's i/o APIs not used directly, but rather all i/o is through this API"
-            // But this is untrue/totally false/misleading. SnapPuller has File all over.
+            // But this is untrue/totally false/misleading. IndexFetcher has File all over.
             for (CoreFileData cfd : resolvedMetadataResult.getFilesToPush()) {
               // Sanity check that we're talking about the same file (just sanity, Solr doesn't update files so should never be different)
               assert cfd.getFileSize() == snapshotIndexDir.fileLength(cfd.getFileName());
@@ -287,7 +280,7 @@ public class CorePushPull {
                   }
 
                   if (createNewIndexDir) {
-                    // point index to the new directory
+                    // point index to the new directory. Method call below always returns true BTW.
                     coreSwitchedToNewIndexDir = solrCore.modifyIndexProps(tempIndexDirName);
                   } else {
                     moveFilesFromTempToIndexDir(solrCore, tempIndexDir, indexDir);
@@ -434,7 +427,7 @@ public class CorePushPull {
 
       String message = String.format(Locale.ROOT,
             "PushPullData=[%s] action=%s storageProvider=%s bucketRegion=%s bucketName=%s "
-            + "runTime=%s startLatency=%s bytesTransferred=%s attempt=%s filesAffected=%s localGenerationNumber=%s blobGenerationNumber=%s ",
+              + "runTime=%s startLatency=%s bytesTransferred=%s attempt=%s filesAffected=%s localGeneration=%s blobGeneration=%s ",
           pushPullData.toString(), action, coreStorageClient.getStorageProvider().name(), coreStorageClient.getBucketRegion(),
           coreStorageClient.getBucketName(), runTime, startLatency, bytesTransferred, attempt, filesAffected,
           solrServerMetadata.getGeneration(), blobMetadata.getGeneration());
--- a/solr/core/src/java/org/apache/solr/store/blob/metadata/CorePushPull.java
+++ b/solr/core/src/java/org/apache/solr/store/blob/metadata/CorePushPull.java
@@ -156,7 +156,7 @@ public class CorePushPull {
 
         // When we build solrServerMetadata we requested to reserve the commit point for some short duration. Assumption is
         // it took less than this duration to get here (didn't do anything blocking) and now we actually save the commit
-        // point for the (pontentially long) time it takes to push all files to the Blob store.
+        // point for the (potentially long) time it takes to push all files to the Blob store.
         Directory coreIndexDir = solrCore.getDirectoryFactory().get(solrCore.getIndexDir(), DirectoryFactory.DirContext.DEFAULT, solrCore.getSolrConfig().indexConfig.lockType);
         solrCore.getDeletionPolicy().saveCommitPoint(solrServerMetadata.getGeneration());
         try {
--- a/solr/core/src/java/org/apache/solr/store/blob/metadata/CorePushPull.java
+++ b/solr/core/src/java/org/apache/solr/store/blob/metadata/CorePushPull.java
@@ -115,74 +115,76 @@ public class CorePushPull {
      */
     public BlobCoreMetadata pushToBlobStore(String currentMetadataSuffix, String newMetadataSuffix) throws Exception {
       long startTimeMs = System.nanoTime();
+      SolrCore solrCore = container.getCore(pushPullData.getCoreName());
+      if (solrCore == null) {
+        throw new Exception("Can't find core " + pushPullData.getCoreName());
+      }
+
       try {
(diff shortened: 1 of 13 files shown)


solr/core/src/java/org/apache/solr/store/blob/metadata/CorePushPull.java
```diff
@@ -96,81 +96,77 @@ protected CoreContainer getContainerFromServerMetadata(ServerSideMetadata solrSe
...
-              assert cfd.getFileSize() == snapshotIndexDir.fileLength(cfd.getFileName());
-
-              String blobPath = pushFileToBlobStore(coreStorageClient, snapshotIndexDir, cfd.getFileName(), cfd.getFileSize());
-              bcmBuilder.addFile(new BlobCoreMetadata.BlobFile(cfd.getFileName(), blobPath, cfd.getFileSize(), cfd.getChecksum()));
-            }
-          } finally {
-            solrCore.getDirectoryFactory().release(snapshotIndexDir);
+        // When we build solrServerMetadata we requested to reserve the commit point for some short duration. Assumption is
+        // it took less than this duration to get here (didn't do anything blocking) and now we actually save the commit
+        // point for the (pontentially long) time it takes to push all files to the Blob store.
```
yonik: "pontentially" spelling bug

solr/core/src/java/org/apache/solr/store/blob/metadata/CorePushPull.java
```diff
@@ -268,7 +264,7 @@ public void pullUpdateFromBlob(long requestQueuedTimeMs, boolean waitForSearcher
                   }
```
mbwaheed: Can't seem to comment on the unchanged lines, therefore, commenting here:
https://github.com/apache/lucene-solr/pull/1055/files?utf8=%E2%9C%93&diff=unified&w=1#diff-d86585f4b79b136f1575702bd7c61e38R226 is not correct, it talks about incorrect understanding of directory locking.

Pull is now an exclusive operation(i.e.no other pull or active indexing should be happening during this time). If local directory contents have changed we fail the pull. Therefore:

1. After closing index writer we should obtain an exclusive lock on directory, similar to https://github.com/apache/lucene-solr/blob/1133bf98a5fd075173efecfb75a51493fceb62b3/solr/core/src/java/org/apache/solr/update/SolrIndexSplitter.java#L161

2. We should close index write and acquire write lock in the beginning of method so that the chances of directory changes(most likely background merges?) are even less.

3. The only purpose of temp directory, I see, is to prevent a partial download of index(transient network error) from corrupting the index directory. Otherwise we could have just downloaded directly into index directory. Comments on temp directory usage need to be updated to reflect that.

solr/core/src/java/org/apache/solr/store/blob/metadata/SharedStoreResolutionUtil.java
```diff
@@ -147,42 +139,53 @@ public static SharedMetadataResolutionResult resolveMetadata(ServerSideMetadata
...
-      // We've computed blobFilesMissingLocally and localFilesMissingOnBlob is empty as it should be.
+      // We've computed blobFilesMissingLocally. localFilesMissingOnBlob is empty as it should be.
       return new SharedMetadataResolutionResult(localFilesMissingOnBlob.values(), blobFilesMissingLocally.values(), blobFilesMissingLocally.values(), false);
     }
 
-    boolean localConflictingWithBlob = false;
+    // If trying to pull files from Blob, make sure similarly named files do not already exist outside the current commit point
+    ImmutableSet<String> allLocalFiles = local.getAllFiles();
+
+    boolean downloadToNewDir = false;
```
mbwaheed: Maybe pullToNewDir? I don't have a strong preference but it would be better remain
consistent and use the same name here https://github.com/apache/lucene-solr/pull/1055/files?utf8=%E2%9C%93&diff=unified&w=1#diff-29111919397c5d8b89434d02aab74127R41.

solr/core/src/java/org/apache/solr/store/blob/metadata/ServerSideMetadata.java
```diff
@@ -132,35 +127,15 @@ public ServerSideMetadata(String coreName, CoreContainer container, boolean take
...
-//        if (allCommits.size() > 1) {
-//          allCommitsBuilder = new ImmutableSet.Builder<>();
-//          for (IndexCommit commit : allCommits) {
-//            // no snapshot for inactive segments files
-//            buildCommitFiles(coreDir, commit, allCommitsBuilder, /* snapshotDir */ null);
-//          }
-//        }
-//        allCommitsFiles = allCommitsBuilder.build();
+        // Need to inventory all local files in case files that need to be pulled from Blob conflict with them.
+        allFiles = ImmutableSet.copyOf(coreDir.listAll());
```
mbwaheed: coreDir.listAll is called twice, once here and once above inside  getSolrDirectoryHash. This can be optimized.

solr/core/src/java/org/apache/solr/store/blob/metadata/SharedStoreResolutionUtil.java
```diff
@@ -147,42 +139,53 @@ public static SharedMetadataResolutionResult resolveMetadata(ServerSideMetadata
...
     // Verify there are no inconsistencies between local index and blob index files
     for (BlobFile bf : distant.getBlobFiles()) {
+      final String solrFilename = bf.getSolrFileName();
       // We remove from map of local files those already present remotely since they don't have to be pushed.
-      localFilesMissingOnBlob.remove(bf.getSolrFileName());
-      CoreFileData cf = allLocalFiles.get(bf.getSolrFileName());
-      if (cf != null) {
+      CoreFileData localFile = localFilesMissingOnBlob.remove(solrFilename);
+      if (localFile != null) {
         // The blob file is present locally. Check if there is a conflict between local and distant (blob) versions of that file.
```
mbwaheed: The blob file is present locally. --> The blob file is present locally and is part of current commit point.

solr/core/src/java/org/apache/solr/store/blob/metadata/SharedStoreResolutionUtil.java
```diff
@@ -147,42 +139,53 @@ public static SharedMetadataResolutionResult resolveMetadata(ServerSideMetadata
...
-              cf.getFileSize(), bf.getFileSize(), cf.getChecksum(), bf.getChecksum());
+              distant.getSharedBlobName(), local.getCoreName(), solrFilename, bf.getBlobName(),
+              localFile.getFileSize(), bf.getFileSize(), localFile.getChecksum(), bf.getChecksum());
           log.info(message);
-          localConflictingWithBlob = true;
+          downloadToNewDir = true;
         }
+      } else if (allLocalFiles.contains(solrFilename)) {
+        // If the file is NOT in current commit point but similar name file exists in local dir, we risk conflict.
+        log.info("File exists locally outside of commit point. sharedShardName=" + distant.getSharedBlobName() + " coreName="+ local.getCoreName()
```
mbwaheed: outside of commit point -- > outside of current commit point

solr/core/src/java/org/apache/solr/store/blob/metadata/ServerSideMetadata.java
```diff
@@ -276,7 +224,7 @@ private String getSolrDirectoryHash(Directory coreDir) throws NoSuchAlgorithmExc
           digest.update(Long.toString(coreDir.fileLength(fileName)).getBytes());
         } catch (FileNotFoundException fnf) {
```
mbwaheed: Need to handle java.nio.file.NoSuchFileException too.

solr/core/src/java/org/apache/solr/store/blob/metadata/ServerSideMetadata.java
mbwaheed: And we don't need directory hash in push scenario, we can consider making it optional so we don't compute it in indexing path.

solr/core/src/java/org/apache/solr/store/blob/metadata/CorePushPull.java
yonik: In general, Solr does not rely on the low level lucene IndexWriter lock.  It's really just an additional safety mechanism, and a failure to aquire the lock really means that something else went badly wrong (a bug, or an external process.) The one true IndexWriter (and changes to it) are generally controlled by the SolrCoreState class.

The comments around pull being an exclusive operation sound correct.

murblanc: Pull is indeed exclusive, but let's not rely on this (i.e. be defensive). We do check the directory hasn't changed during the pull before adding back the files and reopening the IW, so I think we're ok.

Not sure about your reference @mbwaheed to SolrIndexSplitter. The lock is acquired there on a directory that I'm not sure is the index directory (it is searcher.getRawReader().directory() on the passed instance of SolrIndexSearcher). We manipulate solrCore.getIndexDir().

The directory and IndexWriter manipulation in IndexFetcher.fetchLatestIndex() is similar to the one we do here (no surprise, it was used as "inspiration").

solr/core/src/java/org/apache/solr/store/blob/metadata/ServerSideMetadata.java
murblanc: You're right on both counts. In order to limit complexity at this stage, I'm tempted to leave these optimizations for later (likely very minor improvement for indexing given the higher cost of pushing to Blob store).

solr/core/src/java/org/apache/solr/store/blob/metadata/SharedStoreResolutionUtil.java
murblanc: pulling from Blob downloads files :)
Not sure I understand what the link pasted above points to.

solr/core/src/java/org/apache/solr/store/blob/metadata/CorePushPull.java
mbwaheed: I read this and your updated comments in the code. Sounds good.

solr/core/src/java/org/apache/solr/store/blob/metadata/SharedStoreResolutionUtil.java
mbwaheed: The link is not working for me anymore either(it was pointing to a line in the code diff).
It was pointing to SharedMetadataResolutionResult#localConflictingWithBlob.
I was suggesting to keep that consistent with your rename here.

solr/core/src/java/org/apache/solr/store/blob/metadata/ServerSideMetadata.java
mbwaheed: sounds good.

murblanc:
@mbwaheed @yonik I have rebased the changes after Bilal approved the review.

PR #1058, opened 2019-12-05
SOLR-13972: Warn about insecure settings on startup

gerlowskija:
# Description
Solr should warn users on startup if insecure settings are being used.

# Solution
Currently I've added the detection and message for this in bin/solr.  I'm not sure this is the best place for this logic - I'd like to keep logic out of bash/powershell where possible, and a smarter check would look at security.json in ZooKeeper, which is easier to do within the JVM.

But this PR is still worth reviewing if anyone has feedback on the language or appearance of the warning.

# Tests

I've done manual testing, but that's it.  If we had a `bin/solr` test framework as SOLR-11749 proposes, we could conceivably write tests for this, but that's a major task and isn't going to happen anytime soon.

### DIFF
--- a/solr/core/src/java/org/apache/solr/core/CoreContainer.java
+++ b/solr/core/src/java/org/apache/solr/core/CoreContainer.java
@@ -40,6 +40,7 @@ import java.util.concurrent.Future;
 
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Maps;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.http.auth.AuthSchemeProvider;
 import org.apache.http.client.CredentialsProvider;
 import org.apache.http.config.Lookup;
@@ -657,6 +658,7 @@ public class CoreContainer {
 
     securityConfHandler = isZooKeeperAware() ? new SecurityConfHandlerZk(this) : new SecurityConfHandlerLocal(this);
     reloadSecurityProperties();
+    warnUsersIfSecurityDisabled();
     this.backupRepoFactory = new BackupRepositoryFactory(cfg.getBackupRepositoryPlugins());
 
     createHandler(ZK_PATH, ZookeeperInfoHandler.class.getName(), ZookeeperInfoHandler.class);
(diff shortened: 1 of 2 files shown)


solr/core/src/java/org/apache/solr/core/CoreContainer.java
```diff
@@ -897,6 +899,20 @@ private void reloadSecurityProperties() {
     initializeAuditloggerPlugin((Map<String, Object>) securityConfig.getData().get("auditlogging"));
   }
 
+  private void warnUsersIfSecurityDisabled() {
+    if (authenticationPlugin == null || authorizationPlugin == null) {
+      log.warn("Not all security plugins configured!  authentication={} authorization={}.  Solr is only as secure as " +
+          "you make it. Consider configuring authentication/authorization before exposing Solr to users internal or " +
+          "external.  See https://lucene.apache.org/solr/guide/authentication-and-authorization-plugins.html for more info",
```
janhoy: Point to securing-solr parent page instead?

gerlowskija: Sure.

gerlowskija:
> I hoped we had a way to only log such warning if Solr is bound to a public network.
I'd rather keep that sort of logic out of it.  I can imagine situations where Solr is nominally bound to "localhost", but is actually exposed more broadly because of traffic forwarding, iptables rules, etc.

(I'm not the strongest network guy - if someone wants to argue with me that this isn't possible and we can always rely that localhost-binding == no-need-for-auth I'm happy to be corrected.)

janhoy:
Well, if you just download and run Solr without specifying `SOLR_HOST` (or `-Dhost=x`) then earlier it would bind to ALL interfaces and you'd be able to connect to Solr from a 192.x.x.x or 10.x.x.x network or even from a public IP. With [SOLR-13985](https://issues.apache.org/jira/browse/SOLR-13985), Solr will by default bind to 127.0.0.1, and only if you install, say, a proxy software like Nginx on the Solr host would you be able to connect from the outside.

If we remove the noise from logs in localhost (dev) mode, it would give a more welcoming first-use experience and we could allow ourselves to be even more aggressive when *not* in localhost-mode. We could flood the AdminUI with warnings, we could fail to start Solr at all if no security configured, with the need for a `-Di.am.a.fool.to.run.solr.without.security=true` option to override. That would trigger some healthy discussions in various IT depts :)

rmuir:
I don't think we should try to do anything tricky. Warn users when it looks like the configuration is insecure, and give them a way to disable the warning. Make the warning simple and useful. Developers, especially first-use ones, need to see it!

Someone might be doing the TLS+basic auth on their nginx, haproxy, ELB, and its actually secure but doesn't look that way. someone might be bound to all interfaces but not exposed to the outside, because its just docker internal. someone might be bound to what looks like an internal address but its exposed, because that is how some cloud system works, and so on.

solr/core/src/java/org/apache/solr/core/CoreContainer.java
rmuir: should we use an s.apache.org shortened link for this?

solr/core/src/java/org/apache/solr/core/CoreContainer.java
```diff
@@ -657,6 +658,7 @@ public void load() {
 
     securityConfHandler = isZooKeeperAware() ? new SecurityConfHandlerZk(this) : new SecurityConfHandlerLocal(this);
     reloadSecurityProperties();
+    warnUsersIfSecurityDisabled();
```
anshumg: `warnUsersOnInsecureSettings()` ? Security might be enabled in some way here but we still might prompt with a warning here e.g. auth enabled but TLS disabled. Don't want to be a pain with fixing the method names, but I think this change might make the code more readable.

gerlowskija:
Few review comments there is no disagreement on:
1. Change a method name.
2. Change a link referenced in the log message.

I'll make both those changes.  Since there's disagreement on whether or not the suppress the message if localhost-binding is used, I'm going to keep it as is.  I don't have strong opinions here though, so if Jan convinces people otherwise, I'm also happy to change it.

Planning to update the patch today and merge it on Wednesday pending further review feedback.

janhoy:
Progress is better than perfection, so it's your call. Why I'm reluctant is that the developer experience deteriorates for each warning we add to the "normal" case of just starting up a Solr node for the first time on your laptop.

gerlowskija:
I agree about the new-dev/user experience, I just don't know enough about the network environments people deploy Solr in to know whether localhost-binding is a reliable predictor of dev vs. production.

Wondering aloud... if Solr had an explicit `-Dproduction=true|false` System property, we could base these sorts of decisions off of it and wouldn't have to second guess how accurate/inaccurate our heuristic was.  Though we still have to rely on the user knowing to set the system prop....

PR #1068, opened 2019-12-08
SOLR-14033: Fix Hadoop tests with security manager

risdenk:


risdenk:
@rmuir this DEFINITELY isn't ready yet but pushing early to get some eyes on it. The `org.apache.hadoop` copied files are straight from Apache Hadoop 3.2.0 and then modified to fix some of the security manager issues.

I've seen HDFS tests pass on mac with `ant test -Dtests.badapples=false -Dtests.nightly=true -Dtests.slow=true -Dtests.asserts=true -Dtests.awaitsfix=false -D'testcase=*Hdfs*|*HDFS*'` to run just the Hadoop tests. I haven't run everything or precommit yet.

rmuir:
The main question i have with the current approach is that it will allow us to fix issues for tests, but we'd then need a different solution for "real" (https://issues.apache.org/jira/browse/SOLR-13984), because the forked classes are currently just in src/test. Otherwise tests will pass but then users will hit the issues in production. Maybe some stuff (e.g. FileUtil) could be moved to src/java? I don't understand how classpath is fed to jetty yet but the order will be important so that it works. Maybe we get lucky, its easy and "just works"

risdenk:
So thats definitely a valid concern. I have two related thoughts to that:

1) most of the changes are for the server side of Hadoop - NameNode, Datanode, etc for handling the FS itself. Solr is really just a client so I hope not as many changes are needed on the client side.

That being said, I don't know for sure how to separate out the integration tests security requirements from the client side security requirements. I didn't want to package Hadoop code into the actual distribution since that requires us to maintain it more carefully. Currently the client side Hadoop code can be included external to Hadoop (ie: different Hadoop versions if necessary). So your comment about packaging would be an issue I think in real clusters.

2) Worst case need to disable the security manager when using Hadoop with Solr. It would be a sledgehammer, but would enable secure Solr by default and then make the choice about using with Solr or not (FWIW the Elasticsearch HDFS plugin had this exact model for a while until all the HDFS permissions were nailed down in tests).

rmuir:
> That being said, I don't know for sure how to separate out the integration tests security requirements from the client side security requirements. 

I see. the way we fought HDFS this way before for ES was to spin up an external test fixture of the "server" stuff so that the code under test was not mixing the two things: it only needed to depend on the client code. that way you can have a clean integration test and have some real confidence about correctness. I don't think its crazy, it is similar to how you might spin up a real postgresql database or LDAP server for doing integration tests of a webapp. But it would require work on the build side.

today there is a mix if hadoop client, server, test code all in one massive classpath with the rest of solr stuff. so maybe my question was kinda dumb because its impossible for me to tell what-is-what :)

dweiss:
>
> today there is a mix if hadoop client, server, test code all in one
> massive classpath with the rest of solr stuff. so maybe my question was
> kinda dumb because its impossible for me to tell what-is-what :)
>
... and this mix of things propagates into the build classpath, dependency
imports/ exports, etc. making it very difficult to reason about what is
what in dependencies.

Things like integration test fixtures can be implemented in an elegant way
with gradle (dependsOn
startFixtureTask,integTest.finalizedBy(closeFixtureTask)),
but it's still a long way before we get there with the current state of
tests...

Dawid

>

risdenk:
> I see. the way we fought HDFS this way before for ES was to spin up an external test fixture of the "server" stuff so that the code under test was not mixing the two things: it only needed to depend on the client code. that way you can have a clean integration test and have some real confidence about correctness. I don't think its crazy, it is similar to how you might spin up a real postgresql database or LDAP server for doing integration tests of a webapp. But it would require work on the build side.

yea that would be fantastic. testcontainers I've used for a few projects so its pretty integrated into the build and uses Docker for a "real" service. There are probably other ways to do it too. Would be a great goal though to get to at some point.

risdenk:
So fixed precommit and removed the Solr security manager hacks for Hadoop. The tests have passed on Mac. I haven't checked on Windows yet. This is a big improvement though.

@rmuir / @dweiss let me know if you have any concerns w/ this.

PS - There is a race condition or something in the order of initializing the Hadoop UGI so sometimes the tests fail. That is a preexisting condition though just exasperated by trying to set the groups config. I'll track it down. Just don't think tests failing sometimes is as big of an issue we have today w/ the tests failing basically always for HDFS related stuff.

rmuir:
I can test on linux in a VM too. I am starting with a mac just to reproduce what you have done.

In my experience with hadoop stuff, if you have initialization order problems, sometimes an explicit (but documented why) `Class.forName` is a useful hammer.

As a followup we'll also be able to prune away some stupid permissions i added such as `/dev/random` which should not have been necessary in the first place, i was just working around the hadoop hack.

Thank you so much for attacking this!

rmuir:
I had two failures, both involved hadoop trying to execute to get unix supplementary groups:
```
   [junit4]   - org.apache.solr.cloud.TestSolrCloudWithSecureImpersonation (suite)
   [junit4]   - org.apache.solr.security.hadoop.TestImpersonationWithHadoopAuth (suite)
```

Stacks look like this in both cases:
```
   [junit4] ERROR   0.00s J0  | TestImpersonationWithHadoopAuth (suite) <<<
   [junit4]    > Throwable #1: com.google.common.util.concurrent.UncheckedExecutionException: java.security.AccessControlException: access denied ("java.io.FilePermission" "<<ALL FILES>>" "execute")
   [junit4]    > 	at __randomizedtesting.SeedInfo.seed([C1005F0247624]:0)
   [junit4]    > 	at com.google.common.cache.LocalCache$Segment.get(LocalCache.java:2050)
   [junit4]    > 	at com.google.common.cache.LocalCache.get(LocalCache.java:3951)
   [junit4]    > 	at com.google.common.cache.LocalCache.getOrLoad(LocalCache.java:3973)
   [junit4]    > 	at com.google.common.cache.LocalCache$LocalLoadingCache.get(LocalCache.java:4957)
   [junit4]    > 	at org.apache.hadoop.security.Groups.getGroups(Groups.java:228)
   [junit4]    > 	at org.apache.solr.security.hadoop.ImpersonationUtil.getUsersFirstGroup(ImpersonationUtil.java:42)
   [junit4]    > 	at org.apache.solr.security.hadoop.TestImpersonationWithHadoopAuth.setupClass(TestImpersonationWithHadoopAuth.java:74)
   [junit4]    > 	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
   [junit4]    > 	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
   [junit4]    > 	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
   [junit4]    > 	at java.base/java.lang.reflect.Method.invoke(Method.java:567)
   [junit4]    > 	at java.base/java.lang.Thread.run(Thread.java:830)
   [junit4]    > Caused by: java.security.AccessControlException: access denied ("java.io.FilePermission" "<<ALL FILES>>" "execute")
   [junit4]    > 	at java.base/java.security.AccessControlContext.checkPermission(AccessControlContext.java:472)
   [junit4]    > 	at java.base/java.security.AccessController.checkPermission(AccessController.java:1036)
   [junit4]    > 	at java.base/java.lang.SecurityManager.checkPermission(SecurityManager.java:408)
   [junit4]    > 	at java.base/java.lang.SecurityManager.checkExec(SecurityManager.java:658)
   [junit4]    > 	at java.base/java.lang.ProcessBuilder.start(ProcessBuilder.java:1096)
   [junit4]    > 	at java.base/java.lang.ProcessBuilder.start(ProcessBuilder.java:1071)
   [junit4]    > 	at org.apache.hadoop.util.Shell.runCommand(Shell.java:938)
   [junit4]    > 	at org.apache.hadoop.util.Shell.run(Shell.java:901)
   [junit4]    > 	at org.apache.hadoop.util.Shell$ShellCommandExecutor.execute(Shell.java:1213)
   [junit4]    > 	at org.apache.hadoop.security.ShellBasedUnixGroupsMapping.getUnixGroups(ShellBasedUnixGroupsMapping.java:200)
   [junit4]    > 	at org.apache.hadoop.security.ShellBasedUnixGroupsMapping.getGroups(ShellBasedUnixGroupsMapping.java:97)
   [junit4]    > 	at org.apache.hadoop.security.JniBasedUnixGroupsMappingWithFallback.getGroups(JniBasedUnixGroupsMappingWithFallback.java:51)
   [junit4]    > 	at org.apache.hadoop.security.Groups$GroupCacheLoader.fetchGroupList(Groups.java:387)
   [junit4]    > 	at org.apache.hadoop.security.Groups$GroupCacheLoader.load(Groups.java:321)
   [junit4]    > 	at org.apache.hadoop.security.Groups$GroupCacheLoader.load(Groups.java:270)
   [junit4]    > 	at com.google.common.cache.LocalCache$LoadingValueReference.loadFuture(LocalCache.java:3527)
   [junit4]    > 	at com.google.common.cache.LocalCache$Segment.loadSync(LocalCache.java:2276)
   [junit4]    > 	at com.google.common.cache.LocalCache$Segment.lockedGetOrLoad(LocalCache.java:2154)
   [junit4]    > 	at com.google.common.cache.LocalCache$Segment.get(LocalCache.java:2044)
   [junit4]    > 	... 29 more
```

I haven't looked at what is going on here, maybe the user's supplementary groups could be retrieved instead via JAAS functionality, e.g. https://docs.oracle.com/javase/7/docs/jre/api/security/jaas/spec/com/sun/security/auth/module/UnixSystem.html ?

rmuir:
Big disclaimer that definitely didn't dig as to why hadoop would ever need to know this. Versus say just checking that it has access to a file. This can be done in so many ways, including explicit `access(2)` hook via `path.getFileSystem().provider()`:

https://docs.oracle.com/javase/7/docs/api/java/nio/file/spi/FileSystemProvider.html#checkAccess(java.nio.file.Path,%20java.nio.file.AccessMode...)

risdenk:
So this is because the `UserGroupInfromation` is initialized once statically in the JVM. If it isn't configured before being initialized then you end up w/ the wrong group implementation.

I tried to set the group impl here https://github.com/apache/lucene-solr/pull/1068/files#diff-3c36d2ce2df3ab00d607b1c9eb01ae29R210 and then force the UGI initialization here https://github.com/apache/lucene-solr/pull/1068/files#diff-3c36d2ce2df3ab00d607b1c9eb01ae29R132

However if the order of Hadoop tests run where HdfsTestUtil#setupClass isn't run before any other Hadoop related test, the UGI gets initialized incorrectly and then fails with the stacktrace you shared. 


So yea I tried to bypass the whole group lookup crap by putting in a fake group lookup provider, but it requires the UGI to be initialized upfront. I haven't figured out the best place to ensure the UGI is initialized with the right config.

risdenk:
So the ways to fix this:
* Ensure that the UGI is configured up front
* Figure out how to ensure that the group provider mapping is always picked up in default config
* Fix Hadoop to not use shell to get groups

I opted to try to do 1. and just need to make sure that is done in a few more places. 2 seemed risky to try to do in the test framework. 3 isn't going to happen anytime soon.

rmuir:
> So yea I tried to bypass the whole group lookup crap by putting in a fake group lookup provider, but it requires the UGI to be initialized upfront. I haven't figured out the best place to ensure the UGI is initialized with the right config.

Can we try to do this with an explicit static initializer "earlier" up in the file. the order that this happens is defined, there are just too many hooks and its happening too late. it seems everything it needs is static and it could be moved up in the file?

rmuir:
```
However if the order of Hadoop tests run where HdfsTestUtil#setupClass isn't run before any other Hadoop related test (this happens in tests that don't use HdfsTestUtil), the UGI gets initialized incorrectly and then fails with the stacktrace you shared.
```

Can it even happen in SolrTestCaseJ4 so that test order doesn't mess with us? `static void fixHadoop` :)

dweiss:
Anything that can improve the state of those tests I'm +1 for. :)

rmuir:
@dweiss https://issues.apache.org/jira/browse/SOLR-14021 https://issues.apache.org/jira/browse/SOLR-13989

those are two issues that could relieve the pain. if its not worth the battle (2nd bullet from https://github.com/apache/lucene-solr/pull/1068#issuecomment-563372315) then if the hadoop was in its own module we could run just that module naked with no security manager, tell users that here is the big caveat, and stop wrestling with it.

also doing only unit tests with something like mockito and not doing integration tests in the current way would be really helpful. maybe that is somehow easier?

risdenk:
>Can it even happen in SolrTestCaseJ4 so that test order doesn't mess with us? static void fixHadoop :)

Yup had that same thought last night. I was trying to not add more Hadoop specific stuff to "core solr", but that will definitely do the trick in the short term. I can work on that today.

rmuir:
+1

risdenk:
Found a better way to do it. Set core-site.xml on the default test resources so it gets picked up for all `new Configuration()`. Currently just setting the the group mapping that config.

risdenk:
@chatman @gerlowskija FYI based on https://the-asf.slack.com/archives/CEKUCUNE9/p1575957768305100

risdenk:
This is passing tests pretty regularly on Mac. Better than before where all tests were failing.

risdenk:
@rmuir can you take another look over this? I think this is close to ready to merge. I plan to squash the commits and update the commit message. Figure its easier to review in smaller commits though.

rmuir:
I will test it on mac and linux this evening and look over what you have done. thanks again!

gerlowskija:
The duplication of the Hadoop Java files here seems ugly, though I'm sure you're only doing it because there's no other recourse : (

Is it worth putting some sort of header or class-level comment on all these copied-over files to indicate that they were copied verbatim from whatever particular hadoop package?  Or alternatively maybe we should have a comment in ant/ivy where we declare our Hadoop dependency that all these files should be updated whenever we change our hadoop version?

I'm running tests now and will post later when they're done as well.

gerlowskija:
Tests still fail for me with this PR.

To double check, here's how I applied it:
```
git checkout master
git pull
git checkout -b apply_1068
wget https://github.com/apache/lucene-solr/pull/1068
patch -p1 -i 1068.patch
ant clean compile && cd solr && ant server && cd ..
ant clean test -Dtests.iters=5 -Dtestcase=HdfsBackupRepositoryTest
```

And here's my Java version.  Not sure if it's relevant, but...
```
openjdk version "11.0.2" 2019-01-15
OpenJDK Runtime Environment 18.9 (build 11.0.2+9)
OpenJDK 64-Bit Server VM 18.9 (build 11.0.2+9, mixed mode)
```

[ant-output.txt](https://github.com/apache/lucene-solr/files/3946894/ant-output.txt)

Attaching the ant output as well.

rmuir:
Hi @gerlowskija thank you for helping test.

Can you run this for me from your command shell: `echo $HADOOP_HOME` I want to see if it is set to something. it looks to me like hadoop is installed via homebrew or similar mechanism.

If you look at the exception, it happens because code tries to access `/usr/local/Cellar/hadoop/3.1.1` on your local machine, which is not even the correct one for solr tests.

From the stacktrace you find relevant code here:

https://github.com/apache/hadoop/blob/release-3.2.0-RC1/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/util/Shell.java#L417-L440

If this has been really happening, good that its caught now. We may have to explicitly clear this environment variable in tests (or ensure sysprop is set) from the build system or similar.

rmuir:
If this environment variable doesn't yield anything, could you run `env` and manually look for anything suspicious such as `JAVA_TOOL_OPTIONS=` that might contain hadoop related sysprops? I want to know why its looking on that place in your filesystem.

gerlowskija:
Good thinking.

```
➜  lucene-solr-1 git:(SOLR-13999) ✗ echo $HADOOP_HOME
/usr/local/Cellar/hadoop/3.1.1
```

+1 for adding something to unset that env-var if present, though idk where that would happen.

risdenk:
Thanks @gerlowskija. I would bet that problem totally exists today. Hadoop loves to pickup environment variables :( I'll take a look at that separately.

risdenk:
I'm going to merge this here in a few minutes once my precommit run on the squashed commit finishes. Will address any follow up concerns this brings.

PR #1070, opened 2019-12-10
LUCENE-9089: FST Builder fluent-style constructor.

bruno-roustant:
Only FST Builder is refactored. The other classes are touched to adapt to the new fluent style.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/util/fst/Builder.java
+++ b/lucene/core/src/java/org/apache/lucene/util/fst/Builder.java
@@ -51,29 +51,7 @@ import org.apache.lucene.util.fst.FST.INPUT_TYPE; // javadoc
 
 public class Builder<T> {
 
-  /**
-   * Default oversizing factor used to decide whether to encode a node with direct addressing or binary search.
-   * Default is 1: ensure no oversizing on average.
-   * <p>
-   * This factor does not determine whether to encode a node with a list of variable length arcs or with
-   * fixed length arcs. It only determines the effective encoding of a node that is already known to be
-   * encoded with fixed length arcs.
-   * See {@code FST.shouldExpandNodeWithFixedLengthArcs()}
-   * and {@code FST.shouldExpandNodeWithDirectAddressing()}.
-   * <p>
-   * For English words we measured 217K nodes, only 3.27% nodes are encoded with fixed length arcs,
-   * and 99.99% of them with direct addressing. Overall FST memory reduced by 1.67%.
-   * <p>
-   * For worst case we measured 168K nodes, 50% of them are encoded with fixed length arcs,
-   * and 14% of them with direct encoding. Overall FST memory reduced by 0.8%.
-   * <p>
-   * Use {@code TestFstDirectAddressing.main()}
-   * and {@code TestFstDirectAddressing.testWorstCaseForDirectAddressing()}
-   * to evaluate a change.
-   *
-   * @see #setDirectAddressingMaxOversizingFactor
-   */
-  static final float DIRECT_ADDRESSING_MAX_OVERSIZING_FACTOR = 1.0f;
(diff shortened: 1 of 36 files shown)


dweiss:
This is a builder-on-builder now. :) We should have used a different name for the "compiler" of FSTs then it could become:

FST
FSTCompiler
FSTCompiler.Builder

and then:
FSTCompiler.build() -> FSTCompiler.Builder.build() -> FSTCompiler().[add()*|build()] -> FST

But it looks like a nice improvement even now!

bruno-roustant:
Indeed it's a Builder pattern on a class named "Builder", which is odd. That's why I named the methods "construct" (instead of classical "new Builder") and "create" (instead of "build").

I agree with the FSTCompiler and FSTCompiler.Builder names. How do we manage a renaming? These classes are public. Right, they are "expert" and "experimental" but still public. Do we accept to rename them? In 9x only?

dweiss:
Right... I've always wondered how we should handle such things gracefully (and I'm a user of those "secret" APIs myself so a rename affects me too. 

On the other hand if we tangle ourselves with all sorts of deprecations of these experimental APIs then things will build up cruft like the current FST class which is quite horrible (and I'm saying this as a person who is responsible for the current state).

I'd say go ahead and rename on master. My suggestion is not anything you have to hold on to: use your gut feeling to clean up in any way you like.

lucene/core/src/java/org/apache/lucene/util/fst/Builder.java
```diff
@@ -205,22 +156,145 @@ public Builder(FST.INPUT_TYPE inputType, int minSuffixCount1, int minSuffixCount
...
-   * only binary search nodes will be created.
-   *
-   * @see #DIRECT_ADDRESSING_MAX_OVERSIZING_FACTOR
+   * Creates an FST/FSA builder with all the possible tuning and construction tweaks.
+   * Read parameter documentation carefully.
    */
-  public Builder<T> setDirectAddressingMaxOversizingFactor(float factor) {
-    directAddressingMaxOversizingFactor = factor;
-    return this;
+  public static class Constructor<T> {
```
juanka588: re name to builder

msokolov: I agree - Constructor is easily confused with java.util.Constructor

msokolov: Oh o.a.l.util.fst.FSTCompiler or o.a.l.util.fst.Compiler - I guess I have a slight preference for the latter, but up to you

bruno-roustant: I have a preference for fst.FSTCompiler. This may seem redundant when looking at the full path, but inside some code it is clearer to see FSTCompiler than just Compiler. So I'll go with FSTCompiler.

bruno-roustant:
Note that this renaming will touch a lot of classes currently using Builder. I'll also rename the "builder" variables to fstCompiler. Only for master of course.

Related question: since a lot of classes will be touched, it may imply that some other works on those touched classes will have to un-rename when backporting to 8x. So should I wait some time before effectively doing this renaming (e.g. closer to 9x)?

dweiss:
There are not that many people touching the FST classes. I don't think you should wait -- stale patch will require a new round of work, much like backporting a patch that touches the existing FST code.

I'd say go for it.

bruno-roustant:
@dweiss I also reduced the visibility of FSTCompiler inner classes to package (was public). Anyway the inner Node and CompiledNode were already package visible so I don't think Arc was really usable publicly. That makes them easier to evolve without useless constraint on public API.

dweiss:
Yes, sounds good to me.

msokolov:
I was confused at first, since FST.Arc is used elsewhere, but that's right (fst).Builder.Arc is not - it should be encapsulated

PR #1071, opened 2019-12-11
SOLR-14003: Refactor code to avoid reading state from Replica/Slic

noblepaul:


solr/solrj/src/java/org/apache/solr/client/solrj/cloud/DirectShardState.java
```diff
@@ -0,0 +1,68 @@
...
+
+import java.util.function.Predicate;
+
+import org.apache.solr.common.cloud.Replica;
+import org.apache.solr.common.cloud.Slice;
+
+/**Reads the state from state.json
+ *
+ */
+public class DirectShardState implements ShardStateProvider {
```
sigram: IMHO having this class is harmful to the intended goal, because it maintains the illusion that replica state is actually kept up to date inside the `Replica` object. The whole point of this refactoring, as I understand it, is that the state is maintained elsewhere.
I would even go as far as to completely remove `Replica.getState()`.

solr/solrj/src/java/org/apache/solr/client/solrj/cloud/ShardStateProvider.java
```diff
@@ -0,0 +1,56 @@
...
+ * The state is embedded with other topology information in state.json in the original SolrCLoud design.
+ * But, it is possible to have another implementation where the state and leader information can be stored and read
+ * from other places.
+ *
+ */
+public interface ShardStateProvider {
+
+  /**Get the state of  a given {@link Replica}
+   */
+  Replica.State getState(Replica replica);
```
sigram: As I understand it, the true state is no longer maintained in `Replica` so this should be just a default method that delegates to the methods that use only replica identifier(s) (such as collection, shard, coreNode or coreName) and don't need the actual object.

solr/solrj/src/java/org/apache/solr/client/solrj/cloud/ShardStateProvider.java
```diff
@@ -0,0 +1,56 @@
...
+
+  /**Get the leader {@link Replica} of the shard
+   */
+  Replica getLeader(Slice slice);
```
sigram: See the comment above - we're not getting this information from the `Slice` object, do we? Then it should be a default method that simply extracts collection and shard name.

solr/solrj/src/java/org/apache/solr/client/solrj/cloud/ShardStateProvider.java
```diff
@@ -0,0 +1,56 @@
...
+
+  /**Get the leader of the slice. Wait for one if there is no leader
+   * @param timeout how much time to wait for a leader to com eup. -1 means the default value will be used
+   * Throws an {@link InterruptedException} if interrupted in between
+   */
+  Replica getLeader(Slice slice, int timeout) throws InterruptedException;
```
sigram: These two methods should also throw a TimeoutException or something like that when `timeout > 0`, because this value indicates the user's intent to get a definitive answer.

solr/solrj/src/java/org/apache/solr/client/solrj/cloud/ShardStateProvider.java
```diff
@@ -0,0 +1,56 @@
...
+
+
+  Replica getLeader(String collection, String slice, int timeout) throws InterruptedException;
+
+
+  /**CHeck if the replica is active
+   *
+   */
+  boolean isActive(Replica replica);
```
sigram: See comments above about using default methods that extract identifiers from the object.

solr/solrj/src/java/org/apache/solr/common/cloud/Replica.java
```diff
@@ -167,11 +167,15 @@ public String getNodeName() {
     return nodeName;
   }
   
-  /** Returns the {@link State} of this replica. */
+  /** Returns the {@link State} of this replica. use {@link org.apache.solr.client.solrj.cloud.ShardStateProvider#getState(Replica)}  instead*/
+  @Deprecated
   public State getState() {
```
sigram: To be removed in 9.0?

solr/solrj/src/java/org/apache/solr/client/solrj/cloud/ShardStateProvider.java
HoustonPutman: I think that this needs additional documentation stating whether the method returns true if the node which the replica is on is live or not.

solr/core/src/test/org/apache/solr/cloud/SyncSliceTest.java
```diff
@@ -217,11 +218,13 @@ private void waitTillAllNodesActive() throws Exception {
...
       ClusterState clusterState = zkStateReader.getClusterState();
       DocCollection collection1 = clusterState.getCollection("collection1");
+      ShardStateProvider ssp = zkStateReader.getShardStateProvider(collection1.getName());
+
       Slice slice = collection1.getSlice("shard1");
       Collection<Replica> replicas = slice.getReplicas();
       boolean allActive = true;
       for (Replica replica : replicas) {
-        if (!clusterState.liveNodesContain(replica.getNodeName()) || replica.getState() != Replica.State.ACTIVE) {
+        if (!clusterState.liveNodesContain(replica.getNodeName()) || ssp.getState(replica)!= Replica.State.ACTIVE) {
```
HoustonPutman: Why not use a `!ssp.isActive(replica)`?

I see inconsistent usage of this logic throughout.

solr/solrj/src/java/org/apache/solr/client/solrj/cloud/DirectShardState.java
noblepaul: How else can we achieve the goal. The idea is for both of these to co-exist for a few releases. Eventually, this can go away

solr/solrj/src/java/org/apache/solr/client/solrj/cloud/ShardStateProvider.java
noblepaul: OK

solr/solrj/src/java/org/apache/solr/common/cloud/Replica.java
noblepaul: This may not be removed in 9.0. This may impact almost 100% of SolrCloud users. We need to ensure that the new solution is more stable, performant and our users are on board with this. This may go away in 10.0

solr/core/src/test/org/apache/solr/cloud/SyncSliceTest.java
noblepaul: OK

solr/solrj/src/java/org/apache/solr/client/solrj/cloud/ShardStateProvider.java
noblepaul: As explained earlier. The current mode will continue to be used for a while. We will have to support both

solr/solrj/src/java/org/apache/solr/common/cloud/Replica.java
sigram: Then I think this refactoring falls short of its potential - because instead of having a single uniform API we now end up with two APIs that claim to do the same, and the old one still prevents us from moving the state reporting out of the actual Replica object - ie. no matter how clever we are with the `ShardStateProvider` we still end up having to cache the potentially stale state inside `Replica`.

solr/solrj/src/java/org/apache/solr/client/solrj/cloud/DirectShardState.java
sigram: I think we have to decide where we keep (and maintain) the actual state. The API that you proposed provides two ways of getting this info, and two potential locations to get it (which may get out of sync, or have different staleness), which in my opinion is one too many :)

If the replica state resides in `Replica` and this is our source of truth then having a `ShardStateProvider` seems pointless. What is the source of truth then - the state that is cached in `Replica` or the state that we obtain from `ShardStateProvider` implementation, which is guaranteed to the the most up-to-date?

My understanding was that the purpose of introducing `ShardStateProvider` was to provide The single API point that is the only source of truth about the current shard/replica state (similar to other *StateProvider-s). In this model `Replica` becomes a simple static descriptor and it's NOT responsible for caching its state - that's the role of `ShardStateProvider`.

solr/solrj/src/java/org/apache/solr/common/cloud/Replica.java
noblepaul: > the old one still prevents us from moving the state reporting out of the actual Replica object..

We can only deprecate stuff . We could possibly remove it but that should be done later (may be 9.0)

solr/solrj/src/java/org/apache/solr/client/solrj/cloud/DirectShardState.java
noblepaul: >My understanding was that the purpose of introducing ShardStateProvider was to provide The single API point that is the only source of truth about the current shard/replica state (similar to other *StateProvider-s)

The `state` attribute will have to go away when the old model goes away. 

So, if your collection is created with `externalState=true` You really have no information of the state in the `state.json`. Nobody updates it nobody reads from it. Internally we always must rely on the interface to do the right thing.

noblepaul:
@sigram I think there is some miscommunication. Can you take a quick glance at the branch `jira/solr14089` on how this is actually used. May be it can help us get on the same page

chatman:
> @sigram I think there is some miscommunication. Can you take a quick glance at the branch jira/solr14089 on how this is actually used. May be it can help us get on the same page

I think we should hold off on merging this PR until SOLR-14089 is actually ready. This change in isolation is not useful.

solr/solrj/src/java/org/apache/solr/client/solrj/cloud/ShardStateProvider.java
```diff
@@ -0,0 +1,61 @@
...
+   */
+  Replica getLeader(Slice slice);
+
+  Replica getLeader(String coll, String slice);
+
+  /**Get the leader of the slice. Wait for one if there is no leader
+   * @param timeout how much time to wait for a leader to com eup. -1 means the default value will be used
+   * Throws an {@link InterruptedException} if interrupted in between
+   */
+  Replica getLeader(Slice slice, int timeout) throws InterruptedException;
```
madrob: No indication on units for the timeout, consider adding TimeUnit argument.

solr/core/src/java/org/apache/solr/cloud/ElectionContext.java
```diff
@@ -580,7 +580,7 @@ public void publishActiveIfRegisteredAndNotActive(SolrCore core) throws Exceptio
         ClusterState clusterState = zkStateReader.getClusterState();
         Replica rep = getReplica(clusterState, collection, leaderProps.getStr(ZkStateReader.CORE_NODE_NAME_PROP));
         if (rep == null) return;
-        if (rep.getState() != Replica.State.ACTIVE || core.getCoreDescriptor().getCloudDescriptor().getLastPublished() != Replica.State.ACTIVE) {
+        if (zkStateReader.getShardStateProvider(collection).getState(rep) != Replica.State.ACTIVE || core.getCoreDescriptor().getCloudDescriptor().getLastPublished() != Replica.State.ACTIVE) {
```
madrob: Should this be `! isActive(rep)` instead of `getState(rep) != ACTIVE`? There's a lot of places where we still call `getState`.

solr/solrj/src/java/org/apache/solr/client/solrj/cloud/DirectShardState.java
```diff
@@ -0,0 +1,81 @@
...
+  private final Predicate<String> isNodeLive;
+  private final Function<String , DocCollection> collectionProvider;
+
+  public DirectShardState(Predicate<String> isNodeLive, Function<String, DocCollection> collectionProvider) {
+    this.isNodeLive = isNodeLive;
+    this.collectionProvider = collectionProvider;
+  }
+
+  @Override
+  public Replica.State getState(Replica replica) {
```
madrob: I don't understand the goal of this method. Should all usage be switched to `isActive()` and possibly some new `isRecovering()` and `isDown()` if we care about those distinctions?

noblepaul:
Will not fix

PR #1073, opened 2019-12-11
LUCENE-9088: JapaneseNumberFilter uses inaccurate PartOfSpeechAttribute

cbuescher:
Currently the JapaneseNumberFilter reads past a single or multiple numeric
tokens and emits the new composed token with the attributes of the following
token. This will often lead to e.g. wrong part-of-speech attributes on the
numeric token, which in turn can lead to wrong filtering by subsequent filters.

This change keeps track of the state of the last numeric token while iterating
over a number group and restores the last seen state before emiting the composed
numeric token, so we use the attributes of the last one.

### DIFF
--- a/lucene/analysis/kuromoji/src/java/org/apache/lucene/analysis/ja/JapaneseNumberFilter.java
+++ b/lucene/analysis/kuromoji/src/java/org/apache/lucene/analysis/ja/JapaneseNumberFilter.java
@@ -180,8 +180,14 @@ public class JapaneseNumberFilter extends TokenFilter {
     int endOffset = 0;
     State preCompositionState = captureState();
     String term = termAttr.toString();
+    // while iterating over further incoming token, we want to keep track of the last numeral we see
+    // in order to restore its state when we emit the the joined numeral token
+    State lastNumeralTokenState = null;
     boolean numeralTerm = isNumeral(term);
-    
+    if (numeralTerm) {
+      lastNumeralTokenState = captureState();
+    }
+
     while (moreTokens && numeralTerm) {
 
       if (!composedNumberToken) {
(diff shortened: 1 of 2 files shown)


lucene/analysis/kuromoji/src/java/org/apache/lucene/analysis/ja/JapaneseNumberFilter.java
```diff
@@ -218,6 +228,11 @@ public final boolean incrementToken() throws IOException {
         // capture the state of this token and emit it on our next incrementToken()
         state = captureState();
       }
+      // we restore state to when we read the last numeral token to get its attributes (e.g. part-of-speech)
+      if (lastNumeralTokenState != null) {
+        restoreState(lastNumeralTokenState);
```
cbuescher: Note: simply setting the PartOfSpeechAttribute to "noun-numeric" on the emited token wasn't as straight forward as I expected, since the implementation wraps a whole `org.apache.lucene.analysis.ja.Token`. This is why I explored tracking and restoring the last "good" tokens state here.

lucene/analysis/kuromoji/src/java/org/apache/lucene/analysis/ja/JapaneseNumberFilter.java
```diff
@@ -209,6 +216,9 @@ public final boolean incrementToken() throws IOException {
       if (moreTokens) {
         term = termAttr.toString();
         numeralTerm = isNumeral(term) || isNumeralPunctuation(term);
+        if (isNumeral(term)) {
+          // lastNumeralTokenState = captureState();
+        }
```
madrob: What does this block do?

cbuescher: Sorry, this shouldn't be left out, it's a left over from when I was doing some local debugging.
I'll push an update shortly. The classes javadocs say that the filter "will inherit the values of the last token used to compose the normalized number", and in order to do so, that state is saved here on every new numeral token encountered.

PR #1075, opened 2019-12-12
SOLR-14047: Make sure tests don't pickup other Hadoop installs

risdenk:
Checks that `HADOOP_HOME` environment variable isn't set and Java property `hadoop.home.dir` is not set prior to starting Hadoop tests. This will bail out with a nicer error message. It seems better than trying to much with environment variables or system properties.


risdenk:
Merged in 322ed8ce4d8729ad5230a65913b27f7285578a1e with CHANGES entry

risdenk:
Thanks for the review David

PR #1076, opened 2019-12-12
SOLR-14048: Improve Hadoop test sanity checks

risdenk:
* Added javadoc to `org.apache.hadoop` package
* Added a check to ensure that modified Hadoop classes are picked up before testing

### DIFF
--- a/solr/core/src/test/org/apache/solr/cloud/hdfs/HdfsTestUtil.java
+++ b/solr/core/src/test/org/apache/solr/cloud/hdfs/HdfsTestUtil.java
@@ -19,8 +19,10 @@ package org.apache.solr.cloud.hdfs;
 import java.io.File;
 import java.lang.invoke.MethodHandles;
 import java.net.URI;
+import java.util.Arrays;
 import java.util.Enumeration;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Timer;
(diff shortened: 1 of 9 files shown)


solr/core/src/test/org/apache/solr/cloud/hdfs/HdfsTestUtil.java
```diff
@@ -87,6 +99,23 @@ private static void checkHadoopWindows() {
         !Constants.WINDOWS || NativeIO.isAvailable());
   }
 
+  /**
+   * Ensure that the tests are picking up the modified Hadoop classes
+   */
+  private static void checkOverriddenHadoopClasses() {
+    List<Class<?>> modifiedHadoopClasses = Arrays.asList(BlockPoolSlice.class, DiskChecker.class,
```
dweiss: Hi Kevin. Wouldn't it be easier to check the codebase (clazz.getProtectionDomain().getCodeSource()) of these classes and compare it to a known class that should reside in the same codebase? If these are equal then it's "our" code and it wouldn't need those extra fields (or reflection). 


It may need an additional permission though.

risdenk: Hmm well I didn't know that was an option. I don't think adding the new field is really that bad. We already modified the classes enough adding a specific field that jumps out that we added it shouldn't matter.

The only case I can think of that wouldn't be caught is a new modified Hadoop class. I would bet that in 95% of cases, if one class is messed up on the classpath then all of them will be. SOLR-14047 should also provide a decent amount of protection.

dweiss: A field is fine, sure. I just though it'd be more elegant to keep it in one place :)

risdenk:
merged in 3ba005465a5dff3975b85f9c44d365bd3cd36346

PR #1077, opened 2019-12-12
SOLR-14069: Ref guide: overhaul: resources, libs, plugins, config-sets

dsmiley:
* split "resource-and-plugin-loading.adoc" into "resource-loading.adoc" and "libs.adoc" then overhauled both.
* enhanced "config-sets.adoc", moving some content in from elsewhere; bit of an overhaul.
* solr-plugins.adoc is now top-level; overhauled content

Did not touch plugin management stuff or old blob stuff.


dsmiley:
By the way, out of curiosity I tried to see if it was possible to load un-packaged Java classes from a configSet.  That did not work (good, I think).  I also verified that resources can be referenced out of .jar files and that the configSet takes precedence.

chatman:
Looks fantastic, @dsmiley!

solr/solr-ref-guide/src/config-sets.adoc
```diff
@@ -76,4 +82,16 @@ curl -v -X POST -H 'Content-type: application/json' -d '{
...
 
-In SolrCloud mode, you can use the <<configsets-api.adoc#configsets-api,Configsets API>> to manage your configsets.
+In SolrCloud, it's critical to understand that configsets are fundamentally stored in ZooKeeper _and not_ the file system.
+Solr's `_default` configset is uploaded to ZooKeeper on initialization.
+This and some demonstration ones remain on the file system but Solr does not use them whatsoever in this mode.
+
+When you create a collection in SolrCloud, you can specify a named configset -- possibly shared.
+If you don't, then the `_default` will be copied and given a unique name for use by this collection.
+
+A configset can be uploaded to ZooKeeper either via the <<configsets-api.adoc#configsets-api,Configsets API>> or more directly via <<solr-control-script-reference.adoc#upload-a-configuration-set,`bin/solr zk upload`>>.
```
janhoy: Think it is bin/solr zk upconfig

dsmiley: Good catch; I'm glad you are reading everything.  This error was copied from elsewhere; I fixed it.

solr/solr-ref-guide/src/configuring-solrconfig-xml.adoc
```diff
@@ -1,5 +1,16 @@
 = Configuring solrconfig.xml
-:page-children: datadir-and-directoryfactory-in-solrconfig, resource-and-plugin-loading, schema-factory-definition-in-solrconfig, indexconfig-in-solrconfig, requesthandlers-and-searchcomponents-in-solrconfig, initparams-in-solrconfig, updatehandlers-in-solrconfig, query-settings-in-solrconfig, requestdispatcher-in-solrconfig, update-request-processors, codec-factory
+:page-children: datadir-and-directoryfactory-in-solrconfig, \
```
dsmiley: @ctargett I felt it's easier to maintain page children like this.

solr/solr-ref-guide/src/configuring-solrconfig-xml.adoc
```diff
@@ -38,7 +49,6 @@ The `solrconfig.xml` file is located in the `conf/` directory for each collectio
 We've covered the options in the following sections:
 
 * <<datadir-and-directoryfactory-in-solrconfig.adoc#datadir-and-directoryfactory-in-solrconfig,DataDir and DirectoryFactory in SolrConfig>>
-* <<resource-and-plugin-loading.adoc#lib-directives-in-solrconfig,Lib Directives in SolrConfig>>
```
dsmiley: Whoops; I should put this back

solr/solr-ref-guide/src/solr-plugins.adoc
```diff
@@ -17,8 +20,36 @@
 // specific language governing permissions and limitations
 // under the License.
 
-Solr allows you to load custom code to perform a variety of tasks within Solr, from custom Request Handlers to process your searches, to custom Analyzers and Token Filters for your text field. You can even load custom Field Types. These pieces of custom code are called _plugins_.
+One of Solr's strengths is providing a rich platform of functionality with the option of adding your own custom components running within Solr.
+Solr calls such components *plugins* when the implementation is configurable.
```
ctargett: All our HTML pages treat the opening paragraph of a page as the "lede" and CSS makes it larger. For this reason we should avoid large blocks and try to open the page with a sentence that summarizes the feature/function if possible (it can be hard). In this case, the first sentence is rather perfect, so I'd suggest adding a new line here to separate the lede from the next batch of information which starts to get into the details.

solr/solr-ref-guide/src/configuring-solrconfig-xml.adoc
ctargett: OK, +1

ctargett:
I looked over these, and +1.

I left one comment on solr-plugins.adoc about separating the first sentence from the first paragraph from the rest of the paragraph because of the way we treat the opening paragraph of each page - both resource-loading.adoc and libs.adoc could use a similar change. That's mostly a suggestion, though, if you don't feel like making that change I don't think it should hold you up.

PR #1078, opened 2019-12-12
SOLR-14071: Untrusted configsets shouldn't be allowed to use <lib>

chatman:
Added a test for this.

### DIFF
--- a/solr/core/src/test/org/apache/solr/cloud/TestConfigSetsAPI.java
+++ b/solr/core/src/test/org/apache/solr/cloud/TestConfigSetsAPI.java
@@ -348,6 +348,7 @@ public class TestConfigSetsAPI extends SolrTestCaseJ4 {
   @Test
   public void testUploadWithScriptUpdateProcessor() throws Exception {
       // Authorization off
+      unprotectConfigsHandler();
       final String untrustedSuffix = "-untrusted";
       uploadConfigSetWithAssertions("with-script-processor", untrustedSuffix, null, null);
       // try to create a collection with the uploaded configset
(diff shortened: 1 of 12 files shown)


noblepaul:
+1

solr/core/src/test/org/apache/solr/cloud/TestConfigSetsAPI.java
```diff
@@ -369,13 +370,55 @@ public void testUploadWithScriptUpdateProcessor() throws Exception {
...
+    HttpClient cl = null;
+    try {
+      cl = HttpClientUtil.createClient(null);
+      zkClient().setData("/security.json", "{}".getBytes(UTF_8), true);
+    } finally {
+      if (cl != null) {
+        HttpClientUtil.close(cl);
+      }
+    }
+    Thread.sleep(5000); // TODO: Without a delay, the test fails. Some problem with Authc/Authz framework?
```
janhoy: Ok, obviously get rid of this :)

dweiss: Yes, please. The tests already take an hour.

chatman: Yeah, that is so irritating indeed. I remember we put that sleep of 5s in for the authentication to kick in. I'll reduce the sleep to 1s and see if it works or not. Please don't mind the potential noise on Jenkins due to this, if it fails.

dweiss: This won't work: some test machines are slower than others, etc. Please don't use wall time in tests. If something needs an async await then re-check the remote state every .5 seconds and expire the test if it doesn't pick up the required change within a reasonable time frame. There has to be a way to check whether the remote status has changed (right)?

chatman: Okay, I'll dig in to figure out. Maybe there is some way. All the authentication/authorization details are way past my memory.

But anyway, I have limited time today, and lots of issues to get to (before 8.4), please see SOLR-13442). There will be a compromise somewhere or the other unless someone can please help me.

chatman: Wasn't worth wasting my time today over this. I've ignored the test. Lets move on.
I've also added ref guide changes.

dweiss: Well, we're all trying to help one way or another; lack of time is no excuse to add code that will make other people wonder what the heck is going on later.

I have a strong temptation to ban Thread.sleep with forbidden APIs everywhere, including tests. An unconditional sleep is really worthless in a test -- those jenkins VMs are slow and loaded enough that those tests then proceed too soon and fail with false alarms. And on fast machines they make tests run slow when they could happily proceed. A test should be given just as much attention as the code it tests...

solr/solr-ref-guide/src/configsets-api.adoc
```diff
@@ -89,7 +89,7 @@ A configset is uploaded in a "trusted" mode if authentication is enabled and the
 * If specified in the configset, the DataImportHandler's ScriptTransformer will not initialize.
 * The XSLT transformer (`tr` parameter) cannot be used at request processing time.
 * If specified in the configset, the StatelessScriptUpdateProcessor will not initialize.
-* Collections won't initialize if <lib> directives are used in the configset. If you need to include libraries, please add them to the Solr's classpath.
+* Collections won't initialize if <lib> directives are used in the configset. (Note: Libraries added to Solr's classpath don't need the <lib> directive)
```
janhoy: Could we add a link here to the new docs that @dsmiley is about to add? The one where it lists the places you can put libs.

solr/core/src/test/org/apache/solr/cloud/TestConfigSetsAPI.java
chatman: > Well, we're all trying to help one way or another; lack of time is no excuse to add code that will make other people wonder what the heck is going on later.
I agree. That's why I removed the new test. By the way, a 5 second sleep was already there in another test.

> I have a strong temptation to ban Thread.sleep with forbidden APIs everywhere
Totally agree. I want to do that myself.

solr/solr-ref-guide/src/configsets-api.adoc
chatman: Good idea, Jan. @dsmiley Please "do the needful" ;-)
I'm going to merge this to master and branch_8x now.

solr/core/src/test/org/apache/solr/cloud/TestConfigSetsAPI.java
chatman: > lack of time is no excuse to add code that will make other people wonder what the heck is going on later.

Also, I wasn't trying to be so cruel to other people. I had left a TODO comment to help them understand "what the heck is going on". :-)

dweiss: A TODO doesn't help if all you see is an exception/ failure in jenkins logs.

PR #1082, opened 2019-12-14
SOLR-13984: add (experimental, disabled by default) security manager support

rmuir:
The current policy file used by tests is moved to solr/server
Additional permissions are granted for the filesystem locations set by bin/solr, and networking everywhere is enabled.

This takes advantage of the fact that permission entries are ignored if properties are not defined:
https://docs.oracle.com/javase/7/docs/technotes/guides/security/PolicyFiles.html#PropertyExp


solr/bin/solr.cmd
```diff
@@ -994,7 +994,7 @@ set "EXAMPLE_DIR=%SOLR_TIP%\example"
 set TMP_SOLR_HOME=!SOLR_HOME:%EXAMPLE_DIR%=!
 IF NOT "%TMP_SOLR_HOME%"=="%SOLR_HOME%" (
   set "SOLR_LOGS_DIR=%SOLR_HOME%\..\logs"
-  set "LOG4J_CONFIG=file:///%SOLR_SERVER_DIR%\resources\log4j2.xml"
+  set "LOG4J_CONFIG=%SOLR_SERVER_DIR%\resources\log4j2.xml"
```
rmuir: You can't make URLs of things unless you encode them, if there are spaces in `SOLR_SERVER_DIR` this is all messed up. So i fixed it to use a standard path instead: https://logging.apache.org/log4j/log4j-2.1/faq.html#config_location

Looks like the same bug exists in `bin/solr` too, i'll make a space on os X and try it out.

rmuir:
This is working with the examples and so on on *nix and windows, tested with spaces in directory in both cases :)

solr/bin/solr.in.sh
```diff
@@ -203,3 +203,8 @@
 # a -Dsolr.environment property below. Valid values are prod, stage, test, dev, with an optional
 # label or color, e.g. -Dsolr.environment=test,label=Functional+test,color=brown
 #SOLR_OPTS="$SOLR_OPTS -Dsolr.environment=prod"
+
+# Runs solr in java security manager sandbox. This can protect against some attacks.
+# Runtime properties are passed to the security policy file (server/etc/security.policy)
+# This is experimental! It may not work at all with HDFS.
```
risdenk: > It may not work at all with HDFS.

This may not work with any Hadoop related usages. This includes backup/restore to HDFS compatible filesystems (https://lucene.apache.org/solr/guide/8_3/making-and-restoring-backups.html), storing indices on HDFS (https://lucene.apache.org/solr/guide/8_3/running-solr-on-hdfs.html), and Hadoop authentication plugin (https://lucene.apache.org/solr/guide/8_3/hadoop-authentication-plugin.html)

rmuir: Agreed, lets generalize the 'hdfs' here to all hadoop uses.

risdenk:
```
Uncaught exception org.apache.solr.common.SolrException: java.security.AccessControlException: access denied ("java.lang.RuntimePermission" "accessClassInPackage.jdk.internal.reflect") thrown by thread: OverseerThreadFactory-9-thread-1-processing-n:192.168.1.124:8983_solr

**Comment by rmuir** [comment]
Let's open a separate issue for that one. It is really an unrelated issue I think. Java serialization here across the network? What could go wrong :)

**Comment by uschindler** [comment]
I fully agree with Robert. IMHO, ObjectIn/OutputStream should be forboddenapis. It's good that this was catched here. This must really go away. You should never ever send serialized objects with remote streams. This is actually a security issue, a POC should be trivial for any experienced beginner hacker.

**Comment by dweiss** [comment]
bq. a POC should be trivial for any minimally experienced beginner hacker.

I love you, Uwe. :)

**Comment by rmuir** [comment]
i will open issues to fix this after i am done drinking.

**Comment by rmuir** [comment]
https://issues.apache.org/jira/browse/LUCENE-9094
https://issues.apache.org/jira/browse/LUCENE-9095
https://issues.apache.org/jira/browse/SOLR-14095

**Comment by rmuir** [comment]
I merged master with @tflobbe fix for SOLR-14095 and updated the branch here. I repeated the steps @risdenk gave and it works now. Thank you Tomas!

File: solr/bin/solr [hunk 3]
```diff
@@ -2071,6 +2071,17 @@ else
   REMOTE_JMX_OPTS=()
 fi
 
+# Enable java security manager (limiting filesystem access and other things)
+# Don't cache DNS lookups forever, set the value back to the original JDK default.
+if [ "$SOLR_SECURITY_MANAGER_ENABLED" == "true" ]; then
+  SECURITY_MANAGER_OPTS=('-Djava.security.manager' \
+    "-Djava.security.policy=${SOLR_SERVER_DIR}/etc/security.policy" \
+    '-Dsun.net.inetaddr.ttl=30' \
```
shalinmangar: Shouldn't this be configurable by the user? Perhaps we keep this default but make it configurable with a `SOLR_NETWORKADDRESS_CACHE_TTL` variable.

Although most (all?) JVMs still honor the `sun.net.inetaddr.ttl` but it is officially deprecated so we should use `-Dnetworkaddress.cache.ttl`

rmuir: {quote}
Shouldn't this be configurable by the user? Perhaps we keep this default but make it configurable with a SOLR_NETWORKADDRESS_CACHE_TTL variable.
{quote}

It is not configurable today: I'm not trying to add that feature here. I am just preventing dns from cached indefinitely.

{quote}
Although most (all?) JVMs still honor the sun.net.inetaddr.ttl but it is officially deprecated so we should use -Dnetworkaddress.cache.ttl
{quote}

nope: `networkaddress.cache.ttl` is not a system property at all.

rmuir: @shalinmangar i added a commit with a custom `java.security.properties` file, to override this via the `networkaddress.cache.ttl`, setting it back to the default of 30 (same as when no security manager is active). It does the same thing as before, just in a more roundabout way...

If users want to mess with java security properties, they must edit java security properties files either way: no variables can be supported since they are not system properties and can't be set with `-D`.

PR #1084, opened 2019-12-14
SOLR-14066 Deprecate DataImportHandler

janhoy:
# Description

See https://issues.apache.org/jira/browse/SOLR-14066


janhoy:
Link with SOLR-14066

janhoy:
> Should we also print a WARN log about the same for those who missed reading the upgrade notes?

There is a WARN printed to logs during DIH initialization...

janhoy:
We'll not deprecate in 8.4 but take another approach. Closing.

PR #1085, opened 2019-12-15
SOLR-14087 disable package store API if -Denable.packages not set to true

noblepaul:
I wish this to be included in 8.4. There are so many security holes in Solr and we want to play extra safe


chatman:
+1, looks good. Please merge asap!

PR #1086, opened 2019-12-15
SOLR-14072: Deprecate Blob API and runtimeLib

dsmiley:
Bare minimum.

### DIFF
--- a/solr/CHANGES.txt
+++ b/solr/CHANGES.txt
@@ -183,8 +183,8 @@ Upgrade Notes
 
 * SOLR-14065: VelocityResponseWriter has been deprecated and may be removed in a future version.
 
-* SOLR-14072: The "Blob Store" API and "runtimeLib" plugin mechanism that uses it is now considered deprecated.  The
-  replacement to it is the "Package Manager" system, which includes a "File Store".  These are experimental currently
+* SOLR-14072: The "Blob Store" API and "runtimeLib" plugin mechanism that uses it is now considered deprecated. The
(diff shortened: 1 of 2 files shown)


noblepaul:
+1

solr/CHANGES.txt
```diff
@@ -183,6 +183,10 @@ Upgrade Notes
 
 * SOLR-14065: VelocityResponseWriter has been deprecated and may be removed in a future version.
 
+* SOLR-14072: The "Blob Store" API and "runtimeLib" plugin mechanism that uses it is now considered deprecated.  The
+  replacement to it is the "Package Manager" system, which includes a "File Store".  These are experimental currently
```
chatman: Lets call it "package store".

solr/CHANGES.txt
```diff
@@ -183,6 +183,10 @@ Upgrade Notes
 
 * SOLR-14065: VelocityResponseWriter has been deprecated and may be removed in a future version.
 
+* SOLR-14072: The "Blob Store" API and "runtimeLib" plugin mechanism that uses it is now considered deprecated.  The
```
chatman: Should we even deprecare the blob store API itself? I think Metrics system uses it. @sigram any thoughts/ideas?

solr/CHANGES.txt
chatman: Should we call it Package Management system? Also, I think there are two spaces before "These".

solr/CHANGES.txt
chatman: I'm +1 to deprecating it, though. Please go ahead with merge.

PR #1088, opened 2019-12-16
SOLR-14101: Move org.apache.solr.util.TimeOut to SolrJ

noblepaul:


anshumg:
I think this is good to be merged with a CHANGELOG entry. @noblepaul ?

PR #1089, opened 2019-12-17
LUCENE-9097: Decrease the number of shape used in multi-polygon tests

iverase:
Decrease the number so test run faster and minimise the possibility of OOM errors.


iverase:
This was taken care on LUCENE-9167.

PR #1090, opened 2019-12-17
Update README.txt for analysis-extras

tuftii:
Update the analysis-extras README to include reference to including solr-analysis-extras jar.

I struggled to work out what was missing when I was using ICU for the first time, and this caught me out. There may be a better way of phrasing it.


janhoy:
Thanks for the contribution

PR #1094, opened 2019-12-17
SOLR-14106: Cleanup Jetty SslContextFactory usage

risdenk:
SOLR-14106: Cleanup Jetty SslContextFactory usage

Jetty 9.4.16.v20190411 and up introduced separate
client and server SslContextFactory implementations.
This split requires the proper use of of
SslContextFactory in clients and server configs.

This fixes the following
* SSL with SOLR_SSL_NEED_CLIENT_AUTH not working since v8.2.0
* Http2SolrClient SSL not working in branch_8x

### DIFF
--- a/lucene/replicator/src/test/org/apache/lucene/replicator/ReplicatorTestCase.java
+++ b/lucene/replicator/src/test/org/apache/lucene/replicator/ReplicatorTestCase.java
@@ -61,7 +61,7 @@ public abstract class ReplicatorTestCase extends LuceneTestCase {
     // talking to that server, but for the purposes of testing that should 
     // be good enough
     final boolean useSsl = Boolean.getBoolean("tests.jettySsl");
-    final SslContextFactory sslcontext = new SslContextFactory(false);
+    final SslContextFactory.Server sslcontext = new SslContextFactory.Server();
     
     if (useSsl) {
       if (null != System.getProperty("javax.net.ssl.keyStore")) {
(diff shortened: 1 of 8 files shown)


janhoy:
I still get the stack trace from https://issues.apache.org/jira/browse/SOLR-14105 when applying your patch.

I think `Http2SolrClient.getDefaultSslContextFactory()` returns a Client instead of Server, which then crashes with the stack shown.

risdenk:
Thanks for reviewing @janhoy. I fixed the deprecation warning. I'm going to need to spend more time looking at the `Http2SolrClient` stuff since it is pretty intertwined.

risdenk:
@janhoy and @ErickErickson pretty sure this is ready for review. Sanity checks locally all looked good. Let me know what you think :)

risdenk:
> We should probably have more tests. Perhaps a test similar to SolrCloudExampleTest but which starts Solr with SSL and validates that it comes up? We can open another JIRA for that.

There is https://github.com/apache/lucene-solr/blob/master/solr/core/src/test/org/apache/solr/cloud/TestMiniSolrCloudClusterSSL.java which I think covers what you are asking for?

risdenk:
Thanks for testing Jan!

risdenk:
Merged in aab3c5faa338e158bcab2fb5f09b309ee455a4c5

janhoy:
> There is https://github.com/apache/lucene-solr/blob/master/solr/core/src/test/org/apache/solr/cloud/TestMiniSolrCloudClusterSSL.java which I think covers what you are asking for?

So why did we not see test failures for branch_8x? I guess on MacOS SSL would never be enabled due to this other issue, so that was not helpful either.

risdenk:
> So why did we not see test failures for branch_8x? I guess on MacOS SSL would never be enabled due to this other issue, so that was not helpful either.

Thats definitely a valid question and I don't know. I have a few theories about maybe different test Jetty config vs regular Jetty config. Maybe clients aren't properly doing TLS at all in tests? I didn't look into it a ton.

PR #1095, opened 2019-12-17
SOLR-14095: Let the overseer use javabin to store responses in ZooKeeper

tflobbe:
The Overseer currently uses java serialization to store command responses in ZooKeeper. This commit changes the code to use Javabin instead, while allowing Java serialization with a System property in case it's needed for compatibility.
This PR is for merge into master and branch_8. I'll have a new PR after this is merged to remove the back compat code from master.

### DIFF
--- a/solr/core/src/java/org/apache/solr/cloud/OverseerSolrResponse.java
+++ b/solr/core/src/java/org/apache/solr/cloud/OverseerSolrResponse.java
@@ -55,6 +55,13 @@ public class OverseerSolrResponse extends SolrResponse {
     return responseList;
   }
 
+  /**
+   * This method serializes the content of an {@code OverseerSolrResponse}. Note that:
+   * <ul>
+   * <li>The elapsed time is not serialized</li>
+   * <li>"Unknown" elements for the Jababin format will be serialized as Strings. See {@link org.apache.solr.common.util.JavaBinCodec#writeVal}</li>
+   * </ul>
+   */
   @SuppressWarnings("deprecation")
   public static byte[] serialize(OverseerSolrResponse responseObject) {
     Objects.requireNonNull(responseObject);
--- a/solr/core/src/java/org/apache/solr/cloud/OverseerSolrResponse.java
+++ b/solr/core/src/java/org/apache/solr/cloud/OverseerSolrResponse.java
@@ -17,11 +17,17 @@
 package org.apache.solr.cloud;
 
 import org.apache.solr.client.solrj.SolrResponse;
+import org.apache.solr.common.SolrException;
+import org.apache.solr.common.SolrException.ErrorCode;
 import org.apache.solr.common.util.NamedList;
+import org.apache.solr.common.util.Utils;
+
+import java.io.IOException;
+import java.util.Objects;
 
 public class OverseerSolrResponse extends SolrResponse {
   
-  NamedList responseList = null;
+  NamedList<Object> responseList = null;
 
   private long elapsedTime;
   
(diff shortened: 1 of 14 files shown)


solr/core/src/java/org/apache/solr/cloud/OverseerSolrResponse.java
```diff
@@ -48,5 +54,40 @@ public void setElapsedTime(long elapsedTime) {
   public NamedList<Object> getResponse() {
     return responseList;
   }
+
+  /**
+   * This method serializes the content of an {@code OverseerSolrResponse}. Note that:
+   * <ul>
+   * <li>The elapsed time is not serialized</li>
+   * <li>"Unknown" elements for the Jababin format will be serialized as Strings. See {@link org.apache.solr.common.util.JavaBinCodec#writeVal}</li>
```
rmuir: typo: Jababin

solr/core/src/java/org/apache/solr/cloud/OverseerSolrResponse.java
```diff
@@ -48,5 +54,40 @@ public void setElapsedTime(long elapsedTime) {
...
+   * </ul>
+   */
+  @SuppressWarnings("deprecation")
+  public static byte[] serialize(OverseerSolrResponse responseObject) {
+    Objects.requireNonNull(responseObject);
+    if (Boolean.getBoolean("solr.unsafeOverseerResponseSerilization")) {
```
rmuir: lets fix the spelling for the property to be "Serialization"

solr/core/src/test/org/apache/solr/cloud/OverseerSolrResponseUnsafeSerializationTest.java
```diff
@@ -0,0 +1,36 @@
...
+
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
+
+public class OverseerSolrResponseUnsafeSerializationTest extends OverseerSolrResponseTest {
+  
+  @BeforeClass
+  public static void setUpClass() {
+    System.setProperty("solr.unsafeOverseerResponseSerilization", "true");
+    System.setProperty("solr.unsafeOverseerResponseDeserilization", "true");
```
rmuir: Is there a use-case for two separate properties or should it be folded into one?

noblepaul:
Please do not use a binary format. The optimization is not worth the complexity. Stick to JSON

solr/core/src/test/org/apache/solr/cloud/OverseerSolrResponseUnsafeSerializationTest.java
tflobbe: I originally had a single property, but then I was thinking of the rolling restart case. We have the Overseer writing and other nodes consuming this data. If it's a single property there is no way to make the Overseer start writing Javabin while the consumers can consume both. An alternative is to only have `solr.unsafeOverseerResponseDeserilaization`, the catch there is that the Overseer needs to be the last one to be upgraded, otherwise it'll start writing a format that others may not be able to read. Other idea?

dweiss: You could have a single property with a string value. -Dsolr.unsafeOverseerResponse=[serialization,deserialization,all]

Or even parse a comma-separated list of values; this would take all out of the list.

tflobbe: I don't think "serialization" makes a ton of sense. I'll have two options "true" (which turns on serialization and deserialization) and "deserialization", which turns on deserialization only.

PR #1099, opened 2019-12-18
SOLR-14111: Revert SOLR13541 which breaks SSL client auth

janhoy:
See https://issues.apache.org/jira/browse/SOLR-14111 for details


janhoy:
Won't fix. Instead we'll backport the "real" fix.

PR #1100, opened 2019-12-18
SOLR-13579: Create resource management API

sigram:
Please see JIRA for details.


solr/core/src/java/org/apache/solr/managed/ResourceManagerPool.java
```diff
@@ -0,0 +1,216 @@
...
+    listeners.remove(listener);
+  }
+
+
+  /**
+   * Get the current monitored values from all resources. Result is a map with resource names as keys,
+   * and param/value maps as values.
+   */
+  public Map<String, Map<String, Object>> getCurrentValues() throws InterruptedException {
+    updateLock.lockInterruptibly();
```
shalinmangar: I'm having trouble understanding why the updateLock is necessary in this method? It is iterating over the values of a concurrent hash map (`components`) and reading the id and monitored value of the component. Both seem to be thread-safe to me.

shalinmangar: The updateLock is acquired only in getCurrentValues() (called by doManage and ResourceManagerHandler.STATUS) and in manage() but manage() is called only by a scheduled executor which guarantees a happens-before relationship to previous executions. I think we can get rid of the lock entirely?

solr/core/src/java/org/apache/solr/managed/types/CacheManagerPool.java
```diff
@@ -0,0 +1,329 @@
...
+    // (because a lower hit ratio is still acceptable if it means saving resources) and
+    // expand the size of caches with too low hitRatio
+    final AtomicReference<Double> newTotalValue = new AtomicReference<>(totalValue);
+    components.forEach(component -> {
+      long currentLookups = ((Number)currentValues.get(component.getManagedComponentId().toString()).get(SolrCache.LOOKUPS_PARAM)).longValue();
+      long lastLookups = lookups.computeIfAbsent(component.getManagedComponentId().toString(), k -> 0L);
+      if (currentLookups < lastLookups + lookupDelta) {
+        // too little data, skip the optimization
+        return;
+      }
```
shalinmangar: What if a commit happened between the last adjustment and now and therefore the current value of `lookups` has been reset?

shalinmangar: Perhaps we can use cumulative lookups to check delta calculations but use lookups to make sure a minimum number of lookups have actually happened to give us meaningful hit ratio data

solr/core/src/java/org/apache/solr/managed/types/CacheManagerPool.java
```diff
@@ -0,0 +1,329 @@
...
+        // can't optimize because we don't know how far off we are from the initial setting
+        return;
+      }
+      if (currentHitRatio < targetHitRatio) {
+        if (changeRatio.get() < 1.0) {
+          // don't expand if we're already short on resources
+          return;
+        }
+        // expand to increase the hitRatio, but not more than maxAdjustRatio from the initialLimit
+        double newLimit = currentLimit * changeRatio.get();
```
shalinmangar: Shouldn't the limit be adjusted only if there are evictions because of the old limit? Otherwise we are increasing the limits for no benefit

solr/core/src/java/org/apache/solr/handler/admin/ResourceManagerHandler.java
```diff
@@ -0,0 +1,334 @@
...
+    }
+    String poolAction = req.getParams().get(POOL_ACTION_PARAM);
+    String resAction = req.getParams().get(RES_ACTION_PARAM);
+    if (poolAction != null) {
+      handlePoolRequest(req, rsp);
+    } else if (resAction != null) {
+      handleResourceRequest(req, rsp);
+    } else {
+      throw new SolrException(SolrException.ErrorCode.BAD_REQUEST, "Missing " + POOL_ACTION_PARAM + " and " + RES_ACTION_PARAM + ": " + req.getParams());
+    }
```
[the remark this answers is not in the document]
sigram: Good point, I think we should throw an exception.

solr/core/src/java/org/apache/solr/managed/ResourceManagerPool.java
sigram: Yes, it's a left-over from a previous refactoring.

solr/core/src/java/org/apache/solr/managed/types/CacheManagerPool.java
sigram: If the lookups has been reset due to commit we should simply skip any optimization - because we really don't know how useful is the content of the cache at the moment.

Using cumulative stats is a double-edged sword, specifically because they are never reset... Take for example cumulative hit ratio - it's calculated using cumulative_lookups and cumulative_hits. This means that short-term fluctuations are averaged, but it also means that short-term effects of any adjustments are also averaged (ie. they don't seem to affect the hit ratio in short term), which during thee next round of optimization will cause more and more adjustments, and the controller will eventually grossly overshoot its target. This issue becomes more and more serious as the absolute numbers increase because it's more and more difficult to sway this average.

solr/core/src/java/org/apache/solr/managed/types/CacheManagerPool.java
sigram: Good point, fixed.

solr/core/src/java/org/apache/solr/managed/DefaultResourceManager.java
```diff
@@ -0,0 +1,231 @@
...
+
+  private TimeSource timeSource;
+
+  /**
+   * Thread pool for scheduling the pool runs.
+   */
+  private ScheduledThreadPoolExecutor scheduledThreadPoolExecutor;
+
+  protected boolean isClosed = false;
+  protected boolean enabled = true;
```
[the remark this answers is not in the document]
sigram: This is modified in superclass `setEnabled(Boolean)`, which in turn is invoked by `PluginUtils.invokeSetters`.

solr/core/src/java/org/apache/solr/managed/DefaultResourceManager.java
```diff
@@ -0,0 +1,231 @@
...
+
+
+  private TimeSource timeSource;
+
+  /**
+   * Thread pool for scheduling the pool runs.
+   */
+  private ScheduledThreadPoolExecutor scheduledThreadPoolExecutor;
+
+  protected boolean isClosed = false;
```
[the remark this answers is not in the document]
sigram: Fixed.

solr/core/src/java/org/apache/solr/managed/DefaultResourceManager.java
```diff
@@ -0,0 +1,231 @@
...
+
+  @Override
+  public void close() throws IOException {
+    synchronized (this) {
+      isClosed = true;
+      log.debug("Closing all pools.");
+      for (ResourceManagerPool pool : resourcePools.values()) {
+        IOUtils.closeQuietly(pool);
+      }
+      resourcePools.clear();
```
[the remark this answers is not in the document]
shalinmangar: This is still not addressed. Why is the synchronized block required here?

solr/core/src/java/org/apache/solr/managed/ResourceManager.java
```diff
@@ -0,0 +1,305 @@
...
+      throw new IllegalStateException("Not initialized.");
+    }
+  }
+
+  @Override
+  public void close() throws Exception {
+    synchronized (this) {
+      isClosed = true;
+      SolrMetricProducer.super.close();
+    }
```
shalinmangar: The usage of isClosed and initialized in this class is not thread safe. Different threads read and set these values so the latest values are not guaranteed to be visible. Both can be made AtomicBoolean instances and then the synchronized(this) block here is not needed at all.

solr/core/src/java/org/apache/solr/managed/ResourceManagerPool.java
```diff
@@ -0,0 +1,294 @@
...
+
+  /**
+   * Pool context used for managing additional pool state.
+   */
+  public ResourcePoolContext getResourcePoolContext() {
+    return poolContext;
+  }
+
+  public void manage() {
+    manageLock.lock();
```
shalinmangar: I think we can get rid of the manageLock completely. The manage method is only called by the scheduled executor which guarantees that the effect of previous executions happens-before the subsequent ones.

solr/core/src/java/org/apache/solr/managed/ResourceManager.java
sigram: Fixed.

PR #1103, opened 2019-12-19
LUCENE-9102: add maxQueryLength option to DirectSpellChecker

andywebb1975:
# Description

Attempting to spellcheck some long query terms can trigger org.apache.lucene.util.automaton.TooComplexToDeterminizeException.

(See also https://github.com/apache/lucene-solr/pull/1107 for a corresponding Solr change.)

# Solution

This change adds a maxQueryLength option to DirectSpellchecker so that Lucene can be configured to not attempt to spellcheck terms over a specified length.

# Tests

A new test checks that terms longer than maxQueryLength are not spellchecked.

# Checklist

### DIFF
--- a/lucene/suggest/src/java/org/apache/lucene/search/spell/DirectSpellChecker.java
+++ b/lucene/suggest/src/java/org/apache/lucene/search/spell/DirectSpellChecker.java
@@ -75,6 +75,8 @@ public class DirectSpellChecker {
   private float thresholdFrequency = 0f;
   /** minimum length of a query word to return suggestions */
   private int minQueryLength = 4;
+  /** maximum length of a query word to return suggestions */
+  private int maxQueryLength = 0;
   /** value in [0..1] (or absolute number &gt;= 1) representing the maximum
    *  number of documents (of the total) a query term can appear in to
    *  be corrected. */
@@ -198,6 +200,19 @@ public class DirectSpellChecker {
     this.minQueryLength = minQueryLength;
   }
 
+  /** Get the maximum length of a query term to return suggestions */
+  public int getMaxQueryLength() {
+    return maxQueryLength;
+  }
+
+  /** 
+   * Set the maximum length of a query term (default: 0, i.e. no maximum)
+   * to return suggestions. 
+   */
+  public void setMaxQueryLength(int maxQueryLength) {
+    this.maxQueryLength = maxQueryLength;
+  }
+
   /**
    * Get the maximum threshold of documents a query term can appear in order
    * to provide suggestions.
@@ -317,7 +332,11 @@ public class DirectSpellChecker {
       SuggestMode suggestMode, float accuracy) throws IOException {
     final CharsRefBuilder spare = new CharsRefBuilder();
     String text = term.text();
-    if (minQueryLength > 0 && text.codePointCount(0, text.length()) < minQueryLength)
+
+    int textLength = text.codePointCount(0, text.length());
+    if (minQueryLength > 0 && textLength < minQueryLength)
+      return new SuggestWord[0];
+    if (maxQueryLength > 0 && textLength > maxQueryLength)
       return new SuggestWord[0];
     
     if (lowerCaseTerms) {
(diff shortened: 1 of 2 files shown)


lucene/suggest/src/java/org/apache/lucene/search/spell/DirectSpellChecker.java
```diff
@@ -75,6 +75,8 @@
   private float thresholdFrequency = 0f;
   /** minimum length of a query word to return suggestions */
   private int minQueryLength = 4;
+  /** maximum length of a query word to return suggestions */
+  private int maxQueryLength = 0;
```
madrob: Do we want validation somewhere that max >= min? Or simply treat the max < min case as ignoring max?

lucene/suggest/src/java/org/apache/lucene/search/spell/DirectSpellChecker.java
```diff
@@ -198,6 +200,19 @@ public void setMinQueryLength(int minQueryLength) {
...
 
+  /** Get the maximum length of a query term to return suggestions */
+  public int getMaxQueryLength() {
+    return maxQueryLength;
+  }
+
+  /** 
+   * Set the maximum length of a query term (default: 0, i.e. no maximum)
+   * to return suggestions. 
+   */
```
madrob: Can you add an explanation here on why this would be useful, i.e. too long terms will eat lots of memory.

lucene/suggest/src/java/org/apache/lucene/search/spell/DirectSpellChecker.java
bruno-roustant: I prefer to initialize maxQueryLength to Integer.MAX_VALUE, it's more intuitive, so that there is no special <= 0 or < min value.

lucene/suggest/src/java/org/apache/lucene/search/spell/DirectSpellChecker.java
```diff
@@ -317,7 +332,11 @@ public void setDistance(StringDistance distance) {
       SuggestMode suggestMode, float accuracy) throws IOException {
     final CharsRefBuilder spare = new CharsRefBuilder();
     String text = term.text();
-    if (minQueryLength > 0 && text.codePointCount(0, text.length()) < minQueryLength)
+
+    int textLength = text.codePointCount(0, text.length());
+    if (minQueryLength > 0 && textLength < minQueryLength)
+      return new SuggestWord[0];
+    if (maxQueryLength > 0 && textLength > maxQueryLength)
```
bruno-roustant: Here we could group the "if" check, remove the maxQueryLength > 0 condition (with maxQueryLength initialized to Integer.MAX_VALUE), and have only one return empty array because this is the same semantic (and add brackets).

lucene/suggest/src/java/org/apache/lucene/search/spell/DirectSpellChecker.java
andywebb1975: Thanks for the comments - I've put in validation and switched to using Integer.MAX_VALUE.

lucene/suggest/src/java/org/apache/lucene/search/spell/DirectSpellChecker.java
andywebb1975: Yes to both!

lucene/suggest/src/java/org/apache/lucene/search/spell/DirectSpellChecker.java
andywebb1975: I've added a second line - happy to take others' input here if there are better words to use!

lucene/suggest/src/java/org/apache/lucene/search/spell/DirectSpellChecker.java
```diff
@@ -195,9 +197,27 @@ public int getMinQueryLength() {
    * metric.
    */
   public void setMinQueryLength(int minQueryLength) {
+    if (minQueryLength > this.maxQueryLength)
```
bruno-roustant: Actually I think we don't need these checks because whatever the value for minQueryLength and maxQueryLength, the logic is always the same, no surprise.
So personally I would remove these checks, but I'm ok to keep them, your choice.

PR #1104, opened 2019-12-20
SOLR-14129: Reuse Jackson ObjectMapper in AuditLoggerPlugin

janhoy:
See https://issues.apache.org/jira/browse/SOLR-14129


janhoy:
@chatman can you have a look?

PR #1105, opened 2019-12-20
LUCENE-9105: UniformSplit postings format detects corrupted index

bruno-roustant:
and better handles IO exceptions.

To correctly handle IOException, we do not use Supplier<IndexDictionary.Browser> anymore, it is replaced by IndexDictionary.BrowserSupplier which throws (propagates) IOException.
This leads to some disrupting refactoring: method signature change, and one class IndexDictionaryBrowserSupplier moved as an inner class FSTDictionary.BrowserSupplier.

I think it does not prevent us to push this change to 8x as I don't think UniformSplit is already extended by the community (was released recently in 8.3).


lucene/codecs/src/java/org/apache/lucene/codecs/uniformsplit/IndexDictionary.java
```diff
@@ -97,31 +100,47 @@
...
+
+  /**
+   * Supplier for a new stateful {@link Browser} created on the immutable {@link IndexDictionary}.
+   * <p>
+   * The immutable {@link IndexDictionary} is lazy loaded thread safely. This lazy loading allows
+   * us to load it only when {@link org.apache.lucene.index.TermsEnum#seekCeil} or
+   * {@link org.apache.lucene.index.TermsEnum#seekExact} are called (it is not loaded for a direct
+   * all-terms enumeration).
+   */
+  interface BrowserSupplier extends Accountable {
```
dsmiley: See `org.apache.lucene.util.IOSupplier` which this basically duplicates.  I'm not sure we need to mandate Accountable; it could be optionally implemented by an implementation?

bruno-roustant: Thanks. I looked externally for such supplier but I didn't think of looking in Lucene util...
Accountable is leveraged in UniformSplitTerms.getDictionaryRamBytesUsed() when computing the heap space. We could indeed put there an instanceof check and optionally cast to Accountable. But I'd prefer to mandate Accountable so that any implementation doesn't forget to provide it.
What about keeping BrowserSupplier to join IOSupplier and Accountable (without additional interface contract)?

dsmiley: > What about keeping BrowserSupplier to join IOSupplier and Accountable (without additional interface contract)?

+1

PR #1106, opened 2019-12-20
LUCENE-9106: UniformSplit postings format allows extension of block/line serializers.

bruno-roustant:
There are 4 public method signature changes in UniformSplitTermsReader, UniformSplitTermsWriter and their shared-terms extensions. However this seems to me it is still ok to push that to 8x.


lucene/codecs/src/java/org/apache/lucene/codecs/uniformsplit/UniformSplitTermsReader.java
```diff
@@ -68,16 +68,18 @@
    *                     It can be used for decompression or decryption.
    */
   public UniformSplitTermsReader(PostingsReaderBase postingsReader, SegmentReadState state, BlockDecoder blockDecoder) throws IOException {
-    this(postingsReader, state, blockDecoder, NAME, VERSION_START, VERSION_CURRENT,
+    this(postingsReader, state, blockDecoder, new FieldMetadata.Serializer(), NAME, VERSION_START, VERSION_CURRENT,
```
dsmiley: Since it appears the default instance has no state, can't we re-use a shared static instance of this one?

bruno-roustant: Good point.

PR #1108, opened 2019-12-20
SOLR-14125 : Streaming expressions to be loadable from packages

noblepaul:


noblepaul:
It's in a good shape. reviews are welcome

solr/solrj/src/java/org/apache/solr/client/solrj/io/stream/expr/StreamFactory.java
```diff
@@ -24,8 +24,8 @@
 import java.util.HashMap;
 import java.util.List;
 import java.util.Locale;
-import java.util.Map;
 import java.util.Map.Entry;
+import java.util.function.Supplier;
```
joel-bernstein: Why the use of the Supplier? Is this a cosmetic change or is there some functionality here that I'm not seeing? I'm not against a cosmetic change but changes to StreamFactory are going to slow down landing this patch quite a bit because it's such an important class for the Streaming Expressions. So if it's purely cosmetic and we want to land the other changes faster, let's do this in another ticket.

noblepaul: In the new scheme of things you can't hold direct reference to a `class`. 

If a plugin can be reloaded, it's best to just hold an indirect reference. The `Supplier` makes it cleaner. For the in-built classes, the supplier is just a simple wrapper. But for the plugins loaded from packages, it is listens to package changes and gives  the right class on-demand

joel-bernstein: Ok, this makes sense.

noblepaul:
SOLR-14125: Make <expressible> plugins work with packages

PR #1109, opened 2019-12-20
More pervasive use of PackageLoader / PluginInfo

dsmiley:
WIP don't merge.
This is what I did during my "hack day".  I'd be happy to discuss.
CC @noblepaul @chatman

### DIFF
--- a/solr/contrib/velocity/src/java/org/apache/solr/response/VelocityResponseWriter.java
+++ b/solr/contrib/velocity/src/java/org/apache/solr/response/VelocityResponseWriter.java
@@ -45,6 +45,7 @@ import org.apache.solr.client.solrj.response.SolrResponseBase;
 import org.apache.solr.common.SolrException;
 import org.apache.solr.common.params.CommonParams;
 import org.apache.solr.common.util.NamedList;
+import org.apache.solr.core.PluginInfo;
 import org.apache.solr.core.SolrCore;
 import org.apache.solr.request.SolrQueryRequest;
 import org.apache.solr.util.plugin.SolrCoreAware;
@@ -275,7 +276,7 @@ public class VelocityResponseWriter implements QueryResponseWriter, SolrCoreAwar
         for (Map.Entry<String, String> entry : customTools.entrySet()) {
           String name = entry.getKey();
           // TODO: at least log a warning when one of the *fixed* tools classes is same name with a custom one, currently silently ignored
-          Object customTool = SolrCore.createInstance(entry.getValue(), Object.class, "VrW custom tool: " + name, request.getCore(), request.getCore().getResourceLoader());
+          Object customTool = SolrCore.newInstance(new PluginInfo(name, entry.getValue()), Object.class, request.getCore(), request.getCore().getResourceLoader());
           if (customTool instanceof LocaleConfig) {
             ((LocaleConfig) customTool).configure(toolConfig);
           }
--- a/solr/core/src/java/org/apache/solr/core/DirectoryFactory.java
+++ b/solr/core/src/java/org/apache/solr/core/DirectoryFactory.java
@@ -420,7 +420,7 @@ public abstract class DirectoryFactory implements NamedListInitializedPlugin,
     final DirectoryFactory dirFactory;
     if (info != null) {
       log.debug(info.className);
-      dirFactory = config.getResourceLoader().newInstance(info.className, DirectoryFactory.class);
+      dirFactory = config.getResourceLoader().newInstance(info, DirectoryFactory.class);
       // allow DirectoryFactory instances to access the CoreContainer
       dirFactory.initCoreContainer(cc);
       dirFactory.init(info.initArgs);
--- a/solr/core/src/java/org/apache/solr/core/PluginBag.java
+++ b/solr/core/src/java/org/apache/solr/core/PluginBag.java
@@ -129,9 +129,8 @@ public class PluginBag<T> implements AutoCloseable {
   public PluginHolder<T> createPlugin(PluginInfo info) {
     if ("true".equals(String.valueOf(info.attributes.get("runtimeLib")))) {
       log.debug(" {} : '{}'  created with runtimeLib=true ", meta.getCleanTag(), info.name);
-      LazyPluginHolder<T> holder = new LazyPluginHolder<>(meta, info, core, RuntimeLib.isEnabled() ?
-          core.getMemClassLoader() :
-          core.getResourceLoader(), true);
+      //nocommit breaking runtimeLib here
+      LazyPluginHolder<T> holder = new LazyPluginHolder<T>(meta, info, core, core.getResourceLoader(), true);
 
       return meta.clazz == UpdateRequestProcessorFactory.class ?
           (PluginHolder<T>) new UpdateRequestProcessorChain.LazyUpdateProcessorFactoryHolder(holder) :
@@ -140,11 +139,10 @@ public class PluginBag<T> implements AutoCloseable {
       log.debug("{} : '{}' created with startup=lazy ", meta.getCleanTag(), info.name);
       return new LazyPluginHolder<T>(meta, info, core, core.getResourceLoader(), false);
     } else {
-      if (info.pkgName != null) {
-        PackagePluginHolder<T> holder = new PackagePluginHolder<>(info, core, meta);
-        return holder;
+      if (core.getResourceLoader().getPackage(info.pkgName) != null) {
+        return new PackagePluginHolder<>(info, core, meta);
       } else {
-        T inst = core.createInstance(info.className, (Class<T>) meta.clazz, meta.getCleanTag(), null, core.getResourceLoader(info.pkgName));
+        T inst = core.newInstance(info, (Class<T>) meta.clazz, null, core.getResourceLoader());
         initInstance(inst, info);
         return new PluginHolder<>(info, inst);
       }
@@ -403,23 +401,16 @@ public class PluginBag<T> implements AutoCloseable {
     private final SolrConfig.SolrPluginInfo pluginMeta;
     protected SolrException solrException;
     private final SolrCore core;
-    protected ResourceLoader resourceLoader;
+    protected SolrResourceLoader resourceLoader; // nocommit breaking runtimeLib here
     private final boolean isRuntimeLib;
 
-
-    LazyPluginHolder(SolrConfig.SolrPluginInfo pluginMeta, PluginInfo pluginInfo, SolrCore core, ResourceLoader loader, boolean isRuntimeLib) {
+    LazyPluginHolder(SolrConfig.SolrPluginInfo pluginMeta, PluginInfo pluginInfo, SolrCore core, SolrResourceLoader loader, boolean isRuntimeLib) {
       super(pluginInfo);
       this.pluginMeta = pluginMeta;
       this.isRuntimeLib = isRuntimeLib;
       this.core = core;
       this.resourceLoader = loader;
-      if (loader instanceof MemClassLoader) {
-        if (!RuntimeLib.isEnabled()) {
-          String s = "runtime library loading is not enabled, start Solr with -Denable.runtime.lib=true";
-          log.warn(s);
-          solrException = new SolrException(SolrException.ErrorCode.SERVER_ERROR, s);
-        }
-      }
+      // nocommit breaking runtimeLib here
     }
 
     @Override
@@ -443,16 +434,15 @@ public class PluginBag<T> implements AutoCloseable {
     private synchronized boolean createInst() {
       if (lazyInst != null) return false;
       log.info("Going to create a new {} with {} ", pluginMeta.getCleanTag(), pluginInfo.toString());
-      if (resourceLoader instanceof MemClassLoader) {
-        MemClassLoader loader = (MemClassLoader) resourceLoader;
-        loader.loadJars();
-      }
+      // nocommit breaking runtimeLib here
+
       Class<T> clazz = (Class<T>) pluginMeta.clazz;
       T localInst = null;
       try {
-        localInst = core.createInstance(pluginInfo.className, clazz, pluginMeta.getCleanTag(), null, resourceLoader);
+        localInst = core.newInstance(pluginInfo, clazz, null, resourceLoader);
       } catch (SolrException e) {
-        if (isRuntimeLib && !(resourceLoader instanceof MemClassLoader)) {
+        // nocommit breaking runtimeLib here
+        if (isRuntimeLib) {
           throw new SolrException(SolrException.ErrorCode.getErrorCode(e.code()),
               e.getMessage() + ". runtime library loading is not enabled, start Solr with -Denable.runtime.lib=true",
               e.getCause());
--- a/solr/core/src/java/org/apache/solr/core/SolrCore.java
+++ b/solr/core/src/java/org/apache/solr/core/SolrCore.java
@@ -63,7 +63,6 @@ import com.codahale.metrics.Timer;
 import com.google.common.collect.Iterators;
 import com.google.common.collect.MapMaker;
 import org.apache.commons.io.FileUtils;
-import org.apache.lucene.analysis.util.ResourceLoader;
 import org.apache.lucene.codecs.Codec;
 import org.apache.lucene.index.DirectoryReader;
 import org.apache.lucene.index.IndexDeletionPolicy;
@@ -284,13 +283,13 @@ public final class SolrCore implements SolrInfoBean, SolrMetricProducer, Closeab
   }
 
   /** Gets the SolrResourceLoader for a given package
-   * @param pkg The package name
+   * @param pkg The package name. If null or not resolved, we return {@link #getResourceLoader()}.
    */
-  public SolrResourceLoader getResourceLoader(String pkg) {
-    if (pkg == null) {
+  public SolrResourceLoader getPackageResourceLoader(String pkg) {
+    PackageLoader.Package aPackage = resourceLoader.getPackage(pkg);
+    if (aPackage == null) {
       return resourceLoader;
     }
-    PackageLoader.Package aPackage = coreContainer.getPackageLoader().getPackage(pkg);
     PackageLoader.Package.Version latest = aPackage.getLatest();
     return latest.getLoader();
   }
@@ -519,7 +518,7 @@ public final class SolrCore implements SolrInfoBean, SolrMetricProducer, Closeab
     final PluginInfo info = solrConfig.getPluginInfo(IndexDeletionPolicy.class.getName());
     final IndexDeletionPolicy delPolicy;
     if (info != null) {
-      delPolicy = createInstance(info.className, IndexDeletionPolicy.class, "Deletion Policy for SOLR", this, getResourceLoader());
+      delPolicy = newInstance(info, IndexDeletionPolicy.class, this, getResourceLoader());
       if (delPolicy instanceof NamedListInitializedPlugin) {
         ((NamedListInitializedPlugin) delPolicy).init(info.initArgs);
       }
@@ -716,12 +715,12 @@ public final class SolrCore implements SolrInfoBean, SolrMetricProducer, Closeab
     final RecoveryStrategy.Builder rsBuilder;
     if (info != null && info.className != null) {
       log.info(info.className);
-      rsBuilder = getResourceLoader().newInstance(info.className, RecoveryStrategy.Builder.class);
+      rsBuilder = getResourceLoader().newInstance(info, RecoveryStrategy.Builder.class);
     } else {
       log.debug("solr.RecoveryStrategy.Builder");
       rsBuilder = new RecoveryStrategy.Builder();
     }
-    if (info != null) {
+    if (info != null) { //TODO could this be done generically?
       rsBuilder.init(info.initArgs);
     }
     return rsBuilder;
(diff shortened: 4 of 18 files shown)


solr/core/src/java/org/apache/solr/core/SolrResourceLoader.java
```diff
@@ -529,6 +536,18 @@ public String resourceLocation(String resource) {
...
     Class<? extends T> clazz = null;
     try {
+      // If there is a package name prefix ...
+      Pair<String, String> pkgClassPair = PluginInfo.parseClassName(cname);
+      PackageLoader.Package pkg = getPackage(pkgClassPair.first());
+      if (pkg == null) {
+        // essentially, remove the package prefix and continue as normal.  Maybe it'll be found.
+        cname = pkgClassPair.second();
+      } else {
+        // TODO what version?
```
noblepaul: Trying to load the latest always? Why? Need to revisit.

solr/core/src/java/org/apache/solr/core/SolrResourceLoader.java
```diff
@@ -954,4 +987,46 @@ public static void persistConfLocally(SolrResourceLoader loader, String resource
     }
   }
 
+  // TODO document these methods...
```
noblepaul: What is the motivation behind `SolrResourceLoader` returning packages?

solr/core/src/java/org/apache/solr/core/TracerConfigurator.java
```diff
@@ -43,8 +43,7 @@ public static void loadTracer(SolrResourceLoader loader, PluginInfo info, ZkStat
         GlobalTracer.get().setSamplePercentage(0.0);
       }
     } else {
-      TracerConfigurator configurator = loader
-          .newInstance(info.className, TracerConfigurator.class);
+      TracerConfigurator configurator = loader.newInstance(info, TracerConfigurator.class);
```
noblepaul: no , thought given to reload

solr/core/src/java/org/apache/solr/core/backup/repository/BackupRepositoryFactory.java
```diff
@@ -71,7 +71,7 @@ public BackupRepository newInstance(SolrResourceLoader loader, String name) {
     PluginInfo repo = Objects.requireNonNull(backupRepoPluginByName.get(name),
         "Could not find a backup repository with name " + name);
 
-    BackupRepository result = loader.newInstance(repo.className, BackupRepository.class);
+    BackupRepository result = loader.newInstance(repo, BackupRepository.class);
```
noblepaul: no thought given to reload

solr/core/src/java/org/apache/solr/metrics/MetricSuppliers.java
```diff
@@ -283,7 +283,7 @@ public Histogram newMetric() {
 
     MetricRegistry.MetricSupplier<Counter> supplier;
     try {
-      supplier = loader.newInstance(info.className, MetricRegistry.MetricSupplier.class);
+      supplier = loader.newInstance(info, MetricRegistry.MetricSupplier.class);
```
noblepaul: How do we reload this?

solr/core/src/java/org/apache/solr/schema/IndexSchemaFactory.java
```diff
@@ -62,7 +62,7 @@ public static IndexSchema buildIndexSchema(String resourceName, SolrConfig confi
     PluginInfo info = config.getPluginInfo(IndexSchemaFactory.class.getName());
     IndexSchemaFactory factory;
     if (null != info) {
-      factory = config.getResourceLoader().newInstance(info.className, IndexSchemaFactory.class);
+      factory = config.getResourceLoader().newInstance(info, IndexSchemaFactory.class);
```
noblepaul: Do we even support the `packageName:ClassName` scheme in `schema.xml` at all? How does it play with this?
Need more discussions

solr/contrib/velocity/src/java/org/apache/solr/response/VelocityResponseWriter.java
```diff
@@ -275,7 +276,7 @@ private VelocityContext createContext(SolrQueryRequest request, SolrQueryRespons
         for (Map.Entry<String, String> entry : customTools.entrySet()) {
           String name = entry.getKey();
           // TODO: at least log a warning when one of the *fixed* tools classes is same name with a custom one, currently silently ignored
-          Object customTool = SolrCore.createInstance(entry.getValue(), Object.class, "VrW custom tool: " + name, request.getCore(), request.getCore().getResourceLoader());
+          Object customTool = SolrCore.newInstance(new PluginInfo(name, entry.getValue()), Object.class, request.getCore(), request.getCore().getResourceLoader());
```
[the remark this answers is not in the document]
dsmiley: I'll add a code review comment over on that method in SolrCore so we can discuss it in context; I renamed it.  What is the correct classloader it should use?

solr/core/src/java/org/apache/solr/core/SolrCore.java
```diff
@@ -810,28 +809,30 @@ void initIndex(boolean passOnPreviousState, boolean reload) throws IOException {
...
-      //most of the classes do not have constructors which takes SolrCore argument. It is recommended to obtain SolrCore by implementing SolrCoreAware.
-      // So invariably always it will cause a  NoSuchMethodException. So iterate though the list of available constructors
-      Constructor<?>[] cons = clazz.getConstructors();
-      for (Constructor<?> con : cons) {
-        Class<?>[] types = con.getParameterTypes();
-        if (types.length == 1 && types[0] == SolrCore.class) {
-          return cast.cast(con.newInstance(core));
+      //TODO separate out "core" scenario to another method
+      if (pluginInfo.pkgName == null && core != null) {
+        Class<? extends T> clazz = resourceLoader.findClass(pluginInfo.className, cast);
```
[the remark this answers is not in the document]
dsmiley: Quite right Noble!  In so doing, with low effort we enable all plugins to be loaded at runtime.  "Hot update" is for another day.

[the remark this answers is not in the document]
dsmiley: RE SolrCore.newInstance:  II renamed SolrCore.createInstance to SolrCore.newInstance so to be more consistent with ResourceLoader.newInstance.    Also I changed the signature of it a bit... I have some TODOs there.  I'm kinda dubious about that method on SolrCore... seems could be dropped and just use ResourceLoader.newInstance however I'm not sure how to handle plugins that take a SolrCore in its constructor.

solr/core/src/java/org/apache/solr/core/PluginBag.java
```diff
@@ -140,11 +139,10 @@ public static void initInstance(Object inst, PluginInfo info) {
       log.debug("{} : '{}' created with startup=lazy ", meta.getCleanTag(), info.name);
       return new LazyPluginHolder<T>(meta, info, core, core.getResourceLoader(), false);
     } else {
-      if (info.pkgName != null) {
-        PackagePluginHolder<T> holder = new PackagePluginHolder<>(info, core, meta);
-        return holder;
+      if (core.getResourceLoader().getPackage(info.pkgName) != null) {
```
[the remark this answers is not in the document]
dsmiley: that method SolrResourceLoader.getPackage will fail fast it can't load it for whatever reason.  See `handleUnresolvedPackage` which it calls.

solr/core/src/java/org/apache/solr/core/DirectoryFactory.java
```diff
@@ -420,7 +420,7 @@ static DirectoryFactory loadDirectoryFactory(SolrConfig config, CoreContainer cc
     final DirectoryFactory dirFactory;
     if (info != null) {
       log.debug(info.className);
-      dirFactory = config.getResourceLoader().newInstance(info.className, DirectoryFactory.class);
+      dirFactory = config.getResourceLoader().newInstance(info, DirectoryFactory.class);
```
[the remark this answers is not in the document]
dsmiley: Why?  FYI the ResourceLoader on the config is the same as that of the core, I believe.

solr/core/src/java/org/apache/solr/core/SolrCore.java
```diff
@@ -810,28 +809,30 @@ void initIndex(boolean passOnPreviousState, boolean reload) throws IOException {
...
+        //most of the classes do not have constructors which takes SolrCore argument. It is recommended to obtain SolrCore by implementing SolrCoreAware.
```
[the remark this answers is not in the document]
dsmiley: I changed this method on SolrCore (formerly createInstance, now newInstance) to take a PluginInfo.  This is important because the caller often has one, and the PluginInfo has both the package and class name separated out.  Previously, this method only took a String className but that value tended to be pluginInfo.className from the caller which thus stripped out the important information of which package it's on.

solr/core/src/java/org/apache/solr/core/SolrCore.java
```diff
@@ -890,7 +897,7 @@ private UpdateHandler createReloadedUpdateHandler(String className, String msg,
   }
 
   private UpdateHandler createUpdateHandler(String className) {
-    return createInstance(className, UpdateHandler.class, "Update Handler", this, getResourceLoader());
+    return newInstance(new PluginInfo("updateHandler", className), UpdateHandler.class, this, getResourceLoader());
```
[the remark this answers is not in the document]
dsmiley: Yes, I strongly believe hot-updates should occur in a separate ticket given that with relative ease we can get all plugins to load via the amazing plugin system :-)

solr/core/src/java/org/apache/solr/core/SolrResourceLoader.java
dsmiley: Yes, I have a "TODO" at this line.  How do I get the right version?

solr/core/src/java/org/apache/solr/core/SolrResourceLoader.java
dsmiley: I came to think of the SolrResourceLoader as being rather related to packages.  SRL has methods that lookup class names to actual Classes and then instances, so the focal point of my approach is to have `findClass` here work with the package system.

solr/core/src/java/org/apache/solr/schema/IndexSchemaFactory.java
dsmiley: > Do we even support the packageName:ClassName scheme in schema.xml at all?

It does!  Woohoo!  It's because all loading occurs via a SolrResourceLoader in Solr and so that's where the focal point of the approach in this PR is.

dsmiley:
I think one of the most important next steps for Solr is that it support the vast majority of plugin abstractions, and that is the goal of this PR.  For example the schema's abstractions are super important.  I don't know what it takes for a plugin abstraction to support hot-loading; I deliberately overlooked that.  Can they both be done easily at the same time?  If not I suggest a separating that out.

I know there is no test here yet.  I had something very WIP I didn't commit but proved very useful in trying out the schema.  I had it in `TestPackages`; does that seem suitable?  I wonder if we should have a test configSet in which _every_ plugin specification point is in its own package as a way of testing total coverage.  WDYT?

One thing *not* working that is important is lifecycle methods like `SolrCoreAware`; maybe others.

solr/core/src/java/org/apache/solr/core/DirectoryFactory.java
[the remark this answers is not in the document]
noblepaul: It's possible that we can share instances of SolrConfig between cores. We should always use SRL from core

noblepaul:
Technically, Hot-reloading means reloading an object without restarting the JVM. So, it's OK to reload an object by reloading a core. 

The problem with this PR is that it leaves the system in an inconsistent state. Consider the following scenario

-  `ComponentX` uses `PackageA` `1.0`
-  User updates `PackageA` to `2.0`
- A core that got started after the package update uses `2.0` & all other cores use `1.0`

We really cannot leave the system in an inconsistent state and say we will deal with the inconsistencies in another ticket.

What is the solution?
There are 2 ways to use packages. 

1. Ensure that the plugin has true hot reloading. As it is done for the plugins we support today
2. Or, we identify a list of plugins that do not need to be `hot-reloaded`. If a package is updated and such a plugin is using that package, we just reload the core

Wherever possible we strive to implement option#1 because reloading cores can cause instability in the system & we try to minimize it as much as possible. If option#1 is not possible we MUST implement option#2.

solr/core/src/java/org/apache/solr/core/SolrResourceLoader.java
noblepaul: SRL has nothing to do with packages. It's an anti pattern. SRL is used by PackageLaoder and SRL is totally agnostic of the package system. We should just limit exposure of Packages to 2 places
- `CoreContainer` : This is where the package loader is stored
- `SolrCore` : This is where the correct version of a package for a plugin can be identified and listeners for package updates are registered

Without the correct version, we leave the system inconsistent

solr/core/src/java/org/apache/solr/core/PluginBag.java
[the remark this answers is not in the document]
noblepaul: The idea of `SRL` trying to resolve the correct package is wrong. Let's not do this. SRL has no idea what version of a package should be used to load a plugin. Only a `SolrCore` does

solr/core/src/java/org/apache/solr/schema/IndexSchemaFactory.java
noblepaul: That's plain wrong. SRL has no idea which version of package should be loaded

janhoy: Sound like an exciting design, SRL in response of loading stuff. Could not SRL ask SolrCore about version details if it is a core-level package?

solr/core/src/java/org/apache/solr/core/DirectoryFactory.java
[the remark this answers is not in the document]
dsmiley: IndexSchema can be shared ("shareSchema" option) but SolrConfig at present cannot be.  Lets suppose it can be for the sake of argument -- wouldn't this be a reason to _not_ use SolrCore related things?  Otherwise another Collection using the same SolrConfig might theoretically refer to something it shouldn't.

solr/core/src/java/org/apache/solr/core/SolrCore.java
```diff
@@ -519,7 +518,7 @@ private IndexDeletionPolicyWrapper initDeletionPolicy(IndexDeletionPolicyWrapper
     final PluginInfo info = solrConfig.getPluginInfo(IndexDeletionPolicy.class.getName());
     final IndexDeletionPolicy delPolicy;
     if (info != null) {
-      delPolicy = createInstance(info.className, IndexDeletionPolicy.class, "Deletion Policy for SOLR", this, getResourceLoader());
+      delPolicy = newInstance(info, IndexDeletionPolicy.class, this, getResourceLoader());
```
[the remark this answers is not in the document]
dsmiley: This ultimately calls SRL with PackageInfo which resolves the right package classloader there.  That's the central idea of the approach in this PR -- SRL owns the resolution.

solr/core/src/java/org/apache/solr/core/SolrResourceLoader.java
dsmiley: The SRL is pervasively used across the codebase to load stuff (hence "resource loader" in its name).  If CoreContainer/SolrCore is now the right place, then this is a big change; no?

BTW, Wouldn't the ConfigSet object be a more befitting place to store the version association than the SolrCore?  Isn't params.json (which is where the association is) stored in the config set?

solr/core/src/java/org/apache/solr/schema/IndexSchemaFactory.java
dsmiley: "That's plain wrong. "  Please work on your tone @noblepaul .  There is no right/wrong here; we are exploring what the future holds together; it has not been written yet with no spec to judge right/wrong.  It would be more correct to say "According to my vision, the SRL should not know about packages".

noblepaul: > There is no right/wrong here; 

This is not a personal remark. What I'm saying is "it's plain wrong to load using the latest version". 

This is not how the rest of the system work. So, we have to ensure that what should be the expectation of the user

noblepaul:
@dsmiley The comments are not supposed to be personal. These are my observations on the code/PR. I'm sure somebody is going to remove that "WIP" tag pretty soon. There is a ton of work that needs to be done before we can do that. So, I'm typing down my observations as I see them.

solr/core/src/java/org/apache/solr/core/DirectoryFactory.java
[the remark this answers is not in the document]
noblepaul: yeah, schema is shared. SolrConfig is not. Maybe it's safe

noblepaul:
> RE "inconsistent state": Put differently, I think you are just pointing out that this PR is not loading the correct version? Yes; this is a known blocker/nocommit. Did I mention this was a hackday project ;-)
> 
> The latter part of your message seems to point to an as-yet unimplemented improvement proposal for the package management system (is this the right name?) to track which plugin abstractions are hot-loadable or not. I don't think that's in scope of this PR. BTW so that we don't confuse each other, I propose that "hot-loadable" imply _no_ core reload.


We should document clearly, which plugins are

-  reloaded without reloading core
- which are hot loaded
- or somethings cannot even be reloaded and it will fail

Hot reloading is not important. But clearly documenting and setting expectations is important

I clearly understand this is a hackday PR. I have no complaints about it. You seem to have misconstrued by feedback as a criticism. I just wrote down my observations. (Some observations may even be wrong as I wouldn't have thoroughly gone through the huge PR).

We definitely want more review comments on our PRs and not less. So, let's make life a bit easier for the reviewers.

solr/core/src/java/org/apache/solr/schema/IndexSchemaFactory.java
dsmiley: > am I supposed to make comments on what it does today?

Yes (agreed); and not the future.

> I never said SRL loading classes is "plain wrong

I'm glad to hear that.  You made this statement immediately after what I said, which was "It does! Woohoo! It's because all loading occurs via a SolrResourceLoader in Solr and so that's where the focal point of the approach in this PR is."  So I thought you meant my most recent statement "was plain wrong" vs some statement in some other comment thread (notice this thread doesn't discuss versions).  Can you see how I came to this interpretation?

I am working this week, albeit lightly BTW.

dsmiley:
I like the 3 categories of plugins very much, and also especially glad and surprised to hear you state "Hot reloading is not important.".  Eventually we'll want to document this in some way by-plugin abstraction so users understand in the ref guide.  And if you try to refer to a plugin when we know we can't support it, then we should tell the user more explicitly so.

I still need to look at your schema PR #1124

noblepaul:
> I like the 3 categories of plugins very much, and also especially glad and surprised to hear you state "Hot reloading is not important."

The reason why I went with hot reloading for certain components was that it was easy to implement and tests were less likely to fail. My experience with `runtimeLib` taught me that core reloads are not very reliable during tests. They fail sporadically. Moreover, we should aim to reload the smallest possible entity for sake of efficiency and stability. For instance, reloading an instance of a `SearchComponent` may happen in sub milliseconds, but a core reload can take an unpredictable amount of time depending on various other factors

The idea of SRL resolving the class name to appropriate package works, but it's hard to get it right. Hot reloading is not the problem, consistency is. Consider the following

- packages can be updated any time
- cores can be reloaded any time
- nodes can be restarted at any time

The challenge is to ensure that every single plugin in every single node in the cluster has the same version of the package being used. If we cannot have a mechanism to somehow reload the plugin, we cannot ensure consistency. Every time the package is updated, we must ensure that the plugin is refreshed. SRL is generic. It doesn't know how to reload some plugin optimally.

Today, all the implementations treat the plugin itself as the atomic unit (even then there are 3 types : Search components, Streaming expressions and other components).  #1124 treats the `IndexSchema` as the atomic unit. We may have another set of plugins that use the `SolrCore` as the atomic unit. There are so many ways we can do this. But, we CANNOT say that we will somehow load the classes today and we will worry about reloading it later. It works fine for a single node Solr , it cannot work correctly for a cluster of nodes. So, when I say something is "plain wrong", it just means that the design choices are making it impossible for us to give appropriate consistency guarantees for the entire system. 

Then we have another often overlooked factor of "diagnosability". How do we give the user visibility into what version of the plugin is used in a given node/core? There should be an appropriate end point where people can probe the system to get that information.

solr/core/src/java/org/apache/solr/core/PluginBag.java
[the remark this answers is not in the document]
dsmiley: > The idea of SRL trying to resolve the correct package is wrong.

You seem pretty adamant about this; I suggest saying you don't like that approach instead of it being "wrong".  In this PR, SRL resolving the package is a central theme and so if we don't agree on wether SRL should do this then this PR has no future (ah well).  Maybe we will return to this idea if there are holes in other approaches.  

I view the lack of correct version resolution in the PR (still only one commit as I write this) as merely a necessary TODO, and I don't forsee it'd be hard.  If you would like me to explore adding this next, let me know.

PR #1110, opened 2019-12-21
SOLR-14138: enable request log via environ var

rmuir:
Today you have to edit XML to enable the jetty request log. This makes it work via environment variable.

It also gets this thing off the currently deprecated RequestLog class.
SOLR_LOGS_DIR is now respected by default (today it uses hardcoded `logs/`)


rmuir:
ok, i fixed it to respect SOLR_LOGS_DIR as well, and moved it to the new non-deprecated way. I think its in much better shape.

PR #1114, opened 2019-12-23
LUCENE-9109: Use stack walker to implement TestSecurityManager's detection of test JVM exit

uschindler:
This is just a small improvement in Lucene/Solr master (Java 11) to detect exit of JVM in our test framework. There are other places in Lucene that use ineffective ways to inspect the stack trace.

This one optimizes the implementation of TestSecurityManager#checkExit(status) to disallow all JVM exits outside of the official test runner by using StackWalker. In addition this needs no additional permissions, because we do not instruct StackWalker to fetch all crazy stuff like Class instances of stack elements.

The way how this works is: Walk through stack trace:

- skip all internal frames (those which come before the actual exit call)
- skip all frmes with the actual exit call
- limit to one more frame (the method calling System.exit())
- check if that remaining frame is on our whitelist

This can only be commited to master (9.0), as it requires Java 9.

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -62,6 +62,9 @@ Improvements
   previously treated as beginning a line-ending comment (Satoshi Kato and Masaru Hasegawa via
   Michael Sokolov)
 
+* LUCENE-9109: Use StackWalker to implement TestSecurityManager's detection
+  of JVM exit (Uwe Schindler)
+
 Bug fixes
 
 * LUCENE-8663: NRTCachingDirectory.slowFileExists may open a file while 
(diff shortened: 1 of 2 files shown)


uschindler:
You are right:

> If there is a security manager, and this thread is not the current thread, then the security manager's checkPermission method is called with a RuntimePermission("getStackTrace") permission to see if it's ok to get the stack trace.

As we were only looking at current thread it was ok. Not sure why we had the permission stuff.

uschindler:
I just changed the static final predicate to a static method.

dweiss:
LGTM!

PR #1116, opened 2019-12-23
SOLR-14135: Utils.toJavabin returns a byte[] instead of InputStream

tflobbe:
I'm not too convinced about this PR honestly, I started thinking in doing this mostly because in the 8x branch we can't use InputStream's `readAllBytes();` method, but this may actually hurt future consumers of this method, if they don't need to read all bytes at once. I'll leave this PR for now, worst case I'll keep the tests.


noblepaul:
You can create another method with the appropriate return type, correct?

tflobbe:
Yes, the thing is that this method is not being used anywhere else at this time

PR #1119, opened 2019-12-24
SOLR-14146: Zookeeper version not resolved in RefGuide

janhoy:
See https://issues.apache.org/jira/browse/SOLR-14146


ctargett:
+1 - I didn't test it, but looked up how it broke (SOLR-8346) and your change makes sense. If you backport to `branch_8_4` also, it will be fixed in the final release of the 8.4 Ref Guide too.

PR #1120, opened 2019-12-24
LUCENE-9110: Switch tests to use StackWalker

uschindler:
There are a lot of tests (especially around IndexWriter) that look into stack traces to inject failures or check that some methods were called in their call stack.

This issue will refactor all those tests by adding a few methods to LuceneTestCase that make it easy to verify if some method call/class is in stack trace. On master (Java 11) we can use StackWalker to do this checks, which has a speedup of sometimes >>2 times (depending on how deep you dive into call stack).

There are a few tests (only 3) that do more complex stack trace analysis. Those should be refactored at some point. For now I added a deprecated method to get the whole StackTrace in Java 11, which is still 2 times faster than using an Exception.

For branch 8.x i will apply the same patch, just the LuceneTestCase methods use the old "Java 8" way to inspect stack trace using the thread's stack trace (which is very expensive).

So this issue is mainly about refactoring the tests to use a common method pattern to check the existence of stack frames.

One important thing is: Using StackWalker makes sure that the stack is "correct". Stacks from Thread or Exception may miss some frames, as it does not deoptimize the code. So depending on JVMs and optimizations (e.g. Graal), call stacks may change if we still use old code for analysis. This is no longer an issue for Java 8, but may be in future.

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -65,6 +65,9 @@ Improvements
 * LUCENE-9109: Use StackWalker to implement TestSecurityManager's detection
   of JVM exit (Uwe Schindler)
 
+* LUCENE-9110: Refactor stack analysis in tests to use generalized LuceneTestCase
+  methods that use StackWalker (Uwe Schindler)
+
 Bug fixes
 
 * LUCENE-8663: NRTCachingDirectory.slowFileExists may open a file while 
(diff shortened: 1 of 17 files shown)


rmuir:
Should we introduce some forbidden APIs here? At least for test code? I am thinking something like Thread#getStackTrace / Throwable#getStackTrace? If i do `git grep getStackTrace` it seems nearly all usages are tests doing this kind of thing.

uschindler:
@rmuir: I was thinking about the same. Maybe at a later stage. The risk is that this code breaks in future (with Graal), as the stack trace of Exceptions or Thread may be corrumpted if optimizer jumps in. To get the "real" stack trace, you should use StackWalker.

uschindler:
I removed the remaining 4 tests that required the full stack trace and rewrote them to use the new methods. I don't care of requesting a StackWalker 2 or 3 times as the code gets much cleaner. It should be faster anyways. And debugging shows that the stack trace is requested very seldom.

I think we are close to put getStackTrace() or forbiddenapis.

PR #1121, opened 2019-12-25
SOLR-11207: Add OWASP dependency checker to gradle build

janhoy:
PR against gradle-master branch to add owasp dependency checker plugin.
See https://issues.apache.org/jira/browse/SOLR-11207

To invoke, run

    ./gradlew dependencyCheckAnalyze
    open ./build/reports/dependency-check-report.html

There is also a `dependencyCheckAggregate` target that goes deeper and finds more.


dweiss:
This looks good to me, Jan. The only question is whether we want to include it now or later -- the dependencies is something still being worked on?....

janhoy:
Sure, we can include it closer to master merge.

gradle/validation/dependency-check.gradle
```diff
@@ -0,0 +1,12 @@
+// This adds OWASP vulnerability validation of project dependencies
+
+// This should be false only for debugging.
+def failOnError = true
+
+dependencyCheck {
+  autoUpdate=false
```
risdenk: Why not `autoUpdate=true`? I think this is for auto updating the definitions.

From https://jeremylong.github.io/DependencyCheck/dependency-check-gradle/configuration.html

```
Sets whether auto-updating of the NVD CVE/CPE data is enabled. It is not recommended that this be turned to false.
```

gradle/validation/dependency-check.gradle
```diff
@@ -0,0 +1,12 @@
+// This adds OWASP vulnerability validation of project dependencies
+
+// This should be false only for debugging.
+def failOnError = true
```
risdenk: It might make sense to set this to false to get this into the gradle-master build and then once merged to master change the default.

I personally find failing the build on OWASP errors can be really frustrating with new CVEs out all the time. I like to see the warnings/errors but not fail the build. This depends on the severity of the CVE as well sometimes.

dweiss: It can be a property. Then you can only warn on default builds but have it fail the build if requested ("-Pvalidation.failOn=owasp,..."). A CI job could run these so that we get a warning.

dweiss: It can be a property. Then you can only warn on default builds but have it fail the build if requested ("-Pvalidation.failOn=owasp,..."). A CI job could run these so that we get a warning.

janhoy: It is now configurable with -Pvalidation.owasp.fail=true (default not fail)

janhoy:
Updated the PR for master.

I have not enabled OWASP as part of 'check' yet, mainly due to the large initial download that it must do, and the fact that we should first exclude false positives and fix real vulnerabilities first, so that warnings actually flag something developers should look into.

So you run it with `gradlew dependencyCheckAnalyze` for a report. Supported gradle properties are: `validation.owasp.fail=true` to fail the build on a CVSS score >= 7. This is for Jenkins to use. `validation.owasp.skip=true` to skip the task. This will only make sense when we make it part of check task by default, as a way to disable owasp since it requires internet access and is heavy.

janhoy:
We could have a validation.owasp.threshold as alternative to the “fail” prop, to decide on what CVSS score to fail the build. Now it is hard coded to 7.

dweiss:
I think it should just fail if enabled. Then it'd be a single property: "validation.owasp=[true|false]?

build.gradle
```diff
@@ -58,6 +59,7 @@ apply from: file('gradle/validation/versions-props-sorted.gradle')
 apply from: file('gradle/validation/validate-source-patterns.gradle')
 apply from: file('gradle/validation/config-file-sanity.gradle')
 apply from: file('gradle/validation/rat-sources.gradle')
+apply from: file('gradle/validation/dependency-check.gradle')
```
dweiss: Perhaps change the name to "owasp-dependency-check" so that it's distinctly different from other dependency checks (there are some -- checksums, licenses, etc.)?

gradle/validation/dependency-check.gradle
```diff
@@ -0,0 +1,14 @@
+// This adds OWASP vulnerability validation of project dependencies
+// Not part of 'check' task by default, must be called explicitly, e.g. gradlew dependencyCheckAnalyze
+// Start build with -Pvalidation.owasp.fail=true to fail build on owasp errors (CVSS >= 7)
+// Start build with -Pvalidation.owasp.skip=true to skip OWASP checks during 'check' phase
+
+dependencyCheck {
```
dweiss: Does it automatically hook to all projects or is it configured for root project only?

janhoy: I don’t know. Have not tested other than root.

dweiss: You can try to run gradlew -p lucene/core ... and see what happens. :)
If it's just applicable to the root project (which it may be) then I'd wrap it in a configure(rootProject) { ... } closure so that it's explicit in the code.

janhoy: > Task 'dependencyCheckAnalyze' not found in project ':lucene:core'.

Will wrap it

janhoy:
Documentation added to workflow.txt
File renamed to owasp-dependency-check
Task included in 'check' by default, but skipped since property `-Pvalidation.owasp=[true|false]` defaults to false for now.
Caller can choose to override cvss threshold with `-Pvalidation.owasp.threshold=<int>`

dweiss:
Thanks Jan. I'll merge it in tomorrow, have a day off.

janhoy:
Thanks Dawid.

PR #1124, opened 2019-12-26
SOLR-14151 :Schema components to be loadable from packages

noblepaul:


solr/core/src/java/org/apache/solr/core/SolrCore.java
```diff
@@ -935,6 +945,7 @@ public SolrCore(CoreContainer coreContainer, String name, String dataDir, SolrCo
       MDCLoggingContext.setCore(this);
 
       resourceLoader = config.getResourceLoader();
+      resourceLoader.core = this;
```
janhoy: Is this safe? Isn’t SRL shared by many cores?

solr/core/src/java/org/apache/solr/core/PluginInfo.java
```diff
@@ -63,29 +63,42 @@ public PluginInfo(String type, Map<String, String> attrs, NamedList initArgs, Li
...
-  static Pair<String,String > parseClassName(String name) {
-    String pkgName = null;
-    String className = name;
-    if (name != null) {
-      int colonIdx = name.indexOf(':');
-      if (colonIdx > -1) {
-        pkgName = name.substring(0, colonIdx);
-        className = name.substring(colonIdx + 1);
+
+  public static class ClassName {
```
janhoy: Find a better name for this class?

solr/core/src/java/org/apache/solr/schema/IndexSchema.java
```diff
@@ -1981,4 +2089,10 @@ public PayloadDecoder getPayloadDecoder(String field) {
     return decoders.computeIfAbsent(ft, f -> PayloadUtils.getPayloadDecoder(ft));
   }
 
+  @Override
+  public void close() throws IOException {
+    if (pluginLoader instanceof PackageAwarePluginLoader) {
```
janhoy: Why the instanceof check here?

solr/core/src/java/org/apache/solr/core/SolrCore.java
noblepaul: NO, SRL is per core

solr/core/src/java/org/apache/solr/core/PluginInfo.java
noblepaul: I didn't like the name either. Suggestions welcome

solr/core/src/java/org/apache/solr/schema/IndexSchema.java
noblepaul: defensive , programming 😉

solr/core/src/java/org/apache/solr/core/ConfigSet.java
```diff
@@ -51,8 +53,11 @@ public SolrConfig getSolrConfig() {
     return solrconfig;
   }
 
+  public Supplier<IndexSchema> getIndexSchemaSupplier() {
```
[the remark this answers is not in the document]
noblepaul: Yeah, that is a problem. Can you tell me how the `Future` would look?

solr/core/src/java/org/apache/solr/core/ConfigSet.java
```diff
@@ -28,17 +30,17 @@
 
   private final SolrConfig solrconfig;
 
-  private final IndexSchema indexSchema;
+  private volatile Supplier<IndexSchema> indexSchemaSupplier;
```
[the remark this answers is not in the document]
noblepaul: It was a mistake , fixed

noblepaul:
> I see this introduces a new interface "PluginLoader" that is implemented by SolrResourceLoader and a new PackageAwarePluginLoader, and that it's only the latter that parses out the plugin information from a class name and not SolrResourceLoader (unlike the approach in my hackday). Avoiding SolrResourceLoader doing this is only feasible for Solr specific plugin loading code but not for code in Lucene that is based on a Lucene ResourceLoader. Just one example is SynonymGraphFilterFactory that can _itself_ load a Tokenizer and Analyzer via a provided ResourceLoader. Having SolrResourceLoader itself do plugin resolution will resolve this and doesn't introduce new abstractions (e.g. PluginLoader). WDYT?

I have renamed `PluginLoader` -> `SolrClassLoader` in the the latest commit

We need abstractions. Using concrete impls is one of the biggest reasons why we have very little modularity. 

So, the solution to the problem is we will make `SolrClassLoader extends  ResourceLoader` and this will be the interface which will be passed around. We will stop relying on SRL because SRL does too many things already. Piling on more things into an already bloated class is at best a hack. It's not a good idea to say that A package retsurns an SRL because most of the methods are invalid for such a use case

solr/core/src/java/org/apache/solr/core/PluginLoader.java
```diff
@@ -0,0 +1,30 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.solr.core;
+
+/**A generic interface to load classes
```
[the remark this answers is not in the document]
noblepaul: 👍

solr/core/src/java/org/apache/solr/schema/IndexSchema.java
janhoy: Ok, I see - we create this ourselves and then we have to close it :)

solr/core/src/java/org/apache/solr/core/MemClassLoader.java
```diff
@@ -53,6 +53,16 @@ public MemClassLoader(List<PluginBag.RuntimeLib> libs, SolrResourceLoader resour
     this.libs = libs;
   }
 
+  @Override
+  public <T> T newInstance(String cname, Class<T> expectedType, String... subpackages) {
+    return null;
```
dsmiley: I suggest adding "nocommit" strings to remind us there is work to do

solr/core/src/java/org/apache/solr/handler/SchemaHandler.java
```diff
@@ -206,6 +206,10 @@ private void handleGET(SolrQueryRequest req, SolrQueryResponse rsp) {
     }
   }
 
+  /**If a plugin is loaded from a package, the version of the package being used should be added
```
dsmiley: minor: nor formatted well (and same goes for many other added javadocs)
I love the new comment though

solr/core/src/java/org/apache/solr/pkg/PackageAwareSolrClassLoader.java
```diff
@@ -0,0 +1,154 @@
...
+public class PackageAwareSolrClassLoader implements SolrClassLoader {
+  final SolrCore core;
+  final SolrResourceLoader loader;
+  private Map<String, PackageAPI.PkgVersion> classNameVsPkg = new HashMap<>();
+
+  private final List<PackageListeners.Listener> listeners = new ArrayList<>();
+  private final Runnable reloadRunnable;
+
+
+  public PackageAwareSolrClassLoader(SolrCore core, SolrResourceLoader loader, Runnable runnable) {
```
dsmiley: What's the runnable for?

dsmiley:
> So, the solution to the problem is we will make `SolrClassLoader extends ResourceLoader` and this will be the interface which will be passed around. 

Sounds workable.

> We will stop relying on SRL because SRL does too many things already. Piling on more things into an already bloated class is at best a hack. It's not a good idea to say that A package returns an SRL because most of the methods are invalid for such a use case

I'm curious; what is some of the bloat of SRL that you think it should not do?

> The behavior of `SolrCoreAware` and `ResourceLoaderAware` is fixed in the latest commit

I'm glad to see that!

solr/core/src/java/org/apache/solr/core/MemClassLoader.java
noblepaul: Actually , we are getting rid of `MemClassLoader`. It's a part of the `runtimeLib` feature. So, we  don't need to implement any of these

solr/core/src/java/org/apache/solr/pkg/PackageAwareSolrClassLoader.java
noblepaul: It's to reload your set of plugins. In the case of schema it's a `refreshSchema()` , If you wish to reload core or something else , you could do it here

dsmiley:
> The problem is that SRL is directly used in all APIs

I don't consider this a problem.  Solr has a lot of plugin abstractions and resources that need loading and at least there is a single abstraction (SRL) that serves this purpose.  If there were a bunch of loading abstractions then _that_ may be a problem (IMO), but that is not the case yet thankfully.  It may be becoming the case here so I'm paying attention to complexity.

> ```
>   public String getConfigDir() 
> 
>   public String getDataDir() 
> 
>   public Properties getCoreProperties() 
> 
>   public InputStream openConfig(String name) throws IOException 
>  
>   public InputStream openSchema(String name)
> ```

I don't like those methods being there either!  Yay; we agree.  There are some more that seem dubious to me too, e.g. `persistConfLocally`; _blech!_

> So the common sense thing to do is to create an interface and hide these methods 

I think that's one option (the one you are taking).  I don't care to say what is or isn't "common sense".  A down side to this approach is that the system is now more complex as a whole; the SRL you don't like is still there.  I'm curious; do you think these unsightly methods on SRL (that we agree should go away) are really fundamental problems with SRL and get in the way of what we are doing today?

> Slapping package loading responsibilities on to SRL is the most dangerous thing to do. We have no separation of responsibilities, and we have a class we are all afraid to touch because it may break something and we nobody knows how it works. Sol already has enough such pieces

You are a man of strong opinions.  I very much disagree with you concerning your view on SRL.  SRL has been loading plugins since before a package management system came along.  If it worked with a package manager, that seems to me the same responsibility -- it still loads plugins.  The implementation would then change to also work with a package manager but it would still be responsible for loading plugins.  

Perhaps we actually agree much more than it appears.  You are proposing a new abstraction that looks _a lot_ like SRL: SolrClassLoader.  It's as if you can't bear to say you like SRL so you create something else in its image but without the dead weight we agree shouldn't be in SRL.  Can't we instead work to "fix" SRL and have one abstraction?

I very much share your concern of being wary of code that "nobody knows how it works" but I don't view SRL as such.  You can count me as someone who can help you or anyone  understand SRL should you need help ;-). Now _IndexFetcher_ there's a beast I'm scared of.

noblepaul:
>It's as if you can't bear to say you like SRL so you create something else in its image but without the dead weight we agree shouldn't be in SRL. 

SRL is fine today as a class. Did I create something in its image? NO. I just applied some lipstick on to that cute little piglet. It's just an interface. I don't even know why we are discussing SRL here. It's unproductive and we are not  solving any problems with that.

Now, we have a solution that works. This already covers 90% of what most users would need. We will have more productive debate if we start discussing

- Is there a problem with the PR? Is it bad code? Is it working wrong? Is it inefficient?
- How do we cover the last mile?


If  'n' developers are asked to solve a problem, we will get 'n' solutions. We can keep on bikeshedding why solution A is not as good as solution B or vice versa. We just solve it and move on. Solr has a lot of problems to solve. We are just worried about the color of the bike shed

noblepaul:
All tests are passing and I'm planning to merge this soon

dsmiley:
I appreciate that your perspective is that I am engaging in "bike shedding" -- debating matters of little substance.  From my perspective, I am concerned with complexity, and thus the _how_ matters.  I've realized a long time ago that writing code that works (notwithstanding unforeseen bugs) is only part of what matters and doesn't count for much.  What also matters is lots of quality "ilities" -- complexity (/ simplicity), extensibility, readability, efficiency, and various others.  All are somewhat subjective characteristics that differentiate one working solution from another. The solutions developed by the 'n' developers in your example are not equivalent just because they all work.  These quality characteristics impact the future of the project and all of us maintaining it going forward.

I fear the complexity impact in this PR.  A new abstraction that will be used widely that competes with an existing one isn't my biggest concern but it's one I thought I could articulate well (though it seems I'm doing poorly).  My biggest complexity concern is the impact of supporting hot-loading.  Hot-loading is a cool feature to tell users about but the complexity implications of the implementation scare me, and more so here at the schema.  It's not just some feature that stays on the sidelines if not used (e.g. some misc query parser); this affects our foundation.  By "scare me"; it's like what you were saying earlier about code we are scared to touch because we don't understand it (due to complexity of course).

I confess; some of my fear is mere ignorance of what I know little about.  You could help with that.  Perhaps help explain how schema hot reloading works.  And then ultimately put that in javadoc somewhere.  Frankly this patch is getting too complex for me to see for myself; I've never actually gotten through the whole thing.  I can't see the forest for the trees.

-0 to commit as is.  I can understand you want to commit any way; you seem abundantly confident in everything you've done.

noblepaul:
I would like to take a stab at what is the overall design 

## Various classes
- `interface ResourceLoader` : The lucene interface that loads classes and resources
- `interface SolrClassLoader extends ResourceLaoder` : Has some extra overloaded methods to load plugins with a few extra paramaters
- `class SolrResourceLoader implements SolrClassLoader` : implements loading classes from a set of files on disk and it also has some extra methods pertaining to Solr (data dir, conf dir etc, etc). These things do not work with packages b/c packages do not support concepts like data dir, schema config etc.
- `class PackageAwareSolrClassLoader implements SolrClassLoader`:  Has extra functionality of listening to package changes and give callback to reload something if anything changes. This outsources the classloading from file system to `SolrResourceLoader` itself. 

 ## Hotloading

There are 2 types of hot loading implemented as of today
1.  In place update of individual plugins: Most components configured in `solrconfig.xml` today support hotloading. Each object listens to its package changes and reload
2. schema : This treats the entire `IndexSchema` object as a single entity. Each object listens to package changes , if any package is modified, the entire `IndexSchema` is reloaded
3. TODO : We can support a new class of plugins which actually can reload the `SolrCore` when packages are updated. #1109 can use the `PackageAwareSolrClassLoader` to listen to and reload core

Your worst fears can be addressed with the following 
- Is there any component in IndexSchema that is referred and stored outside of IndexSchema object?
- The following method in `SolrCore` must be broken , if this is true.
```
public void setLatestSchema(IndexSchema replacementSchema)
```
 AFAIK, there are no known issues

In the worst case we can provide an extra hook for you to reload core always when schema components are changed

solr/core/src/java/org/apache/solr/core/ConfigSet.java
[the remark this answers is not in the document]
dsmiley: Nevermind Future; I was thinking FutureTask specifically but another look shows it'd block.  However Guava `Suppliers.memoize` is perfect

noblepaul:
After shared schema for SolrCloud, this PR is incompatible. I will need to build another one

chatman:
I don't understand. This PR works fine, is able to load all schema plugins. If this is not compatible with schema sharing functionality, can't we just use this for regular case and explicitly specify that this is not supported for shared schema based collections?

noblepaul:
I've opened another #1669 that's up-to-date with the current master

PR #1126, opened 2019-12-26
LUCENE-4702: Terms dictionary compression.

jpountz:
Compress blocks of suffixes in order to make the terms dictionary more
space-efficient. Two compression algorithms are used depending on which one is
more space-efficient:
 - LowercaseAsciiCompression, which applies when all bytes are in the
   `[0x1F,0x3F)` or `[0x5F,0x7F)` ranges, which notably include all digits,
   lowercase ASCII characters, '.', '-' and '_', and encodes 4 chars on 3 bytes.
   It is very often applicable on analyzed content and decompresses very quickly
   thanks to auto-vectorization support in the JVM.
 - LZ4, when the compression ratio is less than 0.75.

I was a bit unhappy with the complexity of the high-compression LZ4 option, so
I simplified it in order to only keep the logic that detects duplicate strings.
The logic about what to do in case overlapping matches are found, which was
responsible for most of the complexity while only yielding tiny benefits, has
been removed.


lucene/core/src/java/org/apache/lucene/codecs/blocktree/SegmentTermsEnumFrame.java
```diff
@@ -163,15 +179,48 @@ void loadBlock() throws IOException {
...
-    if (suffixBytes.length < numBytes) {
-      suffixBytes = new byte[ArrayUtil.oversize(numBytes, 1)];
+    if (version >= BlockTreeTermsReader.VERSION_COMPRESSED_SUFFIXES) {
+      final long codeL = ste.in.readVLong();
+      isLeafBlock = (codeL & 0x04) != 0;
+      final int numSuffixBytes = (int) (codeL >>> 3);
+      if (suffixBytes.length < numSuffixBytes) {
+        suffixBytes = new byte[ArrayUtil.oversize(numSuffixBytes, 1)];
+      }
+      compressionAlg = (int) codeL & 0x03;
```
dweiss: An enum with final code field for clarity in this and other places, maybe? Should be effectively the same at runtime.

jpountz: I pushed a change that introduces an enum, does it look better?

dweiss: Even fancier than I though it'd have to be! Looks great I think.

lucene/core/src/java/org/apache/lucene/codecs/blocktree/CompressionAlgorithm.java
```diff
@@ -0,0 +1,93 @@
...
+      org.apache.lucene.util.compress.LZ4.decompress(in, len, out, 0);
+    }
+
+    @Override
+    public String toString() {
+      return "lz4";
+    }
+  };
+
+  private static final CompressionAlgorithm[] BY_CODE = new CompressionAlgorithm[3];
```
dsmiley: Why bother with this -- explicit "code" in constructors (when the enum's intrinsic ordinal will do) and explicit BY_CODE construction when you could do `private static final CompressionAlgorithm[] BY_CODE = values();`

lucene/core/src/java/org/apache/lucene/codecs/blocktree/CompressionAlgorithm.java
```diff
@@ -0,0 +1,93 @@
...
+
+  LZ4(0x02) {
+
+    @Override
+    void read(DataInput in, byte[] out, int len) throws IOException {
+      org.apache.lucene.util.compress.LZ4.decompress(in, len, out, 0);
+    }
+
+    @Override
+    public String toString() {
```
dsmiley: What's wrong with the default enum toString?

lucene/core/src/java/org/apache/lucene/codecs/blocktree/CompressionAlgorithm.java
jpountz: I'm fine either way but I know some people strongly prefer decoupling ids - which are used for serialization and shouldn't change - from enum ordinals that might change if new constants are inserted or if values are reordered.

lucene/core/src/java/org/apache/lucene/codecs/blocktree/CompressionAlgorithm.java
jpountz: I wanted to keep the string representations lowercased in Stats#toString like in the previous iteration of this pull request, but no strong feeling, I don't mind removing these `toString` implementations.

lucene/core/src/java/org/apache/lucene/codecs/blocktree/CompressionAlgorithm.java
dsmiley: I understand that but it sandbags the current implementation with verbosity it doesn't need.  We can adjust in the future if we actually need to.

lucene/core/src/java/org/apache/lucene/codecs/blocktree/CompressionAlgorithm.java
dsmiley: I suggest then adding a default implementation that toLowerCase's name().  It would also be helpful to add a comment explaining where this is used; I didn't know.

lucene/core/src/java/org/apache/lucene/codecs/blocktree/CompressionAlgorithm.java
dweiss: Exactly. I prefer the explicit code. Gives you control and context. Ordinals are really internal details of a particular implementation of enums.

dsmiley: I don't view the public methods of an Enum as "internals".

dsmiley: BTW I can understand not wanting clients of an enum to call ordinal() on it; they should call a `getCode()` method, and the implementation of that could be ordinal().

lucene/core/src/java/org/apache/lucene/codecs/blocktree/CompressionAlgorithm.java
jpountz: I switched back to the default implementation.

lucene/core/src/java/org/apache/lucene/codecs/blocktree/CompressionAlgorithm.java
dweiss: You are entitled to your opinion, David. But the comment Adrien made still holds: when you reorder or add/ remove enum constants, their ordinal will change and this side effect is worth guarding against. The "code" may be verbose but it is also explicit and makes making a mistake more difficult. I think it's valuable here.

Anyway. My original comment about hardcoded values was only to point out that code used the same constants in different places. I suggested an enum (like Adrien implemented) but it could as well be a set of constant integers declared in one place. I don't care about this (but I do care about using or overriding ordinal()...).

jpountz: It doesn't address the concern that something that looks as innocuous as reordering constants or adding one at any position but in the end breaks serialization. I have a slight preference for being explicit like Dawid, but I've been in the middle of this controversy many times already and it's a matter of robustness vs. conciseness in the end. If you feel strongly about it @dsmiley I'll switch back to using `#ordinal()`.

jpountz: Oops I had not seen @dweiss 's comment when writing mine.

mikemccand: +1 for the explicit codes too.  Relying on enum ordinals is dangerously fragile ...

lucene/core/src/java/org/apache/lucene/codecs/blocktree/Stats.java
```diff
@@ -75,6 +75,17 @@
   /** Total number of bytes used to store term suffixes. */
   public long totalBlockSuffixBytes;
 
+  /**
+   * Number of times each compression method has been used.
+   * 0 = uncompressed
+   * 1 = lowercase_ascii
+   * 2 = LZ4
+   */
+  public final long[] compressionAlgorithms = new long[3];
```
mikemccand: Cool that you track this in BlockTree stats!  Did you post the stats somewhere?  Edit: ahh, I see the [cool stats here](https://issues.apache.org/jira/browse/LUCENE-4702?focusedCommentId=17003640&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17003640), thanks.

lucene/core/src/java/org/apache/lucene/util/compress/LZ4.java
```diff
@@ -0,0 +1,397 @@
...
+
+import org.apache.lucene.store.DataInput;
+import org.apache.lucene.store.DataOutput;
+import org.apache.lucene.util.packed.PackedInts;
+
+/**
+ * LZ4 compression and decompression routines.
+ *
+ * http://code.google.com/p/lz4/
+ * http://fastcompression.blogspot.fr/p/lz4.html
```
mikemccand: Are these also Apache 2.0 licensed?

jpountz: Not ASL2 but an ASL2-compatible license: BSD 2-clause.

dweiss: I wonder: can you put an ASL license header on files that are under a different license (even if it's permissive/ compatible)?

jpountz: I think some would argue you can here since it's a rewrite, but I agree it's probably safer to retain the BSD license instead. I changed it.

PR #1131, opened 2019-12-30
SOLR-14134: Add lazy and time-based evictiction of shared core concurrency metada…

andyvuong:
Two Main Changes:
- This change switches the cache from a ConcurrentHashMap to a TimeAwareLruCache with its underlying implementation simply being a Caffeine Cache. The default configurations, time delay before least-recently accessed (read or write) are evicted and maximum cache size are currently arbitrarily chosen and will later be configurable.
- In ZkController#register, we evict any entry that already exists in the SharedCoreConcurrencyController cache if it already exists. This supports the collection reuse case described in the JIRA. This is done here because all ADDREPLICA paths flow through here logically and we want to evict entries before the replica gets marked ACTIVE (done here)


solr/core/src/java/org/apache/solr/store/shared/TimeAwareLruCache.java
```diff
@@ -0,0 +1,137 @@
...
+      cache.put(key, value);
+    }
+    
+    @Override
+    public void cleanUp() {
+      cache.cleanUp();
+    }
+
+    @Override
+    public boolean invalidate(K key) {
```
ben-manes: This could be a one-liner :)
```java
return cache.asMap().remove(key) != null;
```

ben-manes:
Just curious if you foresee alternative implementations? If not, you might consider just using the API directly to avoid the facade and test cases that assert basic implementation details that the library's tests check as well.

mbwaheed:
> * This is done here because all ADDREPLICA paths flow through here logically and we want to evict entries before the replica gets marked ACTIVE (done here)

This path is called on each ZK reconnect too and that made me think about, with new push and pull locks instances, what would happen to concurrency if any existing indexing might be running with older locks. That further led me to look into where the right place could be. The answer seems simple: It is when the SolrCore instance is created.  And for regular eviction we do not need this tracking either, we can simply do that in SolrCore#close. The basic assumption is: if core container can hold so many SolrCore instances we can very easily hold this simple metadata. The problem of many cores is planned to be addressed by transient cores or zero replica design. With that I don't think this cache need to worry about its own size as long as it ties itself with SolrCore instances. 
Sorry I did not realize all this when I read the JIRA earlier.

andyvuong:
> And for regular eviction we do not need this tracking either, we can simply do that in SolrCore#close. The basic assumption is: if core container can hold so many SolrCore instances we can very easily hold this simple metadata. The problem of many cores is planned to be addressed by transient cores or zero replica design. With that I don't think this cache need to worry about its own size as long as it ties itself with SolrCore instances.

@mbwaheed we can evict lazily on on SolrCore instance creation in addition to SolrCore#close. To clarify you're also saying we can scope this item smaller and stick with a simpler cache (simple map, no size/time based eviction as done here) and let that future item handle whatever is needed ?

mbwaheed:
> To clarify you're also saying we can scope this item smaller and stick with a simpler cache (simple map, no size/time based eviction as done here) and let that future item handle whatever is needed ?

@andyvuong Yes. Many SolrCore instances is a bigger problem that needs to be handled either by transient cores or zero replica(w/ autoscaling). For this cache to grow and shrink with number of SolrCore instances is good enough.

andyvuong:
cc @mbwaheed - I moved the eviction on creation into registerCore, a layer above the actual ZooKeeper registration it was previously at after checking that all of the new SolrCore instances get created and go through this code path. Also switch to simple cache until the future item.

ErickErickson:
I was just skimming to see if this is related to transient cores (it doesn't appear to be) and noticed that the tests set up a new cluster in each test. That's quite a bit of work, why not set up the cluster in BeforeClass and dispose of it in AfterClass? Since each test creates its own collection, there shouldn't be any confusion.

AddReplicaTest shows one way to do this.

andyvuong:
> I was just skimming to see if this is related to transient cores (it doesn't appear to be) and noticed that the tests set up a new cluster in each test. That's quite a bit of work, why not set up the cluster in BeforeClass and dispose of it in AfterClass? Since each test creates its own collection, there shouldn't be any confusion.
> 
> AddReplicaTest shows one way to do this.

@ErickErickson A few of the tests we set up this way to test specific scenarios relying on N nodes per case but I agree its heavyweight and there's likely room for unit test improvements.

solr/core/src/java/org/apache/solr/core/CoreContainer.java
```diff
@@ -1191,9 +1192,11 @@ protected SolrCore registerCore(CoreDescriptor cd, SolrCore core, boolean regist
      * set both the name of the descriptor and the name of the
      * core, since the descriptors name is used for persisting.
      */
-
     core.setName(cd.getName());
-
+    
+    // Check and evict any pre-existing concurrency metadata for cores belonging to shared replicas
+    evictSharedCoreMetadata(cd);
```
mbwaheed: This call is not exclusive to cloud mode(zkSys.getZkController() != null). Inside evictShareCoreMetadata, getSharedStoreManager() can be null. Need zkSys.getZkController() != null protection here or inside evictShareCoreMetadata. I would put it inside the method.

solr/core/src/java/org/apache/solr/store/shared/SharedCoreConcurrencyController.java
```diff
@@ -129,18 +130,29 @@
...
+   * Maximum number of SharedCoreVersionMetadata to keep in an in-memory cache before we start 
+   * evicting entries 
+   */
+  public static int MAX_CACHE_SIZE = 5000;
+
+  /**
+   * Time in seconds that needs to pass before evicting entries from the in-memory cache that haven't 
+   * accessed recently
+   */
+  public static int CACHE_EXPIRATION_TIME_SECONDS = 60*60*12;
```
mbwaheed: This can be deleted.

PR #1136, opened 2020-01-02
LUCENE-9113: Speed up merging doc values' terms dictionaries.

jpountz:
This makes the merged view call `TermsEnum#next` on its subs rather than `#lookupOrd`.


jpountz:
Thanks @rmuir !

PR #1142, opened 2020-01-05
SOLR-14166: fq cache=false should use TwoPhaseIterator

dsmiley:
See https://issues.apache.org/jira/browse/SOLR-14166


solr/contrib/ltr/src/java/org/apache/solr/ltr/feature/SolrFeature.java
```diff
@@ -237,21 +237,25 @@ public FeatureScorer scorer(LeafReaderContext context) throws IOException {
...
+        throw new SolrException(SolrException.ErrorCode.BAD_REQUEST,
+            "PostFilter queries are not supported");
       }
-
-      return idIter;
+      Query q = pf.filter;
+      if (q == null) {
+        q = new MatchAllDocsQuery(); // usually never happens?
+      }
+      Scorer scorer = q.createWeight(searcher, ScoreMode.COMPLETE_NO_SCORES, 1f).scorer(context);
```
jpountz: Is it guaranteed that the query is rewritten at this stage?

solr/contrib/ltr/src/java/org/apache/solr/ltr/feature/SolrFeature.java
```diff
@@ -237,21 +237,25 @@ public FeatureScorer scorer(LeafReaderContext context) throws IOException {
...
+      if (scorer != null) {
+        return scorer.iterator();
```
jpountz: Assuming the query produces two-phase iterators, this will create an iterator that wraps the two-phase iterator and mostly defeats the point of using two-phase iterators. Should we return the scorer directly and let the caller do the right thing, ie. use the two-phase iterator if the query produces one, or the iterator otherwise (possibly reusing` ConjunctionDISI#intersectScorers`)?

solr/contrib/ltr/src/java/org/apache/solr/ltr/feature/SolrFeature.java
dsmiley: Ah; good catch!  No it's not.

solr/contrib/ltr/src/java/org/apache/solr/ltr/feature/SolrFeature.java
dsmiley: This particular method on this class for LTR wants to return a DocIdSetIterator.  You are correct that this method will not completely benefit from TwoPhaseIterator as-designed.  It will benefit from the cost ordering aspect though.  I like your suggestion of returning a Scorer, thus enabling the caller to _potentially_ use it better (it does not today).  But I don't want to scope creep this PR into the LTR module more than necessary to accomplish the primary goal of the PR.  If what you propose is pretty simple then it can be done now but I don't see it.

dsmiley:
Moved to the new Solr repo https://github.com/apache/solr/pull/57

PR #1144, opened 2020-01-06
SOLR-13756 updated restlet mvn repository url.

zsgyulavari:
# Description

Updated old repository URL for the restlet framework to the current official stated at:
https://restlet.talend.com/downloads/current/

# Solution

The old repository URL does a redirect for the new one, but Ivy fails to follow it on some platforms. The redirect also points to the updated URL.

# Tests

It could be compiled even after deleting the local ivy cache using `rm -rf ~/.ivy2/cache` .

# Checklist

- [n/a] I have added tests for my changes.
- [n/a] I have added documentation for the [Ref Guide](https://github.com/apache/lucene-solr/tree/master/solr/solr-ref-guide) (for Solr changes only).


zsgyulavari:
@joel-bernstein
Cloudera is hosting a mirror of the restlet repository now, so I've added the new repo url instead of the old one which is not in use since 6.6/7.0.
(dependent feat added in https://issues.apache.org/jira/browse/SOLR-1301, removed in https://issues.apache.org/jira/browse/SOLR-9221 )

uschindler:
This is not everything. You also need to change the Pom.xml.template files.

uschindler:
Please don't change the symbolic names. Only update urls.

I am not fully happy with adding a repository that contains stuff mirrored from maven Central. So the newly added Cloudera one should not contain stuff except Cloudera stuff and restlet. Please not a mirror of significant parts of Maven Central!

zsgyulavari:
Thanks for the quick review.
I will revert the symbolic name of the restlet repository, but for the cloudera one it's not just an URL update, but a whole different mirror, so it might be justified if it doesn't break something I'm unaware of.

I'll have to check back next week about whether we can set up a new mirror just for the restlets and nothing else. The original cloudera releases repo is no longer needed imho.

zsgyulavari:
@uschindler, @joel-bernstein I've checked what I can do. The decision is that we won't be maintaining a separate public mirror for restlets. Currently the cloudera repo resolves last, so it might be a good backup, but if there is a trust issue we can/should remove it altogether since the old one is not used since 6.6.

On another note I should make it work with gradle as well now I guess. Can you please decide on whether I should remove or keep the cloudera repo meanwhile?

zsgyulavari:
Rebased it on top of master to have the gradle part sorted out as well.

zsgyulavari:
rebased on top of master and went on to remove the unused cloudera repos.

PR #1145, opened 2020-01-06
LUCENE-9115: NRTCachingDirectory shouldn't cache files of unknown size.

jpountz:
See https://issues.apache.org/jira/browse/LUCENE-9115.


jpountz:
This is a good question, I don't know the answer. I had two failing seeds with OOME and while the memory usage was higher on my machine with these seeds because of NRTCachingDirectory and high multipliers, I still had some room before OOMing. I dug a bit and couldn't find other suspicious memory users.

PR #1146, opened 2020-01-06
SOLR-6613: TextField.analyzeMultiTerm does not throw an exception…

bruno-roustant:
when Analyzer returns no term.

### DIFF
--- a/solr/CHANGES.txt
+++ b/solr/CHANGES.txt
@@ -212,7 +212,7 @@ Bug Fixes
 
 * SOLR-14122: SimUtils converts v2 to v1 request params incorrectly. (Li Cao, ab)
 
-* SOLR-6613: TextField.analyzeMultiTerm does not throw an exception when Analyzer returns no term. (Bruno Roustant)
+* SOLR-6613: TextField.analyzeMultiTerm does not throw an exception when Analyzer returns no terms. (Bruno Roustant)
 
 Other Changes
 ---------------------
--- a/solr/CHANGES.txt
+++ b/solr/CHANGES.txt
@@ -212,6 +212,8 @@ Bug Fixes
 
 * SOLR-14122: SimUtils converts v2 to v1 request params incorrectly. (Li Cao, ab)
 
+* SOLR-6613: TextField.analyzeMultiTerm does not throw an exception when Analyzer returns no term. (Bruno Roustant)
+
 Other Changes
 ---------------------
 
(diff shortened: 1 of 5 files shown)


solr/CHANGES.txt
```diff
@@ -212,6 +212,8 @@ Bug Fixes
 
 * SOLR-14122: SimUtils converts v2 to v1 request params incorrectly. (Li Cao, ab)
 
+* SOLR-6613: TextField.analyzeMultiTerm does not throw an exception when Analyzer returns no term. (Bruno Roustant)
```
dsmiley: "no term" -> "no terms"

solr/core/src/java/org/apache/solr/schema/TextField.java
```diff
@@ -173,8 +173,10 @@ public static BytesRef analyzeMultiTerm(String field, String part, Analyzer anal
 
       TermToBytesRefAttribute termAtt = source.getAttribute(TermToBytesRefAttribute.class);
 
-      if (!source.incrementToken())
-        throw  new SolrException(SolrException.ErrorCode.BAD_REQUEST,"analyzer returned no terms for multiTerm term: " + part);
+      if (!source.incrementToken()) {
+        // Accept no token because it may have been filtered out by a StopFilter for example.
+        return null;
+      }
```
dsmiley: Optional: Consider adding javadocs for TextField.analyzeMultiTerm().  Yes, someone else should have long ago

solr/core/src/test/org/apache/solr/schema/TestTextField.java
```diff
@@ -0,0 +1,54 @@
...
+    assertNull(termBytes);
+
+    // One term provided by the WhitespaceTokenizer for the multi-term part.
+    // This is the regular case. Check TextField.analyzeMultiTerm returns it (and does not throw an exception).
+    termBytes = TextField.analyzeMultiTerm("field", "Sol", new WhitespaceAnalyzer());
+    assertEquals("Sol", termBytes.utf8ToString());
+
+    // Two terms provided by the WhitespaceTokenizer for the multi-term part.
+    // This is not allowed. Expect an exception.
+    try {
```
dsmiley: See expectThrows() in-scope, instead.

solr/CHANGES.txt
bruno-roustant: Hum right, this strange rule of negation and plural. There is the same notion in French and I don't like it neither. To me it means "not a single term" or "zero term" so why should we put plural.
But ok I'll change that.

solr/core/src/test/org/apache/solr/schema/TestTextField.java
bruno-roustant: Thanks

PR #1151, opened 2020-01-07
SOLR-13890: Add "top-level" DVTQ implementation

gerlowskija:
# Description

The "docValuesTermsFilter" method for TermsQParserPlugin uses per-segment docvalue structures to find matches.  This works well when the number of query terms is small, but as it gets over a few hundred, this quickly becomes less efficient than using a "top-level" or global docvalues structure.

# Solution

This PR introduces a query that uses the top-level structure and takes advantage of the Two Phase Iterator (TPI) interface to handle queries with large numbers of terms much more efficiently than any of the existing methods.

# Tests

This PR adds `TestTermsQParserPlugin` with some tests to cover the new code added here as well as the rest of the terms qparser (which was largely untested).  I also wrote a JUnit performance driver to exhibit perf gains, but that no longer appears in this PR.  See patches on the JIRA issue for that.

# Checklist


lucene/core/src/java/org/apache/lucene/search/TwoPhaseIterator.java
```diff
@@ -64,7 +64,7 @@ public static TwoPhaseIterator unwrap(DocIdSetIterator iterator) {
 
     TwoPhaseIteratorAsDocIdSetIterator(TwoPhaseIterator twoPhaseIterator) {
       this.twoPhaseIterator = twoPhaseIterator;
-      this.approximation = twoPhaseIterator.approximation;
+      this.approximation = twoPhaseIterator.approximation();
```
dsmiley: Is this just a minor improvement or is it necessary?  If just some minor improvement, I recommend you don't touch Lucene in a Solr PR.

solr/CHANGES.txt
```diff
@@ -187,6 +187,11 @@ Improvements
   hl.fragsizeIsMinimum, with defaults that aim to better center matches in fragments than previously. See the ref guide.
   Regardless of the settings, the passages may be sized differently than before. (Nándor Mátravölgyi, David Smiley)
 
+* SOLR-13890: Add "top-level" DV implementation for {!terms} queries.  This approach tends to be more efficient for
+  queries withvery large numbers of terms.  The new implementation is used by default for method=docValuesTermsFilter
```
dsmiley: "withvery" -- add space.

solr/core/src/java/org/apache/solr/search/TermsQParserPlugin.java
```diff
@@ -79,20 +88,34 @@ Query makeFilter(String fname, BytesRef[] byteRefs) {
...
         Automaton union = Automata.makeStringUnion(Arrays.asList(byteRefs)); // input must be sorted
         return new AutomatonQuery(new Term(fname), union);//constant scores
       }
     },
     docValuesTermsFilter {//on 4x this is FieldCacheTermsFilter but we use the 5x name any way
       @Override
-      Query makeFilter(String fname, BytesRef[] byteRefs) {
-        return new DocValuesTermsQuery(fname, byteRefs);//constant scores
+      Query makeFilter(String fname, BytesRef[] byteRefs, SolrParams localParams) {
+        final String type = localParams.get(SUBMETHOD);
```
dsmiley: Instead of yet another local-param, maybe "method" could be docValuesTermsFilterTopLevel and docValuesTermsFilterTopLevel.  The existing docValuesTermsFilter could dispatch to either of those.

solr/core/src/java/org/apache/solr/search/TermsQParserPlugin.java
```diff
@@ -102,6 +125,9 @@ public QParser createParser(String qstr, SolrParams localParams, SolrParams para
       public Query parse() throws SyntaxError {
         String fname = localParams.get(QueryParsing.F);
         FieldType ft = req.getSchema().getFieldTypeNoEx(fname);
+        if (ft == null) {
+          throw new SolrException(SolrException.ErrorCode.BAD_REQUEST, "Field name [" + fname + "] does not exist");
```
dsmiley: Instead the line above "NoEx" could be removed and then we get an exception

solr/core/src/java/org/apache/solr/search/TermsQParserPlugin.java
```diff
@@ -138,8 +164,128 @@ public Query parse() throws SyntaxError {
           bytesRefs[i] = term.toBytesRef();
         }
 
-        return method.makeFilter(fname, bytesRefs);
+        return method.makeFilter(fname, bytesRefs, localParams);
       }
     };
   }
+
+  private static abstract class TopLevelTwoPhaseIterator extends TwoPhaseIterator {
```
dsmiley: I don't think you need this class; it seems more confusing than it is useful

solr/core/src/java/org/apache/solr/search/TermsQParserPlugin.java
```diff
@@ -138,8 +164,128 @@ public Query parse() throws SyntaxError {
...
+    private final int nextDocBase;
+    public TopLevelTwoPhaseIterator(DocIdSetIterator approximation, int docBase, int nextDocBase) {
+      super(new PerSegmentViewDocIdSetIterator(approximation, docBase, nextDocBase));
+
+      this.docBase = docBase;
+      this.nextDocBase = nextDocBase;
+    }
+  }
+
+  private static class TopLevelDocValuesTermsQuery extends DocValuesTermsQuery implements ExtendedQuery{
```
dsmiley: I wouldn't bother implementing ExtendedQuery.  If we want to tweak defaults, we should probably do so to affect all docValuesTermsQuery (both sub-methods) and that can be done by replacing QParser.localParams wrapped with one setting defaults we choose in our QParser.

solr/core/src/java/org/apache/solr/search/PerSegmentViewDocIdSetIterator.java
```diff
@@ -0,0 +1,80 @@
...
+
+import org.apache.lucene.search.DocIdSetIterator;
+
+/**
+ * A {@link org.apache.lucene.search.DocIdSetIterator} which allows single-segment access to a wrapped top-level (multi-segment) DISI.
+ *
+ * Useful as a shim between a top-level data structure, and logic that attempts to operate on it using segment-level docIds.
+ * The same wrapped DISI can be used serially by multiple PerSegmentViewDocIdSetIterators
+ */
+public class PerSegmentViewDocIdSetIterator extends DocIdSetIterator {
```
dsmiley: Can you make this an inner class of topLevelDocValuesTermsQuery or anonymous?  If there is a future need to share it _then_ I think we should pull it out.  I doubt it's as generally useful to be a standalone public class.

After further review, I don't think we need the functionality here _at all_... see my future feedback

solr/core/src/java/org/apache/solr/search/TermsQParserPlugin.java
```diff
@@ -138,8 +164,128 @@ public Query parse() throws SyntaxError {
...
+        long ord = lookupTerm(values, term, lastOrdFound);
+        if (ord >= 0L) {
+          matchesAtLeastOneTerm = true;
+          queryTermOrdinals.set(ord);
+          lastOrdFound = ord;
+        }
+      }
+
+      return new ConstantScoreWeight(this, boost) {
+        public Scorer scorer(LeafReaderContext context) throws IOException {
```
dsmiley: I see that you are using the top level sortedSetDocValues as a DocIdSetIterator approximation.  I can see this works but I think we needn't get that complex, and we miss out on other optimizations by doing so.  I'll post PR comment on how to replace it in a way that also removes TopLevelTwoPhaseIterator.

solr/core/src/java/org/apache/solr/search/TermsQParserPlugin.java
```diff
@@ -138,8 +164,128 @@ public Query parse() throws SyntaxError {
...
+    private int cost;
+
+
+    public TopLevelDocValuesTermsQuery(String field, BytesRef... terms) {
+      super(field, terms);
+      this.fieldName = field;
+    }
+
+    public Weight createWeight(IndexSearcher searcher, final ScoreMode scoreMode, float boost) throws IOException {
+      values = DocValues.getSortedSet(((SolrIndexSearcher)searcher).getSlowAtomicReader(), fieldName);
```
dsmiley: I suggest renaming to "topDocValues".   The "top"-ness vs "segment" needs to be made more clear in all variables names here because we dance between them.

mkhludnev: Another naming proposal is ```global_ordinals```.

gerlowskija:
David's Scorer implementation is waay simpler than mine.  I ran a performance test last night and it performs just as well.  (see below).  Not sure what I was thinking with the overly complex PerSegmentView stuff.  Thanks David!  Hoping to get address review comments addressed and pushed up shortly.

perf-with-davids-improvement

solr/core/src/java/org/apache/solr/search/TermsQParserPlugin.java
gerlowskija: I went with `topLevelDocValues`, as `topDocValues` might sounds like a PQueue or a list to those unfamiliar with the code.

solr/core/src/java/org/apache/solr/search/TermsQParserPlugin.java
gerlowskija: Will that work?  My understanding was that SolrIndexSearcher uses cost/cache iff it notices its working with an ExtendedQuery that has `getCost`/`getCache` methods it can use. (i.e. it uses those specific methods, it doesn't root around in SolrParams objects looking for cost/cache param values.)

dsmiley: It will because QParser.getQuery will wrap the query produced by parse() if it's not already an ExtendedQuery.

solr/solr-ref-guide/src/other-parsers.adoc
```diff
@@ -1031,13 +1031,17 @@ The field on which to search. This parameter is required.
...
++
+`docValuesTermsFilterTopLevel` uses top-level doc values data structures to find results.  These data structures are more efficient as the number of query terms grows high (over several hundred). But they are also expensive to build and need to be populated lazily after each commit, causing a sometimes-noticeable slowdown on the first query after each commit.  If you commit frequently, you may benefit from adding a static warming query to your `solrconfig.xml` so that this is done as a part of the commit itself and not attached directly to user requests.
```
dsmiley: The advice about commit frequently and thus add a static warming query is very debatable.  "commit frequently" ~= NRT and NRT users want to dial down caching, even eliminate.  But but maybe they can tolerate longer latency for freshness provided the query response time isn't incurring costs post-commit.  There's no universal answer.  Yet you did say "you *may* benefit" so I guess what you wrote is fine.

solr/solr-ref-guide/src/other-parsers.adoc
```diff
@@ -1031,13 +1031,17 @@ The field on which to search. This parameter is required.
...
+`docValuesTermsFilter` chooses between the `docValuesTermsFilterTopLevel` and `docValuesTermsFilterPerSegment` methods (see below) using the number of query terms as a rough heuristic.  Users should typically use this method instead of using `docValuesTermsFilterTopLevel` or `docValuesTermsFilterPerSegment` directly, unless they've done performance testing to validate one or the other methods on queries of all sizes. Depending on the implementation picked, this method may rely on expensive data structures which are lazily populated after each commit.  If you commit frequently, you may benefit from adding a static warming query to `solrconfig.xml` so that this is done as a part of the commit itself and not attached directly to user requests.
```
dsmiley: Maybe right up front here first declare that this method is only appropriate when the field has docValues?  That's more important and differentiates from terms index based methods.

solr/solr-ref-guide/src/other-parsers.adoc
```diff
@@ -1031,13 +1031,17 @@ The field on which to search. This parameter is required.
...
++
+`docValuesTermsFilterTopLevel` uses top-level doc values data structures to find results.  These data structures are more efficient as the number of query terms grows high (over several hundred). But they are also expensive to build and need to be populated lazily after each commit, causing a sometimes-noticeable slowdown on the first query after each commit.  If you commit frequently, you may benefit from adding a static warming query to your `solrconfig.xml` so that this is done as a part of the commit itself and not attached directly to user requests.
++
+`docValuesTermsFilterPerSegment` uses segment-level doc values data structures to find results.  This is more general purpose than the corresponding "top level" method, but doesn't have the same advantages at large numbers of query terms (see above).
```
dsmiley: "doesn't have the same advantages" without also listing some positives here makes me think this method is not useful when really it's excellent.  Lots of users would say 300 queries is a "large number", and this method is appropriate.  Also this method is very NRT friendly.

solr/core/src/java/org/apache/solr/search/TermsQParserPlugin.java
```diff
@@ -142,4 +158,112 @@ public Query parse() throws SyntaxError {
...
+    private int cost;
+
+
+    public TopLevelDocValuesTermsQuery(String field, BytesRef... terms) {
+      super(field, terms);
+      this.fieldName = field;
+    }
+
+    public Weight createWeight(IndexSearcher searcher, final ScoreMode scoreMode, float boost) throws IOException {
+      topLevelDocValues = DocValues.getSortedSet(((SolrIndexSearcher)searcher).getSlowAtomicReader(), fieldName);
```
dsmiley: If not a SolrIndexSearcher than return super.createWeight    (your idea based on Mikhail's comment)

solr/solr-ref-guide/src/other-parsers.adoc
gerlowskija: Hmm, I see your point.  I don't want to take the advice out entirely because the first-query QTime is pretty startling and I don't want that to scare people away.  Maybe I'll leave the warning in but reword it to be a bit more conditional.

gerlowskija:
@dsmiley ready for re-review when you get another chance

solr/core/src/java/org/apache/solr/search/TermsQParserPlugin.java
```diff
@@ -142,4 +158,100 @@ public Query parse() throws SyntaxError {
...
+                if (topLevelTermOrdinals.get(ord)) {
+                  return true;
+                }
+              }
+
+              return false;
+            }
+
+            public float matchCost() {
+              return 3.0F;
```
dsmiley: Compared to other TPIs, I think 3f is too low.  For example DVTQ (in the superclass) is 3f but we also must call advanceExact on the top level DV.  I suggest 10f.  It's a bike-shed number but I just want more than 3 and much less than 100 which is what some of the no-idea but probably expensive TPIs have.

solr/core/src/java/org/apache/solr/search/TermsQParserPlugin.java
```diff
@@ -88,7 +91,20 @@ Query makeFilter(String fname, BytesRef[] byteRefs) {
...
       Query makeFilter(String fname, BytesRef[] byteRefs) {
-        return new DocValuesTermsQuery(fname, byteRefs);//constant scores
+        // TODO Further tune this heuristic number
+        return (byteRefs.length > 700) ? new TopLevelDocValuesTermsQuery(fname, byteRefs) : new DocValuesTermsQuery(fname, byteRefs);
+      }
+    },
+    docValuesTermsFilterTopLevel {
+      @Override
+      Query makeFilter(String fname, BytesRef[] byteRefs) {
+        return new TopLevelDocValuesTermsQuery(fname, byteRefs);
```
dsmiley: I believe Joel argued strongly we should default both DV impls to cache=false?  You can do that here (and in the other method) easily via WrappedQuery wrapper and setting cache=false.  I mentioned another approach that messes with this.localParams but perhaps that might feel hacky compared to WrappedQuery.  Especially so since makeFilter doesn't have access to change the field this.localParams.

gerlowskija: I'd forgotten about this, done.

FWIW I buy Joel's argument that most real-world use-cases are going to require `cache=false`.  But I worry that users who want caching will be bitten by this.  I guess it depends how hands-on you expect users to be with their query tuning.  For hands-off users, this change is best. But for hands-on users, we've just created a "gotcha".

I'll make sure to add this to the ref-guide docs at the least.

gerlowskija:
Ready for review again when you get a chance.

solr/core/src/java/org/apache/solr/search/TermsQParserPlugin.java
dsmiley: This is just about a default.  So-called "Hands-on users" (experts) will be able to be explicit with cache=true.

gerlowskija: Sure, setting`cache=true` is trivial.  It's _knowing_ that you need to be explicit for this one query in particular that's the tricky part.

But sure, I'm just hand-wringing; I've already made the change.  It's documented in the ref-guide, so that's as good as we're going to get.

solr/core/src/java/org/apache/solr/search/TermsQParserPlugin.java
```diff
@@ -92,22 +92,28 @@ Query makeFilter(String fname, BytesRef[] byteRefs) {
       @Override
       Query makeFilter(String fname, BytesRef[] byteRefs) {
         // TODO Further tune this heuristic number
-        return (byteRefs.length > 700) ? new TopLevelDocValuesTermsQuery(fname, byteRefs) : new DocValuesTermsQuery(fname, byteRefs);
+        return disableCacheByDefault((byteRefs.length > 700) ? new TopLevelDocValuesTermsQuery(fname, byteRefs) : new DocValuesTermsQuery(fname, byteRefs));
```
dsmiley: right here simply call docValuesTermsFilterTopLevel.makeFilter or docValuesTermsFilterPerSegment.makeFilter.  Yes, enum methods may refer to other enums in the same enum :-)

solr/solr-ref-guide/src/other-parsers.adoc
```diff
@@ -1037,11 +1037,11 @@ An optional parameter used to determine which of several query implementations s
...
+`docValuesTermsFilter` can only be used on fields with docValues data.  The `cache` parameter is false by default.  Chooses between the `docValuesTermsFilterTopLevel` and `docValuesTermsFilterPerSegment` methods using the number of query terms as a rough heuristic.  Users should typically use this method instead of using `docValuesTermsFilterTopLevel` or `docValuesTermsFilterPerSegment` directly, unless they've done performance testing to validate one of the methods on queries of all sizes.  Depending on the implementation picked, this method may rely on expensive data structures which are lazily populated after each commit.  If you commit frequently and your use-case can tolerate a static warming query, consider adding one to `solrconfig.xml` so that this work is done as a part of the commit itself and not attached directly to user requests.
```
dsmiley: Just a side-comment about our ref docs:  In the ref docs I write or update, I convert them to one sentence per line.  This makes the diffs easy to read!  The pain of no newlines is very apparent here.  Change or not as you wish.  CC @ctargett

gerlowskija: I thought about this too when I saw the diff.  I remember it coming up as a discussion point when the ref-guide was initially being hashed out too.

I won't change my docs here, as it's probably worth keeping the file as a whole consistent in the strategy it uses.  But maybe it's worth revisiting at some point.

solr/core/src/java/org/apache/solr/search/TermsQParserPlugin.java
gerlowskija: Hah.  I don't know whether to think that's awesome or revolting.  Done.

dsmiley: LOL there is something slightly unsettling about it I admit.  Though it's not confusing or anything.  Any way I value brevity over a lot and I'll take it!

solr/solr-ref-guide/src/other-parsers.adoc
ctargett: When I think of it, I've been changing to one sentence per line when I'm editing also (but I often forget). I don't know that it really changes that much to have a whole page consistent - we should probably encourage one sentence per line, but if it's a big file and you can only do a little, I think that's fine and maybe someone else will get inspired later to do the rest.

PR #1152, opened 2020-01-07
SOLR-14172: Collection metadata remains in zookeeper if too many shards requested

asalamon74:
# Description

When a Collection CREATE command fails because of too many requested shards error the collection metadata remains in ZooKeeper

# Solution

Delete the metadata from zookeeper if this error happens.

# Tests

Created a new unit tests which fails against master but works after this modification.

# Checklist

### DIFF
--- a/solr/core/src/java/org/apache/solr/cloud/api/collections/CreateCollectionCmd.java
+++ b/solr/core/src/java/org/apache/solr/cloud/api/collections/CreateCollectionCmd.java
@@ -190,10 +190,12 @@ public class CreateCollectionCmd implements OverseerCollectionMessageHandler.Cmd
       try {
         replicaPositions = buildReplicaPositions(ocmh.cloudManager, clusterState, clusterState.getCollection(collectionName), message, shardNames, sessionWrapper);
       } catch (Assign.AssignmentException e) {
-        ZkNodeProps deleteMessage = new ZkNodeProps("name", collectionName);
-        new DeleteCollectionCmd(ocmh).call(clusterState, deleteMessage, results);
+        deleteCollection(clusterState, results, collectionName);
         // unwrap the exception
         throw new SolrException(ErrorCode.SERVER_ERROR, e.getMessage(), e.getCause());
+      } catch (SolrException e) {
+        deleteCollection(clusterState, results, collectionName);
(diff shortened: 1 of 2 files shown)


solr/core/src/java/org/apache/solr/cloud/api/collections/CreateCollectionCmd.java
```diff
@@ -190,10 +190,12 @@ public void call(ClusterState clusterState, ZkNodeProps message, NamedList resul
       try {
         replicaPositions = buildReplicaPositions(ocmh.cloudManager, clusterState, clusterState.getCollection(collectionName), message, shardNames, sessionWrapper);
       } catch (Assign.AssignmentException e) {
-        ZkNodeProps deleteMessage = new ZkNodeProps("name", collectionName);
-        new DeleteCollectionCmd(ocmh).call(clusterState, deleteMessage, results);
+        deleteCollection(clusterState, results, collectionName);
         // unwrap the exception
         throw new SolrException(ErrorCode.SERVER_ERROR, e.getMessage(), e.getCause());
+      } catch (SolrException e) {
```
risdenk: So one question I have is why is the error coming back from `buildReplicaPositions` not an `Assign.AssignmentException`? Is it because it is wrapped in a `SolrException` from the remote node?

asalamon74: `buildReplicaPositions` checks the max shard number [itself](https://github.com/apache/lucene-solr/blob/releases/lucene-solr/8.4.1/solr/core/src/java/org/apache/solr/cloud/api/collections/CreateCollectionCmd.java#L389-L390) and throws a new `SQLException`. It also calls `Assign.*` methods, those might throw the `Assign.AssignmentException` exceptions.

shalinmangar:
Thanks @asalamon74 for the PR. This bug is my fault. I think it is best to change the SolrException being thrown inside `buildReplicaPositions` to an `AssignmentException`. After all it is the intended use of that exception. Also, the SolrException thrown in the catch clause inside CreateCollectionCmd.call should be a bad request instead of a server error as it is today. I'll put up a patch on Jira with these changes and commit.

PR #1155, opened 2020-01-07
LUCENE-8962: Add ability to selectively merge on commit

msfroh:
# Description

If we have many index writer threads, they will flush their DWPT buffers
on commit, resulting in many small segments, which can be merged before
the commit returns.

# Solution

This adds a new "findCommitMerges" method to MergePolicy, which can specify merges to be executed as part of preparing a commit. By default, we return null for backward compatibility (i.e. don't merge on commit).

In IndexWriter, we call findCommitMerges from prepareCommitInternal, after we have cloned the SegmentInfos. If we found commit merges, we wrap each OneMerge so that on completion, they update the cloned SegmentInfos and reference counts of segment files. Outside the flush lock, we wait (up to max time specified in IndexWriterConfig) for the commit merges to complete before calling startCommit. 

Also, added an IndexWriterEvents callback (configurable through IndexWriterConfig) so that consuming code can be notified of merge on commit events (to emit metrics on time spent waiting for the merges to complete, for example).

# Tests

1. Added findCommitMerges implementation to MockRandomMergePolicy.
2. Added explicit test case for merge-on-commit behavior in TestIndexWriterMergePolicy.

# Checklist

- [✓] I have reviewed the guidelines for [How to Contribute](https://wiki.apache.org/solr/HowToContribute) and my code conforms to the standards described there to the best of my ability.
- [✓] I have created a Jira issue and added the issue ID to my pull request title.
- [✓] I have given Solr maintainers [access](https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork) to contribute to my PR branch. (optional but recommended)
- [✓] I have developed this patch against the `master` branch.
- [✓] I have run `ant precommit` and the appropriate test suite.
- [✓] I have added tests for my changes.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
+++ b/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
@@ -3149,10 +3149,9 @@ public class IndexWriter implements Closeable, TwoPhaseCommit, Accountable,
     }
   }
 
-  private MergePolicy waitForMergeOnCommitPolicy(MergePolicy source, final SegmentInfos toCommit,
-                                                 AtomicReference<CountDownLatch> mergeLatchRef) {
-    return new OneMergeWrappingMergePolicy(source, (toWrap) -> new MergePolicy.OneMerge(toWrap.segments) {
-      @Override
+  private MergePolicy.OneMerge updateSegmentInfosOnMergeFinish(MergePolicy.OneMerge merge, final SegmentInfos toCommit,
+                                                                AtomicReference<CountDownLatch> mergeLatchRef) {
+    return new MergePolicy.OneMerge(merge.segments) {
       public void mergeFinished() throws IOException {
         super.mergeFinished();
         CountDownLatch mergeAwaitLatch = mergeLatchRef.get();
@@ -3162,23 +3161,28 @@ public class IndexWriter implements Closeable, TwoPhaseCommit, Accountable,
         }
         if (isAborted() == false) {
           deleter.incRef(this.info.files());
-          toCommit.add(this.info.clone());
-          long segmentCounter = Long.parseLong(this.info.info.name.substring(1), Character.MAX_RADIX);
-          toCommit.counter = Math.max(toCommit.counter, segmentCounter + 1);
-          Set<String> segmentNamesToRemove = new HashSet<>();
+          // Resolve "live" SegmentInfos segments to their toCommit cloned equivalents, based on segment name.
+          Set<String> mergedSegmentNames = new HashSet<>();
           for (SegmentCommitInfo sci : this.segments) {
             deleter.decRef(sci.files());
-            segmentNamesToRemove.add(sci.info.name);
+            mergedSegmentNames.add(sci.info.name);
           }
-          for (int i = toCommit.size() - 1; i >= 0; i--) {
-            if (segmentNamesToRemove.contains(toCommit.info(i).info.name)) {
-              toCommit.remove(i);
+          List<SegmentCommitInfo> toCommitMergedAwaySegments = new ArrayList<>();
+          for (SegmentCommitInfo sci : toCommit) {
+            if (mergedSegmentNames.contains(sci.info.name)) {
+              toCommitMergedAwaySegments.add(sci);
             }
           }
+          // Construct a OneMerge that applies to toCommit
+          MergePolicy.OneMerge applicableMerge = new MergePolicy.OneMerge(toCommitMergedAwaySegments);
+          applicableMerge.info = this.info.clone();
+          long segmentCounter = Long.parseLong(this.info.info.name.substring(1), Character.MAX_RADIX);
+          toCommit.counter = Math.max(toCommit.counter, segmentCounter + 1);
+          toCommit.applyMergeChanges(applicableMerge, false);
         }
         mergeAwaitLatch.countDown();
       }
-    });
+    };
   }
 
   private long prepareCommitInternal() throws IOException {
@@ -3203,7 +3207,7 @@ public class IndexWriter implements Closeable, TwoPhaseCommit, Accountable,
       SegmentInfos toCommit = null;
       boolean anyChanges = false;
       long seqNo;
-      MergePolicy.MergeSpecification commitMerges = null;
+      List<MergePolicy.OneMerge> commitMerges = null;
       AtomicReference<CountDownLatch> mergeAwaitLatchRef = null;
 
       // This is copied from doFlush, except it's modified to
@@ -3260,31 +3264,25 @@ public class IndexWriter implements Closeable, TwoPhaseCommit, Accountable,
               toCommit = segmentInfos.clone();
 
               if (anyChanges) {
-                mergeAwaitLatchRef = new AtomicReference<>();
-                MergePolicy mergeOnCommitPolicy = waitForMergeOnCommitPolicy(config.getMergePolicy(), toCommit, mergeAwaitLatchRef);
-
                 // Find any merges that can execute on commit (per MergePolicy).
-                commitMerges = mergeOnCommitPolicy.findCommitMerges(segmentInfos, this);
-                if (commitMerges != null && commitMerges.merges.size() > 0) {
-                  int mergeCount = 0;
-                  for (MergePolicy.OneMerge oneMerge : commitMerges.merges) {
-                    if (registerMerge(oneMerge)) {
-                      mergeCount++;
-                    } else {
+                MergePolicy.MergeSpecification mergeSpec =
+                    config.getMergePolicy().findCommitMerges(segmentInfos, this);
+                if (mergeSpec != null && mergeSpec.merges.size() > 0) {
+                  int mergeCount = mergeSpec.merges.size();
+                  commitMerges = new ArrayList<>(mergeCount);
+                  mergeAwaitLatchRef = new AtomicReference<>(new CountDownLatch(mergeCount));
+                  for (MergePolicy.OneMerge oneMerge : mergeSpec.merges) {
+                    MergePolicy.OneMerge trackedMerge =
+                        updateSegmentInfosOnMergeFinish(oneMerge, toCommit, mergeAwaitLatchRef);
+                    if (registerMerge(trackedMerge) == false) {
                       throw new IllegalStateException("MergePolicy " + config.getMergePolicy().getClass() +
                           " returned merging segments from findCommitMerges");
                     }
+                    commitMerges.add(trackedMerge);
                   }
-                  if (mergeCount > 0) {
-                    if (infoStream.isEnabled("IW")) {
-                      infoStream.message("IW", "Registered " + mergeCount + " commit merges");
-                      infoStream.message("IW", "Before executing commit merges, had " + toCommit.size() + " segments");
-                    }
-                    mergeAwaitLatchRef.set(new CountDownLatch(mergeCount));
-                  } else {
-                    if (infoStream.isEnabled("IW")) {
-                      infoStream.message("IW", "Registered no commit merges");
-                    }
+                  if (infoStream.isEnabled("IW")) {
+                    infoStream.message("IW", "Registered " + mergeCount + " commit merges");
+                    infoStream.message("IW", "Before executing commit merges, had " + toCommit.size() + " segments");
                   }
                 }
               }
@@ -3318,18 +3316,18 @@ public class IndexWriter implements Closeable, TwoPhaseCommit, Accountable,
         maybeCloseOnTragicEvent();
       }
 
-      if (mergeAwaitLatchRef != null && mergeAwaitLatchRef.get() != null) {
+      if (mergeAwaitLatchRef != null) {
         CountDownLatch mergeAwaitLatch = mergeAwaitLatchRef.get();
         // If we found and registered any merges above, within the flushLock, then we want to ensure that they
         // complete execution. Note that since we released the lock, other merges may have been scheduled. We will
         // block until  the merges that we registered complete. As they complete, they will update toCommit to
         // replace merged segments with the result of each merge.
+        config.getIndexWriterEvents().beginMergeOnCommit();
         mergeScheduler.merge(this, MergeTrigger.FULL_FLUSH, true);
         long mergeWaitStart = System.nanoTime();
         int abandonedCount = 0;
         long waitTimeMillis = (long) (config.getMaxCommitMergeWaitSeconds() * 1000.0);
         try {
-          config.getIndexWriterEvents().beginMergeOnCommit();
           if (mergeAwaitLatch.await(waitTimeMillis, TimeUnit.MILLISECONDS) == false) {
             synchronized (this) {
               // Need to do this in a synchronized block, to make sure none of our commit merges are currently
@@ -3338,7 +3336,7 @@ public class IndexWriter implements Closeable, TwoPhaseCommit, Accountable,
               // usual, but when they finish, they won't attempt to update toCommit or modify segment reference
               // counts.
               mergeAwaitLatchRef.set(null);
-              for (MergePolicy.OneMerge commitMerge : commitMerges.merges) {
+              for (MergePolicy.OneMerge commitMerge : commitMerges) {
                 if (runningMerges.contains(commitMerge) || pendingMerges.contains(commitMerge)) {
                   abandonedCount++;
                 }
--- a/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
+++ b/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
@@ -3266,7 +3266,7 @@ public class IndexWriter implements Closeable, TwoPhaseCommit, Accountable,
               if (anyChanges) {
                 // Find any merges that can execute on commit (per MergePolicy).
                 MergePolicy.MergeSpecification mergeSpec =
-                    config.getMergePolicy().findCommitMerges(segmentInfos, this);
+                    config.getMergePolicy().findFullFlushMerges(MergeTrigger.COMMIT, segmentInfos, this);
                 if (mergeSpec != null && mergeSpec.merges.size() > 0) {
                   int mergeCount = mergeSpec.merges.size();
                   commitMerges = new ArrayList<>(mergeCount);
@@ -3323,7 +3323,7 @@ public class IndexWriter implements Closeable, TwoPhaseCommit, Accountable,
         // block until  the merges that we registered complete. As they complete, they will update toCommit to
         // replace merged segments with the result of each merge.
         config.getIndexWriterEvents().beginMergeOnCommit();
-        mergeScheduler.merge(this, MergeTrigger.FULL_FLUSH, true);
+        mergeScheduler.merge(this, MergeTrigger.COMMIT, true);
         long mergeWaitStart = System.nanoTime();
         int abandonedCount = 0;
         long waitTimeMillis = (long) (config.getMaxCommitMergeWaitSeconds() * 1000.0);
--- a/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
+++ b/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
@@ -3331,7 +3331,7 @@ public class IndexWriter implements Closeable, TwoPhaseCommit, Accountable,
           if (mergeAwaitLatch.await(waitTimeMillis, TimeUnit.MILLISECONDS) == false) {
             synchronized (this) {
               // Need to do this in a synchronized block, to make sure none of our commit merges are currently
-              // executing mergeFinished (since mergeFinished itself is called from with the IndexWriter lock).
+              // executing mergeFinished (since mergeFinished itself is called from within the IndexWriter lock).
               // After we clear the value from mergeAwaitLatchRef, the merges we schedule will still execute as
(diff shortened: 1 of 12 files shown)


lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
```diff
@@ -3147,6 +3149,38 @@ public final boolean flushNextBuffer() throws IOException {
...
+          long segmentCounter = Long.parseLong(this.info.info.name.substring(1), Character.MAX_RADIX);
+          toCommit.counter = Math.max(toCommit.counter, segmentCounter + 1);
+          Set<String> segmentNamesToRemove = new HashSet<>();
+          for (SegmentCommitInfo sci : this.segments) {
+            deleter.decRef(sci.files());
+            segmentNamesToRemove.add(sci.info.name);
+          }
+          for (int i = toCommit.size() - 1; i >= 0; i--) {
+            if (segmentNamesToRemove.contains(toCommit.info(i).info.name)) {
+              toCommit.remove(i);
```
msokolov: This relies on some peculiar semantics of SegmentInfos, namely that it manages Segment internally as a List. I guess that's true now, but I wonder if we shouldn't make SegmentInfos actually implement List, have asList return a modifiable List, or at least document and/or add a unit test to TestSegmentInfos enforcing that you can remove while iterating and expect a certain ordering of the elements? It seems to be implicit in the contract given that segments are ordered, but it's not explicitly stated in the javadocs at least.

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
```diff
@@ -3223,15 +3259,44 @@ private long prepareCommitInternal() throws IOException {
               // sneak into the commit point:
               toCommit = segmentInfos.clone();
 
+              if (anyChanges) {
+                mergeAwaitLatchRef = new AtomicReference<>();
+                MergePolicy mergeOnCommitPolicy = waitForMergeOnCommitPolicy(config.getMergePolicy(), toCommit, mergeAwaitLatchRef);
+
+                // Find any merges that can execute on commit (per MergePolicy).
+                commitMerges = mergeOnCommitPolicy.findCommitMerges(segmentInfos, this);
+                if (commitMerges != null && commitMerges.merges.size() > 0) {
```
msokolov: Would we sometimes get !=null and == 0??

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
```diff
@@ -3223,15 +3259,44 @@ private long prepareCommitInternal() throws IOException {
...
+                  int mergeCount = 0;
+                  for (MergePolicy.OneMerge oneMerge : commitMerges.merges) {
+                    if (registerMerge(oneMerge)) {
+                      mergeCount++;
+                    } else {
+                      throw new IllegalStateException("MergePolicy " + config.getMergePolicy().getClass() +
```
msokolov: I If we fail to registerMerge does it necessarily mean the segment was already being merged?

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
```diff
@@ -3223,15 +3259,44 @@ private long prepareCommitInternal() throws IOException {
               // sneak into the commit point:
               toCommit = segmentInfos.clone();
 
+              if (anyChanges) {
+                mergeAwaitLatchRef = new AtomicReference<>();
```
msokolov: could we wait to create this until we create its CountdownLatch and simplify the null checks below?

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
```diff
@@ -3252,6 +3317,53 @@ private long prepareCommitInternal() throws IOException {
...
+        mergeScheduler.merge(this, MergeTrigger.FULL_FLUSH, true);
+        long mergeWaitStart = System.nanoTime();
+        int abandonedCount = 0;
+        long waitTimeMillis = (long) (config.getMaxCommitMergeWaitSeconds() * 1000.0);
+        try {
+          config.getIndexWriterEvents().beginMergeOnCommit();
+          if (mergeAwaitLatch.await(waitTimeMillis, TimeUnit.MILLISECONDS) == false) {
+            synchronized (this) {
+              // Need to do this in a synchronized block, to make sure none of our commit merges are currently
+              // executing mergeFinished (since mergeFinished itself is called from with the IndexWriter lock).
```
msokolov: "... from within ..."

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
```diff
@@ -3252,6 +3317,53 @@ private long prepareCommitInternal() throws IOException {
...
+              // After we clear the value from mergeAwaitLatchRef, the merges we schedule will still execute as
+              // usual, but when they finish, they won't attempt to update toCommit or modify segment reference
+              // counts.
```
msokolov: Typically why would it happen that we abandon some merges? timeout? Would there be any sense in trying to interrupt them?

lucene/core/src/test/org/apache/lucene/index/TestIndexWriterMergePolicy.java
```diff
@@ -277,6 +285,92 @@ public void testSetters() {
...
+      });
+      t.start();
+      indexingThreads.add(t);
+    }
+    startingGun.countDown();
+    for (Thread t : indexingThreads) {
+      t.join();
+    }
+    writerWithMergePolicy.commit();
+    assertEquals(1, writerWithMergePolicy.listOfSegmentCommitInfos().size());
```
msfroh: I just found that this assertion sometimes fails. If there are some pending/running merges left over from the indexing threads, the segments associated with those merges will be excluded from merging on commit. I'll update this test to wait for pending merges to finish before committing.

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
msfroh: Based on e.g. NoMergePolicy, it looks like the convention to say "I found no merges" is to return null, but I can't see anything preventing a MergePolicy from returning an empty MergeSpecification.

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
msfroh: This is a little bit of hackery to share state between this thread and the threads that do the merges. 

We initialize the ref here, pass it to `waitForMergeOnCommitPolicy` on the next line to make sure it gets shared with any computed `OneMerge`s. Then, before we release the IW lock (so we're guaranteed that those `OneMerge`s haven't run yet), we populate the ref with the `CountdownLatch` (once we know what we're counting down).

That said, I think I could simplify things a lot by not using `OneMergeWrappingMergePolicy`, but rather decorating the returned `OneMerge`s (if applicable) directly. I'm going to take a stab at that approach in my next commit.

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
msfroh: Yeah -- maybe we only want to wait a short time for these merges to complete. If a merge doesn't complete, we should move forward with the commit. Since these merges will still update the "real" `SegmentInfos`, any progress they've made is arguably still useful. That's why I didn't want to interrupt them.

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
msfroh: Yes -- or it means that the `OneMerge` contains a segment that doesn't exist (anymore?).

Since we execute this while holding `IndexWriter` lock, it should not be the case that the merge policy could have returned a `OneMerge` that contains a segment that no longer exists (because updating the live `SegmentInfos` should only occur within the IW lock).

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
msfroh: I've now forgotten my motivation for traversing the list backwards. I think it was just to avoid needing to do some potentially messy/error-prone index manipulation.

In my latest commit, I realized that I could leverage `SegmentInfos.applyMergeChanges` if I replace the merged segments with their `toCommit` doppelgängers. I think that provides better encapsulation and lets `SegmentInfos` manage its own internal state better. (Also `applyMergeChanges` does some weird and wonderful list manipulation of its own that I definitely didn't want to replicate into `IndexWriter`.)

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
msfroh: D'oh! Fixed, thanks!

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
msokolov: I see, thanks for explaining!

msokolov:
Thanks, @msfroh this looks good, and I appreciate the reduction in segment count, which can lead to nice performance gains! Let's let this bake a little longer to see if anybody has concerns, then I think we can push. @ErickErickson, @dsmiley I know you've been concerned about merge policies in the past - I just want to draw your attention to this; maybe it addresses use cases you've seen?

dsmiley:
Do no push; please see the parent JIRA issue @msokolov

lucene/core/src/java/org/apache/lucene/index/IndexWriterConfig.java
```diff
@@ -484,6 +487,23 @@ public IndexWriterConfig setCommitOnClose(boolean commitOnClose) {
     return this;
   }
 
+  /**
+   * Expert: sets the amount of time to wait for merges returned by MergePolicy.findCommitMerges(...).
+   * If this time is reached, we proceed with the commit based on segments merged up to that point.
```
mikemccand: Maybe also say that any merges still running after this time will still run to completion, so they are not wasted.

lucene/core/src/java/org/apache/lucene/index/LiveIndexWriterConfig.java
```diff
@@ -480,6 +488,22 @@ public String getSoftDeletesField() {
     return softDeletesField;
   }
 
+  /**
+   * Expert: return the amount of time to wait for merges returned by by MergePolicy.findCommitMerges(...).
+   * If this time is reached, those merges will be aborted and we will wait again. If this time limit is reached again,
```
mikemccand: This is no longer true?  We will just wrap up the commit w/ those merges that did finish, and let the still running merges finish to completion?

lucene/core/src/java/org/apache/lucene/index/MergePolicy.java
```diff
@@ -526,6 +526,19 @@ public abstract MergeSpecification findForcedMerges(
   public abstract MergeSpecification findForcedDeletesMerges(
       SegmentInfos segmentInfos, MergeContext mergeContext) throws IOException;
 
+  /**
+   * Identifies merges that we want to execute (synchronously) on commit. By default, do not synchronously merge on commit.
+   *
+   * Implementers of this method should use isMergingSegment to exclude any already-merging segments from the returned
```
mikemccand: What is `isMergingSegment` here?

lucene/core/src/java/org/apache/lucene/index/MergePolicy.java
```diff
@@ -526,6 +526,19 @@ public abstract MergeSpecification findForcedMerges(
...
+   * {@link MergeSpecification}. If a segment already registered in a merge is returned, then the commit will fail.
```
mikemccand: Is this true?  Or will IW (silently) disregard that requested merge?

msfroh: The code in this PR throws an exception if IW fails to register the merge:

```
                    if (registerMerge(trackedMerge) == false) {
                      throw new IllegalStateException("MergePolicy " + config.getMergePolicy().getClass() +
                          " returned merging segments from findCommitMerges");
                    }
```

I suppose we could silently disregard the merge if we remember to count down the latch for the merge that fails to register. 

What would be less surprising to users? Explosive failure if an invalid `MergeSpecification` is returned or letting the commit happen without trying to merge? I lean slightly toward explosive failure (with what I hope is a clear explanation), but I'm happy to write it either way.

lucene/core/src/java/org/apache/lucene/index/MergePolicy.java
msfroh: Oops... that's from an earlier implementation, before I realized that `MergeContext` provides information on merging segments.

lucene/core/src/java/org/apache/lucene/index/LiveIndexWriterConfig.java
msfroh: Fixed!

lucene/core/src/java/org/apache/lucene/index/IndexWriterConfig.java
msfroh: Done

lucene/core/src/java/org/apache/lucene/index/MergePolicy.java
mikemccand: OK I agree to explosive failure!  Less trappy.

lucene/core/src/java/org/apache/lucene/index/IndexWriterEvents.java
```diff
@@ -0,0 +1,57 @@
...
+ */
+
+// Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+package org.apache.lucene.index;
+
+/**
+ * Callback interface to signal various actions taken by IndexWriter.
+ */
+public interface IndexWriterEvents {
```
mikemccand: Can you add `@lucene.experimental` here?  That reserves the right to make breaking changes even on feature releases ...

lucene/core/src/java/org/apache/lucene/index/MergePolicy.java
```diff
@@ -526,6 +526,19 @@ public abstract MergeSpecification findForcedMerges(
   public abstract MergeSpecification findForcedDeletesMerges(
       SegmentInfos segmentInfos, MergeContext mergeContext) throws IOException;
 
+  /**
+   * Identifies merges that we want to execute (synchronously) on commit. By default, do not synchronously merge on commit.
+   *
+   * If a returned {@link OneMerge} includes a segment already included in a registered merge, then the commit will fail.
```
mikemccand: Can you say what exception will be thrown in that case (or add an `@throws`, below)?

lucene/core/src/java/org/apache/lucene/index/MergePolicy.java
```diff
@@ -526,6 +526,19 @@ public abstract MergeSpecification findForcedMerges(
...
+   * Use {@link MergeContext#getMergingSegments()} to determine which segments are currently registered to merge.
+   *
+   * @param segmentInfos the total set of segments in the index (while preparing the commit)
+   * @param mergeContext the IndexWriter to find the merges on
```
mikemccand: Hmm, maybe improve this (it's not an `IndexWriter`) to say the merge policy may use this to find already merging segments?

lucene/core/src/java/org/apache/lucene/index/MergePolicy.java
```diff
@@ -526,6 +526,19 @@ public abstract MergeSpecification findForcedMerges(
   public abstract MergeSpecification findForcedDeletesMerges(
       SegmentInfos segmentInfos, MergeContext mergeContext) throws IOException;
 
+  /**
+   * Identifies merges that we want to execute (synchronously) on commit. By default, do not synchronously merge on commit.
```
mikemccand: Can you describe a bit more?  E.g. explain that any merges returned here which then finish within the allotted time will be reflected in the commit, so that the number of segments in the commit point may be decreased?

msfroh: While updating the comments here, it occurred to me that if we want to leverage this same method in future to trigger a merge on NRT refresh, would it make sense to call it e.g. `findFlushMerges` or `findFullFlushMerges` instead?

mikemccand: Good thinking!  +1 for `findFullFlushMerges`, to make it clear it's a FULL flush (`commit` or `refresh`) and not an ordinary flush e.g. when the in-memory segments are taking too much memory and some are flushed.

And then I guess the method could look at `MergeTrigger` to differentiate whether it's a `commit` or a `refresh`?

msfroh: Renamed in latest commit and added a new MergeTrigger.

Passing the `MergeTrigger` to the `MergeScheduler` got me thinking that it might be possible to raise the priority of these full-flush merges, though it would be easier to manage that in the call to `registerMerge`, which could insert them at the front of the pending merge queue. I think that's beyond the scope of this change, but it could make for an interesting follow-up.

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
```diff
@@ -3147,6 +3149,42 @@ public final boolean flushNextBuffer() throws IOException {
...
+            if (mergedSegmentNames.contains(sci.info.name)) {
+              toCommitMergedAwaySegments.add(sci);
+            }
+          }
+          // Construct a OneMerge that applies to toCommit
+          MergePolicy.OneMerge applicableMerge = new MergePolicy.OneMerge(toCommitMergedAwaySegments);
+          applicableMerge.info = this.info.clone();
+          long segmentCounter = Long.parseLong(this.info.info.name.substring(1), Character.MAX_RADIX);
+          toCommit.counter = Math.max(toCommit.counter, segmentCounter + 1);
+          toCommit.applyMergeChanges(applicableMerge, false);
```
dnhatn: We should modify `toCommit` under `IndexWrite.this` lock (or a private synchronization between this method and `commitInternal`).

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
```diff
@@ -3252,6 +3315,53 @@ private long prepareCommitInternal() throws IOException {
...
+        long waitTimeMillis = (long) (config.getMaxCommitMergeWaitSeconds() * 1000.0);
+        try {
+          if (mergeAwaitLatch.await(waitTimeMillis, TimeUnit.MILLISECONDS) == false) {
+            synchronized (this) {
+              // Need to do this in a synchronized block, to make sure none of our commit merges are currently
+              // executing mergeFinished (since mergeFinished itself is called from within the IndexWriter lock).
+              // After we clear the value from mergeAwaitLatchRef, the merges we schedule will still execute as
+              // usual, but when they finish, they won't attempt to update toCommit or modify segment reference
+              // counts.
+              mergeAwaitLatchRef.set(null);
```
dnhatn: I think we should set `mergeAwaitLatchRef` in the `else` branch.

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
```diff
@@ -3147,6 +3149,42 @@ public final boolean flushNextBuffer() throws IOException {
...
   }
 
+  private MergePolicy.OneMerge updateSegmentInfosOnMergeFinish(MergePolicy.OneMerge merge, final SegmentInfos toCommit,
+                                                                AtomicReference<CountDownLatch> mergeLatchRef) {
+    return new MergePolicy.OneMerge(merge.segments) {
+      public void mergeFinished() throws IOException {
+        super.mergeFinished();
+        CountDownLatch mergeAwaitLatch = mergeLatchRef.get();
+        if (mergeAwaitLatch == null) {
+          // Commit thread timed out waiting for this merge and moved on. No need to manipulate toCommit.
```
dnhatn: We need a stronger synchronization to make sure that we won't modify `toCommit` if `commitInternal` has stopped waiting for these merges.

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
dnhatn: Sorry I misread this.

dnhatn:
Hmm, I missed the fact that `mergeFinished` is executed under IndexWriter lock. I will dig into this again. Please ignore my previous comments.

lucene/core/src/java/org/apache/lucene/index/IndexWriterEvents.java
```diff
@@ -0,0 +1,59 @@
...
+
+    @Override
+    public void abandonedMergesOnCommit(int abandonedCount) { }
+  };
+
+  /**
+   * Signals the start of waiting for a merge on commit, returned from
+   * {@link MergePolicy#findFullFlushMerges(MergeTrigger, SegmentInfos, MergePolicy.MergeContext)}.
+   */
+  void beginMergeOnCommit();
```
s1monw: I am not really happy with this interface. First and foremost it's only partially used in this PR. I also think it doesn't belong here but rather into a merge policy? I think IW and merge lifecycle should not be tightly coupled. Can we achieve the same with an interface a MP can provide to the IW rather than setting it on the IW config. A pull model should be used here instead IMO.

PR #1157, opened 2020-01-09
Add RAT check using Gradle

madrob:
Still a WIP, but most of the way there.

Need to move the task invocation out to the modules where it belongs. I didn't completely understand the logic for what was included and excluded and how we made those decisions initially.

Am not sure if we should be printing all the violations by default or hiding that behind a `--info` or `--verbose` flag.

### DIFF
--- a/gradle/validation/rat-sources.gradle
+++ b/gradle/validation/rat-sources.gradle
@@ -36,13 +36,9 @@ subprojects {
             description = 'Runs Apache Rat checks.'
         }
 
-        // ant had an exclude for ./lucene/analysis/kuromoji/src/test/org/apache/lucene/analysis/ja/bocchan.utf-8
-        // lucene/analysis/kuromoji/build.xml:  <property name="rat.excludes" value="**/*.txt,**/bocchan.utf-8"/>
-        // why don't we need it anymore?
-
         if (project.path == ":lucene:analysis:common") {
             rat {
-                excludes += [
+                srcExcludes += [
                     "**/*.aff",
                     "**/*.dic",
                     "**/charfilter/*.htm*",
@@ -51,9 +47,31 @@ subprojects {
             }
         }
 
-        if (project.path == ":lucene:highlighter") {
+        if (project.path == ":lucene:analysis:kuromoji") {
+            rat {
+                srcExcludes += [
+                    // whether rat detects this as binary or not is platform dependent?!
+                    "**/bocchan.utf-8"
+                ]
+            }
+        }
+
+        if (project.path == ":lucene:analysis:opennlp") {
             rat {
                 excludes += [
+                    "src/tools/test-model-data/chunks.txt",
+                    "src/tools/test-model-data/lemmas.txt",
+                    "src/tools/test-model-data/ner.txt",
+                    "src/tools/test-model-data/pos.txt",
+                    "src/tools/test-model-data/sentences.txt",
+                    "src/tools/test-model-data/tokenizer.txt",
+                ]
+            }
+        }
+
+        if (project.path == ":lucene:highlighter") {
+            rat {
+                srcExcludes += [
                     "**/CambridgeMA.utf8"
                 ]
             }
@@ -61,17 +79,40 @@ subprojects {
 
         if (project.path == ":lucene:suggest") {
             rat {
-                excludes += [
+                srcExcludes += [
                     "**/Top50KWiki.utf8",
                     "**/stop-snowball.txt"
                 ]
             }
         }
 
+        // TODO this doesn't have a corresponding gradle project so it never gets triggered
+        if (project.path == ":lucene:tools") {
+            rat {
+                includes += [
+                    "forbiddenApis/**",
+                    "prettify/**"
+                ]
+            }
+        }
+
+
         if (project.path == ":solr:core") {
             rat {
-                excludes += [
+                srcExcludes += [
                     "**/htmlStripReaderTest.html"
+                ]
+            }
+        }
+
+        if (project.path == ":solr:webapp") {
+            rat {
+                includes = [ "**" ]
+                excludes += [
+                    "web/img/**",
+                    "*.iml",
+                    "build.gradle",
+                    "build/**",
                 ]
             }
         }
@@ -82,19 +123,21 @@ subprojects {
 class RatTask extends DefaultTask {
     @Input
     List<String> includes = [
+        "*.xml",
+        "src/tools/**"
     ]
 
     @Input
-    List<String> excludes = [
+    List<String> excludes = []
+
+    @Input
+    List<String> srcExcludes = [
         "**/TODO",
         "**/*.txt",
         "**/*.iml",
         "**/*.gradle",
         "build/**"
     ]
-
-    @Input
-    List<String> additionalExcludes = []
 
     def reportDir = new File(project.buildDir, 'rat')
     def xmlReport = new File(reportDir, 'rat-report.xml')
@@ -107,13 +150,8 @@ class RatTask extends DefaultTask {
         def rat = NamespaceBuilder.newInstance(ant, uri)
         rat.report(format: 'xml', reportFile: xmlReport, addDefaultLicenseMatchers: true) {
             ant.fileset(dir: "${project.projectDir}") {
-                ant.include(name: '*.xml')
-                includes.each {
-                    ant.include(name: it)
-                }
-                additionalExcludes.each {
-                    ant.exclude(name: it)
-                }
+                includes.each { pattern -> ant.include(name: pattern) }
+                excludes.each { pattern -> ant.exclude(name: pattern) }
             }
 
             [
@@ -121,7 +159,7 @@ class RatTask extends DefaultTask {
                 project.sourceSets.test.java.srcDirs,
             ].flatten().each { srcLocation ->
                 ant.fileset(dir: srcLocation, erroronmissingdir: false) {
-                    excludes.each { x -> ant.exclude(name: x) }
+                    srcExcludes.each { pattern -> ant.exclude(name: pattern) }
                 }
             }
 
--- a/gradle/validation/rat-sources.gradle
+++ b/gradle/validation/rat-sources.gradle
@@ -83,7 +83,7 @@ allprojects {
             srcExcludes += [
                 "**/CambridgeMA.utf8"
             ]
-    }
+        }
     }
 
     if (project.path == ":lucene:suggest") {
--- a/gradle/validation/rat-sources.gradle
+++ b/gradle/validation/rat-sources.gradle
@@ -29,10 +29,25 @@ configure(rootProject) {
     }
 }
 
-subprojects {
+allprojects {
     task("rat", type: RatTask) {
         group = 'Verification'
         description = 'Runs Apache Rat checks.'
+    }
+
+    if (project == rootProject) {
+        rat {
+            includes += [
+                "buildSrc/**/*.java",
+                "lucene/tools/forbiddenApis/**",
+                "lucene/tools/prettify/**",
+                // "dev-tools/**"
+            ]
+            excludes += [
+                // Unclear if this needs ASF header, depends on how much was copied from ElasticSearch
+                "**/ErrorReportingTestListener.java"
+            ]
+        }
     }
 
     if (project.path == ":lucene:analysis:common") {
@@ -58,12 +73,7 @@ subprojects {
     if (project.path == ":lucene:analysis:opennlp") {
         rat {
             excludes += [
-                "src/tools/test-model-data/chunks.txt",
-                "src/tools/test-model-data/lemmas.txt",
-                "src/tools/test-model-data/ner.txt",
-                "src/tools/test-model-data/pos.txt",
-                "src/tools/test-model-data/sentences.txt",
-                "src/tools/test-model-data/tokenizer.txt",
+                "src/tools/test-model-data/*.txt",
             ]
         }
     }
@@ -73,7 +83,7 @@ subprojects {
             srcExcludes += [
                 "**/CambridgeMA.utf8"
             ]
-        }
+    }
     }
 
     if (project.path == ":lucene:suggest") {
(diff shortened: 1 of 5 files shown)


gradle/validation/rat-sources.gradle
```diff
@@ -0,0 +1,167 @@
...
+            System.setProperty("file.encoding", "UTF-8") //affects the output of the ant rat task
+            generateXmlReport(reportDir)
+            printUnknownFiles()
+        } finally {
+            System.setProperty("file.encoding", origEncoding)
+        }
+    }
+}
+
+class RatPlugin implements Plugin<Project> {
```
dweiss: I wouldn't go into plugin creation at this moment. You don't need a plugin to define a task - just define the task class and add tasks of that type on projects where they're applicable. You can use ant tasks from gradle directly - no need to isolate anything.

gradle/validation/rat-sources.gradle
```diff
@@ -0,0 +1,167 @@
...
+
+    def reportDir = project.file('build/rat')
+    def xmlReport = new File(reportDir, 'rat-report.xml')
+
+    def generateXmlReport(File reportDir) {
+        // Probably better to use the IsolatedAntBuilder if we can, but it seems to have issues with substringMatcher
+        // def antBuilder = services.get(IsolatedAntBuilder)
+
+        def ratClasspath = project.configurations.rat
+        def projectPath = project.getRootDir().getAbsolutePath()
```
dweiss: This is wrong I think. It should be project.projectDir

gradle/validation/rat-sources.gradle
```diff
@@ -0,0 +1,167 @@
...
+        if (unknownLicenses > 0) {
+            throw new GradleException("Found " + unknownLicenses + " files with " +
+                    "unknown licenses.")
+        }
+    }
+
+    @TaskAction
+    def rat() {
+        if (!reportDir.exists()) {
+            reportDir.mkdirs()
```
dweiss: mkdirs alone is fine; it'll check for existence anyway.

gradle/validation/rat-sources.gradle
```diff
@@ -0,0 +1,167 @@
...
+        }
+        def origEncoding = System.getProperty("file.encoding")
+        try {
+            System.setProperty("file.encoding", "UTF-8") //affects the output of the ant rat task
```
dweiss: This is crappy. An XML file should be UTF-8 unless it declares a different encoding in the preamble. We should not have to adjust file.encoding (in fact, we must not) - it seems like a bug somewhere down the toolchain.

madrob: Yea, I was a little surprised at it too, but didn't want to touch it when copying from the original implementation. I'm assuming they ran into some issue that we may or may not see. I'll comment it out for now, but leave a note so if we're in trouble later we might be able to identify why a bit easier.

madrob:
Cleaned a bunch of stuff up, included your review feedback. Thanks, @dweiss!

One piece that I'm still struggling with is that `./gradlew rat` will execute on the root but doesn't delegate to `:lucene:rat` and `:solr:rat` (and in fact those targets don't even exist), and then I'd like those to delegate to rat in all the sub-modules. I tried going through our other validation tasks to look for examples and quickly got lost - can you point me to the best practices here?

madrob:
Almost there! Precommit currently fails with some license failures, I'll need to look at that deeper what exclusions we're actually using in the ant build, but I think we're super close now.

dweiss:
I'll take a look later, Mike. As for applying tasks and anything else -- think of the project structure as a graph. You attach things to this graph in two passes (evaluation, configuration), followed by execution of tasks attached to this graph (in topological order of dependencies).

It is conceptually simple. The devil hides in details of how gradle is evaluated, deferred evaluated-collections, etc. This should be helpful:

https://docs.gradle.org/current/userguide/build_lifecycle.html

I'll review the patch and maybe correct it before committing; when you take a look at the commit vs. your patch you'll see the differences made - I think it'll be easier and faster than explaining (but go ahead and ask if you don't understand something).

gradle/validation/rat-sources.gradle
```diff
@@ -0,0 +1,174 @@
...
+        rat
+    }
+
+    dependencies {
+        rat 'org.apache.rat:apache-rat-tasks:0.13'
+    }
+}
+
+subprojects {
+    plugins.withId("java", {
```
dweiss: These rat checks seem to apply not just to java projects but to other files (non-java files) as well. If so then the task should be applicable to any project, not just those with the java plugin?

I think we'd need to cross-check with what the ant script does (whether it runs those checks on non-java folders).

gradle/validation/rat-sources.gradle
```diff
@@ -0,0 +1,174 @@
...
+                ant.include(name: '*.xml')
+                includes.each {
+                    ant.include(name: it)
+                }
+                additionalExcludes.each {
+                    ant.exclude(name: it)
+                }
+            }
+
+            // There may be a more efficient way to write these next two blocks since they use the same logic
```
dweiss: something like [project.sourceSets.main, project.sourceSets.test].each { location -> ... }

gradle/validation/rat-sources.gradle
```diff
@@ -0,0 +1,174 @@
...
+            project.sourceSets.main.java.srcDirs.each {
+                ant.fileset(dir: it, erroronmissingdir: false) {
```
dweiss: I'd be explicit and use variable for closures rather than "it", especially in larger code blocks.

gradle/validation/rat-sources.gradle
```diff
@@ -0,0 +1,174 @@
...
+            approvedLicense(familyName: "Generated")
+        }
+    }
+
+    def printUnknownFiles() {
+        def ratXml = new XmlParser().parse(xmlReport)
+        def unknownLicenses = 0
+        ratXml.resource.each { resource ->
+            if (resource.'license-approval'.@name[0] == "false") {
+                println('Unknown license: ' + resource.@name)
```
dweiss: collect offending resources to an array and throw the exception with all offenders -- see other validation tasks (license checksum validation for example)?

this matters when tasks run concurrently - easier to spot the problem (reported by gradle at the end of the build).

gradle/validation/rat-sources.gradle
```diff
@@ -0,0 +1,174 @@
...
+        if (unknownLicenses > 0) {
+            throw new GradleException("Found " + unknownLicenses + " files with " +
+                    "unknown licenses.")
+        }
+    }
+
+    @TaskAction
+    def rat() {
+        reportDir.mkdirs()
+        def origEncoding = System.getProperty("file.encoding")
```
dweiss: I can't believe ant rat would change the default encoding (does it?!). If it does it's crazy. File encoding should be a property of rat's ant task not the system global default

dweiss:
[patch.txt](https://github.com/apache/lucene-solr/files/4053932/patch.txt)
I reworked a few things, Mike. Now... rat hangs for me (?). I wonder if we could bypass ant and use rat classes directly - the conversion between gradle and ant filesets is a bit clumsy (it's fine as a first draft though!).

Inclusion/ exclusion patterns have to be reviewed and consolidated with ant: some projects override these, gradle should exclude project-local build/**, etc.

dweiss:
It doesn't hang. It just takes a loooooong time on the highlighter - don't know why. The two failures are related to extra exclusions present in the ant build file of the corresponding project.

dweiss:
[patch.txt](https://github.com/apache/lucene-solr/files/4054015/patch.txt)

Updated patch. Some files still fail; it'd have to be investigated.

dweiss:
Once you apply the patch included in my last comment change the hardcoded &quot; here:

pattern(substring: "Licensed under the Apache License, Version 2.0 (the &quot;License&quot;)") 

gradle files are not XMLs so this is responsible for at least some failures.

gradle/validation/rat-sources.gradle
madrob: This is a quirk of Gradle where the Java plugin is what actually creates sourceSets.main and test, afaict. It also filters out projects where sourceSets.main doesn't exist, which was throwing exceptions for me.

madrob:
Updated based on your patch, pushed latest to this PR.

We're missing rat check on `buildSrc` (no ant equivalent), `lucene/tools` (does not have build.gradle file), and `dev-tools` (ant rat misses this too) - not sure which of those need to be addressed here.

There's definitely something else going on here as one of the excluded files from ant is either skipped or passes on Gradle, so more digging needs to happen here.

One other question of best practices - do we want all the project exclusions in a central rat.gradle file, or spread out and local to each project?

madrob:
One other thought - we probably shouldn't exclude the `*.gradle` files - there's no reason for those not to have license headers like our `build.xml` files currently do. WDYT?

dweiss:
To be honest I would prefer not to add those license headers to build files, unless it's a requirement of apache legal -- this would have to be verified. For one thing, I don't perceive build files as something particularly valuable as an intellectual property (although it surely does require intellectual input to write them). But even if then the distribution bundle comes with a top-level license file that covers them?

The downside of requiring this boilerplate is that they bloat each and every script.

dweiss:
bq. One other question of best practices - do we want all the project exclusions in a central rat.gradle file, or spread out and local to each project?

This is a good question and I think the answer is really subjective. 

Both approaches have advantages and disadvantages. I personally like the "aspect" oriented approach when everything related to a particular build function is gathered in a single file. So yes, for rat it'd be just that single file -- any special handling of that aspect, exclusions, etc. would be collected there. In an extreme case you should be able to enable/ disable rat just by commenting out the apply block in the master build file.

As the build evolves over time it may need to be changed. For example when you move parts of the build into buildSrc and wrap it in plugins it will become necessary to move project-specific logic outside.

For now I'd rather keep it in this "aspect-oriented" form if you don't mind (but this is a subjective decision, not any better or common practice).

gradle/validation/rat-sources.gradle
dweiss: That's not a quirk :) What I'm saying is that you explicitly rely on the project to be a java convention project. We don't need to require this but then have to decide what to use as source folders.

madrob: We _could_ explicitly code the `src/java` and `src/test` paths here, but we are also already doing this in `gradle/ant-compat/folder-layout.gradle` and I'd like to stick to doing it in one place. As for projects with files but not the java plugin... do we have those? I'll need to figure out what happens if we have a failed file in lucene/ directory directly...

madrob:
One final round of updates:
* I moved the java plugin check to inside of the task and verified that the other projects would actually get called/fail with bad headers.
* Added support for incremental builds so that should decrease the impact time to most developers
* Added in the missing directories that didn't have their own projects. And a landmine if those do end up becoming projects so that somebody pays attention and fixes it (probably one of us)

I think that's everything in this PR, and we're ready to merge? 

We can go and switch from ant task to rat classes directly in a later change if we decide that's worthwhile.

> To be honest I would prefer not to add those license headers to build files, unless it's a requirement of apache legal -- this would have to be verified. For one thing, I don't perceive build files as something particularly valuable as an intellectual property (although it surely does require intellectual input to write them). But even if then the distribution bundle comes with a top-level license file that covers them?

I guess we'll hear from LEGAL about this one way or the other. Not going to bother adding them for this PR until somebody does the checks.

> For now I'd rather keep it in this "aspect-oriented" form if you don't mind (but this is a subjective decision, not any better or common practice).

That's fine. I just noticed this and was curious since we were moving from the project oriented approach with ant to this approach and wanted to make sure it was a conscious decision.

dweiss:
Thanks Mike. I'll merge it in today. I actually looked at apache rat sources yesterday -- they are indeed riddled with encoding issues (writer on top of printstream, etc.) and they're not easy to use directly to apply just to a bunch of files instead of directory+inclusions/exclusions.

My motivation here was to replace the whole fileset trickery with gradle's file collections. Then it's easier to pick what you need, pass it to task inputs, etc. Also, what rat does is merely simple pattern detection (line matching)... this could also be written in just plain groovy with relative ease - perhaps at some point we can just switch over to having our own license check where we can control things better. 

For now I think it's fine.

gradle/validation/rat-sources.gradle
```diff
@@ -0,0 +1,266 @@
...
+    if (project == rootProject) {
+        rat {
+            includes += [
+                "buildSrc/**/*.java",
+                "lucene/tools/forbiddenApis/**",
+                "lucene/tools/prettify/**",
+                // "dev-tools/**"
+            ]
+            excludes += [
+                // Unclear if this needs ASF header, depends on how much was copied from ElasticSearch
```
dweiss: The idea was but I changed a whole bunch of stuff along the way. There are similarities and the attribution is there but I think we can consider it our own.

gradle/validation/rat-sources.gradle
```diff
@@ -0,0 +1,266 @@
...
+                "**/charfilter/*.htm*",
+                "**/*LuceneResourcesWikiPage.html"
+            ]
+        }
+    }
+
+    if (project.path == ":lucene:analysis:kuromoji") {
+        rat {
+            srcExcludes += [
+                // whether rat detects this as binary or not is platform dependent?!
```
dweiss: This is one of the reasons I'd rather have our own check - we don't have to rely on a black box (or look inside rat's code to figure out what it does).

gradle/validation/rat-sources.gradle
```diff
@@ -0,0 +1,266 @@
...
+            ]
+        }
+    }
+
+    if (project.path == ":lucene:tools") {
+        rat {
+            includes += [
+                "forbiddenApis/**",
+                "prettify/**",
+                // If/when :lucene:tools becomes a gradle project, then the following line will fail
```
dweiss: I don't think there will be lucene:tools anymore once we depart from ant - I don't think there is any relevant code in there that should stay.

versions.props
```diff
@@ -72,6 +72,7 @@ org.apache.opennlp:opennlp-tools=1.9.1
 org.apache.pdfbox:*=2.0.17
 org.apache.pdfbox:jempbox=1.8.16
 org.apache.poi:*=4.1.1
+org.apache.rat:apache-rat:0.11
```
dweiss: Did you run gradlew precommit? :) Because I think the lock update file isn't included in the patch.

dweiss:
Thanks Mike. I'll fix the remaining tiny issues on gradle-master.

PR #1160, opened 2020-01-10
LUCENE-9125: Improve Automaton.step() with binary search

bruno-roustant:
and introduce Automaton.next().

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/util/automaton/Automaton.java
+++ b/lucene/core/src/java/org/apache/lucene/util/automaton/Automaton.java
@@ -656,32 +656,7 @@ public class Automaton implements Accountable {
    * @return destination state, -1 if no matching outgoing transition
    */
   public int step(int state, int label) {
-    assert state >= 0;
-    assert label >= 0;
-    int stateIndex = 2 * state;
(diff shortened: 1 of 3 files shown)


lucene/core/src/java/org/apache/lucene/util/automaton/Automaton.java
```diff
@@ -658,22 +658,84 @@ public String toDot() {
...
+    // Copy of step() method with
+    // - binary search 'low' bound initialized to transition.transitionUpto.
+    // - param transition .dest/.min/.max/.transitionUpto set to the matching transition.
+    assert transition.source >= 0;
+    assert label >= 0;
+    int stateIndex = 2 * transition.source;
+    int firstTransitionIndex = states[stateIndex];
+    int numTransitions = states[stateIndex + 1];
+
+    // Since transitions are sorted,
```
dweiss: can we extract the binary search subroutine into a separate method? It'd inline anyway, very likely and it'd result in less code.

bruno-roustant: I'll try.

PR #1163, opened 2020-01-13
SOLR-14186: Enforce CRLF in Windows files with .gitattributes

gerlowskija:
# Description

We have nothing that ensures that our Windows files (solr.cmd, etc.) keep the CRLF files they need to run.

# Solution

Introduces a .gitattributes file right under the solr/ directory to ensure that our various OS-specific files are handled correctly.

# Tests
Manual testing on Windows and Linux checkouts to make sure that CRLF is retained in our Windows-specific files.

# Checklist


dweiss:
This should be a precommit check (in addition to whatever editing magic you specify via .editor, .gitattributes, etc.).

gerlowskija:
I thought about a precommit check, but discarded the idea for a few reasons.

It would require duplicating the whitelist/blacklist of file types/locations in two separate formats.  Which is fine if it buys us something that we don't get with `gitattributes` alone, but I'm not sure that it does.  Is there a specific case that `gitattributes` would get wrong that precommit would catch for us?

dweiss:
bq. It would require duplicating the whitelist/blacklist of file types/locations in two separate formats.

It doesn't have to be that general. You only want to check a few launch scripts after all. It can wait until gradle lands in - then it's literally a few lines.

I've had some issues in the past with .gitattributes: third-party tools not understanding it or ignoring it, etc. Perhaps the situation has improved but a precommit check is fairly definitive: you don't rely on any external tool.

janhoy:
I thought it was git itself that enforces gitattributes, not 3rd party tools. If editors would parse and respect .gitattributes then there would be no need for .editorconfig?

Don’t know if git uses it only when checking out files from repo or also when committing changes? I hope both.

dweiss:
You assume a single "git" but there are multiple versions of the tool and even multiple implementations (jgit, for example). An assumption editors support or understand .gitattributes (or .editorconfig) is probably overly optimistic either... Not that I use nodepad.exe... but I'm sure there are folks who do! :)

And more seriously: I've experimented with both options in the past and I use them in other projects but still prefer an explicit check for sanity. Ideally an integration test that takes the final ZIP/TGZ bundle and verifies distribution sanity - this also covers the (not unlikely) possibility of build tools messing up files while doing content-filtering, copying, whatever else.

gerlowskija:
> I thought it was git itself that enforces gitattributes, not 3rd party tools.
That's what I thought too.  What sort of editor/tool issues have you seen Dawid?

> Don't know if git uses it only when checking out files from repo or also when committing changes?
As far as I can tell from the docs [here,](https://git-scm.com/docs/gitattributes#_checking_out_and_checking_in) it looks like both.

dweiss:
You asked about what can go wrong with git-specific stuff, @gerlowskija -- try to apply a patch we have for Kuromoji dictionaries (Noun.proper.csv.patch) with git and then try the same with IntellIJ IDEA's "apply patch); or any other software that uses jgit for that matter... you'll see why I insisted on having validation checks on build side. :)

gerlowskija:
Fair enough.  Did you attach that patch or is it on a jira somewhere?  I'm curious to see what happens now...

dweiss:
Look at jgit's implementation of ApplyCommand and I think you'll see what's wrong.
https://github.com/eclipse/jgit/blob/master/org.eclipse.jgit/src/org/eclipse/jgit/api/ApplyCommand.java

https://github.com/eclipse/jgit/blob/master/org.eclipse.jgit/src/org/eclipse/jgit/api/ApplyCommand.java#L231-L234

PR #1170, opened 2020-01-15
LUCENE-9141: Simplify LatLonShapeXQuery API.

iverase:
This PR tries to simplify the LatLonShape query API. The trick for this simplification is to introduce a new interface called `LatLonGeometry` which is implemented by objects defining queryable geometries, that is `Polygon`, `Line` and the newly introduced `Point`. 

The main feature of those geometries is that they can return a `Component2D` object that can be used for query the index.


iverase:
Update the PR by making LatLonGeometry an abstract class so the method that return a Component2D can be made protected.

lucene/core/src/java/org/apache/lucene/document/LatLonShape.java
```diff
@@ -107,40 +111,36 @@ public static Query newBoxQuery(String field, QueryRelation queryRelation, doubl
...
+    Point[] pointArray = new Point[points.length];
+    for (int i =0; i < points.length; i++) {
+      pointArray[i] = new Point(points[i][0], points[i][1]);
+    }
+    return newGeometryCollectionQuery(field, queryRelation, pointArray);
+  }
+
+  /** create a query to find all indexed geo shapes that intersect a provided geometry collection.
+   **/
+  public static Query newGeometryCollectionQuery(String field, QueryRelation queryRelation, LatLonGeometry... latLonGeometries) {
```
nknize: nit: do you think we could just name this `newGeometryQuery`? This way we just treat geometries and single and/or collections? Feels a little less confusing IMHO.

iverase: Yes, I think it is a better name

PR #1171, opened 2020-01-15
SOLR-13892: Add 'top-level' docValues Join implementation

gerlowskija:
# Description

Some join usecases are better served (performance wise) by using top-level docValues structures, because they avoid some wasteful seeking and iteration that the per-segment alternatives do.

# Solution
This PR introduces a "TopLevelJoinQuery" which can be run on join's where the fields in question have docValues enabled.  This scales much better with the number of "from" query matches than existing alternatives.  Since it's implemented as a Two-Phase Iterator it scales less well by the number of overall matches.  See SOLR-13892 for more in-depth performance discussion.

# Tests

Beefed up tests in `TestJoin` to prove correctness.  Wrote a performance test driver `TestJoinQueryPerformance` that allows reproduction of the perf results.  (Though this will not be committed in the final version) 

# Checklist


gerlowskija:
I have mixed feelings about the last commit in the series I just pushed up (`Remove single-/multi-value duplication`).  It helps get rid of a lot of duplication across SortedDocValues and SortedSetDocValues codepaths, but it also addresses a problem that I imagine people are running into all over the place.  Is there some established way for handling this in Solr?  I imagine tons of query implementations must run into this...

Need to think it over a bit...

dsmiley:
See Lucene's SingletonSortedSetDocValues

gerlowskija:
That's exactly what I was looking for.  I knew it had to exist already.  Thanks David!

solr/core/src/java/org/apache/solr/search/JoinQParserPlugin.java
```diff
@@ -59,67 +60,124 @@
...
+          if (fromHolder != null) fromCoreOpenTime = fromHolder.get().getOpenNanoTime();
+        } finally {
+          otherReq.close();
+          fromCore.close();
+          if (fromHolder != null) fromHolder.decref();
+        }
+      } else {
+        coreName = null;
+        QParser fromQueryParser = qparser.subQuery(v, null);
+        fromQueryParser.setIsFilter(true);
```
dsmiley: Perhaps these two lines, setting is filter on the QParser and calling getQuery should be pulled out of the block as it is (or should be) in common with the positive side of the if-else?

solr/core/src/java/org/apache/solr/search/JoinQParserPlugin.java
```diff
@@ -139,11 +197,16 @@ public static Query createJoinQuery(Query subQuery, String fromField, String toF
 
 
 class JoinQuery extends Query {
+  private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
```
dsmiley: Not used?

solr/core/src/java/org/apache/solr/search/JoinQParserPlugin.java
```diff
@@ -139,11 +197,16 @@ public static Query createJoinQuery(Query subQuery, String fromField, String toF
...
+
   String fromField;
   String toField;
   String fromIndex; // TODO: name is missleading here compared to JoinQParserPlugin usage - here it must be a core name
   Query q;
   long fromCoreOpenTime;
+  private boolean cache;
```
dsmiley: Not used?

solr/core/src/java/org/apache/solr/search/TopLevelJoinQuery.java
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    final LeafReader leafReader = solrSearcher.getSlowAtomicReader();
+    if (field.multiValued()) {
+      return DocValues.getSortedSet(leafReader, fieldName);
+    }
+    return DocValues.singleton(DocValues.getSorted(leafReader, fieldName));
+  }
+
+  private LongBitSet findOrdinalsMatchingFromQuery(SolrIndexSearcher fromSearcher, SortedSetDocValues fromDocValues) throws IOException {
```
dsmiley: Couldn't you rename the method and it's variables to remove "From" and it would be just as valid?  Okay I see why not; this is an instance method that refers to 'fromField' but if that's the only distinction then maybe it should be a static method and take that argument?

solr/core/src/java/org/apache/solr/search/join/MultiValueTermOrdinalCollector.java
```diff
@@ -0,0 +1,72 @@
...
+import org.apache.lucene.search.ScoreMode;
+import org.apache.lucene.util.LongBitSet;
+import org.apache.solr.search.DelegatingCollector;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Populates a bitset of (top-level) ordinals based on field values in a multi-valued field.
+ */
+public class MultiValueTermOrdinalCollector extends DelegatingCollector {
```
dsmiley: Why extend DelegatingCollector?

solr/core/src/java/org/apache/solr/search/join/MultiValueTermOrdinalCollector.java
```diff
@@ -0,0 +1,72 @@
...
+    this.fieldName = fieldName;
+    this.topLevelDocValues = topLevelDocValues;
+    this.topLevelDocValuesBitSet = topLevelDocValuesBitSet;
+  }
+
+  public ScoreMode scoreMode() {
+    return ScoreMode.COMPLETE_NO_SCORES;
+  }
+
+  public boolean needsScores(){
```
dsmiley: What's this for?  Looks old/legacy before needsScores became scoreMode yet you just wrote this code.

solr/core/src/java/org/apache/solr/search/join/MultiValueTermOrdinalCollector.java
```diff
@@ -0,0 +1,72 @@
...
+import org.apache.lucene.util.LongBitSet;
+import org.apache.solr.search.DelegatingCollector;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Populates a bitset of (top-level) ordinals based on field values in a multi-valued field.
+ */
+public class MultiValueTermOrdinalCollector extends DelegatingCollector {
+  private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
```
dsmiley: Nothing to log

solr/core/src/java/org/apache/solr/search/join/MultiValueTermOrdinalCollector.java
```diff
@@ -0,0 +1,72 @@
...
+
+  private int docBase;
+  private SortedSetDocValues topLevelDocValues;
+  private final String fieldName;
+  private final LongBitSet topLevelDocValuesBitSet;
```
dsmiley: document this is the outgoing data structure expressing the results.  Could be a trivial comment

solr/core/src/java/org/apache/solr/search/join/TopLevelDVTermsCollector.java
```diff
@@ -0,0 +1,78 @@
...
+import org.apache.lucene.util.LongBitSet;
+import org.apache.solr.search.DelegatingCollector;
+
+/**
+ * Collects all documents with a field value matching a set value in an ordinal bitset.
+ *
+ * Implementation is similar to {@link org.apache.lucene.search.join.TermsCollector}, but uses top-level ordinals
+ * explicitly and has wider visibility.
+ */
+public class TopLevelDVTermsCollector extends DelegatingCollector {
```
dsmiley: This class seems unused; no?

solr/solr-ref-guide/src/other-parsers.adoc
```diff
@@ -591,36 +591,95 @@ The hash range query parser uses a special cache to improve the speedup of the q
...
-You can optionally use the `score` parameter to return scores of the subordinate query. The values to use for this parameter define the type of aggregation, which are `avg` (average), `max` (maximum), `min` (minimum) `total`, or `none`.
+`method`::
+An optional parameter used to determine which of several query implementations should be used by Solr.
+Options are restricted to: `index`, `dvWithScore`, and `topLevelDV`.
+If unspecified the default value is `index`, unless the `score` parameter is present which overrides it to `dvWithScore`.
+Each implementation has its own performance characteristics, and users are encouraged to experiment to determine which implementation is most performant for their use-case.
+Details and performance heuristics are given below.
++
+`index` the default `method` unless the `score` parameter is specified.
+Uses inverted index structures to process the request.
```
dsmiley: Maybe instead of "inverted index" (which might confuse people) simply say terms index structure?

solr/solr-ref-guide/src/other-parsers.adoc
```diff
@@ -591,36 +591,95 @@ The hash range query parser uses a special cache to improve the speedup of the q
...
+Performance scales linearly with the number of values matched in the "from" field.
+Consider this method when the "from" query matches few documents, when the "to" side returns a large number of documents, or when sporadic post-commit slowdowns cannot be tolerated (this a disadvantage of other methods that `index` avoids).
```
dsmiley: "this a" -> "this is a"

solr/solr-ref-guide/src/other-parsers.adoc
```diff
@@ -591,36 +591,95 @@ The hash range query parser uses a special cache to improve the speedup of the q
...
 
-.Score parameter and single value numerics
++
+`dvWithScore` returns an optional "score" statistic alongside result documents.
+Uses docValues structures if available, but falls back to uninverted values when necessary.
```
dsmiley: "uninverted values" -> "the field cache"
I believe "uninvert" is more of an internal word whereas publicly we say "field cache"?

solr/solr-ref-guide/src/other-parsers.adoc
```diff
@@ -591,36 +591,95 @@ The hash range query parser uses a special cache to improve the speedup of the q
...
+Uninversion slows down the initial requests following a commit and takes up additional space on the JVM heap, so docValues are recommended in most situations.
```
dsmiley: "Uninversion" -> or maybe "First access to the field cache" ?

solr/core/src/test/org/apache/solr/search/TestJoinQueryPerformance.java
```diff
@@ -0,0 +1,235 @@
...
+import org.apache.solr.client.solrj.SolrQuery;
+import org.apache.solr.client.solrj.SolrRequest;
+import org.apache.solr.client.solrj.impl.HttpSolrClient;
+import org.apache.solr.client.solrj.response.QueryResponse;
+import org.apache.solr.common.SolrInputDocument;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+
+public class TestJoinQueryPerformance extends SolrTestCaseJ4 {
```
mkhludnev: How long does it take to run? Souldn't it be @Slow?

solr/core/src/test/org/apache/solr/search/TestJoinQueryPerformance.java
```diff
@@ -0,0 +1,235 @@
...
+        final String standardJoin = joinQueryBase + "}" + fromQuery;
+        final String noScoreJoin = joinQueryBase + " score=none}" + fromQuery;
+        final String tpiJoin = joinQueryBase + " toplevel=true}" + fromQuery;
+
+        final QueryResponse standardJoinRsp = client.query(TO_COLLECTION_NAME, new SolrQuery("q", mainQuery, "fq", standardJoin), SolrRequest.METHOD.POST);
+        final QueryResponse tpiJoinRsp = client.query(TO_COLLECTION_NAME, new SolrQuery("q", mainQuery, "fq", tpiJoin), SolrRequest.METHOD.POST);
+        final QueryResponse noScoreJoinRsp = client.query(TO_COLLECTION_NAME, new SolrQuery("q", mainQuery, "fq", noScoreJoin), SolrRequest.METHOD.POST);
+        final long numFound = tpiJoinRsp.getResults().getNumFound();
+
+        System.out.println(i + "," + numFound + "," + standardJoinRsp.getQTime() + "," + noScoreJoinRsp.getQTime() + "," + tpiJoinRsp.getQTime());
```
mkhludnev: Printing to stdout is prosecuted.

solr/core/src/test/org/apache/solr/search/TestJoinQueryPerformance.java
```diff
@@ -0,0 +1,235 @@
...
+   *  - "from" collection matches (with "to" matches held constant)
+   *  - "to" collection matches (with "from" matches held constant)
+   *
+   * So I think I should index a finite number of docs
+   */
+
+  @BeforeClass
+  public static void setUpCluster() throws Exception {
+    loadDictionary();
+    //loadCollectionData(DType.USER);
```
mkhludnev: Shouldn't it load at least anything?

solr/core/src/test/org/apache/solr/search/TestJoinQueryPerformance.java
gerlowskija: Hey Mikhail, thanks for the review.  I have no intentions of committing this file.  I kept it in the PR up to this point so that people could reproduce/tweak the perf test I ran.  But I'm going to remove it prior to commit.

(In fact, I thought I already had removed it, but there was one last "cleanup" commit I'd neglected to push until right now.  You should see this file gone from the PR now.)

solr/core/src/java/org/apache/solr/search/join/MultiValueTermOrdinalCollector.java
gerlowskija: Because it saves me from reimplementing `getLeafCollector()` and some other methods.  If you're wondering why DelegatingCollector as opposed to SimpleCollector or other options, there's not a great answer - DelegatingCollector was needed in some earlier revision when things were postfilter based.  I've changed it to use SimpleCollector; hopefully that addresses your concern.

gerlowskija:
> In JoinQuery.rewrite I see it says explicitly "don't rewrite the subQuery" but why not? If it never gets rewritten then that's a bug.

That's a question for the original Join author maybe?  I'm not familiar enough with how rewrites work to speak to it.

solr/core/src/java/org/apache/solr/search/TopLevelJoinQuery.java
```diff
@@ -0,0 +1,218 @@
...
+import org.apache.lucene.util.BytesRef;
+import org.apache.lucene.util.LongBitSet;
+import org.apache.solr.common.SolrException;
+import org.apache.solr.schema.IndexSchema;
+import org.apache.solr.schema.SchemaField;
+import org.apache.solr.search.join.MultiValueTermOrdinalCollector;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class TopLevelJoinQuery extends JoinQuery {
```
dsmiley: Always add at least one sentence javadoc for a class

gerlowskija:
Ready for review again when you get the chance @dsmiley

solr/solr-ref-guide/src/other-parsers.adoc
```diff
@@ -591,36 +591,95 @@ The hash range query parser uses a special cache to improve the speedup of the q
...
+`method`::
+An optional parameter used to determine which of several query implementations should be used by Solr.
+Options are restricted to: `index`, `dvWithScore`, and `topLevelDV`.
+If unspecified the default value is `index`, unless the `score` parameter is present which overrides it to `dvWithScore`.
+Each implementation has its own performance characteristics, and users are encouraged to experiment to determine which implementation is most performant for their use-case.
+Details and performance heuristics are given below.
++
+`index` the default `method` unless the `score` parameter is specified.
+Uses the terms index structures to process the request.
+Performance scales linearly with the number of values matched in the "from" field.
```
mkhludnev: Here I would disagree. Old solr's join loops term enums. So, number of terms and posings in `from`  dominates over number of `from` side matches in the overall complexity. see the thread linked at SOLR-14230, and also something like this is explained in LW's blogpost.

solr/solr-ref-guide/src/other-parsers.adoc
```diff
@@ -591,36 +591,95 @@ The hash range query parser uses a special cache to improve the speedup of the q
...
+Consider this method when the "from" query matches few documents, when the "to" side returns a large number of documents, or when sporadic post-commit slowdowns cannot be tolerated (this is a disadvantage of other methods that `index` avoids).
 
-.Score parameter and single value numerics
++
+`dvWithScore` returns an optional "score" statistic alongside result documents.
+Uses docValues structures if available, but falls back to the field cache when necessary.
```
mkhludnev: That what I always think, but giving the mail linked to SOLR-14230 there might be something broken in _fallback_. I'll clarify.

dsmiley:
>  I'm not familiar enough with how rewrites work to speak to it.

I recommend you become familiar with that, as it's a step in overall query processing and you are here writing a new Query, so....    Just saying.

solr/solr-ref-guide/src/other-parsers.adoc
gerlowskija: Thanks for the clarification.  Let me know if you're ok with the new wording.

solr/solr-ref-guide/src/other-parsers.adoc
gerlowskija: I was surprised in that thread that the user got no results without docValues.  The comments and code itself on ScoreJoinQParser make it look like docValues aren't strictly required, but you're right - it's totally possible this hasn't been working.

I'll leave this wording as-is for now, and wait for clarification as you investigate SOLR-14230.  If you don't have time to address 14230 soon, I'd rather not leave this blocked, so I'll figure out some intermediate wording.

mkhludnev: Sure, @gerlowskija please go ahead with this. It's fine. Sadly, I hardly check 14230 any time soon.

gerlowskija:
Fair enough - I'll add it to my list of things to brush up on.  Always plenty to learn.

PR #1172, opened 2020-01-15
SOLR-14189 switch from String.trim() to StringUtils.isBlank()

andywebb1975:
# Description

The edismax and some other query parsers treat pure whitespace queries as empty queries, but they use Java's String.trim() method to normalise queries. That method only treats characters 0-32 as whitespace. Other whitespace characters exist - such as U+3000 IDEOGRAPHIC SPACE - which bypass the test and lead to 400 Bad Request responses.

# Solution

This PR switches from String.trim() to StringUtils.isBlank() which is aware of all whitespace characters.

# Tests

I've added a test which sends U+3000 IDEOGRAPHIC SPACE and checks that it behaves the same way as U+0020 SPACE.

# Checklist

- [ ] (not sure!) I have given Solr maintainers [access](https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork) to contribute to my PR branch. (optional but recommended)
- [ ] (n/a) I have added documentation for the [Ref Guide](https://github.com/apache/lucene-solr/tree/master/solr/solr-ref-guide) (for Solr changes only).


uschindler:
Why not use the much more efficient StringUtils.isWhitespace() for the cases where the result is just checked for null? Just stripping and throwing away the result looks inefficient.

andywebb1975:
That makes total sense, yes! I started out looking for something to replace String.trim() and stopped looking once I'd found StringUtils.stripToNull(). I'll update this to use StringUtils.isWhitespace() instead.

andywebb1975:
I've made that change - it behaves the same as when I was using stripToNull() on my local instance.

cpoerschke:
> ... String.trim() ... StringUtils.stripToNull() ... StringUtils.isWhitespace() ...

I wonder if `StringUtils.isBlank()` might be another option:

```
- if (val == null || StringUtils.isWhitespace(val) ) {
+ if (StringUtils.isBlank(val)) {
```

Not sure about `StringUtils.isWhitespace` vs. `StringUtils.isBlank` efficiency considerations though e.g. https://github.com/apache/lucene-solr/blob/releases/lucene-solr/8.4.1/solr/core/src/java/org/apache/solr/search/SwitchQParserPlugin.java#L166 uses it.

uschindler:
IsBlank() is behaving fine. If you look at its source code all is checked:
- null means blank
- empty means blank
- all remaining characters are tested on isWhitespace()

So this is the best replacement. We should also look for similar patterns at other places in Solr.

+1

andywebb1975:
hi Christine - thanks for looking at this, and for the tip! I've made the change to use `StringUtils.isBlank()`, which efficiency-wise looks just the same as `isWhitespace()` to me.

andywebb1975:
I previously did a quick grep looking for `.trim(` and found these three instances that were obvious ones to change. The edismax instance is biting us as we get occasional exceptions in production due to junk queries, so that was my priority. There are lots more that could change, though I suspect they're much less likely to receive other unusual whitespace chars than the query parsers.

PR #1179, opened 2020-01-17
LUCENE-9147: Move the stored fields index off-heap.

jpountz:
This replaces the index of stored fields and term vectors with two
`DirectMonotonic` arrays. `DirectMonotonicWriter` requires to know the number
of values to write up-front, so incoming doc IDs and file pointers are buffered
on disk using temporary files that never get fsynced, but have index headers
and footers to make sure any corruption in these files wouldn't propagate to the
index.

`DirectMonotonicReader` gets a specialized `binarySearch` implementation that
leverages the metadata in order to avoid going to the IndexInput as often as
possible. Actually in the common case, it would only go to a single
sub `DirectReader` which, combined with the size of blocks of 1k values, helps
bound the number of page faults to 2.


lucene/core/src/java/org/apache/lucene/util/packed/DirectMonotonicReader.java
```diff
@@ -101,20 +104,99 @@ public static LongValues getInstance(Meta meta, RandomAccessInput data) throws I
...
+  public long get(long index) {
+    final int block = (int) (index >>> blockShift);
+    final long blockIndex = index & ((1 << blockShift) - 1);
+    final long delta = readers[block].get(blockIndex);
+    return mins[block] + (long) (avgs[block] * blockIndex) + delta;
+  }
+
+  /** Get lower/upper bounds for the value at a given index without hitting the direct reader. */
+  private long[] getBounds(long index) {
+    final int block = (int) (index >>> blockShift);
```
mikemccand: Do we know this incoming `long index` is small enough not to overflow `int` after right shift?  Should we use `Math.toIntExact` instead to confirm?

jpountz:
We have this information for a few datasets at https://elasticsearch-benchmarks.elastic.co/index.html#tracks/geonames/nightly/default/30d. For instance 3.8MB for the geonames dataset or 6MB for the HTTP logs dataset. It's not much, but these datasets are not very large either (3.1GB and 18.6GB respectively). As you can see it's the main contributor to memory usage after the terms index (which we still load it on-heap for the  `_id` field for now) so if a user loads the terms index off-heap, it may well be that stored fields are the main heap user.

PR #1176, opened 2020-01-16
LUCENE-9143 Add error-prone checks to build, but disabled

madrob:
Fixed a bunch of the smaller ones, ignored a lot of the bigger ones.
Incremental progress is the name of the game here!

This is still a work in progress and should probably not be merged in
current state.

# Description

Please provide a short description of the changes you're making with this pull request.

# Solution

Please provide a short description of the approach taken to implement your solution.

# Tests

Please describe the tests you've developed or run to confirm this patch implements the feature or solves the problem.

# Checklist

Please review the following and check all that apply:

- [ ] I have reviewed the guidelines for [How to Contribute](https://wiki.apache.org/solr/HowToContribute) and my code conforms to the standards described there to the best of my ability.
- [ ] I have created a Jira issue and added the issue ID to my pull request title.
- [ ] I have given Solr maintainers [access](https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork) to contribute to my PR branch. (optional but recommended)
- [ ] I have developed this patch against the `master` branch.
- [ ] I have run `ant precommit` and the appropriate test suite.
- [ ] I have added tests for my changes.
- [ ] I have added documentation for the [Ref Guide](https://github.com/apache/lucene-solr/tree/master/solr/solr-ref-guide) (for Solr changes only).

PR #1181, opened 2020-01-17
LUCENE-9145 First pass addressing static analysis

madrob:
Fixed a bunch of the smaller warnings found by error-prone compiler
plugin, while ignoring a lot of the bigger ones.

This is just the warnings found by #1176 without the build changes


dweiss:
I'd run a full test suite and if it passes just commit it in. Most of these look like legitimate bug fixes!

PR #1183, opened 2020-01-18

ErickErickson:
# Description

The only file that counts here really is lucene/queryparser/build.gradle. All the rest are a result of running the regenerate task. 

There are a couple of dodgy bits, see //nocommit in the build.gradle file.

NOTES:
- this does not address any of the other regenerate tasks yet.
- I'm not going to untangle warnings until later.

# Tests

precommit and the full test suite (both gradle and ant) pass, so I think it's close for this part of the regenerate task


dweiss:
I'll take a look but not until next week.

ErickErickson:
I have a PR that includes removing warnings.

PR #1184, opened 2020-01-18
LUCENE-9142 Refactor IntSet operations for determinize

madrob:
Fix a bug where a frozen set could be symmetrically unequal to the
sorted set that created it because we compared the backing array
instead of only the active elements.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/util/automaton/FrozenIntSet.java
+++ b/lucene/core/src/java/org/apache/lucene/util/automaton/FrozenIntSet.java
@@ -1,0 +1,51 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.lucene.util.automaton;
+
+import java.util.Arrays;
+
+final class FrozenIntSet extends IntSet {
+  final int[] values;
+  final int hashCode;
+  final int state;
+
+  FrozenIntSet(int[] values, int hashCode, int state) {
+    this.values = values;
+    this.hashCode = hashCode;
+    this.state = state;
+  }
+
+  @Override
+  int[] asArray() {
+    return values;
+  }
+
+  @Override
+  int size() {
+    return values.length;
+  }
+
+  @Override
+  public int hashCode() {
+    return hashCode;
+  }
+
+  @Override
+  public String toString() {
+    return Arrays.toString(values);
+  }
+}
--- a/lucene/core/src/java/org/apache/lucene/util/automaton/FrozenIntSet.java
+++ b/lucene/core/src/java/org/apache/lucene/util/automaton/FrozenIntSet.java
@@ -30,7 +30,7 @@ final class FrozenIntSet extends IntSet {
   }
 
   @Override
-  int[] asArray() {
+  int[] getArray() {
     return values;
   }
 
--- a/lucene/core/src/java/org/apache/lucene/util/automaton/IntSet.java
+++ b/lucene/core/src/java/org/apache/lucene/util/automaton/IntSet.java
@@ -1,0 +1,46 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.lucene.util.automaton;
+
+import java.util.Arrays;
+
+abstract class IntSet {
+    /**
+     * Return an array representation of this int set's values. Values are valid for indices [0, {@link #size()}).
+     * If this is a mutable int set, then changes to the set are not guaranteed to be visible in this array.
+     * @return an array containing the values for this set, guaranteed to be at least {@link #size()} elements
+     */
+    abstract int[] asArray();
+
+    /**
+     * Guaranteed to be less than or equal to the length of the array returned by {@link #asArray()}.
+     * @return The number of values in this set.
+     */
+    abstract int size();
+
+    @Override
+    public abstract int hashCode();
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (!(o instanceof IntSet)) return false;
+        IntSet that = (IntSet) o;
+        return hashCode() == that.hashCode()
+            && Arrays.equals(asArray(), 0, size(), that.asArray(), 0, that.size());
+    }
+}
--- a/lucene/core/src/java/org/apache/lucene/util/automaton/IntSet.java
+++ b/lucene/core/src/java/org/apache/lucene/util/automaton/IntSet.java
@@ -24,10 +24,10 @@ abstract class IntSet {
      * If this is a mutable int set, then changes to the set are not guaranteed to be visible in this array.
      * @return an array containing the values for this set, guaranteed to be at least {@link #size()} elements
      */
-    abstract int[] asArray();
+    abstract int[] getArray();
 
     /**
-     * Guaranteed to be less than or equal to the length of the array returned by {@link #asArray()}.
+     * Guaranteed to be less than or equal to the length of the array returned by {@link #getArray()}.
      * @return The number of values in this set.
      */
     abstract int size();
@@ -41,6 +41,6 @@ abstract class IntSet {
         if (!(o instanceof IntSet)) return false;
         IntSet that = (IntSet) o;
         return hashCode() == that.hashCode()
-            && Arrays.equals(asArray(), 0, size(), that.asArray(), 0, that.size());
+            && Arrays.equals(getArray(), 0, size(), that.getArray(), 0, that.size());
     }
 }
--- a/lucene/core/src/java/org/apache/lucene/util/automaton/SortedIntSet.java
+++ b/lucene/core/src/java/org/apache/lucene/util/automaton/SortedIntSet.java
@@ -156,7 +156,7 @@ final class SortedIntSet extends IntSet {
   }
 
   @Override
-  int[] asArray() {
+  int[] getArray() {
       return values;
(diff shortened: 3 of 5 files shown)


lucene/core/src/java/org/apache/lucene/util/automaton/FrozenIntSet.java
```diff
@@ -0,0 +1,38 @@
...
+  public FrozenIntSet(int[] values, int hashCode, int state) {
+    this.values = values;
+    this.hashCode = hashCode;
+    this.state = state;
+  }
+
+  public FrozenIntSet(int num, int state) {
+    this.values = new int[] { num };
+    this.state = state;
+    this.hashCode = 683 + num;
```
dweiss: This is a really bad hash, to be honest.... What is this constructor even used for? It's really sneaky as it allows changes to the content of a "frozen" set of values?

lucene/core/src/java/org/apache/lucene/util/automaton/IntSet.java
```diff
@@ -0,0 +1,63 @@
...
+ * limitations under the License.
+ */
+package org.apache.lucene.util.automaton;
+
+import java.util.Arrays;
+
+public abstract class IntSet {
+    int[] values;
+
+    int hashCode;
```
dweiss: I think the hashCode should be private to the abstract class and anything else should call hashCode() (including equals method). Then subclasses that are mutable could return true hashCode of their content (cached or computed).

lucene/core/src/java/org/apache/lucene/util/automaton/IntSet.java
```diff
@@ -0,0 +1,63 @@
...
+        IntSet intSet = (IntSet) other;
+        if (hashCode != intSet.hashCode) {
+            return false;
+        }
+        return Arrays.equals(values, 0, size(), intSet.values, 0, intSet.size());
+    }
+
+    @Override
+    public String toString() {
+        StringBuilder sb = new StringBuilder().append('[');
```
dweiss: If it's just for debugging then I'd just use a stream and join the values there, it's a one-liner. Space after for(

lucene/core/src/java/org/apache/lucene/util/automaton/SortedIntSet.java
```diff
@@ -151,126 +141,23 @@ public void computeHash() {
...
+   * Create a FrozenIntSet from the current values in this IntSet.
+   *
+   * Note: Must call computeHash() before calling this method
+   *
+   * @param state the state to save
+   * @return a FrozenIntSet that has the same values and hashCode as this set
+   */
   public FrozenIntSet freeze(int state) {
     final int[] c = new int[upto];
     System.arraycopy(values, 0, c, 0, upto);
```
dweiss: isn't there a futurearrays method that does this in a single step (without intermediate zeroed int[])?

lucene/core/src/java/org/apache/lucene/util/automaton/IntSet.java
uschindler: In addition, fields that are initialized in constructors and never changed should always be declared final.

lucene/core/src/java/org/apache/lucene/util/automaton/FrozenIntSet.java
madrob: It's a shortcut for the other constructor, uses the same calculation for hash that is in SortedIntSet.calculateHash, but specialized down to a single value. I'll see if we can easily drop this code since I agree that it adds complexity for maintaners.

lucene/core/src/java/org/apache/lucene/util/automaton/SortedIntSet.java
madrob: Arrays.copyOf does it in two steps, same as here. FutureArrays has compare, equals, and mismatch that I see?

dweiss: You look at Java code but the difference is that Array.copyOf is (I believe) an jvm intrinsic so it should be replaced with a more optimized code. The pair of allocation+arraycopy should as well be optimized for that matter but it requires execution graph analysis and copyOf is straightforward.

lucene/core/src/java/org/apache/lucene/util/automaton/FrozenIntSet.java
dweiss: oh, ok. drop it; the less code to understand the better. I don't think it'll be a particular gain here.

lucene/core/src/java/org/apache/lucene/util/automaton/SortedIntSet.java
madrob: Interestingly enough, forbidden APIs doesn't like Arrays.copyOf... I think we're fine here though.

lucene/core/src/java/org/apache/lucene/util/automaton/FrozenIntSet.java
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.lucene.util.automaton;
+
+public final class FrozenIntSet extends IntSet {
```
mikemccand: Hmm does it really need to be public?  I thought it's only the java sources in this package that use it?

lucene/core/src/java/org/apache/lucene/util/automaton/IntSet.java
```diff
@@ -0,0 +1,53 @@
...
+
+import org.apache.lucene.util.ArrayUtil;
+
+import java.util.Arrays;
+
+/**
+ * An int set backed by an array.
+ */
+public abstract class IntSet {
+    int[] values;
```
mikemccand: `protected`?

lucene/core/src/java/org/apache/lucene/util/automaton/SortedIntSet.java
```diff
@@ -77,6 +88,7 @@ public void incr(int num) {
         values[i] = num;
         counts[i] = 1;
         upto++;
+        stale = true;
```
mikemccand: How come the code didn't need/use `stale` before?

lucene/core/src/java/org/apache/lucene/util/automaton/SortedIntSet.java
```diff
@@ -16,59 +16,70 @@
...
-// BasicOperations.determinize
-final class SortedIntSet {
-  int[] values;
+// int[] count per state.  Used by Operations.determinize()
+final class SortedIntSet extends IntSet {
   int[] counts;
   int upto;
-  private int hashCode;
+
+  // Tracks if the hashCode computation is out of date and also if the array is out of sync with the map
```
mikemccand: Isn't the array discarded as soon as we switch over to the map (so how could it be out of sync)?

lucene/core/src/java/org/apache/lucene/util/automaton/SortedIntSet.java
```diff
@@ -149,128 +156,32 @@ public void computeHash() {
...
     }
+    stale = false;
   }
 
+  /**
+   * Create a FrozenIntSet from the current values in this IntSet.
+   *
+   * @return a FrozenIntSet that has the same values and hashCode as this set
+   */
+  @SuppressForbidden(reason = "Arrays.copyOf is a candidate for JVM intrinsic and we know our bounds are good")
```
mikemccand: Hmm why did we mark this API forbidden again?

lucene/core/src/test/org/apache/lucene/util/automaton/TestIntSet.java
```diff
@@ -0,0 +1,86 @@
...
+import org.junit.runner.RunWith;
+
+import static org.apache.lucene.util.LuceneTestCase.random;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotEquals;
+import static org.junit.Assert.assertTrue;
+
+@RunWith(com.carrotsearch.randomizedtesting.RandomizedRunner.class)
+public class TestIntSet {
+    @Test
```
mikemccand: We don't need `@Test` annotations -- `LuceneTestCase` runner knows `test*` methods are tests.

lucene/core/src/java/org/apache/lucene/util/automaton/SortedIntSet.java
```diff
@@ -16,59 +16,70 @@
...
+  boolean stale = true;
+
+  int hashCode;
 
   // If we hold more than this many states, we switch from
   // O(N^2) linear ops to O(N log(N)) TreeMap
-  private final static int TREE_MAP_CUTOVER = 30;
+  private final static int TREE_MAP_CUTOVER = 32;
 
   private final Map<Integer,Integer> map = new TreeMap<>();
```
mikemccand: Would be nice to use a native map from HPPC instead -- does HPPC have a sorted map?  Hmm, but this is core, and we don't have any dependencies in Lucene's core :)

lucene/core/src/java/org/apache/lucene/util/automaton/SortedIntSet.java
madrob: Because it didn't track whether the cached hash code was stale and relied on the caller to do it. I feel like it is more clear to manage it internally.

lucene/core/src/java/org/apache/lucene/util/automaton/SortedIntSet.java
madrob: The array is used for equality comparison.

lucene/core/src/java/org/apache/lucene/util/automaton/SortedIntSet.java
dweiss: Because copyOf was considered trappy -- the new length can be larger than original size of the input array (and is then null or zero padded)... I'm not sure it's worth having System.arraycopy instead... Maybe we could have a utility method that assert-verifies newLength <= length and uses Arrays.copyOf. Not sure.

lucene/core/src/java/org/apache/lucene/util/automaton/SortedIntSet.java
dweiss: oh, found the replacement we should use: ArrayUtil.copyOfSubArray

madrob:
Started over on this, tried to be less ambitious but still improving the situation I think.

Unclear why precommit failed according to GitHub, when I click through it looks like it succeeds?

@mikemccand @dweiss @uschindler can you take another look?

lucene/core/src/java/org/apache/lucene/util/automaton/FrozenIntSet.java
```diff
@@ -0,0 +1,51 @@
...
+  final int state;
+
+  FrozenIntSet(int[] values, int hashCode, int state) {
+    this.values = values;
+    this.hashCode = hashCode;
+    this.state = state;
+  }
+
+  @Override
+  int[] asArray() {
```
mikemccand: Could we rename to `getArray`?  `asArray` makes me think it might make a copy ...

lucene/core/src/test/org/apache/lucene/util/automaton/TestIntSet.java
mikemccand: I think you can remove these `@Test` annotations?

lucene/core/src/java/org/apache/lucene/util/automaton/FrozenIntSet.java
madrob: In the case of SortedIntSet, it might make a copy, I wanted to leave it flexible for the underlying implementation on how to handles this. Renaming to getArray makes sense though.

madrob:
> Who are the heavy consumers of determinize?

Fuzzy queries using multi-byte code points. Spell check (I think again, with multi-byte points). Possibly some of the graph queries, but I don't completely understand the context of how those build up their automata.

PR #1186, opened 2020-01-19

ErickErickson:
The only file that counts here really is lucene/queryparser/build.gradle. All the rest are a result of running the regenerate task.

There are a couple of dodgy bits, see //nocommit in the build.gradle file.

NOTES:
    this does not address any of the other regenerate tasks yet.
    This also incorproates LUCENE-9151, I'll close that JIRA.

Tests

precommit and the full test suite (both gradle and ant) pass, so I think it's close for this part of the regenerate task


gradle/defaults-java.gradle
```diff
@@ -6,6 +6,8 @@ allprojects {
     targetCompatibility = "11"
 
     compileJava.options.encoding = "UTF-8"
+    compileJava.options.compilerArgs << '-Xlint:unchecked'
```
dweiss: I added linting options in a separate commit.

lucene/queryparser/build.gradle
```diff
@@ -7,3 +7,224 @@ dependencies {
...
+  @TaskAction
+  void javacc() {
+
+
+    String javaCCClasspath = project.project(":lucene:queryparser").configurations.javaCCDeps.asPath
+    String javaCCHome = javaCCClasspath.substring(0, javaCCClasspath.lastIndexOf("/"))
+
+    // This bit seems really awkward, but I didn't find a good way to either convince the ant task to accept a different
+    // name than javacc.jar...
+    // nocommit So I'm taking the javacc-5.0.jar file that's downloaded to Gradle's cache and renaming it.
```
dweiss: You shouldn't be doing anything with those files. I think it'd be better to use javacc directly rather than through ant - then you could just invoke the javacc with the classpath you declared in the dependency, that's it.

ErickErickson: OK,I'll try it that way.

gradle/defaults-java.gradle
ErickErickson: Yep, saw that and I'll update this to master and resolve.

ErickErickson:
Dawid:

Thanks, I’ll implement your comments.

Meanwhile I’m trying to get the regenerate task to _not_ kick in when executing things like “gw assemble” by using constructs like:

regenerate.onlyIf {
  gradle.startParameter.taskRequests.args.findAll { String s -> s == "regenerate" }.size() > 0
}

is there a better way? So far this works on an individual task in a gradle file. I’d like a way for it to apply to _all_ tasks in a particular gradle file, including dependency resolution…. Digging.

As for the other files that changed, almost all of the differences are results of things I did intentionally. 

- Some tasks themselves use a replaceregexp with quoted lines that I decided to “fix” indentation on. Unnecessary and distracting I agree.

- The SuppressWarnings were moved around so that once this runs, people don’t have to go back in and hand-edit the generated files to suppress them.

- This little jewel: “implemetation’s” is because javacc apparently generates the misspelling it was hand-corrected.

- As for changes like: 

private java.util.List<int[]> jj_expentries = new java.util.ArrayList<>(); 
changed to:
private java.util.List<int[]> jj_expentries = new java.util.ArrayList<int[]>();

removing the redundancies were hand-edits, see: LUCENE-5512. and I don’t see the harm in leaving them in. Hmmm. I’ll try doing a replaceregexp on them to avoid compiler warnings and thrashing on it.

- I can’t explain at all why invoking javacc in this environment decided to move some methods around in StandardSyntaxParser.java I certainly changed the jj file in order to put the SuppressWarnings in, so it makes sense that the java file was regenerated. But the methods are identical, just in different places so I’m not worrying about it.

- As for the cast changes, all I can say is that I get no warnings. Hmmm, let me add -Xlint:cast to the parameters and see if they come back and fix as part of the task if so.

Thanks!

dweiss:
Looking at the patch this task should not be triggered when you run ./gradlew assemble... You don't need to complicate it with onlyIf or any such directives - a task is only triggered if it's attached to something executed in the task graph.

ErickErickson:
Thanks for the tip. Maybe after banging my head against this wall for a couple of hours I finally understand what the doLast{} bit is about. I’d broken out some tasks in the gradle file to do things like clean regenerated files first but _not_ wrapped them in doLast{} blocks, so they were being executed at config time, and removing things that subsequently didn’t get rebuilt because there was no execution phase on that task.

Straightened out now.

You know, if I actually read the damn output when I added printlns into the tasks under the "> Configure project :lucene:core” section and figured it out faster. Siiigggghhhh..

Erick

lucene/queryparser/build.gradle
ErickErickson: Guess what? I found a plugin for Gradle that runs and resolves the javacc task, it'll be in the next PR.

dweiss: I know there are plugins for that. We should stick to the exact javacc version ant currently uses though (so that the output code is the same). By the way - the shuffled generated code may be a result of a HashMap somewhere in the javacc code. Sigh.

ErickErickson:
New PR .

PR #1187, opened 2020-01-20
LUCENE-9152: Improve line intersection detection for polygons

iverase:
This patch changes the logic so we consider the boundary if no points of the triangle are inside the polygon. If all points are inside, then the boundary is not consider.


nknize:
LGTM!

PR #1189, opened 2020-01-21
SOLR-14196: AdminUI login not working for JWTAuth when blockUnknown=false

janhoy:
See https://issues.apache.org/jira/browse/SOLR-14196


janhoy:
Closing, wrong branch

PR #1191, opened 2020-01-21
SOLR-14197 Reduce API of SolrResourceLoader

dsmiley:
https://issues.apache.org/jira/browse/SOLR-14197

### DIFF
--- a/solr/core/src/java/org/apache/solr/core/SolrCore.java
+++ b/solr/core/src/java/org/apache/solr/core/SolrCore.java
@@ -326,6 +326,11 @@ public final class SolrCore implements SolrInfoBean, Closeable {
     return schema;
   }
 
+  /** The core's instance directory. */
+  public Path getInstancePath() {
+    return getCoreDescriptor().getInstanceDir();
+  }
+
   /**
    * Sets the latest schema snapshot to be used by this core instance.
    * If the specified <code>replacementSchema</code> uses a {@link SimilarityFactory} which is
@@ -949,8 +954,7 @@ public final class SolrCore implements SolrInfoBean, Closeable {
       this.dataDir = initDataDir(dataDir, solrConfig, coreDescriptor);
       this.ulogDir = initUpdateLogDir(coreDescriptor);
 
-      log.info("[{}] Opening new SolrCore at [{}], dataDir=[{}]", logid, resourceLoader.getInstancePath(),
-          this.dataDir);
+      log.info("[{}] Opening new SolrCore at [{}], dataDir=[{}]", logid, getInstancePath(), this.dataDir);
 
       checkVersionFieldExistsInSchema(schema, coreDescriptor);
       setLatestSchema(schema);
@@ -1169,7 +1173,7 @@ public final class SolrCore implements SolrInfoBean, Closeable {
     parentContext.gauge(() -> name == null ? "(null)" : name, true, "coreName", Category.CORE.toString());
     parentContext.gauge(() -> startTime, true, "startTime", Category.CORE.toString());
     parentContext.gauge(() -> getOpenCount(), true, "refCount", Category.CORE.toString());
-    parentContext.gauge(() -> resourceLoader.getInstancePath().toString(), true, "instanceDir", Category.CORE.toString());
+    parentContext.gauge(() -> getInstancePath().toString(), true, "instanceDir", Category.CORE.toString());
     parentContext.gauge(() -> isClosed() ? "(closed)" : getIndexDir(), true, "indexDir", Category.CORE.toString());
     parentContext.gauge(() -> isClosed() ? 0 : getIndexSize(), true, "sizeInBytes", Category.INDEX.toString());
     parentContext.gauge(() -> isClosed() ? "(closed)" : NumberUtils.readableSize(getIndexSize()), true, "size", Category.INDEX.toString());
--- a/solr/core/src/java/org/apache/solr/core/SolrCore.java
+++ b/solr/core/src/java/org/apache/solr/core/SolrCore.java
@@ -960,6 +960,7 @@ public final class SolrCore implements SolrInfoBean, Closeable {
           this.dataDir);
 
       checkVersionFieldExistsInSchema(schema, coreDescriptor);
+      setLatestSchema(schema);
 
       // initialize core metrics
       initializeMetrics(solrMetricsContext, null);
@@ -969,8 +970,6 @@ public final class SolrCore implements SolrInfoBean, Closeable {
       // also register it here for back-compat
       solrFieldCacheBean.initializeMetrics(solrMetricsContext, "core");
       infoRegistry.put("fieldCache", solrFieldCacheBean);
-
-      initSchema(config, schema);
 
       this.maxWarmingSearchers = config.maxWarmingSearchers;
       this.slowQueryThresholdMillis = config.slowQueryThresholdMillis;
@@ -1152,20 +1151,6 @@ public final class SolrCore implements SolrInfoBean, Closeable {
     }
     infoRegistry.put("updateHandler", newUpdateHandler);
     return newUpdateHandler;
-  }
-
-  /**
-   * Initializes the "Latest Schema" for this SolrCore using either the provided <code>schema</code>
-   * if non-null, or a new instance build via the factory identified in the specified <code>config</code>
-   *
-   * @see IndexSchemaFactory
-   * @see #setLatestSchema
-   */
-  private void initSchema(SolrConfig config, IndexSchema schema) {
-    if (schema == null) {
-      schema = IndexSchemaFactory.buildIndexSchema(IndexSchema.DEFAULT_SCHEMA_FILE, config);
-    }
-    setLatestSchema(schema);
   }
 
   /**
--- a/solr/core/src/java/org/apache/solr/core/SolrCore.java
+++ b/solr/core/src/java/org/apache/solr/core/SolrCore.java
@@ -134,7 +134,6 @@ import org.apache.solr.rest.ManagedResourceStorage.StorageIO;
 import org.apache.solr.rest.RestManager;
 import org.apache.solr.schema.FieldType;
 import org.apache.solr.schema.IndexSchema;
-import org.apache.solr.schema.IndexSchemaFactory;
 import org.apache.solr.schema.ManagedIndexSchema;
 import org.apache.solr.schema.SimilarityFactory;
 import org.apache.solr.search.QParserPlugin;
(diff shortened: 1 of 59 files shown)


noblepaul:
Lucene `ResourceLoader` is an interface, why can't we make `SolrResourceLoader` an interface. Having concrete classes as public APIs is not desirable. Once we make something an interface , it becomes much easier to plug in alternate impls or wrap it.

dsmiley:
An interface sounds fine but it will take some steps to get to that juncture, method-by-method reducing needless scope of the very use of SRL that needn't be there.  The PR is maybe 80% of the way there; close.

BTW when looking over this, be sure to look commit-by-commit and not all at once since each commit is usually fairly focused.  Extracting "properties" was a bit hard.  getInstancePath was too but that one isn't even done.

dsmiley:
I rebased off master since there were some upstream changes.  Also resolved some getInstancePath callers though didn't actually remove it yet.  I think it can move (not remain in SRL) to ZkSolrResourceLoader (until 9x, then remove) and a new StandaloneSolrResourceLoader.  I think a StandaloneSolrResourceLoader is the next step at this point.

madrob:
This looks pretty nice and was something I had been thinking about as well. Skimmed the first handful of commits and things made sense, I'll try to take a deeper look at this tomorrow or Wednesday!

dsmiley:
Perhaps the remaining larger changes relating to new classes (e.g. StandaloneSolrResourceLoader) should wait for a follow-on commit; there's plenty here already.  Maybe a few static methods could/should move elsewhere but this is ready for a review I think.

dsmiley:
I think this PR is ready for review @madrob since there are a lot of changes even without introducing SRL subclasses.

dsmiley:
FYI tomorrow I'll be on vacation for a week with minimal internet access, so I may not respond much for a bit.  Again, I think the issue is ready for review.

solr/core/src/java/org/apache/solr/core/SolrResourceLoader.java
```diff
@@ -201,24 +198,19 @@ void addToClassLoader(List<URL> urls) {
...
-  }
-  
   /**
    * Reloads all Lucene SPI implementations using the new classloader.
    * This method must be called after {@link #addToClassLoader(List)}
    * and before using this ResourceLoader.
    */
   void reloadLuceneSPI() {
+    // TODO improve to use a static Set<URL> to check when we need to
+    if (!needToReloadLuceneSPI) {
```
madrob: is this thread safe? worst case we end up reloading more often than necessary, which isn't a big deal, right?

solr/core/src/test/org/apache/solr/core/ResourceLoaderTest.java
```diff
@@ -169,12 +169,13 @@ public void testClassLoaderLibs() throws Exception {
...
     SolrResourceLoader loader = new SolrResourceLoader(tmpRoot);
+    loader.addToClassLoader(SolrResourceLoader.getURLs(lib));
 
-    // ./lib is accessible by default
+    // check "lib/aLibFile"
     assertNotNull(loader.getClassLoader().getResource("aLibFile"));
 
     // add inidividual jars from other paths
-    loader.addToClassLoader(otherLib.resolve("jar2.jar").toUri().toURL());
+    loader.addToClassLoader(Arrays.asList(otherLib.resolve("jar2.jar").toUri().toURL()));
```
madrob: nit: Collections.singletonList is a tiny bit more readable IMO

solr/core/src/java/org/apache/solr/core/SolrCore.java
```diff
@@ -909,15 +907,9 @@ public CoreContainer getCoreContainer() {
...
-   *
-   * @param dataDir the index directory
-   * @param config  a solr config instance
-   * @param schema  a solr schema instance
-   * @since solr 1.3
    */
-  public SolrCore(CoreContainer coreContainer, String name, String dataDir, SolrConfig config,
-                  IndexSchema schema, NamedList configSetProperties,
-                  CoreDescriptor coreDescriptor, UpdateHandler updateHandler,
+  private SolrCore(CoreContainer coreContainer, String name, ConfigSet configSet, CoreDescriptor coreDescriptor,
```
madrob: For this one and the other constructors that we changed, can we improve the javadoc to note to the caller that configSet will be used to create the resource loader? That might be too technical for a general audience, but it also seems like something that might cause a debugging headache down the line.

solr/core/src/java/org/apache/solr/core/SolrXmlConfig.java
```diff
@@ -128,19 +129,23 @@ public static NodeConfig fromFile(SolrResourceLoader loader, Path configFile, Pr
...
+  public static NodeConfig fromString(Path solrHome, String xml) {
+    return fromInputStream(
+        solrHome,
+        new ByteArrayInputStream(xml.getBytes(StandardCharsets.UTF_8)),
+        new Properties());
   }
 
-  public static NodeConfig fromInputStream(SolrResourceLoader loader, InputStream is, Properties substituteProps) {
+  public static NodeConfig fromInputStream(Path solrHome, InputStream is, Properties substituteProps) {
+    SolrResourceLoader loader = new SolrResourceLoader(solrHome);
```
madrob: Does this need to be closed?

solr/core/src/test/org/apache/solr/core/TestSolrXml.java
```diff
@@ -118,16 +110,16 @@ public void testPropertySub() throws IOException {
     System.setProperty("socketTimeout", "220");
     System.setProperty("connTimeout", "200");
 
-    File testSrcRoot = new File(SolrTestCaseJ4.TEST_HOME());
-    FileUtils.copyFile(new File(testSrcRoot, "solr-50-all.xml"), new File(solrHome, "solr.xml"));
+    Path testSrcRoot = Paths.get(SolrTestCaseJ4.TEST_HOME());
```
madrob: Should this be TEST_PATH like above?

solr/core/src/test/org/apache/solr/core/TestConfigSets.java
```diff
@@ -73,11 +73,11 @@ public void testConfigSetServiceFindsConfigSets() {
     CoreContainer container = null;
     try {
       container = setupContainer(TEST_PATH().resolve("configsets").toString());
-      Path testDirectory = container.getResourceLoader().getInstancePath();
+      Path solrHome = container.solrHome;
```
madrob: For consistency, prefer `getSolrHome()`

solr/core/src/java/org/apache/solr/core/SolrPaths.java
```diff
@@ -0,0 +1,130 @@
...
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Set;
+import java.util.concurrent.ConcurrentSkipListSet;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Utility methods about about paths in Solr.
```
madrob: repeated word: "about about"

solr/core/src/java/org/apache/solr/core/SolrPaths.java
```diff
@@ -0,0 +1,130 @@
...
+ */
+public class SolrPaths {
```
madrob: class can be final?

solr/core/src/java/org/apache/solr/core/SolrPaths.java
```diff
@@ -0,0 +1,130 @@
...
+   * <li>The system property solr.solr.home</li>
+   * <li>Look in the current working directory for a solr/ directory</li>
+   * </ol>
+   * <p>
+   * The return value is normalized.  Normalization essentially means it ends in a trailing slash.
+   *
+   * @return A normalized solrhome
+   * @see #normalizeDir(String)
+   */
+  public static Path locateSolrHome() {
```
madrob: I'm assuming this is a verbatim copy from the old code, so I'm not looking at it too closely. Let me know if that's not the case.

madrob:
Also, general comment that the scope of each commit was just about perfect and I really appreciated that as a reviewer.

solr/core/src/java/org/apache/solr/core/SolrResourceLoader.java
dsmiley: Hey good point; needToReloadLuceneSPI should be volatile and reloadLuceneSPI() should be synchronized.  In practice it'd only happen if someone had some strange Solr hacks / plugins because the only callers of the couple methods that touch this boolean do so once and then never touch it again.  But still, lets be thorough.

solr/core/src/java/org/apache/solr/core/SolrCore.java
dsmiley: I dunno... it's just one detail of a great many details that might be said.  And when you document something you run the risk of reality changing from what you documented and so need to keep in sync.  _Shrug._

solr/core/src/java/org/apache/solr/core/SolrXmlConfig.java
dsmiley: I looked into this a bit tonight; I didn't consider the close-ability of SRLs before.  Apparently it's because of the embedded URLClassLoader which _ought to_ be closed.  This was discussed when it was made closeable in https://issues.apache.org/jira/browse/SOLR-4791 and it seems to only affect Windows practically speaking.  The CoreContainer.shutdown method will close it's resourceLoader which is fetched from this SolrXmlConfig.  That's the particular instance you comment on here.  But I don't see that a Core's SRL will get closed.  Shrug; I dunno if I should try to address that in this PR.

solr/core/src/test/org/apache/solr/core/TestSolrXml.java
dsmiley: Indeed; a nice consistency improvement.

solr/core/src/java/org/apache/solr/core/SolrResourceLoader.java
```diff
@@ -104,7 +104,7 @@
   private RestManager.Registry managedResourceRegistry;
 
   /** @see #reloadLuceneSPI() */
-  private boolean needToReloadLuceneSPI = false;
+  private boolean needToReloadLuceneSPI = false; // requires synchronization
```
madrob: You mentioned in your reply that this needs to be `volatile`

dsmiley: Yeah I did... but then I found it difficult to reason about the sequence of when it's set to true/false... vs when the actual work is done and so I instead made both methods that touch it synchronized to be clearer/safer.

dsmiley:
@madrob can you approve this PR or suggest improvements?

What do we think about master only vs 8x?  I'm not sure honestly... I don't think the changes are drastic for plugin writers in 8x.  As @mkhludnev advised me on another issue recently, perhaps we immediately (for 8.5) mark these methods deprecated and then do the change for 8.6.

dsmiley:
I propose that in 8.x, I leave the methods that were moved but mark them as deprecated.  Otherwise, the rest of the changes are rather internal and can be applied to 8x, thus helping reduce possible merge conflicts for our future work.

I'll put an entry in Other Changes for 9.x: "SolrResourceLoader remove deprecated methods (numerous)."  Simple and succinct.  The commit message will have lots of info.  On 8x I'll say "SolrResourceLoader: marked many methods as deprecated, and in some cases rerouted exiting logic to avoid them".  Since this is very internal stuff, people can go see for themselves.

PR #1195, opened 2020-01-21
SOLR-13101: Log accurate file counts for Push and Pull in CorePushPull

ebehrendt:
# Description

Number of files and bytes pushed/pulled are logged when pushing to or pulling from blobstore, but values are only accurate when operation is successful. Log expected and actual file count, actual byte count, and add back boolean for whether operation was successful.

# Solution

Count number of files/bytes actually transferred

# Tests

Ran ant test, and working on adding additional unit tests.

Logline for successful push:
`PushPullData=[collectionName=gettingstarted shardName=shard1 sharedStoreName=gettingstarted_shard1 coreName=gettingstarted_shard1_replica_s1] action=PUSH storageProvider=LOCAL_FILE_SYSTEM bucketRegion=N/A bucketName=N/A runTime=44325264 startLatency=44325264 bytesTransferred=2076 attempt=0 expectedFilesAffected=10 actualFilesAffected=10 isSuccessful=true localGeneration=2 blobGeneration=-1 `

Logline for successful pull:
`PushPullData=[collectionName=pulltest2 shardName=shard2 sharedStoreName=pulltest2_shard2 coreName=pulltest2_shard2_replica_s6] action=PULL storageProvider=LOCAL_FILE_SYSTEM bucketRegion=N/A bucketName=N/A runTime=70088178 startLatency=70090032 bytesTransferred=9828 attempt=0 expectedFilesAffected=46 actualFilesAffected=46 isSuccessful=true localGeneration=1 blobGeneration=6`

Logline for exception thrown partway through push:
`PushPullData=[collectionName=gettingstarted shardName=shard1 sharedStoreName=gettingstarted_shard1 coreName=gettingstarted_shard1_replica_s1] action=PUSH storageProvider=LOCAL_FILE_SYSTEM bucketRegion=N/A bucketName=N/A runTime=6363713 startLatency=6363713 bytesTransferred=1132 attempt=0 expectedFilesAffected=10 actualFilesAffected=6 isSuccessful=false localGeneration=2 blobGeneration=-1`

Logline for exception thrown partway through pull:
`PushPullData=[collectionName=pulltest2 shardName=shard1 sharedStoreName=pulltest2_shard1 coreName=pulltest2_shard1_replica_s1] action=PULL storageProvider=LOCAL_FILE_SYSTEM bucketRegion=N/A bucketName=N/A runTime=9548098 startLatency=9548350 bytesTransferred=1404 attempt=0 expectedFilesAffected=19 actualFilesAffected=6 isSuccessful=false localGeneration=1 blobGeneration=3`

### DIFF
--- a/solr/core/src/java/org/apache/solr/store/blob/metadata/CorePushPull.java
+++ b/solr/core/src/java/org/apache/solr/store/blob/metadata/CorePushPull.java
@@ -22,7 +22,6 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.lang.invoke.MethodHandles;
-import java.nio.charset.StandardCharsets;
 import java.util.Collection;
 import java.util.HashSet;
 import java.util.Iterator;
@@ -39,6 +38,7 @@ import org.apache.lucene.store.IOContext;
 import org.apache.lucene.store.IndexInput;
 import org.apache.lucene.store.IndexOutput;
 import org.apache.solr.common.SolrException;
+import org.apache.solr.common.util.Pair;
 import org.apache.solr.core.CoreContainer;
 import org.apache.solr.core.DirectoryFactory;
 import org.apache.solr.core.SolrCore;
@@ -119,6 +119,10 @@ public class CorePushPull {
       if (solrCore == null) {
         throw new Exception("Can't find core " + pushPullData.getCoreName());
       }
+      
+      long actualFilesAffected = 0;
+      long bytesTransferred = 0;
+      boolean isSuccessful = false;
 
       try {
         // Creating the new BlobCoreMetadata as a modified clone of the existing one
@@ -159,6 +163,7 @@ public class CorePushPull {
         // point for the (potentially long) time it takes to push all files to the Blob store.
         Directory coreIndexDir = solrCore.getDirectoryFactory().get(solrCore.getIndexDir(), DirectoryFactory.DirContext.DEFAULT, solrCore.getSolrConfig().indexConfig.lockType);
         solrCore.getDeletionPolicy().saveCommitPoint(solrServerMetadata.getGeneration());
+        
         try {
           // Directory's javadoc says: "Java's i/o APIs not used directly, but rather all i/o is through this API"
           // But this is untrue/totally false/misleading. IndexFetcher has File all over.
@@ -168,6 +173,9 @@ public class CorePushPull {
 
             String blobPath = pushFileToBlobStore(coreStorageClient, coreIndexDir, cfd.getFileName(), cfd.getFileSize());
             bcmBuilder.addFile(new BlobCoreMetadata.BlobFile(cfd.getFileName(), blobPath, cfd.getFileSize(), cfd.getChecksum()));
+            
+            actualFilesAffected++;
+            bytesTransferred += cfd.getFileSize();
           }
         } finally {
           solrCore.getDeletionPolicy().releaseCommitPoint(solrServerMetadata.getGeneration());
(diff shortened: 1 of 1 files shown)


solr/core/src/java/org/apache/solr/store/blob/metadata/CorePushPull.java
```diff
@@ -419,45 +433,51 @@ protected String pushFileToBlobStore(CoreStorageClient blob, Directory dir, Stri
...
 
     /**
      * Downloads files from the Blob store 
      * @param destDir (temporary) directory into which files should be downloaded.
      * @param filesToDownload blob files to be downloaded
+     * @return number of files downloaded and number of bytes transferred successfully
      */
     @VisibleForTesting
-    protected void downloadFilesFromBlob(Directory destDir, Collection<? extends BlobFile> filesToDownload) throws Exception {
+    protected Pair<Long, Long> downloadFilesFromBlob(Directory destDir, Collection<? extends BlobFile> filesToDownload) throws Exception {
```
ebehrendt: I considered creating an inner wrapper class for the return values but it seemed like overkill, so I went with returning a pair

solr/core/src/java/org/apache/solr/store/blob/metadata/CorePushPull.java
```diff
@@ -419,45 +433,51 @@ protected String pushFileToBlobStore(CoreStorageClient blob, Directory dir, Stri
...
       for (BlobFile bf: filesToDownload) {
         log.info("About to create " + bf.getSolrFileName() + " for core " + pushPullData.getCoreName() +
             " from index on blob " + pushPullData.getSharedStoreName());
-        IndexOutput io = destDir.createOutput(bf.getSolrFileName(), DirectoryFactory.IOCONTEXT_NO_CACHE);
 
-        try (OutputStream outStream = new IndexOutputStream(io);
-          InputStream bis = coreStorageClient.pullStream(bf.getBlobName())) {
-          IOUtils.copy(bis, outStream);
-        }
+          try (IndexOutput io = destDir.createOutput(bf.getSolrFileName(), DirectoryFactory.IOCONTEXT_NO_CACHE);
```
ebehrendt: Moved IndexOutput initialization into the try with resources since it can throw an exception it also implements Closable

solr/core/src/java/org/apache/solr/store/blob/metadata/CorePushPull.java
```diff
@@ -419,45 +433,51 @@ protected String pushFileToBlobStore(CoreStorageClient blob, Directory dir, Stri
...
+              OutputStream outStream = new IndexOutputStream(io);
+            InputStream bis = coreStorageClient.pullStream(bf.getBlobName())) {
+            IOUtils.copy(bis, outStream);
+            
+            filesDownloaded++;
+            bytesTransferred += bf.getFileSize();
+          } catch (Exception e) {
+            return new Pair<>(filesDownloaded, bytesTransferred);
```
ebehrendt: Anyone have a recommendation on how to do this more gracefully than return the same values in and outside of the catch? I want to return these values whether exception is thrown or not, but cannot return from a finally.

solr/core/src/java/org/apache/solr/store/blob/metadata/CorePushPull.java
```diff
@@ -419,45 +433,51 @@ protected String pushFileToBlobStore(CoreStorageClient blob, Directory dir, Stri
 
     /**
      * Logs soblb line for push or pull action
```
andyvuong: This comment is also out of date (soblb doesn't exist) so we can remove it.

solr/core/src/java/org/apache/solr/store/blob/metadata/CorePushPull.java
andyvuong: This will swallow exceptions from requests failing with the blob store. I'd prefer to keep the same behavior and let those errors propagate up.

solr/core/src/java/org/apache/solr/store/blob/metadata/CorePushPull.java
andyvuong: Looked at the code and `OutputStream outStream = new IndexOutputStream(io)` was already in the try-with block and will end up closing the IndexOutput you moved. But the change has no affect if close if called twice so no change needed here, just wanted to note that.

yonik:
Thanks Emily!

PR #1202, opened 2020-01-22
SOLR-14149: CHANGES.txt Remove off-topic stuff

dsmiley:
https://issues.apache.org/jira/browse/SOLR-14149

### DIFF
--- a/dev-tools/scripts/addVersion.py
+++ b/dev-tools/scripts/addVersion.py
@@ -200,15 +200,7 @@ def get_solr_init_changes():
   return dedent('''
     Consult the LUCENE_CHANGES.txt file for additional, low level, changes in this release.
 
-    Versions of Major Components
-    ---------------------
-    Apache Tika %(org.apache.tika.version)s
-    Carrot2 %(/org.carrot2/carrot2-mini)s
-    Velocity %(/org.apache.velocity/velocity-engine-core)s and Velocity Tools %(org.apache.velocity.tools.version)s
-    Apache ZooKeeper %(/org.apache.zookeeper/zookeeper)s
-    Jetty %(org.eclipse.jetty.version)s
-
-    ''' % parse_properties_file('lucene/ivy-versions.properties'))
+    ''')
   
 def main():
   if not os.path.exists('lucene/version.properties'):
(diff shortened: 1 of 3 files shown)


dsmiley:
Yes 8x too.  Thanks for the reminder about 8.5.  I moved the 8.5 upgrade notes to the asciidoc page.  In 9x I left a reminder to move that once branch_9x is created since I think we need to maintain 9x upgrade notes on CHANGES.txt.  Or maybe we don't and this is a difference between both branches for the asciidoc file similarly to how CHANGES.txt is necessarily different?

What is the official link to master's live ref guide?  I usually use the build server for that: https://builds.apache.org/job/Solr-reference-guide-master/javadoc/   Speaking of which, it's strange that "javadoc" is in the URL there.

Running `ant package-src-tgz` in the solr dir runs the perl script that process CHANGES.txt into Changes.html.  It works on the CHANGES.txt that is committed to git, and thus doesn't see local edits, by the way.

janhoy:
> Or maybe we don't and this is a difference between both branches for the asciidoc file similarly to how CHANGES.txt is necessarily different?

I believe this would work. The master branch should contain what would go in a 9.0 release if we released today, so why not maintain the Upgrade Notes there from the start? Since it is easier to forget to touch two files, perhaps this should be a new checkbox in GH PR template "[ ] I have added an Upgrade Note to RefGuide if there are breaking changes".

dsmiley:
> this is a difference between both branches 

Good; we agree.  I'll update the PR.

> perhaps this should be a new checkbox in GH PR template

I disagree on that because I view the audience of that template to be our contributors, not us committers.  I delete it as step 1 of a PR.  Notice this comment at the very top of the template that I added:
_(If you are a project committer then you may remove some/all of the following template.)_

I think it's enough for the CHANGES.txt header to refer to solr-upgrade-notes (in some fashion) as that is an inherent reminder that there are other upgrade notes.  Besides, this is really for us committers to do and we will form new habits.

janhoy:
Agree on the PR template. An even more “live” page for master upgrade notes would be https://github.com/apache/lucene-solr/blob/master/solr/solr-ref-guide/src/solr-upgrade-notes.adoc

dsmiley:
LOL wow I'm ashamed for not knowing the ref guide looks pretty good served from GitHub.

PR #1205, opened 2020-01-23
SOLR-14206: Annotate HttpSolrCall as thread-safe

anshumg:
Annotate HttpSolrCall and V2HttpCall as thread-safe


madrob:
I think you already did this in #1203

anshumg:
@madrob  - this is a PR against 8x.

PR #1117, opened 2019-12-24
SOLR-13101: ant precommit fixes

megancarey:
# Description

This PR is just to fix precommit errors.

# Solution

Added licenses, updated code to remove forbidden method calls, and fixed other miscellaneous issues.

# Tests

N/A - I ran existing tests for the files changed to ensure no change in functionality.

# Checklist

Please review the following and check all that apply:

- [x] I have reviewed the guidelines for [How to Contribute](https://wiki.apache.org/solr/HowToContribute) and my code conforms to the standards described there to the best of my ability.
- [x] I have created a Jira issue and added the issue ID to my pull request title.
- [x] I have given Solr maintainers [access](https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork) to contribute to my PR branch. (optional but recommended)
- [ ] I have developed this patch against the `master` branch.
- [x] I have run `ant precommit` and the appropriate test suite.
- [ ] I have added tests for my changes.
- [ ] I have added documentation for the [Ref Guide](https://github.com/apache/lucene-solr/tree/master/solr/solr-ref-guide) (for Solr changes only).

PR #1208, opened 2020-01-24
SOLR-13101: Convert nanotime to ms

ebehrendt:
# Description

Continue logging ms, but use System.nanoTime instead of System.currentTimeMillis for more precise measurement of elapsed time. Specifically updating references from this previous PR: https://github.com/apache/lucene-solr/pull/1117/

# Solution

Divide nanoseconds by 1000000 to get milliseconds. Converting nanoTime to milliseconds is more precise than using currentTimeMillis because it does not use the system clock time which could face disruptions

# Tests

Ran ant test and manually examined loglines

### DIFF
--- a/solr/core/src/java/org/apache/solr/store/blob/metadata/CorePushPull.java
+++ b/solr/core/src/java/org/apache/solr/store/blob/metadata/CorePushPull.java
@@ -229,7 +229,6 @@ public class CorePushPull {
      */
     public void pullUpdateFromBlob(long requestQueuedTimeMs, boolean waitForSearcher, int attempt) throws Exception {
         long startTimeMs = System.nanoTime() / 1000000;
-        boolean isSuccessful = false;
 
         try {
           SolrCore solrCore = container.getCore(pushPullData.getCoreName());
--- a/solr/core/src/java/org/apache/solr/store/blob/metadata/CorePushPull.java
+++ b/solr/core/src/java/org/apache/solr/store/blob/metadata/CorePushPull.java
@@ -114,7 +114,7 @@ public class CorePushPull {
      * @param newMetadataSuffix suffix of the new core.metadata file to be created as part of this push
      */
     public BlobCoreMetadata pushToBlobStore(String currentMetadataSuffix, String newMetadataSuffix) throws Exception {
-      long startTimeMs = System.nanoTime() / 1000000;
+      long startTimeMs = BlobStoreUtils.getCurrentNanoTimeInMs();
       SolrCore solrCore = container.getCore(pushPullData.getCoreName());
       if (solrCore == null) {
         throw new Exception("Can't find core " + pushPullData.getCoreName());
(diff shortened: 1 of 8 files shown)


ebehrendt:
FYI @andyvuong

solr/core/src/java/org/apache/solr/store/blob/metadata/CorePushPull.java
```diff
@@ -499,7 +500,7 @@ void enqueueForHardDelete(BlobCoreMetadataBuilder bcmBuilder) throws Exception {
     @VisibleForTesting
     protected boolean okForHardDelete(BlobCoreMetadata.BlobFileToDelete file) {
       // For now we only check how long ago the file was marked for delete.
-      return System.nanoTime() - file.getDeletedAt() >= deleteManager.getDeleteDelayMs();
+      return (System.nanoTime() / 1000000) - file.getDeletedAt() >= deleteManager.getDeleteDelayMs();
```
ebehrendt: From what I could find, deletedAt and deleteDelay is only set using System.nanoTime/1000000, so there should not be any mismatch from comparing times that used nanoTime and times that used currentTimeMillis

andyvuong: It might be a good idea to update the BlobFileToDelete class to make it explicit via the javadoc (actually it doesn't have any) that the getDeletedAt() returns the timestamp in milliseconds

solr/core/src/java/org/apache/solr/store/blob/metadata/CorePushPull.java
```diff
@@ -114,7 +114,7 @@ protected CoreContainer getContainerFromServerMetadata(ServerSideMetadata solrSe
      * @param newMetadataSuffix suffix of the new core.metadata file to be created as part of this push
      */
     public BlobCoreMetadata pushToBlobStore(String currentMetadataSuffix, String newMetadataSuffix) throws Exception {
-      long startTimeMs = System.nanoTime();
+      long startTimeMs = System.nanoTime() / 1000000;
```
andyvuong: I'd consider making System.nanoTime() / 1000000 a util method in BlobStoreUtils

solr/core/src/java/org/apache/solr/store/blob/util/BlobStoreUtils.java
```diff
@@ -75,6 +76,16 @@ public static String generateMetadataSuffix() {
     return UUID.randomUUID().toString();
   }
 
+  /**
+   * Returns current nano time in millisecond resolution for use in measuring elapsed time.
+   * 
+   * Note: Do not combine this value with currentTimeMillis - currentTimeMillis will return ms relative to epoch
+   * while this method returns ms relative to some arbitrary time
+   */
+  public static long getCurrentNanoTimeInMs() {
```
andyvuong: Why not just call it getCurrentTimeMs and let the javadoc make it clear it's not the same as System.currentTimeMillis(), the main reason being it was changed originally to fix ant precommit failures?

ebehrendt: Sure - I had named this way to avoid combining with currentTimeMillis but if the javadoc is good enough for that I'll rename

PR #1140, opened 2020-01-03
SOLR-14165 set SolrResponse's serialVersionUID explicitly

andywebb1975:
# Description

The serialVersionUID of SolrResponse changed in https://github.com/apache/lucene-solr/pull/929 making Solr nodes before/after that change mutually incompatible for some operations.

# Solution

This sets the serialVersionUID of SolrResponse explicitly to its previous value.

# Tests

I've run manual builds with/without this patch and seen that /solr/admin/collections?action=overseerstatus against stops giving the exception `java.io.InvalidClassException: org.apache.solr.client.solrj.SolrResponse; local class incompatible: stream classdesc serialVersionUID = -7931100103360242645, local class serialVersionUID = 2239939671435624715` for hybrid clusters when the patch is in place.

# Checklist

Please review the following and check all that apply:

- [x] I have reviewed the guidelines for [How to Contribute](https://wiki.apache.org/solr/HowToContribute) and my code conforms to the standards described there to the best of my ability.
- [x] I have created a Jira issue and added the issue ID to my pull request title.
- [x] I have given Solr maintainers [access](https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork) to contribute to my PR branch. (optional but recommended)
- [x] I have developed this patch against the `master` branch.
- [x] I have run `ant precommit` and the appropriate test suite.
- (n/a) I have added tests for my changes.
- (n/a) I have added documentation for the [Ref Guide](https://github.com/apache/lucene-solr/tree/master/solr/solr-ref-guide) (for Solr changes only).

PR #1210, opened 2020-01-26
SOLR-14219 force serialVersionUID of OverseerSolrResponse

andywebb1975:
# Description

When the useUnsafeOverseerResponse=true option introduced in SOLR-14095 is used, the serialized OverseerSolrResponse has a different serialVersionUID to earlier versions, making it backwards-incompatible.

# Solution

This PR forces the serialVersionUID of OverseerSolrResponse to be the same value as before.

# Tests

Tested in a prototype environment with a mixed 8.4.1 and master-branch Solr nodes.

# Checklist

### DIFF
--- a/solr/core/src/java/org/apache/solr/cloud/OverseerSolrResponse.java
+++ b/solr/core/src/java/org/apache/solr/cloud/OverseerSolrResponse.java
@@ -26,7 +26,9 @@ import java.io.IOException;
 import java.util.Objects;
 
 public class OverseerSolrResponse extends SolrResponse {
-  
+ 
+  private static final long serialVersionUID = 4721653044098960880L;
+
   NamedList<Object> responseList = null;
 
   private long elapsedTime;


solr/core/src/java/org/apache/solr/cloud/OverseerSolrResponse.java
```diff
@@ -26,7 +26,9 @@
 import java.util.Objects;
 
 public class OverseerSolrResponse extends SolrResponse {
-  
+ 
+  private static final long serialVersionUID = 4721653044098960880L;
```
tflobbe: My understanding is that this number can actually vary depending on the compiler, so setting it to a specific value like this (expecting it to be number the class had in previous versions) may not work for everyone. 
Since the changes done in SOLR-14095 are just addition of static methods, maybe the better solution is to just revert them from OverseerSolrResponse and put them in some other util class

andywebb1975: hi Tomas, that's my understanding too - and it's possible that people may be running builds of earlier Solr versions whose serialVersionUID for this class differ from 472165... . 

I've some (slightly shaky!) evidence in support of doing it this way: in https://github.com/apache/lucene-solr/pull/1140 for SOLR-14165 I saw that the same earlier UID for SolrResponse had been generated by several different stacks, but it's possible the stacks weren't sufficiently different. Also I can see about a dozen instances of serialVersionUID being set to explicit values (other than 1) in the Lucene/Solr codebase, presumably for reasons similar to the current one - though to be fair I've no way to tell if these have caused compatibility issues with custom builds in the past.

I do think that serialVersionUID should be set explicitly for all serializable classes, and that setting it to the value used in the earlier official release builds is the best value to use when it's being set retrospectively like this.

In the interests of finding other approaches to making the serialization vs javabin change backwards-compatible, I've just run a build that didn't set serialVersionUID explicitly but made useUnsafeSerialization and useUnsafeDeserialization private, to see if this would give me the same default UID as before. It didn't - I got a third value -550706... instead. So as far as I can see the only options here are to remove all the changes from OverseerSolrResponse and put them elsewhere as you describe above, or to set its serialVersionUID to a possibly-incompatible value as I've done in this PR.

hope this helps!
Andy

tflobbe: I agree, everything that uses java serialization should be setting a serialVersionUID, my concern is that it may be too late now. I think you discovered a bug with your test (thanks!), but I believe it's too late to add a serialVersionUID now because for some users it could mean exactly the same as if we'd had one before and now we are changing it.
Hopefully this won't be an issue once the serialization is in javabin and the Java serialization part is removed.

andywebb1975: hi Tomas,

I still think it'd be better to set the serialVersionUID. It's _possible_ (though I think unlikely*) that there are systems where the previous (computed) value is different to `472165...`, but the new computed value (with the change to the class) would be different anyway, so either way they'll see an incompatibility. On systems using the standard build, we can make the new class backwards-compatible by adding serialVersionUID. It's guaranteed to be incompatible for everyone if we don't.

Andy

\* My reading of https://docs.oracle.com/javase/7/docs/platform/serialization/spec/class.html is that the computed UID is independent of the Java version, and that it should be set to the previously-computed value in version 2+ of a class.

tflobbe: > It's possible (though I think unlikely*) that there are systems where the previous (computed) value is different to 472165..., but the new computed value (with the change to the class) would be different anyway

That's why I suggest that we revert the changes to the class

> On systems using the standard build, we can make the new class backwards-compatible by adding serialVersionUID. It's guaranteed to be incompatible for everyone if we don't.

But removing the changes in OverseerSolrResponse also fixes the compatibility issue, for all cases in theory. For versions higher than 8.5, the solution will be to move to Javabin. @andywebb1975, I proposed the other approach in https://github.com/apache/lucene-solr/pull/1227. mind taking a look?

tflobbe:
Superseded by https://github.com/apache/lucene-solr/pull/1227

PR #1212, opened 2020-01-27
LUCENE-9176: Handle the case when there is only one leaf node in TestEstimatePointCount

iverase:
see https://issues.apache.org/jira/browse/LUCENE-9176

### DIFF
--- a/lucene/core/src/test/org/apache/lucene/codecs/lucene60/TestLucene60PointsFormat.java
+++ b/lucene/core/src/test/org/apache/lucene/codecs/lucene60/TestLucene60PointsFormat.java
@@ -132,7 +132,7 @@ public class TestLucene60PointsFormat extends BasePointsFormatTestCase {
     PointValues points = lr.getPointValues("f");
 
     // If all points match, then the point count is numLeaves * maxPointsInLeafNode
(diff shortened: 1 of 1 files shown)


lucene/core/src/test/org/apache/lucene/codecs/lucene60/TestLucene60PointsFormat.java
```diff
@@ -132,7 +132,7 @@ public void testEstimatePointCount() throws IOException {
     PointValues points = lr.getPointValues("f");
 
     // If all points match, then the point count is numLeaves * maxPointsInLeafNode
-    final int numLeaves = (int) Math.ceil((double) points.size() / maxPointsInLeafNode);
+    final int numLeaves = (int) Math.max(Math.ceil((double) points.size() / maxPointsInLeafNode), 1);
```
jpountz: I don't think this line needs to change, we already get a number of leaves equal to 1 when there are fewer points than number of points per leaf?

iverase: Yes, you are right, I reverted that change.

PR #1213, opened 2020-01-27
LUCENE-8143: Remove SpanBoostQuery

romseygeek:
`SpanBoostQuery` is essentially a null op, as boosts on nested span queries are
ignored by the scorer, unless it is used to wrap a query at the top level.  In this 
case it can simply be replaced by a normal `BoostQuery`.  It is confusing to have
a query that adds essentially nothing, and only works under certain circumstances.

This commit removes `SpanBoostQuery` entirely.


romseygeek:
This was merged into lucene 9 by https://github.com/apache/lucene/pull/155

PR #1214, opened 2020-01-27
LUCENE-9074: Slice Allocation Control Plane For Concurrent Searches

atris:
This commit introduces a mechanism to control allocation of threads to slices planned for a query.
The default implementation uses the size of backlog queue of the executor to determine if a slice
should be allocated a new thread


atris:
@jpountz Raised this PR for the approach that we discussed. Please let me know your thoughts and comments

atris:
@jpountz Please let me know your thoughts on this. Happy to iterate

lucene/core/src/java/org/apache/lucene/search/QueueSizeBasedExecutionControlPlane.java
```diff
@@ -0,0 +1,88 @@
...
+    if (tasks == null) {
+      throw new IllegalArgumentException("Tasks is null");
+    }
+
+    if (executor == null) {
+      throw new IllegalArgumentException("Executor is null");
+    }
+
+    ThreadPoolExecutor threadPoolExecutor = null;
+    if ((executor instanceof ThreadPoolExecutor) == false) {
```
jpountz: This is a bit trappy, if you wrap a ThreadPoolExecutor using the Executor API, you'd get different behavior than if you provided the ThreadPoolExecutor directly? Maybe this class should take a ThreadPoolExecutor, and there'd be a different simpler implementation for regular implementations. We'd keep an IndexSearcher constructor that takes an executor and documentation would say that we'd use this more sophisticated approach if it implements ThreadPoolExecutor?

lucene/core/src/java/org/apache/lucene/search/QueueSizeBasedExecutionControlPlane.java
```diff
@@ -0,0 +1,88 @@
...
+    for (FutureTask task : tasks) {
+      boolean shouldExecuteOnCallerThread = false;
+
+      if (isThresholdCheckEnabled && threadPoolExecutor.getQueue().size() <
+          (NUMBER_OF_PROCESSORS * LIMITING_FACTOR)) {
+        shouldExecuteOnCallerThread = true;
+      }
+
+      try {
+        executor.execute(task);
```
jpountz: hmm shouldn't we only do this if `shouldExecuteOnCallerThread` is false?

lucene/core/src/java/org/apache/lucene/search/QueueSizeBasedExecutionControlPlane.java
```diff
@@ -0,0 +1,88 @@
...
+import java.util.Collection;
+import java.util.List;
+import java.util.concurrent.Executor;
+import java.util.concurrent.Future;
+import java.util.concurrent.FutureTask;
+import java.util.concurrent.RejectedExecutionException;
+import java.util.concurrent.ThreadPoolExecutor;
+
+/**
+ * Implementation of SliceExecutionControlPlane with queue backpressure based thread allocation
```
jpountz: can you give more details about what this class tries to achieve?

lucene/core/src/java/org/apache/lucene/search/QueueSizeBasedExecutionControlPlane.java
```diff
@@ -0,0 +1,88 @@
...
+      threadPoolExecutor = (ThreadPoolExecutor) executor;
+    }
+
+    List<Future<C>> futures = new ArrayList();
+
+    for (FutureTask task : tasks) {
+      boolean shouldExecuteOnCallerThread = false;
+
+      if (isThresholdCheckEnabled && threadPoolExecutor.getQueue().size() <
+          (NUMBER_OF_PROCESSORS * LIMITING_FACTOR)) {
```
jpountz: should it be `threadPoolExecutor#getMaxPoolSize()` instead of `NUMBER_OF_PROCESSORS`?

lucene/core/src/java/org/apache/lucene/search/SliceExecutionControlPlane.java
```diff
@@ -0,0 +1,38 @@
...
+import java.util.concurrent.Future;
+import java.util.concurrent.FutureTask;
+
+
+/**
+ * Execution control plane which is responsible
+ * for execution of slices based on the current status
+ * of the system and current system load
+ */
+public interface SliceExecutionControlPlane<C> {
```
jpountz: why do we need generics on the class, it should only be needed on the method?

lucene/core/src/java/org/apache/lucene/search/SliceExecutionControlPlane.java
```diff
@@ -0,0 +1,38 @@
...
+  /**
+   * Invoke all slices that are allocated for the query
+   */
+  List<Future<C>> invokeAll(Collection<FutureTask> tasks);
```
jpountz: maybe require a less invasive abstraction for tasks, e.g. Callable?

jpountz: I also wonder whether it should go one step further and return a `List<C>` instead of a list of futures.

lucene/core/src/java/org/apache/lucene/search/QueueSizeBasedExecutionControlPlane.java
```diff
@@ -0,0 +1,88 @@
...
+import java.util.concurrent.ThreadPoolExecutor;
+
+/**
+ * Implementation of SliceExecutionControlPlane with queue backpressure based thread allocation
+ */
+public class QueueSizeBasedExecutionControlPlane<C> implements SliceExecutionControlPlane<C> {
+  private static final double LIMITING_FACTOR = 1.5;
+  private static final int NUMBER_OF_PROCESSORS = Runtime.getRuntime().availableProcessors();
+
+  private Executor executor;
```
jpountz: let's make it final?

lucene/core/src/java/org/apache/lucene/search/QueueSizeBasedExecutionControlPlane.java
```diff
@@ -0,0 +1,88 @@
...
+      // No effect
+      isThresholdCheckEnabled = false;
+    } else {
+      threadPoolExecutor = (ThreadPoolExecutor) executor;
+    }
+
+    List<Future<C>> futures = new ArrayList();
+
+    for (FutureTask task : tasks) {
+      boolean shouldExecuteOnCallerThread = false;
```
jpountz: should we always execute on the caller thread for the last task like IndexSearcher used to?

lucene/core/src/java/org/apache/lucene/search/QueueSizeBasedExecutionControlPlane.java
atris: Agreed. Reverted the Executor changes and added the abstraction while updating the docs for IndexSearcher

atris:
@jpountz Updated, please see and let me know your thoughts

lucene/test-framework/src/java/org/apache/lucene/util/LuceneTestCase.java
```diff
@@ -113,38 +139,10 @@
...
-import com.carrotsearch.randomizedtesting.annotations.ThreadLeakScope;
-import com.carrotsearch.randomizedtesting.annotations.ThreadLeakZombies.Consequence;
-import com.carrotsearch.randomizedtesting.annotations.ThreadLeakZombies;
-import com.carrotsearch.randomizedtesting.annotations.TimeoutSuite;
-import com.carrotsearch.randomizedtesting.generators.RandomPicks;
-import com.carrotsearch.randomizedtesting.rules.NoClassHooksShadowingRule;
-import com.carrotsearch.randomizedtesting.rules.NoInstanceHooksOverridesRule;
-import com.carrotsearch.randomizedtesting.rules.StaticFieldsInvariantRule;
-
-import junit.framework.AssertionFailedError;
```
jpountz: let's undo changes to this class?

lucene/core/src/java/org/apache/lucene/search/DefaultExecutionControlPlane.java
```diff
@@ -0,0 +1,96 @@
...
+      }
+    }
+
+    if (!shouldExecuteOnCallerThread) {
+      futures.add(task);
+    } else {
+      try {
+        futures.add(CompletableFuture.completedFuture(task.get()));
+      } catch (Exception e) {
+        throw new RuntimeException(e.getMessage());
```
jpountz: please preserve the stack trace of the cause

lucene/core/src/java/org/apache/lucene/search/DefaultExecutionControlPlane.java
```diff
@@ -0,0 +1,96 @@
...
+        // Execute on caller thread
+        shouldExecuteOnCallerThread = true;
+      }
+    }
+
+    if (shouldExecuteOnCallerThread) {
+      try {
+        task.run();
+      } catch (Exception e) {
+        throw new RuntimeException(e.getMessage());
```
jpountz: please preserve the stack trace of the cause

lucene/core/src/java/org/apache/lucene/search/IndexSearcher.java
```diff
@@ -210,12 +216,27 @@ public IndexSearcher(IndexReader r, Executor executor) {
...
     leafContexts = context.leaves();
     this.leafSlices = executor == null ? null : slices(leafContexts);
   }
 
+  /**
+   * We do this elaborate dance as to have only one constructor with a nullable second parameter
+   * See the next constructor for more clarification
+   * Only for testing
+   */
+  IndexSearcher(IndexReaderContext context, Executor executor,
```
jpountz: this doesn't need an executor, does it?

lucene/core/src/java/org/apache/lucene/search/SliceExecutionControlPlane.java
```diff
@@ -0,0 +1,35 @@
...
+ * Execution control plane which is responsible
+ * for execution of slices based on the current status
+ * of the system and current system load
+ */
+public interface SliceExecutionControlPlane <C, T extends Runnable> {
+  /**
+   * Invoke all slices that are allocated for the query
+   */
+  C invokeAll(Collection<T> tasks);
+}
```
jpountz: the generics on this interface look over-engineered?

lucene/core/src/java/org/apache/lucene/search/IndexSearcher.java
atris: It is needed since executor is a final member of IndexSearcher hence needs to be initialized? We could potentially always set it to null since it anyways wont be used but that seemed a bit counter intuitive. WDYT?

lucene/core/src/java/org/apache/lucene/search/SliceExecutionControlPlane.java
atris: Not sure about that. What would you advice it look like?

lucene/test-framework/src/java/org/apache/lucene/util/LuceneTestCase.java
atris: Yeah, seems like the original strategy caused a lot of changes to files inadvertently. I have taken a clean slate approach with minimal code changes and opened a new PR which addresses your comments.

Sorry about that, but refactoring the original changes to this PR seemed riskier than raising a new PR with only the necessary changes.

The new PR is #1294 . Lets continue the discussion there?

atris:
Superseded by #1294

PR #1217, opened 2020-01-28
SOLR-14223 PublicKeyHandler consumes a lot of entropy during tests

madrob:
Use a non-blocking implementation of SecureRandom for generating RSA Keys.


solr/test-framework/src/java/org/apache/solr/util/NotSecurePsuedoRandom.java
```diff
@@ -0,0 +1,73 @@
...
+
+/**
+ * A mocked up instance of SecureRandom that just uses {@link Random} under the covers.
+ * This is to prevent blocking issues that arise in platform default
+ * SecureRandom instances due to too many instances / not enough random entropy.
+ * Tests do not need secure SSL.
+ */
+public class NotSecurePsuedoRandom extends SecureRandom {
+  public static final SecureRandom INSTANCE = new NotSecurePsuedoRandom();
+  private static final Random RAND = new Random(42);
```
dweiss: This must *not* be static or shared across test instances. A better solution would be to create this off an initial long seed and this seed should be taken from RandomizedContext.current..().random().nextLong().

madrob: Hmm, not sure how that would work. Not every test runs with RandomizedContext, and even for those that do according to the code "static test class initializers are not permitted to access random contexts"

madrob: If the concern is that the random seeds are not reproducible across test runs... nothing using SecureRandom right now is reproducible anyway by design, so this wouldn't be a regression.

rmuir:
Is the goal of this JIRA just to address the entropy usage during tests or also to address the expensive key generation too? Can we fix both by just pregenerating a couple of keys for the tests and put the .pem files into test-files folder?

solr/test-framework/src/java/org/apache/solr/util/NotSecurePsuedoRandom.java
dweiss: All tests should extend from LuceneTestCase. If they don't, it's likely a problem already (because this also dodges thread leak checks, etc.). As for static variables - don't make it a static instance (it wasn't before if I see the diff right). Static instances make the reasoning about tests a lot more difficult and order-dependent.

bq. "static test class initializers are not permitted to access random contexts"

Correct. Because this happens outside of the control of the test runner. Again - it makes reasoning about leaked threads (and anything else) more difficult.

madrob:
@rmuir that's a great idea, and seems pretty straightforward. I'll take a look at what it would take to get there.

madrob:
Current state is partial implementation of Robert's suggestion, the other half that we would need is to plumb the creation logic through to solr.xml maybe?

madrob:
Wired this up so that we can get the keys loaded from disk - this setup seems to work for tests in `core` but not in `solar` or `contrib` modules because they have different test sources? Should I copy the keys to each module, or is there a more elegant way to handle that?

rmuir:
Maybe rather than "file", an idea is it could be a "resource" (e.g. Class.getResource) in solr-test-framework src/resources, matching the base class that seems to be using them. It would essentially be in the jar file, for any subclasses anywhere.

PR #1218, opened 2020-01-28
LUCENE-9134: Javacc skeleton

ErickErickson:
Here's the build changes to get javacc to run, modeled on the jflex changes , many thanks for the model. Only two files changed here ;)

If the structure is OK, I'll fill in the "doLast" blocks with the cleanup code and maybe be able extract some common parts. NOTE: you can't even compile the result of running this because I wanted the changes to the build structure to be clear first so didn't include the cleanup tasks yet.

So if this structure is OK, should I merge it into master before or after the rest of the cleanup? My assumption is after. I want to try to get all the warnings etc. out of the generated code in the next phase to reduce the temptation for people to make hand-edits.


gradle/defaults-java.gradle
```diff
@@ -25,13 +25,13 @@ allprojects {
     tasks.withType(JavaCompile) {
       options.encoding = "UTF-8"
       options.compilerArgs += [
-        "-Xlint",
```
dweiss: Are these differences on EOLs? I think they've been normalized recently so it should be LF.

gradle/generation/javacc.gradle
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  @TaskAction
+  def generate() {
+    if (!javaccFile || !javaccFile.exists()) {
+      throw new RuntimeException("JavaCC input file does not exist: ${javaccFile}")
+    }
+    // Remove old files so we can regenerate them
+    def parentDir = javaccFile.parentFile
+    parentDir.eachFileMatch FILES, ~/.*\.java/, { file ->
```
dweiss: If these files get overwritten I don't think we should care about explicit deletions (?).

dweiss: Oh, what's FILES -- I don't know this construct.

gradle/generation/javacc.gradle
```diff
@@ -0,0 +1,102 @@
...
+    }
+    logger.lifecycle("Regenerating JavaCC:\n  from: ${javaccFile}\n    to: ${parentDir}")
+
+    project.javaexec {
+      classpath {
+        project.rootProject.configurations.javacc
+      }
+      main = "org.javacc.parser.Main"
+      args += "-OUTPUT_DIRECTORY=${parentDir}"
+      args += [javaccFile]
```
dweiss: no need for array around javaccFile?

gradle/generation/javacc.gradle
```diff
@@ -0,0 +1,102 @@
...
+configure(project(":lucene:queryparser")) {
+  task javaccParserClassic(type: JavaCCTask) {
+    description "Regenerate classic query parser from java CC.java"
+    group "generation"
+
+    javaccFile = file('src/java/org/apache/lucene/queryparser/classic/QueryParser.jj')
+    def parent = javaccFile.parentFile.toString() // I'll need this later.
+
+    doLast {
+      // There'll be a lot of cleanup in here to get precommits and builds to pass, but as long as we don't
```
dweiss: I think it'd be ideal to regenerate with ant first (to eliminate any overlays that have accumulated),  commit that, then regenerate with gradle. With any local patches applied the result should be identical -- that's how you'll know the process is the same as with ant (git diff should be empty)?

ErickErickson:
Didn't title it right.

gradle/defaults-java.gradle
ErickErickson: OK, I'll check. I don't even know how they got changed frankly, I'll revert

gradle/generation/javacc.gradle
ErickErickson: I copied it from some example and it worked.......

ErickErickson: Actually, they aren't overwritten. If they're not deleted you get messages during execution like: "Warning: TokenMgrError.java: File is obsolete.  Please rename or delete this file so that a new one can be generated for you."

gradle/generation/javacc.gradle
ErickErickson: I'll change.

gradle/generation/javacc.gradle
ErickErickson: That _should_ be the end product already, that's one of the reasons I spent so much time on the ant version and why all those files were changed when I committed. At least I _think_ I got them all. At least that's what I remember doing... That said I'll try not to go off in the weeds.

Now that I've got the structure right, I'll see if I can get this to happen. Shouldn't actually be that much.

Oh, and ignore PR 1219, I had a bad title for this PR and it didn't link. When I changed the title of this one it took a while to show up and I got impatient. 1219 and 1218 are identical.

Finally, many thanks for your coaching (well, ok, outright fixing things)!

dweiss: Thanks Erick. I'm exhausted today - will commit it in tomorrow though. Good one with the coaching... some would call it whining!

gradle/defaults-java.gradle
ErickErickson: Done.

gradle/generation/javacc.gradle
ErickErickson: Leaving this in. We could make it conditional on whether there's anything to do, but that's waaaay more work than the benefit warrants for the 5 seconds this takes every two years when we actually need to regenerate these.

gradle/generation/javacc.gradle
ErickErickson: Dawid already combined these two.

gradle/generation/javacc.gradle
ErickErickson: I'd call it saving my reputation.
Working on this now, no need to keep this comment open.

PR #1222, opened 2020-01-29
LUCENE-9414 EagerCheapMergePolicy

dsmiley:
https://issues.apache.org/jira/browse/LUCENE-8962
(contributed by Salesforce)

See early comment:
// Extends TieredMergePolicy for convenience but we could adjust to delegate or modify TieredMergePolicy


lucene/core/src/java/org/apache/lucene/index/EagerCheapMergePolicy.java
```diff
@@ -0,0 +1,202 @@
+/*
```
mikemccand: Can we move this to sandbox instead?

lucene/core/src/java/org/apache/lucene/index/EagerCheapMergePolicy.java
```diff
@@ -0,0 +1,202 @@
...
+  /**
+   * Only find a cheap merge when the number of existing merging segments is less than or equal to
+   * this number.  In all likelihood, those segments are being merged concurrently. By default this
+   * setting is 0, thus won't find cheap merges during concurrent merging. If a higher number is
+   * used (e.g. 10) then there is a likely chance {@link org.apache.lucene.index.ConcurrentMergeScheduler}
+   * will intentionally stall the current thread if it's thresholds have been reached.  This may be
+   * okay if there are multiple merge threads (thus less likely to reach such thresholds) or if it's
+   * deemed more important to do cheap merges at the expense of slowing index throughput further.
+   */
+  public EagerCheapMergePolicy setCheapLimitConcurrentMergeSegments(int cheapLimitConcurrentMergeSegments) {
```
mikemccand: I think this is somewhat dangerous .... when a large merge kicks off, and runs for minutes maybe, it means we would suppress these "cheap merges on refresh/commit" while it is running?

lucene/core/src/java/org/apache/lucene/index/EagerCheapMergePolicy.java
```diff
@@ -0,0 +1,202 @@
...
+      }
+      accumSize += segSize;
+    }
+    if (numSegmentsToMerge < cheapMinMergeAtOnce) {
+      return null; // too few segments to merge
+    }
+    return infosSortedAsc.subList(0, numSegmentsToMerge);
+  }
+
+  private static long size(SegmentCommitInfo infoPerCommit) {
```
mikemccand: Rename to `sizeInBytes`?

lucene/core/src/java/org/apache/lucene/index/EagerCheapMergePolicy.java
```diff
@@ -0,0 +1,202 @@
...
+        ", cheapMinMergeAtOnce=" + cheapMinMergeAtOnce +
+        ", cheapLimitConcurrentMergeSegments=" + cheapLimitConcurrentMergeSegments;
+  }
+
+  @Override
+  public MergeSpecification findMerges(MergeTrigger mergeTrigger, SegmentInfos segmentInfos, MergeContext mergeContext) throws IOException {
+    InfoStream infoStream = mergeContext.getInfoStream();
+    boolean verbose = infoStream.isEnabled(INFO_STREAM_CATEGORY);
+
+    if (mergeTrigger == MergeTrigger.FULL_FLUSH) { // i.e. commit
```
mikemccand: Also `refresh`?

lucene/core/src/java/org/apache/lucene/index/EagerCheapMergePolicy.java
dsmiley: Why do you think?  It'd be nice for users to get an implementation of findFullFlushMerges without requiring sandbox.

lucene/core/src/java/org/apache/lucene/index/EagerCheapMergePolicy.java
dsmiley: Yes but why is that dangerous?  The intent of this was to protect from excessive merging / i/o.

lucene/core/src/java/org/apache/lucene/index/EagerCheapMergePolicy.java
dsmiley: Intentionally this is named consistently with `MergePolicy.size()`.  Otherwise I'd prefer the unit as well.

lucene/core/src/java/org/apache/lucene/index/EagerCheapMergePolicy.java
dsmiley: I don't see MergeTrigger.REFRESH; maybe you meant something else?  Any way this is now obsolete given that I updated this PR to use `findFullFlushMerges`

lucene/core/src/java/org/apache/lucene/index/EagerCheapMergePolicy.java
mikemccand: Well, dangerous because it leads to merge starvation.

Those tiny/fast merges really need to run at all times.  If you stop them because a huge merge is running it can lead to an explosion of small segments in the index, GC problems, running out of file descriptors, etc.

lucene/core/src/java/org/apache/lucene/index/EagerCheapMergePolicy.java
mikemccand: I think this a rather dangerous merge policy, at least as I currently understand it, so I think baking for a while in `sandbox` so we can confirm for more general users this is indeed safe.

lucene/core/src/java/org/apache/lucene/index/EagerCheapMergePolicy.java
```diff
@@ -0,0 +1,204 @@
...
+  /**
+   * Limit the maximum ratio of size between a candidate segment (that is deemed cheap) and the sum
+   * of all smaller segments. Example: If it's 2.0, a 20KB segment will merge with a 10KB
+   * segment (or say two 5KB segments) but will not with a 9KB segment.
+   * The limit should be tuned if {@link #setCheapMinMergeAtOnce(int)} is changed. It should be more
+   * than 1/(min-1).
+   */
+  public EagerCheapMergePolicy setCheapMaxSizeRatio(double cheapMaxSizeRatio) {
+    if (cheapMaxSizeRatio <= 0.0) {
+      throw new IllegalArgumentException("must be > 0");
```
mikemccand, line 90: Include the actual (invalid) value that was passed in here, in the exception message?

lucene/core/src/java/org/apache/lucene/index/EagerCheapMergePolicy.java
```diff
@@ -0,0 +1,204 @@
...
+        ", cheapLimitConcurrentMergeSegments=" + cheapLimitConcurrentMergeSegments;
+  }
+
+  @Override
+  public MergeSpecification findFullFlushMerges(MergeTrigger mergeTrigger, SegmentInfos segmentInfos, MergeContext mergeContext) throws IOException {
+    InfoStream infoStream = mergeContext.getInfoStream();
+    boolean verbose = infoStream.isEnabled(INFO_STREAM_CATEGORY);
+
+    if (mergeContext.getMergingSegments().size() <= cheapLimitConcurrentMergeSegments) {
+      OneMerge cheapMerge = findCheapMerge(segmentInfos, mergeContext);
```
mikemccand, line 130: Hmm, why only one merge?  If enough concurrent threads are indexing, there could be many small segments written during `commit`.

lucene/core/src/java/org/apache/lucene/index/EagerCheapMergePolicy.java
mikemccand: Let's not propagate the bad naming?  Just because an existing method is poorly named doesn't mean we should name new methods poorly too :)  Maybe open an issue to deprecate/rename `MergePolicy.size`?

Also, why not just use `MergePolicy.size` instead of adding a new dangerous method?

lucene/core/src/java/org/apache/lucene/index/EagerCheapMergePolicy.java
```diff
@@ -0,0 +1,204 @@
...
+      accumSize += segSize;
+    }
+    if (numSegmentsToMerge < cheapMinMergeAtOnce) {
+      return null; // too few segments to merge
+    }
+    return infosSortedAsc.subList(0, numSegmentsToMerge);
+  }
+
+  private static long size(SegmentCommitInfo infoPerCommit) {
+    //note: We could discount for deleted docs (as TieredMergePolicy does) but lets keep this simple/cheap.
```
mikemccand, line 196: But it is important for a merge policy to favor merging segments with many deletions?  Or, are you thinking that on commit the newly flushed segments will not have many deletions typically so why bother?  I think it would be simpler / less risky to just use the existing `MergePolicy.size`?

lucene/core/src/test/org/apache/lucene/index/EagerCheapMergePolicyTest.java
```diff
@@ -0,0 +1,86 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.List;
+
+import org.apache.lucene.util.LuceneTestCase;
+import org.apache.lucene.util.StringHelper;
+
+public class EagerCheapMergePolicyTest extends LuceneTestCase {
```
mikemccand, line 31: You could also insert it into `LuceneTestCase.newIndexWriterConfig` maybe?

lucene/core/src/java/org/apache/lucene/index/EagerCheapMergePolicy.java
```diff
@@ -0,0 +1,204 @@
...
+
+import org.apache.lucene.util.InfoStream;
+
+/**
+ * Eagerly proposes cheap merges on a commit. The goal is to help keep the segment count low where
+ * we can do so cheaply, particularly on a commit so that the IndexSearcher has fewer segments to
+ * search over.  This policy only overrides {@link #findFullFlushMerges(MergeTrigger, SegmentInfos, MergeContext)}; the
+ * rest are delegated to another policy.
+ */
+public class EagerCheapMergePolicy extends FilterMergePolicy {
```
mikemccand, line 35: Could we rename this to `MergeOnCommitMergePolicy` perhaps?

lucene/core/src/java/org/apache/lucene/index/EagerCheapMergePolicy.java
mikemccand: Actually, I was confused here -- since this merge policy is only picking (cheap) merges during commit, it is not as dangerous as I thought.  I had thought it was doing this in general (for natural and forced merges too).

dsmiley:
The `setCheapLimitConcurrentMergeSegments` limit was simply a precautionary addition given that I knew these cheap merges would up the write amplification and so I wanted to counter-act it in ways.  Basically, I wanted this MP to _not_ be "dangerous".

lucene/core/src/java/org/apache/lucene/index/EagerCheapMergePolicy.java
dsmiley: That's a nice vote of confidence I suppose because that's the generic name and it'd mean attempts from others to add a MP that merges on commit likewise would get the most obvious name.  For example "Highlighter" is forever people's go-to highlighter despite there being a better option.   If you like it so much, I do :-), maybe TieredMergePolicy might have this algorithm a setMaxCommitMergeWaitSeconds of 0 would mean it's never called any way and is thus "safe" to those who don't want it.  Just a thought and perhaps deserves JIRA level commentary instead of hidden in a GitHub PR code review.

PR #1223, opened 2020-01-30
SOLR-14213: Configuring Solr Cloud to use Shared Storage

andyvuong:
This is an initial PR that allows Solr Cloud to be configured to use shared storage or not. Prior to this, shared storage is "on" by default with components be lazily initialized as needed e.g. when a collection of type shared is created. This of course is difficult to do correctly. 

The solution proposed by this PR is check for the presence of a <sharedStorage> section in solr.xml that indicates the user's intention to use shared storage and initiate the components on startup. This also sets the stage for future work we need to do to move some hard coded values into configurable parameters that would be specified in sharedStorage. See the JIRA for further discussion. 

Changes:
- solr.xml supports a bare-bone <sharedStorage></sharedStorage> section that should be added to enable shared storage on the cluster
- If core discovery discovers a shared collection but the cluster is not shared storage enabled, this is considered a fatal error (user error) and solr will not start in stable state
- Collections of type shared can only be created if the cluster is shared storage enabled (all shared functionality is already gated behind the presence of shared replicas within shared collections)
- Added a test in SimpleSharedStorageCollectionTest to ensure proper feature gating
- Refactored our test setup for anything using SolrCloudSharedStoreTestCase to automatically start a mini solr cluster with a solr.xml (added solr-sharedstorage.xml) containing the sharedStorage section. Also made a change moving the local FS client setup into this parent class  so that any extending test classes don't need to worry about setting the shared directory up unless they want to mock the client themselves

### DIFF
--- a/solr/core/src/java/org/apache/solr/core/CoreContainer.java
+++ b/solr/core/src/java/org/apache/solr/core/CoreContainer.java
@@ -742,7 +742,7 @@ public class CoreContainer {
       if (cfg.getSharedStoreConfig() != null) {
         log.info("Shared storage is enabled in Solr Cloud. Initiating SharedStoreManager.");
         sharedStoreManager = new SharedStoreManager(getZkController());
-        sharedStoreManager.load();
+        sharedStoreManager.load(this);
       }
     }
 
@@ -1089,8 +1089,7 @@ public class CoreContainer {
         }
         
         if (sharedStoreManager != null) {
-          sharedStoreManager.getBlobProcessManager().shutdown();
-          sharedStoreManager.getBlobDeleteManager().shutdown();
+          sharedStoreManager.shutdown();
         }
       }
 
--- a/solr/core/src/java/org/apache/solr/core/CoreContainer.java
+++ b/solr/core/src/java/org/apache/solr/core/CoreContainer.java
@@ -741,8 +741,8 @@ public class CoreContainer {
       
       if (cfg.getSharedStoreConfig() != null) {
         log.info("Shared storage is enabled in Solr Cloud. Initiating SharedStoreManager.");
-        sharedStoreManager = new SharedStoreManager(getZkController());
-        sharedStoreManager.load(this);
+        sharedStoreManager = new SharedStoreManager(this);
+        sharedStoreManager.load();
       }
     }
 
--- a/solr/core/src/java/org/apache/solr/core/CoreContainer.java
+++ b/solr/core/src/java/org/apache/solr/core/CoreContainer.java
@@ -229,6 +229,8 @@ public class CoreContainer {
   protected volatile AutoscalingHistoryHandler autoscalingHistoryHandler;
   
   protected SharedStoreManager sharedStoreManager;
+  
+  private volatile boolean discoveredSharedCollection = false;
 
 
   // Bits for the state variable.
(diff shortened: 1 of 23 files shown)


solr/core/src/java/org/apache/solr/core/CoreContainer.java
```diff
@@ -756,6 +763,12 @@ public void load() {
         // is set to true for them
         // TODO: should this go behind some config?
         List<CoreDescriptor> additionalCoreDescriptors = discoverAdditionalCoreDescriptorsForSharedReplicas(cds);
+        if (!isSharedStoreEnabled() && discoveredSharedCollection) {
+          throw new SolrException(SolrException.ErrorCode.SERVER_ERROR,
```
andyvuong: This doesn't actually stop the process from starting up and currently not sure of a good way to abort the process if it's misconfigured. There used to be a setting by that was deprecated a long time ago.

solr/core/src/java/org/apache/solr/store/shared/SharedStoreManager.java
```diff
@@ -43,68 +43,38 @@
...
+    sharedCoreConcurrencyController = new SharedCoreConcurrencyController(sharedShardMetadataController);
   }
   
-  /*
-   * Initiates a SharedShardMetadataController if it doesn't exist and returns one 
+  /**
+   * Start blob processes that depend on an initiated SharedStoreManager
    */
+  public void load() {
+    blobCoreSyncer = new BlobCoreSyncer();
```
mbwaheed: It is not clear why BlobCoreSyncer is initialized here inside load method instead of the constructor. 
The comment says things here depend on initiated SharedStoreManager. That statement is incomplete. It is that plus the CoreContainer#sharedStoreManager is initialized because CorePullerFeeder depends on that. Also blobCoreSyncer and blobProcessUtil are two fields that are not initialized yet. The comment suggests all the public facing fields would have been initialized by now. Even if that is not what the comment means, there is nothing preventing them from being accessed in null state.

The other problem is that if the CoreContainer#load fails then CoreContainer@shutdown is called for cleanup. At least inside CoreContainer#createAndLoad. That can hit an NPE when calling sharedStoreManager.getBlobProcessManager().shutdown() because it is possible the exception that started the shutdown was in sharedStoreManager#load before blobProcessUtil was even initialized.

The second problem can be solved by creating a shutdown method inside SharedStoreManager and moving  sharedStoreManager.getBlobProcessManager().shutdown() there, protected by blobProcessUtil!= null check.
Though, I don't have any good ideas for the first problem.

solr/core/src/test/org/apache/solr/cloud/api/collections/SimpleSharedStorageCollectionTest.java
```diff
@@ -76,6 +62,28 @@ public void testCreateCollection() throws Exception {
...
     assertTrue(cloudClient.getZkStateReader().getZkClient().exists(ZkStateReader.COLLECTIONS_ZKNODE + "/" + collectionName, false));
   }
+  
+  /**
+   * Test that verifies that a collection creation command for a "shared" type collection fails
+   * if the cluster was not enabled with shared storage
+   */
+  @Test
+  public void testCreateCollectionSharedDisabled() throws Exception {
+    setupClusterSharedDisable(3);
```
mbwaheed: Don't need 3 node cluster here?

solr/core/src/java/org/apache/solr/store/shared/SharedStoreManager.java
andyvuong: For the first problem, there are shared storage components that have corecontainer injected explicitly in their api methods or via other dependencies that have getters opening access to it (zkcontroller for example). There are also shared storage components that have it injected in the constructor. Thinking about and looking at it again, it's kind of a mess identifying the initialization flows/orders and I might need to refactor a bunch of things here for better consistency

andyvuong: Looking at it again I don't think I needed to do any big refactor. Both blobCoreSyncer and blobProcessUtil are initialized in the constructor now but blobProcessUtil now starts CorePullFeeder in load(CoreContainer).

andyvuong:
cc @yonik can you merge when you get a chance. Thanks

PR #1224, opened 2020-01-30
LUCENE-9194: Simplify XYShapeQuery API

iverase:
This PR tries to simplify the XYShape query API. The trick for this simplification is to introduce a new abstract called XYGeometry.

In addition this patch makes package private most of the internal classes under geo package.


iverase:
@nknize let me know what you think.

PR #1225, opened 2020-01-30
SOLR-14233: JsonSchemaCreator should support Map payloads

noblepaul:


solr/solrj/src/java/org/apache/solr/common/util/JsonSchemaCreator.java
```diff
@@ -57,10 +57,18 @@
...
-      Type typ = ((ParameterizedType) t).getActualTypeArguments()[0];
-      map.put("type", "array");
-      map.put("items", getSchema(typ));
+    } else if (t instanceof ParameterizedType) {
+      if (((ParameterizedType) t).getRawType() == List.class) {
+        Type typ = ((ParameterizedType) t).getActualTypeArguments()[0];
+        map.put("type", "array");
+        map.put("items", getSchema(typ));
+      } else if (((ParameterizedType) t).getRawType() == Map.class) {
+        Type typ = ((ParameterizedType) t).getActualTypeArguments()[0];
```
sigram: This line can be removed - it's a no-op, I put it there because I thought I could use the type for something analogous to the List's "items".

noblepaul: yeah, it's useless if you use a Map with unknown types

PR #1226, opened 2020-01-30
LUCENE-9134 Port ant-regenerate tasks to Gradle build

ErickErickson:
I think this is the final version, I'll merge upstream over the weekend absent objections. The cleanup does three things:

1> makes it all compile
2> allows precommit to pass
3> fixes all the compiler warnings except deprecations. Which means that the "./gradlew assemble" produces clean output since the default lint settings don't report them except to say "there are some". This should reduce the frequency of people hand-editing these files.

A couple of notes:

1> the deprecations warning are mostly from FuzzyQuery, from changes a long time ago and I don't have the energy to tackle that, it should be a separate JIRA anyway.

2> cast and unchecked SuppressWarnings are no longer necessary, you'll note they are missing. Instead of suppressing warnings, the cleanup code fixes the source. We're using Javacc 5.0 and compiling against Java 11, so until we upgrade JavaCC they should stay I think.

3> all precommit and tests pass under Ant as well as Gradle.


gradle/defaults-java.gradle
```diff
@@ -25,13 +25,13 @@ allprojects {
...
       options.compilerArgs += [
-        "-Xlint", 
+        "-Xlint",
         "-Xlint:-deprecation",
         "-Xlint:-serial",
 
         "-Xdoclint:all/protected",
-        "-Xdoclint:-missing", 
-        "-Xdoclint:-accessibility", 
+        "-Xdoclint:-missing",
```
dweiss: extra white spaces.

ErickErickson: I wish I knew how this keeps happening, saw that after the PR. I commented out the -deprecation. Then uncommented. Then saved it. Didn't change he other lines at all AFAIK. I can fix it again, and apparently never touch that file again.

PR #1231, opened 2020-02-02
SOLR-13132: single sweep iteration over base, foreground, and backgro…

treygrainger:
Updating to master


treygrainger:
Ignore this one.

PR #1232, opened 2020-02-03
LUCENE-9171: Add BoostAttribute handling to QueryBuilder

romseygeek:
This commit allows QueryBuilder to add boosts to term queries and synonym queries 
when a boost is specified on a particular term via a BoostAttribute.


romseygeek:
This was merged as part of 663611c99c7d48dd31d53ea17644fcecd5e0fad7

PR #1233, opened 2020-02-03
LUCENE-9202: refactor leaf collectors in TopFieldCollector

msokolov:
Refactoring the two LeafCollector subclasses here into a shared `TopFieldLeafCollector` class to make the code easier on the eyes and amenable to further complications

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/TopFieldCollector.java
+++ b/lucene/core/src/java/org/apache/lucene/search/TopFieldCollector.java
@@ -69,6 +69,79 @@ public abstract class TopFieldCollector extends TopDocsCollector<Entry> {
     }
   }
 
+  private abstract class TopFieldLeafCollector extends MultiComparatorLeafCollector {
+
+    final boolean canEarlyTerminate;
+    boolean collectedAllCompetitiveHits = false;
+
+    TopFieldLeafCollector(FieldValueHitQueue<Entry> queue, Sort sort, LeafReaderContext context) throws IOException {
+      super(queue.getComparators(context), queue.getReverseMul());
+      final Sort indexSort = context.reader().getMetaData().getSort();
+      canEarlyTerminate = canEarlyTerminate(sort, indexSort);
+    }
+
+    void countHit(int doc) throws IOException {
+      ++totalHits;
+      hitsThresholdChecker.incrementHitCount();
+
+      if (minScoreAcc != null && (totalHits & minScoreAcc.modInterval) == 0) {
+        updateGlobalMinCompetitiveScore(scorer);
+      }
+    }
+
+    boolean thresholdCheck(int doc) throws IOException {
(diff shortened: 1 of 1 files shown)


lucene/core/src/java/org/apache/lucene/search/TopFieldCollector.java
```diff
@@ -555,7 +519,7 @@ public static void populateScores(ScoreDoc[] topDocs, IndexSearcher searcher, Qu
 
   final void add(int slot, int doc) {
     bottom = pq.add(new Entry(slot, docBase + doc));
-    queueFull = totalHits == numHits;
+    queueFull = slot == numHits - 1;
```
mikemccand: Hmm why was this needed?  Did we lose (easy) access to `totalHits`?

Can we `assert slot < numHits`?

lucene/core/src/java/org/apache/lucene/search/TopFieldCollector.java
```diff
@@ -69,6 +69,79 @@ public void setScorer(Scorable scorer) throws IOException {
     }
   }
 
+  private abstract class TopFieldLeafCollector extends MultiComparatorLeafCollector {
```
mikemccand: Should we maybe pull this out into its own `TopFieldLeafCollector.java` source?

msokolov: Well it has many reliances on `TopFieldCollector`'s members, and is never created on its own, so I think it's probably best as an inner class.

msokolov: I was mostly focused on the code duplicattion. However there are other opportunities to simplify - eg we could merge with `MultiComparatorLeafCollector`.

lucene/core/src/java/org/apache/lucene/search/TopFieldCollector.java
msokolov: It's because in `PagingFieldCollector` we care about the number of hits collected since the first page, rather than `totalHits`, and this abstracts away the difference, since `slot` is either `totalHits-1` or `hitsCollected-1`

jpountz: maybe leave a comment about it there?

lucene/core/src/java/org/apache/lucene/search/TopFieldCollector.java
mikemccand: OK let's do that in followon issue.  Progress not perfection!

lucene/core/src/java/org/apache/lucene/search/TopFieldCollector.java
msokolov: Thanks, I did

msokolov:
pushed via command line

PR #1235, opened 2020-02-03
LUCENE-8929: parallel early termination sharing counts and max scores across leaves

msokolov:
# Description
Briefly, introduces a new object for sharing scores and counts in `TopFieldCollector` in order to get earlier, but still accurate, termination. See the Jira for more dicsussion


mikemccand:
Hmm the strange looking empty file `lucene/core/.attach_pid19164` was added here.

lucene/core/src/java/org/apache/lucene/search/MaxScoreTerminator.java
```diff
@@ -0,0 +1,170 @@
...
+        // no longer contribute any top hits
+        return true;
+      }
+    }
+    return false;
+  }
+
+  private void updateWorstHit() {
+    //System.out.println("scoreboard updateWorstHit");
+    if (leafStates.size() > 1) {
```
mikemccand: Hmm, but if there is exactly one leaf we never set the worst `LeafState`?  Do we somehow bypass all this logic if the index has only one segment?

lucene/core/src/java/org/apache/lucene/search/MaxScoreTerminator.java
```diff
@@ -0,0 +1,170 @@
...
+      Collections.sort(leafStates);
```
mikemccand: A priority queue should be better here, in general?  It can insert and return the worst `LeafState` in `O(log(N))` time, amortized, whereas this sort is `O(N * log(N))` cost.  But maybe since the number of leaves is typically "smallish", it would not matter much in practice (the constant factors hidden in those big-Os would offset any gains)?

lucene/core/src/test/org/apache/lucene/search/TestTopFieldCollector.java
```diff
@@ -690,4 +697,94 @@ public void testRandomMinCompetitiveScore() throws Exception {
...
+        leafCollector1.collect(0);
+        leafCollector0.collect(0); // we have collected numHits=2 docs, but there are better ones remaining
+        expectThrows(CollectionTerminatedException.class, () -> leafCollector1.collect(1));
+        expectThrows(CollectionTerminatedException.class, () -> leafCollector0.collect(1)); // now we have all the top docs
+        expectThrows(CollectionTerminatedException.class, () -> leafCollector2.collect(0));
+      }
+    }
+  }
+
+  public void testRandomMaxScoreTermination() throws Exception {
```
mikemccand: Have you tried beasting this test?  If there are any bugs in the termination logic it seems like this test (given enough iterations) might uncover them ... `newIndexWriterConfig` and `RandomIndexWriter` are especially fun since they so heavily randomize how the docs are written to segments ...

lucene/core/src/java/org/apache/lucene/search/TopFieldCollector.java
```diff
@@ -310,14 +338,18 @@ public void collect(int doc) throws IOException {
 
   final int numHits;
   final HitsThresholdChecker hitsThresholdChecker;
-  final FieldComparator.RelevanceComparator firstComparator;
+  final FieldComparator.RelevanceComparator relevanceComparator;
   final boolean canSetMinScore;
+  final FieldComparator<?> firstComparator;
 
-  // an accumulator that maintains the maximum of the segment's minimum competitive scores
+  // an accumulator that maintains the maximum of the segments' minimum competitive scores
```
mikemccand: ;)

lucene/core/src/java/org/apache/lucene/search/TopFieldCollector.java
```diff
@@ -310,14 +338,18 @@ public void collect(int doc) throws IOException {
...
   final MaxScoreAccumulator minScoreAcc;
   // the current local minimum competitive score already propagated to the underlying scorer
   float minCompetitiveScore;
 
+  // Enables global early termination using minimum competitive scores and collected counts of all segments
```
mikemccand: Maybe also say "with concurrent search threads"?

lucene/core/src/java/org/apache/lucene/search/TopFieldCollector.java
```diff
@@ -69,6 +69,107 @@ public void setScorer(Scorable scorer) throws IOException {
     }
   }
 
+  private abstract class TopFieldLeafCollector extends MultiComparatorLeafCollector {
```
mikemccand: This source (`TopFieldCollector.java`) has really become too crazy ... we should at some point pull all these inner classes out to their own source files ... thank you for opening https://issues.apache.org/jira/browse/LUCENE-9202

lucene/core/src/java/org/apache/lucene/search/MaxScoreTerminator.java
```diff
@@ -0,0 +1,170 @@
...
+ * limitations under the License.
+ */
+package org.apache.lucene.search;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * MaxScoreTerminator is notified periodically by leaf collectors calling {@link #update}
```
mikemccand: Maybe state up front that this only applies in the static (index time) sort case, when the query's requested sort order is congruent (also sorting by that index time sort)?

And also (ideally?) applies to the concurrent case, where separate threads are searching each segment.

lucene/core/src/java/org/apache/lucene/search/MaxScoreTerminator.java
```diff
@@ -0,0 +1,170 @@
...
+
+    void update(float score) {
+      assert score >= this.worstScore;
+      this.worstScore = score;
+      ++this.resultCount;
+    }
+
+    @Override
+    public int compareTo(LeafState o) {
+      return Float.compare(worstScore, o.worstScore);
```
mikemccand: Hmm why `Float` and not `Double`?  What if the sort was a compound sort (sorting on multiple fields)?

lucene/core/src/java/org/apache/lucene/search/MaxScoreTerminator.java
```diff
@@ -0,0 +1,170 @@
...
+  synchronized LeafState add() {
+    LeafState newLeafState = new LeafState();
+    leafStates.add(newLeafState);
+    if (leafState == null) {
+      leafState = newLeafState;
+    }
+    return newLeafState;
+  }
+
+  // for testing
```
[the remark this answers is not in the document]
msokolov: Yeah, I'll add a javadoc explaining how to call this and what value to use

lucene/core/src/java/org/apache/lucene/search/MaxScoreTerminator.java
```diff
@@ -0,0 +1,170 @@
...
+ * exceed numHits by a small factor.  When the documents are evenly distributed among N segments,
+ * we expect to collect approximately (N+1/N) * numHits documents. In a worst case, where *all*
+ * the best documents are in a single segment, we expect to collect something O(log N) ie
+ * (1/N + 1/N-1 + ... + 1) * numHits documents, which is still much better than
+ * the N * numHits we would collect with a naive strategy.
+ * </p>
+ */
+class MaxScoreTerminator {
+
+  // we use 2^5-1 to check the remainder with a bitwise operation
```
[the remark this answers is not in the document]
msokolov: Yes, that's right. I'm changing to 5 (leaf collectors should call for every 32d collected hit), but this nominal value is never used in practice.

lucene/core/src/java/org/apache/lucene/search/MaxScoreTerminator.java
msokolov: We only terminate based on the first component of the score; we basically behave as if subsequent components were all equal. Hmm could be a lurking bug there. We might have to limit this optimization to Sort where this only a single field. Given that, is float sufficient? I guess we could be sorting on a double-valued field, so we should switch to double.

lucene/core/src/java/org/apache/lucene/search/MaxScoreTerminator.java
```diff
@@ -0,0 +1,170 @@
...
+  private final List<LeafState> leafStates;
+
+  /** The total number of docs to collect: from the Collector's numHits */
+  final int totalToCollect;
+
+  /** An upper bound on the number of docs "excluded" from max-score accounting due to early termination. */
+  private int numExcludedBound;
+
+  /** A lower bound on the total hits collected by all leaves */
+  int totalCollected;
```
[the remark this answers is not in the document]
msokolov: It's only ever written or read (except in a nonthreaded unit test) in `updateWorstHit`, called by `updateLeafState`, which is synchronized, so I think it's correct?

lucene/core/src/java/org/apache/lucene/search/MaxScoreTerminator.java
```diff
@@ -0,0 +1,170 @@
...
+    intervalMask = interval - 1;
+  }
+
+  /**
+   * Called by leaf collectors periodically to update their progress.
+   * @param newLeafState the leaf collector's current lowest score
+   * @return whether the collector should terminate
+   */
+  synchronized boolean update(LeafState newLeafState) {
+    totalCollected += interval;
```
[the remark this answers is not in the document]
msokolov: Yes it's true that calling threads must abide by the calling frequency, but the distinction between competitive and noncompetitive hits is moot since it will never be called in the non-competitive case: if the leaf knows its hit is noncompetitive, it is expected to terminate. I'll beef up the javadocs here

lucene/core/src/java/org/apache/lucene/search/MaxScoreTerminator.java
msokolov: IndexSearcher bypasses this `if (executor == null || leafSlices.length <= 1) {` . We update leafStates by removing them as we terminate the leaves though, so this will eventually drop to 1.

lucene/core/src/java/org/apache/lucene/search/MaxScoreTerminator.java
msokolov: Well, also we do not care what the state is in between the relatively infrequent events (I think?) when we cross the `totalToCollect` boundary, so we save the effort of many calls to `updateTop` or similar. In fact I think we can improve this here by only calling `updateWorstHit` when `totalCollected - numExcludedBound >= totalToCollect` rather than the current `totalCollected >= totalToCollect`

lucene/core/src/test/org/apache/lucene/search/TestTopFieldCollector.java
msokolov: Good idea - I'll try it! There is definitely the potential for a lurking bug here.

lucene/core/src/java/org/apache/lucene/search/TopFieldCollector.java
msokolov: hmm maybe we can do more there, but given how this is stacked on the refactoring PR in that issue I'd rather avoid to much more refactoring rn.

lucene/core/src/java/org/apache/lucene/search/MaxScoreTerminator.java
msokolov: I suppose we could re-insert all items into a PQ here instead of strictly sorting them though, yeah might be worth it

lucene/core/src/test/org/apache/lucene/search/TestTopFieldCollector.java
msokolov: OK I beasted at 1000 times, all good, but then I added extra SortField to the Sort and it hit issues, so I'll need to address.

lucene/core/src/java/org/apache/lucene/search/MaxScoreTerminator.java
msokolov: OK multiple fields is not a problem. After a long detour (where I learned how our test seeds work), I found and fixed a small bug when we use multiple sort keys with inverse ordering. Also, I am posting a new commit that switches to using double for scores.

msokolov:
This PR can no longer be rebased cleanl and I can't see what combination of git magic to use to preserve history here while rebasing; also, since it uses the all-caps label, it generates a lot of email noise. I'll open a new PR with the latest so this history can be preserved

PR #1236, opened 2020-02-04
Add back assertions removed by LUCENE-9187.

jpountz:
This time they would only apply to TestFastLZ4/TestHighLZ4 and avoid slowing
down all tests.


rmuir:
I think the randomize() stuff in {{src/java}} is confusing to follow and does not belong there.

maybe we should defer/split that and just get the assertion working again? It seems enough to add back the assertReset, but only call it from AssertingHashTable like you did here?

rmuir:
+1, thanks

PR #1237, opened 2020-02-04
SOLR-14238: Fix HdfsDirectory to no longer overwrite existing files.

jpountz:


jpountz:
@risdenk Correct. I'd also like to add support for `createTempOutput` in a follow-up change. The lack of support for it in `HdfsDirectory` means Solr tests don't pass on my other pull request https://github.com/apache/lucene-solr/pull/1179, and one way to make it work would be to generate temporary file names until creating a file with this name doesn't throw a `FileAlreadyExistsException`, like we do in `FSDirectory`.

jpountz:
Thanks @risdenk !

PR #1240, opened 2020-02-05
SOLR-14242: HdfsDirectory#createTempOutput.

jpountz:
JIRA: https://issues.apache.org/jira/browse/SOLR-14242


dsmiley:
Why create temporary data in HDFS when a local disk could be used, perhaps /tmp ?  Maybe because coding it this way was simply easy?

itygh:

PR #1241, opened 2020-02-05
Gradle util

ErickErickson:
This adds the generation targets for util/packed and util/automaton.

Interestingly, for whatever reason my local Python doesn't do anything weird like it did when regenerating the html entities, the generated code is identical.

One thing I'd like to draw attention to is that I had to change createLevAutomata.py to path to the new place moman is downloaded to.

I'll merge upstream sometime over the weekend probably barring objections.

I think this finishes off the regeneration work, so I'll close LUCENE-9134 after merging.


ErickErickson:
Didn't link appropriately,  I wondered why nobody replied.

PR #1243, opened 2020-02-07
LUCENE-9212: Intervals.multiterm() should take CompiledAutomaton

romseygeek:
Currently it takes `Automaton` and then compiles it internally, but we need to do things
like check for binary-vs-unicode status; it should just take `CompiledAutomaton` instead,
and put responsibility for determinization, binaryness, etc, on the caller.


romseygeek:
Merged as ffb7cafe9351cd6cd5181bc06dd053d586f6d63f

PR #1245, opened 2020-02-07
LUCENE-9146: Create gradle precommit action

anshumg:
This adds a gradle precommit action w/ Java11 for all branches.

### DIFF
--- a/.github/workflows/gradle-precommit.yml
+++ b/.github/workflows/gradle-precommit.yml
@@ -1,0 +1,23 @@
+name: Gradle Precommit
+
+on: 
+  pull_request:
+    branches:
+    - '*'
+
+jobs:
+  test:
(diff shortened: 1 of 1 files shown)


.github/workflows/gradle-precommit.yml
```diff
@@ -0,0 +1,23 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up JDK 11
+      uses: actions/setup-java@v1
+      with:
+        java-version: 11
+    - name: Grant execute permission for gradlew
+      run: chmod +x gradlew
```
dweiss: gradlew should have this permission already when you do a git clone? Why is it explicit?

PR #1247, opened 2020-02-10
SOLR-14252 use double rather than Double to avoid NPE

andywebb1975:
# Description

The getMax and getMin methods in AggregateMetric can throw an NPE if _only_ non-Number values are present when it tries to cast a null Double to a double.

# Solution

This PR switches to using primitive doubles, defaulting to zero, and logs when non-Number values are provided (we've seen `false` and `LocalStatsCache`).

# Tests

TBC

# Checklist


andywebb1975:
The PR really just changes an exception to a warning - it may be papering over another issue. I'm going to try changing `public Object value;` to `public Number value;` at line 41 in order to trigger earlier exceptions.

solr/core/src/java/org/apache/solr/metrics/AggregateMetric.java
```diff
@@ -93,16 +99,13 @@ public double getMax() {
     if (values.isEmpty()) {
       return 0;
     }
-    Double res = null;
+    double res = 0;
     for (Update u : values.values()) {
       if (!(u.value instanceof Number)) {
+        log.warn("not a Number: " + u.value);
```
andywebb1975: Note I'm not completely clear whether `u.value` is ever _expected_ to not be a `Number` - have seen this line report `false` and `LocalStatsCache` and I'm tracing back through to find out why these occur.

solr/core/src/java/org/apache/solr/metrics/AggregateMetric.java
```diff
@@ -51,6 +61,18 @@ public String toString() {
           ", updateCount=" + updateCount +
           '}';
     }
+
+    public double toDouble() {
```
tflobbe: Does this need to be `public`?

solr/core/src/java/org/apache/solr/metrics/AggregateMetric.java
```diff
@@ -93,18 +115,11 @@ public double getMax() {
     if (values.isEmpty()) {
       return 0;
     }
-    Double res = null;
+    double res = 0;
```
tflobbe: Is it guaranteed that values are going to be `>0`? how about a negatives?

solr/core/src/java/org/apache/solr/metrics/AggregateMetric.java
```diff
@@ -114,18 +129,11 @@ public double getMin() {
...
```
tflobbe: same here

solr/core/src/java/org/apache/solr/metrics/AggregateMetric.java
```diff
@@ -51,6 +61,18 @@ public String toString() {
           ", updateCount=" + updateCount +
           '}';
     }
+
+    public double toDouble() {
+      if (value instanceof Boolean) {
+        return 0;
+      } 
+      if (!(value instanceof Number)) {
+        log.debug("not a Number: " + value);
```
tflobbe: Can you use parametrized logging?

solr/core/src/java/org/apache/solr/metrics/AggregateMetric.java
andywebb1975: That's a good point - I don't know. (Who might know?)

It may be better to switch to creating a list of numerical values then finding its min/max/mean/etc (ideally using standard library functions) - but this may be overkill.

solr/core/src/java/org/apache/solr/metrics/AggregateMetric.java
andywebb1975: yes - if this logging is useful, I may just remove it instead.

solr/core/src/java/org/apache/solr/metrics/AggregateMetric.java
andywebb1975: I don't think so, no - and I think some/all of the the other `public` things in these classes could be made less accessible too.  Happy to take a look at this, but would like to make sure the metrics aggregation is working correctly first.

andywebb1975:
Note it's reporter config like the below that triggers the exceptions - the cache metrics include `LocalStatsCache` values:

```
<reporter name="test" group="shard" class="org.apache.solr.metrics.reporters.solr.SolrShardReporter">
      <int name="period">10</int>
      <str name="filter">UPDATE\./update/.*requests</str>
      <str name="filter">QUERY\./select.*requests</str>
      <str name="filter">CACHE\.searcher.*</str>
</reporter>
```

solr/core/src/java/org/apache/solr/metrics/AggregateMetric.java
tflobbe: Sure, but since this method is new lets make it private from start

solr/core/src/java/org/apache/solr/metrics/AggregateMetric.java
tflobbe: Maybe @sigram ?

solr/core/src/java/org/apache/solr/metrics/AggregateMetric.java
andywebb1975: Yup, made it private.

solr/core/src/java/org/apache/solr/metrics/AggregateMetric.java
andywebb1975: I've added a (temporary) warning when negative numbers are encountered - this immediately gave me this in the log, which indicates my assumption was incorrect:

```
o.a.s.m.AggregateMetric negative metric value: -1.0
```

So I think I need to go back to the drawing board with this. Will review the tests and to ensure the negative number situation is covered.

solr/core/src/test/org/apache/solr/util/stats/MetricUtilsTest.java
```diff
@@ -103,16 +108,29 @@ public void testMetrics() throws Exception {
...
         Map<String, Object> values = (Map<String, Object>)v.get("values");
         assertNotNull(values);
-        assertEquals(2, values.size());
+        assertEquals(4, values.size());
         Map<String, Object> update = (Map<String, Object>)values.get("foo");
         assertEquals(10, update.get("value"));
         assertEquals(1, update.get("updateCount"));
         update = (Map<String, Object>)values.get("bar");
         assertEquals(2, update.get("value"));
+        // assertEquals(1.5, update.get("mean"));
```
andywebb1975: This and the other temporarily commented-out tests report `null` instead of the expected value. I'll dig into this further, but if anyone can spot why this fails please let me know!

solr/core/src/test/org/apache/solr/util/stats/MetricUtilsTest.java
```diff
@@ -103,16 +108,29 @@ public void testMetrics() throws Exception {
...
         assertEquals(2, update.get("updateCount"));
+        update = (Map<String, Object>)values.get("baz");
+        assertEquals(false, update.get("value"));
+        // assertEquals(0, update.get("mean"));
+        // assertEquals(0, update.get("min"));
+        // assertEquals(0, update.get("max"));
+        assertEquals(1, update.get("updateCount"));        
+        update = (Map<String, Object>)values.get("neg");
+        assertEquals(-10, update.get("value"));
+        // assertEquals(-10, update.get("max"));
```
andywebb1975: I'm expecting this test (if it worked!) to fail at the moment, reporting 0 instead of -10.

solr/core/src/test/org/apache/solr/util/stats/MetricUtilsTest.java
andywebb1975: I figured it out!

solr/core/src/java/org/apache/solr/metrics/AggregateMetric.java
andywebb1975: I have a breaking test now - have figured out how MetricUtilsTest works.

andywebb1975:
I've opened a new clean PR: https://github.com/apache/lucene-solr/pull/1265 to replace this one, taking a different approach to avoiding the NPE.

PR #1248, opened 2020-02-10

ErickErickson:
This adds the generation targets for util/packed and util/automaton.

For whatever reason my local Python doesn't do anything weird like it did when regenerating the html entities, the generated code is identical.

One thing I'd like to draw attention to is that I had to change createLevAutomata.py to path to the new place moman is downloaded to.

I'll merge upstream in the next day or two barring objections.

I think this finishes off the regeneration work, so I'll close LUCENE-9134 after merging.


gradle/generation/util.gradle
```diff
@@ -0,0 +1,107 @@
...
+      ['gen_BulkOperation.py', 'gen_Packed64SingleBlock.py'].each { prog ->
+        logger.lifecycle("Executing: ${prog} in ${workDir}")
+        project.exec {
+          workingDir workDir
+          executable "python"
+          args = ['-B', "${prog}"]
+        }
+      }
+      // Correct line endings for Windows.
+      ['Packed64SingleBlock.java', 'BulkOperation*.java'].each { files ->
```
madrob: Does this need to be an `each` block, or can we specify multiple includes for the ant execution?

gradle/generation/util.gradle
```diff
@@ -0,0 +1,107 @@
...
+            includes: files,
+            encoding: 'UTF-8',
+            eol: 'lf'
+        )
+      }
+    }
+  }
+}
+
+configure(project(":lucene:core")) {
```
madrob: I think you can combine this with the previous configure block, I don't think separating them adds readability. Let me know if you did this intentionally though

lucene/core/src/java/org/apache/lucene/util/automaton/createLevAutomata.py
```diff
@@ -22,7 +22,7 @@
 import os
 import sys
 # sys.path.insert(0, 'moman/finenight/python')
-sys.path.insert(0, '../../../../../../../../build/core/moman/finenight/python')
+sys.path.insert(0, '../../../../../../../../../build/moman/finenight/python')
```
madrob: I think this has been answered before, but please remind me, does this break `ant regenerate` meaning that both cannot co-exist?

ErickErickson: Yeah. I'm pretty sure it does break the ant regenerate. Since the whole regenerate process is apparently run extremely rarely (like every couple of years or so from what I can tell), I think we'll be on gradle exclusively the next time this is run.

gradle/generation/util.gradle
ErickErickson: Good point, done.

gradle/generation/util.gradle
ErickErickson: True, I'll  change it.

ErickErickson:
I made the changes Mike mentioned, but I won't create another PR for a bit to give others a chance to look

gradle/generation/util.gradle
```diff
@@ -0,0 +1,102 @@
...
+ * limitations under the License.
+ */
+apply plugin: "de.undercouch.download"
+
+configure(rootProject) {
+  configurations {
+    utilgen
+  }
+
+  dependencies {
```
madrob: nit: drop this empty block?

lucene/core/src/java/org/apache/lucene/util/automaton/createLevAutomata.py
dweiss: If so then we should add a fail to ant and just say "use gradle?". These python scripts could take an argument - the path to moman passed from gradle script. Then it'd be elegant and clear without those ugly relative paths.

gradle/generation/util.gradle
ErickErickson: Yeah, I'll nuke that.

lucene/core/src/java/org/apache/lucene/util/automaton/createLevAutomata.py
ErickErickson: Dawid:

Yeah, I wondered about that. But please don't make me run the whole regenerate task in ant ;).

I'll change this to take an argument and get it working in Gradle, and leave a comment in the python code about having to change things a bit if running from Ant.

ErickErickson:
New PR momentarily.

PR #1249, opened 2020-02-11
LUCENE-9217: Add validation to XYGeometries

iverase:
This PR adds validation for XYGeometries, in particular checking for non-valid values like NaN, INF and -INF.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/geo/XYRectangle.java
+++ b/lucene/core/src/java/org/apache/lucene/geo/XYRectangle.java
@@ -16,6 +16,8 @@
  */
 package org.apache.lucene.geo;
 
+import static org.apache.lucene.geo.XYEncodingUtils.checkVal;
+
 /** Represents a x/y cartesian rectangle. */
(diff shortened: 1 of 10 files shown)


lucene/core/src/java/org/apache/lucene/geo/XYRectangle.java
```diff
@@ -29,12 +31,16 @@
 
   /** Constructs a bounding box by first validating the provided x and y coordinates */
   public XYRectangle(double minX, double maxX, double minY, double maxY) {
-    this.minX = minX;
-    this.maxX = maxX;
-    this.minY = minY;
-    this.maxY = maxY;
-    assert minX <= maxX;
-    assert minY <= maxY;
+    if (minX > maxX) {
```
iverase: I wonder if a XYRectangle should be initialise with floats instead of doubles like the other XYGeometries?

iverase:
I have opened #1252 that should supersede this one.

iverase:
fixed in #1252

PR #1250, opened 2020-02-11
SOLR-14250 Fix error logging on Expect: 100-continue

janhoy:
See https://issues.apache.org/jira/browse/SOLR-14250

With this PR, we'll still always try to consume the stream, 
but if the input stream is not available due to Expect: header,
and an error has already been sent by Solr, the resulting IOException 
from Jetty will not be logged at INFO level but instead be a simple
line on DEBUG level.

The PR does not try to add a test to validate behaviour
since this is a logging change only, so there is no risk
that the stream is not consumed anymore.


janhoy:
I tested manually that the new DEBUG log line is printed when hitting a non-existent URL, e.g.

    curl -H "Content-Type: application/json" -H "Expect: 100-continue" http://localhost:8983/solr/foo/update2

So I think this is good to go. Will leave it sitting here to collect feedback for a few days.

PR #1251, opened 2020-02-11
LUCENE-9134: Port ant-regenerate tasks to Gradle build (util and packed)

ErickErickson:
This should be the final patch. It turned out to be a 5 minute task to put the hard-coded path in the Ant build so I did that too.

I'll commit in a few and close this JIRA.


gradle/generation/util.gradle
```diff
@@ -0,0 +1,100 @@
...
+      }
+      project.ant.fixcrlf(
+          srcDir: workDir,
+          includes: '*ParametricDescription.java',
+          encoding: 'UTF-8',
+          eol: 'lf'
+      )
+    }
+  }
+}
```
ErickErickson: I'll put this lf in

PR #1252, opened 2020-02-12
LUCENE-9218: XYGeometries should expose values as floats

iverase:
Boxing the values to doubles happen when creating component2D objects.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/geo/XYEncodingUtils.java
+++ b/lucene/core/src/java/org/apache/lucene/geo/XYEncodingUtils.java
@@ -33,11 +33,12 @@ public final class XYEncodingUtils {
   private XYEncodingUtils() {
   }
 
-  /** validates value is within +/-{@link Float#MAX_VALUE} coordinate bounds */
-  public static void checkVal(double x) {
-    if (Double.isNaN(x) || x < MIN_VAL_INCL || x > MAX_VAL_INCL) {
+  /** validates value is a number and finite */
+  static float checkVal(float x) {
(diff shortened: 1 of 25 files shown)


lucene/core/src/java/org/apache/lucene/geo/XYEncodingUtils.java
```diff
@@ -33,11 +33,12 @@
   private XYEncodingUtils() {
   }
 
-  /** validates value is within +/-{@link Float#MAX_VALUE} coordinate bounds */
-  public static void checkVal(double x) {
-    if (Double.isNaN(x) || x < MIN_VAL_INCL || x > MAX_VAL_INCL) {
+  /** validates value is a number and finite */
+  static float checkVal(float x) {
+    if (Float.isNaN(x) || Float.isInfinite(x)) {
```
jpountz: This could be a single test.
```suggestion
    if (Float.isFinite(x) == false) {
```

lucene/core/src/java/org/apache/lucene/geo/XYEncodingUtils.java
```diff
@@ -68,7 +68,15 @@ public static double decode(int encoded) {
    * @param offset offset into {@code src} to decode from.
    * @return decoded value.
    */
-  public static double decode(byte[] src, int offset) {
+  public static float decode(byte[] src, int offset) {
     return decode(NumericUtils.sortableBytesToInt(src, offset));
   }
+
+  static double[] floatToDouble(float[] f) {
```
jpountz: nit: maybe add `Array` to the method name

PR #1255, opened 2020-02-13
SOLR-14216 Exclude Healthcheck from auth (do not merge)

janhoy:
This PR is just to demonstrate how we whitelist paths that should not be authenticated


epugh:
I have a build on github, and one of the things it checks is that after the docker-compose based enivronment is done running, that various services are up.  and of couse, now that it's wrapped in JWT, I get an error....     Thoughts on reopineing this?  I could use it, and it doesn't seem hacky at all...

PR #1256, opened 2020-02-13
SOLR-10306: Document in Reference Guide how to disable or reduce swapping

janhoy:
See https://issues.apache.org/jira/browse/SOLR-10306


solr/solr-ref-guide/src/taking-solr-to-production.adoc
```diff
@@ -302,6 +307,44 @@ Check these limits every time you upgrade your kernel or operating system. These
 If these limits are exceeded, the problems reported by Solr vary depending on the specific operation responsible for exceeding the limit. Errors such as "too many open files", "connection error", and "max processes exceeded" have been reported, as well as SolrCloud recovery failures.
 ====
 
+=== Avoid swapping
```
ctargett: We use "headline case" - capitalizing every major word - for section titles pretty much everywhere, so this should be changed to be consistent. One of my pre-publication steps is to look for all the places where this was missed and fixing them, so I'll fix it eventually if you don't have time to fix it now (for this and all the other section titles in this PR).

janhoy:
Done some changes. There is probably some sub-par language as well, don't be afraid to shoot me down :)

PR #1257, opened 2020-02-14
SOLR-14258: DocList should not extend DocSet

dsmiley:
https://issues.apache.org/jira/browse/SOLR-14258


dsmiley:
@mkhludnev Thanks for the approval on the first commit about SOLR-14258 (DocList shouldn't implement DocSet).  I forged ahead with SOLR-14256 (remove HashDocSet) on the same PR because I think in the end, one commit doing both is even more compelling in totality.

In short:  DocSet is now immutable and is now always in a natural document ID order.  There are now only exactly two implementations (newly enforced).  The guaranteed O(1) set detection is now handled with a getBits method, and this is more elegant for the callers that need it than the previous code idiom that had to do instanceof checks.  The back-compat concern of all changes here is pretty low, I think.

I'm less than 100% sure it's okay that SortedIntDocSet's getBits has a getLength that'll typically be less than maxDoc of the segment.  Tests pass, so... okay?  If we're not okay with this, we'll need the constructor to pass in maxDoc just as already occurs for FixBitSet used in BitDocSet.  WDYT @yonik 

I could imagine abandoning SortedIntDocSet in favor of only BitDocSet with modifications to be more general by supporting SparsedFixedBitSet (thus both BitSet implementations).  Or maybe practically speaking it'd need another class; I dunno.  Definitely not something I want to explore at this time though.

dsmiley:
BTW I'm about to step away on vacation for a week with spotty internet access and no computer.

PR #1259, opened 2020-02-14
LUCENE-9226: Return CELL_CROSSES_QUERY when point inside the triangle

iverase:
return the right relation when point inside a triangle.


iverase:
I push it as it is trivial

PR #1261, opened 2020-02-14
SOLR-14260: Make SchemaRegistryProvider pluggable in HttpClientUtil

athrog:
# Description
HttpClientUtil.java defines and uses an abstract SchemaRegistryProvider for mapping a protocol to an Apache ConnectionSocketFactory. There is only one implementation of this abstract class (outside of test cases). Currently, it is not override-able at runtime.

# Solution
Adds the ability to override the schema registry provider at runtime, using the class name value provided by "solr.schema.registry.provider", similar to how this class allows for choosing the HttpClientBuilderFactory at runtime.

# Tests
We have been using this patch in our internal fork of SOLR. We have verified that SOLR communication is encrypted, and since this patch helps us enable that encryption by setting a custom SSLContext for HTTP clients, we know this patch is working as expected. 

# Checklist

### DIFF
--- a/solr/core/src/java/org/apache/solr/update/UpdateShardHandler.java
+++ b/solr/core/src/java/org/apache/solr/update/UpdateShardHandler.java
@@ -95,9 +95,9 @@ public class UpdateShardHandler implements SolrInfoBean {
   private int connectionTimeout = HttpClientUtil.DEFAULT_CONNECT_TIMEOUT;
 
   public UpdateShardHandler(UpdateShardHandlerConfig cfg) {
-    updateOnlyConnectionManager = new InstrumentedPoolingHttpClientConnectionManager(HttpClientUtil.getSchemaRegisteryProvider().getSchemaRegistry());
-    recoveryOnlyConnectionManager = new InstrumentedPoolingHttpClientConnectionManager(HttpClientUtil.getSchemaRegisteryProvider().getSchemaRegistry());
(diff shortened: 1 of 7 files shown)


athrog:
FYI @dsmiley

athrog:
Hi David, thank you for the feedback and my apologies in the delayed response. I agree, 'schema' is an overloaded word in Solr that we could avoid. Your suggestion of `connectionSocketFactoryRegistry` is better, but there's still a Provider on top of that, and having a `connectionSocketFactoryRegistryProvider` sounds like too much Java word-salad to me. Maybe just drop the "connection" (since it's implied by a socket)? Or `connectionSocketFactoriesProvider`?

As to using a Supplier, we could use `Supplier<Registry<ConnectionSocketFactory>>`, but that would be a dodgy cast (at HttpClientUtil.java:161) due to the generics so I think I'll keep it a concrete class for now.

dsmiley:
We agree connectionSocketFactoryRegistryProvider is a bit much.  I had dropped "Provider" as it seems kinda meta and somewhat redundant with the word also having a "Factory" (Provider ~ Factory), but I can see it could be there.  I like your proposal of keeping either one of connection or socket.  I propose keep the word socket and drop connection.  Or abbreviate "connection" to "conn" which is sufficiently shorter to make me feel better about the fully complete word.  Ah naming...

I can see generics may be an issue so yeah keep the class.

athrog:
Fixed the obsolete comments, just did a grep for old instances and found a few more that needed changing

PR #1262, opened 2020-02-16
LUCENE-9220: regenerate all stemmers/stopwords/test data from snowball 2.0

rmuir:
Instead of patching them after-the-fact (both manually and
automatically over the years) we patch the generator.

This is easier to maintain than patches/changes against generated code.
See LUCENE-9220 for more information.

There is a remaining nocommit, test data. Also need to hook in and test
the new languages that are added here.


rmuir:
I've got gradle logic working to regenerate snowball. Its currently slow because the script is slow (vim-reformat takes forever). I will see if its enough to fix indentation with simple whitespace replacement. @dweiss will certainly hate it :), but the whole procedure was never automated before, had to start somewhere.

rmuir:
I think this is ready, I plan to push it later today. You have to run regenerate on linux: but that is limited by snowball makefile which will not work on windows. so if you want to do this stuff on windows it is best to start over there :)

gradle/generation/snowball.gradle
```diff
@@ -0,0 +1,118 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+apply plugin: "de.undercouch.download"
+
+configure(rootProject) {
+  ext {
+    snowballWorkDir    = file("${buildDir}/snowball")
+    snowballStemmerDir = file("${buildDir}/snowball/stemmers")
```
dweiss: the latter could be defined relative to snowballWorkDir (${snowballWorkDir}/stemmers)?

gradle/generation/snowball.gradle
```diff
@@ -0,0 +1,118 @@
...
+    cachedETagsFile file("${snowballWorkDir}/data.json")
+
+    doLast {
+      ant.unzip(src: dataZip, dest: snowballDataDir, overwrite: "true") {
+        ant.cutdirsmapper(dirs: "1")
+      }
+    }
+  }
+
+  task cleanSnowballCheckout(type: Delete) {
```
dweiss: The "cleanXXX" is a rule in gradle... don't know if it's not going to clash with something. Perhaps rename to "recompileSnowballCheckout" or something like this?

dweiss: Oh. Could also be a doFirst { project.delete snowballPatchedDir } on the downloadSnowballStemmers?

gradle/generation/snowball.gradle
```diff
@@ -0,0 +1,118 @@
...
+  task patchSnowball(type: Copy) {
+    dependsOn cleanSnowballCheckout
+
+    from fileTree(snowballStemmerDir) {
+      include '**/*'
+    }
+    into snowballPatchedDir
+
+    doLast {
+      ant.patch(patchfile: rootProject.file("gradle/generation/snowball.patch"), dir: snowballPatchedDir, strip: "1")
```
dweiss: ant.patch just redirects to local filesystem patch command (doesn't work on Windows). Sigh. Fine with me though.

gradle/generation/snowball.gradle
```diff
@@ -0,0 +1,118 @@
...
+    }
+  }
+}
+
+configure(project(":lucene:analysis:common")) {
+  task snowballGen() {
+    description "Patch and Regenerate snowball sources, stopwords, and tests"
+    group "generation"
+    dependsOn rootProject.patchSnowball
```
dweiss: Why not move all these tasks into lucene:analysis:common configuration? Are they reused anywhere else? If not then it'd make everything local to the project.

gradle/generation/snowball.sh
```diff
@@ -0,0 +1,122 @@
...
+test $# -eq 4
+
+trap 'echo "*** BUILD FAILED ***" $BASH_SOURCE:$LINENO: error: "$BASH_COMMAND" returned $?' ERR
+set -eEuo pipefail
+
+# reformats file indentation to kill the crazy space/tabs mix.
+# prevents early blindness !
+function reformat_java() {
+  # convert tabs to 8 spaces, then reduce indent from 4 space to 2 space
+  sed --in-place -e 's/\t/        /g' -e 's/    /  /g' $1
```
dweiss: Love the old school (sed). :)

lucene/analysis/common/build.xml
```diff
@@ -124,14 +122,4 @@
 
   <target name="regenerate" depends="jflex,unicode-data"/>
 
-  <target name="patch-snowball" description="Patches all snowball programs in '${snowball.programs.dir}' to make them work with MethodHandles">
```
dweiss: Thanks for cleaning these up from ant.

gradle/generation/snowball.gradle
rmuir: yeah mainly i'm after two things:
* don't download stuff if you have downloaded it before (uses ETag check). Some files are quite large.
* before patching, make a clean copy of checkout. This way we don't ever try to apply patch twice and don't force downloading again.

We can change to a doFirst I think. Lemme try.

gradle/generation/snowball.gradle
rmuir: Yes, duh. I'll fix!

gradle/generation/snowball.gradle
rmuir: I think if you have patch.exe on windows then this part will work?

gradle/generation/snowball.gradle
dweiss: At first hand the doFirst trick should be equivalent to the task dependencies the way you defined them. Let me know if this turns into a problem though.

A way to properly "download once" would be to define a configuration dependency with a custom resolver and let gradle cache the artifact... but it's too much of work for a small gain I think. What you did is already an improvement and we regenerate that stuff infrequently.

gradle/generation/snowball.gradle
rmuir: I can look into it. I am pretty clueless on gradle, I based this on the `moman` regeneration that is located in `gradle/generation/util.gradle`. Because it is doing similar stuff (downloading zip and running commands).

dweiss: Commit it in, I'll correct it afterwards, no worries.

gradle/generation/snowball.gradle
dweiss: Not many people do (I don't). You'd need to run under cygwin or have msys tools installed (which are recompiled for Windows). I don't think this is a problem. I don't think there is a single Windows developer who isn't fluent on other environments. If a regeneration task does not work on Windows so be it (I'd make it complain gracefully but this can come later).

dweiss: I looked into platform-independent patch command implementation but I didn't find any (jgit's is inherently broken; @kojisekig wrote something for jenkins but I didn't look into that - maybe he'll have some more insight).

gradle/generation/snowball.gradle
rmuir: ok I think this is now better. Now tasks are under analysis common, downloads stuff to `lucene/analysis/common/build` instead of `build/` and all tasks are under that project, except a single top-level one with the description that depends on it.

dweiss: Thanks Robert. FYI: the moman build could be modified the same way. There is a root project dependency section in javacc.gradle but the rationale behind it is that it's a configuration dependency then shared by different subprojects (and a task that requires javacc).

gradle/generation/snowball.gradle
rmuir: It worked fine as a doFirst. Here is the issue: it is nice that `gradlew snowball` is fast and not downloading 30MB files (!) every time you run it, if you are trying to iterate locally, for example to actually upgrade it.

So while it may be infrequent, I wanted to make it more pleasant for running this over and over until you get it right, because you will be dealing with plenty of other unpleasant stuff...

gradle/generation/snowball.gradle
rmuir: Maybe we can focus on trying to remove the patch completely (try to package up some PRs for snowball team) and it gets fixed even before their Makefile supports windows :)

I guess I optimistically hope the patch solution is a temporary one. If you look deeper into shell script and snowball GNUMakefile, it may be obvious that we have other issues to fix for windows first.

rmuir:
I am looking into trying to improve the situation to allow it to regenerate on the mac too.

Two initial problems found:
* need to cleanup sed usage to work with both GNU and BSD sed
* different test zip files are created because the random seed for sampling is not reproducible. This is because `openssl` is really LibreSSL on the mac...

gradle/generation/snowball.gradle
dweiss: That download task is broken in that it still executes actions even if the artifact is up to date... You can just delete the unpacked stuff and not the downloaded zip; then it'll wipe and unzip but not download again?

rmuir: @dweiss I can assure you it does not download again. Please look at the PR, maybe the names are confusing and we can clean it up? I think logic is correct.

Here is what happens (first run):
* download `github.com/snowball/xyz.zip` to `snowball.zip`. It extracts this to `stemmers/`. It saves Etag to `stemmers.json`.
* create a clean copy of `stemmers/` named `patched/`.
* apply patch to `patched/`.

Second run:
* Tries to download, but sends request with `If-None-Match: a4gk4k4...`. Remote server returns a HTTP 304 (Not Modified). This would not be the case if, e.g. someone bumped the snowball commit hash locally.
* create a clean copy of `stemmers/` named `patched/`.
* apply patch to `patched/`.

dweiss: I know it doesn't download twice but if it declared inputs/ outputs or upToDateWhen correctly then it shouldn't run any actions attached to the task (doFirst/ doLast) and it does it now, regardless of whether it downloaded the artifact or not, right (that was my impression from other sections of the code, no this patch). 

Anyway, if it's working then fine. As for different versions of patch -- throw git/ jgit's implementation in the mix... they all seem to be behaving differently in corner cases.

rmuir: note this same trick (HTTP cache revalidation) cannot be used for moman, because its server disables caching completely and doesn't send such headers. In fact the server does not even respond correctly to requests with `curl` (unless you modify User-Agent header) so it is clear they are not excited about lots of automated requests, so it should be improved some other way...

rmuir: @dweiss ok, I will look at inputs/outputs/upToDateWhen. Maybe a simpler way is for these tools to download to a file with commit's hash in the name and declare such stuff? We'd be able to fix moman to not download over and over too (cache headers cannot work over there).

dweiss: This isn't something you have to look into, Rob. I'm just adding some comments, that's it. The patch is fine.

The simplest way of making the task not download the file again would be to declare an inputs.file(path-to-zip) I think. But it's really a marginal issue, I wouldn't spend any special time on this if it works.

rmuir: @dweiss So we can add these dependencies, but I just don't want it to patch twice :) That's kinda why their is an explicit "clean" here to create patched/ work area.

I think behavior is fast enough for a first go and we can followup? It takes 26s the first time and 4s the second time for me.

I am worried about making it too smart and declaring such dependencies incorrectly. I want to make sure it does the right thing if you e.g. update commit hash/patch file, or change script, so that its easier to upgrade snowball.

dweiss: It wasn't my intention to change the patch - these were my random musings about gradle, the download task, etc. It's definitely fine for committing in and can be improved later (or never if we can convince people from snowball to generate better code).

rmuir: Yes but if you teach a man to fish... I will look into it. Its just tricky, for example patching task in my mind depends on "cleanCheckoutOfCurrentlySpecifiedCommitHash + input patch file itself". and maybe we must "touch" an output file indicating patch was applied. This way it doesn't try to apply it twice. At least this is how i would do it with something like `make`

dweiss: Declare inputs on the zip and move applying the patch to inside the download task action (you don't even have to copy the unpacked folder then), execute patching in doLast; if it fails the entire task won't be up-to-date (will be re-run)? Just a thought.

rmuir:
OK, `gradlew snowball` now works on my mac and linux and regenerates all files without any changes.

rmuir:
@dweiss OK I got it working not downloading every time, and avoids relying on HTTP caching. instead things are named by commit hashes. I had to add an explicit input dependency on the patch file, so that it redownloads and patches if needed.

Now you see on second run:
{noformat}
> Task :lucene:analysis:common:downloadSnowballData UP-TO-DATE
> Task :lucene:analysis:common:downloadSnowballStemmers UP-TO-DATE
> Task :lucene:analysis:common:downloadSnowballWebsite UP-TO-DATE
{noformat}

PR #1266, opened 2020-02-20
SOLR-14270 export command to have an option to write to a zip file

noblepaul:


solr/core/src/java/org/apache/solr/util/ExportTool.java
```diff
@@ -390,6 +400,8 @@ void exportDocs() throws Exception {
             //ignore
           }
         }
+        System.out.println("\nTotal Docs exported: "+ (docsWritten.get() -1)+
```
anshumg: @noblepaul  - the precommit failed due to the use of `System.out`

noblepaul: fixing...

noblepaul:
SOLR-14270 export command to have an option to write to a zip file

PR #1268, opened 2020-02-20
SOLR-14271: Remove duplicate async id check meant for pre Solr 8 versions

anshumg:


madrob:
LGTM

anshumg:
Fixed the precommit issue that was introduced by another commit on master. The failure here shouldn't be a problem.

PR #1271, opened 2020-02-20
LUCENE-9155: Add Apache License header to the Kuromoji dictionary compilation

anshumg:


dweiss:
Hmmm... I swear I've ran precommit before committing... but I reverted to that commit locally and see it fails properly. Something went wrong - thank you for correcting, @anshumg

PR #1274, opened 2020-02-21
LUCENE-9164: Prevent IW from closing gracefully if threads are still modifying

s1monw:
This is an alternative approach to https://github.com/apache/lucene-solr/pull/1215 It's still a rough cut and more up for discussion. @dnhatn @jpountz do you wanna take a look


dnhatn:
@s1monw There are two related test failures: testHangDuringRollback and testCloseWithThreads. We might need to adjust them.

s1monw:
@dnhatn I did change some parts slightly to wait for threads after all resources are closed like merge schedulers etc. I am still a bit nervous about this change but tests are passing.

s1monw:
> This looks great. I've gone through all methods in IndexWriter. I think prepareCommitInternal, tryModifyDocument, and deleteUnusedFiles should be executed under a modification lease.

Thanks for checking this. We can't and shouldn't ever acquire a modification lease while we are inside a `synchronized(this)` block. I know concurrency is a little mess in IW and needs to be reworked. I do have some ideas for this to streamline it more or at lease make it less trappy. I left some comments. For `prepareCommitInternal` we own the `commitLock` so we are good to go too.

s1monw:
I'd still want @mikemccand to look at this

mikemccand:
I'll have a look!

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
```diff
@@ -1554,20 +1554,24 @@ public long deleteDocuments(Query... queries) throws IOException {
...
+    try (Closeable finalizer = acquireModificationLease()) {
+      return maybeProcessEvents(docWriter.deleteQueries(queries));
     } catch (VirtualMachineError tragedy) {
       tragicEvent(tragedy, "deleteDocuments(Query..)");
       throw tragedy;
     }
   }
 
+  private Closeable acquireModificationLease() {
+    assert Thread.holdsLock(this) == false : "Can't acquire modification lock while holding the IW lock";
```
mikemccand: Can/should we also assert the reverse?  I.e. whenever we `sync` on `this`, `assert` that the `modificationLease` is not held?  They are mutually exclusive?

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
```diff
@@ -2244,21 +2233,22 @@ public void rollback() throws IOException {
     // Ensure that only one thread actually gets to do the
     // closing, and make sure no commit is also in progress:
     if (shouldClose(true)) {
-      rollbackInternal();
+      rollbackInternal(true);
     }
   }
 
-  private void rollbackInternal() throws IOException {
+  private void rollbackInternal(boolean gracefully) throws IOException {
```
mikemccand: Can you add private javadocs explaining what `gracefully` means?

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
```diff
@@ -2275,57 +2265,73 @@ private void rollbackInternalNoCommit() throws IOException {
...
-        if (pendingCommit != null) {
-          pendingCommit.rollbackCommit(directory);
-          try {
-            deleter.decRef(pendingCommit);
-          } finally {
-            pendingCommit = null;
-            notifyAll();
+      // we might rollback due to a tragic event which means we potentially already have
+      // a lease in this case we we can't acquire all leases. In this case rolling back in best effort in terms
+      // off letting all threads gracefully finish. In such a situation some threads might run into
```
mikemccand: s/`off`/`of`

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
```diff
@@ -2275,57 +2265,73 @@ private void rollbackInternalNoCommit() throws IOException {
...
+      // AlreadyClosedExceptions in places they normally wouldn't which doesn't have any impact on
+      // correctness or consistency. The tragic event is fatal anyway.
+      final int leases;
+      if (gracefully) { // in the case
+        leases = Integer.MAX_VALUE;
+        modificationLease.acquireUninterruptibly(leases);
+      } else {
+        // still try to drain all permits to prevent any new threads modifying the index.
+        leases = modificationLease.drainPermits();
```
mikemccand: This will block until all threads return their leases?  Can you add a comment to that effect?

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
```diff
@@ -2275,57 +2265,73 @@ private void rollbackInternalNoCommit() throws IOException {
...
-            notifyAll();
+      // we might rollback due to a tragic event which means we potentially already have
+      // a lease in this case we we can't acquire all leases. In this case rolling back in best effort in terms
+      // off letting all threads gracefully finish. In such a situation some threads might run into
+      // AlreadyClosedExceptions in places they normally wouldn't which doesn't have any impact on
+      // correctness or consistency. The tragic event is fatal anyway.
+      final int leases;
+      if (gracefully) { // in the case
+        leases = Integer.MAX_VALUE;
+        modificationLease.acquireUninterruptibly(leases);
```
mikemccand: This will also block until any other threads holding modification leases close them, and then will acquire all leases and prevent other threads from doing so?  Can you add a comment?

s1monw:
> It's scary we are adding yet another synchronization tool to IndexWriter; I think we need to add comments to make it clear how it interacts with at least IndexWriter's monitor lock (synchronized).

I agree with you. I do wonder if it's worth it. It might be better to go back and simplify things (I try for years now) rather than making it more complicated.  the change proposed here might be the better solution after all https://github.com/apache/lucene-solr/pull/1215

mikemccand:
> the change proposed here might be the better solution after all #1215

Yeah, +1 -- that change is very simple.  I do think it's still a workaround for true simplification for IW (I agree this is long overdue and also very challenging!!) since `AlreadyClosedException` really should never be thrown and caught by a tragic handler.  But it is a very simple solution to the issue.

s1monw:
@mikemccand @dnhatn I explored one more idea that less intrusive and more contained. I like this one much better https://github.com/apache/lucene-solr/pull/1319

s1monw:
closing in favor of #1319

PR #1281, opened 2020-02-24
LUCENE-9245: Reduce AutomatonTermsEnum memory usage.

bruno-roustant:


lucene/core/src/java/org/apache/lucene/util/automaton/Operations.java
```diff
@@ -1091,25 +1091,33 @@ public static String getCommonPrefix(Automaton a) {
    * @return common prefix, which can be an empty (length 0) BytesRef (never null)
    */
   public static BytesRef getCommonPrefixBytesRef(Automaton a) {
```
bruno-roustant: Lazy create structures (WildcardQuery may not have a first single-path state), and only one visited.add() call. This speeds up mainly WildcardQuery (a couple % in my luceneutil measures).

lucene/core/src/java/org/apache/lucene/index/AutomatonTermsEnum.java
```diff
@@ -54,18 +56,20 @@
   private final boolean finite;
   // array of sorted transitions for each state, indexed by state number
   private final Automaton automaton;
-  // for path tracking: each long records gen when we last
+  // for path tracking: each short records gen when we last
   // visited the state; we use gens to avoid having to clear
-  private final long[] visited;
```
bruno-roustant: Main change is replacing the long[] by a short[] to reduce memory usage (the size is the number of automaton states). I also tried a FixedBitSet but it impacts perf.

lucene/core/src/java/org/apache/lucene/index/AutomatonTermsEnum.java
```diff
@@ -160,17 +161,18 @@ private void setLinear(int position) {
     if (maxInterval != 0xff)
       maxInterval++;
     int length = position + 1; /* position + maxTransition */
-    if (linearUpperBound.bytes.length < length)
-      linearUpperBound.bytes = new byte[length];
+    if (linearUpperBound == null) {
+      linearUpperBound = new BytesRef(ArrayUtil.oversize(Math.max(length, 16), Byte.BYTES));
+    } else if (linearUpperBound.bytes.length < length) {
+      linearUpperBound.bytes = new byte[ArrayUtil.oversize(length, Byte.BYTES)];
```
rmuir: I don't think we should have the additional null check path here. It is not worth it to save 10 bytes :). Let's make linearUpperBound final again. Better to initialize it to `new BytesRef()` if you really want to save 10 bytes for the case that its not used, does not require additional branches in the code, it will just get extended by the length check.

bruno-roustant: +1 thanks

lucene/core/src/java/org/apache/lucene/index/AutomatonTermsEnum.java
```diff
@@ -188,7 +188,11 @@ private boolean nextString() {
     savedStates.setIntAt(0, 0);
     
     while (true) {
-      curGen++;
+      if (++curGen == 0) {
+        // Clear the visited states every time curGen overflows (so very infrequently to not impact average perf).
+        curGen++;
```
rmuir: Can we remove this unnecessary increment. Also i'd change the comment from `overflows` to `wraps`.

lucene/core/src/java/org/apache/lucene/index/AutomatonTermsEnum.java
rmuir: visited-state-tracking is only needed when the automaton accepts an infinite language. We use it for loop detection. I think before we get too fancy with how we clear it, we should first stop being stupid about it?

So it is wasteful that we do this stuff when `finite == true` (example: fuzzy query) because we will never even look for a loop. its just that the current code unconditionally records states that it visited.

I think first, in the ctor when `finite == true`, `visited[]` can be initialized to `null` or `new long[0]` or something, and we change this line:
```
visited[state] = curGen;
```
to something like this:
```
if (!finite)
  visited[state] = curGen;
```

I agree we should separately avoid tracking 64 bits per state when only 1 is needed. But before optimizing the storage, first lets avoid doing this stuff at all for ones like complex fuzzy queries?

lucene/core/src/java/org/apache/lucene/util/automaton/Operations.java
rmuir: I don't think this is the right tradeoff. It makes the code more complex, saving the cost of creating a few simple ordinary objects. I hate to say i don't trust your benchmark, but I don't trust your benchmark represents a typical case here. We should keep this code simple, there are other ways it can be improved.

lucene/core/src/java/org/apache/lucene/index/AutomatonTermsEnum.java
bruno-roustant: Good catch.

lucene/core/src/java/org/apache/lucene/util/automaton/Operations.java
bruno-roustant: Ok, removed.

PR #1288, opened 2020-02-25
SOLR-14281: Make sharedLib configurable through SysProp

janhoy:
...and allow multiple comma separated paths

See https://issues.apache.org/jira/browse/SOLR-14281


solr/CHANGES.txt
```diff
@@ -110,6 +110,9 @@ Improvements
 
 * SOLR-14114: Add WARN to Solr log that embedded ZK is not supported in production (janhoy)
 
+* SOLR-14281: Make sharedLib configurable through SysProp and allow multiple comma separated paths.
```
dsmiley: SysProp should by "sys-prop" or system property.  The way you capitalized it makes it seem like some class.

solr/core/src/java/org/apache/solr/core/CoreContainer.java
```diff
@@ -607,19 +609,31 @@ public PackageStoreAPI getPackageStoreAPI() {
   public void load() {
     log.debug("Loading cores into CoreContainer [instanceDir={}]", loader.getInstancePath());
 
+    // Always add $SOLR_HOME/lib to the shared resource loader
+    Set<String> libDirs = new HashSet<>();
+    libDirs.add("lib");
+
+    if (!StringUtils.isBlank(cfg.getSharedLibDirectory())) {
+      List<String> sharedLibs = Arrays.asList(cfg.getSharedLibDirectory().split("\\s*,\\s*"));
```
dsmiley: I'm almost certain we've got a utility that does this so that you don't have to define your regexp

solr/core/src/java/org/apache/solr/core/CoreContainer.java
```diff
@@ -607,19 +609,31 @@ public PackageStoreAPI getPackageStoreAPI() {
   public void load() {
     log.debug("Loading cores into CoreContainer [instanceDir={}]", loader.getInstancePath());
 
+    // Always add $SOLR_HOME/lib to the shared resource loader
+    Set<String> libDirs = new HashSet<>();
```
dsmiley: Use LinkedHashSet

solr/CHANGES.txt
janhoy: Re-phrased

solr/core/src/java/org/apache/solr/core/CoreContainer.java
janhoy: I tried `StrUtils.splitSmart()` but we don't need the smartness of not splitting inside quotes etc, and it does not trim whitespace around comma. So I'll keep this unless you know about a better split-util somewhere.

PR #1289, opened 2020-02-26
LUCENE-9250: Add support for Circle2d#intersectsLine around the dateline.

iverase:
This was actually an oversight in LUCENE-8707. The original PR did support crossing over the dateline added in b59b24d but it was lost during a refactor of the classes. This PR adds the capability again. It adds more documentation to the class as well.


iverase:
I will push this shortly as it is a left over from the original commit for supporting distance queries over LatLonShapes.

PR #1290, opened 2020-02-26
LUCENE-9251: Filter equal edges with different value on isEdgeFromPolygon

iverase:
See https://issues.apache.org/jira/browse/LUCENE-9251

Unfortunately, I haven't been able to produce a meaningful unit test as I can only reproduce the issue with the complex polygon provided by the user. Still I think the change make sense.


lucene/core/src/java/org/apache/lucene/geo/Tessellator.java
```diff
@@ -915,13 +914,14 @@ private static final Node filterPoints(final Node start, Node end) {
       continueIteration = false;
       nextNode = node.next;
       prevNode = node.previous;
-      //We can filter points when they are the same, if not and they are co-linear we can only
-      //remove it if both edges have the same value in .isNextEdgeFromPolygon
-      if (isVertexEquals(node, nextNode)  ||
-          (prevNode.isNextEdgeFromPolygon == node.isNextEdgeFromPolygon &&
+      // we can filter points when:
+      if (isVertexEquals(node, nextNode)  ||   // 1. they are the same,
+         // isVertexEquals(prevNode, nextNode) || // 2.- each one starts and ends in each other
```
nknize: nit: remove comment

iverase: ups... that was a left over while testing

iverase:
I agree but the issue is that the polygon is too big to add it to the unit test and there is no really framework to read it from a file. My suggestion is to push this change as it is and open a Lucene issue to be able to test this kind go big polygons. Probably adding it to the test framework? wdyt?

iverase:
test added

lucene/core/src/test/org/apache/lucene/geo/TestTessellator.java
```diff
@@ -561,6 +561,18 @@ public void testComplexPolygon39() throws Exception {
     checkPolygon(wkt);
   }
 
+  @Nightly
+  public void testComplexPolygon40() throws Exception {
+    String wkt = GeoTestUtil.readShape("lucene-9251.wkt.gz");
```
nknize: nice!

PR #1293, opened 2020-02-27
SOLR-14044: Delete collection bug fix by changing sharedShardName to use the same blob delimiter

andyvuong:
In shared storage, deletion from the blob store works by listing files beginning with a certain prefix and reading from a blob store requires knowing some or all of the file name.

The sharedShardName, or the identifier prefixing all blob storage files per shard works by concatenating the collection name (uniqueness assumption in solr cloud overall) and the shard name with "_" joining the two. "_" is allowed in collection names which allows for a bug in listing files from other collections unintentionally.

This change changes the delimiter to use "/" which is more consistent with how index files stored in blob are structured anyway <collection identifier>/<index_file_blob_name>. Piggy-backed a minor unrelated test fix in DistributedZkUpdateProcessorTest.java that I missed earlier.

Change Summary
- Use "/" in sharedShardName
- Update tests
- Fix DistributedZkUpdateProcessorTest which tries to delete all collections after every test. In the same cleanup method the entire tmp directory we use as a local blob store gets deleted at the same time so the collection deletion errors out and leaks threads. This is a test setup issue vs a functional one. We spin up a new cluster in the different test cases anyway so we don't need to delete all collections.

### DIFF
--- a/solr/core/src/test/org/apache/solr/store/blob/metadata/CorePushPullTest.java
+++ b/solr/core/src/test/org/apache/solr/store/blob/metadata/CorePushPullTest.java
@@ -27,6 +27,7 @@ import org.apache.solr.SolrTestCaseJ4;
 import org.apache.solr.cloud.api.collections.Assign;
 import org.apache.solr.core.CoreContainer;
 import org.apache.solr.core.SolrCore;
+import org.apache.solr.store.blob.client.BlobClientUtils;
 import org.apache.solr.store.blob.client.BlobCoreMetadata;
 import org.apache.solr.store.blob.client.BlobCoreMetadataBuilder;
 import org.apache.solr.store.blob.client.CoreStorageClient;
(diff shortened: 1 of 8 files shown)


solr/core/src/test/org/apache/solr/store/blob/metadata/CorePushPullTest.java
```diff
@@ -52,7 +53,7 @@
   private static CoreStorageClient storageClient;
   private static Path localBlobDir;
   
-  private String sharedBlobName = "collectionTest_shardTest";
+  private String sharedBlobName = "collectionTest" + BlobClientUtils.BLOB_FILE_PATH_DELIMITER + "shardTest";
```
mbwaheed: Minor: use Assign.buildSharedShardName?

andyvuong:
cc @yonik can you merge. Thanks

PR #1294, opened 2020-02-27
LUCENE-9074: Slice Allocation Control Plane For Concurrent Searches

atris:
This commit introduces a mechanism to control allocation of threads to slices planned for a query.
The default implementation uses the size of backlog queue of the executor to determine if a slice
should be allocated a new thread

Supersedes #1214


lucene/core/src/java/org/apache/lucene/search/IndexSearcher.java
```diff
@@ -211,6 +215,18 @@ public IndexSearcher(IndexReaderContext context, Executor executor) {
...
     reader = context.reader();
     this.executor = executor;
+    this.sliceExecutionControlPlane = executor == null ? null : getSliceExecutionControlPlane(executor);
+    this.readerContext = context;
+    leafContexts = context.leaves();
+    this.leafSlices = executor == null ? null : slices(leafContexts);
+  }
+
+  // Package private for testing
+  IndexSearcher(IndexReaderContext context, Executor executor, SliceExecutionControlPlane sliceExecutionControlPlane) {
```
atris: To quote @jpountz :
"this doesn't need an executor, does it?"

It is needed since executor is a final member of IndexSearcher hence needs to be initialized? We could potentially always set it null since it anyways wont be used but that seemed a bit counter intuitive. WDYT?

lucene/core/src/java/org/apache/lucene/search/SliceExecutionControlPlane.java
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * Execution control plane which is responsible
+ * for execution of slices based on the current status
+ * of the system and current system load
+ */
+public interface SliceExecutionControlPlane<C, T extends Runnable> {
+  /**
+   * Invoke all slices that are allocated for the query
+   */
+  C invokeAll(Collection<T> tasks);
```
atris: To quote @jpountz 
"the generics on this interface look over-engineered?"

I am not sure as to what the correct generics should be. Could you please advice?

lucene/core/src/java/org/apache/lucene/search/DefaultSliceExecutionControlPlane.java
```diff
@@ -0,0 +1,103 @@
...
+
+    if (shouldExecuteOnCallerThread) {
+      try {
+        task.run();
+      } catch (Exception e) {
+        throw new RuntimeException(e);
+      }
+    }
+
+    if (!shouldExecuteOnCallerThread) {
```
msokolov: else?

msokolov: could we add this to the first if block? ie executor.execute() .. futures.add() ...?  Then use early return and you don't need to modify the incoming parameter

lucene/core/src/java/org/apache/lucene/search/SliceExecutionControlPlane.java
atris: Thanks for taking a look. I have fixed your comments. I have introduced the interface to allow easy extensibility. I agree that the current use case is with Future and FutureTask but that should not restrict the future use cases that are possible on top of the interface.

Regarding the name, SliceRunner does not seem to convey the right message since the class can control the way the execution of the query happens without any internal directions hence it is vital to highlight that the execution of a query can be influenced by the decisions of this class.

msokolov: This is an internal detail of IndexSearcher, right? We're always free to change the method signatures later (if we keep the classes package-private - we should!). Maybe it would help if you were to explain what extension you have in mind.

By the way, using force push makes it more difficult for reviewers since we can't easily see what changed from one version to the next. In general it's better to push your commits and then squash-merge them at the end (github will even do this for you I think)

msokolov: Also- I'm curious if you saw any performance impact from the back pressure here?

atris: I do not have a very strong preference to the extensibility model of the interface -- so removed it per your recommendation. Let me know if you have any further comments or thoughts

atris: > Also- I'm curious if you saw any performance impact from the back pressure here?

When testing with artificial CPU stress and a small threadpool size, the worst case performance tends towards sequential since all tasks will be scheduled on the caller thread

lucene/core/src/java/org/apache/lucene/search/IndexSearcher.java
jpountz: Is there anything we need to do with the executor that we couldn't do with the sliceExecutionControlPlane?

lucene/core/src/java/org/apache/lucene/search/IndexSearcher.java
```diff
@@ -662,34 +676,19 @@ public TopFieldDocs reduce(Collection<TopFieldCollector> collectors) throws IOEx
       }
       query = rewrite(query);
       final Weight weight = createWeight(query, scoreMode, 1);
-      final List<Future<C>> topDocsFutures = new ArrayList<>(leafSlices.length);
-      for (int i = 0; i < leafSlices.length - 1; ++i) {
+      final List<FutureTask> listTasks = new ArrayList<>();
```
jpountz: Let's avoid introducing warnings about generics, FutureTask needs to be parameterized?

lucene/core/src/java/org/apache/lucene/search/SliceExecutionControlPlane.java
```diff
@@ -0,0 +1,105 @@
...
+import java.util.concurrent.Future;
+import java.util.concurrent.FutureTask;
+import java.util.concurrent.RejectedExecutionException;
+
+/**
+ * Execution control plane which is responsible
+ * for execution of slices based on the current status
+ * of the system and current system load
+ */
+class SliceExecutionControlPlane {
```
jpountz: nit: I'd prefer a simpler name, e.g. `SliceExecutor`

lucene/core/src/java/org/apache/lucene/search/IndexSearcher.java
```diff
@@ -211,6 +213,18 @@ public IndexSearcher(IndexReaderContext context, Executor executor) {
...
+    leafContexts = context.leaves();
+    this.leafSlices = executor == null ? null : slices(leafContexts);
+  }
+
+  // Package private for testing
+  IndexSearcher(IndexReaderContext context, Executor executor, SliceExecutionControlPlane sliceExecutionControlPlane) {
+    assert context.isTopLevel: "IndexSearcher's ReaderContext must be topLevel for reader" + context.reader();
+    reader = context.reader();
+    this.executor = executor;
+    this.sliceExecutionControlPlane = executor == null ? null : sliceExecutionControlPlane;
```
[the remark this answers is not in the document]
atris: Not sure if I understood your point. The passed in instance is the one being assigned to the member?

atris:
@jpountz Raised another iteration, please let me know your thoughts and comments.

atris:
@jpountz Any thoughts on this one?

lucene/core/src/java/org/apache/lucene/search/IndexSearcher.java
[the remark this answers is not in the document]
jpountz: My point was that it sounds like a bug on the caller of this constructor to pass a null executor and a non-null sliceExecutionControlPlane? So I'd rather have validation around it rather than be lenient and ignore the provided sliceExecutionControlPlane if the executor is null?

lucene/core/src/java/org/apache/lucene/search/IndexSearcher.java
```diff
@@ -211,6 +213,18 @@ public IndexSearcher(IndexReaderContext context, Executor executor) {
     assert context.isTopLevel: "IndexSearcher's ReaderContext must be topLevel for reader" + context.reader();
     reader = context.reader();
     this.executor = executor;
+    this.sliceExecutor = executor == null ? null : getSliceExecutionControlPlane(executor);
+    this.readerContext = context;
+    leafContexts = context.leaves();
+    this.leafSlices = executor == null ? null : slices(leafContexts);
```
jpountz: maybe this should delegate to the below constructor?

lucene/core/src/java/org/apache/lucene/search/SliceExecutor.java
```diff
@@ -0,0 +1,105 @@
...
+    }
+  }
+
+  // Private helper method to run a task on the caller thread
+  private void runTaskOnCallerThread(FutureTask task) {
+    try {
+      task.run();
+    } catch (Exception e) {
+      throw new RuntimeException(e);
+    }
```
jpountz: we don't need this catch block as task.run() doesn't declare any non-runtime exception?

lucene/core/src/java/org/apache/lucene/search/IndexSearcher.java
```diff
@@ -933,6 +932,13 @@ public Executor getExecutor() {
     return executor;
   }
 
+  /**
+   * Returns this searchers slice execution control plane or <code>null</code> if no executor was provided
+   */
+  public SliceExecutor getSliceExecutor() {
```
jpountz: we shouldn't make this method public if it returns a pkg-private class, let's make the method pkg-private too? Or even remove it entirely as I'm not seeing any call site for it?

lucene/core/src/java/org/apache/lucene/search/SliceExecutor.java
```diff
@@ -0,0 +1,105 @@
...
+ * of the system and current system load
+ */
+class SliceExecutor {
+  private final Executor executor;
+
+  public SliceExecutor(Executor executor) {
+    this.executor = executor;
+  }
+
+  public <C> List<Future<C>> invokeAll(Collection<FutureTask<C>> tasks) {
```
jpountz: I wonder whether this is the right API. We could change the return type to `void` and use `Runnable` instead of `FutureTask` and that would still work, right? The return value isn't really useful since it has the same content as the input collection? So what about making it just: `public void invokeAll(Collection<? extends Runnable> tasks)`?

lucene/core/src/java/org/apache/lucene/search/SliceExecutor.java
```diff
@@ -0,0 +1,105 @@
...
+
+    if (executor == null) {
+      throw new IllegalArgumentException("Executor is null");
+    }
+
+    List<Future<C>> futures = new ArrayList();
+
+    int i = 0;
+
+    for (FutureTask task : tasks) {
```
jpountz: we should never use generic types without type parameters, can you address all these compilation warnings?

lucene/core/src/java/org/apache/lucene/search/SliceExecutor.java
```diff
@@ -0,0 +1,105 @@
...
+        return;
+      } catch (RejectedExecutionException e) {
+        // Execute on caller thread
+      }
+    }
+
+    runTaskOnCallerThread(task);
+
+    try {
+      futures.add(CompletableFuture.completedFuture(task.get()));
```
jpountz: this has the same effect as `futures.add(task)` unless I'm missing something

lucene/core/src/java/org/apache/lucene/search/SliceExecutor.java
atris: Agreed, fixed, thanks

lucene/core/src/java/org/apache/lucene/search/IndexSearcher.java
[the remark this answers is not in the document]
atris: +1, did not add it since it was a test method but I see your point. Added now, thanks!

atris:
@jpountz Fixed, please see and let me know your thoughts.

atris:
@jpountz Let me know if this looks fine.

lucene/core/src/test/org/apache/lucene/search/TestIndexSearcher.java
```diff
@@ -347,4 +370,82 @@ public void execute(final Runnable runnable) {
...
+        for (Sort sort : sorts) {
+          searcher.search(query, Integer.MAX_VALUE);
+          searcher.searchAfter(after, query, Integer.MAX_VALUE);
+          if (sort != null) {
+            searcher.search(query, Integer.MAX_VALUE, sort);
+            searcher.search(query, Integer.MAX_VALUE, sort, true);
+            searcher.search(query, Integer.MAX_VALUE, sort, false);
+            searcher.searchAfter(after, query, Integer.MAX_VALUE, sort);
+            searcher.searchAfter(after, query, Integer.MAX_VALUE, sort, true);
+            searcher.searchAfter(after, query, Integer.MAX_VALUE, sort, false);
```
jpountz: should we assert on the results?

lucene/core/src/test/org/apache/lucene/search/TestIndexSearcher.java
```diff
@@ -347,4 +370,82 @@ public void execute(final Runnable runnable) {
       throw new RejectedExecutionException();
     }
   }
+
+  public void testQueueSizeBasedCP() throws Exception {
```
jpountz: what is CP the acronym of?

lucene/core/src/test/org/apache/lucene/search/TestIndexSearcher.java
```diff
@@ -70,13 +75,31 @@ public void setUp() throws Exception {
...
+      doc.add(newStringField("field2", Boolean.toString(i % 2 == 0), Field.Store.NO));
+      doc.add(new SortedDocValuesField("field2", new BytesRef(Boolean.toString(i % 2 == 0))));
+      iw2.addDocument(doc);
+
+      if (random.nextBoolean()) {
+        iw2.commit();
+      }
+    }
+    reader2 = iw2.getReader();
+    iw2.close();
```
jpountz: why are we not reusing reader1?

atris: reader2 has random commits baked into it to ensure random slice distributions each time the test is run.

lucene/core/src/java/org/apache/lucene/search/SliceExecutor.java
```diff
@@ -0,0 +1,85 @@
...
+      }
+    }
+
+    runTaskOnCallerThread(task);
+  }
+
+  // Private helper method to run a task on the caller thread
+  private void runTaskOnCallerThread(Runnable task) {
+    task.run();
+  }
```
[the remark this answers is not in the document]
atris: Fair point -- I dont know what I was thinking :)

lucene/core/src/test/org/apache/lucene/search/TestIndexSearcher.java
jpountz: Would it be a problem to add these random commits to the existing reader?

atris: I did not want to touch the existing reader since it is used by many tests -- dont see any risk though, given the specifics of the tests that use the existing reader, should be safe. Updated, thanks

atris:
@jpountz Updated. Please see and let me know your comments and thoughts

lucene/core/src/java/org/apache/lucene/search/IndexSearcher.java
```diff
@@ -208,9 +210,21 @@ public IndexSearcher(IndexReader r, Executor executor) {
...
   public IndexSearcher(IndexReaderContext context, Executor executor) {
+    this(context, executor, getSliceExecutionControlPlane(executor));
+  }
+
+  // Package private for testing
+  IndexSearcher(IndexReaderContext context, Executor executor, SliceExecutor sliceExecutor) {
     assert context.isTopLevel: "IndexSearcher's ReaderContext must be topLevel for reader" + context.reader();
+    if (executor == null) {
+      assert sliceExecutor == null;
+    }
```
jpountz, line 221: this is covered by the below assert already?

lucene/core/src/test/org/apache/lucene/search/TestIndexSearcher.java
```diff
@@ -347,4 +354,89 @@ public void execute(final Runnable runnable) {
...
+            assert topDocs.totalHits.value > 0;
+
+            topDocs = searcher.searchAfter(after, query, Integer.MAX_VALUE, sort);
+            assert topDocs.totalHits.value > 0;
+
+            topDocs = searcher.searchAfter(after, query, Integer.MAX_VALUE, sort, true);
+            assert topDocs.totalHits.value > 0;
+
+            topDocs = searcher.searchAfter(after, query, Integer.MAX_VALUE, sort, false);
+            assert topDocs.totalHits.value > 0;
```
jpountz, line 419: we should use assertTrue rather than assert in tests, since the latter only runs when assertions are enabled, and it can be useful to run tests with assertions disabled to make sure that the logic remains correct when assertions are off

atris:
Thank you @jpountz

lucene/core/src/java/org/apache/lucene/search/SliceExecutor.java
```diff
@@ -0,0 +1,80 @@
...
+import java.util.Collection;
+import java.util.concurrent.Executor;
+import java.util.concurrent.RejectedExecutionException;
+
+/**
+ * Executor which is responsible
+ * for execution of slices based on the current status
+ * of the system and current system load
+ */
+class SliceExecutor {
```
16pierre, line 29: What there a reason for not making `SliceExecutor` public class ?
I'd want to inject my own `SliceExecutor` without having to fork Lucene

atris: Not really. I will fix this and add to current release

16pierre: Great thanks a lot !

While I'm at it: 
the `reduce` operation also runs on the main thread, it would be lovely to make this configurable.

Glad to put up a PR for it if I find time. I'd hesitate between adding some more methods on `SliceExecutor`, or plumbing the `reduce` executor totally independently. 
The name `SliceExecutor` feels awkward if we also add logic for `reduce` there. Since `SliceExecutor` wasn't public, maybe we can still afford a rename ?

PR #1296, opened 2020-02-27
LUCENE-9253: Support custom dictionaries in KoreanTokenizer

danmuzi:
KoreanTokenizer does not support custom dictionaries(system, unknown) now, even though Nori provides DictionaryBuilder that creates custom dictionary.
In the current state, it is very difficult for Nori users to use a custom dictionary.
Therefore, we need to open a new constructor that uses it.

JIRA : https://issues.apache.org/jira/browse/LUCENE-9253


lucene/analysis/nori/src/java/org/apache/lucene/analysis/ko/KoreanTokenizer.java
```diff
@@ -185,16 +185,43 @@ public KoreanTokenizer(AttributeFactory factory, UserDictionary userDictionary,
...
+   * This constructor provides an entry point for users that want to construct custom language models
+   * that can be used as input to {@link org.apache.lucene.analysis.ko.util.DictionaryBuilder}.</p>
+   *
+   * @param factory the AttributeFactory to use
+   * @param systemDictionary a custom known token dictionary
+   * @param unkDictionary a custom unknown token dictionary
+   * @param connectionCosts custom token transition costs
+   * @param userDictionary Optional: if non-null, user dictionary.
+   * @param mode Decompound mode.
+   * @param outputUnknownUnigrams If true outputs unigrams for unknown words.
```
msokolov: Don't capitalize "If"

lucene/analysis/nori/src/java/org/apache/lucene/analysis/ko/dict/BinaryDictionary.java
```diff
@@ -150,7 +150,18 @@ protected final InputStream getResource(String suffix) throws IOException {
         throw new IllegalStateException("unknown resource scheme " + resourceScheme);
     }
   }
-  
+
+  public static InputStream getResource(ResourceScheme scheme, String path) throws IOException {
```
msokolov: so .. this basically follows the pattern from JapaneseTokenizer, I think. .. but somehow I don't see where we defined ResourceScheme? We're not referencing the one in kuromoji, right?

lucene/analysis/nori/src/java/org/apache/lucene/analysis/ko/KoreanTokenizer.java
danmuzi: Oh, I did that because it was capitalized before.
I'll change the other constructors as well.

lucene/analysis/nori/src/java/org/apache/lucene/analysis/ko/dict/BinaryDictionary.java
danmuzi: It's the same as Kuromoji's pattern.

But there is no relation with Kuromoji.
It's already written as an enum in this class.
https://github.com/apache/lucene-solr/blob/d5e44e95175dbf027915e162925057bbcc14200b/lucene/analysis/nori/src/java/org/apache/lucene/analysis/ko/dict/BinaryDictionary.java#L45-L47

danmuzi:
Thanks for your review! @msokolov

I changed some upper cases in Javadoc.
And I added a change log for this patch.

danmuzi:
@jimczi 
I'll push it to master and 8.x branch because the branch 8.5 cut date is almost up.
Please let me know if there is a problem :)

PR #1298, opened 2020-02-27
SOLR-14289 Skip ZkChroot check when not necessary

madrob:


solr/core/src/java/org/apache/solr/servlet/SolrDispatchFilter.java
```diff
@@ -285,7 +285,7 @@ public static NodeConfig loadNodeConfig(Path solrHome, Properties nodeProperties
           if (zkClient.exists("/solr.xml", true)) {
             log.info("solr.xml found in ZooKeeper. Loading...");
             byte[] data = zkClient.getData("/solr.xml", null, null, true);
-            return SolrXmlConfig.fromInputStream(loader, new ByteArrayInputStream(data));
+            return SolrXmlConfig.fromInputStream(loader, new ByteArrayInputStream(data), "zookeeper");
```
dsmiley: So, what other than "zookeeper" is used here?  Do we just need a boolean if this is the only value?

madrob: Yea, a boolean would be sufficient. I was thinking about if we need to have other sources in the future, but we can change this to a string/enum later.

dsmiley: Then I much prefer a boolean as it'd be much clearer -- "isInZooKeeper" or some-such.

madrob:
Switched to a boolean, and rebased to account for SOLR-14197 changes

PR #1304, opened 2020-03-01
LUCENE-9242: generate javadocs by calling Ant javadoc task

mocobeta:
## Description

Draft PR that adds a gradle task to generate javadocs by invoking Ant javadoc task.
All generated javadocs passed "checkMissingDocs" check.

Related Jira: https://issues.apache.org/jira/browse/LUCENE-9242


mocobeta:
To confirm if the inter-module links are correctly generated, the "broken links check" task will be of help for us (it isn't ported to gradle yet).

gradle/invoke-javadoc.gradle
```diff
@@ -0,0 +1,335 @@
...
+
+      inputs.property("linksource", "no")
+      inputs.property("linkJUnit", false)
+      inputs.property("linkHref", [])
+
+      inputs.files sourceSets.main.java.asFileTree
+      outputs.dir project.javadocRoot
+
+      doFirst {
+        srcDirs.each { srcDir ->
```
dweiss: This could also be task-per sourceset as in ecj-lint snippet... It also makes me wonder why would each source set render a different javadoc output and when is this practical... Do we even have projects that have multiple primary sourceSets?

gradle/invoke-javadoc.gradle
```diff
@@ -0,0 +1,335 @@
...
+    invokeJavadoc {
+      [':lucene:memory',
+       ':lucene:highlighter',
+       ':lucene:analysis:common',
+       ':lucene:queryparser',
+       ':lucene:facet',
+       ':lucene:spatial-extras'].collect { path ->
+        dependsOn "${path}:invokeJavadoc"
+      }
+      inputs.property('linkHref', inputs.properties.linkHref + [ '../memory', '../highlighter', '../analyzers-common', '../queryparser', '../facet', '../spatial-extras' ])
```
dweiss: These relative paths should ideally go through project names and transform into relative paths. So that in case we change project paths it'd fail early and not silently point at a wrong folder?

[":lucene:memory", ":lucene:highlighter", ...].collect { prjPath -> "../" + project(prjPath).name }

mocobeta:
@dweiss thanks for your comments, I will try to update the PR.

> I have mixed feelings about the centralized output folder for these javadocs... It makes projects very inter-connected and relying on each other. I don't know if the linking could be done any different so I think this can be left as is and maybe somebody will figure it out later.

I think there is a natural option (and may be only one option as far as I know) to not to use centralized output folder and preserve the per-project outputs. Solr points to Lucene javadocs via absolute URLs like this. I'm sure we can take the same approach for all inter-project links and this will not affect on existing (published) folder structure but only on the "href" attributes.
https://github.com/mocobeta/lucene-solr/blob/master/solr/common-build.xml#L298

I have no strong opinions about how the links should be, and it's okay to me we change the relative paths to absolute urls when switching to gradle build. One concern about switching to use absolute urls is, I think, we will not be able to check broken links without fixing `checkJavadocLinks.py` as before (this doesn't check if the external links exist or not).

Should we remove all relative paths for inter-module links, to keep projects strictly separated? (It looks like just a matter of decision-making...)

dweiss:
Let's leave this for a follow-up issue, Tomoko. This patch is already a lot of work, improvements can come later.

gradle/invoke-javadoc.gradle
```diff
@@ -0,0 +1,335 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// invoke javadoc tool
+
+allprojects {
+
+  ext {
```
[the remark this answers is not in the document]
mocobeta: moved `ext{...}` under plugins.withType().

gradle/invoke-javadoc.gradle
```diff
@@ -0,0 +1,335 @@
...
+    javadocRoot = project.path.startsWith(':lucene') ? project(':lucene').file("build/docs") : project(':solr').file("build/docs")
+    javadocDestDir = "${javadocRoot}/${project.name}"
+  }
+
+  plugins.withType(JavaPlugin) {
+    def libName = project.path.startsWith(":lucene") ? "Lucene" : "Solr"
+    def title = "${libName} ${project.version} ${project.name} API".toString()
+    def srcDirs = sourceSets.main.java.srcDirs.findAll { dir -> dir.exists() }
```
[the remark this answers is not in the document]
mocobeta: moved `def srcDirs = ...` into doFirst.

gradle/invoke-javadoc.gradle
```diff
@@ -0,0 +1,335 @@
...
+
+    task invokeJavadoc {
```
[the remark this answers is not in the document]
mocobeta: changed the name to `renderJavadoc`.

gradle/invoke-javadoc.gradle
```diff
@@ -0,0 +1,335 @@
...
+      description "Generates Javadoc API documentation for the main source code. This invokes Ant Javadoc Task."
+      group "documentation"
+
+      dependsOn sourceSets.main.compileClasspath
+
+      inputs.property("linksource", "no")
```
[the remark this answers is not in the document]
mocobeta: removed the inputs properties and defined them as the task ext properties.

gradle/invoke-javadoc.gradle
```diff
@@ -0,0 +1,335 @@
...
+      inputs.property("linkJUnit", false)
+      inputs.property("linkHref", [])
+
+      inputs.files sourceSets.main.java.asFileTree
```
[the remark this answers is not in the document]
mocobeta: changed to `inputs.files { sourceSets.main.java.asFileTree }`

gradle/invoke-javadoc.gradle
mocobeta: Instead of repeatly invoking ant.javadoc, I added (possibly multiple) `packageset`s to the javadoc task. For now, all projects have only one main source directory, so only one packageset element will be created by this code.    
```
srcDirs.collect { srcDir ->
    packageset(dir: srcDir)
}
```
Does this change make sense?

gradle/invoke-javadoc.gradle
mocobeta: The relative paths are not necessarily inferred from their project name.
- ":lucene:analysis:icu" => docs/analyzers-icu
- ":solr:solrj" => docs/solr-solrj
- ":solr:contrib:extraction" => docs/solr-cell (!)
...

How can we properly map the projects and javadoc output directories (without messing up the task definitions any further)...?

gradle/invoke-javadoc.gradle
```diff
@@ -0,0 +1,335 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// invoke javadoc tool
+
+allprojects {
+
+  ext {
+    javadocRoot = project.path.startsWith(':lucene') ? project(':lucene').file("build/docs") : project(':solr').file("build/docs")
```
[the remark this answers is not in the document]
mocobeta: yes we will be able to fix this problem in a follow-up issue...

mocobeta:
Hi @dweiss,
I updated the scripts and added some replies to your comments. Can you please take a look again?

gradle/validation/missing-docs-check.gradle
```diff
@@ -34,8 +34,9 @@ allprojects {
         ':lucene:spatial3d',
     ]
 
-    task checkMissingDocsDefault(type: CheckMissingDocsTask, dependsOn: 'javadoc') {
-      dirs += [ project.javadoc.destinationDir ]
+    task checkMissingDocsDefault(type: CheckMissingDocsTask, dependsOn: 'renderJavadoc') {
+      //dirs += [ project.javadoc.destinationDir ]
```
dweiss: Remove the commented bit maybe?

gradle/invoke-javadoc.gradle
dweiss: I don't know if it makes more sense but I like it. :)

gradle/invoke-javadoc.gradle
dweiss: This code is emulating ant - hence the folder name difference (I created gradle project structure with focus on clarity, not conformance to ant). You could set up a map of project paths to those "ant" folder names and then look up the corresponding path from project but I'm fine if you don't want to spend any more time on it.

gradle/validation/missing-docs-check.gradle
mocobeta: Yes, it was removed. Thank you.

gradle/invoke-javadoc.gradle
mocobeta: I'd like to leave it as it is for now, because it's still not clear to me where/how the map should be placed.
One thing I noticed is, we might also need this map (project to javadoc folder name - in other words, ant project name) when publishing the docs to apache.org web site, not just for resolving internal links?
e.g.)
https://lucene.apache.org/core/8_4_0/analyzers-icu/index.html
https://lucene.apache.org/solr/8_4_1/solr-solrj/index.html

dweiss: I don't think we should try to emulate ant here, Tomoko. Ideally, I'd rather have the documentation follow gradle project path here. So ".../core/[ver]/analysis/icu/index.html", for example. The fewer exceptions and hand-written mappings the easier it is to change things later on.

mocobeta: > Ideally, I'd rather have the documentation follow gradle project path here. So ".../core/[ver]/analysis/icu/index.html", for example. 

I agree with that it makes things easy and i myself would love to throw away the convention come from Ant for simplicity. 
But changing folder structure on lucene.apahce.org would be a bit big change for devs, maybe it's required to gain an agreement with others...?

dweiss: I think this change should follow dropping ant build entirely from master. Then things become simpler to manage. These links are prefixed with version number anyway so I don't think there is harm in the docs structure changing for /9_0_0/....

mocobeta: Ok I will create another patch to make the folder structure follow the gradle project path, instead of ant one.

dweiss: You can also create a placeholder issue for such a change, Tomoko. I don't think we can introduce it until we drop ant anyway (?).

PR #1310, opened 2020-03-03
SOLR-13350: Multithreaded search using collector managers

chatman:
This is almost complete. Here, all queries are multi-threaded.
TODO: Implement a query time parameter (default: off) to enable multi-threaded searching.

Tests pass.


solr/core/src/java/org/apache/solr/core/CoreContainer.java
```diff
@@ -543,6 +547,7 @@ protected CoreContainer(Object testConstructor) {
     cfg = null;
     containerProperties = null;
     replayUpdatesExecutor = null;
+    this.collectorExecutor = ExecutorUtil.newMDCAwareCachedThreadPool(new DefaultSolrThreadFactory("searcherCollectorExecutor"));
```
dsmiley: Wait; didn't you already initialize this?

lucene/core/src/java/org/apache/lucene/search/MultiCollector.java
```diff
@@ -115,6 +116,11 @@ public ScoreMode scoreMode() {
     return scoreMode;
   }
 
+  // nocommit: need to raise a LUCENE jira for this?
+  public List<Collector> getCollectors() {
```
dsmiley: This change is small and simple enough that I suggest just at-mentioning one or two people you think would care about this to get their opinion if it's worth a separate JIRA, and separately if it's even worth a Lucene changes.txt entry.  I think not worth it.  Remember the higher javadoc standards :-)

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -1643,6 +1660,83 @@ public ScoreMode scoreMode() {
...
+        DocSet docSet = new BitDocSet(new FixedBitSet(maxDoc())); // TODO: if docset is not needed, avoid this initialization
+        int i = 0;
+        for (MultiCollector multiCollector: multiCollectors) {
+          int c = 0;
+          List<Collector> subCollectors = multiCollector.getCollectors();
+          TopDocsCollector topDocsCollector = needTopDocs? ((TopDocsCollector) subCollectors.get(c++)): null;
+          MaxScoreCollector maxScoreCollector = needMaxScore? ((MaxScoreCollector) subCollectors.get(c++)): null;
+          DocSetCollector docSetCollector = needDocSet? ((DocSetCollector) subCollectors.get(c++)): null;
+          
+          if (needTopDocs)  topDocs[i++] = (topDocsCollector instanceof TopFieldCollector)? ((TopFieldCollector)topDocsCollector).topDocs(0, len): topDocsCollector.topDocs(0, len);
```
dsmiley: The ternary check & cast is not necessary here?

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -1643,6 +1660,83 @@ public ScoreMode scoreMode() {
...
+        if (needMaxScore) collectors.add(new MaxScoreCollector());
+        if (needDocSet) collectors.add(new DocSetCollector(maxDoc()));
+        return (MultiCollector) MultiCollector.wrap(collectors);
+      }
+
+      @Override
+      public CollectorManagerResult reduce(Collection<MultiCollector> multiCollectors) throws IOException {
+        final TopDocs[] topDocs = new TopDocs[multiCollectors.size()];
+        float maxScore = 0.0f;
+        DocSet docSet = new BitDocSet(new FixedBitSet(maxDoc())); // TODO: if docset is not needed, avoid this initialization
```
dsmiley: I disagree that we even need this.  Init to null.  Instead of calling union, if null just set it.

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -1643,6 +1660,83 @@ public ScoreMode scoreMode() {
...
+        return new CollectorManagerResult(mergedTopDocs, docSet, maxScore, totalHits);
+      }
+
+    };
+
+    CollectorManagerResult ret;
+    try {
+      ret = super.search(query, manager);
+    } catch (Exception ex) {
+      if (ex instanceof RuntimeException &&
```
dsmiley: Yuck!  Any ideas to improve this?

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -1643,6 +1660,83 @@ public ScoreMode scoreMode() {
...
+          && ex.getCause().getCause() != null && ex.getCause().getCause() instanceof RuntimeException) {
+        throw (RuntimeException)ex.getCause().getCause();
+      } else {
+        throw ex;
+      }
+    }
+    return ret;
+  }
+
+  class CollectorManagerResult {
```
dsmiley: Make all fields final if you can

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -1643,6 +1660,83 @@ public ScoreMode scoreMode() {
...
   }
 
+  CollectorManagerResult searchCollectorManagers(int len, QueryCommand cmd, Query query,
+      boolean needTopDocs, boolean needMaxScore, boolean needDocSet) throws IOException {
+    CollectorManager<MultiCollector, CollectorManagerResult> manager = new CollectorManager<MultiCollector, CollectorManagerResult>() {
+      @Override
+      public MultiCollector newCollector() throws IOException {
+        // nocommit: Here, creating a MultiCollector for every segment (correctness > speed).
+        // Need to explore sharing a single MultiCollector with every segment. Are these
+        // sub-collectors thread-safe? DocSetCollector seems like not thread-safe, does someone know?
```
dsmiley: DocSetCollector is *not* threadsafe.  That is something Etsy tried to tackle.  I think if we had per-segment DocSets (and thus segment oriented filter cache), then this would be a non-issue?  I don't think you should tackle this in this issue; it's a TODO for another day.

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -1712,33 +1806,47 @@ public ScoreMode scoreMode() {
...
-      DocSetCollector setCollector = new DocSetCollector(maxDoc);
-      MaxScoreCollector maxScoreCollector = null;
-      List<Collector> collectors = new ArrayList<>(Arrays.asList(topCollector, setCollector));
-
-      if ((cmd.getFlags() & GET_SCORES) != 0) {
-        maxScoreCollector = new MaxScoreCollector();
-        collectors.add(maxScoreCollector);
+      TopDocs topDocs;
+      if (pf.postFilter != null || cmd.getSegmentTerminateEarly() || cmd.getTimeAllowed() > 0
+          || query instanceof RankQuery || query instanceof GraphQuery) {
```
dsmiley: What makes GraphQuery so special?  Maybe we should define a marker interface here?  What if the Query is a BooleanQuery that somewhere inside _contains_ a GraphQuery?

solr/core/src/test/org/apache/solr/search/TestFiltering.java
```diff
@@ -86,21 +86,38 @@ public void testLiveDocsSharing() throws Exception {
...
+        assertTrue( equals(live, set) );
       }
 
     } finally {
       req.close();
     }
   }
 
+  boolean equals(DocSet ds1, DocSet ds2) {
+    DocSet smaller = ds1.getFixedBitSet().length() < ds2.getFixedBitSet().length() ? ds1: ds2;
```
dsmiley: Do we really need this much code here to check this?  I think I've seen DocSet equality elsewhere; perhaps DocSet should implement this for real.  I had taken this away recently but it could return.

solr/core/src/java/org/apache/solr/request/SolrRequestInfo.java
```diff
@@ -67,6 +70,19 @@ public static void setRequestInfo(SolrRequestInfo info) {
     threadLocal.set(info);
   }
 
+  public static void init() {
```
dsmiley: No usages found

chatman:
PLanning to update this and revive this effort shortly, hopefully for 8.7.

chatman:
@atris I've brought your changes and updated this PR. There are test failures now, seems like something may have changed in master to have caused these since the time David reviewed it (tests passed fine back then).

solr/core/src/java/org/apache/solr/core/NodeConfig.java
```diff
@@ -309,6 +317,7 @@ public boolean isFromZookeeper() {
     private final String nodeName;
 
     public static final int DEFAULT_CORE_LOAD_THREADS = 3;
+    public static final int DEFAULT_COLLECTORS_POOL_SIZE = 32768;
```
mkhludnev, line 320: Is it number of threads? Ins't it too many?

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
mkhludnev: > What if the Query is a BooleanQuery that somewhere inside _contains_ a GraphQuery?

`QueryVisitor` to rescue.

dsmiley: I should have clarified that I knew the answer is QueryVisitor :-). I wanted those working on the PR to remember that in general it's questionable to aQuery instanceof XYZ.

cpoerschke:
newer PR here: https://github.com/apache/solr/pull/2248

PR #1313, opened 2020-03-03
LUCENE-8962: Split test case

msfroh:
The testMergeOnCommit test case was trying to verify too many things
at once: basic semantics of merge on commit and proper behavior when
a bunch of indexing threads are writing and committing all at once.

Splitting the test into two should make the tests more robust - one
will verify basic behavior, with strict assertions on invariants, while
the other just verifies that everything gets indexed and we don't throw
an exception when multiple threads are indexing and merging on commit.

Also, the part of the test that is now testMultithreadedMergeOnCommit
can take several seconds to run, so moving it to the @Nightly set.

# Description

Fixing an intermittent test failure on testMergeOnCommit.

# Solution

Split the logic from testMergeOnCommit into two test cases. The basic test has consistently passed, and actually verifies the merge on commit invariants. The more complicated, more potentially-brittle multithreaded test doesn't necessarily satisfy clear invariants (as we may be merging on commit from multiple threads, which could result in multiple segments in the end), but it should never throw an exception or lose any updates.

# Tests

Split existing test case into two test cases. Ran tests multiple times.

# Checklist

### DIFF
--- a/lucene/core/src/test/org/apache/lucene/index/TestIndexWriterMergePolicy.java
+++ b/lucene/core/src/test/org/apache/lucene/index/TestIndexWriterMergePolicy.java
@@ -36,7 +36,28 @@ import org.apache.lucene.util.LineFileDocs;
 import org.apache.lucene.util.LuceneTestCase;
 
 public class TestIndexWriterMergePolicy extends LuceneTestCase {
-  
+
+  private static final MergePolicy MERGE_ON_COMMIT_POLICY = new LogDocMergePolicy() {
+    @Override
+    public MergeSpecification findFullFlushMerges(MergeTrigger mergeTrigger, SegmentInfos segmentInfos, MergeContext mergeContext) {
+      // Optimize down to a single segment on commit
+      if (mergeTrigger == MergeTrigger.COMMIT && segmentInfos.size() > 1) {
+        List<SegmentCommitInfo> nonMergingSegments = new ArrayList<>();
+        for (SegmentCommitInfo sci : segmentInfos) {
+          if (mergeContext.getMergingSegments().contains(sci) == false) {
+            nonMergingSegments.add(sci);
+          }
+        }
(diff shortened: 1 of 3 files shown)


lucene/core/src/test/org/apache/lucene/index/TestIndexWriterMergePolicy.java
```diff
@@ -298,63 +320,44 @@ public void testMergeOnCommit() throws IOException, InterruptedException {
...
+      assertEquals(6, reader.numDocs());
+      assertEquals(6, searcher.count(new MatchAllDocsQuery()));
+    }
+
+    writerWithMergePolicy.close();
+    dir.close();
+  }
+
+   // Test that when we have multiple indexing threads merging on commit, we never throw an exception.
+  @Nightly
```
msokolov: I ran this test a few times, and noticed it takes 2-3 minutes to complete. Was it this slow before? I don't remember it taking so long in the previous version. Perhaps it's because it's Nightly now so the constants are made larger?

msfroh: Oh, that's probably the case.

Given that there are other tests that flex indexing across multiple threads and `MockRandomMergePolicy` will occasionally merge on commit, maybe this test doesn't add any value?

msokolov: Yes, I think given it does not assert anything -- just makes sure no exceptions occur -- we should already be well-covered.

msokolov:
I verified this fixes the `TestIndexWriterExceptions2.testBasics` reported by @jpountz and also beasted that test 1000x just in case. I think we need to get ahead of this given all the fail emails from these tests, and the upcoming 8.5 release, so I'll push today

dnhatn:
I've left some comments in https://github.com/apache/lucene-solr/pull/1155.

PR #1316, opened 2020-03-04
LUCENE-8929 parallel early termination in TopFieldCollector using minmin score

msokolov:
This is a followup to https://github.com/apache/lucene-solr/pull/1235. I found some bugs in luceneutil measuring impact on concurrent search that were making progress extremely difficult! Finally I am able to measure consistently, my sanity is mostly restored, and I think this is very nearly ready to go.  I'll post a better explanation of what this is, along with some performance measurements, in the JIRA issue.


msokolov:
A few TODOs remain:

1. Consider switching to max-min termination for lower values of N, where it seems to perform better
2. Add a unit test to verify that we respect docid tie-break

msokolov:
I think this is pretty much done now. The main changes in the last batch of commits are:
1. We now switch between the two termination strategies depending on the value of numHits, using this one for larger values
2. I tightened up the code, removing unneeded conditionals
3. Improved the explanations in comments
I'll post updated (improved) performance measurements to JIRA

msokolov:
I've done pretty extensive performance testing, results are good, and unit tests are passing, but I would really appreciate some eyeballs if anyone has the time because this is a pretty sensitive area, and the implementation I've tested extensively in the wild is not exactly the same as this one, although it implements the same strategy. 

I'm also particularly interested, @jimczi, if you can comment on whether the `MaxScoreAccumulator` still provides additional benefit alongside this opto? I haven't tried removing it, but I wonder if it might be doing something redundant now - I'm not totally clear what impact setMinCompetitiveScore will have.

jimczi:
Thanks for the ping @msokolov .

>  if you can comment on whether the MaxScoreAccumulator still provides additional benefit alongside this opto? I haven't tried removing it, but I wonder if it might be doing something redundant now - I'm not totally clear what impact setMinCompetitiveScore will have.

It's redundant in spirit but the MaxScoreAccumulator is for queries sorted by relevancy. So it used so that queries  sorted by relevancy can use `setMinCompetitiveScore` even if they have a tiebreaker on another field (using TopFieldCollector). The logic is similar to what you added in the `MaxScoreTerminator` except that the side effects of changes in the maximum score are handled by the top collectors directly. I left a comment in the original issue but I think we should try to merge the optimization you have for the sorted index case in the current logic or create a new top field collector dedicated to optimize the retrieval of large top N on sorted indices. With this pr we would have 3 different objects used by concurrent requests to speed up search but I think it would be preferrable to specialize at this point. What do you think ?

msokolov:
> I think it would be preferrable to specialize at this point. What do you think ?

I tried this out, and although there is some duplication, it is much easier to read now - thanks for the idea! I also found a bunch of edge cases while doing some deeper testing, and cleaned up the code in IndexSearcher that creates the collector managers. Performance improvements are basically unchanged; maybe a bit better. I'll post numbers in the JIRA.

msokolov:
@jimczi I will push soon; LMK if you have any reservations

lucene/core/src/java/org/apache/lucene/search/ParallelSortedCollector.java
```diff
@@ -0,0 +1,612 @@
...
+    }
+
+    /**
+     * @param sort the query sort
+     * @param numHits the number of hits requested
+     * @return whether the Sort is compatible with early termination using {@link MaxScoreTerminator} and the number of hits
+     * requested is large enough to make this optimization likely to help.
+     */
+    static boolean isApplicable(Sort sort, int numHits) {
+      return numHits > TERMINATION_STRATEGY_HIT_THRESHOLD && sortIsNumeric(sort);
```
jimczi: `intervalMask` could be used as a heuristic (`numHits > interval x N`)  ?

lucene/core/src/java/org/apache/lucene/search/ParallelSortedCollector.java
```diff
@@ -0,0 +1,612 @@
...
+import org.apache.lucene.search.FieldValueHitQueue.Entry;
+import org.apache.lucene.search.TotalHits.Relation;
+
+/**
+ * A {@link Collector} for results sorted by field, optimized for early termination in
+ * the case where the {@link Sort} matches the index and the search is executed in parallel,
+ * using multiple threads.
+ *
+ * @lucene.experimental
+ */
```
jimczi: I wonder why it should be reserved for parallel collection ? Sorry I wasn't clear but I thought that specializing collectors for sorted index could open the door for more optimizations. For instance the priority queue is not needed since the index is already sorted, a lot of comparisons could be saved ? 
Regarding the usage of the `MaxScoreTerminator`, I think it would be simpler if the logic to track leaf state remains in the `LeafCollector`. We have to track the global bottom value but I don't understand why `MaxScoreTerminator` also handles leave states. This is similar to the `MaxScoreAccumulator` so you could apply the logic for termination  directly in the collector:
````
 if (hitsThresholdChecker.isThresholdReached() 
        && Double.compare(maxScoreTerminator.getCurrent(), leafTerminationState) > 0)) {
   // should early terminated
````
Checking the current bottom score does not require extensive synchronization so the interval at which the bottom value is checked is more to avoid extra comparison than thread contention.

File: lucene/core/src/java/org/apache/lucene/search/ParallelSortedCollector.java [hunk 3]
```diff
@@ -0,0 +1,612 @@
...
+          throw new CollectionTerminatedException();
+        }
+      }
+    }
+
+    @Override
+    public void setScorer(Scorable scorer) throws IOException {
+      this.scorer = scorer;
+      comparator.setScorer(scorer);
+      minCompetitiveScore = 0f;
```
jimczi: Should we check early termination here too ?
`````
if (maxScoreTerminator.updateLeafState(leafTerminationState)) {
          // Stop if across all segments we have collected enough, and our scores are no longer competitive
          totalHitsRelation = Relation.GREATER_THAN_OR_EQUAL_TO;
          //System.out.println("scoreboard terminate leaf " + context.ord + " doc=" + context.docBase + "+" + doc + " totalHits=" + totalHits + " score=" + (long) score);
          throw new CollectionTerminatedException();
        }
````

File: lucene/core/src/java/org/apache/lucene/search/ParallelSortedCollector.java [hunk 4]
```diff
@@ -0,0 +1,612 @@
...
+    }
+
+    void countHit() {
+      ++totalHits;
+      // TODO: replace hitsThresholdChecker with something simpler
+      hitsThresholdChecker.incrementHitCount();
+    }
+
+    void collectHitIfCompetitive(int doc) throws IOException {
+      if (reverseMul * comparator.compareBottom(doc) > 0) {
```
jimczi: if we specialize for parallel sorted  index we should get ride of this comparison ? A linked list should also be enough to keep track of the top N  per leave ?

lucene/core/src/java/org/apache/lucene/search/ParallelSortedCollector.java
```diff
@@ -0,0 +1,612 @@
...
+      intervalMask = interval - 1;
+    }
+
+    /**
+     * Must be called by leaf collectors on every interval hits to update their progress; they should
+     * call when ((leafCollector.numCollected &amp; this.intervalMask) == 0).
+     * @param newLeafState the leaf collector's current lowest score
+     * @return whether the collector should terminate
+     */
+    synchronized boolean updateLeafState(LeafState newLeafState) {
```
jimczi: Updating the max score should be synchronized but I don't think we should check termination under a lock. It should be ok to check an outdated max score as long as checks are  frequents.

lucene/core/src/java/org/apache/lucene/search/ParallelSortedCollector.java
```diff
@@ -0,0 +1,612 @@
...
+        if (newLeafState.compareTo(thresholdState) <= 0) {
+          // Tell the current leaf collector to terminate since it can no longer contribute any top hits
+          //System.out.println("  terminate " + newLeafState + " b/c it is > " + thresholdState);
+          return true;
+        }
+      }
+      return false;
+    }
+
+    private void excludeSuperfluousLeaves() {
```
jimczi: This whole logic seems wrong to me. We should track state on the leaves directly.

msokolov:
Thanks for the comments @jimczi - I'll get back to this soon, and have a look at the other PR you linked.

msokolov:
Hi Jim - thanks so much for diving in! You have some great suggestions, which I'll work to incorporate. I agree that specializing for collection over a sorted index that can be early terminated makes sense as a single collector, both for parallel and serial cases. I'm not so sure that the sorting opto Mayya is working on is related to this though - it seems like something that would help when sorting by a field in general, but isn't needed (I think?) in the specific case when the sort is by the index sort, since then no skipping is possible, although I may have misunderstood?

lucene/core/src/java/org/apache/lucene/search/ParallelSortedCollector.java
msokolov: Oh I like it!

lucene/core/src/java/org/apache/lucene/search/ParallelSortedCollector.java
msokolov: > I wonder why it should be reserved for parallel collection ?
ah, true, although somehow I think it would be less impactful in the serial case.  Also the terminator does not have to be synchronized then, I guess we could have two implementations like the threshold checker. I'll see if I can get the serial case covered.

> ...  why MaxScoreTerminator also handles leave states
The idea is that when one leaf terminates, the global max score may be updated to the max score of *some other leaf*.  That's why we have this queue of leaf states. We don't want to wait for that other leaf to collect() so it can check in to perform this update since it may already have finished collection. Also, we track not only the global max score, but also the number of hits collected by terminated collectors, and their last docid, for tiebreaking. I'll see if I can move more of the logic into the leaf collector, I agree that would be preferable to duplicating state like this, but given the need to sort the leaves (not the leaf collectors), I'm not sure how much we can do.

lucene/core/src/java/org/apache/lucene/search/ParallelSortedCollector.java
msokolov: I don't understand - do you mean check a flag like `canEarlyTerminate` that we have in TopFieldLeafCollector? -- this is only created when early termination is possible (see isApplicable)

lucene/core/src/java/org/apache/lucene/search/ParallelSortedCollector.java
msokolov: thanks, so many improvements here. I think I would be tempted to put this off for a separate issue though, given how long this one has been stewing, and the significant gains it offers already.

lucene/core/src/java/org/apache/lucene/search/ParallelSortedCollector.java
msokolov: hmm, I'm not sure - see my comment above about the need to sort the leaves (or at least find the worst leaf after terminating another).

lucene/core/src/java/org/apache/lucene/search/ParallelSortedCollector.java
msokolov: I tried to explain above -- this is the whole idea, really. Maybe there is something else we can do that would be as good or better, but I don't see it yet. Given the gains here, I'm tempted to go forward and then I'd be *perfectly happy* to see this all thrown away in favor of something neater if we can achieve the same result.

lucene/core/src/java/org/apache/lucene/search/ParallelSortedCollector.java
msokolov: I started to look at this and I see what you mean - we don't need a PQ for a given leaf. However we do need to merge hits from other leaves. It's hard to see how to do that efficiently without a PQ. We could always collect a single leaf per collector and then merge sort in the collector manager? How were you thinking to approach that?

PR #1317, opened 2020-03-05
SOLR-13101: Create metadataSuffix znode only at common shard creating api calls

andyvuong:
Shared collections make use of something called the metadataSuffix znode per shard in ZooKeeper to understand what core.metadata file to pull from blob store and hydrate a replica with the most up-to-date and correct index data. Shared collections would create the node on-demand with a default value (e.g. indexing) but this is not correct behavior when considering restore operations in the future. We'll want to create the node with a non-default value down the line. 

This change initiates the node when we create the shard in common collection API commands and includes CREATE, CREATESHARD, SPLITSHARD. Other operations such as RESTORE are not currently supported by solr shared storage so are not in scope of this change.

### DIFF
--- a/solr/core/src/java/org/apache/solr/store/shared/metadata/SharedShardMetadataController.java
+++ b/solr/core/src/java/org/apache/solr/store/shared/metadata/SharedShardMetadataController.java
@@ -1,4 +1,5 @@
 /*
+
  * Licensed to the Apache Software Foundation (ASF) under one or more
  * contributor license agreements.  See the NOTICE file distributed with
  * this work for additional information regarding copyright ownership.
@@ -54,11 +55,12 @@ public class SharedShardMetadataController {
   /**
    * Creates a new metadata node if it doesn't exist for shared shared index whose correctness metadata 
    * is managed by ZooKeeper. This node should only be created during Shard creation or Recovery events
+   * and the method will throw an exception if the node already exists.
    *  
    * @param collectionName name of the collection that needs a metadata node
    * @param shardName name of the shard that needs a metadata node
    */
-  public void initiateMetadataNode(String collectionName, String shardName) throws IOException {
+  public void createMetadataNode(String collectionName, String shardName) throws IOException {
     ClusterState clusterState = cloudManager.getClusterStateProvider().getClusterState();
     DocCollection collection = clusterState.getCollection(collectionName);
     if (!collection.getSharedIndex()) {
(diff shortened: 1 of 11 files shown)


solr/core/src/java/org/apache/solr/store/shared/metadata/SharedShardMetadataController.java
```diff
@@ -53,12 +53,12 @@ public SharedShardMetadataController(SolrCloudManager cloudManager) {
...
   /**
    * Creates a new metadata node if it doesn't exist for shared shared index whose correctness metadata 
-   * is managed by ZooKeeper
+   * is managed by ZooKeeper. This node should only be created during Shard creation or Recovery events
    *  
    * @param collectionName name of the collection that needs a metadata node
    * @param shardName name of the shard that needs a metadata node
    */
-  public void ensureMetadataNodeExists(String collectionName, String shardName) throws IOException {
+  public void initiateMetadataNode(String collectionName, String shardName) throws IOException {
```
mbwaheed: Wouldn't it be more appropriate to call it createMetadataNode and fail in case the node already exists?

solr/core/src/java/org/apache/solr/store/shared/metadata/SharedShardMetadataController.java
```diff
@@ -146,20 +148,18 @@ public void cleanUpMetadataNodes(String collectionName, String shardName) {
...
-  private void createPersistentNodeIfNonExistent(String path, byte[] data) {
+  private void createPersistentNode(String path, byte[] data) {
     try {
       if (!stateManager.hasData(path)) {
-        try {
-          stateManager.makePath(path, data, CreateMode.PERSISTENT, /* failOnExists */ false);
-        } catch (AlreadyExistsException e) {
-          // it's okay if another beats us creating the node
-        }
+        stateManager.makePath(path, data, CreateMode.PERSISTENT, /* failOnExists */ true);
```
andyvuong: #makePath will also throw an AlreadyExistsException if the node exists and failOnExists=true

solr/core/src/test/org/apache/solr/store/shared/metadata/SharedShardMetadataControllerTest.java
```diff
@@ -77,10 +77,26 @@ public void cleanup() throws Exception {
    */
   @Test
   public void testSetupMetadataNode() throws Exception {    
-    shardMetadataController.ensureMetadataNodeExists(TEST_COLLECTION_NAME, TEST_SHARD_NAME);
+    shardMetadataController.createMetadataNode(TEST_COLLECTION_NAME, TEST_SHARD_NAME);
     assertTrue(cluster.getZkClient().exists(metadataNodePath, false));
   }
   
+  /**
+   * Test if we try to create the same metadat anode twice, we faile
```
mbwaheed: few typos.

andyvuong:
cc @yonik

PR #1318, opened 2020-03-05
LUCENE-9263: Fix wrong transformation of distance in meters to radians in Geo3DPoint

iverase:
This PR removes the calculation from the planet model and move it to Geo3DUtil. Planet model is generic and it does not need and spheroid representing the earth.

@nknize can you have a look? We should fix it before the 8.5 release?


iverase:
@nknize, I had another iteration and decided to go back to one constructor that defines the semiMajorAxis and semiMinorAxis only (The fattening factor was not used and can be calculated anyway from those two values).

lucene/spatial3d/src/test/org/apache/lucene/spatial3d/TestGeo3DPoint.java
```diff
@@ -1328,6 +1329,32 @@ public void testEncodeDecodeIsStable() throws Exception {
     }
   }
 
+  public void testDistanceQuery() throws Exception {
+    Directory dir = newDirectory();
+    RandomIndexWriter w = new RandomIndexWriter(random(), dir);
+
+    PlanetModel planetModel = PlanetModel.WGS84;
```
nknize: Can we test this with the PlanetModel randomly selected from all three (Clarke, Sphere, WGS84)?

iverase: done

iverase:
One thing that worries me is that we built Planet models using minor and major axis but equality of a planet model is done through xyScaling and zScaling, is that an issue for you?

nknize:
@iverase good point. With these changes the `equals` and `hashCode` methods should also be changed to use major and minor axis

iverase:
updated

PR #1319, opened 2020-03-05
LUCENE-9164: process all events before closing gracefully

s1monw:
This is yet another / simpler approach to https://github.com/apache/lucene-solr/pull/1274 to ensure that all event are processed if we are closing the IW gracefully. This also improves the case where we closing due to a tragic event where we don't try to be heroic and just drop all pending events on the floor.


lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
```diff
@@ -299,7 +300,70 @@ static int getActualMaxDocs() {
   final FieldNumbers globalFieldNumberMap;
 
   final DocumentsWriter docWriter;
-  private final Queue<Event> eventQueue = new ConcurrentLinkedQueue<>();
+  private final CloseableQueue eventQueue = new CloseableQueue();
```
dweiss: Wouldn't it be nicer to make it just Closeable and pass IndexWriter in the constructor (instead of each method)?

s1monw:
thanks for looking @dweiss

mikemccand:
I'll try to review this one soon!

dnhatn:
Thanks, Simon. I will take a look at this tomorrow.

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
```diff
@@ -299,7 +300,76 @@ static int getActualMaxDocs() {
...
+
+    void processEvents() throws IOException {
+      tryAcquire();
+      try {
+        processEventsInternal();
+      }finally {
+        permits.release();
+      }
+    }
+    private void processEventsInternal() throws IOException {
```
dnhatn: nit: add a new line

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
```diff
@@ -299,7 +300,76 @@ static int getActualMaxDocs() {
   final FieldNumbers globalFieldNumberMap;
 
   final DocumentsWriter docWriter;
-  private final Queue<Event> eventQueue = new ConcurrentLinkedQueue<>();
+  private final CloseableQueue eventQueue = new CloseableQueue(this);
+
+  static final class CloseableQueue implements Closeable {
```
dnhatn: I am not sure if `EventQueue` is a better name?

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
```diff
@@ -299,7 +300,76 @@ static int getActualMaxDocs() {
...
+      } finally {
+        permits.release();
+      }
+    }
+
+    void processEvents() throws IOException {
+      tryAcquire();
+      try {
+        processEventsInternal();
+      }finally {
```
dnhatn: nit: space after `{`

lucene/core/src/test/org/apache/lucene/index/TestIndexWriter.java
```diff
@@ -3773,7 +3774,58 @@ public void testRefreshAndRollbackConcurrently() throws Exception {
       stopped.set(true);
       indexer.join();
       refresher.join();
+      if (w.getTragicException() != null) {
+        w.getTragicException().printStackTrace();
```
dnhatn: I think we don't need to print the stack trace here.

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
```diff
@@ -299,7 +300,77 @@ static int getActualMaxDocs() {
...
+      tryAcquire();
+      try {
+        processEventsInternal();
+      } finally {
+        permits.release();
+      }
+    }
+
+    private void processEventsInternal() throws IOException {
+      assert Integer.MAX_VALUE - permits.availablePermits() > 0 : "must acquire a permit before processing events";
```
mikemccand: Maybe rewrite to `assert permits.availablePermits() < Integer.MAX_VALUE` for easier readability?

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
```diff
@@ -299,7 +300,77 @@ static int getActualMaxDocs() {
...
+      Event event;
+      while ((event = queue.poll()) != null) {
+        event.process(writer);
+      }
+    }
+
+    @Override
+    public synchronized void close() throws IOException {
```
mikemccand: Why `synchronized` here (no other method is `synchronized`)?  Just to prevent two threads from calling `close` at once?

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
```diff
@@ -299,7 +300,77 @@ static int getActualMaxDocs() {
...
+    private volatile boolean closed = false;
+    private final Semaphore permits = new Semaphore(Integer.MAX_VALUE);
+    private final Queue<Event> queue = new ConcurrentLinkedQueue<>();
+    private final IndexWriter writer;
+
+    EventQueue(IndexWriter writer) {
+      this.writer = writer;
+    }
+
+    private void tryAcquire() {
```
mikemccand: Can we rename this to `acquire`?  If it fails, it throws an exception, so upon returning, it always succeeds?

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
```diff
@@ -299,7 +300,77 @@ static int getActualMaxDocs() {
...
+      if (permits.tryAcquire() == false) {
+        throw new AlreadyClosedException("queue is closed");
+      }
+      if (closed) {
+        throw new AlreadyClosedException("queue is closed");
```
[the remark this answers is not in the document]
dnhatn: Great catch.

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
s1monw: but this wouldn't catch the case where we didn't acquire any permit?

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
s1monw: correct

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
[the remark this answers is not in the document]
s1monw: I added a test to ensure this works

s1monw:
> I wonder/wish if we could simply use synchronized methods instead of a Semaphore with Integer.MAX_VALUE permits, but I guess we want to allow multiple threads to process events concurrently, and to allow adding new events while events are being processed, etc.

this is indeed the reason

s1monw:
@mikemccand are you ok with pushing this?

PR #1320, opened 2020-03-05
LUCENE-9257: Always keep FST off-heap. Remove FSTLoadMode and Reader attributes.

bruno-roustant:
This PR modifies many classes because it removes Reader attributes now unused because FST is always loaded off-heap.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/codecs/blocktree/FieldReader.java
+++ b/lucene/core/src/java/org/apache/lucene/codecs/blocktree/FieldReader.java
@@ -25,7 +25,6 @@ import org.apache.lucene.index.IndexOptions;
 import org.apache.lucene.index.Terms;
 import org.apache.lucene.index.TermsEnum;
 import org.apache.lucene.store.ByteArrayDataInput;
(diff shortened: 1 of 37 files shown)


lucene/core/src/java/org/apache/lucene/codecs/blocktree/FieldReader.java
```diff
@@ -82,32 +80,11 @@
...
-          throw new IllegalStateException("unknown enum constant: " + fstLoadMode);
-      }
       final IndexInput clone = indexIn.clone();
       clone.seek(indexStartFP);
-      if (isFSTOffHeap) {
-        index = new FST<>(clone, ByteSequenceOutputs.getSingleton(), new OffHeapFSTStore());
-      } else {
-        index = new FST<>(clone, ByteSequenceOutputs.getSingleton());
-      }
+      index = new FST<>(clone, ByteSequenceOutputs.getSingleton(), new OffHeapFSTStore());
```
juanka588: nice

jpountz:
In the spirit of @dsmiley 's recent email, let's add a CHANGES entry?

PR #1322, opened 2020-03-06
Remove some unused lines from addBackcompatIndexes.py related to svn

janhoy:
This is dead code in a python script. We don't use svn anymore. I did not add corresponding git commands since the releaseWizard explicitly does an add after running the script, and noone has complained for so long time :) 

Tagging @sarowe since it seems you have touched this script in the past


sarowe:
+1 LGTM

PR #1327, opened 2020-03-06
SOLR-13942: /api/cluster/zk/* to fetch raw ZK data

noblepaul:


chatman:
This must have the reference guide changes too.

solr/core/src/java/org/apache/solr/handler/admin/ZkRead.java
```diff
@@ -0,0 +1,117 @@
...
+/**Exposes the content of the Zookeeper
+ * This is an expert feature that exposes the data inside the back end zookeeper.This API may change or
+ * be removed in future versions.
+ * This is not a public API. The data that is returned is not guaranteed to remain same
+ * across releases, as the data stored in Zookeeper may change from time to time.
+ */
+@EndPoint(path = "/cluster/zk/*",
+    method = SolrRequest.METHOD.GET,
+    permission = COLL_READ_PERM)
+public class ZkRead {
```
tflobbe: The name of this class is not very representative. Also, all ZooKeeper handlers are called `ZooKeeper...`, we should keep that for consistency

solr/core/src/java/org/apache/solr/handler/admin/ZkRead.java
```diff
@@ -0,0 +1,117 @@
...
+  private final CoreContainer coreContainer;
```
tflobbe: Do we need to keep the CoreContainer? is the zkClient enough?

solr/core/src/java/org/apache/solr/handler/admin/ZkRead.java
```diff
@@ -0,0 +1,117 @@
...
+                ew1.put("cversion", stat.getCversion());
+                ew1.put("czxid", stat.getCzxid());
+                ew1.put("ephemeralOwner", stat.getEphemeralOwner());
+                ew1.put("mtime", stat.getMtime());
+                ew1.put("mzxid", stat.getMzxid());
+                ew1.put("pzxid", stat.getPzxid());
+                ew1.put("dataLength", stat.getDataLength());
+              });
+            } catch (Exception e) {
+              ew.put("s", Collections.singletonMap("error", e.getMessage()));
```
tflobbe: you mean `"s"` or `s`? Also, the idea here is that we'll add an error per child?

solr/core/src/java/org/apache/solr/handler/admin/ZkRead.java
```diff
@@ -0,0 +1,117 @@
...
+                ew1.put("ctime", stat.getCtime());
+                ew1.put("cversion", stat.getCversion());
+                ew1.put("czxid", stat.getCzxid());
+                ew1.put("ephemeralOwner", stat.getEphemeralOwner());
+                ew1.put("mtime", stat.getMtime());
+                ew1.put("mzxid", stat.getMzxid());
+                ew1.put("pzxid", stat.getPzxid());
+                ew1.put("dataLength", stat.getDataLength());
+              });
+            } catch (Exception e) {
```
tflobbe: Why are you catching `Exception` here? instead of a subclass? Also, `exists` throws interrupted exception, you should interrupt (and break)

solr/core/src/java/org/apache/solr/handler/admin/ZkRead.java
```diff
@@ -0,0 +1,117 @@
...
+        d = coreContainer.getZkController().getZkClient().getData(path, null, null, false);
+        if (d == null || d.length == 0) {
+          rsp.add(path, null);
+          return;
+        }
+
+        Map<String, String> map = new HashMap<>(1);
+        map.put(WT, "raw");
+        map.put(OMIT_HEADER, "true");
+        req.setParams(SolrParams.wrapDefaults(new MapSolrParams(map), req.getParams()));
```
tflobbe: Why is this needed?

solr/core/src/java/org/apache/solr/handler/admin/ZkRead.java
```diff
@@ -0,0 +1,117 @@
...
+
+
+        rsp.add(CONTENT, new ContentStreamBase.ByteArrayStream(d, null,
+            d[0] == '{' ? CommonParams.JSON_MIME : BinaryResponseParser.BINARY_CONTENT_TYPE));
+
+      }
+
+    } catch (Exception e) {
```
tflobbe: Same as above, we probably don't need to catch all exceptions, just the ones being thrown. And the same thing with interruption

solr/core/src/test/org/apache/solr/handler/admin/ZookeeperStatusHandlerTest.java
```diff
@@ -74,6 +78,39 @@ public void tearDown() throws Exception {
     super.tearDown();
   }
 
+  @Test
+  public void testZkread() throws Exception {
```
tflobbe: This test doesn't belong in `ZookeeperStatusHandlerTest`

solr/core/src/test/org/apache/solr/handler/admin/ZookeeperStatusHandlerTest.java
```diff
@@ -74,6 +78,39 @@ public void tearDown() throws Exception {
...
+    URL baseUrl = cluster.getJettySolrRunner(0).getBaseUrl();
+    String basezk = baseUrl.toString().replace("/solr", "/api") + "/cluster/zk";
+
+    try(  HttpSolrClient client = new HttpSolrClient.Builder(baseUrl.toString()).build()) {
```
tflobbe: nit. Fix spaces

solr/core/src/java/org/apache/solr/handler/admin/ZkRead.java
```diff
@@ -0,0 +1,117 @@
...
+                ew1.put("pzxid", stat.getPzxid());
+                ew1.put("dataLength", stat.getDataLength());
+              });
+            } catch (Exception e) {
+              ew.put("s", Collections.singletonMap("error", e.getMessage()));
+            }
+          }
+        });
+
+      } else {
```
tflobbe: Should we consider the case where znodes have data and children?

solr/core/src/test/org/apache/solr/handler/admin/ZookeeperStatusHandlerTest.java
```diff
@@ -74,6 +78,39 @@ public void tearDown() throws Exception {
...
+          Utils.JSONCONSUMER );
+      assertNotNull(o);
+      o = Utils.executeGET(client.getHttpClient(),
+          basezk + "/configs",
+          Utils.JSONCONSUMER );
+      assertEquals("0", String.valueOf(getObjectByPath(o,true, split(":/configs:_default:dataLength",':'))));
+      assertEquals("0", String.valueOf(getObjectByPath(o,true, split(":/configs:conf:dataLength",':'))));
+      byte[] bytes = new byte[1024*5];
+      for (int i = 0; i < bytes.length; i++) {
+        bytes[i] = (byte) random().nextInt(128);
```
tflobbe: why make things so complicated? cant you just put some static text? what are you trying to achieve with this randomization?

solr/core/src/test/org/apache/solr/handler/admin/ZookeeperStatusHandlerTest.java
janhoy: That's true. ZKStatusHandler is the handler that pulls 4LW status from ZK TCP API, which is a completely different handler :)

solr/core/src/java/org/apache/solr/handler/admin/ZkRead.java
noblepaul: The zkclient may get closed or recreated. Always safer to get the current one

solr/core/src/java/org/apache/solr/handler/admin/ZkRead.java
noblepaul: good catch. it was supposed to be `s`

solr/core/src/java/org/apache/solr/handler/admin/ZkRead.java
noblepaul: does it even matter? 

Why interrupt the thread when this thread is not doing anything at all. This is the `qtp` thread. Interrupting it makes no sense

solr/core/src/java/org/apache/solr/handler/admin/ZkRead.java
noblepaul: This is important if we are ever writing out raw content that comes from ZK. We should not use any of the `ResponseWriter` in Solr

solr/core/src/java/org/apache/solr/handler/admin/ZkRead.java
noblepaul: it does not matter in this case

solr/core/src/java/org/apache/solr/handler/admin/ZkRead.java
noblepaul: fixing it. you can explicitly request for the node data using a param `leaf=true`

solr/core/src/test/org/apache/solr/handler/admin/ZookeeperStatusHandlerTest.java
noblepaul: this was the easiest way to create a `byte[]` .  could use anything else. But this code was simpler

janhoy:
Not sure I like the magic switch between returning actual data and returning a list of metdatata simply based on whether a znode happens to have children or not. This is not deterministic. I'd rather have an explicit action, e.g. `GET /api/cluster/zookeeper/collections/_children` to list children for a node, and let pure `GET <path>` calls always return UTF-8 data. Futher, it must be possible to retrieve the metadata of any znode. The existing `/admin/zookeeper` API returns both metadata (ctime etc), utf-8 data and children in the same response. So consider either a `GET /api/cluster/zookeeper/collections/_metadata` call, or putting the metadata in HTTP headers `X-solr-zkmeta-ctime: foo`?

An example to show the problem is the `/collections` znode which happens to have no children on a clean node after install. Try this:

    docker run --rm -ti solr:8.4.1 bash
    solr start -c
    curl "http://localhost:8983/solr/admin/zookeeper?path=/collections&detail=true"

It will return `children_count:0`. Your current impl would just return the data of that node, i.e. an empty string, but after creating the first collection it would return a list, and you'd need to add `leaf=true` to get to the data, which is very inconsistent.

solr/core/src/java/org/apache/solr/handler/admin/ZkRead.java
tflobbe: This try/catch is inside a loop that makes requests to ZooKeeper. So you are saying that if the thread is interruped (shutdown, or in tests for example) the plan is to swallow that interruption and continue processing (doing requests to ZooKeeper for every child this zkpath has, with sleeps/retries), and the intention of that is to add an "exception" to one (just one) of the elements in the tree in the response?
Regarding other exceptions, lets say you are getting back a SessionExpiredException, the plan is to continue iterating over all the child docs and continue adding `node -> KeeperException.SessionExpiredException`?
I think the proper handling of most these exceptions is to stop processing and return an error. If you get a NoNodeException, then yes, that could be added in the response if you want, but doing a blind `catch (Exception e)` is a bad idea.

solr/core/src/java/org/apache/solr/handler/admin/ZkRead.java
tflobbe: It does. See my comment above

tflobbe: Also, wouldn't this be returning 200 OK for error cases?

solr/core/src/java/org/apache/solr/handler/admin/ZookeeperRead.java
```diff
@@ -0,0 +1,118 @@
...
+/**Exposes the content of the Zookeeper
+ * This is an expert feature that exposes the data inside the back end zookeeper.This API may change or
+ * be removed in future versions.
+ * This is not a public API. The data that is returned is not guaranteed to remain same
+ * across releases, as the data stored in Zookeeper may change from time to time.
+ */
+@EndPoint(path = "/cluster/zk/*",
+    method = SolrRequest.METHOD.GET,
+    permission = COLL_READ_PERM)
+public class ZookeeperRead {
```
tflobbe: Handlers of requests are typically called `Handler`

solr/core/src/test/org/apache/solr/handler/admin/ZookeeperStatusHandlerTest.java
tflobbe: Wait, can't you do `"this is some test content".getBytes(StandardCharsets.UTF_8);`?

solr/core/src/test/org/apache/solr/handler/admin/ZookeeperReadTest.java
```diff
@@ -0,0 +1,100 @@
...
+      o = Utils.executeGET(client.getHttpClient(),
+          basezk + "/configs",
+          Utils.JSONCONSUMER);
+      assertEquals("0", String.valueOf(getObjectByPath(o, true, split(":/configs:_default:dataLength", ':'))));
+      assertEquals("0", String.valueOf(getObjectByPath(o, true, split(":/configs:conf:dataLength", ':'))));
+
+      o = Utils.executeGET(client.getHttpClient(),
+          basezk + "/configs?leaf=true",
+          Utils.JSONCONSUMER);
+      assertTrue(((Map)o).containsKey("/configs"));
```
tflobbe: can you fix the compiler warnings?

solr/core/src/java/org/apache/solr/handler/admin/ZkRead.java
tflobbe: Ah, true. The problem is that, by requiring the CoreContainer you are making this much more difficult to test (which is why I guess you are starting a Solr cluster just to test this handler). Maybe there is a way to improve the design to improve tests

solr/core/src/java/org/apache/solr/handler/admin/ZkRead.java
tflobbe: Is this only the case for the situation where we write the data? what about the case of the child nodes? can those requests be written in other response formats? If this is the only format that's allowed, should we fail if another format is requested?

solr/core/src/java/org/apache/solr/handler/admin/ZkRead.java
noblepaul: Please take a look at the latest

* NoNodeException returns with a `404`
* Other Exceptions return with `500`

This is supposed to just fail fast if there is a problem. It does not need to be any more sophisticated than that

noblepaul:
> Not sure I like the magic switch between returning actual data and returning a list of metdatata simply based on whether a znode happens to have children or not. This is not deterministic. I'd rather have an explicit action, e.g. `GET /api/cluster/zookeeper/collections/_children` to list children for a node, and let pure `GET <path>` calls always return UTF-8 data. Futher, it must be possible to retrieve the metadata of any znode. The existing `/admin/zookeeper` API returns both metadata (ctime etc), utf-8 data and children in the same response. So consider either a `GET /api/cluster/zookeeper/collections/_metadata` call, or putting the metadata in HTTP headers `X-solr-zkmeta-ctime: foo`?
> 
> An example to show the problem is the `/collections` znode which happens to have no children on a clean node after install. Try this:
> 
> ```
> docker run --rm -ti solr:8.4.1 bash
> solr start -c
> curl "http://localhost:8983/solr/admin/zookeeper?path=/collections&detail=true"
> ```
> 
> It will return `children_count:0`. Your current impl would just return the data of that node, i.e. an empty string, but after creating the first collection it would return a list, and you'd need to add `leaf=true` to get to the data, which is very inconsistent.

I changed it to have 2 paths
* `/api/cluster/zk-data`
* `/api/cluster/zk-ls` 

There is no reason for ambiguity

solr/core/src/test/org/apache/solr/handler/admin/ZookeeperReadTest.java
noblepaul: what are the warnings?

solr/core/src/java/org/apache/solr/handler/admin/ZkRead.java
noblepaul: this is fixed

noblepaul:
I plan to commit this soon

janhoy:
Make sure to announce your intention in the JIRA issue too, and allow a few days lazy consensus.

tflobbe:
@noblepaul, seems to me that there are still some unaddressed comments

solr/core/src/java/org/apache/solr/handler/admin/ZkRead.java
tflobbe: This is still a problem in your latest code

solr/core/src/java/org/apache/solr/handler/admin/ZkRead.java
tflobbe: Have you given any thoughts to this? It's unfortunate to have a feature that has so little (nothing really) to do with other parts of Solr to have to start a full SolrCloud cluster to be able to test.

solr/core/src/test/org/apache/solr/handler/admin/ZookeeperStatusHandlerTest.java
tflobbe: This is still unresolved. Can you clarify why we need this random byte array for the test?

solr/core/src/test/org/apache/solr/handler/admin/ZookeeperReadTest.java
tflobbe: Isn't your compiler complaining about non-parametrized map?

solr/core/src/java/org/apache/solr/handler/admin/ZkRead.java
noblepaul: I fail to see the problem. Could you please explain

solr/core/src/java/org/apache/solr/handler/admin/ZookeeperRead.java
noblepaul: APIs are named as *API

solr/core/src/test/org/apache/solr/handler/admin/ZookeeperReadTest.java
noblepaul: NO. I couldn't see it. Maybe you can just commit that directly

solr/core/src/java/org/apache/solr/handler/admin/ZkRead.java
noblepaul: must be fixed now

tflobbe: Is not yet fixed in the list case

solr/core/src/java/org/apache/solr/handler/admin/ZkRead.java
noblepaul: no if you are requesting data , you should expect raw data. it will not honour the `wt` param

solr/core/src/test/org/apache/solr/handler/admin/ZookeeperStatusHandlerTest.java
noblepaul: I wanted a big enough `byte[]` not a small one.

solr/core/src/java/org/apache/solr/handler/admin/ZkRead.java
noblepaul: We should probably  support a `Supplier<ZkClient>`.

solr/core/src/java/org/apache/solr/handler/admin/ZkRead.java
noblepaul:
```
  public void list(SolrQueryRequest req, SolrQueryResponse rsp) {
      String path = req.getPathTemplateValues().get("*");
      if (path == null || path.isEmpty()) path = "/";
      try {
        List<String> l = coreContainer.getZkController().getZkClient().getChildren(path, null, false);
        String prefix = path.endsWith("/") ? path : path + "/";
        rsp.add(path, (MapWriter) ew -> {
          for (String s : l) {
            Stat stat = null;
            try {
              stat = coreContainer.getZkController().getZkClient().exists(prefix + s, null, false);
            } catch (Exception e) {
              throw new RuntimeException(e);
            }
            printStat(ew, s, stat);
          }
        });
      } catch (KeeperException.NoNodeException e) {
        throw new SolrException(SolrException.ErrorCode.NOT_FOUND, "No such node :"+ path);
      } catch (Exception e) {
        rsp.add(CONTENT, new ContentStreamBase.StringStream(Utils.toJSONString(Collections.singletonMap("error", e.getMessage()))));
      } finally {
        RequestHandlerUtils.addExperimentalFormatWarning(rsp);
      }
    }
```

If there is an exception you will see a 500 error in this code. Am I missing something?

solr/core/src/java/org/apache/solr/handler/admin/ZkRead.java
noblepaul: yes, for list operations , you can get data in any format

PR #1333, opened 2020-03-09
LUCENE-9266 Update smoke test for gradle

madrob:


lucene/common-build.xml
```diff
@@ -598,9 +599,13 @@
     <attribute name="pom.xml"/>
     <attribute name="jar.file" default="${dist.jar.dir.prefix}-${version}/${dist.jar.dir.suffix}/${final.name}.jar"/>
     <sequential>
-      <artifact:install-provider artifactId="wagon-ssh" version="1.0-beta-7"/>
+      <artifact:install-provider artifactId="wagon-ssh" version="1.0-beta-7">
+        <remoteRepository id="${maven.repository.id}" url="${ivy_bootstrap_url1}" />
```
dweiss: Why is this needed? Is this related to gradle?

madrob: This was part of LUCENE-9170, I ran into the same problem. Wasn't sure if it was related to gradle or not.

PR #1336, opened 2020-03-10
LUCENE-9270: Update Javadoc about normalizeEntry in the Kuromoji DictionaryBuilder

danmuzi:
The normalizeEntry option is missing from the Javadoc of Kuromoji DictionaryBuilder.
Without this explanation, users don't know what it means until they see the code.
Also, if user follows the usage of Javadoc, it will not be built.

Please check the following JIRA issue:
[LUCENE-9270](https://issues.apache.org/jira/browse/LUCENE-9270)


danmuzi:
Thanks for your checking, @mocobeta 👍 
I'll push it to master and 8.x branch.

PR #1338, opened 2020-03-11
LUCENE-9271: Move BufferedIndexInput to the ByteBuffer API

jpountz:
See https://issues.apache.org/jira/browse/LUCENE-9271.


lucene/core/src/java/org/apache/lucene/store/BufferedIndexInput.java
```diff
@@ -154,65 +144,60 @@ public final void readBytes(byte[] b, int offset, int len, boolean useBuffer) th
...
       }
     }
   }
 
   @Override
   public final short readShort() throws IOException {
-    if (2 <= (bufferLength-bufferPosition)) {
-      return (short) (((buffer[bufferPosition++] & 0xFF) <<  8) |  (buffer[bufferPosition++] & 0xFF));
+    if (Short.BYTES <= buffer.remaining()) {
+      return buffer.getShort();
```
dweiss: Can't remember off the top of my head - the byte order is always guaranteed here, right? Or should we assert the buffer has the right one?

dweiss:
This looks like a very nice simplification to me.

lucene/core/src/java/org/apache/lucene/store/BufferedIndexInput.java
jpountz: Yes: "Specific byte orders are represented by instances of the ByteOrder class. The initial order of a byte buffer is always BIG_ENDIAN." (https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/nio/ByteBuffer.html)

jpountz: I think it's still a good call to add assertions to make sure that sub classes don't mess up with the ByteBuffer. I'll add some.

dweiss: +1. I checked the limit/remaining and other calls but the correctness there seems to be better verified by tests than by hand. :) Overall, like I said - looks like a nice simplification to me.

rmuir:
If we are replacing the `byte[]` with `ByteBuffer` here, can we try to add some safety at the same time? Can the actual buffer be `private final` and only passed to `readInternal()`? Otherwise, subclasses should only be able to see a read-only view (`.asReadOnlyBuffer`). The little bit of safety should be free, and since we are changing the API anyway, we might as well try to do it now?

rmuir:
Or maybe even simpler: just change `protected` to `private`. I'm not sure if any subclasses really need to mess with the buffer in that way anyway? If there is really a use-case, then we can try to introduce more sanity checks.

jpountz:
Good call.

lucene/core/src/java/org/apache/lucene/store/BufferedIndexInput.java
```diff
@@ -154,65 +146,60 @@ public final void readBytes(byte[] b, int offset, int len, boolean useBuffer) th
...
-        long after = bufferStart+bufferPosition+len;
+        long after = bufferStart+buffer.position()+len;
         if(after > length())
           throw new EOFException("read past EOF: " + this);
-        readInternal(b, offset, len);
+        readInternal(ByteBuffer.wrap(b, offset, len));
         bufferStart = after;
-        bufferPosition = 0;
-        bufferLength = 0;                    // trigger refill() on read
+        buffer.position(0);
```
danmuzi: I think if there is `buffer.limit(0)` on the next line, `buffer.position(0)` doesn't look necessary.
Because there is a code of initializing position value within `buffer.limit(0)`.
```java
// ByteBuffer.java
public abstract class ByteBuffer extends Buffer implements Comparable<ByteBuffer>
{
    ...
    ByteBuffer limit(int newLimit) {
        super.limit(newLimit);
        return this;
    }
    ...
}

// Buffer.java
public abstract class Buffer {
    ...
    public Buffer limit(int newLimit) {
        if (newLimit > capacity | newLimit < 0)
            throw createLimitException(newLimit);
        limit = newLimit;
        if (position > limit) position = limit; // position will be 0.
        if (mark > limit) mark = -1;
        return this;
    }
    ...
}
```

PR #1339, opened 2020-03-11
LUCENE-9272: Move checksum verification of the `.tip` file to `checkIntegrity()`.

jpountz:
See https://issues.apache.org/jira/browse/LUCENE-9272.


rmuir:
+1 for the change. We only did the check in that way because we had to read the bytes anyway on init before.

PR #1343, opened 2020-03-11
LUCENE-8103: Use two-phase iteration in Query- and DoubleValuesSource

micpalmia:
# Description
QueryValueSource (in "queries" module) is a ValueSource representation of a Query; the score is the value. It ought to try to use a TwoPhaseIterator from the query if it can be offered.

# Solution
Always check whether a TwoPhaseIterator can be offered, and use it. Since its `matches()` function may only be called once, while there is no restrictions on QueryValueSource's or DoubleValuesSource's method calling, I unfortunately had to add some state.

# Tests

For each value source, I added a test case that makes use of a query that serves a TwoPhaseIterator. I don't think that's the best solution, but was not sure about how to solve this differently.

# Checklist

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/DoubleValuesSource.java
+++ b/lucene/core/src/java/org/apache/lucene/search/DoubleValuesSource.java
@@ -604,8 +604,13 @@ public abstract class DoubleValuesSource implements SegmentCacheable {
       Scorer scorer = weight.scorer(ctx);
       if (scorer == null)
         return DoubleValues.EMPTY;
-      DocIdSetIterator it = scorer.iterator();
+      TwoPhaseIterator tpi = scorer.twoPhaseIterator();
+      DocIdSetIterator it = (tpi == null) ? scorer.iterator() : tpi.approximation();
+
       return new DoubleValues() {
+        int scorerDoc = it.docID();
+        boolean scorerMatch = tpi == null || (scorerDoc != -1 && tpi.matches());
+
         @Override
         public double doubleValue() throws IOException {
           return scorer.score();
@@ -613,9 +618,11 @@ public abstract class DoubleValuesSource implements SegmentCacheable {
 
         @Override
         public boolean advanceExact(int doc) throws IOException {
-          if (it.docID() > doc)
-            return false;
-          return it.docID() == doc || it.advance(doc) == doc;
+          if (scorerDoc < doc) {
+            scorerDoc = it.advance(doc);
+            scorerMatch = tpi==null || tpi.matches();
+          }
+          return scorerDoc == doc && (tpi == null  || scorerMatch);
         }
       };
     }
--- a/lucene/core/src/java/org/apache/lucene/search/DoubleValuesSource.java
+++ b/lucene/core/src/java/org/apache/lucene/search/DoubleValuesSource.java
@@ -608,25 +608,18 @@ public abstract class DoubleValuesSource implements SegmentCacheable {
       return new DoubleValues() {
         private final TwoPhaseIterator tpi = scorer.twoPhaseIterator();
         private final DocIdSetIterator disi = (tpi == null) ? scorer.iterator() : tpi.approximation();
-        private Boolean thisDocMatches = false;
 
         @Override
         public double doubleValue() throws IOException {
-          return thisDocMatches == Boolean.TRUE ? scorer.score() : Double.NaN;
+          return scorer.score();
         }
 
         @Override
         public boolean advanceExact(int doc) throws IOException {
           if (disi.docID() < doc) {
             disi.advance(doc);
-            thisDocMatches = null;
           }
-          if (disi.docID() == doc) {
-            if (thisDocMatches == null) {
-              thisDocMatches = tpi == null || tpi.matches();
-            }
-            return thisDocMatches;
-          } else return false;
+          return disi.docID() == doc && (tpi == null || tpi.matches());
         }
       };
     }
--- a/lucene/core/src/java/org/apache/lucene/search/DoubleValuesSource.java
+++ b/lucene/core/src/java/org/apache/lucene/search/DoubleValuesSource.java
@@ -612,7 +612,7 @@ public abstract class DoubleValuesSource implements SegmentCacheable {
 
         @Override
         public double doubleValue() throws IOException {
-          return (thisDocMatches != null && thisDocMatches) ? scorer.score() : Double.NaN;
+          return thisDocMatches == Boolean.TRUE ? scorer.score() : Double.NaN;
         }
 
         @Override
--- a/lucene/core/src/java/org/apache/lucene/search/DoubleValuesSource.java
+++ b/lucene/core/src/java/org/apache/lucene/search/DoubleValuesSource.java
@@ -608,22 +608,25 @@ public abstract class DoubleValuesSource implements SegmentCacheable {
       return new DoubleValues() {
         private final TwoPhaseIterator tpi = scorer.twoPhaseIterator();
         private final DocIdSetIterator disi = (tpi == null) ? scorer.iterator() : tpi.approximation();
-
-        private int scorerDoc = -1;
-        private boolean thisDocMatches = false;
+        private Boolean thisDocMatches = false;
 
         @Override
         public double doubleValue() throws IOException {
-          return thisDocMatches ? scorer.score() : Double.NaN;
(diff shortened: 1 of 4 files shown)


micpalmia:
Copying comments by @dsmiley here:

> minor: I suggest using the variable name "disi" for a DocIdSetIterator as this choice is quite common elsewhere.
> 
> In DoubleValueSource I think it'd be better to declare the state you need like tpi, disi inside the anonymous DoubleValues class to keep the scope smaller and to reduce extra hidden indirection that I think Java adds. Also declare final where applicable. scorerMatch boolean is confusing to me; and I think the initial value as written doesn't matter either. I suggest renaming this to thisDocMatches, initialize to false (since we are unpositioned at the start), and then update it appropriately. The logic you have now around scoreMatch seems to do maybe extra work and/or isn't as simple as it could be. If doubleValue is invoked and not thisDocMatches then don't delegate to the scorer, which is probably invalid. Return NaN.
> 
> In QueryValueSource.exists, maybe you could improve this a bit to be simpler; I find the way it was written prior to be confusing.

dsmiley:
BTW thanks for your contributions lately.  Perhaps you will be at Berlin Buzzwords this June; I shall be there.

micpalmia:
Thank you for your time and help! Fwiw, I appreciate it a lot. I'll be to Buzzwords too, hoping it's actually gonna happen.

`QueryValueSource.exists(int doc)` is probably looking more complicated than it could because it assumes docs can be requested out of order and because it keeps state on the availability of a scorer. I tried to make it tidier, but it's purely aesthetical.

lucene/core/src/java/org/apache/lucene/search/DoubleValuesSource.java
```diff
@@ -604,18 +604,26 @@ public DoubleValues getValues(LeafReaderContext ctx, DoubleValues scores) throws
...
         }
 
         @Override
         public boolean advanceExact(int doc) throws IOException {
-          if (it.docID() > doc)
-            return false;
-          return it.docID() == doc || it.advance(doc) == doc;
+          if (scorerDoc < doc) {
+            scorerDoc = disi.advance(doc);
+            thisDocMatches = tpi==null || tpi.matches();
```
dsmiley: If scorerDoc is > "this" doc (doc param), then we should not even compute tpi.matches(); it's wasted potentially expensive computation.  Maybe if thisDocMatches is a Object Boolean, we can differentiate an unkinown state.

Perhaps
```
if (scorerDoc < doc) {
  scorerDoc = disi.advance(doc);
  thisDocMatches = null;
}
if (scorerDoc == doc) {
 if (thisDocMatches == null) {
  thisDocMatches = tpi == null || tpi.matches();
 }
 return thisDocMatches;
} else {
  return thisDocMatches = false;
}
```

And I don't think we even need scorerDoc to be a field since it's disi.doc();  Local-variable is nice.
Hmm; if I'm not mistaken, the very first check (if scorerDoc < doc) can be replaced with an assertion since otherwise the caller is violating the API contract.

lucene/queries/src/java/org/apache/lucene/queries/function/valuesource/QueryValueSource.java
```diff
@@ -156,42 +136,35 @@ public float floatVal(int doc) {
...
 
       if (scorerDoc < doc) {
         scorerDoc = it.advance(doc);
-      }
-
-      if (scorerDoc > doc) {
-        // query doesn't match this document... either because we hit the
-        // end, or because the next doc is after this doc.
-        return false;
+        thisDocMatches = tpi == null || tpi.matches();
```
dsmiley: Again, lets not compute thisDocMatches if the scorerDoc doesn't even align.

lucene/core/src/java/org/apache/lucene/search/DoubleValuesSource.java
micpalmia: I've done as you suggested. I'm a bit concerned about the performance implications of the unboxing though - this implementation is transforming a Boolean to a boolean for every matching document. I can look at running performance tests if you think that would be important.

lucene/queries/src/java/org/apache/lucene/queries/function/valuesource/QueryValueSource.java
```diff
@@ -156,42 +135,38 @@ public float floatVal(int doc) {
...
+        if (scorer == null) {
           noMatches = true;
           return false;
         }
-        it = scorer.iterator();
+        tpi = scorer.twoPhaseIterator();
+        disi = tpi==null ? scorer.iterator() : tpi.approximation();
+        thisDocMatches = false;
       }
+      assert doc >= lastDocRequested : "values requested out of order; last=" + lastDocRequested + ", requested=" + doc;
```
dsmiley: Hmm; I wasn't sure about this but the more I look, it appears that the ValueSource API (here) also has a forward-only requirement.  A few years ago this was not the case.  So we can assume this and throw an exception.  See JoinDocFreqValueSource for an example.  I think this check you added here should be way up front at this method, thus clearly showing as a pre-condition, not down here.

lucene/core/src/java/org/apache/lucene/search/DoubleValuesSource.java
```diff
@@ -608,22 +608,25 @@ public DoubleValues getValues(LeafReaderContext ctx, DoubleValues scores) throws
...
-          return thisDocMatches ? scorer.score() : Double.NaN;
+          return (thisDocMatches != null && thisDocMatches) ? scorer.score() : Double.NaN;
         }
 
         @Override
         public boolean advanceExact(int doc) throws IOException {
-          if (scorerDoc < doc) {
-            scorerDoc = disi.advance(doc);
-            thisDocMatches = tpi==null || tpi.matches();
+          if (disi.docID() < doc) {
```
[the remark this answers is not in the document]
micpalmia: You mean something like the following?
```
assert disi.docID() < doc;
disi.advance(doc);
thisDocMatches = null;
if (disi.docID() == doc) { 
[...]
```

The disi can advance potentially much further than the requested `doc`, and that means that, for potentially many requested `doc`s afterwards ,`disi.docID() > doc` might hold. Tests fail for the edit above.

lucene/queries/src/java/org/apache/lucene/queries/function/valuesource/QueryValueSource.java
micpalmia: I also didn't think so until your last review, but I found the following on the docstring for `ValueSource.getValues`:

> The values must be consumed in a forward docID manner, and you must call this method again to iterate through the values again.

lucene/queries/src/java/org/apache/lucene/queries/function/valuesource/QueryValueSource.java
```diff
@@ -156,42 +135,38 @@ public float floatVal(int doc) {
...
-        scorerDoc = -1;
-        if (scorer==null) {
+        if (scorer == null) {
           noMatches = true;
           return false;
         }
-        it = scorer.iterator();
+        tpi = scorer.twoPhaseIterator();
+        disi = tpi==null ? scorer.iterator() : tpi.approximation();
+        thisDocMatches = false;
```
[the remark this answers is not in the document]
micpalmia: The disi is positioned on -1, so definitely not a match, no?

micpalmia:
> No auto-boxing concerns for Boolean :-)

thanks for the reassurance :) - do you mean the performance hit is too low to be of interest, or that there's no trouble with performances at all? Do you maybe have any references on this?

lucene/core/src/java/org/apache/lucene/search/DoubleValuesSource.java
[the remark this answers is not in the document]
dsmiley: Doh!  Of course; thanks.

lucene/queries/src/java/org/apache/lucene/queries/function/valuesource/QueryValueSource.java
```diff
@@ -139,15 +138,14 @@ public boolean exists(int doc) {
     lastDocRequested = doc;
 
     try {
-      if (noMatches) return false;
       if (scorer == null) {
```
dsmiley: This now needs to check if disi isn't null, otherwise we might call weight.scorer a ton!

micpalmia: 🤦🏻‍♂️

PR #1346, opened 2020-03-12
LUCENE-9276: Use same code-path for updateDocuments and updateDocument

s1monw:
Today we have a large amount of duplicated code that is rather of
complex nature. This change consolidates the code-paths to always
use the updateDocuments path.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/DocumentsWriter.java
+++ b/lucene/core/src/java/org/apache/lucene/index/DocumentsWriter.java
@@ -474,13 +474,6 @@ final class DocumentsWriter implements Closeable, Accountable {
     return seqNo;
   }
 
-
-  long updateDocument(final Iterable<? extends IndexableField> doc, final Analyzer analyzer,
-                      final DocumentsWriterDeleteQueue.Node<?> delNode) throws IOException {
-
-    return updateDocuments(List.of(doc), analyzer, delNode);
-  }
-
   private boolean doFlush(DocumentsWriterPerThread flushingDWPT) throws IOException {
     boolean hasEvents = false;
     while (flushingDWPT != null) {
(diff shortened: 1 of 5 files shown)


s1monw:
@mikemccand @dnhatn wanna take a look

s1monw:
@uschindler I would love to get your input here too

lucene/core/src/java/org/apache/lucene/index/DocumentsWriterPerThread.java
```diff
@@ -346,7 +285,7 @@ public long updateDocuments(Iterable<? extends Iterable<? extends IndexableField
     }
   }
   
-  private long finishDocument(DocumentsWriterDeleteQueue.Node<?> deleteNode) {
+  private long finishDocument(DocumentsWriterDeleteQueue.Node<?> deleteNode, int docCount) {
```
dnhatn: Should we call this finishDocument**S** ?

lucene/core/src/java/org/apache/lucene/index/DocumentsWriter.java
```diff
@@ -474,49 +474,11 @@ long updateDocuments(final Iterable<? extends Iterable<? extends IndexableField>
     return seqNo;
   }
 
+
   long updateDocument(final Iterable<? extends IndexableField> doc, final Analyzer analyzer,
```
dnhatn: Can we also remove this method and delegate `updateDocument` to `updateDocuments` in IndexWriter instead?

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
```diff
@@ -1655,7 +1655,7 @@ private long updateDocument(final DocumentsWriterDeleteQueue.Node<?> delNode,
     ensureOpen();
     boolean success = false;
     try {
-      final long seqNo = maybeProcessEvents(docWriter.updateDocument(doc, analyzer, delNode));
+      final long seqNo = maybeProcessEvents(docWriter.updateDocuments(List.of(doc), analyzer, delNode));
```
dnhatn: Can we just delegate this method to `IndexWriter#updateDocuments`?

PR #1348, opened 2020-03-13
SOLR-14326 Number of tlog replicas off by one when restoring collections

CBWallaby:
https://issues.apache.org/jira/browse/SOLR-14326

When making a request to restore a collection, the quantity of tlog replicas will always be off by one when restoring a collection that doesn't contain nrt replicas or when specifying the quantity of replicas in the request itself.

This is due to a flawed comparison where an int meant to be an iterator for tlog replicas is checked if it is greater than zero, however, since that variable was initialized as 0 just prior it will never be greater than zero. The fix is to compare the desired number of tlog replicas (like nrt) rather than the iterator.


madrob:
Thanks for finding this and taking the time to post a patch! Could you add a test to demonstrate that this fixes the issue and prevent future regressions? Also, can you generate a PR against master and then we will back port from there? I think we are unlikely to see any more releases on the 7.x branch in the future.

PR #1351, opened 2020-03-13
LUCENE-9280: Collectors to skip noncompetitive documents

mayya-sharipova:
Similar how scorers can update their iterators to skip non-competitive
documents, collectors and comparators should also provide and update
iterators that allow them to skip non-competive documents

This could be useful if we want to sort by some field.


mayya-sharipova:
@jimczi I have created a draft PR for comparators and collectors to skip non-competitive docs.  Can you please have a look at it and see if we are happy with this approach.

lucene/core/src/java/org/apache/lucene/search/Weight.java
```diff
@@ -249,6 +249,10 @@ static int scoreRange(LeafCollector collector, DocIdSetIterator iterator, TwoPha
      *  See <a href="https://issues.apache.org/jira/browse/LUCENE-5487">LUCENE-5487</a> */
     static void scoreAll(LeafCollector collector, DocIdSetIterator iterator, TwoPhaseIterator twoPhase, Bits acceptDocs) throws IOException {
       if (twoPhase == null) {
+        if (collector.iterator() != null) {
```
jimczi: I think we should try to handle the collector iterator early in the chain. The default BulkScorer can be overridden so I wonder how this would look like if we build a conjunction from the main query and the collector iterator directly when building the weight (in IndexSearcher) ?

lucene/core/src/java/org/apache/lucene/search/LongDocValuesPointComparator.java
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    @Override
+    public void setScorer(Scorable scorer) throws IOException {}
+
+    public DocIdSetIterator iterator() {
+        return iterator;
+    }
+
+    public void updateIterator() throws IOException {
```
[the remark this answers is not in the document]
mayya-sharipova: Addressed in 6384b15

lucene/core/src/java/org/apache/lucene/search/LongDocValuesPointComparator.java
```diff
@@ -0,0 +1,164 @@
...
+        final byte[] maxValueAsBytes = new byte[Long.BYTES];
+        final byte[] minValueAsBytes = new byte[Long.BYTES];
+        if (reverse == false) {
+            LongPoint.encodeDimension(bottom, maxValueAsBytes, 0);
+        } else {
+            LongPoint.encodeDimension(bottom, minValueAsBytes, 0);
+        };
+
```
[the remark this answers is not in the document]
mayya-sharipova: Addressed in 6384b15

lucene/core/src/java/org/apache/lucene/search/LongDocValuesPointComparator.java
```diff
@@ -0,0 +1,164 @@
...
+                if ((reverse == false) && (Arrays.compareUnsigned(minPackedValue, 0, Long.BYTES, maxValueAsBytes, 0, Long.BYTES) > 0)) {
+                   return PointValues.Relation.CELL_OUTSIDE_QUERY;
+                }
+                if ((reverse == true) && (Arrays.compareUnsigned(maxPackedValue, 0, Long.BYTES, minValueAsBytes, 0, Long.BYTES) < 0)) {
+                    return PointValues.Relation.CELL_OUTSIDE_QUERY;
+                }
+                return PointValues.Relation.CELL_CROSSES_QUERY;
+            }
+        };
+        pointValues.intersect(visitor);
```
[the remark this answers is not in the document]
mayya-sharipova: Addressed in 6384b15

lucene/core/src/java/org/apache/lucene/search/FieldComparator.java
```diff
@@ -928,4 +928,9 @@ public int compareTop(int doc) throws IOException {
     @Override
     public void setScorer(Scorable scorer) {}
   }
+
+  public static abstract class IteratorSupplierComparator<T> extends FieldComparator<T> implements LeafFieldComparator {
+    abstract DocIdSetIterator iterator();
+    abstract void updateIterator() throws IOException;
```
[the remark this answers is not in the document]
mayya-sharipova: Indeed it is more straightforward to just update an iterator in `setBottom` function of a comparator.

But I was thinking it is better to have a special function for two reasons:
1)  After updating an iterator, in `TopFieldCollector` we need to change 
`totalHitsRelation = TotalHits.Relation.GREATER_THAN_OR_EQUAL_TO;`

2)  we also need to check `hitsThresholdChecker.isThresholdReached()`, and passing not strictly related object `hitsThresholdChecker` to a comparator's constructor doesn't look nice to me. 
Please let me know if you think otherwise

lucene/core/src/java/org/apache/lucene/search/Weight.java
mayya-sharipova: @jimczi Thanks for the initial review.   

>  I wonder how this would look like if we build a conjunction from the main query and the collector iterator directly when building the weight (in IndexSearcher) 

I could not think of any clever way  to do this in `IndexSearcher`,  I would appreciate your help if you can suggest any such way.

I looked at classes that override `BulkScorer` and many of them still refer to a default `BulkScorer`, and for those that don't such as `BooleanScorer` I found its logic to be too complex to combine with  a collector's iterator.

lucene/core/src/java/org/apache/lucene/search/Weight.java
```diff
@@ -201,20 +201,54 @@ public long cost() {
...
-          if (twoPhase == null) {
-            doc = iterator.advance(min);
-          } else {
-            doc = twoPhase.approximation().advance(min);
+        if (doc < min) scorerIterator.advance(min);
+        return scoreRange(collector, combinedIterator, twoPhase, acceptDocs, doc, max);
+      }
+    }
+
+    // conjunction iterator between scorer's iterator and collector's iterator
```
jimczi: you can replace this with `ConjunctionDISI#intersectIterators` ?

lucene/core/src/java/org/apache/lucene/search/Weight.java
```diff
@@ -223,44 +257,23 @@ public int score(LeafCollector collector, Bits acceptDocs, int min, int max) thr
...
-          }
-          currentDoc = approximation.nextDoc();
+      while (currentDoc < end) {
+        if ((acceptDocs == null || acceptDocs.get(currentDoc)) && (twoPhase == null || twoPhase.matches())) {
+          collector.collect(currentDoc);
         }
-        return currentDoc;
+        currentDoc = iterator.nextDoc();
       }
+      return currentDoc;
```
jimczi: this change is not required ? I see hotspot in the javadoc comment above so  we shouldn't touch it if it's not  required ;).

lucene/core/src/java/org/apache/lucene/search/Weight.java
```diff
@@ -201,20 +201,54 @@ public long cost() {
...
+        scoreAll(collector, combinedIterator, twoPhase, acceptDocs);
         return DocIdSetIterator.NO_MORE_DOCS;
       } else {
         int doc = scorer.docID();
-        if (doc < min) {
-          if (twoPhase == null) {
-            doc = iterator.advance(min);
-          } else {
-            doc = twoPhase.approximation().advance(min);
+        if (doc < min) scorerIterator.advance(min);
```
jimczi:
```suggestion
        if (doc < min) {
          doc = combinedIterator.advance(min);
        }
```
?

lucene/core/src/java/org/apache/lucene/search/FieldComparator.java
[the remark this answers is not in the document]
jimczi: For 1. we could set the totalHitsRelation when we reach the total hits threshold in the TOP_DOCS mode ? 
For 2. I wonder if we could pass the hitsThresholdChecker to the LeafFieldComparator like we do for the scorer ?
This way we can update the iterator internally when a new bottom is set or when `compareBottom` is called ?

lucene/core/src/java/org/apache/lucene/search/Weight.java
mayya-sharipova: Addressed in https://github.com/apache/lucene-solr/pull/1351/commits/d732d7eb9de67a597f67e91c9774104aa055e293

lucene/core/src/java/org/apache/lucene/search/Weight.java
mayya-sharipova: Addressed in https://github.com/apache/lucene-solr/pull/1351/commits/d732d7eb9de67a597f67e91c9774104aa055e293

lucene/core/src/java/org/apache/lucene/search/Weight.java
mayya-sharipova: https://github.com/apache/lucene-solr/pull/1351/commits/d732d7eb9de67a597f67e91c9774104aa055e293

lucene/core/src/java/org/apache/lucene/search/LongDocValuesPointComparator.java
```diff
@@ -0,0 +1,208 @@
...
+        maxDocVisited = 0;
+        return this;
+    }
+
+    @Override
+    public void setBottom(int slot) throws IOException {
+        this.bottom = values[slot];
+        // can't use hitsThresholdChecker.isThresholdReached() as it uses > numHits,
+        // while we want to update iterator as soon as threshold reaches numHits
+        if (hitsThresholdChecker != null && (hitsThresholdChecker.getHitsThreshold() >= numHits)) {
```
mayya-sharipova: I am not very happy about this change because of 2 reasons: 
1) We  can't use `hitsThresholdChecker.isThresholdReached` as it checks for greater than numHits, but we need to check starting with equal, as if there are no competitive docs later `setBottom` will not be called.
2) totalHitsRelation may not end up to be set to TotalHits.Relation.GREATER_THAN_OR_EQUAL_TO, as we set it only when we have  later competitive hits. 

I will keep thinking how to organize it better

lucene/core/src/java/org/apache/lucene/search/FieldComparator.java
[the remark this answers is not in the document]
mayya-sharipova: @msokolov  Thanks for the suggestion, naming is tough, addressed in 95e1bc1.

lucene/core/src/java/org/apache/lucene/search/LongDocValuesPointComparator.java
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    // update its iterator to include possibly only docs that are "stronger" than the current bottom entry
+    public void updateIterator() throws IOException {
+        updateCounter++;
+        if (updateCounter > 256 && (updateCounter & 0x1f) != 0x1f) { // Start sampling if we get called too much
+            return;
+        }
+
+        final byte[] maxValueAsBytes = reverse == false ? new byte[Long.BYTES] : hasTopValue ? new byte[Long.BYTES]: null;
```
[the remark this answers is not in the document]
mayya-sharipova: @msokolov  Thanks for the suggestion, indeed these values can be initialized in the comparator's constructor. As each topfieldcollector has its own comparator and processes segments sequentially, these values should be shareable.  Addressed in 95e1bc1

mayya-sharipova:
I have run some benchmarking using `luceneutil`.
As the new sort optimization uses a new `LongDocValuesPointSortField` that is not present in `luceneutil`, I had to hack `luceneutil` as follows:

1. I added a  sort task on a long field `TermDateTimeSort`  to `wikimedium.1M.nostopwords.tasks` . This task was present in `wikinightly.tasks` , but was not able for wikimedium 1M and 10M tasks
2. I indexed the corresponding field `lastModNDV` as `LongPoint` as well. It was only indexed as `NumericDocValuesField` before, but for the sort optimization we need long values to be indexed both as docValues and as points.
3. I modified `SearchTask.java` to have `TopFieldCollector` with `totalHitsThreshold` set to `topK`: `final TopFieldCollector c = TopFieldCollector.create(s, topN, null, topN);`   Sort optimization only works when we set total hits threshold.
4. For the patch version , I modified sort in `TaskParser.java`. Instead of `lastModNDVSort = new Sort(new SortField("lastModNDV", SortField.Type.LONG));`  I useed the optimized sort: `lastModNDVSort = new Sort(new LongDocValuesPointSortField("lastModNDV"));`

Here the main point of comparison is `TermDTSort` as it is the only sort on long field. Other sorts are presented to demonstrate a possible regression or absence on them.

---
wikimedium1m
| TaskQPS               | baseline QPS | StdDevQPS | my_modified_version QPS | StdDevQPS |
| --------------------- | -----------: | --------: | ----------------------: | --------: |
| **TermDTSort**        |       507.20 |   (11.2%) |                  550.02 |   (16.1%) |
| HighTermMonthSort     |       550.06 |   (10.4%) |                  443.69 |   (16.1%) |
| HighTermDayOfYearSort |       105.62 |   (24.9%) |                   91.93 |   (22.1%) |
---
wikimedium10m
| TaskQPS               | baseline QPS | StdDevQPS | my_modified_version QPS | StdDevQPS |
| --------------------- | -----------: | --------: | ----------------------: | --------: |
| **TermDTSort**        |       147.64 |   (11.5%) |                  547.80 |    (6.6%) |
| HighTermMonthSort     |       147.85 |   (12.2%) |                  239.28 |    (7.3%) |
| HighTermDayOfYearSort |        74.44 |    (7.7%) |                   42.56 |   (12.1%) |

For wikimedium1m using `LongDocValuesPointSortField` doesn't seem to have much effect. As probably in this index segments are smaller, and probably optimization was completely skipped on those segments.
For wikimedium10m using `LongDocValuesPointSortField`  instead of usual `SortField.Type.LONG` **brings about 3x speedups**.
There is so regression/speedups for the sort tasks of HighTermMonthSort and HighTermDayOfYearSort, which I don't know the reason why, as they should not be effected.

msokolov:
That 3x speedup is very nice! My experience with these benchmarks is they can be pretty noisy, maybe accounting for the regressions? I tend to increase comp.taskRepeatCount = 500. I'd also be interested to see how this optimization fares for higher values of topN - I think the default is 10, but you can edit in benchUtil.py. You did not sort the index right (eg: comp.newIndex('baseline', sourceData, facets=facets, indexSort='lastModNDV:long', addDVFields=True)? It would be interesting to see if this has the same impact for sorted index, large N, especially running with an executor (.competitor(...concurrentSearchers = True ).

mayya-sharipova:
@msokolov Thank for suggesting additional benchmarks that we can use.
Below are the results on the dataset `wikimedium10m`.

First I will repeat the results from the previous round of benchmarking:

topN=10, taskRepeatCount = 20, concurrentSearchers = False

| TaskQPS               | baseline QPS | StdDevQPS | my_modified_version QPS | StdDevQPS |
| --------------------- | -----------: | --------: | ----------------------: | --------: |
| **TermDTSort**        |       147.64 |   (11.5%) |                  547.80 |    (6.6%) |
| HighTermMonthSort     |       147.85 |   (12.2%) |                  239.28 |    (7.3%) |
| HighTermDayOfYearSort |        74.44 |    (7.7%) |                   42.56 |   (12.1%) |



---
  topN=10, **taskRepeatCount = 500**, concurrentSearchers = False
| TaskQPS               | baseline QPS | StdDevQPS | my_modified_version QPS | StdDevQPS |
| --------------------- | -----------: | --------: | ----------------------: | --------: |
| **TermDTSort**        |       184.60 |    (8.2%) |                 3046.19 |    (4.4%) |
| HighTermMonthSort     |       209.43 |    (6.5%) |                  253.90 |   (10.5%) |
| HighTermDayOfYearSort |       130.97 |    (5.8%) |                   73.25 |   (11.8%) |

This seemed to speed up all operations, and here the speedups for `TermDTSort` even bigger: 16.5x times. There is also seems to be more regression for `HighTermDayOfYearSort`.

---
  **topN=500**,  taskRepeatCount = 20, concurrentSearchers = False


| TaskQPS               | baseline QPS | StdDevQPS | my_modified_version QPS | StdDevQPS |
| --------------------- | -----------: | --------: | ----------------------: | --------: |
| **TermDTSort**        |       210.24 |    (9.7%) |                  537.65 |    (6.7%) |
| HighTermMonthSort     |       116.02 |    (8.9%) |                  189.96 |   (13.5%) |
| HighTermDayOfYearSort |        42.33 |    (7.6%) |                   67.93 |    (9.3%) |

With increased `topN` the sort optimization has less speedups up to 2x, as it is expected as it will be possible to run it only after collecting `topN` docs.

---
topN=10, taskRepeatCount = 20, **concurrentSearchers = True**
| TaskQPS               | baseline QPS | StdDevQPS | my_modified_version QPS | StdDevQPS |
| --------------------- | -----------: | --------: | ----------------------: | --------: |
| **TermDTSort**        |       132.09 |   (14.3%) |                  287.93 |   (11.8%) |
| HighTermMonthSort     |       211.01 |   (12.2%) |                  116.46 |    (7.1%) |
| HighTermDayOfYearSort |        72.28 |    (6.1%) |                   68.21 |   (11.4%) |

With the concurrent searchers the speedups are also smaller up to 2x. This is expected as now segments are spread between several TopFieldCollects/Comparators and they don't exchange bottom values.  As a follow-up on this PR, we can think how we can have a global bottom value similar how `MaxScoreAccumulator` is used to set up a global competitive min score.

lucene/core/src/java/org/apache/lucene/search/LongDocValuesPointComparator.java
```diff
@@ -0,0 +1,210 @@
...
+    private final long[] values;
+    private long bottom;
+    private long topValue;
+    protected NumericDocValues docValues;
+    private DocIdSetIterator iterator;
+    private PointValues pointValues;
+    private int maxDoc;
+    private int maxDocVisited;
+    private int updateCounter = 0;
+    private byte[] cmaxValueAsBytes = null;
```
msokolov: Can these be final, and allocated only in the constructor? I think it might be clearer to add a boolean "hasTopValues" and set that in setTopValue, rather than use the existence of these byte[]? Then you could make these final and eliminate the local variables where they get copied below

mayya-sharipova:
@msokolov Thank you for an additional  review.  I realized I ran benchmarks incorrectly, not indexing documents with docValues. Sorry, I am still learning lucene benchmarking tool.  Please disregard the previous benchmarking results, I will be rerunning them.

msokolov:
@mayya-sharipova sounds good - I'd also encourage you to post a PR with your modifications to luceneutil

mayya-sharipova:
@msokolov  Sorry again for reporting incorrect benchmarking results. Below are are my latest results, and I feel  quite confident in their correctness.

First about the benchmarking setup.
1. [Here](https://github.com/mayya-sharipova/luceneutil/commit/e0d86b24053cc8a68796abd9f0fd08dbac899779)  are the changes made to `luceneutil`
2.  `patch` folder is checkout as this PR
3. `trunk` folder is checkout as this PR as well with a modification.  As there is no `LongDocValuesPointSortField` in master, I can't benchmark [sorting using this field](https://github.com/mayya-sharipova/luceneutil/commit/e0d86b24053cc8a68796abd9f0fd08dbac899779#diff-58e50bb4a8f0be480df656bcd84d5b77R76) on master. What I did is just is on `trunk` 
folder delegated sorting to the traditional sorting on a long field like this:
```java
public class LongDocValuesPointSortField extends SortField {
    public LongDocValuesPointSortField(String field) {
        super(field, SortField.Type.LONG);
    }
    public LongDocValuesPointSortField(String field, boolean reverse) {
        super(field, SortField.Type.LONG, reverse);
    }
}
```
So basically I was benchmarking a traditional long sort VS a long sort using a new field `LongDocValuesPointSortField`.


wikimedium10m: 10 millon docs, up to 2x speedups

```
 TaskQPS                     baseline   StdDevQPS     patch     StdDev    Pct diff
             TermDTSort       64.53      (6.4%)      155.29     (42.3%)  140.7% (  86% -  202%)
  HighTermDayOfYearSort       47.63      (5.4%)       50.47      (6.8%)    6.0% (  -5% -   19%)
       HighTermMonthSort      110.07     (7.3%)      121.13      (6.8%)   10.0% (  -3% -   26%)
WARNING: cat=TermDTSort: hit counts differ: 754451 vs 1669+
```

wikimediumall: about 33 million docs, up to 3.5 x speedups
```
 TaskQPS                     baseline   StdDevQPS     patch     StdDev    Pct diff
              TermDTSort       28.96      (4.3%)      108.45     (56.9%)  274.5% ( 204% -  350%)
   HighTermDayOfYearSort        9.69      (5.1%)        9.56      (6.1%)   -1.3% ( -11% -   10%)
       HighTermMonthSort       39.41      (4.7%)       47.99     (10.0%)   21.8% (   6% -   38%)
WARNING: cat=TermDTSort: hit counts differ: 1474717 vs 1070+
```

Please let me know if these results and methodology make sense.

lucene/core/src/java/org/apache/lucene/search/LeafCollector.java
```diff
@@ -93,4 +93,11 @@
    */
   void collect(int doc) throws IOException;
 
+  /*
```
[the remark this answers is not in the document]
mayya-sharipova: Addressed in 6c628f7

lucene/core/src/java/org/apache/lucene/search/LeafCollector.java
```diff
@@ -93,4 +93,11 @@
...
+   * optionally returns an iterator over competitive documents
```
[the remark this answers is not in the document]
mayya-sharipova: Thanks @jpountz 

> It's probably worth making explicit as null iterators are elsewhere interpreted as matching no documents

What is the way to make this explicit?

mayya-sharipova:
@jpountz  Thank you for the review.

> I wonder whether we could make it easier to write implementations. I haven't spent much time thinking about it, but for instance would it be possible to wrap existing comparators to add the skipping functionality? Alternatively we could add the skipping logic to the existing comparators, but the fact that Lucene doesn't require that the same data be stored in indexes and doc values makes me a bit nervous about enabling it by default, and I'd like to avoid adding a new constructor argument.

Would it make sense  for each numeric FieldComparator to add an extra class that would wrap a numeric comparator and provide additional methods for skipping logic (getting an iterator and updating an iterator)?

mayya-sharipova:
@jpountz  What do you think of this design in eeb23c11?

1. `IterableFieldComparator` wraps an `FieldComparator` to provide skipping functionality. All numeric  comparators are wrapped in corresponding iterable comparators.
2.  `SortField` has a new method  `allowSkipNonCompetitveDocs`, that if set will use a comparator that provided skipping functionality.

In this case, we would not need other classes that I previously introduced `LongDocValuesPointComparator` and  `LongDocValuesPointSortField`.

romseygeek:
I like the idea of wrapping things up, and I think we may be able to take this further by pushing more of the logic into the comparator:
* add a `wrapDocIdSetIterator(DocIdSetIterator in)` method to `LeafCollector` that by default returns the passed-in iterator.  This gets called in `DefaultBulkScorer#score` to wrap the iterator for a query.
* add a `wrapDocIdSetIterator(DocIdSetIterator in)` method to `FieldComparator` that by default returns the passed-in iterator.  `TopFieldCollector` delegates its `wrapDocIdSetIterator` method to this method on its first comparator.  This allows us to completely contain the logic that combines a query's iterator with sorting shortcuts to the `SortField` and associated `FieldComparator` implementation.
* Move the logic that checks whether or not to update the iterator into `setBottom` on the leaf comparator.  I know this involves passing the `HitsThresholdChecker` into the leaf comparator constructor, but I think that's reasonable if the point of this API change is to make it possible for comparators to skip hits

mayya-sharipova:
@romseygeek Thank you for the review and suggestions, I will work on them.

mayya-sharipova:
@romseygeek Thanks for the feedback. I have addressed your comments 1 and 2 in 89d241e.   Indeed, the APIs look simpler, I like them more now.  I just renamed `wrapDocIdSetIterator` to `filterIterator`. 

The comment 3 is challenging to address.  I have already tried to do this in d732d7eb9 as a response to @jimczi  feedback, but I reverted this commit because of those challenges.  `TopFieldCollector` has a lot of subtle logic that makes it difficult to reason and imitate in other classes.  The challenges are following:

1. `HitsThresholdChecker`. First we are passing a not strictly related class `hitsThresholdChecker` to `LeafComparator`.  Secondly, it turned out that we can't use `hitsThresholdChecker.isThresholdReached` method in `setBottom` as it starts to return `true` only after we have already collected  hits  more than `numHits`, but in `setBottom` we need to update an iterator as as soon as we have collected `numHits`, because if there are no competitive docs later `setBottom` will never be called again.

2. `TotalHitsRelation`.  If we end up updating the iterator, we need to set it to `TotalHits.Relation.GREATER_THAN_OR_EQUAL_TO` and it is not clear to me when this should be set.

3. If we have a parallel collector and would like to update a global bottom, it is not clear to me how to do this with this model as well.

I guess I need to think more about it.

mayya-sharipova:
@romseygeek  I have tried to address your outstanding feedback in 4448499f0f.  Can you please continue the review when you have time?

> Move the logic that checks whether or not to update the iterator into setBottom on the leaf comparator.

In the new `FilteringFieldComparator` class, the iterator is updated in
- setBottom
- when we change a segment in `getLeafComparator`, so that we can also update iterators of subsequent segments. 
- and also when for the first time queue becomes full and hitsThreshold is reached in `setCanUpdateIterator`, this method is called from a collector.

lucene/core/src/java/org/apache/lucene/search/TopFieldCollector.java
```diff
@@ -302,20 +312,27 @@ private TopFieldCollector(FieldValueHitQueue<Entry> pq, int numHits,
...
       canSetMinScore = true;
     } else {
-      firstComparator = null;
+      relevanceComparator = null;
       scoreMode = needsScores ? ScoreMode.COMPLETE : ScoreMode.COMPLETE_NO_SCORES;
       canSetMinScore = false;
     }
     this.minScoreAcc = minScoreAcc;
+
+    if ((firstComparator instanceof FilteringFieldComparator) && (hitsThresholdChecker.getHitsThreshold() != Integer.MAX_VALUE)) {
```
romseygeek, line 330: Can we merge this with the  `if` statement immediately above it? That way `scoreMode` can stay final

lucene/core/src/test/org/apache/lucene/search/TestSortOptimization.java
```diff
@@ -0,0 +1,160 @@
...
+    for (int i = 0; i < numDocs; ++i) {
+      final Document doc = new Document();
+      doc.add(new NumericDocValuesField("my_field", i));
+      doc.add(new LongPoint("my_field", i));
+      writer.addDocument(doc);
+    }
+    final IndexReader reader = DirectoryReader.open(writer);
+    IndexSearcher searcher = new IndexSearcher(reader);
+    final SortField sortField = new SortField("my_field", SortField.Type.LONG);
+    sortField.allowFilterNonCompetitveDocs();
```
romseygeek, line 48: Can we avoid having to set this explicitly? The fact that we're setting a totalHitsThreshold on the collector should be enough to tell the comparator that we're enabling skipping.  Maybe have a `SkippableSortField` implementation that wraps an existing SortField and overrides `getComparator()`; then in `TopFieldCollector.create()` we check if totalHitsThreshold is set and if so rebuild the sort

lucene/core/src/java/org/apache/lucene/search/ScoreMode.java
```diff
@@ -51,10 +59,51 @@ public boolean needsScores() {
...
   /**
    * Whether this {@link ScoreMode} needs to compute scores.
    */
   public abstract boolean needsScores();
+
+  /**
+   * Returns {@code true} if for this {@link ScoreMode} it is necessary to process all documents,
+   * or {@code false} if is enough to go through top documents only.
+   */
+  public abstract boolean isExhaustive();
```
romseygeek, line 108: Rather than defining these as abstract, maybe have two final fields set in a constructor?  Then the definitions become much less verbose: `TOP_DOCS_WITH_SCORES(true, false);`

lucene/core/src/java/org/apache/lucene/search/TopFieldCollector.java
mayya-sharipova: @romseygeek  Addressed in 719882e

lucene/core/src/java/org/apache/lucene/search/ScoreMode.java
mayya-sharipova: @romseygeek Thanks, a very good suggestion, this indeed made the code more clear and less verbose.  Addressed in 719882e

lucene/core/src/test/org/apache/lucene/search/TestSortOptimization.java
mayya-sharipova: @romseygeek  Thanks for the suggestion.  It is difficult to wrap `SortField` into another class that extends `SortField`, as `SortField` is a concrete class with several constructors. 

Instead, I  am setting `sort.fields[0].allowFilterNonCompetitveDocs()` when we create a `TopFieldCollector` with a set hitsTreshold.   

 Partially addressed in 719882e

romseygeek: I'm a bit wary of this, as I don't think we should be mutating objects passed in to the API that may well be shared (it's already a bit sketchy that SortField doesn't have final fields, mainly due to how the constructors are implemented - also the case for `Sort`, which I think we should make immutable as a follow-up).  There are several specialised implementations of SortField elsewhere in the codebase, so I don't think it should be too hard to create a new SkippingSortField that takes another SortField in its constructor and copies the relevant field, type, etc data from that?

lucene/core/src/java/org/apache/lucene/search/ScoreMode.java
```diff
@@ -24,37 +24,53 @@
...
+   * to optionally allow to skip for non-competitive docs
+   */
+  TOP_DOCS(false, false),
+
+  /**
+   * ScoreMode for top field collectors that can provide their own iterators,
+   * to optionally allow to skip for non-competitive docs.
+   * This mode is used when there is a secondary sort by _score.
+   */
+  TOP_DOCS_WITH_SCORES(false, true);
```
romseygeek, line 52: Do we need to distinguish between `TOP_SCORES` and `TOP_DOCS_WITH_SCORES`?

lucene/core/src/java/org/apache/lucene/search/TopFieldCollector.java
```diff
@@ -30,6 +30,7 @@
 import org.apache.lucene.search.MaxScoreAccumulator.DocAndScore;
 import org.apache.lucene.search.TotalHits.Relation;
 
+
```
romseygeek, line 33: nit: extra line

lucene/core/src/java/org/apache/lucene/search/TopFieldCollector.java
```diff
@@ -432,6 +462,12 @@ static TopFieldCollector create(Sort sort, int numHits, FieldDoc after,
       throw new IllegalArgumentException("hitsThresholdChecker should not be null");
     }
 
+    // here we assume that if hitsThreshold was set, we let the corresponding comparator to skip non-competitive docs
+    // It is beneficial for the 1st field only to skip non-competitive docs
+    if (hitsThresholdChecker.getHitsThreshold() != Integer.MAX_VALUE) {
+      sort.fields[0].allowFilterNonCompetitveDocs();
+    }
+
```
romseygeek, line 470: I think this should be done in `FieldValueHitQueue.create()` below - pass in a boolean to indicate that we want to skip hits if we can, and then do the wrapping/rewriting/whatever it ends up being as a part of building the sort queue

lucene/test-framework/src/java/org/apache/lucene/search/AssertingLeafCollector.java
```diff
@@ -50,5 +50,9 @@ public void collect(int doc) throws IOException {
     lastCollected = doc;
   }
 
+  @Override
+  public DocIdSetIterator filterIterator(DocIdSetIterator scorerIterator) {
+    return super.filterIterator(scorerIterator);
```
romseygeek, line 55: Is this necessary?

mayya-sharipova: @romseygeek Good notice, not it is not necessary. Removed in c84fe5e

lucene/core/src/java/org/apache/lucene/search/ScoreMode.java
mayya-sharipova: Yes, I think this distinction is necessary, if there is an additional sort field on `_score`.
`Weight` and `Scorer` classes check collector's scoreMode for `needsScores()`, and will not calculate scores if `needsScores` is false.  I have added [an additional test](https://github.com/apache/lucene-solr/pull/1351/commits/c84fe5ed90b866dcbe41979dc6ba2ce2a874d59b#diff-cade050face1b04d4f21171fb26a319eR84)  to check that scores computed .

lucene/core/src/java/org/apache/lucene/search/TopFieldCollector.java
mayya-sharipova: @romseygeek  Great suggestion to move wrapping into `FieldValueHitQueue`. I like the APIs more and more now.  Addressed in c84fe5e.
And also looks like we don't even need any changes to `SortField`, we can just wrap comparators.

lucene/core/src/java/org/apache/lucene/search/LeafCollector.java
```diff
@@ -93,4 +93,16 @@
...
+   * Optionally creates a view of the scorerIterator where only competitive documents
+   * in the scorerIterator are kept and non-competitive are skipped.
+   *
+   * Collectors should delegate this method to their comparators if
+   * their comparators provide the skipping functionality over non-competitive docs.
+   * The default is to return the same iterator which is interpreted as the collector doesn't filter any documents.
+   */
+  default DocIdSetIterator filterIterator(DocIdSetIterator scorerIterator) {
+    return scorerIterator;
+  }
```
jpountz, line 106: This allows for some hacks like returning an iterator that matches more hacks than the scorer. I liked the previous approach that returned an iterator better.

lucene/core/src/java/org/apache/lucene/search/ScoreMode.java
romseygeek: But `TOP_SCORES` and `TOP_DOCS_WITH_SCORES` have identical `needsScores()` and `isExhaustive()` values, so I'm not sure why we need both?

lucene/core/src/java/org/apache/lucene/search/LeafCollector.java
romseygeek: Oh that's a good point. +1 to just return an iterator based on the comparator, and do the conjuncion/combination in `BulkScorer`

mayya-sharipova: Thanks, makes sense. Addressed in d7ef9b6

mayya-sharipova:
I have run another round of benchmarks, this time comparing the performance of this PR VS master as we don't need any special sort field.  [Here](https://github.com/mayya-sharipova/luceneutil/commit/c3166e4fc44e7fcddcd1672112c96364d9f464e5) are the changes made to luceneutil.


**wikimedium10m**
```
 TaskQPS                     baseline   StdDevQPS     patch     StdDev    Pct diff
   HighTermDayOfYearSort       50.93      (5.6%)       49.31     (10.9%)   -3.2% ( -18% -   14%)
              TermDTSort       83.37      (5.9%)      129.95     (41.2%)   55.9% (   8% -  109%)
WARNING: cat=HighTermDayOfYearSort: hit counts differ: 541957 vs 541957+
WARNING: cat=TermDTSort: hit counts differ: 506054 vs 1861+
```
Here we have two sorts:
-  Int sort on a day of year. Slight decrease of performance: -3.2%. There was an attempt to do the optimization, but the optimization was eventually not run as every time [estimatedNumberOfMatches](https://github.com/apache/lucene-solr/pull/1351/files#diff-aff67e212aa0edd675ec31c068cb642bR268) was not selective enough. The reason for that the data here a day of the year in the range [1, 366], and all segments contain various values through a segment.

- Long sort on date field (msecSinceEpoch).  Speedups: 55.9%.

mayya-sharipova:
@msokolov  @jimczi @jpountz  I was wondering if you have any other additional comments for this change?

mayya-sharipova:
Sorry for bringing this up and not finishing, but I thought that is also worth to report the test results on a smaller collection `wikimedium1m`:

```
 TaskQPS                     baseline   StdDevQPS     patch     StdDev    Pct diff
              TermDTSort      292.71     (15.1%)       59.60      (4.9%)  -79.6% ( -86% -  -70%)
   HighTermDayOfYearSort       60.01     (44.0%)       33.75     (13.6%)  -43.8% ( -70% -   24%)
WARNING: cat=HighTermDayOfYearSort: hit counts differ: 65216 vs 65093+
WARNING: cat=TermDTSort: hit counts differ: 68644 vs 507+
```

Here there is a substantial reduction in performance by using the proposed sort optimization.  

As the data in these indexes are not monotonically increasing `setBottom` is called many times.  
Looks like for smaller indexes (especially with data that is not monotonically increasing) it is faster just to do the conventional sort than the proposed optimization.  

I am not sure how significant is this reduction. 
- **Should we apply the optimization only for segments over 1 million docs?**
- **Should we apply the optimization only when the data is diverse enough?**

Or we can follow up on these proposals in subsequent PRs?

mayya-sharipova:
I thought I also report benchmarking results if we apply the optimization only on segments over 1 million docs .  Here we don't have any significant reductions, but also able to achieve speedups.

wikimedium1m
```
  TaskQPS baseline      StdDevQPS my_modified_version      StdDev                Pct diff
              TermDTSort      395.99      (8.0%)      360.72     (11.5%)   -8.9% ( -26% -   11%)
   HighTermDayOfYearSort       49.51     (19.8%)       51.95     (14.0%)    4.9% ( -24% -   48%)
WARNING: cat=HighTermDayOfYearSort: hit counts differ: 541658 vs 541658+
WARNING: cat=TermDTSort: hit counts differ: 68644 vs 68644+
```

wikimedium10m
```
  TaskQPS baseline      StdDevQPS my_modified_version      StdDev                Pct diff
              TermDTSort       83.37      (5.1%)      111.73     (30.4%)   34.0% (  -1% -   73%)
   HighTermDayOfYearSort       52.46      (6.9%)       46.76     (12.4%)  -10.9% ( -28% -    9%)
WARNING: cat=HighTermDayOfYearSort: hit counts differ: 496079 vs 496079+
WARNING: cat=TermDTSort: hit counts differ: 506054 vs 44560+
```

wikimediumall
```
  TaskQPS baseline      StdDevQPS my_modified_version      StdDev                Pct diff
              TermDTSort       32.23      (3.2%)       85.28     (26.2%)  164.6% ( 131% -  200%)
   HighTermDayOfYearSort       14.46      (5.0%)       13.93      (6.6%)   -3.7% ( -14% -    8%)
WARNING: cat=HighTermDayOfYearSort: hit counts differ: 2485178 vs 2485178+
WARNING: cat=TermDTSort: hit counts differ: 1474717 vs 106400+
```

romseygeek:
Have you tried increasing the sampling rate?  256 might be too low, what about 1024 instead?

mayya-sharipova:
@romseygeek  Are you suggesting to do 
```java
if (updateCounter > 1024 && (updateCounter & 0x1f) != 0x1f) {
```
but this will run optimization even more often which we want to avoid, no?

It the wikimedium1m TermDTSort  case,  `updateCounter` doesn't even reach 20 (so the optimization doesn't called that many times), but enough to make it slower than the traditional sort.

mayya-sharipova:
I have caught up with @jimczi  offline, and it could be how selective a query iterator is important for performance. It is possible that if  a query iterator is already selective enough there is no point to materialize a collector's iterator based on points.  
I am going to run benchmarks on MatchAll query to investigate that.

mikemccand:
The `wikimedium1m` corpus is really too small to draw strong conclusions -- I would use it to run a quick performance test, e.g. to see that it can run to completion, not dying with an exception, but then run the real test on `wikimediumall`.

Do you know why you are seeing these warnings?

```
WARNING: cat=HighTermDayOfYearSort: hit counts differ: 541658 vs 541658+
WARNING: cat=TermDTSort: hit counts differ: 68644 vs 68644+
```

Oh, I guess this is expected, since the optimization may skip some non-competitive matches by skipping forward to find possibly competitive hits based on current PQ bottom?  But then, why is the warning not something more like:

```
WARNING: cat=HighTermDayOfYearSort: hit counts differ: 541658 vs 5000+
```

I.e., it looks like (for these benchmark queries anyways) the optimization did not wind up skipping any hits (though, it thought it may have, hence the added `+`)?

mikemccand:
> I.e., it looks like (for these benchmark queries anyways) the optimization did not wind up skipping any hits (though, it thought it may have, hence the added +)?

OK, I see in some of the other runs above, that the optimization does sometimes reduce the actual number of hits visited, nice!  E.g.:

```
WARNING: cat=TermDTSort: hit counts differ: 754451 vs 1669+
```

lucene/core/src/java/org/apache/lucene/search/FilteringFieldComparator.java
```diff
@@ -0,0 +1,350 @@
...
+            maxDoc = context.reader().maxDoc();
+            maxDocVisited = 0;
+            updateCompetitiveIterator(); // update an iterator if we have a new segment
+            return this;
+        }
+
+        // update its iterator to include possibly only docs that are "stronger" than the current bottom entry
+        public void updateCompetitiveIterator() throws IOException {
+            if (canUpdateIterator == false) return;
+            if (pointValues == null) return;
```
jpountz, line 206: I believe that this `pointValues == null` aims at disabling the optimization when the field has doc values but is not indexed. However it can also be null if the field is indexed but doesn't have values on this segment. It's fine to not optimize this case right now, but let's at least leave a comment about it?

lucene/core/src/java/org/apache/lucene/search/FilteringFieldComparator.java
```diff
@@ -0,0 +1,350 @@
...
+        }
+
+        @Override
+        public void setScorer(Scorable scorer) throws IOException {}
+
+        @Override
+        public final LeafFieldComparator getLeafComparator(LeafReaderContext context) throws IOException {
+            ((NumericComparator) in).doSetNextReader(context);
+            pointValues = context.reader().getPointValues(field);
+            iterator = pointValues == null ? null : ((NumericComparator)in).currentReaderValues;
```
jpountz, line 196: I don't think we should steal the iterator of the wrapped comparator, it's a hacky and I believe that it could cause bugs due to the iterator being advanced on doc IDs out-of-order since the same iterator has two consumers that don't know about each other.

jpountz: Furthermore we use `iterator.cost() >>> 3` below as a condition. I think that the first time we should instead use the cost of the scorer otherwise this approach is going to slow down selective queries.

lucene/core/src/java/org/apache/lucene/search/ScoreMode.java
jpountz: We need both because the first one sorts by score first and should use e.g. block-max WAND while the latter sorts by field so block-max WAND isn't relevant, but we still need to disable bulk scoring. `needsScores` and `isExhaustive` are not complete descriptions of these enum constants.

lucene/core/src/java/org/apache/lucene/search/Weight.java
```diff
@@ -201,19 +202,20 @@ public long cost() {
     @Override
     public int score(LeafCollector collector, Bits acceptDocs, int min, int max) throws IOException {
       collector.setScorer(scorer);
+      DocIdSetIterator scorerIterator = twoPhase == null? iterator : twoPhase.approximation();
```
jpountz, line 205:
```suggestion
      DocIdSetIterator scorerIterator = twoPhase == null ? iterator : twoPhase.approximation();
```

lucene/core/src/java/org/apache/lucene/search/FilteringFieldComparator.java
```diff
@@ -0,0 +1,350 @@
...
+                }
+            };
+            final long threshold = iterator.cost() >>> 3;
+            long estimatedNumberOfMatches = pointValues.estimatePointCount(visitor); // runs in O(log(numPoints))
+            if (estimatedNumberOfMatches >= threshold) {
+                // the new range is not selective enough to be worth materializing, it doesn't reduce number of docs at least 8x
+                return;
+            }
+            pointValues.intersect(visitor);
+            this.iterator = result.build().iterator();
```
jpountz, line 274: I believe that doing it is incorrect if a missing value is configured and would be competitive.

dsmiley:
@mikemccand I'm curious about your statement:

> The wikimedium1m corpus is really too small to draw strong conclusions -- I would use it to run a quick performance test, e.g. to see that it can run to completion, not dying with an exception, but then run the real test on wikimediumall.

Should we infer that you don't think a 1M doc corpus is realistic in many production settings of Lucene?  That's the implication I'm reading from your statement, which I definitely do not agree with.  I've seen plenty of e-commerce search engines in the few-hundred K range (and lower), and also multi-tenant search where most tenants don't have much data in their system.

lucene/core/src/java/org/apache/lucene/search/FilteringFieldComparator.java
```diff
@@ -0,0 +1,350 @@
...
+        if (comparator instanceof FieldComparator.FloatComparator){
+            return new FilteringFieldComparator.FilteringFloatComparator((FieldComparator.FloatComparator) comparator, reverse);
+        }
+        return comparator;
+    }
+
+    /**
+     * A wrapper over {@code NumericComparator} that adds a functionality to filter non-competitive docs.
+     */
+    public static abstract class FilteringNumericComparator<T extends Number> extends FilteringFieldComparator<T> implements LeafFieldComparator {
```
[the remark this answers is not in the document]
mayya-sharipova: @jpountz  Thanks for the review, very good points. I tried to address all your feedback in 24c94ff

lucene/core/src/java/org/apache/lucene/search/FilteringFieldComparator.java
```diff
@@ -0,0 +1,350 @@
...
+ * Decorates a wrapped FieldComparator to add a functionality to skip over non-competitive docs.
+ * FilteringFieldComparator provides two additional functions for a FieldComparator:
+ * 1) {@code competitiveIterator()} that returns an iterator over
+ *      competitive docs that are stronger than already collected docs.
+ * 2) {@code setCanUpdateIterator()} that notifies the comparator when it is ok to start updating its internal iterator.
+ *  This method is called from a collector to inform the comparator to start updating its iterator.
+ */
+public abstract class FilteringFieldComparator<T> extends FieldComparator<T> {
+    final FieldComparator<T> in;
+    protected DocIdSetIterator iterator = null;
```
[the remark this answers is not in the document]
mayya-sharipova: addressed in  24c94ff

lucene/core/src/java/org/apache/lucene/search/FilteringFieldComparator.java
mayya-sharipova: addressed in  24c94ff

lucene/core/src/java/org/apache/lucene/search/FilteringFieldComparator.java
mayya-sharipova: addressed in  24c94ff

lucene/core/src/java/org/apache/lucene/search/FilteringFieldComparator.java
mayya-sharipova: addressed in  24c94ff

mayya-sharipova:
@mikemccand Thanks for looking at this

> Do you know why you are seeing these warnings?
WARNING: cat=HighTermDayOfYearSort: hit counts differ: 541658 vs 541658+
WARNING: cat=TermDTSort: hit counts differ: 68644 vs 68644+
... the optimization did not wind up skipping any hits (though, it thought it may have, hence the added +)

Indeed, we set up `totalHitsRelation` to `GREATER_THAN_OR_EQUAL_TO` when we try to run the optimization, but the optimization may end up  not skipping any documents if it is not selective enough.  It looks like some Scorers behave the same way in updating competitive scores (e.g LongDistanceFeatureQuery#DistanceScorer).

jpountz:
> Should we infer that you don't think a 1M doc corpus is realistic in many production settings of Lucene?

It's certainly realistic, but I think that the point still holds that these collections are not very useful for benchmarking as they tend to be more noisy and can easily miss improvements as even a linear scan is fast on a small collection?

dsmiley:
Okay; I get your point about noise.  I think it's also true that smaller (realistic) data sets may expose how well code can scale down and make different / better choices than it should make for larger sizes; and that's not noise.  So both large and small data sets matter for benchmarking.

lucene/core/src/java/org/apache/lucene/search/FilterLeafCollector.java
```diff
@@ -53,4 +53,8 @@ public String toString() {
     return name + "(" + in + ")";
   }
 
+  @Override
+  public DocIdSetIterator competitiveIterator() {
+    return in.competitiveIterator();
+  }
```
jpountz, line 59: We've had endless discussions about whether or not to delegate in FilterXXX classes and I think that the consensus is that we should only delegate abstract methods. Since this one has a default implementation, let's not delegate and look for extensions of FilterCollector that should delegate it? (e.g. asserting collectors)

lucene/core/src/java/org/apache/lucene/search/FilteringFieldComparator.java
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * Decorates a wrapped FieldComparator to add a functionality to skip over non-competitive docs.
+ * FilteringFieldComparator provides two additional functions for a FieldComparator:
+ * 1) {@code competitiveIterator()} that returns an iterator over
+ *      competitive docs that are stronger than already collected docs.
+ * 2) {@code setCanUpdateIterator()} that notifies the comparator when it is ok to start updating its internal iterator.
+ *  This method is called from a collector to inform the comparator to start updating its iterator.
+ */
+public abstract class FilteringFieldComparator<T> extends FieldComparator<T> {
+    final FieldComparator<T> in;
```
jpountz, line 30: indentation looks wrong, Lucene uses 2-spaces indentation

lucene/core/src/java/org/apache/lucene/search/FilteringFieldComparator.java
```diff
@@ -0,0 +1,83 @@
...
+
+    public FilteringFieldComparator(FieldComparator<T> in) {
+        this.in = in;
+    }
+
+    protected abstract DocIdSetIterator competitiveIterator();
```
jpountz, line 36: Let's only have this method on LeafFieldCompatarors, e.g. by doing something like this? FieldComparators are top-level objects so it doesn't make sense to have leaf-level objects defined on them like DocIdSetIterators.

```suggestion
  @Override
  public abstract FilteringLeafFieldComparator getLeafComparator(LeafReaderContext context) throws IOException; // covariant return type

  public static interface FilteringLeafFieldComparator extends LeafFieldComparator {
    DocIdSetIterator competitiveIterator();
  }
```

Then there's a question of whether setCanUpdateIterator should go there too as well or not.

lucene/core/src/java/org/apache/lucene/search/FilteringFieldComparator.java
```diff
@@ -0,0 +1,83 @@
...
+
+    protected abstract void setCanUpdateIterator() throws IOException;
```
jpountz, line 38: can you add javadocs?

lucene/core/src/java/org/apache/lucene/search/FilteringNumericComparator.java
```diff
@@ -0,0 +1,329 @@
...
+    @Override
+    public final LeafFieldComparator getLeafComparator(LeafReaderContext context) throws IOException {
+        ((NumericComparator) in).doSetNextReader(context);
+        maxDoc = context.reader().maxDoc();
+        maxDocVisited = 0;
+        pointValues = context.reader().getPointValues(field);
+        // TODO: optimize a case when pointValues are missing only on this segment
+        competitiveIterator = pointValues == null ? null : DocIdSetIterator.all(maxDoc);
+        iteratorCost = 0;
+        return this;
```
jpountz, line 152: Related to my above comment, I think we could make things cleaner by implementing the comparator and the leaf comparator in different classes. Then each class would be smaller and many fields could become final.

lucene/core/src/java/org/apache/lucene/search/FilteringNumericComparator.java
```diff
@@ -0,0 +1,329 @@
...
+        ((NumericComparator) in).copy(slot, doc);
+    }
+
+    @Override
+    public void setScorer(Scorable scorer) throws IOException {
+        ((NumericComparator) in).setScorer(scorer);
+        if (scorer instanceof Scorer) {
+            iteratorCost = ((Scorer) scorer).iterator().cost(); // starting iterator cost is the scorer's cost
+            updateCompetitiveIterator(); // update an iterator when we have a new segment
+        }
```
jpountz, line 140: Let's open a separate issue about adding `Scorable.cost()` to make such optimizations easier?

lucene/core/src/java/org/apache/lucene/search/FilteringNumericComparator.java
```diff
@@ -0,0 +1,329 @@
...
+
+    @Override
+    public final LeafFieldComparator getLeafComparator(LeafReaderContext context) throws IOException {
+        ((NumericComparator) in).doSetNextReader(context);
+        maxDoc = context.reader().maxDoc();
+        maxDocVisited = 0;
+        pointValues = context.reader().getPointValues(field);
+        // TODO: optimize a case when pointValues are missing only on this segment
+        competitiveIterator = pointValues == null ? null : DocIdSetIterator.all(maxDoc);
+        iteratorCost = 0;
```
jpountz, line 151: I'd rather have maxDoc as a default, the reason being that this optimization otherwise wouldn't apply with MatchAllDocsQuery, which would be a pity.

lucene/core/src/java/org/apache/lucene/search/FilterLeafCollector.java
mayya-sharipova: addressed in 7120424ff

lucene/core/src/java/org/apache/lucene/search/FilteringFieldComparator.java
mayya-sharipova: Thanks @jpountz, indeed the suggested structure is much better. addressed in 7120424ff

lucene/core/src/java/org/apache/lucene/search/FilteringFieldComparator.java
mayya-sharipova: addressed in 7120424ff

lucene/core/src/java/org/apache/lucene/search/FilteringNumericComparator.java
mayya-sharipova: Thanks @jpountz, indeed the suggested structure is much better. addressed in 7120424

lucene/core/src/java/org/apache/lucene/search/FilteringNumericComparator.java
mayya-sharipova: addressed in 7120424

lucene/core/src/java/org/apache/lucene/search/FilteringFieldComparator.java
mayya-sharipova: addressed in b8e138c

lucene/core/src/java/org/apache/lucene/search/FilteringNumericComparator.java
mayya-sharipova: Opened  [LUCENE-9352](https://issues.apache.org/jira/browse/LUCENE-9352)

mayya-sharipova:
@jpountz @jimczi I was wondering if you have any further comments for this PR.

lucene/core/src/java/org/apache/lucene/search/FieldComparator.java
```diff
@@ -159,14 +159,15 @@ protected void doSetNextReader(LeafReaderContext context) throws IOException {
     protected NumericDocValues getNumericDocValues(LeafReaderContext context, String field) throws IOException {
       return DocValues.getNumeric(context.reader(), field);
     }
+
```
jimczi, line 162: nit: remove line

lucene/core/src/test/org/apache/lucene/search/TestSortOptimization.java
```diff
@@ -0,0 +1,294 @@
...
+import org.apache.lucene.index.IndexWriter;
+import org.apache.lucene.index.IndexWriterConfig;
+import org.apache.lucene.store.Directory;
+import org.apache.lucene.util.LuceneTestCase;
+
+import java.io.IOException;
+
+import static org.apache.lucene.search.SortField.FIELD_SCORE;
+
+public class TestSortOptimization extends LuceneTestCase {
```
jpountz, line 36: Let's use a less generic name, e.g. `TestFieldSortOptimization` to make it clear it doesn't apply to scores, or maybe something even more describing such as `TestFieldSortOptimizationSkipping`?

mayya-sharipova: @jpountz Thank you for the feedback,  all last comments are addressed in 1ab2a6e.  
I have also created a [LUCENE-9384](https://issues.apache.org/jira/browse/LUCENE-9384)  for backporting to 8.x.

msokolov:
This is so close! just needs precommit fix I think?

mayya-sharipova:
@jpountz I am wondering if you have any further feedback for this PR?

mikemccand:
Do we expect the nightly benchmarks (`luceneutil`) to move from this?  Does it speed up any non-relevance (e.g. field'd) sort by default?

mayya-sharipova:
@mikemccand We should see speed-ups by default on sort for numeric fields, as long as these fields are indexed both with docValues and points, and a full total hits count is not needed.
I will be submitting a PR for `luceneutil` for this use-case.

lucene/core/src/java/org/apache/lucene/search/FilteringLeafFieldComparator.java
```diff
@@ -0,0 +1,39 @@
...
+ * FilteringLeafFieldComparator provides two additional functions to a LeafFieldComparator:
+ *  {@code competitiveIterator()} and {@code setCanUpdateIterator()}.
+ */
+public interface FilteringLeafFieldComparator extends LeafFieldComparator {
+  /**
+   * Returns a competitive iterator
+   * @return an iterator over competitive docs that are stronger than already collected docs
+   * or {@code null} if such an iterator is not available for the current segment.
+   */
+  DocIdSetIterator competitiveIterator();
```
jpountz, line 32: maybe we should throw IOException here to not force some implementations to catch/rethrow an an UncheckedIOException

lucene/core/src/java/org/apache/lucene/search/FilteringNumericComparator.java
```diff
@@ -0,0 +1,53 @@
...
+      return new FilteringNumericLeafComparator.FilteringIntLeafComparator((FieldComparator.IntComparator) inLeafComparator, context,
+          ((IntComparator) inLeafComparator).field, reverse, singleSort, hasTopValue);
+    } else if (comparatorClass == FieldComparator.DoubleComparator.class) {
+      return new FilteringNumericLeafComparator.FilteringDoubleLeafComparator((FieldComparator.DoubleComparator) inLeafComparator, context,
+          ((DoubleComparator) inLeafComparator).field, reverse, singleSort, hasTopValue);
+    } else if (comparatorClass == FieldComparator.FloatComparator.class) {
+      return new FilteringNumericLeafComparator.FilteringFloatLeafComparator((FieldComparator.FloatComparator) inLeafComparator, context,
+          ((FloatComparator) inLeafComparator).field, reverse, singleSort, hasTopValue);
+    } else {
+      assert false: "Unexpected class for [FieldComparator]!";
```
jpountz, line 48: add the class to the error message?

lucene/core/src/java/org/apache/lucene/search/FilteringNumericComparator.java
```diff
@@ -0,0 +1,53 @@
...
+      return null;
```
jpountz, line 49: should we throw an IllegalStateException instead of being lenient when assertions are disabled?

lucene/core/src/java/org/apache/lucene/search/FilteringNumericLeafComparator.java
```diff
@@ -0,0 +1,340 @@
...
+        adder.add(docID); // doc is competitive
+      }
+
+      @Override
+      public PointValues.Relation compare(byte[] minPackedValue, byte[] maxPackedValue) {
+        boolean maxValueOutsideQuery = false;
+        boolean minValueOutsideQuery = false;
+        if (maxValueAsBytes != null) {
+          int cmp = Arrays.compareUnsigned(minPackedValue, 0, bytesCount, maxValueAsBytes, 0, bytesCount);
+          maxValueOutsideQuery = cmp > 0 || (singleSort && cmp == 0);
```
jpountz, line 199: should we return CELL_OUTSIDE_QUERY here when appropriate instead of delaying?

PR #1352, opened 2020-03-14
LUCENE-9278: Make javadoc folder structure follow Gradle project path

mocobeta:
# Description

See https://issues.apache.org/jira/browse/LUCENE-9278

### DIFF
--- a/gradle/render-javadoc.gradle
+++ b/gradle/render-javadoc.gradle
@@ -25,7 +25,7 @@ allprojects {
   plugins.withType(JavaPlugin) {
     ext {
       javadocRoot = project.path.startsWith(':lucene') ? project(':lucene').file("build/docs") : project(':solr').file("build/docs")
-      javadocDestDir = "${javadocRoot}/" + pathToDocdir(project.path)
+      javadocDestDir = "${javadocRoot}/${pathToDocdir(project.path)}"
     }
 
     task renderJavadoc {
@@ -331,15 +331,6 @@ configure(project(':solr:contrib:dataimporthandler-extras')) {
         dependsOn "${path}:renderJavadoc"
         linkHref += [ "../../" + pathToDocdir(path) ]
       }
-    }
-  }
-}
-
-
-configure(project(':solr:contrib:extraction')) {
-  plugins.withType(JavaPlugin) {
-    ext {
-      javadocDestDir = "${javadocRoot}/contrib/cell"
     }
   }
 }
--- a/gradle/render-javadoc.gradle
+++ b/gradle/render-javadoc.gradle
@@ -17,11 +17,15 @@
 
 // generate javadocs by using Ant javadoc task
 
+// utility function to convert project path to document output dir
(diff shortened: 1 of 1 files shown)


gradle/render-javadoc.gradle
```diff
@@ -17,11 +17,15 @@
 
 // generate javadocs by using Ant javadoc task
 
+// utility function to convert project path to document output dir
+// e.g.: ':lucene:analysis:common' => 'analysis/common'
+def pathToDocdir = { path -> path.split(':').drop(2).join('/') }
```
mocobeta: There might be better way?

gradle/render-javadoc.gradle
```diff
@@ -294,35 +305,41 @@ configure(project(':solr:core')) {
...
     }
   }
 }
 
+
 configure(project(':solr:contrib:extraction')) {
   plugins.withType(JavaPlugin) {
     ext {
-      javadocDestDir = "${javadocRoot}/solr-cell"
+      javadocDestDir = "${javadocRoot}/contrib/cell"
```
mocobeta: I'm not sure this tweak is needed for gradle build. Simply, "solr/contrib/extraction" (same as other contrib modules) may be okay?

gradle/render-javadoc.gradle
dweiss: This seems fine to me.

gradle/render-javadoc.gradle
dweiss: The "solr-cell" is what it's called in ant. I'm not sure why. I would prefer to stick to folder/ project path naming here too (so contrib/extraction for me).

gradle/render-javadoc.gradle
```diff
@@ -17,11 +17,15 @@
...
+// utility function to convert project path to document output dir
+// e.g.: ':lucene:analysis:common' => 'analysis/common'
+def pathToDocdir = { path -> path.split(':').drop(2).join('/') }
+
 allprojects {
   plugins.withType(JavaPlugin) {
     ext {
       javadocRoot = project.path.startsWith(':lucene') ? project(':lucene').file("build/docs") : project(':solr').file("build/docs")
-      javadocDestDir = "${javadocRoot}/${project.name}"
+      javadocDestDir = "${javadocRoot}/" + pathToDocdir(project.path)
```
dweiss: You can call the function inside the evaluated string too; "${javadocRoot}/${pathToDocDir(project.path)}"

gradle/render-javadoc.gradle
mocobeta: I removed the entire configure block, so now the docs will go to `contrib/extraction`.

gradle/render-javadoc.gradle
mocobeta: Fixed, thanks.

mocobeta:
I mean to close this PR because we will eventually replace all relative paths here with absolute URLs, sooner or later. I'll try to make some changes to renderJavadoc task to drop all relative paths (and move all generated docs under _project_/build again).

PR #1353, opened 2020-03-14
LUCENE-9279: Update dictionary version for Ukrainian analyzer

arysin:
Update morfologik dictionary version to 4.9.0 for Ukrainian analyzer.
There's about 80k of new lemmas


arysin:
Looks like needs a bit more work

PR #1354, opened 2020-03-15
LUCENE-9279: Update dictionary version for Ukrainian analyzer

arysin:


dweiss:
Looks good to me, thanks Andriy!

dweiss:
Hmm... Had to revert it because ant builds started complaining about offending license files. Could be line endings - will have to check.

dweiss:
The checksums didn't have trailing LF. Use 'gradlew updateLicenses' - makes life easier.

PR #1358, opened 2020-03-17
SOLR-14256: replaced EMPTY with empty() to fix deadlock

dsmiley:
The previous change has caused sporadic test failures relating to class loading race conditions.  Basically an abstract base class cannot declare a static field initialized to that of a subclass.  Apparently on interfaces, this is okay -- DocSet was an interface before.  I had changed from an interface to abstract class before as a simplification and because I wanted to lock down the implementations to the only two that existed, and additionally many methods took a pair of DocSets and made assumptions bout the only implementations that exist.  I still like that, even if I need to not have a public static final EMPTY.

### DIFF
--- a/solr/core/src/java/org/apache/solr/search/DocSet.java
+++ b/solr/core/src/java/org/apache/solr/search/DocSet.java
@@ -36,6 +36,21 @@ public abstract class DocSet implements Accountable, Cloneable /* extends Collec
     assert this instanceof BitDocSet || this instanceof SortedIntDocSet;
   }
 
+  // we can't simply use a trivial static initializer "= new SortedIntDocSet" because it can lead to classloader deadlock
+  private static DocSet EMPTY;
+
+  /** An empty instance. */
+  public static DocSet empty() {
+    // The static field EMPTY is set with an immutable instance lazily.  SortedIntDocSet is an "effectively final" class
+    //  and so this is thread-safe, as it will not be partially-constructed.  It's no big deal to create more than
+    //  if multiple threads create this as they are relatively cheap and equivalent.
+    DocSet empty = EMPTY;
+    if (empty == null) {
+      empty = EMPTY = new SortedIntDocSet(new int[0]);
+    }
+    return empty;
+  }
+
   /**
    * Returns the number of documents in the set.
    */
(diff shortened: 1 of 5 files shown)


dsmiley:
@mkhludnev I can't seem to add you as a reviewer, but please take a look at this.

dsmiley:
TestSmileRequest would hit this with seed -Dtests.seed=761C7515C02CEBF1 typically.

solr/core/src/java/org/apache/solr/search/DocSet.java
```diff
@@ -36,6 +36,21 @@
     assert this instanceof BitDocSet || this instanceof SortedIntDocSet;
   }
 
+  // we can't simply use a trivial static initializer "= new SortedIntDocSet" because it can lead to classloader deadlock
+  private static DocSet EMPTY;
+
+  /** An empty instance. */
+  public static DocSet empty() {
```
madrob: If you do the singleton holder idiom then you don't need to have a null check (at the cost of an extra class, but the VM takes care of this quickly) [Effective Java 3rd Ed, Item 83] 

```
    private static class LazyHolder {
        static final DocSet INSTANCE = new SortedIntDocSet(new int[0]);
    }

    public static DocSet empty() {
        return LazyHolder.INSTANCE;
    }

dsmiley: A fine suggestion. Do you think that approach is better?  For this case I find both no better/worse.

madrob: I'll refer to the much wiser than myself, Josh Bloch (emphasis mine) -

> In summary, you should initialize most fields normally, not lazily. If you must initialize a field lazily in order to... break a harmful initialization circularity, then use the appropriate lazy initialization technique. For instance fields, it is the double-check idiom; **for static fields, the lazy initialization holder class idiom**. For instance fields that can tolerate repeated initialization, you may also consider the single-check idiom.

There's actually a lot more nuance in the surrounding paragraphs in the book, most of which I don't understand, but the summary appears to suggest that this is the fastest, simplest way.

dsmiley:
Merged: 4fd96bedc27adff61f3487539adbd67011181b90

PR #1359, opened 2020-03-18
SOLR-13101: Make dir hash computation optional and resilient

andyvuong:
Slight optimizations in ServerSideMetadata given that push operations do not require computing a directory hash of a core's index directory. This change makes the computation optional and resilient by also moving the optional computation to the latest commit capture's retry loop. 

If we hit an exception (file not found, etc) either capturing files from the latest commit or during the hash computation (if captureDirHash=true), we should try and recompute both snapshots anyway.

### DIFF
--- a/solr/core/src/java/org/apache/solr/store/blob/metadata/ServerSideMetadata.java
+++ b/solr/core/src/java/org/apache/solr/store/blob/metadata/ServerSideMetadata.java
@@ -104,10 +104,11 @@ public class ServerSideMetadata {
    *
    * @param reserveCommit if <code>true</code>, latest commit is reserved for a short while ({@link #RESERVE_COMMIT_DURATION})
    *                      to (reasonably) allow the caller to save it while it pushes files to Blob.
-   *
+   * @param captureDirHash if <code>true</code>, a directory hash is computed for the contents of the specified Core's
+   *                      index directory.  
    * @throws Exception if core corresponding to <code>coreName</code> can't be found.
    */
-  public ServerSideMetadata(String coreName, CoreContainer coreContainer, boolean reserveCommit) throws Exception {
+  public ServerSideMetadata(String coreName, CoreContainer coreContainer, boolean reserveCommit, boolean captureDirHash) throws Exception {
     this.coreName = coreName;
     this.container = coreContainer;
     this.core = coreContainer.getCore(coreName);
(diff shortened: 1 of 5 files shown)


solr/core/src/java/org/apache/solr/store/blob/metadata/ServerSideMetadata.java
```diff
@@ -214,23 +228,23 @@ public String getDirectoryHash() {
 
   /**
    * Returns <code>true</code> if the contents of the directory passed into this method is identical to the contents of
-   * the directory of the Solr core of this instance, taken at instance creation time.<p>
+   * the directory of the Solr core of this instance, taken at instance creation time. If the directory hash was not 
+   * computed at the instance creation time, then this returns <code>false</code>
```
mbwaheed: If directory hash was not computed I would throw IllegalStateException instead of returning false. This will help identify the programming error faster.

solr/core/src/java/org/apache/solr/store/blob/process/CorePullTask.java
```diff
@@ -286,7 +286,7 @@ void pullCoreFromBlob(boolean isLeaderPulling) throws InterruptedException {
       }
 
       // Get local metadata + resolve with blob metadata. Given we're doing a pull, don't need to reserve commit point
-      ServerSideMetadata serverMetadata = new ServerSideMetadata(pullCoreInfo.getCoreName(), storeManager.getCoreContainer(), false);
+      ServerSideMetadata serverMetadata = new ServerSideMetadata(pullCoreInfo.getCoreName(), storeManager.getCoreContainer(), false, true);
```
mbwaheed: Maybe add a short comment explaining the passed value of captureDirHash. Same comment for BlobStoreUtils.java and CorePusher.java.

solr/core/src/java/org/apache/solr/store/blob/metadata/ServerSideMetadata.java
```diff
@@ -212,9 +212,12 @@ public long getGeneration() {
   }
 
   /**
-   * @return Null if the directory hash was not computed for the given Core 
+   * @throw IllegalStateException is this instance was not created with a computed directoryHash
```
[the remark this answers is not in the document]
mbwaheed: Actually I don't see this method being used any where. It'll be better to delete the method.

solr/core/src/java/org/apache/solr/store/blob/metadata/ServerSideMetadata.java
```diff
@@ -229,13 +232,18 @@ public String getDirectoryHash() {
   /**
    * Returns <code>true</code> if the contents of the directory passed into this method is identical to the contents of
    * the directory of the Solr core of this instance, taken at instance creation time. If the directory hash was not 
-   * computed at the instance creation time, then this returns <code>false</code>
+   * computed at the instance creation time, then we throw an IllegalStateException indicating a programming error.
    *
    * Passing in the Directory (expected to be the directory of the same core used during construction) because it seems
    * safer than trying to get it again here...
+   * 
+   * @throw IllegalStateException is this instance was not created with a computed directoryHash
```
mbwaheed: typo is=if?

PR #1360, opened 2020-03-18
LUCENE-9281: Retire SPIClassIterator from master because Java 9+ uses different mechanism to load services when module system is used

uschindler:
See https://issues.apache.org/jira/browse/LUCENE-9281 for more details:

We currently have our own implementation of the service loader standard (SPI) fo several reasons:

(1) In some older JDKs the order of classpath was not respected and this lead to wrong order of codecs implementing the same SPI name. This caused tests to sometimes use wrong class (we had this in Lucene 4 where we had a test-only read/write Lucene3 codec that was listed before the read-only one). That's no longer an issue, the order of loading does not matter. In addition, Java now does everything correct.

(2) In Analysis, we require SPI classes to have a constructor taking args (a Map of params in our case). We also extract the NAME from a static field. Standard service loader does not support this, it tries to instantiate the class with default ctor.

With Java 9+, the ServiceLoader now has a stream() method that allows to filter and preprocess classes: https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/ServiceLoader.html#stream()
This allows us to use the new interface and just get the loaded class (which may come from module-info.class or a conventional SPI file): https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/ServiceLoader.Provider.html#type()

This change allows us to convert Lucene to modules listing all SPIs in the module-info.java.


lucene/analysis/common/src/java/org/apache/lucene/analysis/util/AnalysisSPILoader.java
```diff
@@ -143,6 +144,24 @@ public S newInstance(String name, Map<String,String> args) {
...
+
+  /**
+   * Looks up SPI name (static "NAME" field) with appropriate modifiers.
+   * Also it must be a String class and declared in the concrete class.
+   * @return the SPI name
+   * @throws NoSuchFieldException - if the "NAME" field is not defined.
+   * @throws IllegalAccessException - if the "NAME" field is inaccessible.
+   * @throws IllegalStateException - if the "NAME" field does not have appropriate modifiers or isn't a String field.
+   */
+  public static String lookupSPIName(Class<? extends AbstractAnalysisFactory> service) throws NoSuchFieldException, IllegalAccessException, IllegalStateException {
```
mocobeta: This could be package private (since the original method was not public and still this is called only from classes in the same package)?

uschindler: I thought that this method is useful if you want to figure out the name of a factory where you only have class reference. There is no reason to hide it.

lucene/CHANGES.txt
```diff
@@ -47,6 +47,13 @@ API Changes
 * LUCENE-9264: SimpleFSDirectory has been removed in favor of NIOFSDirectory.
   (Yannick Welsch)
 
+* LUCENE-9281: Use java.util.ServiceLoader to load codec components and analysis
+  factories to be compatible with Java Module System. This allows to load factories
+  without META-INF/service from a Java module exposing the factory in the module
+  descriptor. This breaks backwards compatibility as custom factories must now
+  also implement the default constructor (see MIGRATE.txt).
+  (Uwe Schindler, Dawid Weiss, Tomoko Uchida)
```
mocobeta: i did nothing here and had no idea about this issue...

uschindler: Removed you, I just added you because you were doing a lot of preparatory work and also commented on some changes.

mocobeta: Thanks... I just looked through the changes and approved (though i don't fully understand about it).

PR #1361, opened 2020-03-18
LUCENE-8118: Throw exception if DWPT grows beyond it's maximum ram limit

s1monw:
Today if you add documents that cause a single DWPT to grow beyond its maximum ram buffer size we just keep on indexing. If a user  misuses our IndexWriter#addDocuments API and provides a very very large iterable we will run into `ArrayIndexOutOfBoundsException` down the road and abort the IndexWriter. This change is not bulletproof but best effort to ensure that we fail with a better exception message and don't abort the IndexWriter.


lucene/core/src/java/org/apache/lucene/index/DocumentsWriterPerThread.java
```diff
@@ -219,6 +222,9 @@ final void testPoint(String message) {
   /** Anything that will add N docs to the index should reserve first to
    *  make sure it's allowed. */
   private void reserveOneDoc() {
+    if (numDocsInRAM != 0 && hardMaxBytesPerDWPT < bytesUsed()) { // TODO should we check this after each field as well to also preempt within one doc?
```
mikemccand: Or even per-token while inverting?  I think we already detect this today and throw aborting exception?  But we could fix that to instead throw a non-aborting exception, so the one document is deleted, but IW/DWPT are not aborted.  But we should do that later ...

mikemccand: Could you reverse the conditional to `bytesUsed() > hardMaxBytesPerDWPT`?  I think it's more quickly readable in that order?

lucene/core/src/java/org/apache/lucene/index/DocumentsWriterPerThread.java
```diff
@@ -219,6 +222,9 @@ final void testPoint(String message) {
...
+      throw new IllegalArgumentException("RAM used by a single DocumentsWriterPerThread can not exceed: " + hardMaxBytesPerDWPT / 1024 / 1024 + "MB");
```
mikemccand: s/`can not`/`cannot`

s1monw:
> 2) fixing exception handling in DW.updateDocuments to handle this exception gracefully (non-aborting).

the problem in this method was that we need to flush the DWPT to disk even if we hit an non-aborting exception. Otherwise the next doc would hit the same exception or we would violate the assumption that we never receive a flush pending DWPT. 

One thing that I keep thinking about is what happens if we index documents close to the limit and then the next batch of documents would exceed that limit. That would mean that we reject docs event though we could index them into a different DWPT. Should we at least try to index it again in a fresh DWPT unless the DWPT that rejected it was empty? I mean that would prevent the issue and should be straight forward to implement. I can work on this in a followup.

lucene/core/src/test/org/apache/lucene/index/TestIndexWriter.java
```diff
@@ -3941,4 +3943,108 @@ public void testRandomOperationsWithSoftDeletes() throws Exception {
...
+        public Document next() {
+          if (!hasNext()) {
+            throw new AssertionError();
+          }
+          count.incrementAndGet();
+          return doc;
+        }
+      });
+    } catch (IllegalArgumentException ex) {
+      assertEquals("RAM used by a single DocumentsWriterPerThread can exceed: 1945MB", ex.getMessage());
```
dnhatn: can -> can't

lucene/core/src/test/org/apache/lucene/index/TestIndexWriter.java
```diff
@@ -3941,4 +3943,108 @@ public void testRandomOperationsWithSoftDeletes() throws Exception {
...
+
+        @Override
+        public boolean hasNext() {
+          return count.get() < 100_000_000;
+        }
+
+        @Override
+        public Document next() {
+          if (!hasNext()) {
+            throw new AssertionError();
```
dnhatn: I think we should throw an error in `hasNext` instead?

mikemccand:
> > 1. fixing exception handling in DW.updateDocuments to handle this exception gracefully (non-aborting).
> 
> the problem in this method was that we need to flush the DWPT to disk even if we hit an non-aborting exception. Otherwise the next doc would hit the same exception or we would violate the assumption that we never receive a flush pending DWPT.

Thanks for the explanation; make sense.

> One thing that I keep thinking about is what happens if we index documents close to the limit and then the next batch of documents would exceed that limit. That would mean that we reject docs event though we could index them into a different DWPT. Should we at least try to index it again in a fresh DWPT unless the DWPT that rejected it was empty? I mean that would prevent the issue and should be straight forward to implement. I can work on this in a followup.

+1, that's an awesome idea!  In general when picking among the N free DWPTs when a new document arrives shouldn't we pick the DWPT that's smallest (least `.ramBytesUsed()`)?

s1monw:
@mikemccand @dnhatn I added the retry logic, added a test and fixed some minor things. It's changed quite a bit I think it warrants another review.

>+1, that's an awesome idea! In general when picking among the N free DWPTs when a new document arrives shouldn't we pick the DWPT that's smallest (least .ramBytesUsed())?

I will address picking the smallest DWPT in a separate PR.

s1monw:
@dnhatn that's a good question. One option here would be to fix Lucene's `Field` to accept a `Supplier<Reader>` instead of a reader instance. That way we can consume it multiple times and for back-compat we can keep the reader ctor. This is something that we could benefit from in multiple other places too I guess. @mikemccand WDYT?

lucene/core/src/java/org/apache/lucene/index/DocumentsWriter.java
```diff
@@ -435,55 +436,81 @@ private void ensureInitialized(ThreadState state) throws IOException {
...
+      flushingDWPT = flushControl.doAfterDocument(perThread, isUpdate);
+    } catch (MaxBufferSizeExceededException ex) {
+      if (perThread.dwpt.isAborted()) {
+        throw ex;
+      } else {
+        // we hit an exception but still need to flush this DWPT
+        // let's run postUpdate to make sure we flush stuff to disk
+        // in the case we exceed ram limits etc.
+        hasEvents = doAfterDocumentRejected(perThread, isUpdate, hasEvents);
+        // we retry if we the DWPT had more than one document indexed and was flushed
```
mikemccand: s/`we the`/`the`

lucene/core/src/java/org/apache/lucene/index/DocumentsWriter.java
```diff
@@ -435,55 +436,81 @@ private void ensureInitialized(ThreadState state) throws IOException {
...
+        boolean shouldRetry = perThread.dwpt == null && numDocsBefore > 0;
+        if (shouldRetry) {
+          try {
+            // we retry into a brand new DWPT, if it doesn't fit in here we can't index the document
```
mikemccand: You mean retry into another DWPT right?  Not necessarily a brand new (newly created, 0 docs) one?  Maybe tweak comment?

lucene/core/src/java/org/apache/lucene/index/DocumentsWriter.java
```diff
@@ -435,55 +436,81 @@ private void ensureInitialized(ThreadState state) throws IOException {
...
+      if (perThread.dwpt.isAborted()) {
+        throw ex;
+      } else {
+        // we hit an exception but still need to flush this DWPT
+        // let's run postUpdate to make sure we flush stuff to disk
+        // in the case we exceed ram limits etc.
+        hasEvents = doAfterDocumentRejected(perThread, isUpdate, hasEvents);
+        // we retry if we the DWPT had more than one document indexed and was flushed
+        boolean shouldRetry = perThread.dwpt == null && numDocsBefore > 0;
+        if (shouldRetry) {
```
mikemccand: This is wild!  So it means if the doc was so big it caused all RAM used across all DWPTs to be too much (> 2 X the IW's ram buffer?), causing the initially selected DWP to then flush, then we will try (just once) to pick another DWPT and process the doc?

Or, is the `MaxBufferSizeExceededException` just for one DWPT exceeding its single hard max RAM usage (1945 MB by default)?

s1monw: yes this is only one DWPT not across all DWPTs

lucene/core/src/java/org/apache/lucene/index/DocumentsWriter.java
s1monw: no that's a correct comment. We flush the old one and then initialize the ThreadState again with a brand new DWPT

mikemccand: Oh, I see: we create a new DWPT and send the doc there, ok.

lucene/core/src/java/org/apache/lucene/index/DocumentsWriter.java
mikemccand: Got it.

mikemccand:
Another thing we could consider is changing DWPT's postings addressing from `int` to `long` so we don't need retry logic.

We could do it, always, increasing the per-unique-term memory cost.  Or we could maybe find a way to do it conditionally, when a given DWPT wants to exceed the 2.1 GB limit, but that'd be trickier.

s1monw:
> Another thing we could consider is changing DWPT's postings addressing from int to long so we don't need retry logic.

I will look into this. My biggest fear is that we miss places where we exceed limits and testing will be difficult. Yet, there is no rush with this and I have time to explore fixing IntBlockPool addressing.

PR #1364, opened 2020-03-19
SOLR-14335: Lock Solr's memory to prevent swapping

janhoy:
See https://issues.apache.org/jira/browse/SOLR-14335
Goal of this PR is to start Solr with a custom bootstrap class, which delegates to Jetty's main. In the custom bootstrap class, we will then implement memroy locking through JNA


solr/server/build.gradle
```diff
@@ -111,7 +112,12 @@ task assemblePackaging(type: Sync) {
     into "lib/ext"
   })
 
-  from { project.configurations.startJar.singleFile } {
+  from("${buildDir}/libs", {
```
dweiss: This doesn't seem to be right. you rename twice into the same solr.jar. And I don't think it's elegant that the server depends on the launcher (jetty-start). It's a smell.

janhoy: It's just a hack so far to get it building. I thought the bootstrap class belongs in server module, but there is probably a better way to copy the lib in place. Or we could have a completely new solr/bootstrap module with its own build file just to produce solr.jar?

dweiss: I think a new module would make more sense (put a comment in jira).

janhoy:
I updated to have a separate `bootstrap` module, only generating a `solr.jar` witha single class in it. Also the MANIFEST now specifies main class and classpath, so you can run `java -jar solr.jar --module=http` and it will start Solr.

This is still gradle only, so `ant server && bin/solr start` will not work...

solr/bootstrap/src/java/org/apache/solr/bootstrap/NativeLibrary.java
```diff
@@ -0,0 +1,139 @@
...
+    }
+  }
+
+  private NativeLibrary() {
+  }
+
+  /**
+   * @return the detected OSType of the Operating System running the JVM using crude string matching
+   */
+  private static OSType getOsType() {
```
uschindler: we already detect the OS type using `oal.util.Constants`. We should not duplicate code here.

solr/bootstrap/src/java/org/apache/solr/bootstrap/SolrBootstrap.java
```diff
@@ -0,0 +1,51 @@
...
+    org.eclipse.jetty.start.Main.main(args);
+  }
+
+  private void memLockMaybe() {
+    if (Boolean.getBoolean("solr.memory.lock")) {
+      if (NativeLibrary.isAvailable()) {
+        StartLog.info("Attempting to lock Solr's memory to prevent swapping...");
+        NativeLibrary.tryMlockall();
+      } else {
+        StartLog.debug("JNA not available, cannot lock memory");
```
uschindler: This should also be "warn" has it is a bad idea to not tell the user that it does not work at all. If user wants to get rid of warning, he should disable the solr.memory.lock sysprop (resp. env var) from startfile.

solr/bootstrap/src/java/org/apache/solr/bootstrap/NativeLibrary.java
janhoy: This module runs before Lucene's classpath has been established, so does not have access to any of that. That's why we cannot even log with slf4j.

solr/bootstrap/src/java/org/apache/solr/bootstrap/SolrBootstrap.java
janhoy: Agree

solr/bootstrap/src/java/org/apache/solr/bootstrap/NativeLibrary.java
uschindler: Ah right!

solr/bootstrap/build.gradle
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+
+apply plugin: 'java-library'
+
+jar {
+  manifest {
+    attributes(
+            'Main-Class': 'org.apache.solr.bootstrap.SolrBootstrap',
+            'Class-Path': '. lib/start.jar jna.jar'
```
dweiss: This is a chance to clean up manual jar naming, actually. The classpath here should be something like this:
"Class-Path": configurations.runtimeClasspath.collect { "lib/${it.getName()}" }.join(' '))
then bootstrap would have accurate coordinates at the exact library versions it needs to launch. Plus -- the "rename" trickery wouldn't be needed anymore in server.gradle.

The name of the output artifact (bootstrap.jar) can be set in this project as well (so that it doesn't have a version in the name). 
jar.archiveName = "boostrap.jar"

janhoy: I tried this, but got

> * What went wrong:
> A problem occurred evaluating project ':solr:bootstrap'.
> > Not allowed to resolve configuration ':solr:bootstrap:runtimeClasspath' at configuration time (https://guides.gradle.org/performance/#don_t_resolve_dependencies_at_configuration_time). Please upgrade your plugins and double-check your gradle scripts (see stacktrace)
>

janhoy: If you get it working, you may in fact push a change to the PR branch

janhoy:
Lucene and Solr development has moved to separate git repositories and this PR is being bulk-closed.\nPlease open a new PR against https://github.com/apache/solr or https://github.com/apache/lucene if your contribution is still relevant to the project.

PR #1368, opened 2020-03-20
SOLR-14351: Fix/improve MDCLoggingContext usage

dsmiley:
https://issues.apache.org/jira/browse/SOLR-14351

### DIFF
--- a/solr/core/src/java/org/apache/solr/servlet/HttpSolrCall.java
+++ b/solr/core/src/java/org/apache/solr/servlet/HttpSolrCall.java
@@ -613,8 +613,6 @@ public class HttpSolrCall {
         t = t.getCause();
       }
       return RETURN;
-    } finally {
-      MDCLoggingContext.clear(); // admittedly setCore might not have been called but that's okay.
     }
 
   }
--- a/solr/core/src/java/org/apache/solr/servlet/HttpSolrCall.java
+++ b/solr/core/src/java/org/apache/solr/servlet/HttpSolrCall.java
@@ -307,7 +307,6 @@ public class HttpSolrCall {
 
     // With a valid core...
     if (core != null) {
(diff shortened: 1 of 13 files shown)


dsmiley:
I'll commit in two days according to lazy consensus.
Proposed log entry:
OTHER:
* Improved / hardened MDCLoggingContext usage.

dsmiley:
I was thinking that maybe there is a way to reduce the number of places we even have to think about set/clear of the MDC.  In the last commit here I explore that idea.  SolrCore.open (called whenever a core is retrieved from CoreContainer) now calls set on the MDC, and close (it's ref-counted) clears it.  I was then able to remove a bunch of callers, and perhaps in the process enhanced the scope of this context to some callers' callers that made sense.  On the down side, it might be argued this is surprising side-effect behavior that doesn't _belong_ on the open/close of getting a core reference; I dunno.  I think I like it though I'm slightly worried I missed something since this sort of thing isn't really tested well so I'm relying on all of us to notice and report missing or wrong MDC info.

solr/core/src/java/org/apache/solr/servlet/HttpSolrCall.java
```diff
@@ -615,7 +614,7 @@ public Action call() throws IOException {
       }
       return RETURN;
     } finally {
-      MDCLoggingContext.clear();
+      MDCLoggingContext.clear(); // admittedly setCore might not have been called but that's okay.
```
dsmiley: This probably isn't needed anymore either

dsmiley:
There's some stupid precommit import or something I'll fix but otherwise I think this is ready.  I slept on this one a bit and reviewed some pieces of code that gave me doubts, and I'm comfortable with my latest commits on the PR.  I'll commit Friday.

PR #1369, opened 2020-03-21
SOLR-14213: Allow enabling shared store to be scriptable

andyvuong:
Enhancing the configuration section to allow shared storage feature to be enabled via system properties vs just checking for presence of <sharedStore> section in solr.xml.
```
<sharedStore>
  <bool name="sharedStoreEnabled">${sharedStoreEnabled:false}</bool>
</sharedStore>
```
Shared storage is now only enabled if cloud mode is on, <sharedStore> section is present, and the field sharedStoreEnabled is present and set to true. If <sharedStore> is present but the field isn't, startup fails. The field supports system properties overriding whatever default is specified. This can be passed in via the solr binary such as 
`bin/solr start -e cloud -noprompt -DsharedStoreEnabled=true` 
or set by specifying the environment variable SHARE_STORE_ENABLED.

### DIFF
--- a/solr/bin/solr
+++ b/solr/bin/solr
@@ -1975,6 +1975,10 @@ if [ "$SOLR_MODE" == 'solrcloud' ]; then
 
     CLOUD_MODE_OPTS+=('-DzkRun')
   fi
+  
+  if [ -n "$SHARE_STORE_ENABLED" ]; then
+    CLOUD_MODE_OPTS+=("-DsharedStoreEnabled=$SHARE_STORE_ENABLED")
+  fi
 
   # and if collection1 needs to be bootstrapped
   if [ -e "$SOLR_HOME/collection1/core.properties" ]; then
--- a/solr/core/src/java/org/apache/solr/core/SolrXmlConfig.java
+++ b/solr/core/src/java/org/apache/solr/core/SolrXmlConfig.java
@@ -100,8 +100,8 @@ public class SolrXmlConfig {
     
     SharedStoreConfig sharedStoreConfig = null;
     if (config.getNodeList("solr/sharedStore", false).getLength() > 0) {
-      sharedStoreConfig = SharedStoreConfig.loadSharedStoreConfig(
-          readNodeListAsNamedList(config, "solr/sharedStore/*[@name]", "<sharedStore>"));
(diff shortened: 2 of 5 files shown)


solr/core/src/java/org/apache/solr/core/SolrXmlConfig.java
```diff
@@ -568,4 +568,17 @@ private static PluginInfo getTracerPluginInfo(XmlConfigFile config) {
     Node node = config.getNode("solr/tracerConfig", false);
     return (node == null) ? null : new PluginInfo(node, "tracerConfig", false, true);
   }
+  
+  private static SharedStoreConfig loadSharedStoreConfig(NamedList<Object> nl) {
```
andyvuong: fyi I moved the actual initialization method from SharedStoreConfig to SolrXmlConfig to avoid duplicating the existing field extraction methods or changing their visibility and so it aligns with how the rest of the section objects are being created. I've kept the SharedStoreConfig since I anticipate we'll start using it once we complete the enabling configurable's work.

solr/bin/solr
```diff
@@ -1975,6 +1975,10 @@ if [ "$SOLR_MODE" == 'solrcloud' ]; then
 
     CLOUD_MODE_OPTS+=('-DzkRun')
   fi
+  
+  if [ -n "$SHARE_STORE_ENABLED" ]; then
```
mbwaheed: Why not SHARE**D**_STORE_ENABLED?

solr/core/src/java/org/apache/solr/core/SolrXmlConfig.java
```diff
@@ -568,4 +568,17 @@ private static PluginInfo getTracerPluginInfo(XmlConfigFile config) {
...
+    if (nl == null) {
+      // shared store is not configured
+      return null;
+    } 
+    boolean enabled = Boolean.parseBoolean(
+        required("sharedStore", "sharedStoreEnabled", removeValue(nl, "sharedStoreEnabled")));
+    if (enabled) {
+      return new SharedStoreConfig();
+    }
```
mbwaheed: Maybe it works for now, but later we might want to properly add an enabled property to SharedStoreConfig and create SharedStoreConfig even enabled is false.

PR #1371, opened 2020-03-21
SOLR-14333: print readable version of CollapsedPostFilter query

gunasekhardora:
# Description

Prints a readable version of parsed_filter_queries in CollapsedPostFilter query

# Solution

The debug response will now be in the lines of
```
"parsed_filter_queries": [
"CollapsingPostFilter({!collapse field=id, nullPolicy=expand, groupHeadSelectorText=_version_ asc, groupHeadSelectorType=SORT, hint=top_fc, size=5000})"
]
```

# Checklist

### DIFF
--- a/solr/core/src/java/org/apache/solr/search/CollapsingQParserPlugin.java
+++ b/solr/core/src/java/org/apache/solr/search/CollapsingQParserPlugin.java
@@ -188,6 +188,11 @@ public class CollapsingQParserPlugin extends QParserPlugin {
       return 17 * (31 + selectorText.hashCode()) * (31 + type.hashCode());
     }
 
+    @Override
+    public String toString(){
+      return "groupHeadSelectorText=" +this.selectorText + ", groupHeadSelectorType=" +this.type;
+    }
+
     /**
      * returns a new GroupHeadSelector based on the specified local params
      */
@@ -279,7 +284,20 @@ public class CollapsingQParserPlugin extends QParserPlugin {
     }
 
     public String toString(String s) {
-      return s;
+      return "{!collapse field=" + this.collapseField +
+          ", nullPolicy=" + getNullPolicyString(this.nullPolicy) + ", " +
+          this.groupHeadSelector.toString() +
+          (hint == null ? "": ", hint=" + this.hint) +
+          ", size=" + this.size
+          + "}";
+    }
+
+    private String getNullPolicyString(int nullPolicy) {
+      switch (nullPolicy) {
+        case NULL_POLICY_COLLAPSE: return NULL_COLLAPSE;
+        case NULL_POLICY_EXPAND: return NULL_EXPAND;
+        default: return NULL_IGNORE;
+      }
     }
 
     public CollapsingPostFilter(SolrParams localParams, SolrParams params, SolrQueryRequest request) {
--- a/solr/core/src/java/org/apache/solr/search/CollapsingQParserPlugin.java
+++ b/solr/core/src/java/org/apache/solr/search/CollapsingQParserPlugin.java
@@ -156,6 +156,21 @@ public class CollapsingQParserPlugin extends QParserPlugin {
     public int getCode() {
       return code;
     }
+
+    public static NullPolicy fromString(String nullPolicy) {
+      if (StringUtils.isEmpty(nullPolicy)) {
+        return DEFAULT_POLICY;
+      }
+      switch (nullPolicy) {
+        case "ignore": return IGNORE;
+        case "collapse": return COLLAPSE;
+        case "expand": return EXPAND;
+        default:
+          throw new SolrException(SolrException.ErrorCode.BAD_REQUEST, "Invalid nullPolicy: " + nullPolicy);
+      }
+    }
+
+    static NullPolicy DEFAULT_POLICY = IGNORE;
   }
 
 
@@ -313,7 +328,7 @@ public class CollapsingQParserPlugin extends QParserPlugin {
     public String toString(String s) {
       return "CollapsingPostFilter[field=" + this.collapseField +
           ", nullPolicy=" + this.nullPolicy.getName() + ", " +
-          this.groupHeadSelector.toString() +
+          this.groupHeadSelector +
           (hint == null ? "": ", hint=" + this.hint) +
           ", size=" + this.size
           + "]";
@@ -391,12 +406,7 @@ public class CollapsingQParserPlugin extends QParserPlugin {
         }
       }
 
-      String nPolicy = localParams.get("nullPolicy",NullPolicy.IGNORE.getName());
-      try {
-        this.nullPolicy = NullPolicy.valueOf(nPolicy);
-      } catch (IllegalArgumentException e) {
-        throw new SolrException(SolrException.ErrorCode.BAD_REQUEST, "Invalid nullPolicy: " + nPolicy);
-      }
+      this.nullPolicy = NullPolicy.fromString(localParams.get("nullPolicy"));
     }
 
     @SuppressWarnings({"unchecked"})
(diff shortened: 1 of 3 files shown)


solr/core/src/java/org/apache/solr/search/CollapsingQParserPlugin.java
```diff
@@ -184,6 +184,11 @@ public boolean equals(final Object other) {
     public int hashCode() {
       return 17 * (31 + selectorText.hashCode()) * (31 + type.hashCode());
     }
+
+    @Override
+    public String toString(){
+      return selectorText;
```
munendrasn: shouldn't we include type?

solr/core/src/java/org/apache/solr/search/CollapsingQParserPlugin.java
```diff
@@ -276,7 +281,19 @@ public int getCost() {
     }
 
     public String toString(String s) {
-      return s;
+      return "!collapse field=" + this.collapseField +
```
munendrasn: Maybe add curly braces for beginning and end

solr/core/src/java/org/apache/solr/search/CollapsingQParserPlugin.java
```diff
@@ -276,7 +281,19 @@ public int getCost() {
...
+          ", nullPolicy=" + getNullPolicyString(this.nullPolicy) +
+          ", groupHeadSelector=" + this.groupHeadSelector.toString() +
+          (hint == null ? "": ", hint=" + this.hint) +
+          ", size=" + this.size;
+    }
+
+    private String getNullPolicyString(int nullPolicy) {
```
munendrasn: check if converting them enum makes it cleaner

munendrasn (commented):
Please add test cases with debug component and normal toString()

solr/core/src/java/org/apache/solr/search/CollapsingQParserPlugin.java
gunasekhardora: Added type too

solr/core/src/java/org/apache/solr/search/CollapsingQParserPlugin.java
gunasekhardora: Added them

solr/core/src/java/org/apache/solr/search/CollapsingQParserPlugin.java
gunasekhardora: Converted this to enum

gunasekhardora:
Added relevant test cases too. @munendrasn Please review

solr/core/src/java/org/apache/solr/search/CollapsingQParserPlugin.java
```diff
@@ -218,7 +245,7 @@ public static GroupHeadSelector build(final SolrParams localParams) {
     public String hint;
     private boolean needsScores = true;
     private boolean needsScores4Collapsing = false;
-    private int nullPolicy;
+    private NullPolicy nullPolicy;
     private Set<BytesRef> boosted; // ordered by "priority"
     public static final int NULL_POLICY_IGNORE = 0;
```
madrob, line 250: These constants can be removed and everything routed through the NullPolicy enum. Same for the `NULL_COLLAPSE` string above and others.

gunasekhardora: @madrob Missed your comment somehow. Added the change that you have suggested. Can you please review this now?

solr/core/src/java/org/apache/solr/search/CollapsingQParserPlugin.java
```diff
@@ -128,6 +128,28 @@ field collapsing (with ngroups) when the number of distinct groups
   public static final String HINT_TOP_FC = "top_fc";
   public static final String HINT_MULTI_DOCVALUES = "multi_docvalues";
 
+  public enum NullPolicy {
+    IGNORE("ignore", 0),
```
madrob, line 132: Can we get rid of the earlier String declarations too? Lines 125-127?

gunasekhardora: @madrob Removed them. Added an unit test to validate if null policy is an illegal argument as well.

solr/core/src/java/org/apache/solr/search/CollapsingQParserPlugin.java
```diff
@@ -188,6 +206,11 @@ public int hashCode() {
       return 17 * (31 + selectorText.hashCode()) * (31 + type.hashCode());
     }
 
+    @Override
+    public String toString(){
+      return "groupHeadSelectorText=" +this.selectorText + ", groupHeadSelectorType=" +this.type;
```
munendrasn, line 211: would it be better to include the class name too here?
```suggestion
      return "GroupHeadSelector[selectorText=" +this.selectorText + ", type=" +this.type + "]";
```

solr/core/src/java/org/apache/solr/search/CollapsingQParserPlugin.java
```diff
@@ -279,7 +299,12 @@ public int getCost() {
     }
 
     public String toString(String s) {
-      return s;
+      return "{!collapse field=" + this.collapseField +
+          ", nullPolicy=" + this.nullPolicy.getName() + ", " +
+          this.groupHeadSelector.toString() +
+          (hint == null ? "": ", hint=" + this.hint) +
+          ", size=" + this.size
+          + "}";
```
munendrasn, line 307: `{!collapse` specifies the query parser, I think we shouldn't include it in toString()
```suggestion
      return "CollapsingPostFilter[field=" + this.collapseField +
          ", nullPolicy=" + this.nullPolicy.getName() + ", " +
          this.groupHeadSelector +
          (hint == null ? "": ", hint=" + this.hint) +
          ", size=" + this.size
          + "]";
```

munendrasn: The above suggestion might lead cause this in debug response
```
CollapsingPostFilter([CollapsingPostFilter field=id, nullPolicy=expand, GroupHeadSelector[selectorText=_version_ asc, type=SORT], hint=top_fc, size=5000])

```
This is caused due to how Query is converted to String by `QueryParsing` Class but I think that should be handled separately as issue exists for different query type
 https://github.com/apache/lucene-solr/blob/ab3f1f0d1d0fcdbe58958197b11ba3837795a24a/solr/core/src/java/org/apache/solr/search/QueryParsing.java#L336

solr/core/src/java/org/apache/solr/search/CollapsingQParserPlugin.java
munendrasn: I agree with @madrob about removing the unused declarations but as access the public for these variables. I think we should deprecate it in this PR(for 8x) and remove it in different PR
wdyt?

munendrasn (commented):
@gunasekhardora 
Apologies for the delay. I have left some suggestions. In case if you don't have the bandwidth right now, let me know I will make those changes

munendrasn:
Fixed the failing tests and updated changes to include deprecation

solr/core/src/java/org/apache/solr/search/CollapsingQParserPlugin.java
```diff
@@ -122,12 +122,57 @@ field collapsing (with ngroups) when the number of distinct groups
...
+  @Deprecated
   public static final String HINT_MULTI_DOCVALUES = "multi_docvalues";
 
+  public enum NullPolicy {
+    IGNORE("ignore", 0),
+    COLLAPSE("collapse", 1),
+    EXPAND("expand", 2);
+
+    private final String name;
+    private final int code;
```
dsmiley, line 145: I love the introduction of this enum, but don't get the point/purpose of the `code` integer.  BTW, in case you something like this was needed (and I don't see it), all Enums have an `ordinal()` ;-)

PR #1373, opened 2020-03-23
SOLR-14340 - ZkStateReader.readConfigName is doing too much work

mariemat:
# Description

Reduced the scope of ZkStateReader.readConfigName() as described in the JIRA by David.

# Solution

* Removed the check to ensure that the Zk Node exists.
* Removed the check for the existence of the Configset entry.
In a local test with 600 collections, it saved 2/5 of the GETCLUSTERSTATUS  execution time.

# Tests

Removed an assert in TestCollectionAPI that tests that a collection with missing configset are not returned. That behavior will not happen anymore.

# Checklist

- [ X] I have reviewed the guidelines for [How to Contribute](https://wiki.apache.org/solr/HowToContribute) and my code conforms to the standards described there to the best of my ability.
- [  ] I have created a Jira issue and added the issue ID to my pull request title.
- [  ] I have given Solr maintainers [access](https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork) to contribute to my PR branch. (optional but recommended)
- [ X] I have developed this patch against the `master` branch.
- [ X] I have run `ant precommit` and the appropriate test suite.
- [  ] I have added tests for my changes.
- [  ] I have added documentation for the [Ref Guide](https://github.com/apache/lucene-solr/tree/master/solr/solr-ref-guide) (for Solr changes only).

### DIFF
--- a/solr/solrj/src/java/org/apache/solr/common/cloud/ZkStateReader.java
+++ b/solr/solrj/src/java/org/apache/solr/common/cloud/ZkStateReader.java
@@ -294,10 +294,6 @@ public class ZkStateReader implements SolrCloseable {
     log.debug("Loading collection config from: [{}]", path);
 
     try {
-      if (zkClient.exists(path, true) == false) {
-        log.warn("No collection found at path {}.", path);
-        throw new KeeperException.NoNodeException("No collection found at path: " + path);
-      }
       byte[] data = zkClient.getData(path, null, null, true);
       if (data == null) {
         log.warn("No config data found at path {}.", path);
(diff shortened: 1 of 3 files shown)


solr/solrj/src/java/org/apache/solr/common/cloud/ZkStateReader.java
```diff
@@ -294,10 +294,6 @@ public String readConfigName(String collection) throws KeeperException {
     log.debug("Loading collection config from: [{}]", path);
 
     try {
-      if (zkClient.exists(path, true) == false) {
-        log.warn("No collection found at path {}.", path);
-        throw new KeeperException.NoNodeException("No collection found at path: " + path);
-      }
       byte[] data = zkClient.getData(path, null, null, true);
```
HoustonPutman: Small thing, without the check above this will throw a NoNodeException if the path doesn't exist. Maybe this can be wrapped in a try/catch just for the NoNodeException, so that the more user-friendly message used above can be thrown instead:

`throw new KeeperException.NoNodeException("No collection found at path: " + path);`

Just a thought, I'm not even sure it's worth it.

mariemat: As the NoNodeException was still thrown, and the message, even not as precise, is still pretty clear, I left it that way. But if you believe it can be useful, I can add that check with the proper message.

dsmiley: I suggest leaving this because it'll be quite moot once the configSet setting moves to state.json.

PR #1374, opened 2020-03-24
LUCENE-9287:Add DocValuesFieldExistsQuery as a never cache query in UsageTrackingQueryCachingPolicy

iverase:
see https://issues.apache.org/jira/browse/LUCENE-9287

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -127,6 +127,9 @@ Optimizations
 * LUCENE-8103: DoubleValuesSource and QueryValueSource now use a TwoPhaseIterator if one is provided by the Query.
   (Michele Palmia, David Smiley)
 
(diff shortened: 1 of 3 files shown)


lucene/CHANGES.txt
```diff
@@ -127,6 +127,9 @@ Optimizations
 * LUCENE-8103: DoubleValuesSource and QueryValueSource now use a TwoPhaseIterator if one is provided by the Query.
   (Michele Palmia, David Smiley)
 
+* LUCENE-9287:Add DocValuesFieldExistsQuery as a never cache query in UsageTrackingQueryCachingPolicy.
```
jpountz: This is an implementation detail of this class, maybe say something like `UsageTrackingQueryCachingPolicy no longer caches DocValuesFieldExistsQuery`?

PR #1376, opened 2020-03-25
LUCENE-9292: BKDWriter refactor: Group point configuration in its own class

iverase:
see https://issues.apache.org/jira/browse/LUCENE-9292

This changes introduces a new class called `BKDConfig` which contains the point configuration,


iverase:
out of date PR

PR #1377, opened 2020-03-25
LUCENE-9293: BKDWriter refactor: Introduce the concept of a leaf block

iverase:
see https://issues.apache.org/jira/browse/LUCENE-9293

This change introduces a new interface called `BKDLeafBlock`  that replaces the functional interface that is currently used to extract packedValues.


iverase:
out of date PR

PR #1379, opened 2020-03-25
SOLR-14363: Separate /get requests into their own type designation

gerlowskija:
# Description
The SolrLogParseTool parses most log records containing the string "QTime=" as being a query. While this is useful, the designation would be a little more helpful if operations with a different performance profile were broken out into their own "type_s" value. That gives users the ability to view them together or separately, as they prefer.

# Solution

Change the parsing code in SolrLogParseTool to check specifically for "/get" requests when parsing query requests.  Other than settings "type_s" and "id_s" all other fields are handled in a way identical to other queries

# Tests
A unit test for the new "rtg" type_s value.

# Checklist

### DIFF
--- a/solr/core/src/java/org/apache/solr/util/SolrLogPostTool.java
+++ b/solr/core/src/java/org/apache/solr/util/SolrLogPostTool.java
@@ -301,7 +301,7 @@ public class SolrLogPostTool {
       } else if(path != null && params.contains("/replication")) {
         doc.addField("type_s", "replication");
       } else if (path != null && path.contains("/get")) {
-        doc.addField("type_s", "rtg");
+        doc.addField("type_s", "get");
       } else {
         doc.addField("type_s", "query");
       }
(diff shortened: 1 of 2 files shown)


dsmiley:
Can we instead dynamically pull out the request handler path no matter what it is?  Then we could index this with https://builds.apache.org/job/Solr-reference-guide-master/javadoc/tokenizers.html#path-hierarchy-tokenizer or similar

gerlowskija:
> Can we instead dynamically pull out the request handler path no matter what it is?
We do this already - it gets indexed into a field called `path_s`.  That happens regardless of this PR.  Changing the tokenization on `path_s` sounds reasonable, but is probably a separate PR since it's not entirely connected.

This PR is mostly about the `type_s` field. `type_s` gets populated with a few enum-like values that aim to represent broad classes/types of requests: "update", "query", "commit", etc.  The idea is that admins can get a sense at a glance as to what sort of traffic is hitting cross sections of their cluster without getting as granular as `path_s` provides.

PR #1381, opened 2020-03-26
SOLR-14364: LTR SolrFeature fq improvements

dsmiley:
https://issues.apache.org/jira/browse/SOLR-14364

Admittedly there is no test change.  Again my primary motivation is really just for this class to not call `SolrIndexSearcher.getProcessedFilter` and otherwise behave as it did before.

CC @cpoerschke

### DIFF
--- a/solr/contrib/ltr/src/java/org/apache/solr/ltr/feature/SolrFeature.java
+++ b/solr/contrib/ltr/src/java/org/apache/solr/ltr/feature/SolrFeature.java
@@ -23,22 +23,23 @@ import java.util.List;
 import java.util.Map;
 
 import org.apache.lucene.index.LeafReaderContext;
-import org.apache.lucene.search.DocIdSet;
-import org.apache.lucene.search.DocIdSetIterator;
 import org.apache.lucene.search.IndexSearcher;
+import org.apache.lucene.search.MatchNoDocsQuery;
 import org.apache.lucene.search.Query;
 import org.apache.lucene.search.ScoreMode;
 import org.apache.lucene.search.Scorer;
 import org.apache.lucene.search.Weight;
-import org.apache.lucene.util.Bits;
 import org.apache.solr.common.params.CommonParams;
 import org.apache.solr.common.util.NamedList;
 import org.apache.solr.core.SolrCore;
 import org.apache.solr.request.LocalSolrQueryRequest;
 import org.apache.solr.request.SolrQueryRequest;
+import org.apache.solr.search.DocSet;
 import org.apache.solr.search.QParser;
+import org.apache.solr.search.QueryUtils;
 import org.apache.solr.search.SolrIndexSearcher;
 import org.apache.solr.search.SyntaxError;
+
 /**
  * This feature allows you to reuse any Solr query as a feature. The value
  * of the feature will be the score of the given query for the current document.
@@ -61,6 +62,8 @@ public class SolrFeature extends Feature {
   private String df;
   private String q;
   private List<String> fq;
+
+  // The setters will be invoked via reflection from the passed in params
 
   public String getDf() {
     return df;
@@ -109,7 +112,7 @@ public class SolrFeature extends Feature {
   public FeatureWeight createWeight(IndexSearcher searcher, boolean needsScores,
       SolrQueryRequest request, Query originalQuery, Map<String,String[]> efi)
           throws IOException {
-    return new SolrFeatureWeight(searcher, request, originalQuery, efi);
+    return new SolrFeatureWeight((SolrIndexSearcher) searcher, request, originalQuery, efi);
   }
 
   @Override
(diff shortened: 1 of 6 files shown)


solr/core/src/java/org/apache/solr/search/grouping/CommandHandler.java
```diff
@@ -228,12 +226,8 @@ private void searchWithTimeLimiter(Query query,
       collector = MultiCollector.wrap(collector, hitCountCollector);
     }
 
-    if (filter.filter != null) {
-      query = new BooleanQuery.Builder()
-          .add(query, Occur.MUST)
-          .add(filter.filter, Occur.FILTER)
-          .build();
-    }
+    query = QueryUtils.combineQueryAndFilter(query, filter.filter);
```
cpoerschke: observation: if `query` were null (which presumably it wouldn't be but haven't looked) then `QueryUtils.combineQueryAndFilter` handles that differently now compared to existing logic

dsmiley: If the parsed query is null (e.g. was stopwords or something), then the end effect is matching nothing.  You can see this in the former logic by seeing solrQueryWeight getting set to null.  In the new logic, this is done via setting query to MatchNoDocsQuery, and thus "query" variable is non-null.

solr/contrib/ltr/src/java/org/apache/solr/ltr/feature/SolrFeature.java
```diff
@@ -120,67 +123,66 @@ protected void validate() throws FeatureException {
...
+        final SolrQueryRequest req = makeRequest(request.getCore(), q, fq, df);
         if (req == null) {
           throw new IOException("ERROR: No parameters provided");
         }
 
+        // Build the scoring query
+        Query scoreQuery;
+        String qStr = q;
+        if (qStr == null || qStr.isEmpty()) {
+          scoreQuery = null; // ultimately behaves like MatchAllDocs
```
cpoerschke: minor: s/MatchAllDocs/MatchAllDocsQuery

solr/contrib/ltr/src/java/org/apache/solr/ltr/feature/SolrFeature.java
```diff
@@ -120,67 +123,66 @@ protected void validate() throws FeatureException {
...
-        final SolrQueryRequest req = makeRequest(request.getCore(), solrQuery,
-            fqs, df);
+        final SolrQueryRequest req = makeRequest(request.getCore(), q, fq, df);
         if (req == null) {
           throw new IOException("ERROR: No parameters provided");
         }
 
+        // Build the scoring query
+        Query scoreQuery;
+        String qStr = q;
```
cpoerschke: observation: i like how use of `qStr` here hints towards String type which is clearer to read than the prior `solrQuery` being String type but namewise hinting towards Query type

solr/contrib/ltr/src/java/org/apache/solr/ltr/feature/SolrFeature.java
```diff
@@ -120,67 +123,66 @@ protected void validate() throws FeatureException {
...
+          scoreQuery = null; // ultimately behaves like MatchAllDocs
+        } else {
+          qStr = macroExpander.expand(qStr);
+          if (qStr == null) {
+            throw new FeatureException(this.getClass().getSimpleName() + " requires efi parameter that was not passed in request.");
+          }
+          scoreQuery = QParser.getParser(qStr, req).getQuery();
+          // note: QParser can return a null Query sometimes, such as if the query is a stopword or just symbols
+          if (scoreQuery == null) {
+            scoreQuery = new MatchNoDocsQuery(); // debatable; all or none?
```
cpoerschke: Would 'all' be consistent with both prior q-is-missing code path using `*:*` and the new `QueryUtils.combineQueryAndFilter` choosing all?

solr/contrib/ltr/src/java/org/apache/solr/ltr/feature/SolrFeature.java
```diff
@@ -120,67 +123,66 @@ protected void validate() throws FeatureException {
...
+          }
+        }
+
         // Build the filter queries
-        queryAndFilters = new ArrayList<Query>(); // If there are no fqs we just want an empty list
-        if (fqs != null) {
-          for (String fq : fqs) {
-            if ((fq != null) && (fq.trim().length() != 0)) {
```
cpoerschke: observation: the `(fq.trim().length() != 0)` is being removed as part of the refactor but that seems fair since blank `fq` or `fq` with trailing blanks would seem unusual

solr/contrib/ltr/src/java/org/apache/solr/ltr/feature/SolrFeature.java
```diff
@@ -120,67 +123,66 @@ protected void validate() throws FeatureException {
...
-              final Query filterQuery = fqp.getQuery();
+              final Query filterQuery = QParser.getParser(fqStr, req).getQuery();
               if (filterQuery != null) {
-                queryAndFilters.add(filterQuery);
+                filterQueries.add(filterQuery);
               }
             }
           }
+
+          DocSet filtersDocSet = searcher.getDocSet(filterQueries); // execute
```
cpoerschke: observation: i've sort of convinced myself that at this point `filterQueries` should be a non-empty list but if it were an empty list then i haven't yet dived fully enough into the `getDocSet` implementation to understand if that is doing what we want or if `filterDocSetQuery` remaining `null` might be clearer. perhaps you've thought about this already too?

solr/contrib/ltr/src/java/org/apache/solr/ltr/feature/SolrFeature.java
```diff
@@ -209,49 +211,11 @@ private LocalSolrQueryRequest makeRequest(SolrCore core, String solrQuery,
...
-        solrScorer = solrQueryWeight.scorer(context);
-      }
-
-      final DocIdSetIterator idItr = getDocIdSetIteratorFromQueries(
-          queryAndFilters, context);
-      if (idItr != null) {
-        return solrScorer == null ? new ValueFeatureScorer(this, 1f, idItr)
-            : new SolrFeatureScorer(this, solrScorer,
-                new SolrFeatureScorerIterator(idItr, solrScorer.iterator()));
-      } else {
```
cpoerschke: Very nice how all this vanishes!

solr/contrib/ltr/src/java/org/apache/solr/ltr/feature/SolrFeature.java
```diff
@@ -280,52 +243,10 @@ public float score() throws IOException {
...
-        return docID;
-      }
-
-      @Override
-      public long cost() {
-        return filterIterator.cost() + scorerFilter.cost();
+        return solrScorer.getMaxScore(upTo);
       }
 
+      // TODO delegate more methods?
```
cpoerschke: Good question. I'll take a look to see if we could have some reflective unit test coverage either way (similar to existing coverage elsewhere) to ensure delegation for any  methods that might be added in future.

solr/core/src/java/org/apache/solr/search/grouping/CommandHandler.java
cpoerschke: Thanks for the pointer, i see the `MatchNoDocsQuery` now in SolrFeature, hadn't looked before.

Outside of `ltr/SolrFeature` though the `QueryUtils.combineQueryAndFilter` is using `MatchAllDocsQuery`, right?

dsmiley: I wouldn't define combineQueryAndFilter by who uses it to reverse engineer how it might work; it has its own javadocs.  It behaves the same no matter who calls it.  Do you think it's defaults are counterintuitive?

BTW I think it's a mistake that Solr's 'q' effectively defaults to semantics of MatchNoDocsQuery.

solr/contrib/ltr/src/java/org/apache/solr/ltr/feature/SolrFeature.java
dsmiley: It seemed like a needless check since a blank fq would likely get parsed to null and we'd drop it then any way.

solr/contrib/ltr/src/java/org/apache/solr/ltr/feature/SolrFeature.java
dsmiley: Good eye.  It's possible filterQueries might be empty if say someone added a blank fq.  And I verified in a debugger get getDocSet does the right thing.  It's taking a long time to do its job however (accumulating a complete docset, bit by bit).  In another issue I'm targeting at 9.0, I optimized this case to a live-docs DocSet, among other things relating to TwoPhaseIterator.  I should pull that out and make it its own improvement, along with a bit more documentation to getDocSet to indicate that's what the semantics are if an empty list is passed.

solr/contrib/ltr/src/java/org/apache/solr/ltr/feature/SolrFeature.java
dsmiley: I think this is simply debatable in general.  What does it mean for a query parser to return a null query, even when there was some input string?  It's a bike-shed and reasonable arguments can be said either way.  If I was in a more ambitious mood, I'd file a JIRA issue that would mandate that QParsers starting in 9.0 must return something other than null because the current situation is that every place that uses a QParser has to make some decision or possibly have subtle bugs if forgotten (which has happened over time).

The semantics of what's here is no change from the prior logic.  It might not be obvious from the diff.  In an earlier iteration of this PR I overlooked something here and the tests thankfully caught it.  It's subtly confusing because there are multiple levels of defaulting -- the input string, the post-macro expansion, and the QParser result.

I feel strongly about how combineQueryAndFilter defaults / does its job though.  It may be that some callers parse queries and produce different match-all vs match-none compared to combineQueryAndFilter if you were to give that a null parameter, but I'm not about to reconsider these places; I just wanted a useful utility method that I hope is intuitive to folks.

solr/contrib/ltr/src/java/org/apache/solr/ltr/feature/SolrFeature.java
cpoerschke: Looks like there could be scope for factoring out a `FilterFeatureScorer` class, I've opened https://issues.apache.org/jira/browse/SOLR-14378 re: that.

solr/core/src/java/org/apache/solr/search/grouping/CommandHandler.java
cpoerschke: Okay, returning to this I think I now understand the `QueryUtils.combineQueryAndFilter` scenarios better:

* There are two inputs and zero, two or one of them could be null.

  * scoreQuery non-null, filterQuery null:
    * (both old and new code) just use `scoreQuery`.

  * scoreQuery non-null, filterQuery non-null:
    * `BooleanQuery.Builder` is used to combine the two (in both old and new code).

  * scoreQuery null, filterQuery null:
    * In the old code in CommandHandler/SolrIndexSearcher/Grouping/ExpandComponent a value of null would have been used for `scoreQuery`, searching on that isn't gonna end well, is it.
    * In the new code `MatchAllDocsQuery` will be used to return everything.

  * scoreQuery null, filterQuery non-null:
    * In the old code `BooleanQuery.Builder` is used to combine the two, regardless of `scoreQuery` being null and so that search is not going to end well too.
    * In the new code the `filterQuery` wrapped up into a `ConstantScoreQuery` is used to search. In terms of search meaning this is equivalent to the `scoreQuery` being a `MatchAllDocsQuery` i.e. with a score query matching everything only the filter query needs to be searched with.

So in the "scoreQuery null" scenario the new code
* no longer results in a failed search,
* but instead returns something,
* and what is returned is consistent w.r.t. any filter query,
* and/or consistent within itself i.e. a null (score or filter) query is taken to mean 'everything' i.e. `MatchAllDocsQuery`.

So yes, I agree, `MatchAllDocsQuery` is the logical choice.

dsmiley:
Are you okay with me merging this now or is there anything left to resolve @cpoerschke ?

solr/contrib/ltr/src/java/org/apache/solr/ltr/feature/SolrFeature.java
cpoerschke: In case it helps, I've branched off your branch here, merged latest master into it and added small [commit](https://github.com/cpoerschke/lucene-solr/commit/56af94b59e0377c4a1279ed5569ba4cc8cea0ee2) for `SolrFeature` here extending `FilterFeatureScorer`.

cpoerschke:
> Are you okay with me merging this now or is there anything left to resolve @cpoerschke ?

Sure, please go ahead. Thanks!

dsmiley:
Thanks for the code review Christine!

PR #1384, opened 2020-03-27
Remove CurrentCoreDescriptorProvider

madrob:
Replace `CurrentCoreDescriptorProvider` with a functional interface so that it is easier to construct since all implementations in our code base were anonymous classes anyway. Added Javadocs explaining the usage instead of relying on class name to convey information.


solr/core/src/java/org/apache/solr/core/ZkContainer.java
```diff
@@ -112,20 +111,14 @@ public void initZooKeeper(final CoreContainer cc, CloudConfig config) {
...
-                List<CoreDescriptor> descriptors = new ArrayList<>(
-                    cc.getLoadedCoreNames().size());
-                Collection<SolrCore> cores = cc.getCores();
-                for (SolrCore core : cores) {
-                  descriptors.add(core.getCoreDescriptor());
-                }
-                return descriptors;
-              }
-            });
+        ZkController zkController = new ZkController(cc, zookeeperHost, zkClientConnectTimeout, config, () -> {
```
dsmiley: I'd prefer that you declare that last parameter as a local typed variable and then pass it in.  This makes it clearer what it is that is being passed to ZkController's c'tor.

solr/core/src/java/org/apache/solr/cloud/ZkController.java
```diff
@@ -287,7 +288,14 @@ public Object call() throws Exception {
     }
   }
 
-  public ZkController(final CoreContainer cc, String zkServerAddress, int zkClientConnectTimeout, CloudConfig cloudConfig, final CurrentCoreDescriptorProvider registerOnReconnect)
+  /**
+   * @param cc Core container associated with this controller. cannot be null.
+   * @param zkServerAddress where to connect to the zk server
+   * @param zkClientConnectTimeout timeout in ms
+   * @param cloudConfig configuration for this controller. TODO: possibly redundant with CoreContainer
+   * @param registerOnReconnect a provider of the current core descriptors. used to know what to reregister after a reconnect
```
dsmiley: I wonder if "registerOnReconnect" is not appropriate for the name.  It's named how it's being used but not for what it is.  Imagine how silly if all things were named with such an approach.  Perhaps simply "descriptorProvider".  Any way, leave or change as you wish.

solr/core/src/java/org/apache/solr/cloud/ZkController.java
```diff
@@ -469,7 +477,7 @@ public boolean isClosed() {
       if (cc != null) cc.securityNodeChanged();
     });
 
-    init(registerOnReconnect);
+    init();
```
dsmiley: Just looking in GitHub review here but did you omit this param because it's already in scope?

madrob: It was unused in the method.

madrob:
Is ZkController public API? Do I need a JIRA and CHANGES entry for this?

PR #1387, opened 2020-03-29
SOLR-14210: Include replica health in healtcheck handler

janhoy:
See https://issues.apache.org/jira/browse/SOLR-14210


solr/core/src/java/org/apache/solr/handler/admin/HealthCheckHandler.java
```diff
@@ -88,15 +95,42 @@ public void handleRequestBody(SolrQueryRequest req, SolrQueryResponse rsp) throw
...
+  }
+
+  /**
+   * Find replicas DOWN or RECOVERING
+   * @param clusterState clusterstate from ZK
+   * @param nodeName this node name
+   * @return list of core names that are either DOWN ore RECOVERING on 'nodeName'
+   */
+  static List<String> findUnhealthyCores(ClusterState clusterState, String nodeName) {
+    return clusterState.getCollectionsMap().values().stream()
```
janhoy: Not sure about performance here if you have hundreds of collections and thousands of replicas in a large cluster.
The alternative is to instead iterate cores on current node, and consult with clusterState their overall state.

janhoy: > maybe we want to return false if there are any replicas from inactive slices on the node

Inactive shards are not searched, so we should not care about those. A shard split will not clean up the old shard, but instead mark it inactive, until user manually deletes those shards, or the Autoscaling framework rules go reap them. That is why I chose to check active shards only. We should be ok if the active shard(s) only are up and active, then k8s can go restart the next node.

If a shard split is currently running (could be long running), on a node being restarted, the split would be aborted but when the node comes up again I believe the overseer might try again??

janhoy: > Also if the clusterState thinks that cores live on this node, but the core directories do not exist, then I think that this handler should respond not healthy.

Yes, we can add an extra check that for each replica in clusterstate for node, we check that it exists locally?

solr/core/src/test/org/apache/solr/handler/admin/HealthCheckHandlerTest.java
```diff
@@ -177,4 +183,13 @@ public void testHealthCheckV2Api() throws Exception {
     }
   }
 
+  @Test
+  public void testFindUnhealthyCores() throws Exception {
+    ZkStateReader reader = ClusterStateMockUtil.buildClusterState("csrr2rDcsr2rR", 1, 1, "node1", "node2");
```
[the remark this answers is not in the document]
janhoy: Will do

janhoy:
I addressed the review comments

solr/core/src/java/org/apache/solr/handler/admin/HealthCheckHandler.java
HoustonPutman: > > Also if the clusterState thinks that cores live on this node, but the core directories do not exist, then I think that this handler should respond not healthy.
> 
> Yes, we can add an extra check that for each replica in clusterstate for node, we check that it exists locally?

I think that the logic as it stands now should work, because the cluster state will report the replica as "DOWN" (From my experience, but we can also add tests around this). The comment was meant to validate the current approach over the other one you mentioned:

> The alternative is to instead iterate cores on current node, and consult with clusterState their overall state.

HoustonPutman: > If a shard split is currently running (could be long running), on a node being restarted, the split would be aborted but when the node comes up again I believe the overseer might try again??

And I understand your decision for the active shards better now. As long as the overseer thing is true, then we should be fine. And if in the future if we need to, we can add another parameter to fail on inactive slices as well.

HoustonPutman:
Are you targeting this for 8.6 or 9.0?

solr/core/src/java/org/apache/solr/handler/admin/HealthCheckHandler.java
```diff
@@ -88,15 +96,46 @@ public void handleRequestBody(SolrQueryRequest req, SolrQueryResponse rsp) throw
...
-    } else {
+    // Fail if not in live_nodes
+    if (!clusterState.getLiveNodes().contains(cores.getZkController().getNodeName())) {
       rsp.add(STATUS, FAILURE);
       rsp.setException(new SolrException(SolrException.ErrorCode.SERVICE_UNAVAILABLE, "Host Unavailable: Not in live nodes as per zk"));
+      return;
     }
 
-    rsp.setHttpCaching(false);
+    // Optionally require that all cores on this node are active if param 'failWhenRecovering=true'
```
HoustonPutman, line 106: `requireHealthyCores` not `failWhenRecovering`

janhoy:
> Are you targeting this for 8.6 or 9.0

I see no reason for not targeting 8.6

solr/core/src/java/org/apache/solr/handler/admin/HealthCheckHandler.java
```diff
@@ -88,15 +96,46 @@ public void handleRequestBody(SolrQueryRequest req, SolrQueryResponse rsp) throw
...
+
+  /**
+   * Find replicas DOWN or RECOVERING, or replicas in clusterstate that do not exist on local node
+   * @param clusterState clusterstate from ZK
+   * @param nodeName this node name
+   * @param allCoreNames list of all core names on current node
+   * @return list of core names that are either DOWN ore RECOVERING on 'nodeName'
+   */
+  static List<String> findUnhealthyCores(ClusterState clusterState, String nodeName, Collection<String> allCoreNames) {
+    return clusterState.getCollectionsMap().values().stream()
```
shalinmangar, line 132: Why does this need to go to cluster state? This becomes a very expensive method when you have a lot of collections because it fetches collection states for all collections from ZK even if those collections have no replicas on the current node. Imagine having 300 collections and doing 299 zk read operations just for a health check if the node hosts replicas for one collection only.

I think this method should be rewritten to iterate over the CloudDescriptors of all local cores and check for `hasRegistered == true` and `lastPublished == ACTIVE`. Those two should be sufficient for a health check functionality. We don't even need to consult the cluster state.

janhoy: I assumed ClusterState object in each node is cached on the node and iterating it will not incur any new ZK calls, but it is uptated by watches? If it incurs connections then I agree with you!

I want to exclude replicas of inactive shards from the check. The only place I could find that info was in Slice inside Clusterstate. Sure, I can iterate each core on local host, find its Slice-ID and then go lookup the Slice in clusterstate to find whether it's active, that was my other alternative but more code.

shalinmangar: The cluster state is a shell object that holds individual collection states that each live in different znodes. Each node watches only those collection states for which it hosts a replica. The rest of the collections exist as a lazy reference which is populated by a live read. The `getCollectionsMap()` method calls `CollectionRef.get()` for all collections so it will cause a live read to zk for all lazy references. The lazy reference can optionally cache the fetched state for 2 seconds (if you call `CollectionRef.get(true)`) but that too is too short an interval for a health check.

> I want to exclude replicas of inactive shards from the check. The only place I could find that info was in Slice inside Clusterstate.

It's more code but it is a good idea for sure. Your idea of skipping recovery_failed cores from the health check is also sound.

Thanks for taking this up!

janhoy: Thanks, this is super valuable. Will switch the loop to only fetch info for collection/replicas that reside on the node!

solr/core/src/java/org/apache/solr/handler/admin/HealthCheckHandler.java
```diff
@@ -88,15 +98,45 @@ public void handleRequestBody(SolrQueryRequest req, SolrQueryResponse rsp) throw
...
+
+  /**
+   * Find replicas DOWN or RECOVERING, or replicas in clusterstate that do not exist on local node.
+   * We first find local cores which are either not registered or unhealthy, and check each of these against
+   * the clusterstate, and return a list of unhealthy replicas that are part of an active shard for an existing collection
+   * @param cores list of core descriptors to iterate
+   * @param clusterState clusterstate from ZK
+   * @return list of core names that are either DOWN ore RECOVERING on 'nodeName'
+   */
+  static List<String> findUnhealthyCores(Collection<CloudDescriptor> cores, ClusterState clusterState) {
```
janhoy, line 134: @shalinmangar Please have a look at the new logic. If all local cores are active and registered, then we do not consult clusterstate at all. And we only consult clusterstate now to filter out replicas from inactive shards.

The only thing I'm unsure of now is whether I feed the correct **slice ID** to `.getActiveSlicesMap().containsKey(**HERE**)`. Is this slice ID the same as `c.getShardId()` on CloudDescriptor? My unit test is a mock, so I cannot be sure :)

shalinmangar: Looks good to me. Yes, the `c.getShardId()` will give the slice ID so checking that against the active slice map is fine. One gotcha is that the `c.getCoreNodeName()` can return null if a core is not fully loaded yet so you might want to return a count from this method instead of a list of core node names where some of them may be null depending on the current state of the system.

janhoy: Ok. Since I want a more informative error message, I instead did a null-check on `coreNodeName` and use `<collectionName>_<shardId>` as a fallback.

HoustonPutman: I think it might be helpful to name the fallback `unknown:<collectionName>_<shardId>` or `code-loading:<collectionName>_<shardId>` to distinguish from the rest of the list which contains coreNodeNames. Otherwise people might be confused why a core that they can't find in their list is registering as unhealthy.

HoustonPutman: Also this isn't a blocker, but I was wondering if you might have some insight @shalinmangar. When the core directory no longer exists (say for example the disk was wiped before starting the node), the clusterState will register the missing replica(s) as DOWN and the logs will error saying that those cores cannot be found. Will those missing cores still be returned within the `cores.getCores()` call? If so then this is a non-issue, but if they aren't included as I suspect, then this will return healthy even when there are replicas in the clusterState scheduled on the node that are not healthy.

Maybe this just requires a fix in a different part of solr to auto-delete replicas that have cores that are missing on startup. Solr used to auto-recreate the cores, but I think that functionality was removed in Solr 6 or 7.

janhoy: > I think it might be helpful to name the fallback `unknown:<collectionName>_<shardId>` or `code-loading:<collectionName>_<shardId>` to distinguish...

I think I have changed my mind and agree with shalin that it is enough to return a count, and if that count > 0 include in the error msg "N out of M cores are still not healthy". That will avoid the confusion and give a clear and short state to caller. Imagine a node with 3000 cores just having been started but not yet recovered, that list of RECOVERING cores would be huge :) If you really need to know which cores are unhealthy, there are ways to find that elsewhere.

shalinmangar: > When the core directory no longer exists (say for example the disk was wiped before starting the node), the clusterState will register the missing replica(s) as DOWN and the logs will error saying that those cores cannot be found. Will those missing cores still be returned within the cores.getCores() call?

No, if the core directories themselves have been wiped off then the node cannot return that core in `cores.getCores()`. 

> but if they aren't included as I suspect, then this will return healthy even when there are replicas in the clusterState scheduled on the node that are not healthy.

Yes but that should be okay? The node itself is in fact healthy.

> Maybe this just requires a fix in a different part of solr to auto-delete replicas that have cores that are missing on startup.

Is this a common case? i.e. wiping disks and putting the nodes back in rotation? It is more common to have nodes which have cores that are not in the cluster state and those are unloaded automatically if those nodes come back up. Also, there's node lost trigger which can be used to delete replicas from cluster state for nodes that go away for a long time.

janhoy: So I changed logic to report number of unhealthy cores instead of list of names.
I agree that the corner case of wiping disk is not within scope of this issue.

janhoy:
I think I have addressed all comments now?

PR #1388, opened 2020-03-29
LUCENE-9278: Use -linkoffline instead of relative paths to make links to other projects

mocobeta:
For gradle build, use absolute urls instead of relative paths to make inter-project links so that javadoc destination directory can be moved under `_project_/build/`.

- docs of ":lucene:core" prj will go to `lucene/core/build/docs/javadoc/`
- docs of ":lucene:analysis:common" will go to `lucene/analysis/common/build/docs/javadoc/`
- docs of ":solr:core" prj will go to `solr/core/build/docs/javadoc/`
- ... and so on

Also, this directly calls [javadoc tool](https://docs.oracle.com/en/java/javase/11/tools/javadoc.html) rather than Ant javadoc task - Ant javadoc task doesn't recognize `element-list`, the successor to `package-list` up until Java 8, so `<link offline="true" ...>` no longer correctly work with JDK11 (https://issues.apache.org/jira/browse/SOLR-14352).

All generated docs passed the "checkJavaDocs.py" check. In other words, there's no missing package summary.

See also https://issues.apache.org/jira/browse/LUCENE-9278

### DIFF
--- a/gradle/render-javadoc.gradle
+++ b/gradle/render-javadoc.gradle
@@ -54,51 +54,66 @@ allprojects {
       doFirst {
         def srcDirs = sourceSets.main.java.srcDirs.findAll { dir -> dir.exists() }
 
+        def opts = []
+        opts += [ "-overview ${file("src/java/overview.html").toString()}" ]
+        opts += [ "-sourcepath ${srcDirs.join(' ')}" ]
+        opts += [ "-subpackages ${project.path.startsWith(':lucene') ? 'org.apache.lucene' : 'org.apache.solr'}"]
+        opts += [ "-d ${project.javadoc.destinationDir.toString()}" ]
+        opts += [ "-protected" ]
+        opts += [ "-encoding UTF-8" ]
+        opts += [ "-charset UTF-8" ]
+        opts += [ "-docencoding UTF-8" ]
+        opts += [ "-noindex" ]
+        opts += [ "-author" ]
+        opts += [ "-version" ]
+        if (linksource) {
+          opts += [ "-linksource" ]
+        }
+        opts += [ "-use" ]
+        opts += [ "-locale en_US" ]
+        opts += [ "-windowtitle '${title}'" ]
+        opts += [ "-doctitle '${title}'" ]
+        if (!sourceSets.main.compileClasspath.isEmpty()) {
+          opts += ["-classpath ${sourceSets.main.compileClasspath.asPath}" ]
+        }
+        opts += [ "-bottom '<i>Copyright &copy; 2000-${buildYear} Apache Software Foundation. All Rights Reserved.</i>'" ]
+
+        opts += [ "-tag 'lucene.experimental:a:WARNING: This API is experimental and might change in incompatible ways in the next release.'" ]
+        opts += [ "-tag 'lucene.internal:a:NOTE: This API is for internal purposes only and might change in incompatible ways in the next release.'" ]
+        opts += [ "-tag 'lucene.spi:t:SPI Name (case-insensitive: if the name is 'htmlStrip', 'htmlstrip' can be used when looking up the service).'" ]
+
+        // resolve links to JavaSE and JUnit API
+        opts += [ "-linkoffline ${javaSEDocUrl} ${project(':lucene').file('tools/javadoc/java11/').toString()}" ]
+        if (linkJUnit) {
+          opts += [ "-linkoffline ${junitDocUrl} ${project(':lucene').file('tools/javadoc/junit').toURL()}" ]
+        }
+        // resolve inter-project links
+        linkLuceneProjects.collect { path ->
+          opts += [ "-linkoffline ${luceneDocUrl}/${pathToDocdir(path)} ${file(project(path).javadoc.destinationDir).toString()}" ]
+        }
+        linkSorlProjects.collect { path ->
+          opts += [ "-linkoffline ${solrDocUrl}/${pathToDocdir(path)} ${file(project(path).javadoc.destinationDir).toString()}" ]
+        }
+
+        opts += [ "--release 11" ]
+        opts += [ "-Xdoclint:all,-missing" ]
+
+        // temporary file that holds all javadoc options for the current run (for debugging).
+        def optionsFile = new File("${it.getTemporaryDir()}/javadoc-options.txt")
+        // delete previous options if it exists
+        optionsFile.delete()
+        if (!optionsFile.createNewFile()) {
+          // ensure write privilege
+          throw new GradleException("Javadoc generation failed: Cannot create options file: ${optionsFile}")
+        }
+        optionsFile.write(opts.join("\n"), "UTF-8")
+
         project.exec {
           executable javadocCmd
 
-          args += [ "-overview", file("src/java/overview.html").toString() ]
-          args += [ "-sourcepath", srcDirs.join(" ") ]
-          args += [ "-subpackages", project.path.startsWith(":lucene") ? "org.apache.lucene" : "org.apache.solr"]
-          args += [ "-d", project.javadoc.destinationDir.toString() ]
-          args += [ "-protected" ]
-          args += [ "-encoding", "UTF-8" ]
-          args += [ "-charset", "UTF-8" ]
-          args += [ "-docencoding", "UTF-8" ]
-          args += [ "-noindex" ]
-          args += [ "-author" ]
-          args += [ "-version" ]
-          if (linksource) {
-            args += [ "-linksource" ]
-          }
-          args += [ "-use" ]
-          args += [ "-locale", "en_US" ]
-          args += [ "-windowtitle", title ]
-          args += [ "-doctitle", title ]
-          if (!sourceSets.main.compileClasspath.isEmpty()) {
-            args += ["-classpath", sourceSets.main.compileClasspath.asPath]
-          }
-          args += [ "-bottom", "<i>Copyright &copy; 2000-${buildYear} Apache Software Foundation. All Rights Reserved.</i>" ]
+          args += [ "@${optionsFile.toString()}"]
 
-          args += [ "-tag", "lucene.experimental:a:WARNING: This API is experimental and might change in incompatible ways in the next release." ]
-          args += [ "-tag", "lucene.internal:a:NOTE: This API is for internal purposes only and might change in incompatible ways in the next release." ]
-          args += [ "-tag", "lucene.spi:t:SPI Name (case-insensitive: if the name is 'htmlStrip', 'htmlstrip' can be used when looking up the service)." ]
-
-          // resolve links to JavaSE and JUnit API
-          args += [ "-linkoffline", javaSEDocUrl, project(":lucene").file("tools/javadoc/java11/").toString() ]
-          if (linkJUnit) {
-            args += [ "-linkoffline", junitDocUrl, project(":lucene").file("tools/javadoc/junit").toURL() ]
-          }
-          // resolve inter-project links
-          linkLuceneProjects.collect { path ->
-            args += [ "-linkoffline", "${luceneDocUrl}/${pathToDocdir(path)}", file(project(path).javadoc.destinationDir).toString() ]
-          }
-          linkSorlProjects.collect { path ->
-            args += [ "-linkoffline", "${solrDocUrl}/${pathToDocdir(path)}", file(project(path).javadoc.destinationDir).toString() ]
-          }
-
-          args += [ "--release", "11" ]
-          args += [ "-Xdoclint:all,-missing" ]
+          // -J flags can't be passed via options file... (an "javadoc: error - invalid flag: -J-Xmx512m" error occurs)
           args += [ "-J-Xmx512m" ]
           // force locale to be "en_US" (fix for: https://bugs.openjdk.java.net/browse/JDK-8222793)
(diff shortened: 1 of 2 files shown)


gradle/render-javadoc.gradle
```diff
@@ -15,93 +15,105 @@
...
-        ) {
-          srcDirs.collect { srcDir ->
-            packageset(dir: srcDir)
+        project.exec {
+          executable javadocCmd
+
+          args += [ "-overview", file("src/java/overview.html").toString() ]
+          args += [ "-sourcepath", srcDirs.join(" ") ]
+          args += [ "-subpackages", project.path.startsWith(":lucene") ? "org.apache.lucene" : "org.apache.solr"]
+          args += [ "-d", project.javadoc.destinationDir.toString() ]
```
mocobeta: I reused the output destination dir for the gradle default "javadoc" task here. This can be moved to another location if needed.

gradle/render-javadoc.gradle
```diff
@@ -15,93 +15,105 @@
...
       def libName = project.path.startsWith(":lucene") ? "Lucene" : "Solr"
       def title = "${libName} ${project.version} ${project.name} API".toString()
 
+      // absolute urls for "-linkoffline" option
+      def javaSEDocUrl = "https://docs.oracle.com/en/java/javase/11/docs/api/"
+      def junitDocUrl = "https://junit.org/junit4/javadoc/4.12/"
+      def luceneDocUrl = "https://lucene.apache.org/core/${project.version.replace(".", "_")}".toString()
+      def solrDocUrl = "https://lucene.apache.org/solr/${project.version.replace(".", "_")}".toString()
+
+      def javadocCmd = org.gradle.internal.jvm.Jvm.current().getJavadocExecutable()
```
dweiss: Not sure if it's using Gradle's "internal" JVM or the "project's" JVM. We should be using the javadoc tool that is aligned with what is used for the compiler and execution (tests). I think you got it right but maybe worth double checking.

https://docs.gradle.org/current/userguide/build_environment.html
(what happens if you set org.gradle.java.home=... to an older JVM, for example).

gradle/render-javadoc.gradle
dweiss: I like it.

gradle/render-javadoc.gradle
```diff
@@ -15,93 +15,105 @@
...
-          tag(name: "lucene.internal", description: "NOTE: This API is for internal purposes only and might change in incompatible ways in the next release.")
-          tag(name: "lucene.spi", description: "SPI Name (Note: This is case-insensitive. e.g., if the name is 'htmlStrip', 'htmlstrip' can be used when looking up the service):", scope: "types")
+          args += [ "-tag", "lucene.experimental:a:WARNING: This API is experimental and might change in incompatible ways in the next release." ]
+          args += [ "-tag", "lucene.internal:a:NOTE: This API is for internal purposes only and might change in incompatible ways in the next release." ]
+          args += [ "-tag", "lucene.spi:t:SPI Name (case-insensitive: if the name is 'htmlStrip', 'htmlstrip' can be used when looking up the service)." ]
 
           // resolve links to JavaSE and JUnit API
-          link(offline: "true", href: "https://docs.oracle.com/en/java/javase/11/docs/api/", packageListLoc: project(":lucene").file("tools/javadoc/java11/").toString())
+          args += [ "-linkoffline", javaSEDocUrl, project(":lucene").file("tools/javadoc/java11/").toURL() ]
```
dweiss: Why is toURL required? It was a path previously.

gradle/render-javadoc.gradle
mocobeta: I set the default java home to JDK14 (that is used to launch gradle) and "org.gradle.java.home" to JDK11.
```
$ echo $JAVA_HOME
/Library/Java/JavaVirtualMachines/adoptopenjdk-14.jdk/Contents/Home
$ java -version
openjdk version "14" 2020-03-17
$ cat gradle.properties
org.gradle.java.home=/Library/Java/JavaVirtualMachines/adoptopenjdk-11.jdk/Contents/Home
```

I did just print debug for `org.gradle.internal.jvm.Jvm.current().getJavadocExecutable()`, that said the `JVM.current()` points the same jvm to org.gradle.java.home property.
```
/Library/Java/JavaVirtualMachines/adoptopenjdk-11.jdk/Contents/Home/bin/javadoc
```

Is this expected behaviour (or is there another configuration point)?

gradle/render-javadoc.gradle
mocobeta: According to -linkoffline documentation the second parameter takes url, so I used toURL() here. But file path is also fine (there is no difference in terms of generated documents). 
https://docs.oracle.com/en/java/javase/11/tools/javadoc.html
```
-linkoffline url1 url2
```

gradle/render-javadoc.gradle
dweiss: I'm not sure, to be honest. In that setting, which JVM is used for running tests? My assumption was that it'd be the first one one from path (or JAVA_HOME) but maybe I'm wrong.

gradle/render-javadoc.gradle
dweiss: This documentation looks weird to me ;) Specifically the "relative" url2 is not clear. If you do toURL() I think it'll be converted into a full file://.../ URL and it's no longer relative... There is also the question of what it does on Windows where file paths are quite odd.

mocobeta: I changed toURL() to toString(). Both methods return absolute url / path and equally effective for me (MacOS and Linux) here, but I cannot confirm the outputs on Windows file system soon...

dweiss: I can check it on Windows. Let me know what to run, specifically, and if you need anything from me.

mocobeta: Thanks, i'd be glad if you check the (absolute) links are correctly created by changes here when built on Windows. :)
```
./gradlew :lucene:analysis:common:renderJavadoc
```

This should generate hyperlinks to core module, for example 
スクリーンショット 2020-03-31 18 36 01
Here, `org.apache.lucene.analysis.Analyzer` should be `<a>` link to 
`https://lucene.apache.org/core/9_0_0-SNAPSHOT/core/org/apache/lucene/analysis/Analyzer.html` (url will be changed depending on project.version).

dweiss: Darn. I get execution error:

    > Task :lucene:core:renderJavadoc FAILED
    javadoc: error - Illegal package name: "<i>Copyright &copy; 2000-2020 Apache Software Foundation. All Rights Reserved.</i>"
    1 error

    FAILURE: Build failed with an exception.

    * Where:
    Script 'O:\repos\lucene-gradle-master\gradle\render-javadoc.gradle' line: 57

    * What went wrong:
    Execution failed for task ':lucene:core:renderJavadoc'.
    > Process 'command 'C:\Tools\java\jdk\bin\javadoc.exe'' finished with non-zero exit value 2

    * Try:
    Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

    * Get more help at https://help.gradle.org

    BUILD FAILED in 15s
    2 actionable tasks: 1 executed, 1 up-to-date

I think it's because something is escaped improperly. Let me look into this.

dweiss: I think the generation is working properly:
```
<li><a href="https://lucene.apache.org/core/9_0_0-SNAPSHOT/core/org/apache/lucene/analysis/Analyzer.html?is-external=true" title="class or interface in org.apache.lucene.analysis" class="externalLink">org.apache.lucene.analysis.Analyzer</a></li>
```

I am surprised this worked for you since the classpath for one of the modules is empty and this created an empty argument (which was skipped). Hence the problem above. I had to add:

          if (!sourceSets.main.compileClasspath.isEmpty()) {
            args += [ "-classpath", sourceSets.main.compileClasspath.asPath ]
          } 

to make it work.

I am also inclined to generate those arguments into a temporary file and use javadoc's "@file" syntax to pass arguments from that file. This is what gradle does; it also makes inspecting exact options passed to javadoc a tad easier (the file is task-temporary but remains under tmp if the task fails).

mocobeta: I fixed this on https://github.com/apache/lucene-solr/pull/1388/commits/f48c6ecbe3b5a47ae2cc677dae8e1c44b513ac9c.

> I am also inclined to generate those arguments into a temporary file and use javadoc's "@file" syntax to pass arguments from that file. This is what gradle does; it also makes inspecting exact options passed to javadoc a tad easier (the file is task-temporary but remains under tmp if the task fails).

Ok, I can do this. Let me know another PR would be preferred for the change.

dweiss: Either way, whatever is more convenient for you. :)

mocobeta: hi @dweiss,
I made following changes to the PR:
- save javadoc options to a file under the tmp directory for each task (_project_/build/tmp/renderJavadoc).
- pass the options file to javadoc tool via "@file" argument.
- when the task fails, it shows raw javadoc tool err message and options file path for debugging.

I've tested it on MacOS/Fedora Linux/Windows10. Works for me, but I'd appreciate if you could cross-check especially if you come up with some edge cases on Windows.

dweiss:
Hi Tomoko. I made a few minor changes. It is a contract of task.getTemporaryDir that the directory is exclusively owned by each task instance so I just simplified the checks there and overwrite any previous file. I also redirected the output to a temporary file and display it only if an error is thrown. I don't think we need to see the verbose messages from javadoc rendering (do we?). If we do then we could add a switch that would provide a TeeOutputStream (commons-io) and tee the output to the console as well.

I ran it all on Windows and seems to be working. Please feel free to commit it in when you're ready.

[patch.txt](https://github.com/apache/lucene-solr/files/4438204/patch.txt)

mocobeta:
Thanks, Dawid, it works for me. I like the fix to redirect stdout/stderr to a temporary file!

gradle/render-javadoc.gradle
mocobeta: > In that setting, which JVM is used for running tests?

Gradle Test task has a customizable property `executable`, the documentation says "The name of the executable to use. Default with java plugin: java command for the current JVM." If my understanding is correct, it is the runtime on which tests are run.
https://docs.gradle.org/current/dsl/org.gradle.api.tasks.testing.Test.html

I tried to print test's executable and javadoc's executable by adding doLast{} clause on the default test task (gradle/default-tests.gradle).
```
      doLast {
        println("Test executable\t " + executable)
        println("Javadoc executable\t" + org.gradle.internal.jvm.Jvm.current().getJavadocExecutable())
      }
```

I set 
`JAVA_HOME=/usr/local/java/adoptopenjdk/jdk-14+36/` and 
`org.gradle.java.home=/usr/local/java/adoptopenjdk/jdk-11.0.3+7/`
then, ran the tests. It said both `executable` point the gradle's java home. (Here we don't change the executable for test, so the default "current" JVM should be used.)

```
Test executable        /usr/local/java/adoptopenjdk/jdk-11.0.3+7/bin/java
Javadoc executable      /usr/local/java/adoptopenjdk/jdk-11.0.3+7/bin/javadoc
```

Looks consistent to me, though i could miss something here... what do you think @dweiss ?

dweiss: I really don't know the details here. Thank you for investigating! I know it is possible to separate gradle's JVM and the JVM used for everything else but I never did this. ES folks do have this separation implemented but I only skipped through that code a while ago.

https://github.com/elastic/elasticsearch/blob/master/buildSrc/src/main/java/org/elasticsearch/gradle/info/GlobalBuildInfoPlugin.java

mocobeta: I did a little more experiments with `org.gradle.internal.jvm.Jvm.current()`, which is used both on compilation and test execution, the search path is
1. org.gradle.java.home
2. $JAVA_HOME
3. user's default java (on $PATH)

It's consistent with their documentation.

Elasticsearch's custom build plugin takes completely different search strategy from gradle's.
1. "compiler.java" system property
2. $JAVA_HOME
3. org.gradle.java.home
4. user's default java (on $PATH)

(I didn't run it but just interpreted this method https://github.com/elastic/elasticsearch/blob/master/buildSrc/src/main/java/org/elasticsearch/gradle/info/GlobalBuildInfoPlugin.java#L209)

mocobeta: I am going to merge it to master branch since I think I understand what I did here with `org.gradle.internal.jvm.Jvm.current()`.

dweiss: Thanks for looking into this, Tomoko. We may have to do something similar to what ES does since we want to be able to run javac, javadocs and tests against new JVMs (which gradle itself may not support yet). It's a different issue though and it can certainly wait.

mocobeta: I just merged it to the master. 

> We may have to do something similar to what ES does since we want to be able to run javac, javadocs and tests against new JVMs (which gradle itself may not support yet).

Should we open an issue for that, or can it be delayed?

dweiss: I've opened https://issues.apache.org/jira/browse/LUCENE-9312 and will perhaps take a look at it this Eastern -- feel free to join! It is important and your investigation helps.

PR #1389, opened 2020-03-30
LUCENE-9298: Improve RAM accounting in BufferedUpdates when deleted doc IDs and terms are cleared

bringyou:
the method clearDeletedDocIds in BufferedUpdates.java has a bug, it can't reset bytesUsed correctly.
```java
void clearDeletedDocIds() {
  deleteDocIDs.clear();
  bytesUsed.addAndGet(-deleteDocIDs.size() * BufferedUpdates.BYTES_PER_DEL_DOCID);
}
```
this PR will fix it.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/BufferedUpdates.java
+++ b/lucene/core/src/java/org/apache/lucene/index/BufferedUpdates.java
@@ -80,6 +80,7 @@ class BufferedUpdates implements Accountable {
 
   private final Counter bytesUsed = Counter.newCounter(true);
   final Counter fieldUpdatesBytesUsed = Counter.newCounter(true);
+  private final Counter termsBytesUsed = Counter.newCounter();
 
   private final static boolean VERBOSE_DELETES = false;
 
@@ -151,7 +152,7 @@ class BufferedUpdates implements Accountable {
     // is done to respect IndexWriterConfig.setMaxBufferedDeleteTerms.
     numTermDeletes.incrementAndGet();
     if (current == null) {
-      bytesUsed.addAndGet(BYTES_PER_DEL_TERM + term.bytes.length + (Character.BYTES * term.field().length()));
+      termsBytesUsed.addAndGet(BYTES_PER_DEL_TERM + term.bytes.length + (Character.BYTES * term.field().length()));
     }
   }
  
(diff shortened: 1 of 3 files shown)


bringyou:
@s1monw sorry to bother, but can you take a look?

bringyou:
> Change looks good to me. Would you mind adding a small test for this issue? Thanks @bringyou!

sorry for the delay~ add a test for `BufferedUpdates` and change a bit more code, please take another look @dnhatn

lucene/core/src/java/org/apache/lucene/index/BufferedUpdates.java
```diff
@@ -176,8 +176,11 @@ void addBinaryUpdate(BinaryDocValuesUpdate update, int docIDUpto) {
   }
 
   void clearDeleteTerms() {
-    deleteTerms.clear();
     numTermDeletes.set(0);
+    deleteTerms.forEach((term, docIDUpto) -> {
```
s1monw, line 180: Instead of counting this here on clear, can we use a second counter for the deleteTerms next to `bytesUsed`? This would be great. It doesn't need to be thread safe IMO

bringyou: add a counter named `termsBytesUsed`

lucene/core/src/java/org/apache/lucene/index/BufferedUpdates.java
```diff
@@ -80,6 +80,7 @@ load factor (say 2 * POINTER).  Entry is object w/
 
   private final Counter bytesUsed = Counter.newCounter(true);
   final Counter fieldUpdatesBytesUsed = Counter.newCounter(true);
+  private final Counter termsBytesUsed = Counter.newCounter();
```
s1monw, line 83:
```suggestion
  private final Counter termsBytesUsed = Counter.newCounter(true);
```

Sorry I looked at it again and we have one case where we use it in a global context in the delete queue so we need the threadsafety. Rest looks great.

s1monw:
I ran tests and we are ready to go. Please add a changes entry like this:

```diff
diff --git a/lucene/CHANGES.txt b/lucene/CHANGES.txt
index b5bc71cb759..2daa8f3c71f 100644
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -238,6 +238,9 @@ Improvements
 * LUCENE-9171: QueryBuilder can now use BoostAttributes on input token streams to selectively
   boost particular terms or synonyms in parsed queries. (Alessandro Benedetti, Alan Woodward)
 
+* LUCENE-9298: Improve RAM accounting in BufferedUpdates when deleted doc IDs and terms are cleared.
+  (Your Name, Simon Willnauer)
+
 Optimizations
 ---------------------
 ```

bringyou:
@s1monw thanks for the reviews, I add the changelogs and edit issue's title

s1monw:
thanks!!

PR #1393, opened 2020-03-31
SOLR-11245 Cloud native Dockerfile

janhoy:
See https://issues.apache.org/jira/browse/SOLR-11245

This is a fork of Solr's official Dockerfile, simplified for quickly building local image. Very much work in progress, it should probably be simplified even more, get rid of most of the scripts, be able to build from gradle's target folder instead of requiring a tgz etc. 

Just dumping a branch out there that I worked on several months ago, so it has deviated from docker-solr a bit. But it works :)

@makuk66 @dsmiley


janhoy:
Abandoned in favor of #1769

PR #1394, opened 2020-04-01
LUCENE-9300: Fix field infos update on doc values update

jimczi:
Today a doc values update creates a new field infos file that contains the original field infos updated for the new generation as well as the new fields created by the doc values update.

However existing fields are cloned through the global fields (shared in the index writer) instead of the local ones (present in the segment).
In practice this is not an issue since field numbers are shared between segments created by the same index writer.
But this assumption doesn't hold for segments created by different writers and added through IndexWriter#addIndexes(Directory).
In this case, the field number of the same field can differ between segments so any doc values update can corrupt the index
by assigning the wrong field number to an existing field in the next generation.

When this happens, queries and merges can access wrong fields without throwing any error, leading to a silent corruption in the index.

This change ensures that we preserve local field numbers when creating
a new field infos generation.


lucene/core/src/test/org/apache/lucene/index/TestNumericDocValuesUpdates.java
```diff
@@ -1483,6 +1484,81 @@ public void testAddIndexes() throws Exception {
     IOUtils.close(dir1, dir2);
   }
 
+  public void testUpdatesAterAddIndexes() throws Exception {
```
juanka588, line 1487: after?

lucene/core/src/java/org/apache/lucene/index/ReadersAndUpdates.java
```diff
@@ -543,27 +543,30 @@ public synchronized boolean writeFieldUpdates(Directory dir, FieldInfos.FieldNum
...
         // create new fields with the right DV type
+        FieldInfos.Builder builder = new FieldInfos.Builder(fieldNumbers);
         for (List<DocValuesFieldUpdates> updates : pendingDVUpdates.values()) {
           DocValuesFieldUpdates update = updates.get(0);
-          FieldInfo fieldInfo = builder.getOrAdd(update.field);
-          fieldInfo.setDocValuesType(update.type);
+          FieldInfo fi = newFields.get(update.field);
+          if (fi == null) {
+            // the field is not present in this segment so we can fallback to the global fields.
+            fi = builder.getOrAdd(update.field);
```
jpountz, line 563: Had a quick discussion about this line with Jim. This isn't good enough as this might return a field number that already exists in the reader if the field exists in the writer with a number that it less than the number of fields in the reader.

jimczi: How to introduce the same bug but the other way around :(. Thanks for catching this Adrien. I pushed https://github.com/apache/lucene-solr/pull/1394/commits/abd5e08170cc8b7963c0aff3d58afa5766afcb32 to handle this case properly.

lucene/core/src/test/org/apache/lucene/index/TestNumericDocValuesUpdates.java
jimczi: thanks, https://github.com/apache/lucene-solr/pull/1394/commits/abd5e08170cc8b7963c0aff3d58afa5766afcb32

lucene/core/src/java/org/apache/lucene/index/ReadersAndUpdates.java
```diff
@@ -543,27 +543,36 @@ public synchronized boolean writeFieldUpdates(Directory dir, FieldInfos.FieldNum
...
+            if (fi.number <= maxFieldNumber) {
+              // the global field number is already used in this segment for a different field so we force a new one locally.
+              fi = cloneFieldInfo(fi, ++maxFieldNumber);
```
jpountz, line 568: Do we need to set the doc-value type before cloning to make sure the doc-value type is set on the writer? I also wonder if we should use `FieldInfos.Builder` since this logic you implemented looks similar to `FieldInfos.Builder#add`.

jimczi: > I also wonder if we should use FieldInfos.Builder since this logic you implemented looks similar to FieldInfos.Builder#add.

I don't think it works since the logic there is to use the global field number if the field already exists in another segment. In the logic above we try to re-assign the global field number locally since it clashes with a local one.

jimczi: > Do we need to set the doc-value type before cloning to make sure the doc-value type is set on the writer?

thanks, I pushed https://github.com/apache/lucene-solr/pull/1394/commits/090ef78171dfe671b0fb1ab238951fdd617bd65c

lucene/core/src/java/org/apache/lucene/index/ReadersAndUpdates.java
```diff
@@ -543,27 +543,39 @@ public synchronized boolean writeFieldUpdates(Directory dir, FieldInfos.FieldNum
...
         }
 
         // create new fields with the right DV type
+        FieldInfos.Builder builder = new FieldInfos.Builder(fieldNumbers);
         for (List<DocValuesFieldUpdates> updates : pendingDVUpdates.values()) {
           DocValuesFieldUpdates update = updates.get(0);
-          FieldInfo fieldInfo = builder.getOrAdd(update.field);
-          fieldInfo.setDocValuesType(update.type);
+          FieldInfo globalFieldInfo =  builder.getOrAdd(update.field);
+          globalFieldInfo.setDocValuesType(update.type);
```
jpountz, line 563: Sorry I suggested this on the previous review, but the doc-value type is actually supposed to be set on the updateDocValue call, so we should even be able to assert that the doc-value type is correct at this stage instead of setting it.

lucene/core/src/java/org/apache/lucene/index/ReadersAndUpdates.java
```diff
@@ -543,27 +543,39 @@ public synchronized boolean writeFieldUpdates(Directory dir, FieldInfos.FieldNum
...
+          FieldInfo segmentFieldInfo = newFields.get(update.field);
+          if (segmentFieldInfo == null) {
+            // the field is not present in this segment so we can fallback to the global fields.
+            if (globalFieldInfo.number <= maxFieldNumber) {
+              // the global field number is already used in this segment for a different field so we force a new one locally.
+              globalFieldInfo = cloneFieldInfo(globalFieldInfo, ++maxFieldNumber);
+            }
```
jpountz, line 570: I think we should either have an `else` branch that updates the `maxFieldNumber` to avoid conflicts (in case dv updates introduce multiple fields) or always clone the `FieldInfo` to update the number. I have a preference for the latter since it requires fewer branches and makes the code easier to reason about?

lucene/core/src/java/org/apache/lucene/index/ReadersAndUpdates.java
```diff
@@ -543,27 +543,39 @@ public synchronized boolean writeFieldUpdates(Directory dir, FieldInfos.FieldNum
...
+            newFields.put(update.field, globalFieldInfo);
+          } else {
+            segmentFieldInfo.setDocValuesType(update.type);
+            newFields.put(update.field, segmentFieldInfo);
```
jpountz, line 574: this `put` call should be unnecessary since the field is already in the map?

lucene/core/src/java/org/apache/lucene/index/ReadersAndUpdates.java
jpountz: maybe add a comment on the above `getOrAdd` call to mention that it never needs to add the field?

lucene/core/src/java/org/apache/lucene/index/ReadersAndUpdates.java
```diff
@@ -543,27 +543,39 @@ public synchronized boolean writeFieldUpdates(Directory dir, FieldInfos.FieldNum
...
         for (FieldInfo fi : reader.getFieldInfos()) {
-          FieldInfo clone = builder.add(fi);
-          // copy the stuff FieldInfos.Builder doesn't copy
-          for (Entry<String,String> e : fi.attributes().entrySet()) {
-            clone.putAttribute(e.getKey(), e.getValue());
-          }
-          clone.setDocValuesGen(fi.getDocValuesGen());
+          // cannot use builder.add(fi) because it does not preserve
+          // the local field number. Field numbers can be different from the global ones
+          // if the segment was created externally (with IndexWriter#addIndexes(Directory)).
```
jpountz, line 552:
```suggestion
          // if the segment was created externally (and added to this index with IndexWriter#addIndexes(Directory)).
```

lucene/core/src/test/org/apache/lucene/index/TestNumericDocValuesUpdates.java
```diff
@@ -1483,6 +1484,83 @@ public void testAddIndexes() throws Exception {
     IOUtils.close(dir1, dir2);
   }
 
+  public void testUpdatesAfterAddIndexes() throws Exception {
```
jpountz, line 1487: It might be better to split into one unit test for the addIndexes case when the updated field already exists in the segment, and another one for the case when the updated field doesn't exist in the segment yet?

lucene/core/src/java/org/apache/lucene/index/ReadersAndUpdates.java
```diff
@@ -644,6 +656,13 @@ public synchronized boolean writeFieldUpdates(Directory dir, FieldInfos.FieldNum
     return true;
   }
 
+  private FieldInfo cloneFieldInfo(FieldInfo fi, int fieldNumber) {
+    return new FieldInfo(fi.name, fieldNumber, fi.hasVectors(), fi.omitsNorms(), fi.hasPayloads(),
+        fi.getIndexOptions(), fi.getDocValuesType(), fi.getDocValuesGen(), new HashMap<>(fi.attributes()),
+        fi.getPointDimensionCount(), fi.getPointIndexDimensionCount(), fi.getPointNumBytes(), fi.isSoftDeletesField());
+
```
s1monw, line 663: extra newline

lucene/core/src/java/org/apache/lucene/index/ReadersAndUpdates.java
```diff
@@ -543,27 +543,39 @@ public synchronized boolean writeFieldUpdates(Directory dir, FieldInfos.FieldNum
...
-          FieldInfo fieldInfo = builder.getOrAdd(update.field);
-          fieldInfo.setDocValuesType(update.type);
+          FieldInfo globalFieldInfo =  builder.getOrAdd(update.field);
+          globalFieldInfo.setDocValuesType(update.type);
+          FieldInfo segmentFieldInfo = newFields.get(update.field);
+          if (segmentFieldInfo == null) {
+            // the field is not present in this segment so we can fallback to the global fields.
+            if (globalFieldInfo.number <= maxFieldNumber) {
+              // the global field number is already used in this segment for a different field so we force a new one locally.
+              globalFieldInfo = cloneFieldInfo(globalFieldInfo, ++maxFieldNumber);
```
s1monw, line 569: we should set the DV type here too no?

lucene/core/src/java/org/apache/lucene/index/ReadersAndUpdates.java
jimczi: ++, https://github.com/apache/lucene-solr/pull/1394/commits/34718cb64e747e56ebc5341f9180d2f12aace029

lucene/core/src/test/org/apache/lucene/index/TestNumericDocValuesUpdates.java
jimczi: ++, I pushed https://github.com/apache/lucene-solr/pull/1394/commits/00d755888d4744ff6a1beba17e20580f43841f12

lucene/core/src/java/org/apache/lucene/index/ReadersAndUpdates.java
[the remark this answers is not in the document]
jimczi: I added an [assertion](https://github.com/apache/lucene-solr/pull/1394/commits/0b5c6c896bcc67f8ec6d9206d5dc9caa5cc5ee81) to check the dv type of the global field but we still need to set the dv type on the cloned field since it is not preserved by the `getOrAdd` call.

jimczi: ++, https://github.com/apache/lucene-solr/pull/1394/commits/0b5c6c896bcc67f8ec6d9206d5dc9caa5cc5ee81

lucene/core/src/java/org/apache/lucene/index/ReadersAndUpdates.java
jimczi: Thanks, I reworked the logic to always clone and reassign the field number.

lucene/core/src/java/org/apache/lucene/index/ReadersAndUpdates.java
jimczi: ++, removed

lucene/core/src/java/org/apache/lucene/index/ReadersAndUpdates.java
jimczi: ++

lucene/core/src/java/org/apache/lucene/index/ReadersAndUpdates.java
jimczi: we should, thanks. I rewrote the logic in my last commits. I thought that `addOrGet` would preserve the dv type but it doesn't.

jimczi:
Thanks @jpountz and @s1monw . I pushed new commits to address your comments.

lucene/core/src/java/org/apache/lucene/index/ReadersAndUpdates.java
```diff
@@ -543,27 +543,37 @@ public synchronized boolean writeFieldUpdates(Directory dir, FieldInfos.FieldNum
       
       try {
         // clone FieldInfos so that we can update their dvGen separately from
-        // the reader's infos and write them to a new fieldInfos_gen file
-        FieldInfos.Builder builder = new FieldInfos.Builder(fieldNumbers);
-        // cannot use builder.add(reader.getFieldInfos()) because it does not
-        // clone FI.attributes as well FI.dvGen
+        // the reader's infos and write them to a new fieldInfos_gen file.
+        int maxFieldNumber = -1;
+        Map<String, FieldInfo> perName = new HashMap<>();
```
jpountz, line 548: nit: we call it `byName`elsewhere

jimczi: https://github.com/apache/lucene-solr/pull/1394/commits/9fed4319409a4d6c3a70e500cc51a58766116e21

PR #1403, opened 2020-04-04
SOLR-9679: Exception when removing zk node /security.json

janhoy:
See https://issues.apache.org/jira/browse/SOLR-9679


solr/core/src/test/org/apache/solr/security/BasicAuthOnSingleNodeTest.java
```diff
@@ -45,13 +45,13 @@ public void setupCluster() throws Exception {
         .setBasicAuthCredentials("solr", "solr")
         .process(cluster.getSolrClient());
     cluster.waitForActiveCollection(COLLECTION, 4, 4);
+  }
 
-    JettySolrRunner jetty = cluster.getJettySolrRunner(0);
```
janhoy, line 49: I removed this part of initialization - don't see why we should tear down and recreate cluster every time?

solr/core/src/test/org/apache/solr/security/BasicAuthOnSingleNodeTest.java
```diff
@@ -45,13 +45,13 @@ public void setupCluster() throws Exception {
...
-    jetty.stop();
-    cluster.waitForJettyToStop(jetty);
-    jetty.start();
-    cluster.waitForAllNodes(30);
-    cluster.waitForActiveCollection(COLLECTION, 4, 4);
+  @Override
+  @After
+  public void tearDown() throws Exception {
+    cluster.shutdown();
```
janhoy, line 53: With just one test in this class, resource cleanup happened automatically at the end I guess, but with two methods, we need to shutdown cluster in `tearDown`...

PR #1405, opened 2020-04-04
Solr GraphTermsQParser simplifications

dsmiley:
No WeightOrDocIdSet; no override of bulkScorer

@joel-bernstein can you please look.  I happened to be viewing this code and didn't understand why the complexity was there (and no comments justifying it either).


dsmiley:
I'll commit this next week.

PR #1407, opened 2020-04-05
Use QueryUtils.combineQueryAndFilter more

dsmiley:
and check query==MatchAllDocsQuery (minor & cheap optimization)

I more thoroughly examined opportunities to use this new function and found them all.  One or two in some spatial code I left for various reasons (e.g. deprecated code).

Note that BooleanQuery.rewrite has an optimization that looks for a MatchAllDocsQuery paired with filters, thus this check here would ultimately not change the fully rewritten query that gets executed.  But, it's such a cheap/easy check and BooleanQuery is a rather heavy wrapper (it's rewrite is extensive).   

CC @cpoerschke

### DIFF
--- a/solr/core/src/java/org/apache/solr/search/QueryUtils.java
+++ b/solr/core/src/java/org/apache/solr/search/QueryUtils.java
@@ -151,7 +151,7 @@ public class QueryUtils {
    * If neither are null then we combine with a BooleanQuery.
    */
   public static Query combineQueryAndFilter(Query scoreQuery, Query filterQuery) {
-    if (scoreQuery == null) {
+    if (scoreQuery == null || scoreQuery instanceof MatchAllDocsQuery) {
       if (filterQuery == null) {
         return new MatchAllDocsQuery(); // default if nothing -- match everything
       } else {
--- a/solr/core/src/java/org/apache/solr/search/QueryUtils.java
+++ b/solr/core/src/java/org/apache/solr/search/QueryUtils.java
@@ -151,6 +151,7 @@ public class QueryUtils {
    * If neither are null then we combine with a BooleanQuery.
    */
   public static Query combineQueryAndFilter(Query scoreQuery, Query filterQuery) {
+    // check for *:* is simple and avoids needless BooleanQuery wrapper even though BQ.rewrite optimizes this away
     if (scoreQuery == null || scoreQuery instanceof MatchAllDocsQuery) {
       if (filterQuery == null) {
         return new MatchAllDocsQuery(); // default if nothing -- match everything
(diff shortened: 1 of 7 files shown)


solr/core/src/java/org/apache/solr/search/QueryUtils.java
```diff
@@ -151,7 +151,7 @@ public static BooleanQuery build(BooleanQuery.Builder builder, QParser parser) {
    * If neither are null then we combine with a BooleanQuery.
    */
   public static Query combineQueryAndFilter(Query scoreQuery, Query filterQuery) {
-    if (scoreQuery == null) {
+    if (scoreQuery == null || scoreQuery instanceof MatchAllDocsQuery) {
```
dsmiley, line 154: I'll add a comment:
`// check for *:* is simple and avoids needless BooleanQuery wrapper even though BQ.rewrite optimizes this away

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -1660,9 +1656,7 @@ private DocSet getDocListAndSetNC(QueryResult qr, QueryCommand cmd) throws IOExc
 
     ProcessedFilter pf = getProcessedFilter(cmd.getFilter(), cmd.getFilterList());
     Query query = QueryUtils.makeQueryable(cmd.getQuery());
-    if (pf.filter != null) {
-      query = new BooleanQuery.Builder().add(query, Occur.MUST).add(pf.filter, Occur.FILTER).build();
-    }
+    query = QueryUtils.combineQueryAndFilter(query, pf.filter);
```
cpoerschke, line 1659:
```suggestion
    final Query query = QueryUtils.combineQueryAndFilter(QueryUtils.makeQueryable(cmd.getQuery()), pf.filter);
```
could provide clarity w.r.t. `query` not subsequently changing though the `QueryUtils...QueryUtils` bit is a bit ugly perhaps.

solr/core/src/java/org/apache/solr/search/QueryUtils.java
cpoerschke: An alternative to a comment could be to include it in the method's javadocs, though perhaps it's more of an implementation detail thingy.

edge case: if `scoreQuery is a MatchAllDocsQuery (which implies it's non-null)` and `filterQuery is null` then would it make sense to return `scoreQuery` itself rather than the current `return new MatchAllDocsQuery();` at line 156?

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
dsmiley: I was thinking the same thing but did not act.  I will now :-)

solr/core/src/java/org/apache/solr/search/QueryUtils.java
dsmiley: Yes I felt this was an internal detail not worth mentioning.  I pondered how I would document it but felt the javadocs of this method are eclipsing the relative simplicity of what the code does.  I don't think it's worth bothering RE matchAllDocsQuery instantiation.  Really there should be a static singleton for MADQ and when that happens this'd be a non-issue.  It's so minor.

PR #1414, opened 2020-04-06
SOLR-14385 added shard name thru MDC and collection name to split histogram logs

saatchibhalla:
https://issues.apache.org/jira/browse/SOLR-14385

# Description

Added shard name and collection name to split histogram logs. 

# Solution

Had access to collection and now accessing shard name directly from MDC.

# Tests

Modified SplitHandlerTest file to include creation of a test cluster with test collection to verify logs. Verified all related tests pass.

# Checklist

- [ 01894589] I have run `ant precommit` and the appropriate test suite.

### DIFF
--- a/solr/core/src/java/org/apache/solr/handler/admin/SplitOp.java
+++ b/solr/core/src/java/org/apache/solr/handler/admin/SplitOp.java
@@ -26,6 +26,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.TreeMap;
 
+import org.apache.log4j.MDC;
 import org.apache.lucene.index.MultiTerms;
 import org.apache.lucene.index.Terms;
 import org.apache.lucene.index.TermsEnum;
@@ -54,6 +55,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import static org.apache.solr.common.cloud.DocCollection.DOC_ROUTER;
+import static org.apache.solr.common.cloud.ZkStateReader.SHARD_ID_PROP;
 import static org.apache.solr.common.params.CommonParams.PATH;
 import static org.apache.solr.common.params.CoreAdminParams.GET_RANGES;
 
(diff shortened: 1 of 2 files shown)


solr/core/src/java/org/apache/solr/handler/admin/SplitOp.java
```diff
@@ -26,6 +26,7 @@
 import java.util.Map;
 import java.util.TreeMap;
 
+import org.apache.log4j.MDC;
```
dsmiley, line 29: This is the wrong MDC; see MDCLoggingContext which imports `org.slf4j.MDC`

solr/core/src/test/org/apache/solr/handler/admin/SplitHandlerTest.java
```diff
@@ -22,24 +22,50 @@
...
+  private static String systemCollNode;
+  private static String COLL_NAME = "collection1";
+  private static String SHARD_NAME = "shard1";
+  
   @BeforeClass
   public static void beforeTests() throws Exception {
     System.setProperty("managed.schema.mutable", "true");  // needed by cloud-managed config set
     initCore("solrconfig.xml","schema_latest.xml");
+    
+    configureCluster(2)
```
dsmiley, line 56: This seems wrong given this test calls initCore and thus isn't a SolrCloud test.  Maybe it needs to _become_ a SolrCloud test?  On the other hand, it's nice to keep a test lightweight; the only motivation here is for logging?  Doesn't seem justified to me IMO.

CC @yonik

dsmiley:
I'm unsure this is needed at all.  See my comment on JIRA.  Basically, doesn't the replica name already in MDC effectively tell you what you want to know?

solr/core/src/test/org/apache/solr/handler/admin/SplitHandlerTest.java
megancarey: That's fair @dsmiley - do you know if a way to create a mock collection so we can leave as SolrTestCaseJ4 but still avoid NPE from the new log line?

dsmiley: No.  Maybe some _other_ test that is _already_ SolrCloud based could assert whatever log statement you want to do?

IMO log statements don't _need_ tests.  A nice to have.  If it's hard, spend your engineering hours on more worthwhile things.

PR #1416, opened 2020-04-07
LUCENE-9286: FST.Arc.BitTable is read directly from the FST bytes.

bruno-roustant:
Arc is lightweight again and Arc.copyFrom() is light.
Faster FSTEnum traversal by keeping the count of bits set in a new FST.Arc field.


lucene/core/src/java/org/apache/lucene/util/fst/FST.java
```diff
@@ -195,10 +209,9 @@
         posArcsStart = other.posArcsStart();
         arcIdx = other.arcIdx();
         numArcs = other.numArcs();
-        if (nodeFlags() == ARCS_FOR_DIRECT_ADDRESSING) {
-          bitTable = other.bitTable() == null ? null : other.bitTable().copy();
-          firstLabel = other.firstLabel();
-        }
+        bitTableStart = other.bitTableStart;
```
dweiss, line 212: I don't see the full code in the diff but I believe all fields should be copied (or cleared, depending on the condition) from other. It may be a bit of an overhead but it'd keep internal state of arcs consistent (and debugging easier). I recall this code was copying just the required fields depending on nodeFlags() - it should clear or copy everything, really.

lucene/core/src/java/org/apache/lucene/util/fst/FSTEnum.java
```diff
@@ -178,7 +179,7 @@ protected void doSeekCeil() throws IOException {
     } else {
       if (targetIndex < 0) {
         targetIndex = -1;
-      } else if (arc.bitTable().isBitSet(targetIndex)) {
```
dweiss, line 181: This is just a side note but it's code like this one that would really benefit from some kind of arc iterator abstraction. :)

lucene/core/src/java/org/apache/lucene/util/fst/BitTableUtil.java
```diff
@@ -0,0 +1,200 @@
...
+
+
+  /**
+   * Counts all bits set in the bit-table.
+   *
+   * @param bitTableBytes The number of bytes in the bit-table.
+   * @param reader        The {@link FST.BytesReader} to read. It must be positioned at the beginning of the bit-table.
+   */
+  static int countBits(int bitTableBytes, FST.BytesReader reader) throws IOException {
+    assert bitTableBytes >= 0 : "bitTableBytes=" + bitTableBytes;
```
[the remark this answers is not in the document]
bruno-roustant: > wouldn't it be more efficient to entire longs + the remainder?

Yes good point. I don't know why I didn't do the same way as countBitsUpTo() below. This effectively avoids many conditions. I'll change that.

> read8bytes - this is effectively the same as reader.readLong for bitcounts?

Not the same, there is a difference in the byte order. reader.readLong() reads 2 ints by reading the high bytes first. read8Bytes() reads 8 bytes with low byte first. This matters when shifting the bit index. I agree that this does not matter here for bit count, but this way read8Bytes() is compatible for future use with bit index. And in addition it uses less operation since it requires 1 less bit shift and bit mask.

lucene/core/src/java/org/apache/lucene/util/fst/FSTEnum.java
bruno-roustant: I agree. I'd like to review FSTEnum soon. First to refactor and share common code like this, second there is still room for some perf improvement I think for seek floor/ceil.

lucene/core/src/java/org/apache/lucene/util/fst/FST.java
bruno-roustant: Ok. You've debugged this code for a long time so now you know :)
I'll change that and I'll put a comment to explain.

lucene/core/src/java/org/apache/lucene/util/fst/BitTableUtil.java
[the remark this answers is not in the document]
dweiss: I noticed the byte order difference (note the "for bitcounts" bit). :) My gut feeling is that pushing reads so that they're aggregated first, followed by a bitcount will still give you a performance improvement. A bit shift and a bit mask is probably dwarfed when hotspot compiles and inlines all this but single-byte get() methods with conditionals inside will typically perform worse than a bulk get. 

This is a scholarly discussion as things will very likely vary from machine to machine and even between hotpot runs, depending on the calling code layout.

PR #1417, opened 2020-04-08
SOLR-12847: Auto-create a policy rule that corresponds to maxShardsPerNode

sigram:
# Description

# Solution

# Tests

# Checklist


solr/core/src/java/org/apache/solr/cloud/api/collections/CreateCollectionCmd.java
```diff
@@ -407,14 +466,81 @@ public void call(ClusterState clusterState, ZkNodeProps message, NamedList resul
...
+    Clause c = Clause.create("{'replica': '<" + maxReplicasPerNode +"' , 'shard': '#ANY', 'node': '#ANY'}");
+    boolean alreadyExists = false;
+    if (clauses != null) {
+      for (Clause clause : clauses) {
+        if (clause.equals(c)) {
+          alreadyExists = true;
+          break;
+        }
+        if (clause.getReplica() != null) {
+          throw new Assign.AssignmentException("Both an existing policy=" + policyName + " and " +
```
shalinmangar, line 513: This block will always execute if were to add a default policy like the one on SOLR-12845. We should test if there is a similar clause in the policy where similar means that only the value of N in `replica < N` is different than maxShardsPerNode

solr/core/src/java/org/apache/solr/cloud/api/collections/CreateCollectionCmd.java
```diff
@@ -407,14 +466,81 @@ public void call(ClusterState clusterState, ZkNodeProps message, NamedList resul
...
           .onNodes(nodeList)
           .build();
-      Assign.AssignStrategyFactory assignStrategyFactory = new Assign.AssignStrategyFactory(cloudManager);
+      // if Strategy.POLICY is in use AND maxShardsPerNode was specified then for back-compat it needs to
+      // be translated into a rule (see SOLR-12847).
+      // Modify the autoscaling rules as needed BEFORE actually calling assign, setting configToRestore as
+      // a side-effect to restore the original config if the creation fails
+      if ((strategy == Assign.AssignStrategyFactory.Strategy.POLICY) &&
+          (maxReplicasPerNode != null && maxReplicasPerNode != Integer.MAX_VALUE)) {
+        maybeAddMaxReplicasRule(cloudManager, maxReplicasPerNode, docCollection, configToRestore);
```
shalinmangar, line 475: I would have preferred to do this outside of this method instead of adding a side effect here. That way the addition of the clause as well as its cleanup can be in one place.

solr/core/src/java/org/apache/solr/cloud/api/collections/DeleteCollectionCmd.java
```diff
@@ -159,6 +165,36 @@ public void call(ClusterState state, ZkNodeProps message, NamedList results) thr
         });
       }
 
+      // remove auto-created policy if this is the only collection that uses it
```
shalinmangar, line 168: Maybe we should disallow people from using these auto created policies in the first place?

solr/solrj/src/java/org/apache/solr/common/params/CollectionAdminParams.java
```diff
@@ -127,4 +127,8 @@
 
   /** Option to follow aliases when deciding the target of a collection admin command. */
   String FOLLOW_ALIASES = "followAliases";
+
+  /** Prefix for automatically created config elements. */
+  String AUTO_PREFIX = ".auto_";
```
shalinmangar, line 132: We use a suffix ".AUTOCREATED" for configsets, maybe we can use the same here?

PR #1418, opened 2020-04-08
LUCENE-9309: Wait for #addIndexes merges when aborting merges

s1monw:
The SegmentMerger usage in IW#addIndexes(CodecReader...) might make changes
to the Directory while the IW tries to clean-up files on rollback. This
causes issues like FileNotFoundExceptions when IDF tries to remove temp files.
This changes adds a waiting mechanism to the abortMerges method that, in addition
to the running merges, also waits for merges in addIndices(CodecReader...)


s1monw:
@dnhatn @mikemccand can you take a look?

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
```diff
@@ -397,11 +393,12 @@ int availablePermits() {
   HashSet<SegmentCommitInfo> mergingSegments = new HashSet<>();
 
   private final MergeScheduler mergeScheduler;
+  private Set<SegmentMerger> runningAddIndexesMerges = new HashSet<>();
   private LinkedList<MergePolicy.OneMerge> pendingMerges = new LinkedList<>();
   private Set<MergePolicy.OneMerge> runningMerges = new HashSet<>();
   private List<MergePolicy.OneMerge> mergeExceptions = new ArrayList<>();
   private long mergeGen;
-  private boolean stopMerges;
+  private boolean stopMerges; // TODOD make sure this is only changed once and never set back to false
```
mikemccand, line 401: Remove extra `D`?

PR #1424, opened 2020-04-11
SOLR-13101: Enable shared store via system property only

andyvuong:
Removing a previous [need](https://github.com/apache/lucene-solr/pull/1223) to configure shared storage feature via solr.xml and only requiring it to be enabled via passed system properties.

Pass in -DsharedStoreEnabled=true to the solr binary or set env variable SHARED_STORE_ENABLED. Note sys props passed to the binary will take precedence.


andyvuong:
@yonik can you merge?

janhoy:
Please link this PR with SOLR-13101 by prefixing the title.

PR #1426, opened 2020-04-12
Do a bit count on 8 bytes from a long directly instead of reading 8 b…

dweiss:
Couldn't resist. Shows a marginal improvement in tight loops. @bruno-roustant ?


dweiss:
I can't say for sure. Like I said - reasoning about hotspot optimizations is difficult and not reproducible across different environments. You could try to dump assembly code and check out the difference there. My experience tells me that even a different generated code shape could be the reason... 

I wanted to get this patch in on Java side mostly because if the reader does optimize getLong somehow then it automatically propagates to any calling code whereas a sequence of readByte() doesn't. These things accumulate over time and it's hard to keep track of them later on.

PR #1427, opened 2020-04-12
LUCENE-9304: Fix IW#getMaxCompletedSequenceNumber() 

s1monw:
After recent refactorings on LUCENE-9304 `IW#getMaxCompletedSequenceNumber()` might return values that belong to non-completed operations if a full flush is running, a new delete queue is already in place but not all DWPTs that participate in the full flush have finished it's in-flight operation. This caused rare failures in `TestControlledRealTimeReopenThread#testControlledRealTimeReopenThread` were documents are not actually visible given the max completed seqNo. This change streamlines the delete queue advance, adds a dedicated testcase and ensures that a delete queues sequence Id space is never exhausted.


s1monw:
@dnhatn @mikemccand can you take a look please

lucene/core/src/test/org/apache/lucene/index/TestIndexWriter.java
```diff
@@ -3949,4 +3950,69 @@ public void testRandomOperationsWithSoftDeletes() throws Exception {
...
+            for (int j = 0; j < numDocs; j++) {
+              Document doc = new Document();
+              String id = idx +"-"+j;
+              doc.add(new StringField("id", id, Field.Store.NO));
+              long seqNo = writer.addDocument(doc);
+              if (maxCompletedSeqID.get() < seqNo) {
+                long maxCompletedSequenceNumber = writer.getMaxCompletedSequenceNumber();
+                manager.maybeRefreshBlocking();
+                long prevValue;
+                while ((prevValue = maxCompletedSeqID.get()) < maxCompletedSequenceNumber) {
```
dnhatn, line 3992: maybe use `updateAndGet` instead of loop.

lucene/core/src/java/org/apache/lucene/index/DocumentsWriterDeleteQueue.java
```diff
@@ -538,17 +546,65 @@ public String toString() {
...
+    } else {
+      // if we haven't advanced the seqNo make sure we fall back to the previous queue
+      long value = previousMaxSeqId.getAsLong();
+      assert value <= startSeqNo : "illegal max sequence ID: " + value + " start was: " + startSeqNo;
+      return value;
+    }
+  }
+
+  /**
+   * Advances the queue to the next queue on flush. This carries over the the generation to the next queue and
```
dnhatn, line 578: s/the the/the

lucene/core/src/test/org/apache/lucene/index/TestIndexWriter.java
```diff
@@ -3949,4 +3950,69 @@ public void testRandomOperationsWithSoftDeletes() throws Exception {
...
+      assertEquals(4, writer.getMaxCompletedSequenceNumber());
+      // commit moves seqNo by 2 since there is one DWPT that could still be in-flight
+      assertEquals(6, writer.commit());
+      assertEquals(6, writer.getMaxCompletedSequenceNumber());
+      assertEquals(7, writer.addDocument(new Document()));
+      writer.getReader().close();
+      // getReader moves seqNo by 2 since there is one DWPT that could still be in-flight
+      assertEquals(9, writer.getMaxCompletedSequenceNumber());
+    }
+    try (Directory dir = newDirectory();
```
dnhatn, line 3971: I ran the second part of this test without the fix several thousand iterations, but all of them passed. It would be great if this test can reproduce the current issue.

lucene/core/src/java/org/apache/lucene/index/DocumentsWriterDeleteQueue.java
```diff
@@ -538,17 +546,65 @@ public String toString() {
...
+  /**
+   * Returns the maximum completed seq no for this queue.
+   */
+  long getMaxCompletedSeqNo() {
+    if (startSeqNo < nextSeqNo.get()) {
+      return getLastSequenceNumber();
+    } else {
+      // if we haven't advanced the seqNo make sure we fall back to the previous queue
+      long value = previousMaxSeqId.getAsLong();
+      assert value <= startSeqNo : "illegal max sequence ID: " + value + " start was: " + startSeqNo;
```
dnhatn, line 572: I think we can harden the assertion as `value` should always be smaller than `startSeqNo`.

lucene/core/src/java/org/apache/lucene/index/DocumentsWriterDeleteQueue.java
```diff
@@ -538,17 +546,65 @@ public String toString() {
...
+   * @return a new queue as a successor of this queue.
+   */
+  synchronized DocumentsWriterDeleteQueue advanceQueue(int maxNumPendingOps) {
+    if (advanced) {
+      throw new IllegalStateException("queue was already advanced");
+    }
+    advanced = true;
+    long seqNo = getLastSequenceNumber() + maxNumPendingOps + 1;
+    maxSeqNo = seqNo;
+    return new DocumentsWriterDeleteQueue(infoStream, generation + 1, seqNo + 1, () -> nextSeqNo.get() - 1);
```
dnhatn, line 594: Should `() -> nextSeqNo.get() - 1` be `this:: getMaxCompletedSeqNo` instead?

lucene/core/src/test/org/apache/lucene/index/TestIndexWriter.java
s1monw: 👍

lucene/core/src/test/org/apache/lucene/index/TestIndexWriter.java
s1monw: I added a very specific testcase that fails always. good call on pushing for a better test!

lucene/core/src/java/org/apache/lucene/index/DocumentsWriterDeleteQueue.java
s1monw: 👍

lucene/core/src/java/org/apache/lucene/index/DocumentsWriterDeleteQueue.java
s1monw: I left a comment why this is not a good idea.

dnhatn: Good point, thanks.

PR #1428, opened 2020-04-13
LUCENE-7788: fail precommit on unparameterised log.trace messages

ErickErickson:
This cleans up two directories. I have yet to go through all the changes with fresh eyes, but gradle check succeeds. NOTE: gradle does not yet include the checker program, nor will it until it runs cleanly.

I intend to push multiple batches of changes to the apache repo, there will be far too many to do all at once. Once that's done, I'll make the changes to Gradle precommit to fail on suspicious log messages.

So unless there are objections, I'll:
- run ant precommit and test on these changes.
- push these changes to Apache tomorrow morning.
- start on another batch after I push.
- rinse, repeat


ErickErickson:
Re-doing

PR #1432, opened 2020-04-14
SOLR-14404 CoreContainer level custom requesthandlers

noblepaul:


solr/core/src/java/org/apache/solr/api/ApiBag.java
```diff
@@ -134,6 +142,14 @@ static void registerIntrospect(List<String> l, PathTrie<Api> registry, Map<Strin
     registry.insert(copy, substitutes, introspect);
   }
 
+  public Api unregister(SolrRequest.METHOD method, String path) {
```
madrob, line 145: Should this be synchronized to match `register`?

solr/core/src/java/org/apache/solr/handler/admin/ContainerPluginsApi.java
```diff
@@ -0,0 +1,178 @@
...
+          payload.addError(info.name + " already exists");
+          return null;
+        }
+        map.put(info.name, info);
+        return map;
+      });
+    }
+
+    @Command(name = "remove")
+    public void remove(SolrQueryRequest req, SolrQueryResponse rsp, PayloadObj<String> payload) throws IOException {
```
madrob, line 97: Should this be METHOD.DELETE instead of a post?

solr/core/src/java/org/apache/solr/pkg/PackageListeners.java
```diff
@@ -63,13 +63,13 @@ public synchronized void removeListener(Listener listener) {
   }
 
   synchronized void packagesUpdated(List<PackageLoader.Package> pkgs) {
-    MDCLoggingContext.setCore(core);
```
madrob, line 66: Why do we need this check? setCore already handles nulls.

solr/core/src/test-files/runtimecode/sig.txt
```diff
@@ -69,6 +69,14 @@ openssl dgst -sha1 -sign ../cryptokeys/priv_key512.pem expressible.jar.bin | ope
 
 ZOT11arAiPmPZYOHzqodiNnxO9pRyRozWZEBX8XGjU1/HJptFnZK+DI7eXnUtbNaMcbXE2Ze8hh4M/eGyhY8BQ==
 
+openssl dgst -sha1 -sign priv_key512.pem containerplugin.v.1.jar.bin | openssl enc -base64 | sed 's/+/%2B/g' | tr -d \\n | sed
```
madrob, line 72: Does this need to be `../cryptokeys/priv_key512.pem`?

solr/core/src/test-files/runtimecode/MyPlugin.java
```diff
@@ -0,0 +1,43 @@
...
+import org.apache.solr.client.solrj.SolrRequest.METHOD;
+import org.apache.solr.core.CoreContainer;
+import org.apache.solr.request.SolrQueryRequest;
+import org.apache.solr.response.SolrQueryResponse;
+import org.apache.solr.security.PermissionNameProvider;
+
+@EndPoint(path = "/plugin/my/path",
+    method = METHOD.GET,
+    permission = PermissionNameProvider.Name.CONFIG_READ_PERM)
+public class MyPlugin {
```
madrob, line 31: I assume this is used to generate containerplugin.v.1.jar.bin and v.2? Should we have two source files for those? Can we do this some other way, besides checking in binaries to the repo?

solr/core/src/test/org/apache/solr/handler/TestContainerPlugin.java
```diff
@@ -0,0 +1,250 @@
...
+    @Command
+    public void read(SolrQueryRequest req, SolrQueryResponse rsp) {
+      rsp.add("testkey", "testval");
+    }
+
+  }
+
+
+  private void expectError(V2Request req, SolrClient client, String errPath, String expectErrorMsg) throws IOException, SolrServerException {
+    try {
```
madrob, line 242: use `expectThrows` here.

solr/core/src/test/org/apache/solr/handler/TestContainerPlugin.java
```diff
@@ -0,0 +1,250 @@
...
+  }
+
+  @After
+  public void teardown() {
+    System.clearProperty("enable.packages");
+  }
+
+  @Test
+  public void testApi() throws Exception {
+    MiniSolrCloudCluster cluster =
```
madrob, line 69: cluster setup and teardown can go in Before/After methods

solr/core/src/test/org/apache/solr/handler/admin/TestApiFramework.java
```diff
@@ -199,6 +200,25 @@ public void testPayload() {
 
   }
 
+  public void testApiWrapper() {
+    Class<ApiWithConstructor> klas = ApiWithConstructor.class;
```
madrob, line 204: unclear what this test is testing

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseHttpSolrClient.java
```diff
@@ -62,6 +63,9 @@ public static RemoteExecutionException create(String host, NamedList errResponse
       if (errObj != null) {
         Number code = (Number) getObjectByPath(errObj, true, Collections.singletonList("code"));
         String msg = (String) getObjectByPath(errObj, true, Collections.singletonList("msg"));
+        if(msg == null) msg = "";
```
madrob, line 66: There's already a null check later, what kinds of messages do we get here that are useful and don't leak too much internals?

solr/core/src/test/org/apache/solr/handler/TestContainerPlugin.java
```diff
@@ -0,0 +1,250 @@
...
+          .build()
+          .process(cluster.getSolrClient());
+      assertEquals(null, rsp._get("/plugin/testplugin/class", null));
+
+    } finally {
+      cluster.shutdown();
+    }
+  }
+  @Test
+  public void testApiFromPackage() throws Exception {
```
madrob, line 124: this is a fairly long test and was somewhat hard to follow, can we break it up into smaller pieces?

solr/core/src/java/org/apache/solr/core/CoreContainer.java
```diff
@@ -648,8 +655,8 @@ public void load() {
       pkiAuthenticationPlugin.initializeMetrics(solrMetricsContext, "/authentication/pki");
       TracerConfigurator.loadTracer(loader, cfg.getTracerConfiguratorPluginInfo(), getZkController().getZkStateReader());
       packageLoader = new PackageLoader(this);
-      containerHandlers.getApiBag().register(new AnnotatedApi(packageLoader.getPackageAPI().editAPI), Collections.EMPTY_MAP);
-      containerHandlers.getApiBag().register(new AnnotatedApi(packageLoader.getPackageAPI().readAPI), Collections.EMPTY_MAP);
+      containerHandlers.getApiBag().registerObject(packageLoader.getPackageAPI().editAPI);
```
madrob, line 658: There's a minor inconsistency here between edit/read and write/read, maybe we can standardize on one?

solr/core/src/java/org/apache/solr/handler/admin/ContainerPluginsApi.java
noblepaul: We could, but , this is the convention that we follow

solr/core/src/java/org/apache/solr/pkg/PackageListeners.java
noblepaul: because we are calling a 

`MDCLoggingContext.clear();` in finally

solr/core/src/test-files/runtimecode/MyPlugin.java
noblepaul: This is for documentation purposes for anyone who wants to know what the code looks like. Yes, we could possibly have 2 files, but then , they will have to be different files

> Can we do this some other way,

Unfortunately no. We need to have it compiled with an old version of java. It has to be precreated

>

solr/core/src/java/org/apache/solr/pkg/PackageListeners.java
madrob: I was under the impression that it setCore/clear still work even if the core was null.

solr/core/src/java/org/apache/solr/api/CustomContainerPlugins.java
```diff
@@ -0,0 +1,283 @@
...
+
+      String path = pluginNameVsPath.get(e.getKey());
+      if (path == null) {
+        // there is a new apiInfo . let's register it
+        try {
+          apiInfo.init();
+          ApiHolder holder = new ApiHolder(apiInfo);
+          plugins.put(holder.key, holder);
+          pluginNameVsPath.put(apiInfo.info.name, holder.key);
+          containerApiBag.register(holder, Collections.EMPTY_MAP);
```
madrob, line 117: prefer Collections.emptyMap()

solr/core/src/java/org/apache/solr/api/CustomContainerPlugins.java
```diff
@@ -0,0 +1,283 @@
...
+  public void writeMap(EntryWriter ew) {
+    plugins.forEach((s, apiHolder) -> ew.putNoEx(s, apiHolder.apiInfo));
+  }
+
+  public CustomContainerPlugins(CoreContainer coreContainer, ApiBag apiBag) {
+    this.coreContainer = coreContainer;
+    this.containerApiBag = apiBag;
+  }
+
+  public void refresh(Map<String, Object> pluginInfos) {
```
madrob, line 79: This parameter is unused?

solr/core/src/java/org/apache/solr/api/CustomContainerPlugins.java
```diff
@@ -0,0 +1,283 @@
...
+    private PackageLoader.Package.Version pkgVersion;
+    EndPoint endPoint;
+
+    private Class klas;
+
+
+    private AnnotatedApi delegate;
+
+
+    public ApiInfo(PluginMeta info, List<String> errs) {
```
madrob, line 201: This feels like a super klunky API writing errors to a string in a constructor argument? Why can't we just throw exceptions here?

solr/core/src/java/org/apache/solr/api/CustomContainerPlugins.java
```diff
@@ -0,0 +1,283 @@
...
+    public String key;
+    @JsonProperty
+    private final PluginMeta info;
+
+    @JsonProperty(value = "package")
+    public final String pkg;
+    private PackageLoader.Package.Version pkgVersion;
+    EndPoint endPoint;
+
+    private Class klas;
```
madrob, line 195: nit: raw types

solr/core/src/java/org/apache/solr/api/CustomContainerPlugins.java
```diff
@@ -0,0 +1,283 @@
...
+      return;
+    }
+    if(pluginInfos.isEmpty()) return;
+
+    for (Map.Entry<String, Object> e : pluginInfos.entrySet()) {
+      PluginMeta info = null;
+      try {
+        info = mapper.readValue(Utils.toJSON(e.getValue()), PluginMeta.class);
+      } catch (IOException ioException) {
+        log.error("Invalid apiInfo configuration :", ioException);
```
madrob, line 93: should this have a `continue`?

solr/core/src/java/org/apache/solr/api/CustomContainerPlugins.java
```diff
@@ -0,0 +1,283 @@
...
+  private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+
+  final CoreContainer coreContainer;
+  final ApiBag containerApiBag;
+  private Map<String, ApiHolder> plugins = new HashMap<>();
+  private Map<String, String> pluginNameVsPath = new HashMap<>();
+
+  @Override
+  public boolean onChange(Map<String, Object> properties) {
+    refresh(null);
```
madrob, line 65: `refresh(properties)`?

solr/core/src/java/org/apache/solr/api/CustomContainerPlugins.java
```diff
@@ -0,0 +1,283 @@
...
+        }
+      }
+    }
+    Set<String> toBeRemoved = new HashSet<>();
+    for (String s : pluginNameVsPath.keySet()) {
+      if (!pluginInfos.containsKey(s)) {
+        toBeRemoved.add(s);
+      }
+    }
+    for (String s : toBeRemoved) {
```
madrob, line 147: why do we need to iterate twice here?

solr/core/src/java/org/apache/solr/api/CustomContainerPlugins.java
```diff
@@ -0,0 +1,283 @@
...
+import org.apache.solr.pkg.PackageLoader;
+import org.apache.solr.request.SolrQueryRequest;
+import org.apache.solr.response.SolrQueryResponse;
+import org.apache.solr.util.SolrJacksonAnnotationInspector;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static org.apache.lucene.util.IOUtils.closeWhileHandlingException;
+
+public class CustomContainerPlugins implements MapWriter, ClusterPropertiesListener {
```
madrob, line 54: there's a lot of nested classes here, can any of those be split out into their own files?

solr/core/src/java/org/apache/solr/handler/admin/ContainerPluginsApi.java
```diff
@@ -0,0 +1,178 @@
...
+      Map<String, Object> clusterPropsJson = (Map<String, Object>) Utils.fromJSON(zkClient.getData(ZkStateReader.CLUSTER_PROPS, null, new Stat(), true));
+      return (Map<String, Object>) clusterPropsJson.computeIfAbsent(PLUGIN, Utils.NEW_LINKED_HASHMAP_FUN);
+    } catch (KeeperException.NoNodeException e) {
+      return new LinkedHashMap<>();
+    } catch (KeeperException | InterruptedException e) {
+      throw new IOException("Error reading cluster property", SolrZkClient.checkInterrupted(e));
+    }
+  }
+
+  private void persistPlugins(Function<Map, Map> modifier) throws IOException {
```
madrob, line 162: nit: raw types on Map

solr/core/src/java/org/apache/solr/api/CustomContainerPlugins.java
noblepaul: The objective is to collect multiple errors and eventually throw one single Exception whihc contains all of them

solr/core/src/test/org/apache/solr/handler/TestContainerPlugin.java
noblepaul: These clusters are created on a per testcase basis. We don't want remnants of other tests to screw up other tests

solr/core/src/java/org/apache/solr/pkg/PackageListeners.java
noblepaul: OK, got it. I misunderstood the comment

solr/core/src/test/org/apache/solr/handler/TestContainerPlugin.java
noblepaul: Splitting these tests will create another cluster and slow down the test execution further. I shall try to add inline comments to make it more clear as to what it is doing

solr/core/src/test/org/apache/solr/handler/TestContainerPlugin.java
noblepaul: `expectThrows` does not work here

solr/core/src/test/org/apache/solr/handler/admin/TestApiFramework.java
noblepaul: it got in there accidentally , removed

solr/core/src/java/org/apache/solr/api/CustomContainerPlugins.java
noblepaul: cleaned up

solr/core/src/java/org/apache/solr/api/CustomContainerPlugins.java
noblepaul: fixed

solr/core/src/java/org/apache/solr/pkg/PackageListeners.java
```diff
@@ -69,7 +69,7 @@ synchronized void packagesUpdated(List<PackageLoader.Package> pkgs) {
         invokeListeners(pkgInfo);
       }
     } finally {
-      MDCLoggingContext.clear();
+      if(core != null) MDCLoggingContext.clear();
```
dsmiley, line 72: No; it was correct before.  Internally, setCore can accept null and doesn't require the caller to care if core is null.  Furthermore, PackageListeners.core will never be null.  It ought to be declared final as well; it's only set in the constructor.

solr/core/src/test/org/apache/solr/handler/TestContainerPlugin.java
dsmiley: How so?  It returns the exception of the expected type and then you can inspect it as you are doing now.

solr/solrj/src/java/org/apache/solr/client/solrj/request/beans/PluginMeta.java
```diff
@@ -0,0 +1,54 @@
...
+ */
+
+package org.apache.solr.client.solrj.request.beans;
+
+import java.util.Objects;
+
+import org.apache.solr.common.annotation.JsonProperty;
+import org.apache.solr.common.util.ReflectMapWriter;
+
+public class PluginMeta implements ReflectMapWriter {
```
dsmiley, line 25: Add at least one line of javadoc for every public class

solr/solrj/src/java/org/apache/solr/common/util/PathTrie.java
```diff
@@ -82,6 +80,22 @@ public boolean add(String s) {
     return parts;
   }
 
+  public T unregister(List<String> path) {
```
dsmiley, line 83: I think the name "unregister" would only make sense if there was a corresponding "register" method.  How about a more generic "remove"?

noblepaul:
thanks @dsmiley , I shall fix them

solr/core/src/java/org/apache/solr/pkg/PackageListeners.java
noblepaul: > PackageListeners.core will never be null

Now it's not.

The schema plugins will no more be associated with a core. They will have it as `null`. I don't need to deal with it in this. So, I'll remove the null check now. `PackageListeners` is not used in ContainerPlugins

solr/solrj/src/java/org/apache/solr/common/util/PathTrie.java
noblepaul: yeah, `remove()` is better for `insert()`

solr/core/src/test/org/apache/solr/handler/TestContainerPlugin.java
noblepaul: fixed

noblepaul:
@dsmiley 
Yes, I want all our components to be written using

- Annotations (even our internal components)
- All input output to use POJOs

PR #1434, opened 2020-04-16
LUCENE-9324: Add an ID to SegmentCommitInfo

s1monw:
We already have IDs in SegmentInfo, as well as on SegmentInfos which are useful to uniquely identify segments and entire commits. Having IDs on SegmentCommitInfo is be useful too in
order to compare commits for equality and make snapshots incremental on generational files.
This change adds a unique ID to SegmentCommitInfo starting from Lucene 8.6. Older segments won't have an ID until the segment receives an update or a delete even if they have been opened and / or committed by Lucene 8.6 or above.


s1monw:
Note: there is still a NOCOMMIT in this pr regarding BWC. It's just an idea we can build on or even move to using `null` as an indicator that there is no id.

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
```diff
@@ -3081,7 +3081,7 @@ private SegmentCommitInfo copySegmentAsIs(SegmentCommitInfo info, String segName
                                           info.info.getUseCompoundFile(), info.info.getCodec(), 
                                           info.info.getDiagnostics(), info.info.getId(), info.info.getAttributes(), info.info.getIndexSort());
     SegmentCommitInfo newInfoPerCommit = new SegmentCommitInfo(newInfo, info.getDelCount(), info.getSoftDelCount(), info.getDelGen(),
-                                                               info.getFieldInfosGen(), info.getDocValuesGen());
+                                                               info.getFieldInfosGen(), info.getDocValuesGen(), info.getId());
```
mikemccand, line 3084: This happens during `IndexWriter.addIndexes(Directory[])` right?  I wonder whether we should give a new id instead of reusing the old one?  E.g. the segment (likely) now has a new name, and is in a different `Directory`, and is copied/forked from a prior segment, so maybe it should get a new `id`?

lucene/core/src/java/org/apache/lucene/index/SegmentCommitInfo.java
```diff
@@ -388,4 +399,17 @@ public SegmentCommitInfo clone() {
...
+
+  private void generationAdvanced() {
+    sizeInBytes = -1;
+    id = null;
+  }
+
+  public byte[] getId() {
+    if (id == null) {
+      // we advanced a generation - need to generate a new ID
+      id = StringHelper.randomId();
```
mikemccand, line 411: Do we need some thread safety here?

lucene/core/src/java/org/apache/lucene/index/SegmentInfos.java
```diff
@@ -374,7 +376,15 @@ public static final SegmentInfos readCommit(Directory directory, ChecksumIndexIn
...
         throw new CorruptIndexException("invalid deletion count: " + softDelCount + delCount + " vs maxDoc=" + info.maxDoc(), input);
       }
-      SegmentCommitInfo siPerCommit = new SegmentCommitInfo(info, delCount, softDelCount, delGen, fieldInfosGen, dvGen);
+      final byte[] sciId;
+      if (format > VERSION_74) {
+        sciId = new byte[StringHelper.ID_LENGTH];
+        input.readBytes(sciId, 0, sciId.length);
+      } else {
+        sciId = infos.id;
+        // NOCOMMIT can we do this? it would at least give us consistent BWC but we can't identify the same SCI in different commits
```
mikemccand, line 385: Good question ... maybe we could use `info.getId()`?   Then it'd be unique across `SCI`, but, shared across `SI` which is also weird.

lucene/core/src/java/org/apache/lucene/index/SegmentCommitInfo.java
s1monw: yeah good question, as far as I can tell we never read or write any of the member vars on this class unless we hold a lock that protects it. But it's a mess so I guess we should. Yet, if we do that we should make every method on this class synced no?

lucene/core/src/java/org/apache/lucene/index/SegmentInfos.java
s1monw: I don't think we should use info.getId() since that would mean that the same SegmentInfo instances are treated the same even if two IW made changes to it and it's generations. The way we have it now it's only considered the same if the overall commit is the same which is good i guess?

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
s1monw: we do share the `info.info.getId()` here as well so I think we should be consistent?

lucene/core/src/java/org/apache/lucene/index/SegmentCommitInfo.java
```diff
@@ -79,8 +85,7 @@
 
   /**
    * Sole constructor.
-   * 
-   * @param info
+   *  @param info
```
jpountz, line 88: nit:

```suggestion
   * @param info
```

lucene/core/src/java/org/apache/lucene/index/SegmentInfos.java
jpountz: When we introduced SegmentInfos#getId, we returned `null` as an ID for older segments. This is probably a safer option here as well, as callers can fall back to whatever behavior makes sense for them such as using a strong hash of the commit files as an ID, or re-downloading all files of the commit all the time and giving up incrementality?

s1monw: I think this case is a bit more complicated due to the changing nature of this ID. For each change (DV, llive docs, fields) we need to move to a new ID. Should we then just accept null and create a new ID once it changes or should we stick with `null` on these segments until they are written first time? Introducing `null` requires quite some changes in how we handle this which we can do, for sure. I still wonder if we can get away with stealing the _parent_ ID and have a smooth upgrade path.

s1monw: I pushed a new commit https://github.com/apache/lucene-solr/pull/1434/commits/f0a72f82bb17bd2582799aa25514ef764e012570 to address this.

jpountz: I don't understand the idea of stealing the parent ID, wouldn't it cause Lucene to consider commits equal when they are not, which would be a much worse problem than considering commits different when they are equal?

jpountz: Ah sorry I got confused because I thought that "parent" was referring to SegmentInfo (no s) instead of SegmentInfos, but I agree that SegmentInfos is not great either.

s1monw:
@mikemccand @jpountz  I pushed new commits that preserves the null value as much as necessary until the segment actually changes. This way we will be able to fall back to different methods to figure out if segments are the _same_. please take another look

lucene/core/src/java/org/apache/lucene/index/SegmentInfos.java
mikemccand: +1 for `null` and have caller deal with the BWC.  Stealing the `id` from another place (either "parent" `SegmentInfos` or the corresponding `SegmentInfo`) seems too dangerous.

lucene/core/src/java/org/apache/lucene/index/SegmentCommitInfo.java
mikemccand: I agree it's messy.  It's just that this sounds like a simple getter w/o side effect, yet, it does have an important and surprising side effect of setting the `id`.  It would be bad if two threads called it at once and got two different `id`s back.

If we really believe all callers are already sync'd (by IW's monitor lock), can we 1) add a comment stating this assumption, and 2) fix the assignment to do something like this instead:

```
  byte[] newID = StringHelper.randomId();
  assert id == null;
  id = newID;
```

?

That is still best effort, but might catch us in tests if we have cases where two threads really are invoking `getId` at once.

Or, maybe we can change this to assign the `id` at the same time as the SCI was changed (e.g. `delGen` increased), instead of at `get` time?

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
mikemccand: ok

lucene/backward-codecs/src/test/org/apache/lucene/index/TestBackwardsCompatibility.java
```diff
@@ -1367,6 +1368,16 @@ public void testIndexCreatedVersion() throws IOException {
     }
   }
 
+  public void testSegmentCommitInfoId() throws IOException {
+    for (String name : oldNames) {
+      Directory dir = oldIndexDirs.get(name);
+      SegmentInfos infos = SegmentInfos.readLatestCommit(dir);
+      for (SegmentCommitInfo info : infos) {
+        assertNull(info.getId());
```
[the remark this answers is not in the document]
s1monw: 👍

lucene/core/src/java/org/apache/lucene/index/SegmentInfos.java
```diff
@@ -374,7 +376,18 @@ public static final SegmentInfos readCommit(Directory directory, ChecksumIndexIn
       if (softDelCount + delCount > info.maxDoc()) {
         throw new CorruptIndexException("invalid deletion count: " + softDelCount + delCount + " vs maxDoc=" + info.maxDoc(), input);
       }
-      SegmentCommitInfo siPerCommit = new SegmentCommitInfo(info, delCount, softDelCount, delGen, fieldInfosGen, dvGen);
+      final byte[] sciId;
+      if (format > VERSION_74 ) {
+        if (input.readByte() == 1) {
```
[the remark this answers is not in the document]
s1monw: 👍

lucene/core/src/java/org/apache/lucene/index/SegmentCommitInfo.java
s1monw: I changed this to roll over to a new ID each time the generation is incremented. I don't really like that we call `StringHelper.randomId()` over and over again. I wonder if it would be better to set a new ID from the outside just before we clone in order to commit instead of doing it inside SegmentCommitInfo. This would make things more explicit IMO

lucene/core/src/java/org/apache/lucene/index/SegmentCommitInfo.java
```diff
@@ -77,9 +83,11 @@
   // this is never written to/read from the Directory
   private long bufferedDeletesGen = -1;
 
+  // is set once any of the generations has been advanced.
+  private boolean hasAdvanced;
```
mikemccand, line 87: We can remove this now?

PR #1435, opened 2020-04-16
SOLR-14410: Switch from SysV init script to systemd service file

mhitza:
# Description

Remove the init.d/solr SysV init script and use a systemd service file instead.

# Solution

I've tried not to diverge as much as possible from the way the installation script used to work before.

# Checklist


solr/solr-ref-guide/src/taking-solr-to-production.adoc
```diff
@@ -365,7 +373,7 @@ There is another issue once the heap reaches 32GB. Below 32GB, Java is able to u
...
+If your use case requires multiple instances, at a minimum you will need unique Solr home directories for each node you want to run; ideally, each home should be on a different physical disk so that multiple Solr nodes don’t have to compete with each other when accessing files on disk. Having different Solr home directories implies that you’ll need a different include file for each node. Moreover, if using the `/etc/systemd/system/solr.service` script to control Solr, then you’ll need a separate service for each node. The easiest approach is to use the service installation script to add multiple services on the same host, such as:
```
janhoy, line 376: Moreover, if using systemctl to control Solr... would be a better wording?

mhitza: Rephrased :+1:

mhitza:
@janhoy just a quick ping, the PR is ready for a new review

phils:
is there no way to progress this further?

epugh:
It's encouraging that there doesn't seem to be a lot of change here, reusing most of what we have.   

Does this have any other ramifications on what it touches?  I quickly looked at the `/solr/docker` dir in master, and that appears to start Solr in a different manner?

Is there any reasonable way of having a test for this for the future?   

Lastly, this change, is there consensus on it?  When I copied the Solr start up scripts over to the Tika project, I had read that systemd was the future, but since I knew the Solr ones, I modeled the Tika scripts on that.   I ask partly because I need to test the start up scripts for Tika 2.0 soon ;-)

janhoy:
I have understood it is the “future”, or rather the new normal. If I’m not mistaken, some people need to install some package before being able to use Solr’s current unit.d script to work?

mhitza:
@epugh The main difference I see between this service file and the docker configuration is that the docker container starts the service in foreground mode. This is to be understood, as docker containers run single services. And also because running systemd within docker is pretty hairy and platform-specific (can be done only if the host system is another system that has systemd available, or at least cgroups that need to be mounted readonly mode within the container).

When proposing this change we discussed on the mailing list, briefly, the option to start Solr in foreground mode (see [mailinglist thread](https://mail-archives.apache.org/mod_mbox/lucene-dev/202004.mbox/%3cCAFSzZzXS+zh1MrSCsjFtyXN0kOD_+6FJoBxd9Zhxt66fHaz6xg@mail.gmail.com%3e)) 

Regarding testing, the only option I can think of is via configuration management (e.g. ansible) targeting a VM (because docker is not as straightforward, as mentioned before, but doable).

@janhoy are you referring to the service file within this PR, or something else. As far as I know, all the common distros are running systemd, unless those peers are on old distros that are no longer maintained. And on systemd systems, there shouldn't be any extra package required to run this (except for the obvious JRE requirement for Solr itself)

epugh:
I'd be open to testing this out.   Also, is this more of a 9.0 thing versus a branch 8?   Seems like changing how you install Solr is a pretty big deal.   I've done the upgrade a few times using the old scripts, but this seems like a breaking change since you would lose the old way right?

mhitza:
The moment when I wrote this systemd service file I was running Solr 8, that is correct. I think it should work with Solr 9 as is.

From my memory when working on updating the installer I think you should be able to run the updated installer and it shouldn't lose the old way. Because on systemd systems all SysV scripts are overtaken/"wrapped around" by systemd. So for example, even if the previous docs stated commands like `service solr start`, it should work with `systemctl start solr` as is.

And you could try out a new installation of Solr using the installer `-s solr2` flag, for example. And then you could start the new service type with `systemctl start solr2` (of course you would need to stop the previous solr instance as they are running on the same port number).

What I haven't tested is what happens when you run the installer without any flags on a system that has already Solr installed. As it should generate a solr.service file (thus having the same service name), but *i think* it would supplant the SysV init script. As in, if a systemd and SysV service exist with the same name the physical solr.service would precede the SysV one.

janhoy:
> are you referring to the service file within this PR, or something else

I think you got me wrong - I said that systemd (the new script) should work ootb but the old initd style may require an extra package in modern Unix systems to even work (service command).

janhoy:
Hi @mhitza, this has fallen a bit behind. If you want to have it merged, please open a new PR against the new repo https://github.com/apache/solr, adding a link to this PR. Then I think there is not much work left to get it in to the 9.0 release.

itygh:

PR #1436, opened 2020-04-17
SOLR-14413: allow timeAllowed and cursorMark parameters

slackhappy:
# Description

[SOLR-14413](https://issues.apache.org/jira/browse/SOLR-14413): allow timeAllowed and cursorMark parameters.

Ever since cursorMarks were introduced in SOLR-5463 in 2014, cursorMark and timeAllowed parameters were not allowed in combination ("Can not search using both cursorMark and timeAllowed"), from [QueryComponent.java](https://github.com/apache/lucene-solr/blob/03363f413f2134594b012175deb3f10ec9384400/solr/core/src/java/org/apache/solr/handler/component/QueryComponent.java#L358):
```java
 if (null != rb.getCursorMark() && 0 < timeAllowed) {
  // fundamentally incompatible
  throw new SolrException(SolrException.ErrorCode.BAD_REQUEST, "Can not search using both " + CursorMarkParams.CURSOR_MARK_PARAM + " and " + CommonParams.TIME_ALLOWED);
}

```
While theoretically impure to use them in combination, it is often desirable to support cursormarks-style deep paging and attempt to protect Solr nodes from runaway queries using timeAllowed, in the hopes that most of the time, the query completes in the allotted time, and there is no conflict.

 
However if the query takes too long, it may be preferable to end the query and protect the Solr node and provide the user with a somewhat inaccurate sorted list. As noted in SOLR-6930, SOLR-5986 and others, timeAllowed is frequently used to prevent runaway load.  In fact, cursorMark and shards.tolerant are allowed in combination, so any argument in favor of purity would be a bit muddied in my opinion.

 

This was discussed once in the mailing list that I can find: https://mail-archives.apache.org/mod_mbox/lucene-solr-user/201506.mbox/%3C5591740B.4080807@elyograg.org%3E It did not look like there was strong support for preventing the combination.

 

I have tested cursorMark and timeAllowed combination together, and even when partial results are returned because the timeAllowed is exceeded, the cursorMark response value is still valid and reasonable.

# Solution

Remove the parameter check and assume that the implementer submitting the query accepts the tradeoff between accurate sorting, and timebounded queries.

# Tests

I removed the corresponding tests that checked the parameter combination - I didn't see any further insight in those tests as to why the combination wasn't allowed, so I assume they were just providing coverage of that check.

# Checklist

### DIFF
--- a/solr/core/src/test/org/apache/solr/CursorPagingTest.java
+++ b/solr/core/src/test/org/apache/solr/CursorPagingTest.java
@@ -25,6 +25,8 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.UUID;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 
 import org.apache.lucene.util.SentinelIntSet;
 import org.apache.lucene.util.TestUtil;
@@ -46,6 +48,7 @@ import org.junit.BeforeClass;
 import static org.apache.solr.common.params.CursorMarkParams.CURSOR_MARK_NEXT;
 import static org.apache.solr.common.params.CursorMarkParams.CURSOR_MARK_PARAM;
 import static org.apache.solr.common.params.CursorMarkParams.CURSOR_MARK_START;
+import static org.apache.solr.common.params.CommonParams.TIME_ALLOWED;
 import static org.apache.solr.common.util.Utils.fromJSONString;
 
 /**
@@ -103,13 +106,6 @@ public class CursorPagingTest extends SolrTestCaseJ4 {
                       "sort", "_docid_ asc, id desc", 
                       CURSOR_MARK_PARAM, CURSOR_MARK_START),
                ErrorCode.BAD_REQUEST, "_docid_");
-
-    // using cursor w/ timeAllowed
-    assertFail(params("q", "*:*", 
-                      "sort", "id desc", 
-                      CommonParams.TIME_ALLOWED, "1000",
-                      CURSOR_MARK_PARAM, CURSOR_MARK_START),
-               ErrorCode.BAD_REQUEST, CommonParams.TIME_ALLOWED);
 
     // using cursor w/ grouping
     assertFail(params("q", "*:*", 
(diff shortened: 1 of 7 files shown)


solr/core/src/test/org/apache/solr/CursorPagingTest.java
```diff
@@ -499,6 +492,61 @@ public void testSimple() throws Exception {
...
+  /**
+   * test that timeAllowed parameter can be used with cursors
+   * uses DelayingSearchComponent in solrconfig-deeppaging.xml
+   */
+  public void testTimeAllowed() throws Exception {
+    String wontExceedTimeout = "10000";
+    int numDocs = 100;
+    // Create a bunch of docs, inspired by createIndex in ExitableDirectoryReaderTest
+    for (int i = 0; i < numDocs; i++) {
+      assertU(adoc("id", Integer.toString(i), "name", "a" + i + " b" + i + " c" + i + " d"+i + " e" + i));
```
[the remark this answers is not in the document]
madrob: Alternatively, that method is `public static`, reuse it so that we don't have this logic duplicated all over the place.

solr/core/src/test/org/apache/solr/CursorPagingTest.java
```diff
@@ -499,6 +492,61 @@ public void testSimple() throws Exception {
...
+        "fl", "id",
+        "sort", "id desc",
+        "rows", "2",
+        "timeAllowed", wontExceedTimeout);
+    cursorMark = assertCursor(req(params, CURSOR_MARK_PARAM, cursorMark)
+        , "/response/numFound==100"
+    );
+    assertNotEquals(cursorMark, partialCursorMark);
+
+    cursorMark = assertCursor(req(params, CURSOR_MARK_PARAM, cursorMark)
```
madrob, line 544: this assignment is not used. Maybe we call another query and make sure we get the same, i.e. we're at the end?

solr/solr-ref-guide/src/pagination-of-results.adoc
```diff
@@ -97,6 +97,7 @@ There are a few important constraints to be aware of when using `cursorMark` par
 
 . `cursorMark` and `start` are mutually exclusive parameters.
 * Your requests must either not include a `start` parameter, or it must be specified with a value of "```0```".
+. When using the <<common-query-parameters.adoc#timeallowed-parameter,`timeAllowed` request param>>, partial results may be returned.  If time expires before the search is complete - as indicated when the `responseHeader` includes `"partialResults": true` - and `cursorMark` matches `nextCursorMark`, you cannot be sure that there are no more results. In this situation, consider increasing `timeAllowed` and reissuing the query.  When the `responseHeader` no longer includes `"partialResults": true` and `cursorMark` matches `nextCursorMark`, there are no more results.
```
madrob, line 100: 👍

solr/solr-ref-guide/src/common-query-parameters.adoc
```diff
@@ -206,7 +206,7 @@ The default value of this parameter is blank, which causes no extra "explain inf
 
 == timeAllowed Parameter
 
-This parameter specifies the amount of time, in milliseconds, allowed for a search to complete. If this time expires before the search is complete, any partial results will be returned, but values such as `numFound`, <<faceting.adoc#faceting,facet>> counts, and result <<the-stats-component.adoc#the-stats-component,stats>> may not be accurate for the entire result set. In case of expiration, if `omitHeader` isn't set to `true` the response header contains a special flag called `partialResults`.
+This parameter specifies the amount of time, in milliseconds, allowed for a search to complete. If this time expires before the search is complete, any partial results will be returned, but values such as `numFound`, <<pagination-of-results.adoc#using-cursors,`nextCursorMark`>>, <<faceting.adoc#faceting,facet>> counts, and result <<the-stats-component.adoc#the-stats-component,stats>> may not be accurate for the entire result set. In case of expiration, if `omitHeader` isn't set to `true` the response header contains a special flag called `partialResults`.
```
madrob, line 209: I don't think it's correct to say that nextCursorMark will be inaccurate. It will be accurate for the next result that should be returned.

This line also makes me think that maybe `timeAllowed+cursorMark+omitHeader` shouldn't be a valid combination. It's not a regression because `timeAllowed+cursorMark` is already disallowed. WDYT?

solr/core/src/test/org/apache/solr/CursorPagingTest.java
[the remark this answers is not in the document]
slackhappy: great point, I'll use ExitableDirectoryReaderTest.createIndex directly.  In my testing, it seemed to be useful to use the higher complexity prefix search to have a better guarantee of triggering partialResults with timeAllowed.  I suspect that is what ExitableDirectoryReaderTest to use that strategy, but to your point, `"a" + i` would be sufficient.

solr/core/src/test/org/apache/solr/CursorPagingTest.java
slackhappy: good idea, maybe I'll run the cursor to the end, adding up the returned docs along the way, making sure we got them all, which would verify the cursor behavior, taking inspiration from Bram's patch: https://issues.apache.org/jira/secure/attachment/13010039/SOLR-14413-bram.patch

solr/solr-ref-guide/src/common-query-parameters.adoc
slackhappy: I was just adding to the existing list there.  I'll pull it out into its own sentence.

In my opinion, `timeAllowed+cursorMark+omitHeader` should be allowed, and SOLR shouldn't be second-guessing the caller's request:

1. The combination `shards.tolerant+cursorMark` has the same issue, and `shards.tolerant+cursorMark+omitHeader` is allowed.
2. There may be a reasonable scenario that simply doesn't call for it, even if it is advisable to keep the header.  Let's say you are implementing a product catalog, and:
    - For whatever reason, you don't need the header, you rely on the http status code
    - You want to use cursorMarks to allow people to scroll through your fine list of shoes or whatever
    - You want to time-bound that resultset and protect your infrastructure from long-running queries
    - It just simply isn't mission critical for that every user to see every shoe.  If some rare expensive query or latency-producing event occurs, and truncates your shoe list, it is okay (you probably wouldn't bother retrying if you knew partialResults were true anyway).

My feeling is also a bit of a reaction to being annoyed about SolrCloud preventing us from creating collections without replicas, because we had legitimate reasons for standing up the replicas ourselves (thankfully, there is a workaround).


Then again, I say this as someone who doesn't have to answer Solr customer questions, so my words are just words.

madrob: Sure, if it's clearly documented that certain combinations of parameters lead to unknowable results, then I think it's fine. Maybe visit the `omitHeader` docs and put a blatant warning there about use in conjunction with shards.tolerant, timeAllowed, or any other partial result operations. (WAND?)

Hronom:
When this will be merged?

PR #1437, opened 2020-04-17
zkRun+https

madrob:
When starting an embedded zk server, if we can detect that we are using https, set the urlScheme cluster property accordingly.


solr/solrj/src/java/org/apache/solr/client/solrj/impl/SolrClientNodeStateProvider.java
```diff
@@ -398,6 +337,39 @@ public Map getZkJson(String path) throws KeeperException, InterruptedException {
...
+          while (t != null) {
+            if (t instanceof IOException) {
+              hasIOExceptionCause = true;
+              break;
+            }
+            t = t.getCause();
+          }
+
+          if (hasIOExceptionCause) {
+            log.info("Error on getting remote info, trying again: " + e.getMessage());
```
dsmiley, line 363: I recommend {} and passing "e".  (note Erick's efforts lately to tame the loggers)

PR #1439, opened 2020-04-19
tiny refguide tweaks

mkhludnev:
It doesn't work otherwise.

Hello, @ctargett I've tried to follow instruction, but turns out it incomplete and loose significant package.
Added one caveat and tweak formatting.


mkhludnev:
pushed https://gitbox.apache.org/repos/asf?p=lucene-solr.git;h=3a743ea

PR #1441, opened 2020-04-20
LUCENE-9332 validate source patterns using gradle

madrob:
If we have a gradle task with declared InputFiles and OutputFile then Gradle can figure out when the check is still up to date and when it needs to be rerun. Most of this is just a straight copy from the groovy script. This can take up to 20s off of precommit.

TODO:
* Figure out how to enable the ratDocument part of this, it was failing on imports for me.
* Split into modules so that we don't run the whole task each time
* Possibly split by file types?

The rat part is the main thing that I need help with, I haven't been able to figure out the right incantation of dependency declarations and build script declarations and whatever else we need to make it in scope. Any pointers appreciated.

### DIFF
--- a/gradle/validation/validate-source-patterns.gradle
+++ b/gradle/validation/validate-source-patterns.gradle
@@ -1,3 +1,5 @@
+import org.gradle.plugins.ide.eclipse.model.Output
+
 /*
  * Licensed to the Apache Software Foundation (ASF) under one or more
  * contributor license agreements.  See the NOTICE file distributed with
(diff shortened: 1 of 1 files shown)


gradle/validation/validate-source-patterns.gradle
```diff
@@ -1,3 +1,5 @@
+import org.gradle.plugins.ide.eclipse.model.Output
```
dweiss, line 1: Huh? Why Eclipse IDE import?

madrob:
> otherwise things will get out of sync
I'm hopeful that there is little change in the script until we can (soon) remove ant. Point taken, though.

> should operate at a project level and be applied to each project
Totally agree. I can tackle that after I get the imports working. Do you have any idea on why I'm not able to resolve the rat dependency here or what I need for it?

I tried playing around with various parameters on the ant task to push the output to a file and allow gradle daemon to know if this check was up to date or not, but wasn't able to do that. Really, I'd much prefer that alternative for now as it is much simpler, but wasn't able to make it work either so I started going down this path.

dweiss:
bq. Do you have any idea on why I'm not able to resolve the rat dependency here or what I need for it?

You removed the dependencies section... perhaps it's because of this? Otherwise I have no clue - should work.

I don't think it needs to be a "task" class either. You can declare inputs/outputs from within a script  (task) too; makes it simpler to experiment.

I also believe a task doesn't have to declare a physical output file to be made incremental - it will re-run when the inputs change. Look at ecj-lint.gradle - it has an example of this.

uschindler:
Hi, I will take some parts to my new pull request #1806. This one does not merge anymore, so it needs to be redone. The problem with RAT is already solved there (the dependencies need to be declared differently).

Closing this as #duplicate of #1806

PR #1444, opened 2020-04-22
LUCENE-9338: Clean up type safety in SimpleBindings

romseygeek:
Replaces SimpleBindings' `Map<String, Object>` with a map of 
`Supplier<DoubleValuesSource>` to improve type safety.  Also moves cycle
detection directly into ExpressionValuesSource.

### DIFF
--- a/lucene/expressions/src/java/org/apache/lucene/expressions/ExpressionValueSource.java
+++ b/lucene/expressions/src/java/org/apache/lucene/expressions/ExpressionValueSource.java
@@ -42,13 +42,17 @@ final class ExpressionValueSource extends DoubleValuesSource {
     this.expression = Objects.requireNonNull(expression);
     variables = new DoubleValuesSource[expression.variables.length];
     boolean needsScores = false;
-    for (int i = 0; i < variables.length; i++) {
-      DoubleValuesSource source = bindings.getDoubleValuesSource(expression.variables[i]);
-      if (source == null) {
-        throw new RuntimeException("Internal error. Variable (" + expression.variables[i] + ") does not exist.");
+    try {
+      for (int i = 0; i < variables.length; i++) {
+        DoubleValuesSource source = bindings.getDoubleValuesSource(expression.variables[i]);
+        if (source == null) {
+          throw new RuntimeException("Internal error. Variable (" + expression.variables[i] + ") does not exist.");
+        }
+        needsScores |= source.needsScores();
+        variables[i] = source;
       }
-      needsScores |= source.needsScores();
-      variables[i] = source;
+    } catch (StackOverflowError e) {
+      throw new IllegalArgumentException("Recursion error: Cycle detected originating in " + this);
     }
     this.needsScores = needsScores;
   }
--- a/lucene/expressions/src/java/org/apache/lucene/expressions/ExpressionValueSource.java
+++ b/lucene/expressions/src/java/org/apache/lucene/expressions/ExpressionValueSource.java
@@ -31,9 +31,8 @@ import org.apache.lucene.search.IndexSearcher;
 /**
  * A {@link DoubleValuesSource} which evaluates a {@link Expression} given the context of an {@link Bindings}.
  */
-@SuppressWarnings({"rawtypes", "unchecked"})
 final class ExpressionValueSource extends DoubleValuesSource {
-  final DoubleValuesSource variables[];
+  final DoubleValuesSource[] variables;
   final Expression expression;
   final boolean needsScores;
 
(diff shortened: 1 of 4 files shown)


lucene/expressions/src/java/org/apache/lucene/expressions/ExpressionValueSource.java
```diff
@@ -42,13 +42,17 @@
...
+        DoubleValuesSource source = bindings.getDoubleValuesSource(expression.variables[i]);
+        if (source == null) {
+          throw new RuntimeException("Internal error. Variable (" + expression.variables[i] + ") does not exist.");
+        }
+        needsScores |= source.needsScores();
+        variables[i] = source;
       }
-      needsScores |= source.needsScores();
-      variables[i] = source;
+    } catch (StackOverflowError e) {
```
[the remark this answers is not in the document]
romseygeek: I've reworked this so that instead of a `Supplier<DoubleValuesSource>` we store a `Function<Bindings, DoubleValuesSource>`, and supply a special `Bindings` implementation in `validate()` that checks for cycles.  Definitely much nicer, thanks for nudging me in the right direction!

romseygeek:
Tricksy, tricksy... I've updated the cycle detection logic to handle multiple levels of recursion, and as a bonus we get a nicer error message as well.

lucene/expressions/src/java/org/apache/lucene/expressions/SimpleBindings.java
```diff
@@ -96,24 +90,51 @@ public DoubleValuesSource getDoubleValuesSource(String name) {
...
+      throw new IllegalArgumentException("Invalid reference '" + name + "'");
+    }
+    return map.get(name).apply(this);
+  }
+
+  /**
+   * Traverses the graph of bindings, checking there are no cycles or missing references
+   * @throws IllegalArgumentException if the bindings is inconsistent
    */
   public void validate() {
```
msokolov, line 109: Have you considered returning the map, or an immutable view on it, so that callers can use this to enumerate all the dependencies? In a similar framework, I've found this to be pretty helpful for analyzing query patterns. It's also nice to know if the same name occurs multiple times in the dependency tree; maybe one should cache its value in that case.

romseygeek: Cacheing is a whole other conversation, which I think is related to the stuff that @mkhludnev is working on around grouping (in that we could plausibly have multiple references to the same iterator all moving in lockstep, where at the moment we pull separate iterators for each reference).  But I think that's for a follow-up really, this issue is just a bit of refactoring to improve type safety.

lucene/expressions/src/java/org/apache/lucene/expressions/SimpleBindings.java
```diff
@@ -96,24 +90,51 @@ public DoubleValuesSource getDoubleValuesSource(String name) {
...
+        throw new UnsupportedOperationException();
     }
   }
   
-  /** 
-   * Traverses the graph of bindings, checking there are no cycles or missing references 
-   * @throws IllegalArgumentException if the bindings is inconsistent 
+  @Override
+  public DoubleValuesSource getDoubleValuesSource(String name) {
+    if (map.containsKey(name) == false) {
```
madrob, line 99: Is `containsKey` boolean check more clear than `get` with a null check? I think the latter is going to be more efficient because it's only a single map operation, but I guess this way might be better for the JVM's escape analysis?

romseygeek: I'm pretty sure this won't be in the hot path - it's part of query setup, not query execution - and I think it reads more clearly.

lucene/expressions/src/java/org/apache/lucene/expressions/SimpleBindings.java
```diff
@@ -96,24 +90,51 @@ public DoubleValuesSource getDoubleValuesSource(String name) {
...
   public void validate() {
-    for (Object o : map.values()) {
-      if (o instanceof Expression) {
-        Expression expr = (Expression) o;
-        try {
-          expr.getDoubleValuesSource(this);
-        } catch (StackOverflowError e) {
-          throw new IllegalArgumentException("Recursion Error: Cycle detected originating in (" + expr.sourceText + ")");
-        }
+    for (String origin : map.keySet()) {
```
jpountz, line 110: nit: use entrySet() since you consume both keys and values?

lucene/expressions/src/test/org/apache/lucene/expressions/TestExpressionValidation.java
```diff
@@ -110,4 +110,15 @@ public void testCoRecursion4() throws Exception {
     });
     assertTrue(expected.getMessage().contains("Cycle detected"));
   }
+
+  public void testCoRecursion42() throws Exception {
```
jpountz, line 114: I provided this test but I don't think we should add it as it relies on iteration order and might get defeated on some JVMs or future versions of Java. I'd suggest to not add this test, or to change the map in SimpleBindings from a HashMap to a TreeMap to be able to better test such cases (in which case we'd have to swap `cycle0`/`cycle2`), this could be done in a follow-up too.

PR #1445, opened 2020-04-22
LUCENE-9339: Only call MergeScheduler when we actually found new merges

s1monw:
IW#maybeMerge calls the MergeScheduler even if it didn't find any merges we should
instead only do this if there is in-fact anything there to merge and safe the call
into a sync'd method.


s1monw:
@mikemccand maybe you can take a look? I think this API is expert enough to also break this in 8.6?

dweiss:
I never fully understood what this parameter would be useful for, actually. LGTM.

PR #1446, opened 2020-04-22
SOLR-14425: Aliases: ZK sync() needs to be synchronous

dsmiley:
https://issues.apache.org/jira/browse/SOLR-14425

CC @dragonsinth @tflobbe Does this make sense?

Separately, I'll file an issue to add CuratorFramework to SolrZkClient

### DIFF
--- a/solr/solrj/src/java/org/apache/solr/common/cloud/SolrZkClient.java
+++ b/solr/solrj/src/java/org/apache/solr/common/cloud/SolrZkClient.java
@@ -32,8 +32,12 @@ import java.lang.invoke.MethodHandles;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Path;
 import java.util.List;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.RejectedExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import java.util.function.BiFunction;
 import java.util.function.Function;
 import java.util.function.Predicate;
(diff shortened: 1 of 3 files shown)


solr/solrj/src/java/org/apache/solr/common/cloud/SolrZkClient.java
```diff
@@ -843,6 +847,31 @@ public void downloadFromZK(String zkPath, Path dir) throws IOException {
...
+   * it, after this method completes.
+   * This should not be called a lot; there may be some delay.  Consider alternative approaches It's often better to try to devise a way to
+   * "watch" for state changed instead of calling this.
+   */
+  public void sync(String path) {
+    // zookeeper.sync is asynchronous; we need to wait till it's done
+    CompletableFuture<Integer> future = new CompletableFuture<>();
+    keeper.sync(path, (rc, path1, ctx) -> future.complete(rc), null);
+    try {
+      Integer status = future.get(zkClientTimeout, TimeUnit.MILLISECONDS);
```
HoustonPutman, line 861: should this method return if the sync was successful or not?

dsmiley: I guess I could do that but it's only caller doesn't care.  It's a best-effort kind of thing.  I should add that to the javadocs; this method should not throw an exception.

dragonsinth: Dumb question from something who hasn't dug into ZK client code in Java in ages, but if you're waiting on the answer anyway, why use a future and then wait it instead of just calling a synchronous method to begin with?  Or is more like, "I'm waiting on potentially lots and lots of answers, I just need it to be up to date with any previous operations"?

dsmiley: Why synchronous method do you refer to?   If there was one that accomplishes this task here, I'd call it!  keeper.sync() is an asynchronous method.

dsmiley:
It appears I am mistaken about this whole thing: https://github.com/Netflix/curator/issues/194
Pretty non-obvious.  I can see why the Zk CLI for sync does wait for the callback but Solr needn't as this is done prior to a get.

PR #1448, opened 2020-04-22
LUCENE-9342: Collector's totalHitsThreshold should not be lower than numHits

tflobbe:
While looking at SOLR-13289 I noticed this situation. If I create a collector with `numHits` greater than `totalHitsThreshold`, and the number of hits in the query is somewhere between those two numbers, collector’s `totalHitRelation` will be `TotalHits.Relation.GREATER_THAN_OR_EQUAL_TO`, however I expect the count to be accurate in this situation, right? Should this be `TotalHits.Relation.EQUAL_TO`? Maybe make `totalHitsThreshold = Max(totalHitsThreshold, humHits)` on creation? Added a test that fails.

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -149,6 +149,9 @@ Improvements
 * LUCENE-9324: Add an ID to SegmentCommitInfo in order to compare commits for equality and make
   snapshots incremental on generational files. (Simon Willnauer, Mike Mccandless, Adrien Grant)
 
+* LUCENE-9342: TotalHits' relation will be EQUAL_TO when the number of hits is lower than TopDocsColector's numHits
+  (Tomás Fernández Löbbe)
+
 Optimizations
 ---------------------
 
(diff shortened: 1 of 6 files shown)


jpountz:
Since the contract is that Lucene might return GREATER_THAN_OR_EQUAL_TO if the hit count is greater than the threshold, both the current behavior and the behavior you suggest look legal to me. I don't have a strong preference. @jimczi As you've been integrating this feature into Elasticsearch, I wonder if you have an opinion?

tflobbe:
Yes, it's definitely legal, I'm just thinking it's unnecessarily more vague than it needs to be. Also, It's weird that someone would be passing parameters like this to Lucene, but may be coming from a default in a higher level app (Solr or Elasticsearch).

jpountz:
Indeed I think it can happen if you use a fixed threshold, say 100, and then perform pagination. If you have pages of 10 hits and go to the 11th page, shards would get passed numHits=110 and totalHitsThreshold=100. Like I said, I don't mind the change, if you'd like to make sure we always count at least `numHits` hits, I'd be happy to approve the PR.

jimczi:
> I don't mind the change, if you'd like to make sure we always count at least numHits hits, I'd be happy to approve the PR.

+1 too.

tflobbe:
Sounds good, I'll take a look.

PR #1451, opened 2020-04-23
LUCENE-9345: Separate MergeSchedulder from IndexWriter

s1monw:
This change extracts the methods that are used by MergeScheduler into
a MergeSource interface. This allows IndexWriter to better ensure
locking, hide internal methods and removes the tight coupling between the two
complex classes. This will also improve future testing.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/ConcurrentMergeScheduler.java
+++ b/lucene/core/src/java/org/apache/lucene/index/ConcurrentMergeScheduler.java
@@ -33,6 +33,7 @@ import org.apache.lucene.store.RateLimitedIndexOutput;
 import org.apache.lucene.store.RateLimiter;
 import org.apache.lucene.util.CollectionUtil;
 import org.apache.lucene.util.IOUtils;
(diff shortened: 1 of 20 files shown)


lucene/core/src/java/org/apache/lucene/index/ConcurrentMergeScheduler.java
```diff
@@ -516,18 +519,18 @@ public synchronized void merge(IndexWriter writer, MergeTrigger trigger) throws
 
     if (verbose()) {
       message("now merge");
-      message("  index: " + writer.segString());
+      message("  index(source): " + mergeSource.toString());
```
jpountz, line 522: did you mean to add `(source)` after `index` or is it a side-effect of a search/replace?

s1monw: I added it on purpose.

s1monw:
Thanks @jpountz - I was wondering if we should break this API in 8.6 already. it's very expert IMO. /cc @mikemccand

PR #1453, opened 2020-04-23
SOLR-14433: Improve SolrShardReporter default metrics list

dsmiley:
https://issues.apache.org/jira/browse/SOLR-14433#

Now includes TLOG and UPDATE./update.
These were small bugs to begin with but from user perspective this is an incremental improvement.

CC @sigram


sigram:
+1, LGTM.

PR #1456, opened 2020-04-24
SOLR-13289: Support for BlockMax WAND

tflobbe:
This PR adds support for BlockMax WAND via a `minExactHits` parameter. Hits will be counted accurately at least until this value, and above that, the count will be an approximation. In distributed search requests, the count will be per shard, so potentially the count will be accurately counted until `numShards * minExactHits`. The response will include the value `numFoundExact` which can be `true` (The value in `numFound` is exact) or `false` (the value in `numFound` is an approximation).
I plan to merge this to master and 8x as it is in this PR


jpountz:
Does minExactHits need to accept longs since the number of hits across shards might exceed Integer.MAX_VALUE?

Also I see you chose to make BMW an opt-in, I hope we'll find ways to enable it by default eventually. :)

tflobbe:
> Does minExactHits need to accept longs since the number of hits across shards might exceed Integer.MAX_VALUE?

Good point, I didn't think about the distributed aspect yet. However, does it make sense to you? Any number > Integer.MAX_VALUE means that each shard will reply with accurate count, right? so the sum of them all will also be accurate. 

> Also I see you chose to make BMW an opt-in, I hope we'll find ways to enable it by default eventually. :)
Yes, I think this should be default in master. I was planning in making it in another PR to make it explicit.

jpountz:
> Any number > Integer.MAX_VALUE means that each shard will reply with accurate count, right? so the sum of them all will also be accurate.

Yes indeed. The perspective I had was rather than since this parameter is a number of hits, as a user I would expect any legal number of hits to also be a legal value for this parameter. So it would be a shame to fail or worse silently cast to an int if a user passes 3B as a value? I believe it would be fine to keep it an integer internally but carefully accepting longs when parsing URL parameters and rounding to the nearest integer, ie. all value greater than Integer.MAX_VALUE would get converted to Integer.MAX_VALUE?

tflobbe:
ah, yes, that makes sense

solr/core/src/java/org/apache/solr/search/MaxScoreCollector.java
```diff
@@ -39,7 +39,7 @@ public float getMaxScore() {
   public ScoreMode scoreMode() {
     // Should be TOP_SCORES but this would wrap the scorer unnecessarily since
     // this collector is only used in a MultiCollector.
-    return ScoreMode.COMPLETE;
+    return ScoreMode.TOP_SCORES;
```
tflobbe, line 42: @jpountz, What did you mean with this comment? MultiCollector will set the `scoreMode` to `COMPLETE` in the case of the main collector being something other than `TOP_SCORES`, right?

solr/core/src/java/org/apache/solr/handler/component/QueryComponent.java
```diff
@@ -401,6 +403,14 @@ public void process(ResponseBuilder rb) throws IOException
     doProcessUngroupedSearch(rb, cmd, result);
   }
 
+  private int getMinExactHits(SolrParams params) {
+    long minExactHits = params.getLong(CommonParams.MIN_EXACT_HITS, Integer.MAX_VALUE);
```
tflobbe, line 407: I plan to move the default value to a solrconfig config in a future PR, for now, MAX_VALUE (disabled)

solr/core/src/java/org/apache/solr/search/MaxScoreCollector.java
jpountz: Yeah I think that at some point we tried not to wrap the scorer when all score modes were `COMPLETE` but apparently now we do, so this comment is stale. https://github.com/apache/lucene-solr/blob/master/lucene/core/src/java/org/apache/lucene/search/MultiCollector.java#L158

solr/core/src/java/org/apache/solr/response/GeoJSONResponseWriter.java
```diff
@@ -292,11 +293,18 @@ else if(geo instanceof WriteableGeoJSON) {
     }
   }
 
+  @Deprecated
   @Override
   public void writeStartDocumentList(String name, 
-      long start, int size, long numFound, Float maxScore) throws IOException
+      long start, int size, long numFound, Float maxScore) throws IOException {
+    throw new UnsupportedOperationException();
```
madrob, line 300: Why? We can't pick a default and still use this?

solr/core/src/java/org/apache/solr/response/JSONWriter.java
```diff
@@ -132,11 +133,16 @@ public void writeSolrDocument(String name, SolrDocument doc, ReturnFields return
   //       that the size could not be reliably determined.
   //
 
+  /**
+   * This method will be removed in Solr 9
```
madrob, line 137: do you mean solr 10? we don't need a major version with it deprecated before we can jettison it? folks upgrading from 8.5->9 might be surprised by this.

solr/core/src/java/org/apache/solr/search/MaxScoreCollector.java
```diff
@@ -37,9 +37,7 @@ public float getMaxScore() {
 
   @Override
   public ScoreMode scoreMode() {
-    // Should be TOP_SCORES but this would wrap the scorer unnecessarily since
-    // this collector is only used in a MultiCollector.
-    return ScoreMode.COMPLETE;
+    return ScoreMode.TOP_SCORES;
```
madrob, line 40: why can we make this change?

solr/core/src/java/org/apache/solr/search/QueryResultKey.java
```diff
@@ -65,6 +70,7 @@ public QueryResultKey(Query query, List<Query> filters, Sort sort, int nc_flags)
       h = h*29 + sf.hashCode();
       ramSfields += BASE_SF_RAM_BYTES_USED + RamUsageEstimator.sizeOfObject(sf.getField());
     }
+    h = h*31 + minExactHits;
```
madrob, line 73: this is going to overflow in the default case, is that ok?

solr/core/src/test/org/apache/solr/request/TestFaceting.java
```diff
@@ -931,5 +934,28 @@ public void testListedTermCounts() throws Exception {
...
+    ModifiableSolrParams params = new ModifiableSolrParams();
+    params.set("q", "title_ws:Book1");
+    params.set(FacetParams.FACET, "true");
+    params.set(FacetParams.FACET_FIELD, "title_ws");
+    assertQ(req(params),
+        "//lst[@name='facet_fields']/lst[@name='title_ws']/int[1][@name='Book1'][.='20']"
+        ,"//*[@hitCountRelation='" + HitCountRelation.EQ + "']"
+        ,"//*[@numFound='" + NUM_DOCS + "']");
+    
+    // It doesn't matter if we request munExactHits, when requesting facets, the numFound value is precise
```
madrob, line 954: s/mun/min

solr/core/src/java/org/apache/solr/response/GeoJSONResponseWriter.java
tflobbe: This method Shouldn't be called at all for writing a response. In other ResponseWriters, I left the code untouched for support in case someone is extending the Writer. In this case, since it's an inner class, I decided to throw an exception to catch potential bugs (internal code calling this)

solr/core/src/java/org/apache/solr/response/JSONWriter.java
tflobbe: My understanding is that it's OK to change methods across major versions? Unless it's something needed for rolling upgrade or something.

solr/core/src/java/org/apache/solr/search/MaxScoreCollector.java
tflobbe: There is a previous discussion in this PR about it:
>  tflobbe:
> @jpountz, What did you mean with this comment? MultiCollector will set the scoreMode to COMPLETE in the case of the main collector being something other than TOP_SCORES, right?
> jpountz:
> Yeah I think that at some point we tried not to wrap the scorer when all score modes were COMPLETE but apparently now we do, so this comment is stale. https://github.com/apache/lucene-solr/blob/master/lucene/core/src/java/org/apache/lucene/search/MultiCollector.java#L158

solr/core/src/java/org/apache/solr/search/QueryResultKey.java
tflobbe: ah! Good question. I guess yes?

solr/solrj/src/test/org/apache/solr/common/util/TestJavaBinCodec.java
```diff
@@ -242,6 +243,7 @@ public void testBackCompatForSolrDocumentWithChildDocs() throws IOException {
   }
 
   @Test
+  @Ignore("This test compares binaries, which change due to SOLR-13289")
```
anshumg, line 246: s/change/changed

solr/core/src/java/org/apache/solr/handler/component/QueryComponent.java
```diff
@@ -401,6 +403,14 @@ public void process(ResponseBuilder rb) throws IOException
...
+    if (minExactHits < 0 || minExactHits > Integer.MAX_VALUE) {
```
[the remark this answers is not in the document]
tflobbe: This is intentional. See the discussion with Adrien in this same PR. We allow longs just because the `minExactHits` is in relation to the `numFound`, which is a long, however, any value greater than `Integer.MAX_VALUE` doesn't make sense, since Lucene doesn't allow more than that in a single shard. In case of distributed queries, the `minExactHits` is used by every shard

solr/core/src/java/org/apache/solr/response/JSONWriter.java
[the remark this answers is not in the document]
tflobbe: I really don't think we need to keep this around longer than 9.0. People upgrading a major version expect to have to change some code. This is a trivial change, they'll get a compile-time error and notice there is a new parameter added. This doesn't even change the client, just plugins.

solr/core/src/test/org/apache/solr/request/TestFaceting.java
```diff
@@ -931,5 +934,28 @@ public void testListedTermCounts() throws Exception {
         "//lst[@name='facet_fields']/lst[@name='title_ws']/int[2][@name='Book2']",
         "//lst[@name='facet_fields']/lst[@name='title_ws']/int[3][@name='Book3']");
   }
+  
+  @Test
+  public void testFacetCountsWithMinExactHits() throws Exception {
+    final int NUM_DOCS = 20;
+    for (int i = 0; i < NUM_DOCS ; i++) {
+      assertU(adoc("id", String.valueOf(i), "title_ws", "Book1"));
+      assertU(commit());
```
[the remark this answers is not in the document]
tflobbe: Actually, I wanted to have multiple segments. I could do something like "sometimes()", but since the number of docs is low, I didn't think it was needed to add any randomization or more complex logic.

solr/core/src/test/org/apache/solr/search/SolrIndexSearcherTest.java
```diff
@@ -0,0 +1,200 @@
...
+public class SolrIndexSearcherTest extends SolrTestCaseJ4 {
+  
+  private final static int NUM_DOCS = 20;
+
+  @BeforeClass
+  public static void setUpClass() throws Exception {
+    initCore("solrconfig.xml", "schema.xml");
+    for (int i = 0 ; i < NUM_DOCS ; i ++) {
+      assertU(adoc("id", String.valueOf(i), "field1_s", "foo", "field2_s", String.valueOf(i % 2), "field3_s", String.valueOf(i)));
+      assertU(commit());
```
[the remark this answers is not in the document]
tflobbe: Same as above, wanted multiple segments

solr/core/src/java/org/apache/solr/search/QueryResultKey.java
tflobbe: Added a test

solr/solrj/src/test/org/apache/solr/common/util/TestJavaBinCodec.java
tflobbe: I fixed this test.

PR #1461, opened 2020-04-27
LUCENE-9348: Add a base grouping test for use with different GroupSelector implementations

romseygeek:
The grouping module tests currently all try and test both grouping by term and 
grouping by ValueSource. They are quite difficult to follow, however, and it is not 
at all easy to add tests for a new grouping type. This commit adds a new
BaseGroupSelectorTestCase class which can be extended to test particular
GroupSelector implementations, and adds tests for TermGroupSelector and
ValueSourceGroupSelector.


romseygeek:
cc @martijnvg @dsmiley 

We should be able to remove `AllGroupHeadsCollectorTest`, `AllGroupsCollectorTest` and `GroupingSearchTest` once this is merged, and possibly `TestGrouping` as well.  It should also make it much easier to add groupers based on Double/LongValuesSource, with the aim of moving the `ValueSourceGroupSelector` into Solr and deprecating ValueSource within lucene.

PR #1463, opened 2020-04-27
SOLR-14440 Cert Auth plugin

madrob:
https://issues.apache.org/jira/browse/SOLR-14440

### DIFF
--- a/solr/CHANGES.txt
+++ b/solr/CHANGES.txt
@@ -10,7 +10,7 @@ Consult the LUCENE_CHANGES.txt file for additional, low level, changes in this r
 
 New Features
 ---------------------
-(No changes)
+* SOLR-14440: Introduce new Certificate Authentication Plugin to load Principal from certificate subject. (Mike Drob)
 
 Improvements
 ----------------------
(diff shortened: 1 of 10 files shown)


noblepaul:
LGTM

madrob:
> Would it be useful to be able to configure from where to pull the username that will be passed in to Solr

I think so, but that is probably better with something like SOLR-10814 in place so that we can keep the full subject name and also define a "short name".

> I'm not familiar with the Principal class being wrapped here so maybe it has what we need, but as I understand there is more than one way to add a userid to a cert.

The principal class works about how you'd expect w.r.t. `getName()`, which is what RuleBasedAuthorizationPlugin uses, and will return something like `CN=Solr User,OU=Engineering,O=Example Inc.,C=US`. Added that to the docs to clarify.

> Wrt Admin UI I suspect it to "just work" if you have the right cert. 

Good ideas, I'll look into this and see if we need to make additions.

madrob:
> But check what it says if you do not have right ssl cert. Ideally it should show the Login screen with a helpful message that Cert Auth is used and you need to provide correct cert.

It does this now! Great suggestion.

janhoy:
Looks good, thanks

PR #1464, opened 2020-04-28
LUCENE-9087: Build always trees with full leaves and lower the default value for maxPointsPerLeafNode

iverase:
This commit changes the logic used to build BKD trees to always construct trees with full leaves (except the last one). This change gives more control in the expected behaviour of the tree. In addition the special logic the we have for 1D trees constructs the same type of trees, therefore removing some discrepancy.

In addition, this commit lowers the default for maxPointsPerLeafNode from 1024 to 512 and simplifies the logic for rotating tree leaves.


lucene/core/src/java/org/apache/lucene/util/bkd/BKDWriter.java
```diff
@@ -730,6 +730,22 @@ private void rotateToTree(int nodeID, int offset, int count, byte[] index, List<
     return newLeafBlockFPs;
   }
 
+  private int getNumLeftLeaveNodes(int numLeaves) {
+    // return the max level for this number of leaves. If level is full it returns the next level
+    int maxLevel = 32 - Integer.numberOfLeadingZeros(numLeaves);
```
jpountz, line 691: is there an off-by-one error? I'd expect maxLevel to be `ceil(log2(numLeaves))`, which would be `32 - Integer.numberOfLeadingZeros(numLeaves-1)` according to https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Integer.html#numberOfLeadingZeros(int)

iverase: I have rewritten the method (and fix some grammatical mistakes) to main more clear. I think the best way is to start calculating the full level which can be done using:

```
int lastFullLevel = 31 - Integer.numberOfLeadingZeros(numLeaves);
```

Then it is clear what we are doing.

jpountz: this makes sense to me

lucene/core/src/java/org/apache/lucene/util/bkd/BKDWriter.java
```diff
@@ -822,39 +804,27 @@ public long finish(IndexOutput out) throws IOException {
...
-            leafBlockFPs = newLeafBlockFPs;
-          }
-          break;
-        }
-
-        levelCount *= 2;
-      }
+    int lastFullLevel = 31 - Integer.numberOfLeadingZeros(numLeaves);
+    int leavesFullLevel = 1 << lastFullLevel;
+    int unbalancedLeafNodes = 2 * (numLeaves - leavesFullLevel);
```
jpountz, line 819: In `getNumLeftLeafNodes` above we do `unbalancedLeafNodes = numLeaves - leavesFullLevel`, which has a factor 2 difference with the above formula. I think I understand the reason but can me make it less confusing and avoid naming the same way different quantities?

lucene/core/src/java/org/apache/lucene/util/bkd/BKDWriter.java
```diff
@@ -420,16 +421,13 @@ private long writeFieldNDims(IndexOutput out, String fieldName, MutablePointValu
...
-    long countPerLeaf = pointCount = values.size();
-    long innerNodeCount = 1;
+    pointCount = values.size();
 
-    while (countPerLeaf > maxPointsInLeafNode) {
-      countPerLeaf = (countPerLeaf+1)/2;
-      innerNodeCount *= 2;
+    int numLeaves = Math.toIntExact(pointCount / maxPointsInLeafNode);
+    if (pointCount % maxPointsInLeafNode != 0) {
+      numLeaves++;
```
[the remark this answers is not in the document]
iverase: Yes, I agree it seems a better approach.

lucene/core/src/java/org/apache/lucene/util/bkd/BKDWriter.java
iverase: renamed to `leavesPartialLevel`

PR #1465, opened 2020-04-28
LUCENE-9349: TermInSetQuery should use consumeMatchingTerms in visit()

romseygeek:
TermInSetQuery currently iterates through all its prefix-encoded terms 
in order to build an array to pass back to its visitor when visit() is called.  
This seems like a waste, particularly when the visitor is not actually 
consuming the terms (for example, when doing a clause-count check 
before executing a search).  This commit changes TermInSetQuery to use 
consumeTermsMatching(), and also changes the signature of this method so 
that it takes a BytesRunAutomaton supplier to allow for lazy instantiation.  In
addition, IndexSearcher's clause count check wasn't counting leaves that
called consumeTermsMatching().

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -142,6 +142,11 @@ API Changes
   interface that offers the basic methods to acquire pending merges, run the merge and do accounting 
   around it. (Simon Willnauer)
 
+* LUCENE-9349: QueryVisitor.consumeTermsMatching() now takes a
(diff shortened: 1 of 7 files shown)


romseygeek:
This will also make FuzzyQuery's visit() method nicer if we switch back to not holding its automaton directly on the query.

mkhludnev:
ok. +1

PR #1467, opened 2020-04-29
LUCENE-9350: Don't hold references to large automata on FuzzyQuery

romseygeek:
LUCENE-9068 moved fuzzy automata construction into FuzzyQuery itself.  However,
this has the nasty side-effect of blowing up query caches that expect queries to be 
fairly small.  This commit restores the previous behaviour of caching the large automata
on an AttributeSource shared between segments, while making the construction a
bit clearer by factoring it out into a package-private `FuzzyAutomatonBuilder`.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/FuzzyAutomatonBuilder.java
+++ b/lucene/core/src/java/org/apache/lucene/search/FuzzyAutomatonBuilder.java
@@ -1,0 +1,88 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.lucene.search;
+
+import org.apache.lucene.util.UnicodeUtil;
+import org.apache.lucene.util.automaton.CompiledAutomaton;
+import org.apache.lucene.util.automaton.LevenshteinAutomata;
+import org.apache.lucene.util.automaton.TooComplexToDeterminizeException;
+
+/**
+ * Builds a set of CompiledAutomaton for fuzzy matching on a given term,
+ * with specified maximum edit distance, fixed prefix and whether or not
+ * to allow transpositions.
+ */
+class FuzzyAutomatonBuilder {
+
+  private final String term;
+  private final int maxEdits;
+  private final LevenshteinAutomata levBuilder;
+  private final String prefix;
+  private final int termLength;
+
+  FuzzyAutomatonBuilder(String term, int maxEdits, int prefixLength, boolean transpositions) {
+    if (maxEdits < 0 || maxEdits > LevenshteinAutomata.MAXIMUM_SUPPORTED_DISTANCE) {
+      throw new IllegalArgumentException("max edits must be 0.." + LevenshteinAutomata.MAXIMUM_SUPPORTED_DISTANCE + ", inclusive; got: " + maxEdits);
+    }
+    if (prefixLength < 0) {
+      throw new IllegalArgumentException("prefixLength cannot be less than 0");
+    }
+    this.term = term;
+    this.maxEdits = maxEdits;
+    int[] codePoints = stringToUTF32(term);
+    this.termLength = codePoints.length;
+    prefixLength = Math.min(prefixLength, codePoints.length);
+    int[] suffix = new int[codePoints.length - prefixLength];
+    System.arraycopy(codePoints, prefixLength, suffix, 0, suffix.length);
+    this.levBuilder = new LevenshteinAutomata(suffix, Character.MAX_CODE_POINT, transpositions);
+    this.prefix = UnicodeUtil.newString(codePoints, 0, prefixLength);
+  }
+
+  CompiledAutomaton[] buildAutomatonSet() {
+    CompiledAutomaton[] compiled = new CompiledAutomaton[maxEdits + 1];
+    for (int i = 0; i <= maxEdits; i++) {
+      try {
+        compiled[i] = new CompiledAutomaton(levBuilder.toAutomaton(i, prefix), true, false);
+      }
+      catch (TooComplexToDeterminizeException e) {
+        throw new FuzzyTermsEnum.FuzzyTermsException(term, e);
+      }
+    }
+    return compiled;
+  }
+
+  CompiledAutomaton buildMaxEditAutomaton() {
+    try {
+      return new CompiledAutomaton(levBuilder.toAutomaton(maxEdits, prefix), true, false);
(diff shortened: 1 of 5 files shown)


lucene/core/src/java/org/apache/lucene/search/FuzzyTermsEnum.java
```diff
@@ -364,4 +325,60 @@ public BytesRef term() throws IOException {
...
+      this.automata = builder.buildAutomatonSet();
+    }
+
+    @Override
+    public void clear() {
+      this.automata = null;
+    }
+
+    @Override
+    public void reflectWith(AttributeReflector reflector) {
```
dsmiley, line 374: Why not implement these like they are done on 8x?  I suspect introspection tools like "luke" / Solr analysis page may choke if this isn't done right.

romseygeek: It's a private implementation and only gets used internally to FuzzyTermsEnum, so I don't see how the analysis page or luke would get presented with one of these?

lucene/core/src/java/org/apache/lucene/search/FuzzyQuery.java
```diff
@@ -237,22 +216,9 @@ public boolean equals(Object obj) {
...
-      return false;
-    if (transpositions != other.transpositions)
-      return false;
-    if (term == null) {
-      if (other.term != null)
-        return false;
-    } else if (!term.equals(other.term))
-      return false;
-    return true;
+    return Objects.equals(maxEdits, other.maxEdits) && Objects.equals(prefixLength, other.prefixLength)
```
dsmiley, line 219: nitpick: compare the term first, as it is most likely to return false for different fuzzy

lucene/core/src/java/org/apache/lucene/search/FuzzyAutomatonBuilder.java
```diff
@@ -0,0 +1,88 @@
...
+        compiled[i] = new CompiledAutomaton(levBuilder.toAutomaton(i, prefix), true, false);
+      }
+      catch (TooComplexToDeterminizeException e) {
+        throw new FuzzyTermsEnum.FuzzyTermsException(term, e);
+      }
+    }
+    return compiled;
+  }
+
+  CompiledAutomaton buildMaxEditAutomaton() {
```
madrob, line 69: I'm confused about the difference between when we would want the full automaton set and when we want the max edit automaton. When is one useful but not the other? Is this a simple optimization to skip building the relatively inexpensive (exponentially less expensive, even) fewer edit automata?

romseygeek: Just the max is useful for highlighting or for other fuzzy searches that don't do the TopTerms pruning (we use it for a fuzzy interval query in elasticsearch, for example)

lucene/core/src/java/org/apache/lucene/search/FuzzyQuery.java
```diff
@@ -183,7 +162,7 @@ public void visit(QueryVisitor visitor) {
       if (maxEdits == 0 || prefixLength >= term.text().length()) {
         visitor.consumeTerms(this, term);
       } else {
-        automata[automata.length - 1].visit(visitor, this, field);
+        visitor.consumeTermsMatching(this, term.field(), () -> getAutomata().runAutomaton);
```
madrob, line 165: Does this still build the automata each time?

lucene/core/src/java/org/apache/lucene/search/FuzzyTermsEnum.java
```diff
@@ -88,43 +89,44 @@
...
-    this.term = term;
     this.atts = atts;
-    this.termLength = termLength;
+    this.term = term;
 
     this.maxBoostAtt = atts.addAttribute(MaxNonCompetitiveBoostAttribute.class);
     this.boostAtt = atts.addAttribute(BoostAttribute.class);
 
-    this.automata = automata;
+    atts.addAttributeImpl(new AutomatonAttributeImpl());
```
madrob, line 123: The java doc on this suggest that we need to be using an AttributeFactory, but that might be overkill here?

lucene/core/src/java/org/apache/lucene/search/FuzzyTermsEnum.java
```diff
@@ -88,43 +89,44 @@
...
+    AutomatonAttribute aa = atts.addAttribute(AutomatonAttribute.class);
```
madrob, line 124: why not getAttribute? If we're sharing this across segments, then I think we want to reuse the same one each time, right?

lucene/core/src/java/org/apache/lucene/search/FuzzyQuery.java
```diff
@@ -193,7 +172,7 @@ protected TermsEnum getTermsEnum(Terms terms, AttributeSource atts) throws IOExc
     if (maxEdits == 0 || prefixLength >= term.text().length()) {  // can only match if it's exact
       return new SingleTermsEnum(terms.iterator(), term.bytes());
     }
-    return new FuzzyTermsEnum(terms, atts, getTerm(), termLength, maxEdits, automata);
+    return new FuzzyTermsEnum(terms, atts, getTerm(), maxEdits, prefixLength, transpositions);
```
madrob, line 175: do we need to update the java docs on multi term query to reflect that fuzzy query also uses the attribute source now as well?

lucene/core/src/java/org/apache/lucene/search/FuzzyTermsEnum.java
romseygeek: `addAttribute` will return the already existing attribute if it's there, but create one if it isn't.  `getAttribute` won't work here because it won't do the create part.

lucene/core/src/java/org/apache/lucene/search/FuzzyTermsEnum.java
romseygeek: I don't think we can use an AttributeFactory here because the attribute source has already been created.  This seems to work fine though - @thetaphi can probably tell me if I've done something wrong?

lucene/core/src/java/org/apache/lucene/search/FuzzyQuery.java
romseygeek: Good idea, will update

lucene/core/src/java/org/apache/lucene/search/FuzzyTermsEnum.java
thetaphi: Uh. You sure you tagged the right sentient here? xD

madrob: @uschindler

lucene/core/src/java/org/apache/lucene/search/FuzzyQuery.java
romseygeek: Only if the visitor implementation actually needs it, we're passing a `Supplier` now.

romseygeek:
I think this is ready @madrob ?

lucene/core/src/java/org/apache/lucene/search/FuzzyTermsEnum.java
uschindler: I think that's fine here. There is only one implementation that's all internally used.

lucene/core/src/java/org/apache/lucene/search/FuzzyTermsEnum.java
uschindler: GetAttribute is also fine, because you added the impl a line before. If the attribute wouldn't be there, adding it won't work, because it's all private. So all: That's just a matter of preference. I'd check how we do this at other places.

PR #1469, opened 2020-04-30
SOLR-14770 Avoid reregistering JVM Guage as well

madrob:


dsmiley:
@sigram should review metric stuff (not me).

PR #1471, opened 2020-04-30
SOLR-14014 PR Against Master

MarcusSorealheis:
# Description

I added a system property to allow users to disable the Solr Admin UI on start.

# Solution

I added this property, which accepts a boolean value, to SOLR_START_OPTS and START_OPTS.

# Tests

Not easily testable in the current project.

# Checklist

### DIFF
--- a/solr/bin/solr
+++ b/solr/bin/solr
@@ -2216,7 +2216,7 @@ function start_solr() {
     # users who don't care about useful error msgs can override in SOLR_OPTS with +OmitStackTraceInFastThrow
     "${SOLR_HOST_ARG[@]}" "-Duser.timezone=$SOLR_TIMEZONE" "-XX:-OmitStackTraceInFastThrow" \
     "-Djetty.home=$SOLR_SERVER_DIR" "-Dsolr.solr.home=$SOLR_HOME" "-Dsolr.data.home=$SOLR_DATA_HOME" "-Dsolr.install.dir=$SOLR_TIP" \
-    "-Dsolr.default.confdir=$DEFAULT_CONFDIR" "${LOG4J_CONFIG[@]}" "${SOLR_OPTS[@]}" "${SECURITY_MANAGER_OPTS[@]}")
+    "-Dsolr.default.confdir=$DEFAULT_CONFDIR" "${LOG4J_CONFIG[@]}" "${SOLR_OPTS[@]}" "${SECURITY_MANAGER_OPTS[@]}" "${SOLR_ADMIN_UI}")
 
   if [ "$SOLR_MODE" == "solrcloud" ]; then
     IN_CLOUD_MODE=" in SolrCloud mode"
--- a/solr/bin/solr
+++ b/solr/bin/solr
@@ -2097,6 +2097,14 @@ else
   SECURITY_MANAGER_OPTS=()
 fi
 
+# Enable ADMIN UI by default, and give the option for users to disable it
+if [ "$SOLR_ADMIN_UI_DISABLED" == "true" ]; then
+  SOLR_ADMIN_UI="-DdisableAdminUI=true"
+  echo -e "ADMIN UI Disabled"
+else
+  SOLR_ADMIN_UI="-DdisableAdminUI=false"
+fi
+
 JAVA_MEM_OPTS=()
 if [ -z "$SOLR_HEAP" ] && [ -n "$SOLR_JAVA_MEM" ]; then
   JAVA_MEM_OPTS=($SOLR_JAVA_MEM)
--- a/solr/bin/solr.cmd
+++ b/solr/bin/solr.cmd
@@ -1199,6 +1199,13 @@ IF "%SOLR_SECURITY_MANAGER_ENABLED%"=="true" (
 -Dsolr.internal.network.permission=*
 )
 
+REM Enable ADMIN UI by default, and give the option for users to disable it
+IF "%SOLR_ADMIN_UI_DISABLED%"=="true" (
+  set DISABLE_ADMIN_UI="true"
+) else (
+  set DISABLE_ADMIN_UI="false"
+)
+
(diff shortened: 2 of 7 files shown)


MarcusSorealheis:
@madrob and @gerlowskija I made a new PR and merged master, but it looks like another typo fix on my local made it in there, which is fine because it should be there. It is not totally relevant so it is not a pure PR, but it is one character added in docs so should be fine.

MarcusSorealheis:
Here, I take UI instance that was previously started when ADMIN UI was enabled, disabled the admin ui via system property, and then show the result. THe first change is because it's a single page app and I didn't reload the page: 

disabled-gif

solr/bin/solr.cmd
```diff
@@ -1288,6 +1295,7 @@ REM '-OmitStackTraceInFastThrow' ensures stack traces in errors,
 REM users who don't care about useful error msgs can override in SOLR_OPTS with +OmitStackTraceInFastThrow
 set "START_OPTS=%START_OPTS% -XX:-OmitStackTraceInFastThrow"
 set START_OPTS=%START_OPTS% !GC_TUNE! %GC_LOG_OPTS%
+set START_OPTS=-DdisableAdminUI=%DISABLE_ADMIN_UI%
```
madrob, line 1298: This clobbers the previous START_OPTS

solr/bin/solr
```diff
@@ -2097,6 +2097,14 @@ else
   SECURITY_MANAGER_OPTS=()
 fi
 
+# Enable ADMIN UI by default, and give the option for users to disable it
+if [ "$SOLR_ADMIN_UI_DISABLED" == "true" ]; then
+  SOLR_ADMIN_UI="-DdisableAdminUI=true"
+  echo -e "ADMIN UI Disabled"
+else
+  SOLR_ADMIN_UI="-DdisableAdminUI=false"
```
madrob, line 2105: It doesn't look like this property is ever passed to the run command?

solr/core/src/java/org/apache/solr/servlet/LoadAdminUiServlet.java
```diff
@@ -15,6 +15,13 @@
  * limitations under the License.
  */
 package org.apache.solr.servlet;
+import javax.servlet.http.HttpServletRequest;
```
madrob, line 18: nit: did you mean to move these imports around?

solr/core/src/java/org/apache/solr/servlet/LoadAdminUiServlet.java
```diff
@@ -24,29 +31,26 @@
...
   @Override
-  public void doGet(HttpServletRequest _request,
-                    HttpServletResponse _response)
-      throws IOException {
+  public void doGet(HttpServletRequest _request, HttpServletResponse _response) throws IOException {
     HttpServletRequest request = SolrDispatchFilter.closeShield(_request, false);
     HttpServletResponse response = SolrDispatchFilter.closeShield(_response, false);
-    
+
+    if(disabled){
```
madrob, line 49: Would this make sense to do before the close shield methods?

solr/core/src/java/org/apache/solr/servlet/LoadAdminUiServlet.java
MarcusSorealheis: Yes. I did that because the precommit failed before locally.

solr/bin/solr
MarcusSorealheis: You have to explicitly pass it, or you can pass the system property

madrob: So if I set `SOLR_ADMIN_UI_DISABLED=true` in my `solr.in.sh`, then it sets `SOLR_ADMIN_UI="-DdisableAdminUI=true"` here, and then... does nothing? Because the `SOLR_ADMIN_UI` variable is never read. What am I missing?

MarcusSorealheis: Yes, that was an error on my part. I forgot to carry that over from the previous PR. I've added it. See line 2219

MarcusSorealheis: https://github.com/apache/lucene-solr/pull/1471/commits/dedcdfe8b9af18d1a4662937be96b4d710963826 @madrob

solr/bin/solr.cmd
MarcusSorealheis: I believe this one is fixed as well @madrob

solr/core/src/java/org/apache/solr/servlet/LoadAdminUiServlet.java
MarcusSorealheis: Not in its current state because response object won't exist, but I could refactor if you have a good reason for that. Or, if I think of one. Gonna peep those methods again now.

MarcusSorealheis: So, I got pulled away by some work. I'll update. I think it makes sense to drop the response before the close shield if possible.

MarcusSorealheis: I'm testing it now.

MarcusSorealheis:
> It would be nice to make the Admin UI Disabled page a little bit prettier if possible, something more akin to the login page, but that might be out of scope of this PR. Feel free to defer that to a follow on if you think it's better to handle separately.

Subsequent PR please

solr/core/src/java/org/apache/solr/servlet/LoadAdminUiServlet.java
MarcusSorealheis: Handling the raw response works better. Looking at the snippet on this page originally I didn't have the full context. Good suggestion. Thanks 👍

MarcusSorealheis:
> Here, I take UI instance that was previously started when ADMIN UI was enabled, disabled the admin ui via system property, and then show the result. The first page change possible is because it's a single page app and I didn't reload the page (but you can see system property):
> 
> disabled-gif

This image doesn't apply anymore as you need to set the environment variable at startup and not use the Java Prop directly.

PR #1475, opened 2020-05-01
LUCENE-9148: Move the BKD index to its own file.

jpountz:
See https://issues.apache.org/jira/browse/LUCENE-9148.

This is only a work-in-progress for now, I'll finish when #1440 is merged to not duplicate work and introduce conflicts.


iverase:
Once we move a codec to backwards codecs, it can only be used for reading and still we are moving the `Lucene60PointsWriter`  to that package. Is that necessary?

uschindler:
I think the writer should only be in the test sources like the RWCodec

iverase:
I missed that, it is on the test sources. Thanks for the clarification!

lucene/core/src/java/org/apache/lucene/codecs/lucene86/Lucene86PointsWriter.java
```diff
@@ -0,0 +1,262 @@
...
+          }
+
+          @Override
+          public Relation compare(byte[] minPackedValue, byte[] maxPackedValue) {
+            return Relation.CELL_CROSSES_QUERY;
+          }
+        });
+
+      // We could have 0 points on merge since all docs with dimensional fields may be deleted:
+      if (writer.getPointCount() > 0) {
```
iverase, line 145: I wonder if we should use the same structure as in the other places by returning here as well a runnable for consistency?

lucene/core/src/java/org/apache/lucene/util/bkd/BKDWriter.java
```diff
@@ -627,20 +638,20 @@ assert valueInOrder(valueCount + leafCount,
...
 
       if (valueCount == 0) {
-        return -1;
+        return null;
       }
 
       pointCount = valueCount;
 
-      long indexFP = out.getFilePointer();
+      long indexFP = indexOut.getFilePointer();
```
iverase, line 654: Not used any more?

PR #1477, opened 2020-05-01
LUCENE-9321: Port markdown task to Gradle

uschindler:
This PR handles the markdown specific parts of the Lucene/Solr documentation:

- copy static assets (unfortunetely those are not yet unified between borth projects)
- convert .md files to HTML using flexmark
- create the index files by first iterating through the projects and collecting all links, creating a markdown representation of all links locally and then write as index.html to docus folder


uschindler:
For now I have no idea how to handle the last task, so I left it as nocommit. @mocobeta: any ideas how to get the relative path and the title of all projects below lucene/solr? I am also not sure if the links hardcoded as absolute links in `render-javadoc.gradle` are consistent with the intended folder structure below `build/documentation`.

These are all open questions and before a decission how to build the f*cking javadocs wthout copying all stuff to another folder (this costs imense amounts of disk space) I refuse to work on that issue/PR - sorry! @dweiss?

gradle/documentation/documentation.gradle
```diff
@@ -34,4 +36,11 @@ configure(subprojects.findAll { it.path == ':lucene' || it.path == ':solr' }) {
   ext {
     docroot = "${project.buildDir}/documentation"
   }
+  
+  task copyDocumentationAssets(type: Copy) {
+    includeEmptyDirs = false
+    from('site/html')      // lucene
```
dweiss, line 42: I'd rather have it as separate configure(":lucene"), configure(":solr") blocks... would be easier to separate the logic between the two?

gradle/documentation/markdown.gradle
```diff
@@ -0,0 +1,95 @@
...
+    classpath 'com.vladsch.flexmark:flexmark-ext-autolink:' + flexmarkVersion
+    classpath 'com.vladsch.flexmark:flexmark-ext-abbreviation:' + flexmarkVersion
+  }
+}
+
+configure(subprojects.findAll { it.path == ':lucene' || it.path == ':solr' }) {
+  task markdownToHtml(type: Copy) {
+    filteringCharset = 'UTF-8'
+    includeEmptyDirs = false
+    from('.') {            // lucene
```
dweiss, line 46: Right... same here.

gradle/documentation/markdown.gradle
```diff
@@ -0,0 +1,95 @@
...
+  }
+  
+  task createDocumentationIndex {
+    // nocommit: this needs to be implemented next
+  }
+}
+
+class MarkdownFilter extends FilterReader {
+  public MarkdownFilter(Reader reader) throws IOException {
+    // this is a hack: it reads whole file, converts it and provides result as a StringReader
```
dweiss, line 66: Looks legit to me, not a hack? :)

gradle/documentation/documentation.gradle
uschindler: I was thinking about that too, but this was easier to begin with. I did not want to clone the whole "Copy" task several times. If you tell me how to do this easier, give me a hint!

Should I remove the `from` here and then add 2 separate configures for lucene and solr just adding the `from`?

gradle/documentation/markdown.gradle
uschindler: It looks like misuse of a FilterReader, because you do everything in the constructor.

gradle/documentation/markdown.gradle
uschindler: I was thinking about that too, but this was easier to begin with. I did not want to clone the whole "Copy" task several times. If you tell me how to do this easier, give me a hint!

Should I remove the `from` here and then add 2 separate configures for lucene and solr just adding the `from`?

mocobeta:
I was not aware of that this task depends on projects' relative paths. To me, before proceeding it we need to manage to reach a consensus about the destination (output) directory for "renderJavadoc" anyway...?

mocobeta:
I am not sure if it's worth here, "render-javadoc.gradle" has a utility method to convert project path into output doc folder path: https://github.com/apache/lucene-solr/blob/master/gradle/render-javadoc.gradle#L21

mocobeta:
Just for clarification... Eventually, the `build/documentation` folder should look like this regardless of which way we choose - keep per-project javadoc outputs, or outputs docs into top-level `documentation` folder from the beginning:

**Lucene**
```
lucene/build/documentation
├── JRE_VERSION_MIGRATION.html
├── MIGRATE.html
├── SYSTEM_REQUIREMENTS.html
├── index.html
├── lucene_green_300.gif
├── analysis
│   ├── common
│   ├── icu
│   ├── kuromoji
│   ├── morfologik
│   ├── nori
│   ├── opennlp
│   ├── phonetic
│   ├── smartcn
│   └── stempel
├── backward-codecs
├── benchmark
├── classification
├── codecs
├── core
├── demo
├── expressions
├── facet
├── grouping
├── highlighter
├── join
├── luke
├── memory
├── misc
├── monitor
├── queries
├── queryparser
├── replicator
├── sandbox
├── spatial-extras
├── spatial3d
├── suggest
└── test-framework
```

**Solr**
```
solr/build/documentation
├── SYSTEM_REQUIREMENTS.html
├── index.html
├── contrib
│   ├── analysis-extras
│   ├── analytics
│   ├── clustering
│   ├── dataimporthandler
│   ├── dataimporthandler-extras
│   ├── extraction
│   ├── jaegertracer-configurator
│   ├── langid
│   ├── ltr
│   ├── prometheus-exporter
│   └── velocity
├── core
├── solrj
└── test-framework
```

The each subproject's javadoc folder structure is consistent with Gradle project path (as I emphasized on LUCENE-9278). Both `build/documentation` folder should be uploaded to lucene.apache.org website on an as-is basis (it's the final purpose of the `documentation` task).

dweiss:
Thanks for clarifying, Tomoko. I'm much in favor of keeping the javadocs in per-project build folders but if Uwe insists this is a problem then would it be a large patch to build those docs under target documentation location?

mocobeta:
> if Uwe insists this is a problem then would it be a large patch to build those docs under target documentation location?

I think all we need is replacing `project.javadoc.destinationDir` variables in this file with the target documentation folder, `_docroot_/${pathToDocdir(project.path)}`. (For now the renderJavadoc task does not care about where is the final destination, so we have to teach it about the value of _docroot_ to the task in some way.)
https://github.com/apache/lucene-solr/blob/master/gradle/render-javadoc.gradle

uschindler:
Hi, thanks for the discussion. @mocobeta explains it correct: You just have to change one thing to change the destination dir in the render-javadoc.

> I was not aware of that this task depends on projects' relative paths. To me, before proceeding it we need to manage to reach a consensus about the destination (output) directory for "renderJavadoc" anyway...?

Actually this is my main concern. I will comment on this on the issue. I have an idea for that. The reason for the issue is because there is conflicting interests: javadocs-jar on Maven central vs. documentation folder on web site and inside tar.gz of whole Lucene bundle.

uschindler:
See my langthly comment here: https://issues.apache.org/jira/browse/LUCENE-9321?focusedCommentId=17097899&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17097899

Please read it fully concentrated!

uschindler:
I can now proceed with this one as #1488 is done.

uschindler:
Hi, I added a first implemntation of `index.html`: 8dca5bb (Lucene only)

It works quite well:
- I added a method that collects all sub-projects with a specific path prefix (`:lucene:`) that have javadocs
- The list is then formatted as markdown (identical to the XSL). The only missing thing is the project description (we have it in Ant, but it's misisng in Gradle)
- The top part of the documentation was also converted to Markdown and is using project variabales with version numbers and the default codec
- default codec is extarcted like with ant, just simpler
- inputs/output of task are trivial

Open questions:
- Should I simple add a `project.description` to all `.gradle` files, so it's consistent with Ant?
- Where to place the 'templated' markdown source code. The XSL is currently a separate file. But as markdown contains Groovy `${...}` I would like to leave it in the gradle file.

Here is how it looks:
image

gradle/documentation/documentation.gradle
uschindler: I did this for the main task. At this place we can also do it, but that's less to maintain.

gradle/documentation/markdown.gradle
```diff
@@ -0,0 +1,190 @@
...
+    
+    doLast {
+      // static Codec defaultCodec   =   LOADER    .   lookup    (   "LuceneXXX"  )   ;
+      def regex = ~/\bdefaultCodec\s*=\s*LOADER\s*\.\s*lookup\s*\(\s*"([^"]+)"\s*\)\s*;/
+      def matcher = regex.matcher(defaultCodecFile.getText('UTF-8'))
+      if (!matcher.find()) {
+        throw GradleException("Cannot determine default codec from file ${defaultCodecFile}")
+      }
+      def defaultCodecPackage = matcher.group(1).toLowerCase(Locale.ROOT)
+      def markdown = """
```
dweiss, line 90: Should we move it to an external file? Then we can have a proper suffix and editor support.

dweiss:
bq. Should I simple add a project.description to all .gradle files, so it's consistent with Ant?

Hmm... What do you mean by adding project.description? We don't need to have per-project build.gradle file (for example if there are no dependencies). If you're thinking of a project property to hold the description then this could be configured from a single file too -- whatever is more convenient, really. I guess this description will come in handy for maven artifact (pom) as well.

uschindler:
The title and metadata of a project should IMHO be part of project's gradle file. You just add a `description = "..."` next to the dependencies, an done. Is this OK to you, or do you want them central?

In Ant this information is part of the root element of the `build.xml`.

uschindler:
> I guess this description will come in handy for maven artifact (pom) as well.

Yes, there it's public and searchable in e.g., search.maven.org

dweiss:
I don't mind either way. Can be per-project gradle file. Seems to fit there nicely.

gradle/documentation/markdown.gradle
uschindler: That was my question. Can we include it in a way so the groovy variables are parsed?

dweiss: https://docs.groovy-lang.org/latest/html/api/groovy/text/SimpleTemplateEngine.html
Should work?

uschindler: Cool. So I will place the file next to the legacy XSL file used by Ant. Once Ant is retired, the latter can go away.

uschindler:
I pushed a first mockup using Groovy's `SimpleTemplateEngine`.

dweiss:
It's very clean and nice with external file. I love it.

uschindler:
I now added a task, so it can be reused in Solr. In solr we have 2 variants: One HTML files that's shipped in ZIP/TGZ and the full blown one on the web. 

I also added more variables to the markdown: I added a binding for `majorVersion`. This one is used to format the line behind the Migration Guide: "What changed in Lucene 9, how to update 8.x code".

gradle/documentation/markdown.gradle
```diff
@@ -0,0 +1,172 @@
...
+  
+    // convert the markdown
+    MutableDataSet options = new MutableDataSet();
+    options.setFrom(ParserEmulationProfile.MARKDOWN);
+    options.set(Parser.EXTENSIONS, [ AbbreviationExtension.create(), AutolinkExtension.create(), AttributesExtension.create() ]);
+    options.set(HtmlRenderer.RENDER_HEADER_ID, true);
+    options.set(HtmlRenderer.MAX_TRAILING_BLANK_LINES, 0);
+    Document parsed = Parser.builder(options).build().parse(markdownSource);
+
+    StringBuilder html = new StringBuilder('<html>\n<head>\n');
```
mocobeta, line 136: Would it be better inserting `<!DOCTYPE html>` at the beginning (as a valid html5)?

uschindler: Yes, if we add CSS that's a good idea. With the current unstyled stuff this was the simplest we can do.

If we change that here, we should also do this in tools/ where Ant's script is located.

mocobeta: Applying CSS was in my mind. I will add the tag later when I add the style file.

uschindler: I would add CSS in a separate issue. I just want to get everything running first. What do you think about current state?

uschindler:
I added `description` metadata for root, core, analysis projects. Looks like it works as expected. I will proceed when coming back (out of office now):
- open build.xml and build.gradle in parallel
- add `description = ''` line
- copypaste stuff from build.xml's `<description/>`

gradle/documentation/markdown.gradle
mocobeta: > I would add CSS in a separate issue. 

Yes I think so too. I will open separate issue for that.

uschindler:
Hi,
I think all is done now. The documentation is generated for both Solr and Lucene.
For Solr this also generates the "minimal" documentation folder as zipped in the release distribution, which just contains a link to the online docs (`solr/build/docs-online`). When zipping we have to make sure to package the right stuff (see ant task).

I think this is ready for merge. I am running the whole stuff one more time with empty build dir.

uschindler:
I also figured out that the assets folder in Solr's site contains some files from the getting started guide which was deleted a while ago. I think we should delete all that, except logo and maybe css.

mocobeta:
I just run `gradlew documentation` on linux PC and manually checked the two index.html files and all hyperlinks; works fine for me.

uschindler:
Hi,
I split the documentation tasks of Lucene and Solr, so you can also just build lucene or just solr documentation (still building all dependencies in a correct way).

This helps with the documentation linter (see #1522), as it can depend on solr/lucene specific documentation.

On top-level we still have the documentation, task too

PR #1480, opened 2020-05-03
SOLR-14456: Fix Content-Type header forwarding on compressed requests

samuelgmartinez:
# Description

`HttpSolrCall` and `HttpSolrClient` were using `Content-Encoding` header incorrectly, as they were setting the charset as the encoding for the request. This bug surfaces only when a Solr node is forwarding a request to another node and tries to forward the original headers.

# Solution

Adjusted the behaviour to set the charset encoding as part of the `Content-Type` (as is specified by the HTTP standard) and `Content-Encoding` is now properly forwarded when available.

# Tests

Run the Solr test suites successfully but only from IntelliJ as from the terminal was failing everywhere (from lucene tests to some other random failures, even with badapples set to false).

Regardin new/modified unit tests, none. Since the `HttpSolrCall` has no tests and this takes a really complicated setup to test it, I just wanted to verify that the introduced changes won't break existing suites. Of course I'm open to add them but I'd appreciate some direction on how to build such a test.

# Checklist


madrob:
Sorry, accidentally closed the PR because I misunderstood what a button in the IntelliJ plugin was doing (I thought it was to close a popup, not close the PR).

HoustonPutman:
I think this issue should exist in `Http2SolrClient`, I see similar logic there.

samuelgmartinez:
> I think this issue should exist in `Http2SolrClient`, I see similar logic there. That was just a cursory look though.

I reviewed it and it's now properly implemented. It relies on two different methods for getting the encoding: one is Jetty's `Response.CompleteListener#getEncoding` (which is the charset not the content-encoding) and then the `Http2SolrClient#getEncoding` (which relies on a "manual" parsing of the charset attribute from the content-type header).

I think I should refactor `Http2SolrClient#getEncoding` to just get the encoding using the HttpClient's `ContentType` classes. I don't think it's perfect as the http2 Jetty based implementation would depend on a HttpClient class, but the dependency [is there already](https://github.com/apache/lucene-solr/blob/13f19f65559290a860df84fa1b5ac2db903b27ec/solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java#L686). Any opinion on the refactor?

HoustonPutman:
Yeah, I think it would be worthwhile for the two classes to act similarly in this regard. I say go ahead and refactor.

samuelgmartinez:
I just pushed the refactor. precommit is failing for an unused import I need to clean up. 

Also, fixed related tests because it no longer throws ParseException and now it always will throw RemoteSolrException when Content-Type is empty.

Tomorrow I’ll build the servers to do a manual round of testing with the cluster setup explained in the issue.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrClient.java
```diff
@@ -568,12 +568,18 @@ private HttpEntityEnclosingRequestBase fillContentStream(SolrRequest request, Co
...
+
+      ContentType contentType = ContentType.get(entity);
+      if (contentType != null) {
+        mimeType = contentType.getMimeType().trim().toLowerCase(Locale.ROOT);
+        charset = contentType.getCharset();
+
+        if (charset != null) {
+          charsetName = charset.name();
+        }
       }
```
HoustonPutman, line 583: I would recommend setting charset to FALLBACK_CHARSET if it's null here, and setting the charsetName afterwards.  You have potential for some nullPointerExceptions below. such as on lines 626 and 634

samuelgmartinez: I made the change already and was abbutto push it when I just wondered if it's the right decision. Here is my thinking: 
* The 626 log line wouldn't cause an NPE but just concat `null`. I agree that it should not be like that and should log a proper value.
* I'm afraid of taking the decision of the default encoding in this class as 'null' encoding for most operations would mean: "either default on Charset.defaultCharset() or take the decision yourself based on whatever you know you are parsing", specially regarding the `ResponseParser`s.

For example, the Json parser is defaulting to UTF-8 when the encoding is passed as null, but it could be that any other parser defaults to JVM `file.encoding` or an XML parser defaults to `ISO-8859-1`. In the original PR I just replaced the hardcoded references to `UTF-8` to the properly referenced `FALLBACK_CHARSET`, while I kept the response's original encoding (whether it's null or not) usages; like the one in line 634, `processor.processResponse` is passing `null` in the original code also.

Do you think that we can safely take the decision at this component layer (and hardcoded it to UTF-8) instead of delegating it to the `ResponseParser`s?

HoustonPutman: That's completely fair. We don't want to dictate the default for each response parser. If you fixed that logging line, I think that's enough. It'd be nice to have UTF8 instead of `null`

samuelgmartinez:
I pushed the changes addressing the exception log lines. 

When you have the opportunity, please review it and give me more feedback 🙂

HoustonPutman:
This looks good to me. I'll add a changes.txt entry and merge later today.

samuelgmartinez:
I’m interested on backporting this to the 7.7.x line. Regarding this I have a couple of questions: 
- Is that even possible? 
- If it is, should I open a PR for all 8.x branches also? 

I’m open to handle the PRs for every branch.

HoustonPutman:
So 7.7.3 was just released, and will likely be the last release of 7.x for a while. But there's no harm backporting it incase of another release. I can do that if you also create a PR against 7x

For 8 I will backport it to `8x` and it will be released in 8.6 (and 9, when that eventually gets cut)

HoustonPutman:
Ok, ran the tests and they pass. Also created a sample collection locally and curl'ed it using `--compressed` and it worked. Anything else I need to do in order to test this out @samuelgmartinez ?

samuelgmartinez:
The verification I ran was something like the one specified in the JIRA issue: 
- create a 2 solr node cluster: solr1(`http://localhost:8983`)  and solr2 (`http://localhost:8984`)
- create a collection "sampleCollection" with one replica and one shard, placed in solr1
- setup an HttpSolrClient with `http://localhost:8984/solr` as baseUrl to force all the requests to be rerouted by solr2 behind the scenes.
- Run a query with the SolrClient and verify it works.

`curl` should also would do the job just fine also (to verify `Content-Encoding` is not broken) as long as the cluster setup is the one I mentioned above and the request is always run to the node that doesn't host the collection.

HoustonPutman:
@samuelgmartinez I backported this to 8x. If you want to include this in 7_7 (although another release might not ever happen), please create another PR against that branch and tag me.

PR #1482, opened 2020-05-04
LUCENE-7822: CodecUtil#checkFooter should throw a CorruptIndexException as the main exception.

jpountz:
See https://issues.apache.org/jira/browse/LUCENE-7822.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/codecs/CodecUtil.java
+++ b/lucene/core/src/java/org/apache/lucene/codecs/CodecUtil.java
@@ -448,24 +448,27 @@ public final class CodecUtil {
       checkFooter(in);
     } else {
       try {
+        // If we have evidence of corruption then we return the corruption as the
+        // main exception and the prior exception gets suppressed. Otherwise we
+        // return the prior exception with a suppressed exception that notifies
+        // the user that checksums matched.
         long remaining = in.length() - in.getFilePointer();
         if (remaining < footerLength()) {
           // corruption caused us to read into the checksum footer already: we can't proceed
-          priorException.addSuppressed(new CorruptIndexException("checksum status indeterminate: remaining=" + remaining +
-                                                                 ", please run checkindex for more details", in));
+          throw new CorruptIndexException("checksum status indeterminate: remaining=" + remaining +
+                                          ", please run checkindex for more details", in);
         } else {
           // otherwise, skip any unread bytes.
           in.skipBytes(remaining - footerLength());
           
           // now check the footer
-          try {
-            long checksum = checkFooter(in);
-            priorException.addSuppressed(new CorruptIndexException("checksum passed (" + Long.toHexString(checksum) + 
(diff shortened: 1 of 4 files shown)


lucene/core/src/java/org/apache/lucene/codecs/CodecUtil.java
```diff
@@ -448,24 +448,27 @@ public static void checkFooter(ChecksumIndexInput in, Throwable priorException)
...
+        // main exception and the prior exception gets suppressed. Otherwise we
+        // return the prior exception with a suppressed exception that notifies
+        // the user that checksums matched.
         long remaining = in.length() - in.getFilePointer();
         if (remaining < footerLength()) {
           // corruption caused us to read into the checksum footer already: we can't proceed
-          priorException.addSuppressed(new CorruptIndexException("checksum status indeterminate: remaining=" + remaining +
-                                                                 ", please run checkindex for more details", in));
+          throw new CorruptIndexException("checksum status indeterminate: remaining=" + remaining +
+                                          ", please run checkindex for more details", in);
```
mikemccand, line 459: Nitpick: `;` instead of `,` since these are really two separate sentences?

lucene/core/src/java/org/apache/lucene/codecs/CodecUtil.java
```diff
@@ -448,24 +448,27 @@ public static void checkFooter(ChecksumIndexInput in, Throwable priorException)
...
           // now check the footer
-          try {
-            long checksum = checkFooter(in);
-            priorException.addSuppressed(new CorruptIndexException("checksum passed (" + Long.toHexString(checksum) + 
-                                                                   "). possibly transient resource issue, or a Lucene or JVM bug", in));
-          } catch (CorruptIndexException t) {
-            priorException.addSuppressed(t);
-          }
+          long checksum = checkFooter(in);
+          priorException.addSuppressed(new CorruptIndexException("checksum passed (" + Long.toHexString(checksum) +
```
mikemccand, line 466: Do we normally (in other places) also use `CorruptIndexException` to indicate a valid checksum?  I feel like we need a `NotCorruptIndexException` for this :)

lucene/core/src/test/org/apache/lucene/util/TestOfflineSorter.java
```diff
@@ -353,12 +352,10 @@ protected void corruptFile() throws IOException {
 
       // This corruption made OfflineSorter fail with its own exception, but we verify it also went and added (as suppressed) that the
```
madrob, line 353: no longer as suppressed

lucene/core/src/java/org/apache/lucene/codecs/CodecUtil.java
jpountz: This case is interesting indeed, it could either be a Lucene bug or a silent corruption.

PR #1486, opened 2020-05-05
SOLR-14423: Use SolrClientCache instance managed by CoreContainer

sigram:
See Jira for more details.


solr/core/src/java/org/apache/solr/handler/sql/CalciteSolrDriver.java
```diff
@@ -59,11 +81,346 @@ public Connection connect(String url, Properties info) throws SQLException {
...
     // Set the default schema
     calciteConnection.setSchema(schemaName);
 
-    return connection;
+    return new SolrCalciteConnectionWrapper(calciteConnection, solrSchema);
+  }
+
+  // the sole purpose of this class is to be able to invoke SolrSchema.close()
+  // when the connection is closed.
+  private static final class SolrCalciteConnectionWrapper implements CalciteConnection {
```
madrob, line 95: Can we have a new top level class that is a no-frills Wrapper, and then extend that here overriding just close? I think it is easier for people to see what is different.

solr/core/src/java/org/apache/solr/handler/GraphHandler.java
```diff
@@ -83,6 +84,7 @@
   private StreamFactory streamFactory = new DefaultStreamFactory();
   private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
   private String coreName;
+  private SolrClientCache solrClientCache;
```
cpoerschke, line 87: minor: here the ordering is coreName/solrClientCache but in the inform method the use order is solrClientCache/coreName. no particular order seems to be necessary technically speaking and so it might be helpful in terms of code readability to use the same ordering (one or the other) in both places?

solr/core/src/java/org/apache/solr/handler/sql/CalciteSolrDriver.java
cpoerschke: > ... a new top level class that is a no-frills Wrapper ...

You mean like a `FilterCalciteConnection` assuming `Filter...` is still the naming convention e.g. as in https://github.com/apache/lucene-solr/blob/releases/lucene-solr/8.5.1/lucene/core/src/java/org/apache/lucene/index/FilterCodecReader.java say?

I was going to suggest w.r.t. test coverage to ensure all that needs to be overridden has been overridden (both now and in future) -- similar to https://github.com/apache/lucene-solr/blob/releases/lucene-solr/8.5.1/lucene/core/src/test/org/apache/lucene/index/TestFilterCodecReader.java#L29 -- and yeah if the no-frills wrapper has the test coverage then it would be easier to see how `SolrCalciteConnectionWrapper extends FilterCalciteConnection` is only different w.r.t. the close() method.

solr/core/src/java/org/apache/solr/metrics/reporters/solr/SolrReporter.java
```diff
@@ -306,11 +312,11 @@ public String toString() {
   // We delegate to registries anyway, so having a dummy registry is harmless.
   private static final MetricRegistry dummyRegistry = new MetricRegistry();
 
-  public SolrReporter(HttpClient httpClient, Supplier<String> urlProvider, SolrMetricManager metricManager,
+  public SolrReporter(SolrClientCache solrClientCache, Supplier<String> urlProvider, SolrMetricManager metricManager,
                       List<Report> metrics, String handler,
                       String reporterId, TimeUnit rateUnit, TimeUnit durationUnit,
                       SolrParams params, boolean skipHistograms, boolean skipAggregateValues,
-                      boolean cloudClient, boolean compact) {
+                      boolean cloudClient, boolean compact, boolean closeClientCache) {
```
cpoerschke, line 319: minor/subjective: `SolrClientCache solrClientCache` and `boolean closeClientCache` arguments being adjacent might help with readability in the builder caller, though of course keeping all the `boolean` arguments together has advantages too

solr/core/src/java/org/apache/solr/handler/sql/CalciteSolrDriver.java
sigram: Sure, good idea.

solr/core/src/java/org/apache/solr/search/join/XCJFQuery.java
```diff
@@ -261,7 +261,7 @@ private TupleStream createSolrStream() {
     }
 
     private DocSet getDocSet() throws IOException {
-      SolrClientCache solrClientCache = new SolrClientCache();
+      SolrClientCache solrClientCache = searcher.getCore().getCoreContainer().getSolrClientCache();
```
cpoerschke, line 264: Need the `solrClientCache.close();` further down in the method be removed since a shared cache is now used?

sigram: Yes, good catch!

solr/core/src/java/org/apache/solr/metrics/reporters/solr/SolrReporter.java
sigram: I went a step back to keep the old constructor for back-compat - the class is public and its constructor is public too.

solr/core/src/java/org/apache/solr/handler/sql/CalciteSolrDriver.java
sigram: Fixed.

solr/core/src/java/org/apache/solr/core/CoreContainer.java
```diff
@@ -636,6 +651,8 @@ public void load() {
     updateShardHandler = new UpdateShardHandler(cfg.getUpdateShardHandlerConfig());
     updateShardHandler.initializeMetrics(solrMetricsContext, "updateShardHandler");
 
+    solrClientCache = new SolrClientCache(updateShardHandler.getDefaultHttpClient());
```
dsmiley, line 654: Maybe this is an aside by why is `UpdateShardHandler` called what it is?  Is it just for "updates"?  It has no javadocs.  It seems a little weird in this line to get an httpclient from an "update" thingy here -- seems arbitrary.  If UpdateShardHandler was perhaps named something more general and not specific to updates, I wouldn't be posing this question.

solr/core/src/java/org/apache/solr/handler/StreamHandler.java
```diff
@@ -88,27 +87,28 @@
  */
 public class StreamHandler extends RequestHandlerBase implements SolrCoreAware, PermissionNameProvider {
 
-  static SolrClientCache clientCache = new SolrClientCache();
-  static ModelCache modelCache = null;
-  static ConcurrentMap objectCache = new ConcurrentHashMap();
+  private ModelCache modelCache = null;
+  private ConcurrentMap objectCache = new ConcurrentHashMap();
```
dsmiley, line 91: Glad to see the removal of nasty statically declared caches!

solr/core/src/test/org/apache/solr/handler/sql/TestFilterCalciteConnection.java
```diff
@@ -0,0 +1,93 @@
...
+import java.sql.DriverManager;
+import java.util.Properties;
+
+import org.apache.calcite.jdbc.CalciteConnection;
+import org.apache.calcite.schema.SchemaPlus;
+import org.apache.solr.cloud.SolrCloudTestCase;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+/**
```
dsmiley, line 32: Does your IDE have a template generating these?  I recommend disabling.

solr/core/src/java/org/apache/solr/core/CoreContainer.java
sigram: This is a good question but it's out of scope for this issue.

This usage comes straight from refactoring - currently whenever the code needs a new SolrClient it uses the HttpClient from UpdateShardHandler.

PR #1487, opened 2020-05-05
SOLR-14426 Move top level classes to nested classes

madrob:
https://issues.apache.org/jira/browse/SOLR-14426

This is as far as I could get before needing to give up. It was partly automated by IntelliJ but still painful manual in some respects. I think I got everything in the analytics module, and made a dent in core.

Probably want to view this comparison ignoring whitespace changes.


HoustonPutman:
This looks good to me. Kind of annoying that the gradle precommit plugin has this bug. I looked around in that repo to try and find something, but I couldn't find anything.

HoustonPutman:
@uschindler have you seen this bug with other applications that use that plugin?

PR #1488, opened 2020-05-06
LUCENE-9321: Refactor renderJavadoc to allow relative links with multiple Gradle tasks

uschindler:
This is a WIP task to convert the renderJavadoc task in the Gradle build to run 2 times:
- Once for the Maven artifacts with absolute links (standard "javadoc" gradle task replacement. Output is project local). This task will be used for precommit (limited checks, as cross-module links cannot be checked. This task is for everyday use
- another task that renders all javadocs to the global documentation folder (one for Lucene, one for Solr). All links inside will be relative.

This PR currently contains:
- Refactor task to own class, so we can create multiple named tasks
- Add task property to generate relative links (a new closure was added to do this: It produces a relative link from the Gradle project path of current project to the linked project


gradle/render-javadoc.gradle
```diff
@@ -430,3 +305,180 @@ configure(subprojects.findAll { it.path in [':solr:solr-ref-guide', ':solr:serve
...
+    // resolve links to JavaSE and JUnit API
+    opts << [ '-linkoffline', javaSEDocUrl, project.project(':lucene').file('tools/javadoc/java11/') ]
+    if (linkJUnit) {
+      opts << [ '-linkoffline', junitDocUrl, project.project(':lucene').file('tools/javadoc/junit/') ]
+    }
+    // resolve inter-project links
+    linkProjects.collect { path ->
+      def docdir = convertPath2Link(path)
+      logger.warn(docdir as String);
+      opts << [ '-linkoffline', docdir, project.project(path).javadoc.destinationDir ]
```
mocobeta: To generate relative links, have you considered to use `-link` (as the original ant task does so) option instead of `-linkoffline` ? The first version of "renderJavadoc" had also used `-link` option with relative file paths before we changed it to absolute urls. (I'm not sure if it makes sense here and have no strong feeling with that, but `-link` seems to be more safe since with the option javadoc tool checks if the files to be linked exist on the local machine.)

uschindler: Ah thanks. I will look into this. From what I have seen as output the links were generated correctly.

So I will use `-link` if the boolean flag enables relative links, otherwise `-linkoffline`. I will check the ANT task, too.

gradle/render-javadoc.gradle
```diff
@@ -57,20 +57,11 @@ configure(subprojects.findAll { it.path.startsWith(':lucene') && it.path != ':lu
...
-            encoding: "UTF-8",
-            file: "${project.javadoc.destinationDir}/element-list",
-            byline: "true",
-            match: "^(\${element-list-regex})\$",
-            replace: "")
+        Set luceneCorePackages = file("${project(':lucene:core').tasks[name].outputDir}/element-list").readLines('UTF-8').toSet();
+        File elementFile = file("${outputDir}/element-list");
+        List elements = elementFile.readLines('UTF-8');
+        elements.removeAll{ luceneCorePackages.contains(it) }
+        elementFile.write(elements.join('\n').concat('\n'), 'UTF-8');
```
dweiss, line 64: Nice!

uschindler:
Hi thanks Dawid for the changes!
I think the whole thing is almost operable:
- instead of root-project we should just detect lucene vs. solr and use the lucene vs solr root project's documentation folder (there is already a extension property available for that)
- for the relative links we have to check if it crosses project boundaries, that's also easy check. 

I think we are close. If I have some time later I will proceed.

uschindler:
P.S.: The new relative path code is fine - I know you hated mine, we just need the additional cross-root-level boundary.

uschindler:
P.P.S.: I have a super-super-wide 4K laptop monitor! 😎

gradle/render-javadoc.gradle
```diff
@@ -42,291 +41,255 @@ allprojects {
...
+  boolean linkJUnit = false
   
   @Input
-  boolean relativeProjectLinks = false;
+  boolean relativeProjectLinks = false
 
   @Input
-  def linkProjects = [];
-  
+  def linkedProjectNames = []
```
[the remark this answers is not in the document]
dweiss: No - linkProjects contains project references that cannot be used as input; the other array contains project paths. Perhaps there is a cleaner way to do this - I didn't investigate deeper.

dweiss:
bq. P.P.S.: I have a super-super-wide 4K laptop monitor! 😎

I realized this pretty soon after looking at your patch. :)

dweiss:
I can finish tomorrow, Uwe. I just had to take care of home businesses today and pushed what I had already.

gradle/render-javadoc.gradle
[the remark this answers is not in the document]
uschindler: That's what I figured out. The names, although unused are good task inputs.

I would just make the other property private.

gradle/render-javadoc.gradle
```diff
@@ -39,394 +35,398 @@ allprojects {
...
+        def base = otherProject.path.startsWith(":lucene") ? luceneDocUrl : solrDocUrl
+        allOfflineLinks.put("${base}/${otherProject.relativeDocPath}/".toString(), otherProject.tasks[name].outputDir)
+      }
+    }
+
+    // Add offline links.
+    // TODO: add some sanity check/ validation here to ensure folder/package-list or folder/element-list is present?
+    allOfflineLinks.each { url, folder ->
+      opts << [ '-linkoffline', url, folder ]
+    }
```
mocobeta, line 367: Would it be better to use Unix terminology "dir" or "directory" for variable names ?

dweiss: Sure, I don't mind. :) Sorry for lack of consistency.

uschindler:
Looks fine now: We must only strip the first "lucene/" or "solr/" from the site's doc path.
I added outputDir.mkdirs(), so the relative links are more predicatable.

uschindler:
OK, I fixed the relative doc path.

I was talking with @rmuir yesterday: He said: At ANT times he hardcoded the dependencies between Javadocs links, so he would propose to just use inter-project dependencies as recorded in Gradle also to add the "-link" relations.

I think this should be easy to do.

Otherwise I think this is almost ready. @dweiss do you have any other things planned? I would like to get thin in quite soon, so I can proceed with tha other issue who builds also the Markdown files and we would then have working documentation with "gradlew documentation".

mocobeta:
+1; `documentation` task works for me, thank you for refactoring it.

I found small differences between generated Solr javadocs (on snapshot build).

With "ant documentation", there's no link to lucene javadocs.
Screenshot from 2020-05-13 22-50-47

With "gradlew documentation", there are (still non-existent) links to lucene javadocs.
Screenshot from 2020-05-13 22-49-20

If it does not have any bad effects on CI job etc., I think it's ok.

uschindler:
Thanks @mocobeta 
I think we explicitly excluded inter-project links.

The problem is that the main task add :lucene:core to all subprojects:

```
configure(subprojects.findAll { it.path.startsWith(':lucene') && it.path != ':lucene:core' }) {
  project.tasks.withType(RenderJavadocTask) {
    addLinkedProject(':lucene:core')
  }
}
```

As I am currebtly looking into removing all dependency stuff, I will take care of that.

uschindler:
Hi,
I did a quick mockup to generate the links automatically:

```
    // Resolve inter-project links.
    // resolve dependencies for links
    def projectDependencies = project.configurations.implementation.allDependencies.withType(ProjectDependency).collect { dep ->
      def otherProject = dep.dependencyProject
      // For relative links we compute the actual relative link between projects.
      def crossLuceneSolr = (otherProject.docroot != project.docroot)
      if (relativeProjectLinks && !crossLuceneSolr) {
        def pathTo = otherProject.tasks[name].outputDir.toPath().toAbsolutePath()
        def pathFrom = outputDir.toPath().toAbsolutePath()
        def relative = pathFrom.relativize(pathTo).toString().replace(File.separator, '/')
        opts << [ '-link', relative ]
      } else {
        // For absolute links, we determine the target URL by assembling the full URL.
        def base = otherProject.path.startsWith(":lucene") ? luceneDocUrl : solrDocUrl
        allOfflineLinks.put("${base}/${otherProject.relativeDocPath}/".toString(), otherProject.tasks[name].outputDir)
      }
    }
```

Very easy! Questions:
- Do we want solr -> lucene deps, or not? If not I would also rmeove the linkoffline above (the else statement can be removed)
- Is the configuration used the correct one? "implementation" vs "api"? I was first trying "api" (because we only look at the API), but these are not all links the current script produces.

Do the order matter, I think not, but maybe sort them alphabetically to have a non-random order. Split Packages are fixed anyways.

uschindler:
I have to add a small fix: I have to exclude projects that have no render task that is enabled. Otherwise solr fails because it depends on solr:server, which is disabled.

Otherwise my approach works. What do you think?

mocobeta:
+1 for the approach generating links from gradle dependency graph, it is smart and also robust. 
I think it'd be better we keep the absolute links from Solr to Lucene Javadocs (since they are legit, ordinary offline links to external artifacts anyway?)

dweiss:
bq. Is the configuration used the correct one? "implementation" vs "api"?

In theory it should be API only because the "API" dependencies are what is part of the public classes (for example arguments or return types on exported methods). Implementation types sholdn't be exposed as they shouldn't be part of the public API.

Of course, this is theory. In practice I don't think we have it cleaned up to this level so I'd keep the implementation. All it does is it links up projects - if they're not used in references nothing will happen.

uschindler:
OK, I pushed the new code for automaically building links from project dependencies. How it works:
- For each project having a RenderJavadocsTask it adds lazily evaluated "dependsOn" using a closure. The closure lists all implementation dependencies  and returns all tasks with same name. This effectively make the render task depend on all tasks with same name in its own dependencies
- During task execution, the task just looks for its own dependencies, filters again on the name and uses the task settings (outputDir and projects's doc-dir) to generate links

dweiss:
Looks excellent to me, Uwe. Also, thanks for eyeballing the outputs with your keen eye, Tomoko.

mocobeta:
Looks excellent to me too.

uschindler:
Hi, I did a small change:
The tasks returned by resolving dependencies are *unsorted*. In case of split packages, the results of javadoc execution are therefore not reproducible. I fixed this by applying a sort order:
- first all projects named "core"
- all other projects sorted lexicographical by (task) path

This makes the `-link` parameters (and also the absolute ones) predictable in order and also implements it in the same way like ANT did.

dweiss:
+1.

uschindler:
After comparing outputs I found out that the whole thing only worked with direct dependencies. Of course we rely only on render tasks of directly referenced projects. So `:solr:core` depends on lots of Lucene stuff and all this is rendered before. This setup works perfectly.
But when we add the `-link` the previous code only searched for renderJavadoc tasks that were direct dependenencies. So Solr contribs were not adding all of lucene, because most Solr contribs nly depend on Solr core. Of course the Javadocs were created before, but we missed them.
I added a new method to the task, that finds all previous render tasks (recursively). The code in the main logic is now also more readable (we just sort the retun value and add links).

Now everything works as expected, I will merge this PR soon. Thanks @dweiss and @mocobeta for help!

gradle/render-javadoc.gradle
```diff
@@ -151,6 +151,18 @@ class RenderJavadocTask extends DefaultTask {
 
   @Input
   def solrDocUrl = project.propertyOrDefault('solr.javadoc.url', "https://lucene.apache.org/solr/${project.baseVersion.replace(".", "_")}")
+  
+  /** Utility method to recursively collect all tasks with same name like this one that we depend on */
+  private Set findPreviousRenderTasks() {
```
dweiss, line 156: findRenderTasksInDependencies maybe?

uschindler: Yeah, much better! This name was puzzling me. I'll change!

uschindler:
I just noticed we have no CHANGES entry for Gradle at all. We should maybe add one for all.
So I will merge this without adding a new changes entry.

mocobeta:
> Also, thanks for eyeballing the outputs with your keen eye, Tomoko.

If I correctly understand the meaning of the phrase, my fellows occasionally say similar things to me (in Japanese). Thanks.  ~(=^‥^)/

PR #1489, opened 2020-05-06
RegEx querying - add support for Java’s predefined character classes like \d for digits

markharwood:
Jira Issue [9336](https://issues.apache.org/jira/browse/LUCENE-9336) proposes adding support for common regex character classes like `\w`.
This PR adds the code to RegExp.java and associated tests.

One other consideration is that the shorthand expressions list could perhaps be made configurable e.g. `\h` might be shorthand used to represent hashtags of the form `#\w*` if that was something users routinely searched for and wanted to add to the regex vocabulary.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/util/automaton/RegExp.java
+++ b/lucene/core/src/java/org/apache/lucene/util/automaton/RegExp.java
@@ -317,7 +317,8 @@ import java.util.Set;
 public class RegExp {
   
   enum Kind {
-    REGEXP_UNION, REGEXP_CONCATENATION, REGEXP_INTERSECTION, REGEXP_OPTIONAL, REGEXP_REPEAT, REGEXP_REPEAT_MIN, REGEXP_REPEAT_MINMAX, REGEXP_COMPLEMENT, REGEXP_CHAR, REGEXP_CHAR_RANGE, REGEXP_ANYCHAR, REGEXP_EMPTY, REGEXP_STRING, REGEXP_ANYSTRING, REGEXP_AUTOMATON, REGEXP_INTERVAL
+    REGEXP_UNION, REGEXP_CONCATENATION, REGEXP_INTERSECTION, REGEXP_OPTIONAL, REGEXP_REPEAT, REGEXP_REPEAT_MIN, REGEXP_REPEAT_MINMAX, REGEXP_COMPLEMENT, REGEXP_CHAR, REGEXP_CHAR_RANGE, REGEXP_ANYCHAR, REGEXP_EMPTY, REGEXP_STRING, REGEXP_ANYSTRING, REGEXP_AUTOMATON, REGEXP_INTERVAL,
+    REGEXP_SHORTHAND
   }
   
   /**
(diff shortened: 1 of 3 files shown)


lucene/core/src/test/org/apache/lucene/search/TestRegexpQuery.java
```diff
@@ -115,6 +145,128 @@ public Automaton getAutomaton(String name) {
     assertEquals(1, searcher.search(query, 5).totalHits.value);
   }
   
+  public void testCoreJavaParity() {
```
romseygeek, line 148: I think these tests should be in `TestRegExp` rather than in `TestRegexpQuery`?

markharwood:
Thanks for the review, @romseygeek 
I've shifted that test.

markharwood:
@jpountz / @jimczi do you want to comment on this?
As Frenchmen you might have opinions on `\w` and the assumption that words are only made of `[a-zA-Z0-9]` characters.
 ;-)

lucene/core/src/java/org/apache/lucene/util/automaton/RegExp.java
```diff
@@ -1101,10 +1116,51 @@ final RegExp parseCharClasses() throws IllegalArgumentException {
...
+      return predefinedExp;
+    }
+        
     int c = parseCharExp();
     if (match('-')) return makeCharRange(c, parseCharExp());
     else return makeChar(c);
   }
+
+  RegExp expandPredefined() {
+    //See https://docs.oracle.com/javase/tutorial/essential/regex/pre_char_classes.html
```
jimczi, line 1130: Can you add the javadocs for this in the class definition ?

lucene/core/src/java/org/apache/lucene/util/automaton/RegExp.java
```diff
@@ -317,7 +317,8 @@
 public class RegExp {
   
   enum Kind {
-    REGEXP_UNION, REGEXP_CONCATENATION, REGEXP_INTERSECTION, REGEXP_OPTIONAL, REGEXP_REPEAT, REGEXP_REPEAT_MIN, REGEXP_REPEAT_MINMAX, REGEXP_COMPLEMENT, REGEXP_CHAR, REGEXP_CHAR_RANGE, REGEXP_ANYCHAR, REGEXP_EMPTY, REGEXP_STRING, REGEXP_ANYSTRING, REGEXP_AUTOMATON, REGEXP_INTERVAL
+    REGEXP_UNION, REGEXP_CONCATENATION, REGEXP_INTERSECTION, REGEXP_OPTIONAL, REGEXP_REPEAT, REGEXP_REPEAT_MIN, REGEXP_REPEAT_MINMAX, REGEXP_COMPLEMENT, REGEXP_CHAR, REGEXP_CHAR_RANGE, REGEXP_ANYCHAR, REGEXP_EMPTY, REGEXP_STRING, REGEXP_ANYSTRING, REGEXP_AUTOMATON, REGEXP_INTERVAL,
+    REGEXP_SHORTHAND
```
jimczi, line 321: Maybe rename to `REGEXP_PRE_CLASS` ?

PR #1490, opened 2020-05-07
SOLR-14461: Replace commons-fileupload with Jetty

dsmiley:
https://issues.apache.org/jira/browse/SOLR-14461


solr/core/src/java/org/apache/solr/servlet/SolrRequestParsers.java
```diff
@@ -571,59 +552,99 @@ public SolrParams parseParamsAndFillStreams(
...
-        factory.setFileCleaningTracker(fileCleaningTracker);
-      }
-      // TODO - configure factory.setSizeThreshold(yourMaxMemorySize);
-      // TODO - configure factory.setRepository(yourTempDirectory);
+    public MultipartRequestParser(int uploadLimitKB) {
+      multipartConfigElement = new MultipartConfigElement(
+          null, // temp dir (null=default)
+          -1, // maxFileSize  (-1=none)
+          uploadLimitKB * 1024, // maxRequestSize
+          1 * 1024 * 1024 ); // fileSizeThreshold after which will go to disk
```
dsmiley, line 566:
```suggestion
          100 * 1024 ); // fileSizeThreshold after which will go to disk
```
Lowering to reduce DoS risk

ErickErickson:
Just wanted to pass on a thanks for taking it out of the gradle build too. That'll make things much easier to deal with when we eventually reconcile it all.

PR #1491, opened 2020-05-07
LUCENE-9366: Remove unused maxDoc parameter from DocValues.emptySortedNumeric()

romseygeek:


romseygeek:
> Is this change master-only or 8.x too?

Given that it's a trivial change to make and in a fairly expert API, I think this is OK to backport to 8.x

PR #1494, opened 2020-05-07
Backport of FuzzyQuery bugfix for issue 9365

markharwood:
Backport of 28e47549c8ba1a7c17ffe7d9e791e88983ef46c2
Fixes false negative in FuzzyQuery when prefix length == search string length.


markharwood:
Aborting - there was a bug with this fix in master. Using RegExpQuery as a fallback is not appropriate because content needs escaping to be legal syntax. Will rework

PR #1496, opened 2020-05-07
Bugfix for FuzzyQuery false negative when prefix length == search term length

markharwood:
Fix for Jira issue 9365 where search for `abc` doesn't match doc `abcd` if prefixlength = 3 and edit distance =1.
The fix is to rewrite the FuzzyQuery as a specialised form of Automaton query when prefix length == search string length.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/FuzzyQuery.java
+++ b/lucene/core/src/java/org/apache/lucene/search/FuzzyQuery.java
@@ -20,13 +20,17 @@ package org.apache.lucene.search;
 import java.io.IOException;
 import java.util.Objects;
 
+import org.apache.lucene.index.IndexReader;
 import org.apache.lucene.index.SingleTermsEnum;
 import org.apache.lucene.index.Term;
 import org.apache.lucene.index.Terms;
 import org.apache.lucene.index.TermsEnum;
 import org.apache.lucene.util.AttributeSource;
+import org.apache.lucene.util.automaton.Automata;
+import org.apache.lucene.util.automaton.Automaton;
 import org.apache.lucene.util.automaton.CompiledAutomaton;
 import org.apache.lucene.util.automaton.LevenshteinAutomata;
+import org.apache.lucene.util.automaton.Operations;
 
 /** Implements the fuzzy search query. The similarity measurement
  * is based on the Damerau-Levenshtein (optimal string alignment) algorithm,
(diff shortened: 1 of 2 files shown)


lucene/core/src/java/org/apache/lucene/search/FuzzyQuery.java
```diff
@@ -99,7 +103,24 @@ public FuzzyQuery(Term term, int maxEdits, int prefixLength, int maxExpansions,
...
     this.maxExpansions = maxExpansions;
-    setRewriteMethod(new MultiTermQuery.TopTermsBlendedFreqScoringRewrite(maxExpansions));
+    if (term.text().length() == prefixLength) {
+      setRewriteAsPrefixedQuery();
+    } else {
+      setRewriteMethod(new MultiTermQuery.TopTermsBlendedFreqScoringRewrite(maxExpansions));
+    }
+  }
+
+  private void setRewriteAsPrefixedQuery() {
```
madrob, line 113: Can we refactor this into a static inner class?

lucene/core/src/test/org/apache/lucene/search/TestFuzzyQuery.java
```diff
@@ -225,6 +225,39 @@ public void testFuzziness() throws Exception {
     directory.close();
   }
   
+  public void testPrefixLengthEqualStringLength() throws Exception {
+    Directory directory = newDirectory();
+    RandomIndexWriter writer = new RandomIndexWriter(random(), directory);
+    addDoc("b*a", writer);
```
madrob, line 231: Can you add a test case with multi-byte strings as well?

lucene/core/src/java/org/apache/lucene/search/FuzzyQuery.java
```diff
@@ -99,7 +103,24 @@ public FuzzyQuery(Term term, int maxEdits, int prefixLength, int maxExpansions,
     this.prefixLength = prefixLength;
     this.transpositions = transpositions;
     this.maxExpansions = maxExpansions;
-    setRewriteMethod(new MultiTermQuery.TopTermsBlendedFreqScoringRewrite(maxExpansions));
+    if (term.text().length() == prefixLength) {
```
madrob, line 106: I think the problem is that we're checking this in `getTermsEnum` and potentially returning a `SingleTermsEnum`. Checking it here and modifying rewriteMethod feels inconsistent.

madrob:
Thanks for catching this, this feels significant enough of a bug fix to benefit from a JIRA issue, could you file one at issues.apache.org/jira and link to this PR?

markharwood:
>Thanks for catching this, this feels significant enough of a bug fix to benefit from a JIRA issue, could you file one and link to this PR?

Thanks for reviewing. The Jira issue is [here](https://issues.apache.org/jira/browse/LUCENE-9365)

markharwood:
Closing - addressed instead by https://gitbox.apache.org/repos/asf?p=lucene-solr.git;h=45611d0

PR #1497, opened 2020-05-07
SOLR-8394: /admin/luke didn't computeindexHeapUsageBytes

dsmiley:
Needed to call FilterLeafReader.unwrap.

https://issues.apache.org/jira/browse/SOLR-8394

CC @sigram


solr/core/src/java/org/apache/solr/handler/admin/LukeRequestHandler.java
```diff
@@ -634,17 +635,19 @@ private static long getSegmentsFileLength(IndexCommit commit) {
...
+        .map(FilterLeafReader::unwrap)
+        .map(leafReader -> {
+          if (leafReader instanceof Accountable) {
+            return ((Accountable) leafReader).ramBytesUsed();
+          } else {
+            return -1L; // unsupported
+          }
+        })
+        .mapToLong(Long::longValue)
+        .reduce(0, (left, right) -> left == -1 || right == -1 ? -1 : left + right);
```
bruno-roustant, line 649: I like the first 4 lines (this is rare for me to see a clear stream). But the last 2 lines are less clear, it took me some time to understand why it's needed.

solr/core/src/java/org/apache/solr/handler/admin/LukeRequestHandler.java
```diff
@@ -634,17 +635,19 @@ private static long getSegmentsFileLength(IndexCommit commit) {
...
-        // Not supported for any reader that is not a SegmentReader
-        return -1;
-      }
-    }
-    return indexHeapRamBytesUsed;
+    return reader.leaves().stream()
+        .map(LeafReaderContext::reader)
+        .map(FilterLeafReader::unwrap)
+        .map(leafReader -> {
+          if (leafReader instanceof Accountable) {
```
juanka588, line 642: can we filter here instead?

juanka588: or map into accountable returning -1L

dsmiley: Feel free to suggest a code snippet; I'm not sure I get it.

One small improvement I see I could make was change the last "map" call to be a "mapToLong" instead, and then I can remove the last mapToLong.

PR #1498, opened 2020-05-08
SOLR-14351: commitScheduler was missing MDC logging

dsmiley:
Adding this to https://issues.apache.org/jira/browse/SOLR-14351 even though it's not a perfect fit.

As an aside, I've seen MDC missing logs from searcher warming QuerySenderListener.  There's another JIRA issue for that one indirectly fixed via stacking SolrRequestInfo.

### DIFF
--- a/solr/core/src/java/org/apache/solr/update/CommitTracker.java
+++ b/solr/core/src/java/org/apache/solr/update/CommitTracker.java
@@ -29,6 +29,7 @@ import java.util.concurrent.atomic.AtomicLong;
 import org.apache.solr.common.SolrException;
 import org.apache.solr.common.params.ModifiableSolrParams;
 import org.apache.solr.core.SolrCore;
+import org.apache.solr.logging.MDCLoggingContext;
 import org.apache.solr.request.LocalSolrQueryRequest;
 import org.apache.solr.request.SolrQueryRequest;
 import org.apache.solr.common.util.SolrNamedThreadFactory;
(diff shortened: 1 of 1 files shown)


solr/core/src/java/org/apache/solr/update/CommitTracker.java
```diff
@@ -57,8 +58,10 @@
   private int docsUpperBound;
   private long timeUpperBound;
   private long tLogFileSizeUpperBound;
-  
-  private final ScheduledExecutorService scheduler = 
+
+  // note: can't use ExecutorsUtil because it doesn't have a *scheduled* ExecutorService.
+  //  Not a big deal but it means we must take care of MDC logging here.
```
madrob, line 63: Is there an issue for improving ExecutorUtil for this?

dsmiley: I thought of that.  There isn't much need for a ScheduledExecutorService throughout Solr, and it'd probably add a bunch of code over in ExecutorsUtil to add one.  I looked at all that ExecutorsUtil does which is kind of a ton and thought all I really want here is a basic MDC to set the core for logging and that's it.  That said, if you think the usage here ought to be doing more than just this, then we should probably go add stuff to ExecutorsUtil for the Scheduled case.

madrob: I think it's fine.

PR #1500, opened 2020-05-08
Backport LUCENE-9350 to branch_8_5

madrob:
It may be worthwhile to review this by individual commits instead of all at once. I do not intend to squash these when committing.

I'm most concerned about whether the changes in commit 3363df7 are sufficient - they are enough for API and compilation, but it's not clear if I needed to update some implementations of QueryVisitor to tie all of this together.

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -7,7 +7,10 @@ http://s.apache.org/luceneversions
 
 Bug Fixes
 ---------------------
-(No changes)
+
+* LUCENE-9350: Partial reversion of LUCENE-9068; holding levenshtein automata on FuzzyQuery can end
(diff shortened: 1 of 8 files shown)


romseygeek:
I don't think we should backport the changes to QueryVisitor, we shouldn't be breaking APIs in a bugfix release even if it is experimental.  Can we change it to build the automaton directly in FuzzyQuery.visit()? It's not ideal, but it's the status quo ante

madrob:
> Can we change it to build the automaton directly in FuzzyQuery.visit()
Sure, I dropped the QueryVisitor changes and we do this now.

I think the changes to use a Supplier in FuzzyTermsEnum are still ok?

PR #1505, opened 2020-05-11
LUCENE-9328: Add PooledDocValuesReader

romseygeek:
This commit adds a PooledDocValuesReader that shares doc values
iterators among several consumers, and adds it to grouping second-pass
collectors.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/PooledDocValuesReader.java
+++ b/lucene/core/src/java/org/apache/lucene/index/PooledDocValuesReader.java
@@ -58,7 +58,7 @@ public class PooledDocValuesReader extends FilterLeafReader {
   private final LeafReaderContext pooledContext;
   private final Predicate<String> fieldFilter;
 
-  private int currentDoc;
+  private int currentDoc = -1;
 
   /**
    * Creates a new PooledDocValuesReader, wrapping an existing reader context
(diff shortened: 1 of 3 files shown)


lucene/core/src/java/org/apache/lucene/index/PooledDocValuesReader.java
```diff
@@ -0,0 +1,548 @@
...
+    public void advanceReal(int target) throws IOException {
+      if (target > realDoc) {
+        realDoc = in.advance(target);
+      }
+      positioned = realDoc == target;
+      if (positioned) {
+        int c = in.docValueCount();
+        values = ArrayUtil.grow(values, c);
+        for (int i = 0; i < c; i++) {
+          values[i] = in.nextValue();
```
mkhludnev, line 286: I don't understand why it's necessary to cache these per doc values. DocValues seems naturally sharable across `SortingGroupHeads` every one of them calls `advanceExact()` which is valid for the same doc that allow to get through the same values again and again.

lucene/core/src/java/org/apache/lucene/index/PooledDocValuesReader.java
```diff
@@ -0,0 +1,548 @@
...
+  }
+
+  /**
+   * Creates a new PooledDocValuesReader, wrapping an existing reader context and
+   * pooling docvalues from fields that match the fieldFilter predicate
+   */
+  public PooledDocValuesReader(LeafReaderContext in, Predicate<String> fieldFilter) {
+    super(in.reader());
+    this.fieldFilter = fieldFilter;
+    this.pooledContext = new LeafReaderContext(in.parent, this, in.ordInParent, in.docBaseInParent, in.ord, in.docBase);
```
mkhludnev, line 78: Aha... thanks for the trick.

PR #1508, opened 2020-05-11
SOLR-14472: AutoScale: "cores" also track lazy, transient

dsmiley:
https://issues.apache.org/jira/browse/SOLR-14472

### DIFF
--- a/solr/solrj/src/java/org/apache/solr/client/solrj/impl/SolrClientNodeStateProvider.java
+++ b/solr/solrj/src/java/org/apache/solr/client/solrj/impl/SolrClientNodeStateProvider.java
@@ -66,7 +66,7 @@ import static org.apache.solr.client.solrj.cloud.autoscaling.Variable.Type.TOTAL
 import static org.apache.solr.client.solrj.cloud.autoscaling.Variable.Type.WITH_COLLECTION;
 
 /**
- *
+ * The <em>real</em> {@link NodeStateProvider}, which communicates with Solr via SolrJ.
  */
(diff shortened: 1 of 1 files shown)


solr/solrj/src/java/org/apache/solr/client/solrj/impl/SolrClientNodeStateProvider.java
```diff
@@ -273,30 +273,31 @@ protected void getRemoteInfo(String solrNode, Set<String> requestedTags, SnitchC
...
         }
         if (requestedTags.contains(CORES)) {
+          NamedList<?> node = (NamedList<?>) metrics.get("solr.node");
           int count = 0;
-          Map cores = (Map) m.get("metrics");
-          for (Object o : cores.keySet()) {
-            if (o.toString().startsWith("solr.core.")) count++;
+          for (String leafCoreMetricName : new String[]{"lazy", "loaded", "unloaded"}) {
+            Number n = (Number) node.get("CONTAINER.cores." + leafCoreMetricName);
+            if (n != null) count += n.intValue();
```
megancarey, line 291: I don't know if this issue was isolated to our use case - but I know we sometimes saw negative values for some of the core counts. Do you want to check if (n != null && n > 0)?

dsmiley: I've thought of that issue.  I think the proper place to fix that is at the source of the metric, not here (consumer).

dsmiley:
Merged d3d0df9da15b9a60280775a0deb4f57e1bd3dea5

PR #1509, opened 2020-05-11
SOLR-10810: Examine precommit lint WARNINGs in non-test code

ErickErickson:
Here's what I've got for removing some of the warnings in solr/core, here for discussion.

There are a lot of changes from static class declarations to nested declarations, and a few wildcard generics that snuck in.

Please discuss freely.


solr/core/src/java/org/apache/solr/handler/AnalysisRequestHandlerBase.java
```diff
@@ -82,7 +82,7 @@ public void handleRequestBody(SolrQueryRequest req, SolrQueryResponse rsp) throw
    *
    * @throws Exception When analysis fails.
    */
-  protected abstract NamedList doAnalysis(SolrQueryRequest req) throws Exception;
+  protected abstract NamedList<?> doAnalysis(SolrQueryRequest req) throws Exception;
```
madrob, line 85: I think best practice is to use `NamedList<Object>` as the return type, and `NamedList<?>` as the argument type in methods, but I can't find a reference for it right now.

ErickErickson:
This has being handled in SOLR-10778

PR #1512, opened 2020-05-12
SOLR-13325: Add a collection selector to ComputePlanAction

shalinmangar:
# Description

The ComputePlanAction supports a `collections` attribute that can be used to filter collections for which the plan is computed. This currently accepts a comma-separated `list` of collection names. If the attribute is empty or omitted then all collections are supposed to match.

However, it is difficult to maintain such a list in a system where collections are added and removed dynamically. It is easier to have a collection selector that matches zero or more collections by filtering them against a given set of collection (property, value) pairs. This can then be used to e.g. match all collections that use a specific policy etc.

# Solution

This PR adds collection selector support to the `ComputePlanAction` which is of the form `collections: {policy: my_policy}`. If more than one key/value pair is specified then all must match for a collection to be considered. The old way of specifying a string for `collections` is preserved for back-compatibility.

After this support, the `AutoAddReplicasPlanAction` becomes a thin shim over ComputePlanAction with a collection selector `collections: {autoAddReplicas: true}`.

The old implementation did not pass collection hints to the policy engine which meant that all possible operations were computed by the policy and then they were filtered out by the ComputePlanAction. In this PR, the collection hints are pushed down to the policy engine so operations for non-matching collections are not even computed in the first place.

# Tests

The ComputePlanActionTest.testSelectedCollections is now refactoted into two test methods viz. `testSelectedCollectionsByName` and `testSelectedCollectionsByPolicy`. The latter tests the collection selector by policy.

# Checklist

- [X] I have added documentation for the [Ref Guide](https://github.com/apache/lucene-solr/tree/master/solr/solr-ref-guide) (for Solr changes only).


solr/solr-ref-guide/src/solrcloud-autoscaling-trigger-actions.adoc
```diff
@@ -29,12 +29,13 @@ commands which can re-balance the cluster in response to trigger events.
 The following parameters are configurable:
 
 `collections`::
-A comma-separated list of collection names. If this list is not empty then
-the computed operations will only calculate collection operations that affect
-listed collections and ignore any other collection operations for collections
+A comma-separated list of collection names. This can also be a selector on the collection property e.g. `collections: {'policy': 'my_custom_policy'}` will match all collections which use the policy named `my_customer_policy`.
```
sigram, line 32: It's a very terse description of a new powerful feature... IMHO at least it should mention that all other collection properties are supported, the property value must be equal, the implicit AND is used when multiple props are specified, and perhaps it should give some other examples.

solr/core/src/java/org/apache/solr/cloud/autoscaling/AutoAddReplicasPlanAction.java
```diff
@@ -18,46 +18,23 @@
...
-        anyCollections = true;
-        suggester.hint(Suggester.Hint.COLL, collection.getName());
-      }
-    }
-
-    if (!anyCollections) return NoneSuggester.get(session);
-    return suggester;
+  public void configure(SolrResourceLoader loader, SolrCloudManager cloudManager, Map<String, Object> properties) throws TriggerValidationException {
+    properties.put("collections", Collections.singletonMap(AUTO_ADD_REPLICAS, "true"));
+    super.configure(loader, cloudManager, properties);
```
sigram, line 38: This is indeed a nice cleanup!

solr/core/src/java/org/apache/solr/cloud/autoscaling/ComputePlanAction.java
```diff
@@ -17,38 +17,28 @@
...
-import org.apache.solr.common.params.CoreAdminParams;
 import org.apache.solr.common.util.Pair;
 import org.apache.solr.common.util.StrUtils;
 import org.apache.solr.core.SolrResourceLoader;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.io.IOException;
+import java.lang.invoke.MethodHandles;
+import java.util.*;
```
sigram, line 37: IIRC using wildcard imports was frowned upon?

solr/solr-ref-guide/src/solrcloud-autoscaling-trigger-actions.adoc
shalinmangar: Thanks. I expanded the description a bit. I added another example that I am not really happy with -- I cannot think of a use-case where something other than policy might be useful especially given the fact that values must match exactly. Suggestions welcome.

solr/core/src/java/org/apache/solr/cloud/autoscaling/ComputePlanAction.java
shalinmangar: I just imported the project in idea using the gradle scripts. precommit also seems happy. I don't recall a rule about wildcards import?

solr/solr-ref-guide/src/solrcloud-autoscaling-trigger-actions.adoc
sigram: There's a typo (name mismatch) in the policy name in the doc - 'my_custom_policy' vs. 'my_customER_policy'.

One other possible property that comes to my mind would be 'router' to separate collections with TRA / CRA .. but in this case the collection could equally well be set up with its own policy - so for now maybe there's no other property that makes sense. :)

solr/core/src/java/org/apache/solr/cloud/autoscaling/ComputePlanAction.java
sigram: Right, there's no formal rule. I checked the expanded list of imports here - indeed it's very long.

solr/solr-ref-guide/src/solrcloud-autoscaling-trigger-actions.adoc
shalinmangar: Thanks. I had reworded the description so the `my_custom_policy` and `my_customer_policy` are both replaced with `my_policy` everywhere. I'll keep the example here because it conveys how to use more than one key/value pair.

PR #1514, opened 2020-05-13
SOLR-13749: Change cross-collection join query syntax to {!join method=crossCollection ...}

danmfox:
# Description

Updates the cross-collection join query in #976 based on the feedback in SOLR-13749.  In that ticket there was a preference to consolidate the cross-collection join functionality into the existing join query parser, rather than creating a new separate query parser.

# Solution

This PR integrates the cross-collection join query parser into the existing join query parser plugin.  The syntax for a cross-collection join changes from `{!xcjf ...}` to `{!join method=crossCollection ...}`.  The arguments that could previously be set on the XCJF query parser plugin can now be set on the join query parser plugin.

# Tests

The XCJFQueryTest class has been updated to use the new query syntax (and renamed to CrossCollectionJoinQueryTest).

# Checklist


solr/core/src/java/org/apache/solr/search/JoinQParserPlugin.java
```diff
@@ -160,23 +176,40 @@ JoinParams parseJoin(QParser qparser) throws SyntaxError {
...
   }
 
+  @Override
+  public void init(NamedList args) {
+    routerField = (String) args.get("routerField");
+    solrUrlWhitelist = new HashSet<>();
+    if (args.get("solrUrlWhitelist") != null) {
+      //noinspection unchecked
+      for (String s : (List<String>) args.get("solrUrlWhitelist")) {
+        if (!StringUtils.isEmpty(s))
```
dsmiley, line 186: This is fine, really, but personally I'd just go with 1 line of code -- `solrUrlWhiteList.addAll((List<String>) args.get("solrUrlWhitelist"))` because the impact of an empty string here is safe.

dsmiley:
If someone uses join QParser with either no method or the default method=index and if fromIndex refers to a non-qualifying collection (sharded or not node-local), they'll get one of the errors printed from this method: `org.apache.solr.search.join.ScoreJoinQParserPlugin#findLocalReplicaForFromIndex`.
* `"SolrCloud join: multiple shards not yet supported " + fromIndex`.  I suggest rewording: `To join with a sharded collection, use method=crossCollection.`
* `"SolrCloud join: No active replicas for "+fromIndex+" found in node " + nodeName`.  I suggest rewording: `To join with a collection that might not be co-located, use method=crossCollection`

There is another case as well when the local replica is not in an Active state but lets ignore that.

dsmiley:
@danmfox Still missing Ref Guide changes to account for the refactor

dsmiley:
Last minute request: also please change solrUrlWhitelist to avoid this "whitelist" word (think current events) to, I propose "allowSolrUrls".  There's a discussion going on now internally that probably should be public about this sort of thing.  There's another patch to add a new similar thing for file system paths deliberately named "allowPaths", so that influenced my suggested new name here.

solr/core/src/java/org/apache/solr/search/JoinQParserPlugin.java
danmfox: Sounds good to me - I made this change while updating the property name.

PR #1519, opened 2020-05-15
SOLR-14482: Fix compile-time warnings in solr/core/search/facet

ErickErickson:
The facet code has a lot of classes declared in a file with a different name. This tries to fix all the warnings in solr/core/search/facet. gradlew check passes. Here for comments.

Plus one log message I noticed that I'd checked in when I was debugging.

Make any comments by EOD Friday, I'll be pushing this over the weekend otherwise after verifying that it all works under ant precommit/test too and checking it over one more time.


ErickErickson:
Reworking, I'll put up another PR.

PR #1521, opened 2020-05-15
Lucene-9371: Allow external access to RegExp's parsed structure

markharwood:
The RegExp parser holds a useful representation of the parsed logic in a regular expression but does not have a public API that would allow the objects to be transformed to new forms e.g. to render to an English explanation of the logic.
This PR exposes that capability by making the internal fields public and final. I toyed with the idea of just adding getter methods but the internal state of this class should ideally be made immutable anyway so instead I opted for making the parsed fields final.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/util/automaton/RegExp.java
+++ b/lucene/core/src/java/org/apache/lucene/util/automaton/RegExp.java
@@ -365,8 +365,43 @@ import java.util.Set;
  */
 public class RegExp {
   
-  enum Kind {
-    REGEXP_UNION, REGEXP_CONCATENATION, REGEXP_INTERSECTION, REGEXP_OPTIONAL, REGEXP_REPEAT, REGEXP_REPEAT_MIN, REGEXP_REPEAT_MINMAX, REGEXP_COMPLEMENT, REGEXP_CHAR, REGEXP_CHAR_RANGE, REGEXP_ANYCHAR, REGEXP_EMPTY, REGEXP_STRING, REGEXP_ANYSTRING, REGEXP_AUTOMATON, REGEXP_INTERVAL,
+  /**
+   * The type of expression represented by a RegExp node.
+   */
+  public enum Kind {
+    /** The union of two expressions */
+    REGEXP_UNION, 
+    /** A sequence of two expressions */
+    REGEXP_CONCATENATION,
+    /** The intersection of two expressions */
+    REGEXP_INTERSECTION,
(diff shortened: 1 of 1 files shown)


lucene/core/src/java/org/apache/lucene/util/automaton/RegExp.java
```diff
@@ -365,8 +365,43 @@
  */
 public class RegExp {
   
-  enum Kind {
-    REGEXP_UNION, REGEXP_CONCATENATION, REGEXP_INTERSECTION, REGEXP_OPTIONAL, REGEXP_REPEAT, REGEXP_REPEAT_MIN, REGEXP_REPEAT_MINMAX, REGEXP_COMPLEMENT, REGEXP_CHAR, REGEXP_CHAR_RANGE, REGEXP_ANYCHAR, REGEXP_EMPTY, REGEXP_STRING, REGEXP_ANYSTRING, REGEXP_AUTOMATON, REGEXP_INTERVAL,
+  /**
+   * The type of expression represented by a RegExp node.
```
mikemccand, line 369: Can we add `@lucene.experimental` or `@lucene.internal`?

Ahh, nevermind -- the class level javadocs say so already.

lucene/core/src/java/org/apache/lucene/util/automaton/RegExp.java
```diff
@@ -468,6 +519,37 @@ public RegExp(String s, int syntax_flags) throws IllegalArgumentException {
     from = e.from;
     to = e.to;
   }
+  
+  RegExp(Kind kind, RegExp exp1, RegExp exp2, String s, int c, int min, int max, int digits, int from, int to){    
+    this.originalString = null;
+    this.kind = kind;
+    this.flags = 0;
+    this.exp1=exp1;
```
mikemccand, line 527: Can we be consistent about the spaces around `=`?

lucene/core/src/java/org/apache/lucene/util/automaton/RegExp.java
```diff
@@ -468,6 +519,37 @@ public RegExp(String s, int syntax_flags) throws IllegalArgumentException {
...
+    this.exp2=exp2;
+    this.s = s;
+    this.c = c;
+    this.min =min;
```
mikemccand, line 531: Hmm also here.  Still looking for the last permutation ...

markharwood:
Thanks for looking at this, Mike!

PR #1522, opened 2020-05-16
LUCENE-9374: Add checkBrokenLinks gradle task

mocobeta:
## Description

~This adds `documentationLint` task to gradle build.~ This adds `checkBrokenLinks` task to gradle build. The task is a placeholder for `:lucene:checkBrokenLinks` and `:solr:checkBrokenLinks` that detect broken or invalid links in docs under `_project_/build/documentation`. 
For now, both check tasks fail as there are broken links in the javadocs.

See also https://issues.apache.org/jira/browse/LUCENE-9374

## Gradle command

```
./gradlew checkBrokenLinks
```

or 

```
./gradlew :lucene:checkBrokenLinks
./gradlew :solr:checkBrokenLinks
```

## Note

- I added `documentationLint` task that wraps `checkBrokenLinks` to align semantics with `documentation` task, but it can be removed if not needed.
- `documentation-lint.gradle` could be placed under `gradle/documentation` ?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -103,6 +103,7 @@ apply from: file('gradle/validation/ecj-lint.gradle')
 apply from: file('gradle/validation/gradlew-scripts-tweaked.gradle')
 apply from: file('gradle/validation/missing-docs-check.gradle')
 apply from: file('gradle/validation/validate-log-calls.gradle')
+apply from: file('gradle/validation/documentation-lint.gradle')
 
 // Source or data regeneration tasks
 apply from: file('gradle/generation/jflex.gradle')
@@ -134,4 +135,3 @@ apply from: file('gradle/ant-compat/forbidden-api-rules-in-sync.gradle')
 apply from: file('gradle/documentation/documentation.gradle')
 apply from: file('gradle/documentation/changes-to-html.gradle')
 apply from: file('gradle/render-javadoc.gradle')
-apply from: file('gradle/validation/documentation-lint.gradle')  // should be placed here; needs to be evaluated after documentation.gradle
--- a/build.gradle
+++ b/build.gradle
@@ -134,3 +134,4 @@ apply from: file('gradle/ant-compat/forbidden-api-rules-in-sync.gradle')
 apply from: file('gradle/documentation/documentation.gradle')
 apply from: file('gradle/documentation/changes-to-html.gradle')
 apply from: file('gradle/render-javadoc.gradle')
+apply from: file('gradle/validation/documentation-lint.gradle')  // should be placed here; needs to be evaluated after documentation.gradle
--- a/build.gradle
+++ b/build.gradle
@@ -115,7 +115,7 @@ apply from: file('gradle/validation/ecj-lint.gradle')
 apply from: file('gradle/validation/gradlew-scripts-tweaked.gradle')
 apply from: file('gradle/validation/missing-docs-check.gradle')
 apply from: file('gradle/validation/validate-log-calls.gradle')
-apply from: file('gradle/validation/documentation-lint.gradle')
+apply from: file('gradle/validation/check-broken-links.gradle')
 
 // Source or data regeneration tasks
 apply from: file('gradle/generation/jflex.gradle')
--- a/gradle/validation/check-broken-links.gradle
+++ b/gradle/validation/check-broken-links.gradle
@@ -17,23 +17,21 @@
 
 configure(rootProject) {
 
-  task documentationLint {
+  task checkBrokenLinks {
     group 'Verification'
-    description 'Verify all documentation'
+    description 'Check broken links in the entire documentation'
 
     dependsOn ':lucene:checkBrokenLinks'
     dependsOn ':solr:checkBrokenLinks'
   }
-}
 
+}
 configure(subprojects.findAll { it.path in [':lucene', ':solr'] }) {
 
   project.afterEvaluate {
     task checkBrokenLinks(type: CheckBrokenLinksTask) {
       docsDir = project.docroot
-      dependsOn subprojects.collect { prj ->
-        prj.tasks.matching { it.name == 'renderSiteJavadoc' }
-      }
+      dependsOn 'documentation'
     }
 
     // TODO: uncomment this line after fixing all broken links.
--- a/gradle/validation/check-broken-links.gradle
+++ b/gradle/validation/check-broken-links.gradle
@@ -37,8 +37,11 @@ configure(subprojects.findAll { it.path in [':lucene', ':solr'] }) {
 
 class CheckBrokenLinksTask extends DefaultTask {
 
+  // wraps input directory location in DirectoryProperty so as to lazily evaluate 'docroot' property
+  // (see gradle/documentation/documentation.gradle)
   @InputDirectory
-  final Provider<File> docsDir = project.providers.provider { project.docroot }
+  final DirectoryProperty docsDir = project.objects.directoryProperty()
+    .fileProvider(project.providers.provider { project.docroot })
 
   @InputFile
   File script = project.rootProject.file("dev-tools/scripts/checkJavadocLinks.py")
@@ -56,7 +59,7 @@ class CheckBrokenLinksTask extends DefaultTask {
         args = [
           "-B",
           script.absolutePath,
-          docsDir.get().absolutePath
+          docsDir.get().getAsFile()
         ]
       }
     }
--- a/gradle/validation/check-broken-links.gradle
+++ b/gradle/validation/check-broken-links.gradle
@@ -37,10 +37,10 @@ configure(subprojects.findAll { it.path in [':lucene', ':solr'] }) {
 
 class CheckBrokenLinksTask extends DefaultTask {
 
-  @Input
+  @InputDirectory
   Provider<File> docsDir = project.providers.provider { project.docroot }
 
-  @Input
+  @InputFile
   File script = project.rootProject.file("dev-tools/scripts/checkJavadocLinks.py")
 
   @TaskAction
--- a/gradle/validation/check-broken-links.gradle
+++ b/gradle/validation/check-broken-links.gradle
@@ -40,6 +40,9 @@ class CheckBrokenLinksTask extends DefaultTask {
   @Input
   Provider<File> docsDir = project.providers.provider { project.docroot }
 
+  @Input
+  File script = project.rootProject.file("dev-tools/scripts/checkJavadocLinks.py")
+
   @TaskAction
   def check() {
     def outputFile = project.file("${getTemporaryDir()}/check-broken-links-output.txt")
@@ -52,7 +55,7 @@ class CheckBrokenLinksTask extends DefaultTask {
         errorOutput = output
         args = [
           "-B",
-          project.rootProject.file("dev-tools/scripts/checkJavadocLinks.py").absolutePath,
+          script.absolutePath,
           docsDir.get().absolutePath
         ]
       }
--- a/gradle/validation/check-broken-links.gradle
+++ b/gradle/validation/check-broken-links.gradle
@@ -38,7 +38,7 @@ configure(subprojects.findAll { it.path in [':lucene', ':solr'] }) {
 class CheckBrokenLinksTask extends DefaultTask {
 
   @InputDirectory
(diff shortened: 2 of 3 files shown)


build.gradle
```diff
@@ -134,3 +134,4 @@ apply from: file('gradle/ant-compat/forbidden-api-rules-in-sync.gradle')
 apply from: file('gradle/documentation/documentation.gradle')
 apply from: file('gradle/documentation/changes-to-html.gradle')
 apply from: file('gradle/render-javadoc.gradle')
+apply from: file('gradle/validation/documentation-lint.gradle')  // should be placed here; needs to be evaluated after documentation.gradle
```
dweiss, line 137: There is already a comment about include ordering above I believe, but fine.

gradle/validation/documentation-lint.gradle
```diff
@@ -0,0 +1,65 @@
...
+ */
+
+configure(rootProject) {
+
+  task documentationLint {
+    group 'Verification'
+    description 'Verify all documentation'
+
+    dependsOn 'documentation'
+    dependsOn subprojects.collect { prj ->
```
dweiss, line 25: This can be explicit since there are only two? (":lucene:checkBrokenLinks", ":solr:checkBrokenLinks")?

gradle/validation/documentation-lint.gradle
```diff
@@ -0,0 +1,65 @@
...
+      prj.tasks.matching { task -> task.name == 'checkBrokenLinks'}
+    }
+  }
+
+  // TODO: uncomment this line after fixing all broken links.
```
dweiss, line 30: I would add (commented out) check.dependsOn checkBrokenLinks to task declaration per-project as well so that when you run gradlew -p lucene check the validation runs on a per-project basis too.

gradle/validation/documentation-lint.gradle
```diff
@@ -0,0 +1,65 @@
...
+  // check.dependsOn documentationLint
+}
+
+configure(subprojects.findAll { it.path in [':lucene', ':solr'] }) {
+  task checkBrokenLinks(type: CheckBrokenLinksTask)
+}
+
+class CheckBrokenLinksTask extends DefaultTask {
+  @Input
+  File docsDir = project.docroot
```
dweiss, line 41: (I don't know how to do this without changing field signature) but ideally this should be a lazy provider returning project.docroot. This would make it evaluation-order independent and would trigger input evaluation only if the task actually runs.

Just a thought if you want to get deeper into gradle. ;)

build.gradle
mocobeta: With lazy evaluation for 'docroot', the evaluation order does not matter so I moved this line to proper location. (I clearly missed the comment "The order of inclusion of these files shouldn't matter".)

gradle/validation/documentation-lint.gradle
mocobeta: I updated the task dependencies. 

Should we remove entire `documentationLink` task, or is it useful to have?

gradle/validation/documentation-lint.gradle
mocobeta: Removed this, instead I added "check.dependsOn checkBrokenLinks".

gradle/validation/documentation-lint.gradle
mocobeta: Thanks, `project.afterEvaluate()` works for me (though can't figure out the way to lazily evaluate ext properties for the task field).

mocobeta: Btw, render-javadoc.gradle (renderSiteJavadoc) also eagarly evaluate the property (I first referred the task and did the same), so it fails if the order is changed in root `build.gradle`. 
https://github.com/apache/lucene-solr/blob/master/gradle/render-javadoc.gradle#L59

dweiss: I wasn't talking about afterEvaluate, actually. I think some properties can be declared as lazy and then assigned a provider - this is different than afterEvaluate. Look here: 
https://docs.gradle.org/current/userguide/lazy_configuration.html

As for order dependencies: afterEvaluate only shifts the problem until later because you still can have dependency ordering problems between afterEvaluate closures... ;) It's not the perfect solution (but of course works reasonably well most of the time).

mocobeta: ah... thanks for correction. I misunderstood the meaning of afterEvaluation. I will come back next week.

gradle/validation/documentation-lint.gradle
```diff
@@ -0,0 +1,72 @@
...
+    dependsOn ':solr:checkBrokenLinks'
+  }
+}
+
+configure(subprojects.findAll { it.path in [':lucene', ':solr'] }) {
+
+  project.afterEvaluate {
+    task checkBrokenLinks(type: CheckBrokenLinksTask) {
+      docsDir = project.docroot
+      dependsOn subprojects.collect { prj ->
```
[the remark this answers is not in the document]
uschindler: Sorry I forgot to submit the review. I hate that github page, especially on smartphone!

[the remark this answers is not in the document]
uschindler: Just depend on `:lucene:documentation`and `:solr:documentation`. Will merge #1477 soon.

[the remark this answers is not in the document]
uschindler: Thats the IMHO "better" code after #1477 is merged:

`dependsOn "${path}:documentation"`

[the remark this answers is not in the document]
uschindler: OK #1477 is merged to master, just update your PR and change dependency. Thanks, Uwe

gradle/validation/documentation-lint.gradle
uschindler: You can have a look at #1477, there I am also using lazy evaluation (see `bindings`) property. It's a MapProvider and the values are lazily evaluated using `project.providers.property{ ... lazy code ... }`

gradle/validation/documentation-lint.gradle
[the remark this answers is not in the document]
mocobeta: Thanks for the change. I merged the latest master and made "checkBrokenLinks" depend on "documentation".

mocobeta:
I changed the top-level 'documentationLint' task name to 'checkBrokenLinks'

gradle/validation/documentation-lint.gradle
uschindler: You can use a `RegularFileProperty` instead of a File and then use `providers.property{...}` to assign value. Gradle DSL does not change, just when using it from Groovy code or Java code, you have to call `get().getAsFile()` (which does lazy evaluation).
I did not use that for render-javadocs, becuase it was not needed.

Another "simple" way for lazy evaluation (works only with Strings, but its often helpful) is to use GString (double quoted Strings): `GString someProperty = "${->project.docroot}"` (please note the `->`). Whenever `toString()` or `as String` is called, its lazinly evaluated. Without the `->` the value is inserted into the string at parsing time.

I had the same problem with #1477, but I will plan to do this for the URL coming from an extension:

```groovy
  @Input
  def luceneDocUrl = "${->project.luceneDocUrl}"

  @Input
  def solrDocUrl = "${->project.solrDocUrl}"  
```

uschindler: See b7c60e53eb8d7c7c01ccd7c00add2662e503fd08

mocobeta: Thanks for the pointers. `project.providers.provider { project.docroot }` perfectly works for me. https://github.com/apache/lucene-solr/pull/1522/commits/ac195d147a3b7d25605ca98034028f9719913782 
(`${->project.docroot}` also works fine; I like using `Provider<File>` signature here.)

On the master branch, `gradlew renderSiteJavadoc` fails with this error when I reverse evaluation order of `gradle/render-javadoc.gradle` and `gradle/documentation/documentation.gradle`.
```
> Could not get unknown property 'docroot' for project ':solr:core' of type org.gradle.api.Project.
```
I think it'd be better doing the same for render-javadoc.gradle just for safety, or add some comments about the ordering on `build.gradle` to prevent someone accidentally changes the file?

gradle/validation/check-broken-links.gradle
```diff
@@ -0,0 +1,68 @@
...
+  task checkBrokenLinks(type: CheckBrokenLinksTask, 'dependsOn': 'documentation')
+
+  // TODO: uncomment this line after fixing all broken links.
+  // (we can't fix the cross-project links until ant build is disabled.)
+  // check.dependsOn checkBrokenLinks
+}
+
+class CheckBrokenLinksTask extends DefaultTask {
+
+  @Input
```
uschindler, line 40: Should be `@InputFile`

gradle/validation/check-broken-links.gradle
```diff
@@ -0,0 +1,68 @@
...
+  Provider<File> docsDir = project.providers.provider { project.docroot }
+
+  @Input
```
uschindler, line 43: same here

mocobeta: Fixed.

gradle/validation/check-broken-links.gradle
mocobeta: I replaced it with `@InputDirectory` (since `@InputFile` does not work for me.)

uschindler: of course it's a directory, sorry!

dweiss: For reference - the difference is that Input is  just the file, InputFIle is the file and its contents (!). 
https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/InputFile.html

dweiss: This works for me and is what I had in mind:
```
 configure(subprojects.findAll { it.path in [':lucene', ':solr'] }) {
 
-  task checkBrokenLinks(type: CheckBrokenLinksTask, 'dependsOn': 'documentation')
+  task checkBrokenLinks(type: CheckBrokenLinksTask, 'dependsOn': 'documentation') {
+    docsDir = project.layout.dir provider { project.docroot }
+  }
 
   // TODO: uncomment this line after fixing all broken links.
   // (we can't fix the cross-project links until ant build is disabled.)
@@ -38,7 +39,7 @@ configure(subprojects.findAll { it.path in [':lucene', ':solr'] }) {
 class CheckBrokenLinksTask extends DefaultTask {
 
   @InputDirectory
-  final Provider<File> docsDir = project.providers.provider { project.docroot }
+  final DirectoryProperty docsDir = project.objects.directoryProperty()
 
   @InputFile
   File script = project.rootProject.file("dev-tools/scripts/checkJavadocLinks.py")
@@ -56,7 +57,7 @@ class CheckBrokenLinksTask extends DefaultTask {
         args = [
           "-B",
           script.absolutePath,
-          docsDir.get().absolutePath
+          docsDir.get().asFile
         ]
       }
     } 
```

Any version is fine though, I don't mind.

mocobeta: Thank you, https://github.com/apache/lucene-solr/pull/1522/commits/778701c97120fbc2e6578a545d7719ac19d82d75 works for me.

uschindler: Thanks!

uschindler:
I think you can do the same "Provider<File> -> DirectoryProperty" change in the changes-to-html task that you modified yesterday.

mocobeta:
I fixed it on https://github.com/apache/lucene-solr/commit/51c15b8660b9d6a3e5979bf1e53a98262db413e9 (and accidentally logged "merge" commit in the master https://github.com/apache/lucene-solr/commit/3b09d30e69795388e655275d23f1e83be0556da7; sorry.)

dweiss:
It's fine, although the snippet I copy-pasted showed a way which I think is more elegant in the long term: separates the actual property configuration bit from its declaration in the task. The difference is subtle - if we move the task to buildSrc, for example, the configuration can still happen in the relevant part of the script  but the task stays precompiled and ready. These are tiny things that definitely can be polished later on.

mocobeta:
Thanks for reviewing. I plan to merge the branch to the master in shortly (if there's no other feedback).

PR #1523, opened 2020-05-17
SOLR-14494: Refactor BlockJoin to not use Filter

dsmiley:
https://issues.apache.org/jira/browse/SOLR-14494

In this PR I may have over-reached the refactorings beyond the scope a bit; it could be scaled back.  I didn't get why the cache interaction was as complex as it was so I simplified it but I may be wrong on this.
CC @mkhludnev


dsmiley:
Just checking -- you have no concerns my simplification of use of the cache, and caching the BitSetProducer as the value instead of some Filter/Query wrapper?

And I'm thinking just 9x only, although backporting wouldn't be a big deal.

PR #1524, opened 2020-05-18
SOLR-15185: Rewrite Hash query

dsmiley:
* Simpler
* Don't use Filter (to be removed)
* Do use TwoPhaseIterator, not PostFilter
* Don't pre-compute matching docs (wasteful)
* Support more fields, and more field types
* Faster hash on Strings (avoid Char conversion)
* Stronger hash when using multiple fields

----
@joel-bernstein, I set out to remove the use of Filter in `{!hash}` (because https://issues.apache.org/jira/browse/SOLR-12336 ) and then I saw a bunch of things that I felt could be improved.  Apparently I had nothing better to do this weekend 😄   j/k     WDYT?

### DIFF
--- a/solr/CHANGES.txt
+++ b/solr/CHANGES.txt
@@ -88,6 +88,9 @@ when told to. The admin UI now tells it to. (Nazerke Seidan, David Smiley)
 * SOLR-15100: Make the ConfigSetService pluggable/configurable via <string name="configSetService" /> in solr.xml
  (baisui)
 
+* SOLR-15185: Various optimizations to the {!hash} QParser, typically used by the parallel()
+  streaming expression.  The hash algorithm changed.  (David Smiley)
+
 Other Changes
 ----------------------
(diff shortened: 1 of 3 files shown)


dsmiley:
@joel-bernstein I'd like to finally get this merged.  If you feel you need more time to review, LMK.  I feel pretty good about it.
The the upgrade docs will mention that the computed hashes are different.

joel-bernstein:
You moved quicker than I thought. Andrzej and I are also working on a revamp of this, but we can start from your PR. Feel free to commit if you're happy with your PR. We'll revisit with another ticket.

joel-bernstein:
It sounds like you had some good performance improvements, do you have any performance numbers? Andrzej and I will be looking primarily performance.

dsmiley:
No numbers.  If you & @sigram are going to revamp then maybe I shouldn't merge this and you can work directly off of this PR (perhaps rebased off of master).  If that's not expected for months (after 9) then I'd prefer to merge now.

sigram:
@dsmiley the improvements Joel and I are considering are totally independent of this cleanup - please proceed with this PR.

PR #1527, opened 2020-05-18
SOLR-14384 Stack SolrRequestInfo

NazerkeBS:
# Description

# Solution

# Tests

# Checklist


solr/core/src/java/org/apache/solr/request/SolrRequestInfo.java
```diff
@@ -52,35 +54,44 @@
...
   public static void setRequestInfo(SolrRequestInfo info) {
-    // TODO: temporary sanity check... this can be changed to just an assert in the future
-    SolrRequestInfo prev = threadLocal.get();
-    if (prev != null) {
-      log.error("Previous SolrRequestInfo was not closed!  req={}", prev.req.getOriginalParams());
-      log.error("prev == info : {}", prev.req == info.req, new RuntimeException());
+    if (info == null) {
+      throw new IllegalArgumentException("SolrRequestInfo is null");
+    } else {
+      threadLocal.get().push(info);
```
mkhludnev, line 65: Shouldn't we limit maximum stack depth?
What if one component leak SolrRequestInfo accidentally, like push it on every request but refuse to pop it when  exception occur. Thus it may steadily flood heap. Or I miss something?

solr/core/src/java/org/apache/solr/servlet/SolrDispatchFilter.java
```diff
@@ -456,6 +457,7 @@ public void doFilter(ServletRequest _request, ServletResponse _response, FilterC
 
       GlobalTracer.get().clearContext();
       consumeInputFully(request, response);
+      SolrRequestInfo.reset();
```
mkhludnev, line 460: there we need to check that stack have fully cleared and disposed

solr/core/src/java/org/apache/solr/request/SolrRequestInfo.java
dsmiley: Good idea!

solr/core/src/java/org/apache/solr/servlet/SolrDispatchFilter.java
dsmiley: That is what `reset` _does_

solr/core/src/java/org/apache/solr/request/SolrRequestInfo.java
NazerkeBS: what would you suggest for the stack size limit?

mkhludnev: 100 for sure

solr/core/src/java/org/apache/solr/request/SolrRequestInfo.java
```diff
@@ -38,7 +40,13 @@
 
 
 public class SolrRequestInfo {
-  protected final static ThreadLocal<SolrRequestInfo> threadLocal = new ThreadLocal<>();
+
+  protected final static int capacity = 150;
+
+  protected final static ThreadLocal<Deque<SolrRequestInfo>> threadLocal = ThreadLocal.withInitial(() -> {
+    Deque<SolrRequestInfo> stack = new ArrayDeque<>(capacity);
```
dsmiley, line 47: I think you can simply do a lambda reference to ArrayDequeue::new.  No need initialize with an internal capacity.  Honestly I'd prefer a LinkedList because in practice this stack will be extremely small (zero, one, sometimes two, very unlikely more).  But I leave that impl choice to you.

solr/core/src/java/org/apache/solr/request/SolrRequestInfo.java
```diff
@@ -38,7 +40,13 @@
 
 
 public class SolrRequestInfo {
-  protected final static ThreadLocal<SolrRequestInfo> threadLocal = new ThreadLocal<>();
+
+  protected final static int capacity = 150;
```
dsmiley, line 44: This should be MAX_STACK_SIZE I think.  
Can anyone ( @mkhludnev ?) venture to guess how, realistically, we might reach upwards of 150?  I can't imagine more than a few, let alone 150.

solr/core/src/java/org/apache/solr/request/SolrRequestInfo.java
```diff
@@ -52,35 +60,48 @@
   private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
 
   public static SolrRequestInfo getRequestInfo() {
-    return threadLocal.get();
+    if (threadLocal.get().isEmpty()) return null;
+    return threadLocal.get().peek();
   }
 
   public static void setRequestInfo(SolrRequestInfo info) {
```
dsmiley, line 67: Add javadoc please

solr/core/src/java/org/apache/solr/request/SolrRequestInfo.java
```diff
@@ -52,35 +60,48 @@
...
+      } else {
+        log.error("SolrRequestInfo Stack is full");
+      }
     }
-    assert prev == null;
-
-    threadLocal.set(info);
   }
 
   public static void clearRequestInfo() {
```
dsmiley, line 79: Add javadoc please

solr/core/src/java/org/apache/solr/request/SolrRequestInfo.java
```diff
@@ -52,35 +60,48 @@
...
-          }
+    if (threadLocal.get().isEmpty()) {
+      log.error("clearRequestInfo called too many times");
+    } else {
+      SolrRequestInfo info = threadLocal.get().pop();
+      closeHooks(info);
+    }
+  }
+
+  public static void reset() {
```
dsmiley, line 88: Add javadoc please

solr/core/src/java/org/apache/solr/request/SolrRequestInfo.java
NazerkeBS: Initially I set it to 100, then there was one test (CloudAuthStreamTest) was failing due to this capacity. So increased it to 150.

dsmiley: I suggest debugging/investigating there to understand _why_.  Maybe this cap has exposed an actual problem?

dsmiley:
Perhaps the reset method should furthermore have an assertion to check that the stack is already empty?  If we decide that reset() is there as a safety measure that isn't to be used instead of clear(), then the assertion can alert us to problems (in tests) rather than hiding them.

solr/core/src/java/org/apache/solr/request/SolrRequestInfo.java
```diff
@@ -52,35 +56,53 @@
   private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
 
   public static SolrRequestInfo getRequestInfo() {
-    return threadLocal.get();
+    if (threadLocal.get().isEmpty()) return null;
```
dsmiley, line 59: Instead of calling threadLocal.get() twice, can you please just call it once?  This goes for all the other methods in this class too.

solr/core/src/java/org/apache/solr/request/SolrRequestInfo.java
```diff
@@ -52,35 +56,53 @@
...
-    if (prev != null) {
-      log.error("Previous SolrRequestInfo was not closed!  req={}", prev.req.getOriginalParams());
-      log.error("prev == info : {}", prev.req == info.req, new RuntimeException());
+    if (info == null) {
+      throw new IllegalArgumentException("SolrRequestInfo is null");
+    } else {
+      if (threadLocal.get().size() <= MAX_STACK_SIZE) {
+        threadLocal.get().push(info);
+      } else {
+        log.error("SolrRequestInfo Stack is full");
```
dsmiley, line 71: add assertion

solr/core/src/java/org/apache/solr/request/SolrRequestInfo.java
```diff
@@ -52,35 +56,53 @@
...
-          }
+    if (threadLocal.get().isEmpty()) {
+      log.error("clearRequestInfo called too many times");
+    } else {
+      SolrRequestInfo info = threadLocal.get().pop();
+      closeHooks(info);
+    }
+  }
+
+  /** Removes all the SolrRequestInfos from the stack */
```
dsmiley, line 86: comment that we expect it to be empty by now because all "set" calls need to be balanced with a "clear".  Thus this reset method is more of a protection mechanism.

solr/core/src/java/org/apache/solr/servlet/HttpSolrCall.java
```diff
@@ -145,6 +145,8 @@
   public static final String INTERNAL_REQUEST_COUNT = "_forwardedCount";
 
   public static final Random random;
+
+  private boolean pushedSolrRequestInfo = false;
```
dsmiley, line 149: I know I suggested this name but lets call this "mustClearSolrRequestInfo"

solr/core/src/java/org/apache/solr/request/SolrRequestInfo.java
```diff
@@ -52,35 +56,60 @@
...
+    boolean isEmpty = stack.isEmpty();
+    while (!stack.isEmpty()) {
+      SolrRequestInfo info = stack.pop();
+      closeHooks(info);
+    }
+    assert isEmpty : "SolrRequestInfo Stack should have been cleared.";
+  }
+
+  private static void closeHooks(SolrRequestInfo info) {
+    if (info != null && info.closeHooks != null) {
```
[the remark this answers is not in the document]
NazerkeBS: it can be null; when getRequestInfo is called, if the stack.isEmpty(), it returns null;

solr/core/src/java/org/apache/solr/client/solrj/embedded/EmbeddedSolrServer.java
```diff
@@ -71,6 +71,7 @@
   protected final String coreName;
   private final SolrRequestParsers _parser;
   private final RequestWriterSupplier supplier;
+  private boolean mustClearSolrRequestInfo = false;
```
dsmiley, line 74: This change is highly suspicious.  Why is this boolean now a field of EmbeddedSolrServer?  It doesn't seem like *state* of this object and so I don't think it belongs here.  It seems like a transient status in the course of work within the method that sets/clears the requestInfo.

solr/core/src/java/org/apache/solr/request/SolrRequestInfo.java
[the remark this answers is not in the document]
dsmiley: The callers of this method will not pass null though; right?

solr/core/src/java/org/apache/solr/client/solrj/embedded/EmbeddedSolrServer.java
```diff
@@ -271,7 +273,9 @@ public void writeResults(ResultContext ctx, JavaBinCodec codec) throws IOExcepti
       throw new SolrServerException(ex);
     } finally {
       if (req != null) req.close();
-      SolrRequestInfo.clearRequestInfo();
+      if (mustClearSolrRequestInfo) {
```
dsmiley, line 276: Hmm; might it simply be enough to call clearRequestInfo if req != null, thus no new boolean?

NazerkeBS: actually that's enough

solr/core/src/java/org/apache/solr/request/SolrRequestInfo.java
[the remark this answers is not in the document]
NazerkeBS: in fact, it's true; as it is a private method and called inside SolrRequestInfo

PR #1530, opened 2020-05-22
SOLR-14690: QueryResultKey.[nc_]minExactCount

cpoerschke:
Allow with-minExactCount searches to use query cache entries from without-minExactCount searches.

https://issues.apache.org/jira/browse/SOLR-14690


cpoerschke:
Transferred to https://github.com/apache/solr/pull/13 pull request.

PR #1537, opened 2020-05-26
LUCENE-9381: Add SortOrder interface

romseygeek:
This commit extracts a new SortOrder interface from SortField, containing only
those methods that are directly necessary to implement a sort.


lucene/backward-codecs/src/test/org/apache/lucene/codecs/lucene70/Lucene70RWSegmentInfoFormat.java
```diff
@@ -89,7 +89,10 @@ public void write(Directory dir, SegmentInfo si, IOContext ioContext) throws IOE
       int numSortFields = indexSort == null ? 0 : indexSort.getSort().length;
       output.writeVInt(numSortFields);
       for (int i = 0; i < numSortFields; ++i) {
-        SortField sortField = indexSort.getSort()[i];
+        if (indexSort.getSort()[i] instanceof SortField == false) {
```
mikemccand, line 92: Hmm, maybe `indexSort` should just continue to take `SortField` for now?  Else this is sort of trappy -- only when `IndexWriter` goes to flush a segment to disk, will we (late) hit this exception?

Or, maybe we could do this check instead in `IndexWriterConfig`?

lucene/core/src/java/org/apache/lucene/index/IndexWriterConfig.java
```diff
@@ -463,16 +465,65 @@ public IndexWriterConfig setCommitOnClose(boolean commitOnClose) {
...
     return this;
   }
 
+  private Set<String> extractFields(Sort sort) {
+    Set<String> fields = new HashSet<>();
+    for (SortOrder sortOrder : sort.getSort()) {
+      IndexSorter sorter = sortOrder.getIndexSorter();
+      assert sorter != null;
+      try {
+        sorter.getDocComparator(new DocValuesLeafReader() {
```
mikemccand, line 484: Hmm this is a little bit scary hackity, compared to what we had before :)

Are you wanting to not add a `SortOrder.getField()`?

lucene/core/src/java/org/apache/lucene/search/Sort.java
```diff
@@ -22,80 +22,7 @@
 
 
 /**
- * Encapsulates sort criteria for returned hits.
- *
- * <p>The fields used to determine sort order must be carefully chosen.
```
mikemccand, line 27: Wow, these docs were quite stale!  They hark from the `FieldCache` days.  Wow, 2004!

I think instead of fully deleting them, we should just updating them to state that you index the corresponding typed doc values field, and then sort by that type?

lucene/core/src/java/org/apache/lucene/search/SortField.java
```diff
@@ -138,7 +146,7 @@ public Provider() {
     }
 
     @Override
-    public SortField readSortField(DataInput in) throws IOException {
+    public SortOrder readSortField(DataInput in) throws IOException {
```
mikemccand, line 149: Hmm, if `SortOrder` is not (necessarily) serializable, and `SortField` is, shouldn't this return a `SortField`?

lucene/core/src/java/org/apache/lucene/search/SortField.java
```diff
@@ -172,8 +180,9 @@ public SortField readSortField(DataInput in) throws IOException {
     }
 
     @Override
-    public void writeSortField(SortField sf, DataOutput out) throws IOException {
-      sf.serialize(out);
+    public void writeSortField(SortOrder sf, DataOutput out) throws IOException {
```
mikemccand, line 183: Similarly, shouldn't this take a `SortField`?

lucene/core/src/java/org/apache/lucene/search/SortField.java
```diff
@@ -172,8 +180,9 @@ public SortField readSortField(DataInput in) throws IOException {
...
+      assert sf instanceof SortField;
+      ((SortField)sf).serialize(out);
```
mikemccand, line 185: And then these spooky `assert` / casts wouldn't be needed -- `javac` would do it for us?

lucene/core/src/java/org/apache/lucene/search/SortField.java
```diff
@@ -172,8 +180,9 @@ public SortField readSortField(DataInput in) throws IOException {
     }
 
     @Override
-    public void writeSortField(SortField sf, DataOutput out) throws IOException {
-      sf.serialize(out);
+    public void writeSortField(SortOrder sf, DataOutput out) throws IOException {
+      assert sf instanceof SortField;
```
mikemccand, line 184: Shouldn't this be a real check?  Caller could legitimately mess this up?

lucene/core/src/java/org/apache/lucene/search/SortField.java
```diff
@@ -226,6 +235,10 @@ else if (missingValue == STRING_FIRST) {
     }
   }
 
+  public String getField() {
+    return name();
```
mikemccand, line 239: Hmm, why not return `field` like we do today?

lucene/core/src/java/org/apache/lucene/search/SortOrder.java
```diff
@@ -0,0 +1,141 @@
...
+  /**
+   * Whether the relevance score is needed to sort documents.
+   */
+  boolean needsScores();
+
+  /**
+   * A name for the sort order
+   */
+  default String name() {
+    return toString();
```
mikemccand, line 54: It's a little scary to me that `getField` comes down to `toString`...

lucene/core/src/java/org/apache/lucene/search/SortOrder.java
```diff
@@ -0,0 +1,141 @@
...
+   *
+   * @param searcher IndexSearcher to use during rewriting
+   * @return New rewritten SortOrder, or {@code this} if nothing has changed.
+   */
+  default SortOrder rewrite(IndexSearcher searcher) throws IOException {
+    return this;
+  }
+
+  /**
+   * Returns an {@link IndexSorter} used for sorting index segments by this SortField.
```
mikemccand, line 68: s/`for sorting index segments`/`for sorting documents within a single index segment`?

lucene/core/src/java/org/apache/lucene/search/SortOrder.java
```diff
@@ -0,0 +1,141 @@
...
+   *
+   * If the SortField cannot be used for index sorting (for example, if it uses scores or
+   * other query-dependent values) then this method should return {@code null}
```
mikemccand, line 71: s/`should`/`must`?

lucene/core/src/java/org/apache/lucene/search/SortOrder.java
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    @Override
+    public IndexSorter getIndexSorter() {
+      return null;
+    }
+
+    @Override
+    public String toString() {
+      return "<score>";
```
mikemccand, line 107: Hmm so callers of `getField()` would think this `<score>` is a field name?

What does `Sort.FIELD_SCORE.getField()` return today?

solr/core/src/java/org/apache/solr/schema/IndexSchema.java
```diff
@@ -1225,6 +1227,16 @@ public SchemaField getFieldOrNull(String fieldName) {
     return f;
   }
 
+  public SchemaField getFieldOrNull(SortOrder sortOrder) {
+    if (sortOrder instanceof SortField) {
```
mikemccand, line 1231: This is another scary hackity code!

solr/core/src/java/org/apache/solr/schema/IndexSchema.java
```diff
@@ -1225,6 +1227,16 @@ public SchemaField getFieldOrNull(String fieldName) {
...
+      String field = sortOrder.name();
```
mikemccand, line 1232: Couldn't you cast to `SortField` and use `.getField()`?

I think this `name`/`field`/`toString` triality is confusing...

lucene/backward-codecs/src/test/org/apache/lucene/codecs/lucene70/Lucene70RWSegmentInfoFormat.java
romseygeek: This is test code for a deprecated segment info format, so users should never hit this in production - might be better as an assertion?

lucene/core/src/java/org/apache/lucene/index/IndexWriterConfig.java
romseygeek: Right, because some sorts refer to multiple fields, or to none.  For example, if we want to add an IndexSorter to expression sorts, we need to check that none of the DV fields referred to in the expression are updateable.

lucene/core/src/java/org/apache/lucene/search/Sort.java
romseygeek: ++

lucene/core/src/java/org/apache/lucene/search/SortOrder.java
romseygeek: It currently returns `null`, which is pretty much the worst of both worlds IMO...

solr/core/src/java/org/apache/solr/schema/IndexSchema.java
romseygeek: We can cut Solr over to using SortOrder instead of SortField as well, but that would make this PR massive, so I was hoping to do that in a followup.

lucene/backward-codecs/src/test/org/apache/lucene/codecs/lucene70/Lucene70RWSegmentInfoFormat.java
mikemccand: Woops, sorry, I missed that this was in `backwards-codecs` for test only, you're right.  Let's leave it as a real if?

lucene/core/src/java/org/apache/lucene/index/IndexWriterConfig.java
mikemccand: Wow, that makes sense -- could you add a comment explaining that the mapping from sort to field is not necessarily "one to one" and so it makes sense we must do something like this to "discover" which fields will be invoked by the sort?  It's still tricky/messy ... I wish we had a cleaner way to ask the `SortOrder` which indexed (doc values?) fields it depends on.

PR #1538, opened 2020-05-27
LUCENE-9368: Use readLELongs to read docIds on BKD leaf nodes

iverase:
This change changes the way we read docIds from the index using readLELongs so we can read most of the docs in one batch into a temporary array. This allows the compiler to run more efficient loops as well as we are copying information between two arrays.

In addition we add two need encoding paths for int8 and int16.


mikemccand:
This looks promising @iverase!  Do you have benchmark results for this approach?

iverase:
Yes, I have done some performance test. As expected the effect is bigger when the overhead of reading the documents is higher to the overhead of reading the points. So running it in the 1D case (~120 million points indexed as double):

|Approach|Index time Dev (sec)|Index time Base (sec)|diff|Force merge time Dev (sec)|Force merge time (sec) Base|diff|Index size Dev (GB)|Index size Base (GB)|diff|Reader heap Dev (MB)|Reader heap (MB) Base|diff|
|--------|:------|:------|:------|:------|:------|:------|:------|:------|:------|:------|:------|:------|
|104.3s|105.6s|-1%|0.0s|0.0s| 0%|0.95|0.95|-0%|0.07|0.07| 0%|

|Approach|M hits/sec  Dev |M hits/sec  Base|diff|QPS  Dev    |QPS  Base|diff |Hit count  Dev    |Hit count   Base |diff|
 |--------|:------|:------|:------|:------|:------|:------|:------|:------|:------|
1D| 114.58 |94.12| 21%|102.21|83.96| 21%|252237120|252237120| 0%|


Running it in the geo benchmarks, still it seems there a bit of a speed-up:

|Approach|Index time Dev (sec)|Index time Base (sec)|diff|Force merge time Dev (sec)|Force merge time (sec) Base|diff|Index size Dev (GB)|Index size Base (GB)|diff|Reader heap Dev (MB)|Reader heap (MB) Base|diff|
|--------|:------|:------|:------|:------|:------|:------|:------|:------|:------|:------|:------|:------|
|points|111.7s|110.0s| 1%|0.0s|0.0s| 0%|0.50|0.50|-0%|0.04|0.04| 0%|
|geo3d|140.0s|140.0s|-0%|0.0s|0.0s| 0%|0.71|0.71|-0%|0.04|0.04| 0%|
|shapes|178.5s|177.3s| 1%|0.0s|0.0s| 0%|1.26|1.26|-0%|0.04|0.04| 0%|


|Approach|Shape|M hits/sec  Dev |M hits/sec  Base|diff|QPS  Dev    |QPS  Base|diff |Hit count  Dev    |Hit count   Base |diff|
 |--------|:------|:------|:------|:------|:------|:------|:------|:------|:------|:------|
|points|polyRussia|17.59|17.35| 1%|5.01|4.95| 1%|3508846|3508846| 0%|
|points|polyMedium|9.67|9.47| 2%|118.43|116.02| 2%|2693559|2693559| 0%|
|points|poly 10|86.99|83.97| 4%|55.01|53.10| 4%|355809475|355809475| 0%|
|points|box|83.15|79.52| 5%|84.61|80.92| 5%|221118844|221118844| 0%|
|points|distance|83.22|79.45| 5%|48.90|46.68| 5%|382961957|382961957| 0%|
|points|nearest 10|0.03|0.03|-0%|2839.79|2847.82|-0%|60844404|60844404| 0%|
|points|sort|38.61|35.49| 9%|39.29|36.11| 9%|221118844|221118844| 0%|
|geo3d|polyRussia|0.67|0.66| 0%|0.19|0.19| 0%|3508671|3508671| 0%|
|geo3d|polyMedium|0.72|0.72| 1%|8.85|8.77| 1%|2693541|2693541| 0%|
|geo3d|poly 10|54.73|52.70| 4%|34.61|33.32| 4%|355855227|355855227| 0%|
|geo3d|box|60.85|56.60| 8%|61.92|57.60| 8%|221118844|221118844| 0%|
|geo3d|distance|76.52|72.77| 5%|44.91|42.71| 5%|383371904|383371904| 0%|
|shapes|polyRussia|10.28|9.94| 3%|2.93|2.83| 3%|3508846|3508846| 0%|
|shapes|polyMedium|3.37|3.43|-2%|41.34|42.04|-2%|2693559|2693559| 0%|
|shapes|poly 10|43.66|42.53| 3%|27.61|26.89| 3%|355809475|355809475| 0%|
|shapes|box|43.73|42.06| 4%|44.49|42.80| 4%|221118844|221118844| 0%|
|shapes|distance|45.79|44.72| 2%|26.90|26.27| 2%|382961957|382961957| 0%|

mikemccand:
Thanks @iverase!

lucene/core/src/java/org/apache/lucene/util/bkd/BKDReader.java
```diff
@@ -498,6 +500,7 @@ public IntersectState(IndexInput in, int numDims,
       this.scratchMinIndexPackedValue = new byte[packedIndexBytesLength];
       this.scratchMaxIndexPackedValue = new byte[packedIndexBytesLength];
       this.index = indexVisitor;
+      this.scratchLongs = new long[maxPointsInLeafNode / 2];
```
jpountz, line 503: does it need to be 
```suggestion
      this.scratchLongs = new long[(maxPointsInLeafNode + 1) / 2];
```
so that it keeps working for even numbers of points per node?

iverase: Not really. We read as many longs as we can so if the count is un-even for 32bits integer, the last four bytes are read from the index. See Line 120 of DocIdsWriter.

iverase:
Yes, that was my idea as a follow up and the reason I added the 8 /16 bits so we can use it later on instead of delta decoding. Adding a special case when docIds are just incrementing by one as well (we are doing that for postings) so base-time index or even counter can greatly benefit.

iverase:
>>> Should we increment the version number in BKDWriter to be able to remove the call to reverseBytes at search time?

We can do that or add different code numbers. We can leave the readers for backwards compatibility?

jpountz:
> Adding a special case when docIds are just incrementing by one as well

I can't think of many cases when that would kick in, this would only be used for single-valued fields that are used for index sorting?

> Yes, that was my idea as a follow up and the reason I added the 8 /16 bits so we can use it later

Do you mean something like stealing a bit of the byte we use to store the number of bits per value to record whether doc IDs are offseted by some constant? If so, then maybe we should do it now so 8/16 bits per value would have a chance of getting used on real data?

iverase:
I did something naive that is to add in an index documents with an increasing integer in a loop. Then you kick in that optimisation. I thought if you are adding documents which have a timestamp on them and you are adding them chronologically, they can benefit of such optimisation. 

The other idea is to remove `vInt` totally and record the lower value of the sorted array. Then you can just index the difference between the values of the array which gives more changes to have very low values.

jpountz:
I like that second idea better because I think it's more likely applicable. The issue with the first one is that it only works if you are indexing from a single thread, if you are indexing from multiple threads then the doc IDs will no longer be sorted after segments get merged.

iverase:
closing it as there has been a better proposal.

PR #1539, opened 2020-05-27
Fix typos in release wizard

madrob:


dev-tools/scripts/releaseWizard.yaml
```diff
@@ -1491,13 +1496,13 @@ groups:
...
-        logfile: add-bakccompat.log
+        logfile: add-backcompat.log
       - !Command
-        cmd: git diff
+        cmd: git diff --staged
         comment: Check the git diff before committing
         tee: true
       - !Command
-        cmd: git add -u .  && git commit -m "Add back-compat indices for {{ release_version }}"  && git push
+        cmd: git commit -m "Add back-compat indices for {{ release_version }}"  && git push
```
janhoy, line 1505: Do you remember why this change is needed?

dev-tools/scripts/releaseWizard.yaml
```diff
@@ -1306,6 +1306,11 @@ groups:
       Make sure you have a PGP enabled email client with your apache key installed.
       There are plugins for popular email programs, as well as browser plugins for webmail.
       See links for help on how to setup your email client for PGP.
+
+      If you prefer to sign the announcements manually rather than using a plugin,
```
janhoy, line 1310: Fair enough to offer that choice, although you'll bypass the auto signature validation that email clients provide...

dev-tools/scripts/releaseWizard.yaml
madrob: Because you already do `git add` on line 1498.

PR #1542, opened 2020-05-28
SOLR-14347: fix cached session update to not depend on Zookeeper state

murblanc:
# Description

Session used for replica placement computation was wrongly built to only account for state already in Zookeeper, therefore ignoring all in progress past Collection API operations. The Session returned to the cache was also the wrong one not including the changes to the cluster state resulting from the placement computation.

# Solution

The Session used for computation is now cloned from the cached session without trying to validate state in Zookeeper where that state might not yet be.

# Tests

Ran solr tests.

# Checklist


noblepaul:
@murblanc 
I've fixed a few things pls review & I can merge this

solr/solrj/src/java/org/apache/solr/client/solrj/cloud/autoscaling/Policy.java
```diff
@@ -644,15 +643,13 @@ public Session cloneToNewSession(SolrCloudManager cloudManager) {
       List<Clause> expandedClauses = new ArrayList<>(this.expandedClauses);
 
       List<Row> matrix = new ArrayList<>(nodes.size());
+      Map<String, Row> copyNodes = new HashMap<>();
+      for (Row oldRow: this.matrix) {
```
murblanc, line 647: Tradeoff with this one time matrix walk and copy is that we copy all nodes, even those that might not be needed (because not present in parameter `nodes`). I do like the approach though it's cleaner than the code I wrote.

Maybe keep this initial loop but just store Row references, then do the copy below when newRow is needed? (there's already an if bloc when it's null, so copy action can easily be placed in else bloc).

murblanc:
> @murblanc
> I've fixed a few things pls review & I can merge this

Added one comment. Looks good, thanks for picking this one up.

solr/solrj/src/java/org/apache/solr/client/solrj/cloud/autoscaling/Policy.java
noblepaul: >even those that might not be needed

I believe it's a reasonable tradeoff because most of the times nodes do not go away. Yes, we make an extra copy in those rare occasions, but it should be OK.

PR #1546, opened 2020-05-31
SOLR: Use absolute paths for server paths.

dsmiley:
Fixes relative path ambiguities when the current working directory is atypical.

I discovered this problem when running Solr standalone from IntelliJ using a CWD that was maybe slightly unexpected. Some relative paths like the core instance dir were getting resolved relative to the CWD instead of the solr home or core root dir.  So I think it's preferable to make the major directories of Solr be absolute to avoid this problem.  Not worth a CHANGES.txt entry, I think.

BTW I *thought* maybe this might be related to https://issues.apache.org/jira/browse/SOLR-7323 because I was trying to reproduce it but no it's not.

### DIFF
--- a/solr/core/src/java/org/apache/solr/core/CoreDescriptor.java
+++ b/solr/core/src/java/org/apache/solr/core/CoreDescriptor.java
@@ -182,7 +182,7 @@ public class CoreDescriptor {
    */
   public CoreDescriptor(String name, Path instanceDir, Map<String, String> coreProps,
                         Properties containerProperties, ZkController zkController) {
-    this.instanceDir = instanceDir;
+    this.instanceDir = instanceDir.toAbsolutePath();
 
     originalCoreProperties.setProperty(CORE_NAME, name);
 
@@ -290,9 +290,7 @@ public class CoreDescriptor {
     return defaultProperties.get(CORE_DATADIR).equals(coreProperties.getProperty(CORE_DATADIR));
   }
 
-  /**
-   * @return the core instance directory
-   */
+  /** The core instance directory (absolute). */
   public Path getInstanceDir() {
     return instanceDir;
   }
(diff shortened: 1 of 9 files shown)


solr/core/src/java/org/apache/solr/handler/admin/CoreAdminOperation.java
```diff
@@ -77,7 +77,7 @@
     String coreName = params.required().get(CoreAdminParams.NAME);
     Map<String, String> coreParams = buildCoreParams(params);
     CoreContainer coreContainer = it.handler.coreContainer;
-    Path instancePath = coreContainer.getCoreRootDirectory().resolve(coreName);
+    Path instancePath;
```
dsmiley, line 80: Just a little change to make the var effectively final, which is clearer.

solr/core/src/java/org/apache/solr/core/CoreDescriptor.java
```diff
@@ -182,7 +182,7 @@ public CoreDescriptor(String coreName, CoreDescriptor other) {
    */
   public CoreDescriptor(String name, Path instanceDir, Map<String, String> coreProps,
                         Properties containerProperties, ZkController zkController) {
-    this.instanceDir = instanceDir;
+    this.instanceDir = instanceDir.toAbsolutePath();
```
janhoy, line 185: A bit late, but I don't think this is necessary, as all callers will send absolute paths. And if you ever get a relative path, resolving it with `toAbsolutePath()` leads to it being relative to whatever CWD the app is started with, while the typical resolving of relative `instanceDir` is to resolve it relative to CoreContainer#coreRootDirectory.

dsmiley: That makes sense; thanks.  I see that all callers send absolute paths already.  So I think I should change this to throw an exception if it isn't absolute to protect us from mistakes.

I'll push a new commit here for review.  I've never done that for an already merged PR; we'll see how it goes.

PR #1547, opened 2020-05-31
SOLR-14525 For components loaded from packages SolrCoreAware, ResourceLoaderAware are not honored

noblepaul:
SOLR-14525
For components loaded from packages SolrCoreAware, ResourceLoaderAware are not honored

### DIFF
--- a/solr/core/src/java/org/apache/solr/core/SolrResourceLoader.java
+++ b/solr/core/src/java/org/apache/solr/core/SolrResourceLoader.java
@@ -538,7 +538,7 @@ public class SolrResourceLoader implements ResourceLoader, Closeable {
     Class<? extends T> clazz = findClass(cName, expectedType, subPackages);
     if (clazz == null) {
       throw new SolrException(SolrException.ErrorCode.SERVER_ERROR,
-          "Can not find class: " + cName + " in " + classLoader);
+              "Can not find class: " + cName + " in " + classLoader);
     }
 
     T obj = null;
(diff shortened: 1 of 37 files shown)


solr/core/src/java/org/apache/solr/pkg/PackageLoader.java
```diff
@@ -17,31 +17,25 @@
...
 import org.apache.solr.common.MapWriter;
+import org.apache.solr.common.SolrException;
 import org.apache.solr.common.cloud.ZkStateReader;
 import org.apache.solr.core.CoreContainer;
 import org.apache.solr.core.SolrCore;
 import org.apache.solr.core.SolrResourceLoader;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.io.Closeable;
```
dsmiley, line 30: Can you please configure your IDE to put java.* import statements up front instead of below?

solr/core/src/java/org/apache/solr/pkg/PackageLoader.java
```diff
@@ -301,6 +295,32 @@ public String toString() {
...
+  static class PackageResourceLoader extends SolrResourceLoader {
+
+    PackageResourceLoader(String name, List<Path> classpath, Path instanceDir, ClassLoader parent) {
+      super(name, classpath, instanceDir, parent);
+    }
+
+    @Override
+    public <T> boolean addToCoreAware(T obj) {
+      //do not do anything
+      //this class is
```
dsmiley, line 307: incomplete comment?  Could use an explanation as to why. SolrCoreAware isn't handled here.

solr/core/src/java/org/apache/solr/pkg/PackageLoader.java
```diff
@@ -301,6 +295,32 @@ public String toString() {
...
+      return false;
+    }
+
+    @Override
+    public <T> boolean addToResourceLoaderAware(T obj) {
+      if (obj instanceof ResourceLoaderAware) {
+        assertAwareCompatibility(ResourceLoaderAware.class, obj);
+        try {
+          ((ResourceLoaderAware) obj).inform(this);
```
dsmiley, line 316: Lets figure out why the base class doesn't just immediately call inform like this here.  What test then fails?  I've read an interesting JIRA issue: https://issues.apache.org/jira/browse/SOLR-8311  (lots of info there; I recommend reading all of it)  It's been on my mind for some time now actually since when I was cleaning up SolrResourceLoader since this "live" wart seems like a hack.

solr/core/src/java/org/apache/solr/pkg/PackagePluginHolder.java
```diff
@@ -117,6 +114,13 @@ protected void initNewInstance(PackageLoader.Package.Version newest) {
     Object instance = SolrCore.createInstance(pluginInfo.className,
         pluginMeta.clazz, pluginMeta.getCleanTag(), core, newest.getLoader());
     PluginBag.initInstance(instance, pluginInfo);
+    if (instance instanceof SolrCoreAware) {
```
dsmiley, line 117: Aha; makes sense.

solr/core/src/java/org/apache/solr/core/SolrResourceLoader.java
```diff
@@ -566,25 +566,42 @@ static void clearCache() {
...
+              "Error instantiating class: '" + clazz.getName() + "'", e);
     }
 
-    if (!live) {
-      if (obj instanceof SolrCoreAware) {
-        assertAwareCompatibility(SolrCoreAware.class, obj);
-        waitingForCore.add((SolrCoreAware) obj);
+    addToCoreAware(obj);
+    addToResourceLoaderAware(obj);
+    if(!live) {
```
dsmiley, line 574: Should this not be in its own method now too?

PR #1550, opened 2020-06-01
LUCENE-9383: benchmark module: Gradle conversion (complete)

dsmiley:
I switched from "java-library" type of Gradle plugin/module to more plainly "java" because this module isn't a library (isn't something depended on by anything), it's closer to an app.  I tried type "application" but I didn't have the same control that the "JavaExec" task gives you.  One consequence of not using "java-library" is that the names of the categories of dependencies are different, and so this appears odd/unusual relative to the other modules.

I did not convert "collation" and "shingle" Ant targets, but I put there the two-line CLI equivalents for both in the form of a comment.  I ram them and they worked... albeit a confusion in one of the perl scripts that thought "darwin" OS was ==~ Windows simply because it contained "win" :-). 

Notice the style of "getEnWiki" and "getGeoNames" and "getTop100kWikiWordFiles":  One task that does all it needs to do by adding a final step in doLast.  Now notice a different style: "reuters" (depending on extractReuters depending on getReuters).  This is more verbose, but admittedly for this case it has to do more.  I'm not well versed enough in Gradle to know which style is preferable.  I lean towards short & concise.  The current state is a nocommit IMO; need to harmonize the approaches.

I did not convert https://www-2.cs.cmu.edu/afs/cs.cmu.edu/project/theo-20/www/data/news20.tar.gz (aka news20) or https://people.csail.mit.edu/u/j/jrennie/public_html/20Newsgroups/20news-18828.tar.gz or https://kdd.ics.uci.edu/databases/20newsgroups/mini_newsgroups.tar.gz (aka mini-news) because I could not find .alg files that used them.

### DIFF
--- a/lucene/benchmark/build.gradle
+++ b/lucene/benchmark/build.gradle
@@ -21,43 +21,41 @@ apply plugin: 'java'
 description = 'System for benchmarking Lucene'
 
 dependencies {  
-  compile project(':lucene:core')
+  implementation project(':lucene:core')
 
-  compile project(':lucene:analysis:common')
-  compile project(':lucene:facet')
-  compile project(':lucene:highlighter')
-  compile project(':lucene:queries')
-  compile project(':lucene:spatial-extras')
-  compile project(':lucene:queryparser')
+  implementation project(':lucene:analysis:common')
+  implementation project(':lucene:facet')
+  implementation project(':lucene:highlighter')
+  implementation project(':lucene:queries')
+  implementation project(':lucene:spatial-extras')
+  implementation project(':lucene:queryparser')
 
-  compile "org.apache.commons:commons-compress"
-  compile "com.ibm.icu:icu4j"
-  compile "org.locationtech.spatial4j:spatial4j"
-  compile("net.sourceforge.nekohtml:nekohtml", {
+  implementation "org.apache.commons:commons-compress"
+  implementation "com.ibm.icu:icu4j"
+  implementation "org.locationtech.spatial4j:spatial4j"
+  implementation("net.sourceforge.nekohtml:nekohtml", {
     exclude module: "xml-apis"
   })
 
-  runtime project(':lucene:analysis:icu')
+  runtimeOnly project(':lucene:analysis:icu')
 
-  testCompile project(':lucene:test-framework')
+  testImplementation project(':lucene:test-framework')
 }
 
-ext {
-  tempDir = file("temp")
-  workDir = file("work")
-}
+def tempDir = file("temp")
+def workDir = file("work")
 
 task run(type: JavaExec) {
   description "Run a perf test (optional: -PtaskAlg=conf/your-algorithm-file -PmaxHeapSize=1G)"
   main 'org.apache.lucene.benchmark.byTask.Benchmark'
   classpath sourceSets.main.runtimeClasspath
   // allow these to be specified on the CLI via -PtaskAlg=  for example
-  def taskAlg = project.properties['taskAlg'] ?: 'conf/micro-standard.alg'
+  def taskAlg = propertyOrDefault('taskAlg', 'conf/micro-standard.alg')
   args = [taskAlg]
 
-  maxHeapSize = project.properties['maxHeapSize'] ?: '1G'
+  maxHeapSize = propertyOrDefault('maxHeapSize', '1G')
 
-  String stdOutStr = project.properties['standardOutput']
+  String stdOutStr = propertyOrDefault('standardOutput', null)
   if (stdOutStr != null) {
     standardOutput = new File(stdOutStr).newOutputStream()
   }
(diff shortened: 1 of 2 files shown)


lucene/benchmark/build.gradle
```diff
@@ -15,27 +15,138 @@
...
+}
+task reuters(type: JavaExec) {
+  dependsOn extractReuters
+  def input = extractReuters.outputs.files[0]
+  def output = "$workDir/reuters-out"
+  inputs.dir(input)
+  outputs.dir(output)
+  main = 'org.apache.lucene.benchmark.utils.ExtractReuters'
+  classpath = sourceSets.main.runtimeClasspath
+  jvmArgs = ['-Xmx1G']
```
madrob, line 131: Use `maxHeapSize`

dweiss:
> I switched from "java-library" type of Gradle plugin/module to more plainly "java" because this module isn't a library (isn't something depended on by anything), it's closer to an app. 

Fine. 

> I tried type "application" but I didn't have the same control that the "JavaExec" task gives you. One consequence of not using "java-library" is that the names of the categories of dependencies are different, and so this appears odd/unusual relative to the other modules.

This isn't accurate. The names of dependency configurations are different (and their setup is different). Gradle plugins (such as java, java-library, etc.) set up defaultsa for essentially the same underlying infrastructure. So you could just stay with 'java-library' and it'd still be fine.

lucene/benchmark/build.gradle
```diff
@@ -15,27 +15,138 @@
...
-  implementation project(':lucene:highlighter')
-  implementation project(':lucene:queries')
-  implementation project(':lucene:spatial-extras')
-  implementation project(':lucene:queryparser')
-
-  implementation "org.apache.commons:commons-compress"
-  implementation "com.ibm.icu:icu4j"
-  implementation "org.locationtech.spatial4j:spatial4j"
-  implementation("net.sourceforge.nekohtml:nekohtml", {
+  compile project(':lucene:core')
```
dweiss, line 24: This is wrong. You used a deprecated configuration name. Please take a look at this:

https://docs.gradle.org/current/userguide/java_plugin.html#sec:java_plugin_and_dependency_management

All these should be implementation dependencies I think.

lucene/benchmark/build.gradle
```diff
@@ -15,27 +15,138 @@
...
+  compile "com.ibm.icu:icu4j"
+  compile "org.locationtech.spatial4j:spatial4j"
+  compile("net.sourceforge.nekohtml:nekohtml", {
     exclude module: "xml-apis"
   })
 
-  testImplementation project(':lucene:test-framework')
+  runtime project(':lucene:analysis:icu')
+
+  testCompile project(':lucene:test-framework')
```
dweiss, line 42: Similar to above.

lucene/benchmark/build.gradle
```diff
@@ -15,27 +15,138 @@
...
+}
+
+ext {
```
dweiss, line 45: Only declare externalized properties if they really have to be externalized (read from somewhere outside the script). Here it's fine to just declare variables.

def tempDir = project.file("temp")

lucene/benchmark/build.gradle
```diff
@@ -15,27 +15,138 @@
...
+  tempDir = file("temp")
+  workDir = file("work")
+}
+
+task run(type: JavaExec) {
+  description "Run a perf test (optional: -PtaskAlg=conf/your-algorithm-file -PmaxHeapSize=1G)"
+  main 'org.apache.lucene.benchmark.byTask.Benchmark'
+  classpath sourceSets.main.runtimeClasspath
+  // allow these to be specified on the CLI via -PtaskAlg=  for example
+  def taskAlg = project.properties['taskAlg'] ?: 'conf/micro-standard.alg'
```
dweiss, line 55: Use global function propertyOrDefault which accepts more than just project.properties.

lucene/benchmark/build.gradle
```diff
@@ -15,27 +15,138 @@
...
+}
+
+task getReuters(type: Download) {
+  // note: there is no HTTPS url and we don't care because this is merely test/perf data
+  src "http://www.daviddlewis.com/resources/testcollections/reuters21578/reuters21578.tar.gz"
+  dest file("$tempDir/${src.file.split('/').last()}")
+  overwrite false
+  compress false
+}
+task extractReuters(type: Copy) {
```
dweiss, line 116: Add newline between tasks for clarity?

lucene/benchmark/build.gradle
```diff
@@ -15,27 +15,138 @@
...
+  dependsOn getReuters
+  from(tarTree(getReuters.dest)) { // can expand a .gz on the fly
+    exclude '*.txt'
+  }
+  into file("$workDir/reuters")
+}
+task reuters(type: JavaExec) {
+  dependsOn extractReuters
+  def input = extractReuters.outputs.files[0]
```
dweiss, line 125: A few problems here - this runs at configuration time so you can't take outputs of another task right away (yes, it'll work but it's not right). Besides, inputs and outputs don't need to be declared for these tasks at all so I'd just leave out inputs.dir and outputs.dir entirely. The task will just always execute.

Args should be moved to doFirst or use absolute location (not the depending task's resolved outputs).

lucene/benchmark/build.gradle
```diff
@@ -15,27 +15,138 @@
...
+}
+
+task getTop100kWikiWordFiles(type: Download) {
+  src "https://home.apache.org/~rmuir/wikipedia/top.100k.words.de.en.fr.uk.wikipedia.2009-11.tar.bz2"
+  dest file("$tempDir/${src.file.split('/').last()}")
+  overwrite false
+  compress false
+
+  doLast {
+    copy {
```
dweiss, line 147: sync rather than copy?

lucene/benchmark/build.gradle
dsmiley: Right; I was confused in my failed experiment with "application"

lucene/benchmark/build.gradle
```diff
@@ -15,13 +15,13 @@
  * limitations under the License.
  */
 
-
-apply plugin: 'java-library'
+apply plugin: 'java'
+// NOT a 'java-library'.  Maybe 'application' but seems too limiting.
```
dweiss, line 19: I think java plugin is more than fine here so remove the comment for the final version, maybe?

lucene/benchmark/build.gradle
```diff
@@ -37,5 +37,121 @@ dependencies {
...
+
+def tempDir = file("temp")
+def workDir = file("work")
+
+task run(type: JavaExec) {
+  description "Run a perf test (optional: -PtaskAlg=conf/your-algorithm-file -PmaxHeapSize=1G)"
+  main 'org.apache.lucene.benchmark.byTask.Benchmark'
+  classpath sourceSets.main.runtimeClasspath
+  // allow these to be specified on the CLI via -PtaskAlg=  for example
+  def taskAlg = propertyOrDefault('taskAlg', 'conf/micro-standard.alg')
```
dweiss, line 53: I'd just inline taskAlg into the array for brevity, but it's fine as is too.

lucene/benchmark/build.gradle
```diff
@@ -37,5 +37,121 @@ dependencies {
...
+  args = [taskAlg]
+
+  maxHeapSize = propertyOrDefault('maxHeapSize', '1G')
+
+  String stdOutStr = propertyOrDefault('standardOutput', null)
```
dweiss, line 58: Just had a random thought that if you don't redirect to a file the process is piped between gradle (parent) and this may cause artificial slowdowns on buffers between processes... Don't know if this matters but an alternative design could create a temporary file (task class has a method for creating task-relative temporary files), redirect the output into that file (always) and only pipe it to the console at the end if stdOutStr is not defined. 

I really don't know how these benchmarks are used in practice but I wanted to signal a potential issue here.

dsmiley: ehh; I'd prefer to keep this the way it is.  The code/scripts in the alg files generally don't print tons of output, so I don't think there's a perf interference concern.

lucene/benchmark/build.gradle
dsmiley: I like that this comment spells out a difference from how all the other modules are.

dweiss: From my (seasoned) gradle viewpoint this comment really doesn't make much sense: it's not an "application" in gradle sense - we launch multiple classes, have infrastructure in the build file, not a main class etc. But fine with me.

lucene/benchmark/build.gradle
dweiss: Sure.

dsmiley:
Merged: 89784ad7be45640ad835fe41981becfc91e62349

PR #1553, opened 2020-06-04
LUCENE-9393: FunctionScoreQuery turns TOP_DOCS to COMPLETE in inner weights

tflobbe:
FunctionScoreQuery can't really use WAND algorithm even if TOP_DOCS score mode is requested. This commit makes the inner weight created use COMPLETE in this case.

### DIFF
--- a/lucene/queries/src/java/org/apache/lucene/queries/function/FunctionScoreQuery.java
+++ b/lucene/queries/src/java/org/apache/lucene/queries/function/FunctionScoreQuery.java
@@ -105,7 +105,13 @@ public final class FunctionScoreQuery extends Query {
 
   @Override
   public Weight createWeight(IndexSearcher searcher, ScoreMode scoreMode, float boost) throws IOException {
-    Weight inner = in.createWeight(searcher, scoreMode.needsScores() && source.needsScores() ? scoreMode : ScoreMode.COMPLETE_NO_SCORES, 1f);
+    ScoreMode sm = scoreMode;
+    if (scoreMode.needsScores() && source.needsScores() && scoreMode == ScoreMode.TOP_SCORES) {
+      sm = ScoreMode.COMPLETE;
+    } else if (!scoreMode.needsScores() || !source.needsScores()) {
+      sm = ScoreMode.COMPLETE_NO_SCORES;
+    }
+    Weight inner = in.createWeight(searcher, sm, 1f);
     if (scoreMode.needsScores() == false)
       return inner;
     return new FunctionScoreWeight(this, inner, source.rewrite(searcher), boost);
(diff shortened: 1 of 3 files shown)


lucene/queries/src/test/org/apache/lucene/queries/function/TestFunctionScoreQuery.java
```diff
@@ -244,4 +247,33 @@ public void testAccessToValueSource() throws Exception {
 
   }
 
+  public void testScoreMode() throws Exception {
+    // Value Source doesn't need scores
+    assertInnerScoreMode(ScoreMode.COMPLETE_NO_SCORES, ScoreMode.COMPLETE, DoubleValuesSource.fromDoubleField("foo"));
```
tflobbe, line 252: While this test makes the current code pass, I'm trying to understand why it is correct. Why do we pass a COMPLETE_NO_SCORES to the underlying weight when the function's ValueSource doesn't need scores?

lucene/queries/src/java/org/apache/lucene/queries/function/FunctionScoreQuery.java
```diff
@@ -105,7 +105,13 @@ public static FunctionScoreQuery boostByQuery(Query in, Query boostMatch, float
 
   @Override
   public Weight createWeight(IndexSearcher searcher, ScoreMode scoreMode, float boost) throws IOException {
-    Weight inner = in.createWeight(searcher, scoreMode.needsScores() && source.needsScores() ? scoreMode : ScoreMode.COMPLETE_NO_SCORES, 1f);
+    ScoreMode sm = scoreMode;
+    if (scoreMode.needsScores() && source.needsScores() && scoreMode == ScoreMode.TOP_SCORES) {
+      sm = ScoreMode.COMPLETE;
+    } else if (!scoreMode.needsScores() || !source.needsScores()) {
+      sm = ScoreMode.COMPLETE_NO_SCORES;
+    }
```
jpountz, line 113: I think we can simplify the above to

```
if (scoreMode.needsScores() && source.needsScores()) {
  sm = ScoreMode.COMPLETE;
} else {
  sm = ScoreMode.COMPLETE_NO_SCORES;
}
```

lucene/queries/src/test/org/apache/lucene/queries/function/TestFunctionScoreQuery.java
jpountz: If the function's values source doesn't need scores, then Scorer#score() will never be called on the inner scorers, so we can treat them as filters, which is what `COMPLETE_NO_SCORES` does.

tflobbe: Ah, right. ValueSources like `MultiplicativeBoostValuesSource` will need scores because it uses the `DoubleValues scores` argument.

PR #1554, opened 2020-06-05
LUCENE-9352: Add iteratorCost function to Scorable

mayya-sharipova:


mayya-sharipova:
@jpountz I am wondering if we want to do anything else for this function?

PR #1556, opened 2020-06-08
SOLR-8392 type safety on SolrParam

madrob:


solr/core/src/java/org/apache/solr/core/RequestParams.java
```diff
@@ -85,32 +85,25 @@ public static ParamSet createParamSet(Map map, Long version) {
...
-      @SuppressWarnings({"rawtypes"})
-      Map.Entry entry = (Map.Entry) o1;
+  private static Map<String,String[]> getMapCopy(Map<String,?> value) {
+    Map<String, String[]> copy = new LinkedHashMap<>();
+    for (Map.Entry<String, ?> entry : value.entrySet()) {
       if ("".equals(entry.getKey())) {
-        copy.put(entry.getKey(), entry.getValue());
-        continue;
+        // nocommit
+        // copy.put(entry.getKey(), entry.getValue());
```
madrob, line 93: I don't understand why this case is special and gets treated differently rather than converting the value to a `String[]`

ErickErickson: This smells badly. If the key is "", then the raw value would be put in the return in the old code, even if null, whereas the changed code if the value was NOT null, it'd be the entry.getValue().toString() rather than entry.getValue().

That said, I think the change is better and any place in the that relies on the old behavior should be fixed.

PR #1558, opened 2020-06-09
LUCENE-9398: BKD tree is now read always off-heap

iverase:
This change makes the BKD index to be read always off-heap.


lucene/core/src/java/org/apache/lucene/util/bkd/BKDReader.java
```diff
@@ -871,7 +746,7 @@ private long estimatePointCount(IntersectState state, byte[] cellMinPacked, byte
 
   @Override
   public long ramBytesUsed() {
-    return packedIndex.ramBytesUsed();
+    return 0;
   }
```
jpountz, line 750: Let's remove Accountable from BKDReader instead?

jpountz:
I was thinking of a middle ground where PointsReader still implements Accountable (I'm planning to remove as part of [LUCENE-9387](https://issues.apache.org/jira/browse/LUCENE-9387)) but our implementations always return 0.

iverase:
I went a bit wild! I revert that commit and just remove accountable from the BKD reader.

PR #1559, opened 2020-06-09
SOLR-14552: Add BMW support to ReRank queries

tflobbe:


solr/core/src/test/org/apache/solr/search/TestReRankQParserPlugin.java
```diff
@@ -645,5 +650,95 @@ public void testReRankQueriesWithDefType() throws Exception {
...
+        "//result/doc[4]/str[@name='id'][.='2']",
+        "//result/doc[5]/str[@name='id'][.='3']",
+        "//result/doc[6]/str[@name='id'][.='4']",
+        "//result/doc[7]/str[@name='id'][.='5']",
+        "//result/doc[8]/str[@name='id'][.='6']",
+        "//result/doc[9]/str[@name='id'][.='7']",
+        "//result/doc[10]/str[@name='id'][.='8']"
+        );
+    
+    params.add(CommonParams.MIN_EXACT_COUNT, "2");
```
vthacker, line 709: So we are allowed to set `MIN_EXACT_COUNT` < `ReRankQParserPlugin.RERANK_DOCS` ?

In this case should we tell the re-rank collector to `ReRankQParserPlugin.RERANK_DOCS` ?

PR #1562, opened 2020-06-10
LUCENE-9400: Tessellator might fail when several holes share the same vertex

iverase:
Make sure we do not build illegal polygons when merging holes that share a single vertex.

see https://issues.apache.org/jira/browse/LUCENE-9400


iverase:
@nknize Can you have a look?

PR #1567, opened 2020-06-10
LUCENE-9402: Let MultiCollector handle minCompetitiveScore

tflobbe:
Here is an idea to make MultiCollector be able to handle `minCompetitiveScore`. Looking at this comment in the code:
```
            // Ignore calls to setMinCompetitiveScore so that if we wrap two
            // collectors and one of them wants to skip low-scoring hits, then
            // the other collector still sees all hits. We could try to reconcile
            // min scores and take the maximum min score across collectors, but
            // this is very unlikely to be helpful in practice.
```
This implementation tries to make it so that all collectors can see all the hits and only allow skipping if all collectors set a min competitive score. The value passed to the inner scorer is the minimum among all collectors (instead of the maximum as the comment suggests).

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/MultiCollector.java
+++ b/lucene/core/src/java/org/apache/lucene/search/MultiCollector.java
@@ -134,7 +134,7 @@ public class MultiCollector implements Collector {
       case 1:
         return leafCollectors.get(0);
       default:
-        return new MultiLeafCollector(leafCollectors, cacheScores);
+        return new MultiLeafCollector(leafCollectors, cacheScores, scoreMode() == ScoreMode.TOP_SCORES);
     }
   }
 
@@ -143,11 +143,15 @@ public class MultiCollector implements Collector {
     private final boolean cacheScores;
     private final LeafCollector[] collectors;
     private int numCollectors;
+    private final float[] minScores;
+    private final boolean skipNonCompetitiveScores;
 
-    private MultiLeafCollector(List<LeafCollector> collectors, boolean cacheScores) {
+    private MultiLeafCollector(List<LeafCollector> collectors, boolean cacheScores, boolean skipNonCompetitive) {
       this.collectors = collectors.toArray(new LeafCollector[collectors.size()]);
       this.cacheScores = cacheScores;
       this.numCollectors = this.collectors.length;
+      this.skipNonCompetitiveScores = skipNonCompetitive;
+      this.minScores = new float[this.skipNonCompetitiveScores ? this.numCollectors : 0];
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


lucene/core/src/java/org/apache/lucene/search/MultiCollector.java
```diff
@@ -143,32 +143,43 @@ public LeafCollector getLeafCollector(LeafReaderContext context) throws IOExcept
...
+    private final float[] minScores;
+    private final boolean skipNonCompetitiveScores;
 
-    private MultiLeafCollector(List<LeafCollector> collectors, boolean cacheScores) {
+    private MultiLeafCollector(List<LeafCollector> collectors, boolean cacheScores, boolean skipNonCompetitive) {
       this.collectors = collectors.toArray(new LeafCollector[collectors.size()]);
       this.cacheScores = cacheScores;
       this.numCollectors = this.collectors.length;
+      this.skipNonCompetitiveScores = skipNonCompetitive;
+      this.minScores = new float[this.skipNonCompetitiveScores ? this.numCollectors : 0];
```
jpountz, line 154: Let's make the array null when `skipNonCompetitiveScores` is false?
```suggestion
      this.minScores = this.skipNonCompetitiveScores ? new float[ this.numCollectors] : null;
```

lucene/core/src/test/org/apache/lucene/search/MultiCollectorTest.java
```diff
@@ -163,4 +163,64 @@ public void testCacheScoresIfNecessary() throws IOException {
     reader.close();
     dir.close();
   }
+  
+  public void testScorerWrappingForTopScores() throws IOException {
+    Directory dir = newDirectory();
+    RandomIndexWriter iw = new RandomIndexWriter(random(), dir);
+    iw.addDocument(new Document());
+    iw.commit();
```
jpountz, line 171: no need to commit, the follow-up call to getReader() creates a NRT segment anyway

tflobbe:
Ah! good Catch, I missed that completely. I'll fix.

lucene/core/src/java/org/apache/lucene/search/MultiCollector.java
```diff
@@ -134,69 +134,110 @@ public LeafCollector getLeafCollector(LeafReaderContext context) throws IOExcept
...
-        public void setMinCompetitiveScore(float minScore) {
-          // Ignore calls to setMinCompetitiveScore so that if we wrap two
-          // collectors and one of them wants to skip low-scoring hits, then
-          // the other collector still sees all hits. We could try to reconcile
-          // min scores and take the maximum min score across collectors, but
-          // this is very unlikely to be helpful in practice.
+      if (skipNonCompetitiveScores) {
+        for (int i = 0; i < collectors.length; ++i) {
+          final LeafCollector c = collectors[i];
+          assert c != null;
```
jpountz, line 163: I don't think that this assertion is right, the collector could be null if the collector already threw a CollectionTerminatedException? (we don't disallow calling `setCollector` after collection started)

tflobbe: hmm I had the null check before, but I thought `setScorer` had to only be called before the `collect` calls because of the javadoc: `Called before successive calls to {@link #collect(int)}.`. I'll put the null checks back.

lucene/core/src/test/org/apache/lucene/search/MultiCollectorTest.java
```diff
@@ -163,4 +163,115 @@ public void testCacheScoresIfNecessary() throws IOException {
...
+
+    Collector mc = MultiCollector.wrap(c1, c2);
+    LeafCollector lc = mc.getLeafCollector(ctx);
+    lc.setScorer(new ScoreAndDoc());
+    lc.collect(0); // OK
+    assertTrue("c1's collect should be called", c1.collectCalled);
+    assertTrue("c2's collect should be called", c2.collectCalled);
+    c1.collectCalled = false;
+    c2.collectCalled = false;
+    lc.collect(1); // OK, but c1 should terminate
```
jpountz, line 263: maybe create a new variant of this test that calls setScorer again after this collect call?

PR #1568, opened 2020-06-11
SOLR-14537 Improve performance of ExportWriter

sigram:
Details in Jira.

Initial changes here implement the "double buffering" approach to increase the throughput - an additional thread is created to fill in a buffer while the main thread writes out the documents from the other buffer.

Lucene TermEnum-s and DocValues are not multi-thread safe, that's why this change required the documents to be fully materialized in the buffer before handing it over to the other thread for writing. I think this is an acceptable tradeoff between reasonable amount of memory in the buffer and speed.


noblepaul:
Any benchmark numbers?

atris:
@sigram Whats the status of this PR? Is it completed?

PR #1569, opened 2020-06-11
LUCENE-9356: Add a test that verifies that Lucene catches bit flips.

jpountz:
Opening a reader and then calling checkIntegrity must throw a `CorruptIndexException` or an `IndexFormatToo(Old|New)Exception`.

### DIFF
--- a/lucene/core/src/test/org/apache/lucene/index/TestAllFilesDetectBitFlips.java
+++ b/lucene/core/src/test/org/apache/lucene/index/TestAllFilesDetectBitFlips.java
@@ -1,0 +1,139 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.lucene.index;
+
+
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.Collections;
+
+import org.apache.lucene.analysis.MockAnalyzer;
+import org.apache.lucene.codecs.CodecUtil;
+import org.apache.lucene.store.BaseDirectoryWrapper;
+import org.apache.lucene.store.Directory;
+import org.apache.lucene.store.IOContext;
+import org.apache.lucene.store.IndexInput;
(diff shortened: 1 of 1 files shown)


lucene/core/src/test/org/apache/lucene/index/TestAllFilesDetectBitFlips.java
```diff
@@ -0,0 +1,139 @@
...
+          try (IndexOutput out = dirCopy.createOutput(name, IOContext.DEFAULT);
+              IndexInput in = dir.openInput(name, IOContext.DEFAULT)) {
+              out.copyBytes(in, flipOffset);
+              out.writeByte((byte) (in.readByte() + TestUtil.nextInt(random(), 0x01, 0xFF)));
+              out.copyBytes(in, victimLength - flipOffset - 1);
+          }
+          try (IndexInput in = dirCopy.openInput(name, IOContext.DEFAULT)) {
+            try {
+              CodecUtil.checksumEntireFile(in);
+              System.out.println("TEST: changing a byte in " + victim + " did not update the checksum)");
```
msokolov, line 117: curious if you saw this much?

jpountz: I haven't seen a single occurrence of it (fortunately! :) )

jpountz:
> So we now check checksums on every file when opening

We only verify checksums of meta file when opening (those that we read entirely anyway). Checksums only get verified on other files when `LeafReader#checkIntegrity` is called.

lucene/core/src/test/org/apache/lucene/index/TestAllFilesDetectBitFlips.java
```diff
@@ -0,0 +1,139 @@
...
+    IndexWriterConfig conf = newIndexWriterConfig(new MockAnalyzer(random()));
+    conf.setCodec(TestUtil.getDefaultCodec());
+
+    if (cfs == false) {
+      conf.setUseCompoundFile(false);
+      conf.getMergePolicy().setNoCFSRatio(0.0);
+    }
+
+    RandomIndexWriter riw = new RandomIndexWriter(random(), dir, conf);
+    // Use LineFileDocs so we (hopefully) get most Lucene features
```
mikemccand, line 62: Woohoo!

lucene/core/src/test/org/apache/lucene/index/TestAllFilesDetectBitFlips.java
```diff
@@ -0,0 +1,139 @@
...
+    }
+    riw.close();
+    checkBitFlips(dir);
+    dir.close();
+  }
+  
+  private void checkBitFlips(Directory dir) throws IOException {
+    for(String name : dir.listAll()) {
+      if (name.equals(IndexWriter.WRITE_LOCK_NAME) == false) {
+        corruptFile(dir, name);
```
mikemccand, line 88: Do you maybe need to exclude `extraN` files for when `ExtraFS` randomly strikes?

Edit: oh, nevermind, I see we suppress `ExtraFS` above :)

lucene/core/src/test/org/apache/lucene/index/TestAllFilesDetectBitFlips.java
jpountz: This is actually copy-pasted from `TestAllFilesDetectTruncation` :)

PR #1571, opened 2020-06-11
SOLR-14560: Interleaving for Learning To Rank

alessandrobenedetti:
# Description

Interleaving is an approach to Online Search Quality evaluation that can be very useful for Learning To Rank models:
https://sease.io/2020/05/online-testing-for-learning-to-rank-interleaving.html
Scope of this issue is to introduce the ability to the LTR query parser of accepting multiple models (2 to start with).
If one model is passed, normal reranking happens.
If two models are passed, reranking happens for both models and the final reranked list is the interleaved sequence of results coming from the two models lists.
As a first step it is going to be implemented through:
TeamDraft Interleaving with two models in input.
In the future, we can expand the functionality adding the interleaving algorithm as a parameter.

# Solution

Change of core LTR classed and addition of a new rescorer

# Tests

Done

# Checklist

- [X ] I have reviewed the guidelines for [How to Contribute](https://wiki.apache.org/solr/HowToContribute) and my code conforms to the standards described there to the best of my ability.
- [X ] I have created a Jira issue and added the issue ID to my pull request title.
- [X ] I have given Solr maintainers [access](https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork) to contribute to my PR branch. (optional but recommended)


alessandrobenedetti:
The Pull Request is in a good status now, reviews can start.
I will keep adding some tests and responding to review feedbacks.

solr/contrib/ltr/src/java/org/apache/solr/ltr/search/LTRQParserPlugin.java
```diff
@@ -146,93 +149,114 @@ public LTRQParser(String qstr, SolrParams localParams, SolrParams params,
...
-
-      // Enable the feature vector caching if we are extracting features, and the features
-      // we requested are the same ones we are reranking with
-      if (featuresRequestedFromSameStore) {
-        scoringQuery.setFeatureLogger( SolrQueryRequestContextUtils.getFeatureLogger(req) );
+     
+      LTRScoringQuery[] rerankingQueries = new LTRScoringQuery[modelNames.length];
+      for (int i = 0; i < modelNames.length; i++) {
+        final LTRScoringQuery rerankingQuery;
+        if (!ORIGINAL_RANKING.equals(modelNames[i])) {
```
cpoerschke, line 161: The model name `originalRanking` is being given a special meaning here. I wonder if perhaps the differences between models could be transferred to the parameter names somehow (e.g. a new `original_model` parameter name alongside the existing `model` parameter name)? Then users could choose any model name they like, including for "original ranking" purposes.

https://github.com/apache/lucene-solr/pull/1705 proposes to factor out a `LTRQParserPlugin.newLTRScoringQuery` method (but I haven't yet explored fully w.r.t. how that might connect up here w.r.t. additional parameter names).

What do you think?

alessandrobenedetti: Hi @cpoerschke , I think you are right, it is probably better to have a separate param to activate the interleaving with the original ranking.
I have just made the change and updated the tests.
They are green locally, I'll just wait for them to be green remotely as well

alessandrobenedetti: Thinking a little bit more about it, even if we solve the "originalRanking" request through a special param, when returning the results, the [interleaving] transformer will need to present for the search results coming from the original ranking a value :

{
        "id":"GB18030TEST",
        "score":1.0005897,
        "[interleaving]":"OriginalRanking"},
      {
        "id":"UTF8TEST",
        "score":0.79656565,
        "[interleaving]":"myModel"}]
  }

This must follow the same format ( I wouldn't add a special field name different from [interleaing], this would complicate evaluators behaviours.

This means that in the end I guess we need to keep a special name to indicate the original ranking, and this name can't be used by the admin for any other model.
Given that, do you think is still beneficial to have a separate parameter?
Or just using the original approach, with the documented exclusivity of a special name, would look cleaner?

cpoerschke: Ah, good point about the special "OriginalRanking" also appearing in the "[interleaving]" transformer!

When using interleaving there's always at least two models to be interleaved, right?

The models could all be actual models or one of them could be the "OriginalRanking" pseudo-model.

I wonder if class inheritance might help us e.g.

```
class InterleavingLTRQParserPlugin extends LTRQParserPlugin
```

and (say)

```
<queryParser name="ltr" class="org.apache.solr.ltr.search.LTRQParserPlugin"/>
<queryParser name="iltr" class="org.apache.solr.ltr.search.InterleavingLTRQParserPlugin"/>
```

where

```
rq={!iltr model=myModelXYZ}
```

can be a convenience equivalent to

```
rq={!iltr model=OriginalRanking model=myModelXYZ}
```

i.e. if only one model is supplied then it is implied that the second model is the original ranking.

And if the special "OriginalRanking" name doesn't suit someone (either because they already have a real model that happens to be called "OriginalRanking" or because they would prefer a different descriptor in the "[interleaving]" transformer output) then something like

```
rq={!iltr originalRankingModel=noModel model=noModel model=someModel}
```

would allow them to call the "OriginalRanking" something else e.g. "noModel" instead.


We could even reject any "OriginalRanking" models that are actual models via something like

```
final String originalRankingModelName = localParams.get("originalRankingModel", "OriginalRanking" /* default */);
if (null != mr.getModel(originalRankingModelName)) {
  throw new SolrException(SolrException.ErrorCode.BAD_REQUEST,
                          "Found an actual '" + originalRankingModelName + "' model, please ...
}
```

----

However, the "[interleaving]" transformer still needs to know about the special name, hmm.


At the moment we have

```
public static boolean isOriginalRanking(LTRScoringQuery rerankingQuery){
  return rerankingQuery.getScoringModel() == null;
}
```

in LTRQParserPlugin and in LTRInterleavingTransformerFactory

```
if (isOriginalRanking(rerankingQuery)) {
  doc.addField(name, ORIGINAL_RANKING);
} else {
  doc.addField(name, rerankingQuery.getScoringModel().getName());
}
```

and if we gave LTRScoringQuery a getScoringModelName method

```
public String getScoringModelName() {
  return ltrScoringModel.getName();
}
```

then in LTRInterleavingTransformerFactory we could have

```
doc.addField(name, rerankingQuery.getScoringModelName());
```

if the

```
new LTRScoringQuery(null);
```

in LTRQParserPlugin becomes


```
new LTRScoringQuery(null) {
  @Override
  public String getScoringModelName() {
    return originalRankingModelName;
  }
};
```

but that wouldn't displace any other `isOriginalRanking(rerankingQuery)` calls, though if we had something like

```
final class OriginalRankingLTRScoringQuery extends LTRScoringQuery {
  private final String originalRankingModelName;
  public OriginalRankingLTRScoringQuery(String originalRankingModelName) {
    super(null /* LTRScoringModel */);
    this.originalRankingModelName = originalRankingModelName;
  }
  @Override
  public String getScoringModelName() {
    return this.originalRankingModelName;
  }
}
```

then `(rerankingQuery instanceof OriginalRankingLTRScoringQuery)` could replace `isOriginalRanking(rerankingQuery)` conceptually.

----

Just thinking out aloud here, haven't yet tried to turn any of the above into code.

Could you see inheritance (for LTRQParserPlugin and/or LTRScoringQuery) work potentially?

alessandrobenedetti: So @cpoerschke , I made a step back with the latest commit, let me explain,
First of all, the approach you detailed can definitely work,
I don't want to undervalue your effort and contribution, I am actually quite grateful for the time you dedicated in reviewing my proposal, so I really appreciate that.
But let's aim to a Keep It Simple Stupid approach here (KISS) and avoid over-complication in a first release, if possible:

1) Apache Solr makes use of "special names" in various places :
-  "_DEFAULT_" to indicate the default feature/model store
-  "_version" to indicate a meta-field used for versioning/optimistic concurrency
ect
I notice the prefix and suffix '_' is used for them.
So let's use it  : "_OriginalRanking_" could be a fair name to identify the Apache Solr ranking pre-re-ranking.
I named it after the "OriginalScore" feature, already in place, happy to use a different name, but wouldn't spend to much time on it.
My proposal would be to use "_OriginalRanking_" or potentially "_OriginalScoreRanking_" if more readable

2) the user interact with the !ltr query parser as usual, and if he/she wants to use interleaving, he/she just passes a second model (yes, we are supporting only 2 interleaved model right now):
model=A model=B
If he/she wants to interleave with the original ranking:
model=A model=_OriginalRanking_
I think this is quite intuitive and easy to be used, quite explicit and following current Apache Solr nomenclature for special named items

3) what happens if a user had/wants to use "_OriginalRanking_" for one of his/her models?
In this unlikely event, a Jira issue can be raised and we design a possible solution at that time.
A possible simple approach could be to make the original ranking name parametric, so the user can specify the name in the solrconfig.xml
But I would delay this reasoning only if we get requests for that functionality from users

What do you think?
If we find a good compromise here, we could move the review to the re-scoring part, that is more delicate and we can make sure no mistake happened there (I definitely don't want to slow down normal re-ranking and I definitely want the interleaved re-ranking to be as fast as possible)

You find the latest changes in the latest commits, I will update the related documentation as well today.

cpoerschke: Hi @alessandrobenedetti! Apologies for not returning to here sooner.

Good point about Apache Solr already using "special names" in various places, sure, let's go with "_OriginalRanking_" as the special name then. Hmm, the UI treats underscore special here, so `"_OriginalRanking_"` is what's in the code currently.

> ... we could move the review to the re-scoring part, that is more delicate ...

"delicate" is an excellent word choice, thank you, I spent some pleasant time today continuing to learn more about the re-scoring part. Took a sort of outside-to-inside approach i.e. starting at `LTRQParserPlugin` and then looking at what changed and how the bits fit together.

If it's okay with you I'll push a branch to my fork and add comments/questions sometimes with commit links here, not quite sure if that might work, let's try?

solr/contrib/ltr/src/java/org/apache/solr/ltr/search/LTRQParserPlugin.java
```diff
@@ -146,93 +149,114 @@ public LTRQParser(String qstr, SolrParams localParams, SolrParams params,
...
-      // External features
-      scoringQuery.setRequest(req);
-
-      return new LTRQuery(scoringQuery, reRankDocs);
     }
   }
+  
+  public static boolean isOriginalRanking(LTRScoringQuery rerankingQuery){
+    return rerankingQuery.getScoringModel() == null;
+  }
```
cpoerschke, line 208: 1/n Instead of having a static `isOriginalRanking(LTRScoringQuery rerankingQuery)` utility method here we could have a `OriginalRankingLTRScoringQuery extends LTRScoringQuery` class and replace `LTRQParserPlugin.isOriginalRanking(query)` calls with `query instanceof OriginalRankingLTRScoringQuery` expressions.

https://github.com/cpoerschke/lucene-solr/commit/6204ead58f1b6378f26d014dbba062423257f04d explores that.

In isolation the change appears to be somewhat stylistic only but it would later  (2/n and 4/n) allow for other possibilities too.

solr/contrib/ltr/src/java/org/apache/solr/ltr/response/transform/LTRInterleavingTransformerFactory.java
```diff
@@ -0,0 +1,120 @@
...
+import org.apache.solr.ltr.LTRScoringQuery;
+import org.apache.solr.ltr.SolrQueryRequestContextUtils;
+import org.apache.solr.request.SolrQueryRequest;
+import org.apache.solr.response.ResultContext;
+import org.apache.solr.response.transform.DocTransformer;
+import org.apache.solr.response.transform.TransformerFactory;
+import org.apache.solr.util.SolrPluginUtils;
+
+import static org.apache.solr.ltr.search.LTRQParserPlugin.ORIGINAL_RANKING;
+import static org.apache.solr.ltr.search.LTRQParserPlugin.isOriginalRanking;
```
cpoerschke, line 32: 2/n https://github.com/cpoerschke/lucene-solr/commit/be20b3a5a94fece4c8387d41831620a0f9497d39 explores making `LTRQParserPlugin.ORIGINAL_RANKING` private. A stylistic thing to some extent though it would in future also make it easier to pass in their own alternative to the special `"_OriginalRanking_"` value as a parameter.

solr/contrib/ltr/src/java/org/apache/solr/ltr/SolrQueryRequestContextUtils.java
```diff
@@ -47,12 +47,12 @@ public static FeatureLogger getFeatureLogger(SolrQueryRequest req) {
 
   /** scoring query accessors **/
 
-  public static void setScoringQuery(SolrQueryRequest req, LTRScoringQuery scoringQuery) {
-    req.getContext().put(SCORING_QUERY, scoringQuery);
+  public static void setScoringQuery(SolrQueryRequest req, LTRScoringQuery[] scoringQueries) {
```
cpoerschke, line 50: 3/n `getScoringQuery` already is renamed to `getScoringQueries` to reflect the changed signature, how about renaming `setScoringQuery` to `setScoringQueries` also?  https://github.com/cpoerschke/lucene-solr/commit/2579aee9f2e76321a08754d89630f635ae6899b7 does that plus `SCORING_QUERY` to `SCORING_QUERIES` too.

solr/contrib/ltr/src/java/org/apache/solr/ltr/LTRScoringQuery.java
```diff
@@ -73,6 +74,8 @@
   final private Map<String,String[]> efi;
   // Original solr query used to fetch matching documents
   private Query originalQuery;
+  // Model was picked for this Docs
+  private Set<Integer> pickedInterleavingDocIds;
```
cpoerschke, line 78: 4.1/n The addition of a new member here which is only sometimes applicable got me wondering about possibly having `LTRInterleavingScoringQuery extends LTRScoringQuery` inheritance.

solr/contrib/ltr/src/java/org/apache/solr/ltr/search/LTRQParserPlugin.java
```diff
@@ -146,93 +149,114 @@ public LTRQParser(String qstr, SolrParams localParams, SolrParams params,
...
-      super(defaultQuery, reRankDocs, new LTRRescorer(scoringQuery));
-      this.scoringQuery = scoringQuery;
+    public LTRQuery(LTRScoringQuery[] rerankingQueries, int rerankDocs) {
+      super(defaultQuery, rerankDocs, new LTRInterleavingRescorer(rerankingQueries));
+      this.rerankingQueries = rerankingQueries;
+    }
+
+    public LTRQuery(LTRScoringQuery rerankingQuery, int rerankDocs) {
+      super(defaultQuery, rerankDocs, new LTRRescorer(rerankingQuery));
+      this.rerankingQueries = new LTRScoringQuery[]{rerankingQuery};
```
cpoerschke, line 224: 4.2/n The differentation of behaviour here based on which constructor variant was called got me wondering about possibly having `LTRInterleavingQuery extends LTRQuery` inheritance.

https://github.com/cpoerschke/lucene-solr/commit/ff2cfe77538ee90cc7e10741ad06bf7347754d4a explores factoring out of `LTRInterleaving[Scoring]Query` classes.

solr/contrib/ltr/src/java/org/apache/solr/ltr/response/transform/LTRFeatureLoggerTransformerFactory.java
```diff
@@ -271,17 +287,23 @@ public void transform(SolrDocument doc, int docid)
...
-        fv = featureLogger.makeFeatureVector(
-            LTRRescorer.extractFeaturesInfo(
-                modelWeight,
-                docid,
-                (docsWereNotReranked ? score : null),
-                leafContexts));
+      LTRScoringQuery rerankingQuery = rerankingQueries[0];
+      LTRScoringQuery.ModelWeight rerankingModelWeight = modelWeights[0];
+      if (rerankingQueries.length > 1 && rerankingQueries[1].getPickedInterleavingDocIds().contains(docid)) {
+        rerankingQuery = rerankingQueries[1];
```
cpoerschke, line 293: 5/n When doing the 4/n stuff a question occurred to me here: `rerankingQueries[0]` and `modelWeights[0]` are the initial values. If the query is changed to `rerankingQueries[1]` then should the weight also be changed to `modelWeights[1]`? (Open question, haven't analysed it further yet nor tried it out.)

solr/contrib/ltr/src/java/org/apache/solr/ltr/search/LTRQParserPlugin.java
```diff
@@ -146,93 +149,114 @@ public LTRQParser(String qstr, SolrParams localParams, SolrParams params,
...
+            throw new SolrException(SolrException.ErrorCode.BAD_REQUEST,
+                "cannot find " + LTRQParserPlugin.MODEL + " " + modelNames[i]);
+          }
+          final String modelFeatureStoreName = ltrScoringModel.getFeatureStoreName();
+          final boolean extractFeatures = SolrQueryRequestContextUtils.isExtractingFeatures(req);
+          final String fvStoreName = SolrQueryRequestContextUtils.getFvStoreName(req);        // Check if features are requested and if the model feature store and feature-transform feature store are the same
+          final boolean featuresRequestedFromSameStore = (modelFeatureStoreName.equals(fvStoreName) || fvStoreName == null) ? extractFeatures : false;
+          if (threadManager != null) {
+            threadManager.setExecutor(req.getCore().getCoreContainer().getUpdateShardHandler().getUpdateExecutor());
+          }
```
cpoerschke, line 173: 6/n Also as part of 4/n noticed that here this if-then-set likely need not be inside the for-loop? The loop runs only once or twice so efficiency wise no concerns but from code reading perspective it not being inside the loop would be clearer.

cpoerschke (commented):
The commits mentioned in the comments are on https://github.com/cpoerschke/lucene-solr/commits/feature/SOLR-14560-cpoerschke-1 branch.

What do you think?

I really enjoyed looking at interleaving changes here today, thank you. :-)

solr/contrib/ltr/src/java/org/apache/solr/ltr/search/LTRQParserPlugin.java
alessandrobenedetti: Thanks @cpoerschke ! Tomorrow I'll address all your comments and think about them!
My intent is to merge within few days, so hopefully we can close this soon!
Thank you again for your time and cntribution, much appreciated!

solr/contrib/ltr/src/java/org/apache/solr/ltr/search/LTRQParserPlugin.java
```diff
@@ -146,93 +149,114 @@ public LTRQParser(String qstr, SolrParams localParams, SolrParams params,
     @Override
     public Query parse() throws SyntaxError {
```
cpoerschke, line 150: 7/n minor comments:
* technically not just `modelNames[0]` but all the model names could be `isEmpty` checked
* `threadManager.setExecutor` (already mentioned in 6/n) need not be inside the loop, likewise `extractFeatures` and `fvStoreName` and `extractEFIParams(localParams)` could move out.

solr/contrib/ltr/src/java/org/apache/solr/ltr/interleaving/TeamDraftInterleaving.java
```diff
@@ -0,0 +1,87 @@
...
+
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.LinkedHashSet;
+import java.util.Random;
+import java.util.Set;
+
+import org.apache.lucene.search.ScoreDoc;
+
+public class TeamDraftInterleaving implements Interleaving{
```
cpoerschke, line 28: 8/n suggestions:
* class level javadocs re: the interleaving algorithm
* comments or javadocs re: any assumptions e.g.
  * must rerankedA.length and rerankedB.length match?
  * can rerankedA and rerankedB contain the same docs?
  * can rerankedA contain the same doc more than once?
  * can rerankedB contain the same doc more than once?
* consider guarding against array-index-out-of-bounds exceptions (even if they shouldn't happen if all assumptions are met)

```
indexA = updateIndex(interleavedResults,indexA,rerankedA);
if (indexA < rerankedA.length) {
  interleavedResults.add(rerankedA[indexA]);
  teamA.add(rerankedA[indexA].doc);
  indexA++;
}
```

cpoerschke (commented):
I agree we should be able to wrap this up soon.

Once I'd grasped how the _Interleaving_ classes fit together with the existing _LTR_ classes I kept reading yesterday evening until I reached the end of the `solr/contrib/ltr/src/java` files. :-)

Will add more comments/questions today, in chunks, interleaved (no pun intended) with meetings and other work.

solr/contrib/ltr/src/java/org/apache/solr/ltr/LTRRescorer.java
```diff
@@ -166,64 +186,77 @@ public void scoreFeatures(IndexSearcher indexSearcher, TopDocs firstPassTopDocs,
...
-      scorer.docID();
-      scorer.iterator().advance(targetDoc);
-
-      scorer.getDocInfo().setOriginalDocScore(hit.score);
-      hit.score = scorer.score();
-      if (hitUpto < topN) {
-        reranked[hitUpto] = hit;
-        // if the heap is not full, maybe I want to log the features for this
-        // document
+      scoreSingleHit(indexSearcher, topN, modelWeight, docBase, hitUpto, hit, docID, scoringQuery, scorer, reranked);
```
cpoerschke, line 189: 9/n observation and suggestions:
* very elegant factoring out of methods for use by `LTRInterleavingRescorer`
* the `rerank` method already takes a `searcher` and so it could determine its own `leaves` from that
* `this.scoringQuery` being passed to the `scoreSingleHit` method as an argument (rather than it using the `this.scoringQuery` directly) is very subtle. it is of course present as an argument because `LTRInterleavingRescorer` will passing `rerankingQueries[i]` for that argument. the subtlety could be removed by making `scoreSingleHit` a static method.

https://github.com/cpoerschke/lucene-solr/commit/16512dbddbabe12ca5a2a26a9180ceeaae62cea2

solr/contrib/ltr/src/java/org/apache/solr/ltr/LTRInterleavingRescorer.java
```diff
@@ -0,0 +1,158 @@
...
+   *          documents to return;
+   */
+  @Override
+  public TopDocs rescore(IndexSearcher searcher, TopDocs firstPassTopDocs,
+      int topN) throws IOException {
+    if ((topN == 0) || (firstPassTopDocs.scoreDocs.length == 0)) {
+      return firstPassTopDocs;
+    }
+    
+    int originalRankingIndex = -1;
```
cpoerschke, line 68: 10/n suggestions:
* calculate originalRankingIndex in the constructor
* use originalRankingIndex to minimise "rerankingQueries[i] is not original ranking" checks
* use originalRankingIndex to remove the "original ranking query is always the last query" assumption
* the 'rerank' method already takes a `searcher` and so it could determine its own `leaves` from that

https://github.com/cpoerschke/lucene-solr/commit/002c31cceaeff411ad35111f00b7cee71c5b11de

solr/contrib/ltr/src/java/org/apache/solr/ltr/search/LTRQParserPlugin.java
alessandrobenedetti: makes sense to me! I resolve this and continue

solr/contrib/ltr/src/java/org/apache/solr/ltr/response/transform/LTRInterleavingTransformerFactory.java
alessandrobenedetti: Good idea!

solr/contrib/ltr/src/java/org/apache/solr/ltr/SolrQueryRequestContextUtils.java
alessandrobenedetti: absolutely reasonable!

solr/contrib/ltr/src/java/org/apache/solr/ltr/LTRInterleavingRescorer.java
```diff
@@ -0,0 +1,156 @@
...
+/**
+ * Implements the rescoring logic. The top documents returned by solr with their
+ * original scores, will be processed by a {@link LTRScoringQuery} that will assign a
+ * new score to each document. The top documents will be resorted based on the
+ * new score.
+ * */
+public class LTRInterleavingRescorer extends LTRRescorer {
+  
+  LTRScoringQuery[] rerankingQueries;
+  Interleaving interleavingAlgorithm = new TeamDraftInterleaving();
```
cpoerschke, line 43: 11.1/n If we go with 4/n then `LTRQParserPlugin.LTRQParser` could pass a `Interleaving interleavingAlgorithm` argument to the `LTRInterleavingQuery` constructor which could pass it to the `LTRInterleavingRescorer` constructor. For now `TeamDraftInterleaving` would be the only supported algorithm but in future other algorithms could then easily be added e.g. based on an additional `ltr` parameter. What do you think? An easy change to make now or something better left for later?

solr/contrib/ltr/src/java/org/apache/solr/ltr/interleaving/TeamDraftInterleaving.java
```diff
@@ -0,0 +1,87 @@
...
+  public static Random RANDOM;
+
+  static {
+    // We try to make things reproducible in the context of our tests by initializing the random instance
+    // based on the current seed
+    String seed = System.getProperty("tests.seed");
```
cpoerschke, line 34: 11.2/n `LTRQParserPlugin.LTRQParser` also has access to the `SolrQueryRequest` and its `SolrCore` object. For some reason I thought that within that some 'official' source of random-ness might be available which could be passed to a `TeamDraftInterleaving(Random)` constructor. And I imagined that our test harnesses would use seeds to make tests reproducible w.r.t. that 'official' source of random-ness. There however doesn't appear to be such a source of non-test official random-ness? `System.getProperty("tests.seed");` being used/available to non-test code seems potentially tricky.

@dweiss would you perhaps have any insights around non-test sources of randomness?

solr/contrib/ltr/src/java/org/apache/solr/ltr/response/transform/LTRFeatureLoggerTransformerFactory.java
```diff
@@ -210,50 +216,59 @@ public void setContext(ResultContext context) {
...
-        scoringQuery.setFeatureLogger( SolrQueryRequestContextUtils.getFeatureLogger(req) );
-      }
-      scoringQuery.setRequest(req);
-
-      featureLogger = scoringQuery.getFeatureLogger();
+      modelWeights = new LTRScoringQuery.ModelWeight[rerankingQueries.length];
+      String featureStoreName = SolrQueryRequestContextUtils.getFvStoreName(req);
+      for (int i = 0; i < rerankingQueries.length; i++) {
+        LTRScoringQuery scoringQuery = rerankingQueries[i];
+        if ((scoringQuery == null || !(scoringQuery instanceof OriginalRankingLTRScoringQuery)) && (docsWereNotReranked || (featureStoreName != null && !featureStoreName.equals(scoringQuery.getScoringModel().getFeatureStoreName())))) {
```
cpoerschke, line 229: 12/n observations/thoughts/questions:

Most tricky to articulate, hence left until last.

Prior to interleaving the existing logic is that if feature vectors are requested and there is no model (or the model is for a different feature store) then a logging model is created.

So now then if we have two models:
* if both models are for the requested feature store then that's great and each document would have been picked by one of the models and so we use the feature vector already previously calculated by whatever model had picked the document.
* if neither model is for the requested feature store then we need to create a logging model, is one logging model sufficient or do we need two? intuitively to me one would seem to be sufficient but that's based on partial analysis only so far.
* if one of the two models (modelA) is for the requested feature store then for the documents picked by modelA we can use the feature vector already previously calculated by modelA. what about documents picked by modelB? it could be that modelA actually has the feature vector for that document but that modelB simply managed to pick the document first. or if modelA does not have the feature vector then we could calculate it for modelA. would a logging model help in this scenario? intuitively to me it would seem that calculating the missing feature vector via modelA or via the logging model would both be equally efficient and hence no logging model would be needed but again that's only based on partial analysis so far.

solr/solr-ref-guide/src/learning-to-rank.adoc
```diff
@@ -247,6 +254,81 @@ The output XML will include feature values as a comma-separated list, resembling
...
+  }}
+----
+
+=== Running a Rerank Query Interleaving a model with the original ranking
+When approaching Search Quality Evaluation with interleaving it may be useful to compare a model with the original ranking. 
+To rerank the results of a query, interleaving a model with the original ranking, add the `rq` parameter to your search, with a model in input and activating the original ranking interleaving, for example:
+
+
+[source,text]
+http://localhost:8983/solr/techproducts/query?q=test&rq={!ltr model=myModel interleaveOriginalRanking=true reRankDocs=100}&fl=id,score
```
cpoerschke, line 299: 13/n minor: `interleaveOriginalRanking=true` --> `model=_OriginalRanking_` assuming we're going with that special value

cpoerschke (commented):
That concludes all the `solr/contrib/ltr/src/java` comments I had. Hope it helps, sorry it's a bit chunked up into individual items and/or commits on the branch.

Haven't yet fully looked at the `solr-ref-guide` changes and not yet at all at the tests though would love to also take a (more relaxed) look but that wouldn't be until next week, say tentatively Monday afternoon at the earliest if I may.

Thank you, this is really nice new Solr LTR functionality!

solr/contrib/ltr/src/java/org/apache/solr/ltr/LTRRescorer.java
alessandrobenedetti: Brilliant observation!

solr/contrib/ltr/src/java/org/apache/solr/ltr/search/LTRQParserPlugin.java
alessandrobenedetti: I have some doubts about this commit, it's not a strong opinion though, your considerations about implementing two subclasses are fair.
The additional property and method added to LTR Query is just: private Set<Integer> pickedInterleavingDocIds; and its getter/setters.

On the other hand with the creation of the subclass there is a little bit of code we are adding around,  I am not sure if it's worth in regarding to readability/maintenance.

I am just wondering if we get substantial benefits from this piece of refactor, if not we may stick with the original code,
what do you think?
Can you elaborate a bit more the expected benefits you were thinking?

solr/contrib/ltr/src/java/org/apache/solr/ltr/response/transform/LTRFeatureLoggerTransformerFactory.java
alessandrobenedetti: Yes, I agree, I'll change that and resolve this later

solr/contrib/ltr/src/java/org/apache/solr/ltr/LTRInterleavingRescorer.java
alessandrobenedetti: Perfectly splendid! I am just waiting to finalise the decision about the subclasses yes/no, and I'll proceed with this commit

solr/contrib/ltr/src/java/org/apache/solr/ltr/interleaving/TeamDraftInterleaving.java
alessandrobenedetti: I spent quite a lot of time on this, to make it compatible with the tests, if anyone else has some better solution I would be more than happy to change it, I was not super happy of what I ended with, but it was the only solution I found to be working at the time.

solr/contrib/ltr/src/java/org/apache/solr/ltr/search/LTRQParserPlugin.java
cpoerschke: That's a fair question.

Apart from the subclass considerations already annotated I see sort of "structural extensibility" type benefits, not sure if that's a good description, let me try to illustrate with two examples:

* example 1: beyond TeamDraftInterleaving: An interleaving algorithm other than `TeamDraftInterleaving` could be supported like I alluded to in 11.1/n though of course with the existing class structure that would also be possible.

```
- public LTRInterleavingQuery(LTRInterleavingScoringQuery[] rerankingQueries, int rerankDocs) {
-     super(null /* scoringQuery */, rerankDocs, new LTRInterleavingRescorer(rerankingQueries));
+ public LTRInterleavingQuery(LTRInterleavingScoringQuery[] rerankingQueries, int rerankDocs, Interleaving interleavingAlgorithm) {
+     super(null /* scoringQuery */, rerankDocs, new LTRInterleavingRescorer(rerankingQueries, interleavingAlgorithm));
```

vs.

```
- public LTRQuery(LTRScoringQuery[] rerankingQueries, int rerankDocs) {
-     super(defaultQuery, rerankDocs, new LTRInterleavingRescorer(rerankingQueries));
+ public LTRQuery(LTRScoringQuery[] rerankingQueries, int rerankDocs, Interleaving interleavingAlgorithm) {
+     super(defaultQuery, rerankDocs, new LTRInterleavingRescorer(rerankingQueries, interleavingAlgorithm));
      this.rerankingQueries = rerankingQueries;
  }

  public LTRQuery(LTRScoringQuery rerankingQuery, int rerankDocs) {
      super(defaultQuery, rerankDocs, new LTRRescorer(rerankingQuery));
      this.rerankingQueries = new LTRScoringQuery[]{rerankingQuery};
  }
```

You'd know better than me what other interleaving algorithms exist and might be interesting for LTR and what sort of parameters they might need from `LTRQParserPlugin`. One variant of the current `TeamDraftInterleaving` that I (naively) could see is for the two models to interleave not with the current 50:50 ratio but with (say) a 80:20 ratio.


* example 2: beyond Interleaving: we've got one model, we've got interleaving of two models, might there be other ways of combining models or of rescoring? I don't know what these might be: choose a document only if both models had picked it? choose a document only if at least 2 of 3 models picked it? hmm, maybe those are variants of interleaving. could one model rescore the documents already scored by another model?

anyhow, whatever the shiny new thing is, there'd be an existing `LTRQuery/LTRScoringQuery/LTRRescorer` + `LTRInterleavingQuery/LTRInterleavingScoringQuery/LTRInterleavingRescorer` pattern that might be repeated then as `LTRShinyNewQuery/LTRShinyNewScoringQuery/LTRShinyNewRescorer` and whilst `LTRInterleaving*` required only the `pickedInterleavingDocIds` as state within `LTR[Interleaving]ScoringQuery` perhaps `LTRShinyNew*` would require something more complex.

----

When you say _"there is a little bit of code we are adding around"_ I agree, yes, there's all the methods twice though slightly different then.
* Is your concern about the repetition as such and/or about the increase in the `LTRQParserPlugin` class in which the `LTR[Interleaving]Query` classes are inner classes?
* It appears that the inner classes are not _but could be_ static inner classes and therefore it should also be possible to promote them to not be inner classes of `LTRQParserPlugin` but classes in their own right and that would make `LTRQParserPlugin` shorter.

solr/contrib/ltr/src/java/org/apache/solr/ltr/search/LTRQParserPlugin.java
alessandrobenedetti: as you will see in my next commit, I added the empty check in the loop and a couple of tests for it, I leave this open, feel free to resolve this once you take a look.

If we don't check for empty, we will get an exception anyway saying:
"cannot find model"
It's up to us if we want to help more the user with the empty message.
I think is fine, telling explicitly in an exception the model is empty could save time for developers/integrators not understanding why a model cannot be found

solr/contrib/ltr/src/java/org/apache/solr/ltr/interleaving/TeamDraftInterleaving.java
cpoerschke: Yes, predictable randomness in tests can be tricky! I had an (unplanned) idea around this at breakfast this morning, trying it out now: https://github.com/cpoerschke/lucene-solr/commit/5fc99898f9bc82f4dabcfe188e132ba2bc727704 is part 1 of 2 (and is independent technically speaking).

solr/contrib/ltr/src/java/org/apache/solr/ltr/interleaving/TeamDraftInterleaving.java
alessandrobenedetti: update the java docs, in regards to the index out of bounds, I think we should raise the exception instead of adding the condition:
if the assumptions are not met, there is a bug to be fixed, so we need interleaving to fail instead of silently operates wrong, am I missing anything?
Please take a look to my new commit, especially for the comments

solr/contrib/ltr/src/java/org/apache/solr/ltr/search/LTRQParserPlugin.java
alessandrobenedetti: I didn't like much the bit in the  LTRFeatureLoggerTransformerFactory.java
LTRInterleavingTransformerFactory.java, but thinking more about it, I don't think is that much of a problem, ad your observations are fair. I am finishing the other comments and I may extract  the two classes

solr/contrib/ltr/src/java/org/apache/solr/ltr/LTRInterleavingRescorer.java
alessandrobenedetti: I added this bit, I think is easy enough and not much invasive to prepare the structure now.

Adding new algorithm would be as easy as :
- fetching the param from the request
- add the algorithm name in the switch clause
- implement the new algorithm

Take a look if you like it

solr/contrib/ltr/src/java/org/apache/solr/ltr/interleaving/TeamDraftInterleaving.java
cpoerschke: https://github.com/cpoerschke/lucene-solr/commit/63a190b57fdc892b98c546f005edbc096e4e4095 is part 2 of 2: it removes the tests.seed System property use in TeamDraftInterleaving by passing a Random object into TeamDraftInterleaving. instead the tests.seed System property use is in a test class (LTRQParserTestPlugin) and tests continue to make their `setRANDOM` calls but instead of `TeamDraftInterleaving.setRANDOM` it's now `LTRQParserTestPlugin.setRANDOM` then.

Two problems with this though:
* a forbidden apis check fails (saying that RandomizedRunner's random() should be used instead but i'm unclear still on how that might be done)
* the tests no longer pass, which baffles me though haven't looked at details yet; speculations:
  * was the randomness in the tests predictable before and now it's predictable also but the number sequence has changed and test expectations need to be adjusted to match?
  * something to do with order of system property setting and non-test/test class loading perhaps and perhaps use of the RandomizedRunner's random would solve it somehow?

solr/contrib/ltr/src/java/org/apache/solr/ltr/search/LTRQParserPlugin.java
alessandrobenedetti: So I have done that in my latest commit, I brought back the original implementation for the transformers, it may be less readable but are we reducing the number of contains call?
Given that we support just interleaving for 2 ranked list, if a document is not assigned to modelB is by default assigned to modelA and no need of checking the  getPickedInterleavingDocIds?

solr/contrib/ltr/src/java/org/apache/solr/ltr/interleaving/TeamDraftInterleaving.java
alessandrobenedetti: Now I remember, I worked on this a lot and I ended up with the current solution even if including the test seed get property.
I have not introduced that approach on my own, but I took inspiration from other areas of Solr code:
such as:
org/apache/solr/cloud/api/collections/OverseerCollectionMessageHandler.java:174
org/apache/solr/core/BlobRepository.java:73
org/apache/solr/servlet/HttpSolrCall.java:152
org/apache/solr/util/tracing/GlobalTracer.java:35

So it is already used quite widely across Solr production code.
Unless you have stronger concerns we can probably leave it as it is and resolve the conversation here :)

solr/contrib/ltr/src/java/org/apache/solr/ltr/response/transform/LTRFeatureLoggerTransformerFactory.java
alessandrobenedetti: Spent a lot of time on this, I will continue on Monday, some considerations so far:
- in interleaving the logging model scoring query keeps also the interleaving picks (currently as a subclass of the scoring query), so we would need 2 logging models, by the way, I had various problems due to class cast exceptions and array around, so I will think a bit more about the subclass implementation to understand if it's actually an headache or full benefit, 
- I think the solution will be to implement only one model, 2 different scoringQuery (that contain the picks) and change the feature vector cache to key on: private static int fvCacheKey(LTRScoringQuery scoringQuery, int docid) {
    return  scoringQuery.getScoringModel().hashCode() + (31 * docid);
  }
I keep you posted on monday

solr/contrib/ltr/src/java/org/apache/solr/ltr/interleaving/TeamDraftInterleaving.java
cpoerschke: Ah, precedent and existing use already, good to know, thanks for sharing!

solr/solr-ref-guide/src/learning-to-rank.adoc
```diff
@@ -247,6 +254,81 @@ The output XML will include feature values as a comma-separated list, resembling
...
+      {
+        "id":"UTF8TEST",
+        "score":0.79656565,
+        "[interleaving]":"myModelA"}]
+  }}
+----
+
+=== Running a Rerank Query Interleaving a model with the original ranking
+When approaching Search Quality Evaluation with interleaving it may be useful to compare a model with the original ranking. 
+To rerank the results of a query, interleaving a model with the original ranking, add the `rq` parameter to your search, with a model in input and activating the original ranking interleaving, for example:
```
cpoerschke, line 295:
```suggestion
To rerank the results of a query, interleaving a model with the original ranking, add the `rq` parameter to your search, passing the special inbuilt `_OriginalRanking_` model identifier as one model and your comparison model as the other model, for example:
```

solr/solr-ref-guide/src/learning-to-rank.adoc
```diff
@@ -247,6 +254,81 @@ The output XML will include feature values as a comma-separated list, resembling
...
+
+
+[source,text]
+http://localhost:8983/solr/techproducts/query?q=test&rq={!ltr model=myModel model=_OriginalRanking_ reRankDocs=100}&fl=id,score
```
cpoerschke, line 299: subjective: might `model=_OriginalRanking_ model=myModel` be more intuitive i.e. the 'from' baseline model on the left and the 'to' alternative model on the right? (i recall that the code had an "original ranking last" assumption before but if that's gone there's a possibility here to swap the order)

solr/solr-ref-guide/src/learning-to-rank.adoc
```diff
@@ -247,6 +254,81 @@ The output XML will include feature values as a comma-separated list, resembling
...
 ----
 
+=== Running a Rerank Query Interleaving Two Models
+
+To rerank the results of a query, interleaving two models (myModelA, myModelB) add the `rq` parameter to your search, passing two models in input, for example:
+
+[source,text]
+http://localhost:8983/solr/techproducts/query?q=test&rq={!ltr model=myModelA model=myModelB reRankDocs=100}&fl=id,score
+
+To obtain the model that interleaving picked for a search result, computed during reranking, add `[interleaving]` to the `fl` parameter, for example:
```
cpoerschke, line 264: question: if myModelA had `[ doc1, doc2, doc3 ]` document order and myModelB had `[ doc1, doc3, doc2 ]` document order i.e. there was agreement between the models re: the first document, will `[interleaving]` return (1) randomly `myModelA` or `myModelB` depending on how the picking actually happened or will it return (2) something else e.g. `myModelA,myModelB` (if myModelA actually picked and myModelB agreed) or `myModelB,myModelA` (if myModelB actually picked and myModelA agreed) or will it return (3) neither since in a way neither of them picked the document since they both agreed on it?

answer-ish: from recalling the implementation the answer is (1) i think though from a user's perspective perhaps it might be nice here to clarify here somehow around that? a subtle aspect being (if i understand things right) that `[features]` and `[interleaving]` could both be requested in the `fl` and whilst myModelA and myModelB might have agreed that `doc1` should be the first document they might have used very different features to arrived at that conclusion and their `score` value could also differ.

solr/solr-ref-guide/src/learning-to-rank.adoc
```diff
@@ -418,6 +500,14 @@ Learning-To-Rank is a contrib module and therefore its plugins must be configure
 </transformer>
 ----
 
+* Declaration of the `[interleaving]` transformer.
++
+[source,xml]
+----
+<transformer name="interleaving" class="org.apache.solr.ltr.response.transform.LTRInterleavingTransformerFactory">
+</transformer>
```
cpoerschke, line 508: minor/subjective: could shorten since there's no parameters

```
<transformer name="interleaving" class="org.apache.solr.ltr.response.transform.LTRInterleavingTransformerFactory"/>
```

solr/solr-ref-guide/src/learning-to-rank.adoc
```diff
@@ -779,3 +869,7 @@ The feature store and the model store are both <<managed-resources.adoc#managed-
 * "Learning to Rank in Solr" presentation at Lucene/Solr Revolution 2015 in Austin:
 ** Slides: http://www.slideshare.net/lucidworks/learning-to-rank-in-solr-presented-by-michael-nilsson-diego-ceccarelli-bloomberg-lp
 ** Video: https://www.youtube.com/watch?v=M7BKwJoh96s
+
```
cpoerschke, line 872: We've got _"... Contributions for further models, features and normalizers are welcome. ..."_ above, any thoughts on adding "interleaving algorithms" to that list?

solr/contrib/ltr/src/java/org/apache/solr/ltr/interleaving/TeamDraftInterleaving.java
```diff
@@ -0,0 +1,121 @@
...
+    interleavedResultArray = interleavedResults.toArray(interleavedResultArray);
+
+    return new InterleavingResult(interleavedResultArray,interleavingPicks);
+  }
+
+  private int updateIndex(LinkedHashSet<ScoreDoc> interleaved, int index, ScoreDoc[] reranked) {
+    boolean foundElementToAdd = false;
+    while (index < reranked.length && !foundElementToAdd) {
+      ScoreDoc elementToCheck = reranked[index];
+      if (interleaved.contains(elementToCheck)) {
```
cpoerschke, line 109: Wondering re: `interleaved.contains(elementToCheck)` vs. `interleaved.contains(elementToCheck).doc` here. I note that the new `ScoreDoc.equals` method considers `doc` and `shardIndex` but not `score` which makes sense in the context here but for `ScoreDoc` alone is perhaps less obvious. The `interleavingPicks` use `doc` only (i.e. not `doc+shardIndex`). I haven't fully thought it through but intuitively if `doc+shardIndex` is required here then would it not also be required for the interleaving picks to avoid a strange edge case when potentially modelA picks "doc 1 from shard 1" and modelB picks "doc 1 from shard 2"?

solr/contrib/ltr/src/java/org/apache/solr/ltr/interleaving/TeamDraftInterleaving.java
cpoerschke: Very nice java docs and comments, thanks!

> ... I think we should raise the exception ...

Yes, you're right, an exception that will lead to a bug being identified and fixed rather than silent failure or (confusing for the user) "array out of bounds exception" would be better.

solr/contrib/ltr/src/java/org/apache/solr/ltr/response/transform/LTRFeatureLoggerTransformerFactory.java
alessandrobenedetti: So I just committed my changes on this, I spent quite a while thinking about the various scenarios, adjusting the code and adding the related tests.

From your observations:
- if both models are for the requested feature store then that's great and each document would have been picked by one of the models and so we use the feature vector already previously calculated by whatever model had picked the document. [OK]
- if neither model is for the requested feature store then we need to create a logging model, is one logging model sufficient or do we need two? intuitively to me one would seem to be sufficient but that's based on partial analysis only so far.
[One is sufficient, and my latest changes do that, anyway I just realized that the loggingModel is not heavy to create and when getting the featureVector, the cache is accessed, the key for that cache doesn't look to the instance but to the content of classes, so two identical logging models would have matched in the feature vector cache, the change was probably not vital, but not harmful either]
- in the third scenario we still need the logging model, because when specifying a featureStore in the featureVector transformer we aim to extract all the features of that store, from efi passed to the transformer, so when explicitly mentioned, we need a logger for both the model again (also for the one that aligns)

alessandrobenedetti: Actually taking a deeper look to the third point, the original implementation was not extracting all the features, but if the explicit featureStore was matching the model featureStore, it was using the model one (no logger).

So I agree with you, in our case, we want to use the model already existent and no logger at all.
I am going to clean up that bit and do a new commit tomorrow

alessandrobenedetti: Ok, I have done an extensive refactor of this bit, following your scenarios guideline, I believe the code is much more readable now.
I added few tests as well.
Thank you very much for the insight, now that part is extremely clear.

Before resolving  this discussion, working on this, another consideration sparkled:

Currently when we use the feature logger transformer, all features are extracted (even the ones not used by the reranking model, if any).
Are we sure we want this behavior ?
We could re-use the extracted feature vector  cached also for logging if we just log the features actually used by the model.
I am just wondering why I would be interested in logging for a document, all the features in a featureStore, including potentially features just used by other models.
This could actually lead to confusion.
If you agree I create a separate Jira for that and I'll implement a solution soon, to avoid to forget and context switch.

solr/solr-ref-guide/src/learning-to-rank.adoc
alessandrobenedetti: Just fixed that, it's not critical as the order is not important, but if it's more readable, let's go for it!

solr/solr-ref-guide/src/learning-to-rank.adoc
alessandrobenedetti: 1 is what is currently implemented and it aligns with the TeamDraft Interleaving papers and evaluation methods.
Your observation is interesting though, but to implement that we should invent a new type of Interleaving algorithm that will do that when interleaving the results and will evaluate the user clicks accordingly later on.

Your observation on the features to log applies as well.
So far  no change is needed in this regard in my opinion.

solr/contrib/ltr/src/java/org/apache/solr/ltr/interleaving/TeamDraftInterleaving.java
alessandrobenedetti: You are right, ScoreDoc.equals alone may be less obvious.
The reason I originally did that was to support sharding, but to be honest, currently Interleaving doesn't support sharding anyway (and adding the support is an extremely difficult topic that would require some serious thinking).

So I rolled back the ScoreDoc, and added an additional data structure on the doc Id, to do the contains.

Let me know if is better!

alessandrobenedetti:
@cpoerschke  I think I finished implementing all the required changes and discussions, let me know with no rush and I think we can move this to the commit phase :)

solr/contrib/ltr/src/java/org/apache/solr/ltr/response/transform/LTRFeatureLoggerTransformerFactory.java
```diff
@@ -208,55 +216,116 @@ public void setContext(ResultContext context) {
       if (threadManager != null) {
         threadManager.setExecutor(context.getRequest().getCore().getCoreContainer().getUpdateShardHandler().getUpdateExecutor());
       }
```
cpoerschke, line 218: LTRFeatureLoggerTransformerFactory.3 - I noted that `threadManager` here is an existing member of the transformer factory and it is initialised as part of request processing. Since there's no locking or anything there could be a chance that multiple threads concurrently call `threadManager.setExecutor()` but the argument to the set call is not specific to the request i.e. all requests would set the same thing (whereas for the logging model different requests could supply a different feature store name via the `fl=[feature store=...]` parameter).

solr/contrib/ltr/src/java/org/apache/solr/ltr/response/transform/LTRFeatureLoggerTransformerFactory.java
```diff
@@ -208,55 +216,116 @@ public void setContext(ResultContext context) {
...
+     * the feature vector can be fetched from the feature vector cache.
+     * 2) If an explicit feature store is passed, and no reranking query uses a model from that featureStore,
+     * There is the need of a logger model to extract the feature vector
+     * 3) If an explicit feature store is passed, and there is a reranking query that uses a model from that featureStore,
+     * It can be re-used
+     * 
+     * @param rerankingQueriesFromContext reranking queries
+     * @param transformerFeatureStore explicit feature store for the transformer
+     * @param transformerExternalFeatureInfo explicit efi for the transformer
+     */
```
cpoerschke, line 270: LTRFeatureLoggerTransformerFactory.4 - Very nice overview of the different code paths here. https://github.com/cpoerschke/lucene-solr/commit/3a61287a0e4fb5a77f080a92e7129582b234cbd7 suggests some tweaks to the implementation, I will annotate comments re: my thinking behind those suggestions. Let me know what you think?

solr/contrib/ltr/src/java/org/apache/solr/ltr/response/transform/LTRFeatureLoggerTransformerFactory.java
```diff
@@ -208,55 +216,116 @@ public void setContext(ResultContext context) {
...
+            threadManager);
+        rerankingQueries = new LTRScoringQuery[]{loggingQuery};
+      } else {
+        rerankingQueries = new LTRScoringQuery[rerankingQueriesFromContext.length];
+        System.arraycopy(rerankingQueriesFromContext, 0, rerankingQueries, 0, rerankingQueriesFromContext.length);
+
+        if (transformerFeatureStore != null) {// explicit feature store for the transformer
+          LTRScoringModel matchingRerankingModel = null;
+          for (LTRScoringQuery rerankingQuery : rerankingQueries) {
+            if (isModelMatchingFeatureStore(transformerFeatureStore, rerankingQuery.getScoringModel())) {
```
cpoerschke, line 285: LTRFeatureLoggerTransformerFactory.5 - the `rerankingQuery` here could be an `OriginalRankingLTRScoringQuery` object which inherently has no feature store. The implementation here works because it relies on `rerankingQuery.getScoringModel()` returning `null` for `OriginalRankingLTRScoringQuery` objects and because `isModelMatchingFeatureStore` checks model null-ness. This is very subtle and it could be made more explicit for code clarity.

solr/contrib/ltr/src/java/org/apache/solr/ltr/response/transform/LTRFeatureLoggerTransformerFactory.java
```diff
@@ -208,55 +216,116 @@ public void setContext(ResultContext context) {
...
+          if (matchingRerankingModel != null) {//one of the LTR query model can be re-used
+            for (LTRScoringQuery rerankingQuery : rerankingQueries) {
+              rerankingQuery.setLtrScoringModel(matchingRerankingModel);
+            }
+          } else {
+            for (LTRScoringQuery rerankingQuery : rerankingQueries) {
+              rerankingQuery.setLtrScoringModel(loggingModel);
+              rerankingQuery.setExtractAllFeatures(true);
+              rerankingQuery.setEfi(transformerExternalFeatureInfo);
+            }
```
cpoerschke, line 298: LTRFeatureLoggerTransformerFactory.6 - the reranking queries from the context were copied and now they are being modified. The "why copy-and-modify instead construct-a-new-one?" question may arise for code readers.

solr/contrib/ltr/src/java/org/apache/solr/ltr/response/transform/LTRFeatureLoggerTransformerFactory.java
```diff
@@ -271,17 +340,24 @@ public void transform(SolrDocument doc, int docid)
...
-      if (fv == null) { // FV for this document was not in the cache
-        fv = featureLogger.makeFeatureVector(
-            LTRRescorer.extractFeaturesInfo(
-                modelWeight,
-                docid,
-                (docsWereNotReranked ? score : null),
-                leafContexts));
+      LTRScoringQuery rerankingQuery = rerankingQueries[0];
+      LTRScoringQuery.ModelWeight rerankingModelWeight = modelWeights[0];
+      if (rerankingQueries.length > 1 && ((LTRInterleavingScoringQuery)rerankingQueries[1]).getPickedInterleavingDocIds().contains(docid)) {
```
cpoerschke, line 345: LTRFeatureLoggerTransformerFactory.7 - here now the reranking queries that were copied from the context (and modified) are being used. It's very subtle but questions that may arise for code readers are (a) whether or not the copy happened before or after the picked interleaving doc ids were set, and/or (b) if the copy is shallow so that the query from the context and the query in the transformer share the same underlying picked ids object, and/or (c) if any shared underlying picked ids object remains shared when 'set' is called on it. https://github.com/cpoerschke/lucene-solr/commit/3a61287a0e4fb5a77f080a92e7129582b234cbd7 proposes to introduce (alongside the `rerankingQueries` array) a `rerankingQueriesFromContext` array which is used only for the "picked doc id" check which would avoid the "was it copied before or after the picked ids were set" questions. An extra implementation subtlety is as follows: get-picked-interleaving-doc-ids applies only to interleaving queries and because of that the "copy-and-modify" approach was required for `rerankingQueries` before but if the already existing queries from the context can be used for the picked-or-not check then the `rerankingQueries` can follow a "construct-a-new-one" approach instead with the constructed queries all being of the same `LTRScoringQuery` base class type.

solr/contrib/ltr/src/java/org/apache/solr/ltr/LTRScoringQuery.java
```diff
@@ -102,6 +102,26 @@ public LTRScoringModel getScoringModel() {
...
+    this.ltrThreadMgr = ltrThreadMgr;
+  }
+
+  public void setExtractAllFeatures(boolean extractAllFeatures) {
+    this.extractAllFeatures = extractAllFeatures;
+  }
+
+  public void setEfi(Map<String, String[]> efi) {
+    this.efi = efi;
+  }
```
cpoerschke, line 119: LTRFeatureLoggerTransformerFactory.8 - with a `rerankedQueriesFromContext` + `rerankedQueries` approach in the feature logger transformer the setters here would not be needed.

cpoerschke:
Hi @alessandrobenedetti, I returned to this pull request today, both the changes above and the tests (which I hadn't looked at before). Very comprehensive test coverage, thank you. Have pushed all my remaining insights to the https://github.com/cpoerschke/lucene-solr/commits/feature/SOLR-14560-cpoerschke-2 branch -- for the https://github.com/cpoerschke/lucene-solr/commit/4912daccd596435f5c61ac1a3cf86eaebb039118 and https://github.com/cpoerschke/lucene-solr/commit/3a61287a0e4fb5a77f080a92e7129582b234cbd7 commits which are perhaps a bit subtle I've added annotations on the pull request here -- the other commits are hopefully relatively self-explanatory. Let me know what you think, I agree the commit phase is fast approaching here :)

solr/contrib/ltr/src/java/org/apache/solr/ltr/interleaving/TeamDraftInterleaving.java
cpoerschke: > ... currently Interleaving doesn't support sharding ...

Let's include that in the documentation somehow, e.g. https://github.com/cpoerschke/lucene-solr/commit/7edd6005d6daa29b5a80bfd2a0e1249c23bd9279 say.

> ... So I rolled back the ScoreDoc ... Let me know if it is better!

Yes it is, thanks!

solr/contrib/ltr/src/java/org/apache/solr/ltr/response/transform/LTRFeatureLoggerTransformerFactory.java
cpoerschke: > ... I believe the code is much more readable now ... now that part is extremely clear.

Yes, I agree, very nice.

> ... another consideration sparkled: ... a separate Jira for that ...

Interesting points, will need to think about them a bit (next week). I agree it's unrelated i.e. not a blocker for this pull request here.

solr/contrib/ltr/src/java/org/apache/solr/ltr/response/transform/LTRFeatureLoggerTransformerFactory.java
alessandrobenedetti: LTRScoringQuery queries may be LTRInterleavingScoringQuery, containing the picked document Ids we don't want to lose

solr/contrib/ltr/src/java/org/apache/solr/ltr/response/transform/LTRFeatureLoggerTransformerFactory.java
alessandrobenedetti: Liked them! Will push them soon

solr/contrib/ltr/src/java/org/apache/solr/ltr/response/transform/LTRFeatureLoggerTransformerFactory.java
alessandrobenedetti: but this is solved with the queries from context :)

solr/contrib/ltr/src/java/org/apache/solr/ltr/response/transform/LTRFeatureLoggerTransformerFactory.java
alessandrobenedetti: Perfectly splendid! I agree!

alessandrobenedetti:
Hi @cpoerschke  I just pushed all your commits, I was happy with all of them :) 
If there's no other observation I would proceed committing next week ( it will be my first direct commit, so I'll need to take a look to official guidelines and target for merging).
I assume we merge to master squashing and then cherry-pick the commit to some other branches?
Do you think we need to target a major release? Or we could add it in the upcoming minors?
Have a nice weekend and thank you for your help, it has been greatly appreciated!

cpoerschke:
> ... I assume we merge to master squashing and then cherry-pick the commit to some other branches? ...

Yes, squash-and-merge to master branch and then cherry-pick to branch_8x from which branch_8_8 would be cut in due course as part of the 8.8.0 release process.

> ... Do you think we need to target a major release? Or we could add it in the upcoming minors? ...

Good question.
* From end users' perspective there are no breaking changes which would point towards targeting 9.x major rather than 8.8+ minor releases.
* From code API perspective, there are some APIs that changed but from what I can see -- https://github.com/cpoerschke/lucene-solr/commits/feature/SOLR-14560-cpoerschke-2 has associated commits -- we could provide backwards compatible deprecated niceties to avoid breaking builds for anyone who perhaps had created their own transformer or plugin class that referenced the changed APIs. Having said that, perhaps for `solr/contrib` code the expectations of not breaking APIs in a given 8.x series are different from `solr/core` code.
* But yes, with or without backcompat niceties, I think we can target the 8.8.0 minor rather than then 9.0 major release.

alessandrobenedetti:
I am not against the back-compat niceties, but is there any automatic tool that will remove them when a release candidate happen for 9.0 ?
If not I suspect we may risk forgetting and in that case, I prefer not to have them (and leave the integrators the responsibility in adapting their plugin when updating).

For the original ranking name parameter I am not sure it's actually necessary, I would probably go ahead with the commit as it is in this current version.

Let me know and I'll commit tomorrow and checcy pick :)

alessandrobenedetti:
Thanks @cpoerschke for the thorough review!

PR #1572, opened 2020-06-11
SOLR-14561 CoreAdminAPI's parameters instanceDir and dataDir are now validated

janhoy:
See https://issues.apache.org/jira/browse/SOLR-14561

The `instanceDir` and `dataDir` params must now be relative to either `SOLR_HOME`, `SOLR_DATA_HOME` or `coreRootDir`.

Added new solr.xml config 'allowPaths', controlled by system property 'solr.allowPaths' that allows you to add other allowed paths when needed.


janhoy:
I think the only thing I'm lacking is a real integration test. I validated manually that core creation fails in `/tmp`, and that setting `-Dsolr.allowPaths=/tmp` allows it:
allowPath

solr/core/src/java/org/apache/solr/core/CoreContainer.java
```diff
@@ -1259,6 +1277,20 @@ public SolrCore create(String coreName, Path instancePath, Map<String, String> p
...
   }
 
+  /**
+   * Checks that the given path is relative to SOLR_HOME, SOLR_DATA_HOME, coreRootDirectory or one of the paths
+   * specified in solr.xml's allowPaths element.
+   * @param path path to check
+   * @throws SolrException if path is outside allowed paths
+   */
+  public void assertPathAllowed(Path path) throws SolrException {
+    if (path.normalize().equals(path) && !path.isAbsolute()) return;
```
tflobbe, line 1287: This doesn't cover the case of a `../foo` path, right? is that covered somewhere else?

janhoy: You are right. We need a more thorough check
* Disallow relative paths starting with "."
* Always `normalize()` the path before `toAbsolutePath()` to catch the `/var/solr/../../etc` case, else that example would return true for `startsWith("/var/solr")`

janhoy: I pushed a commit addressing this. Quite a few changes, but we now detect `..` specifically.

solr/core/src/java/org/apache/solr/core/SolrPaths.java
```diff
@@ -128,4 +130,33 @@ private static void logOnceInfo(String key, String msg) {
...
+   *   <li>Relative paths starting with <code>..</code></li>
+   *   <li>Windows UNC paths (<code>\\host\share\path</code>)</li>
+   *   <li>Absolute paths which are not below the list of allowed paths</li>
+   * </ul>
+   * @param pathToAssert path to check
+   * @param allowPaths list of paths that should be allowed prefixes
+   * @throws SolrException if path is outside allowed paths
+   */
+  public static void assertPathAllowed(Path pathToAssert, Set<Path> allowPaths) throws SolrException {
+    if (OS.isFamilyWindows() && pathToAssert.toString().startsWith("\\\\")) {
```
janhoy, line 147: I have not tested this on Windows. On my mac, the `Path` class uses an OSX implementation so I think it will not detect the UNC style path, it does not manage to normalize or make it absolute, so I scoped the check for Windows only. I test on the string version before normalizing since normalize may mess up UNC paths.

I decided to block UNC totally instead of trying to be smart about it. Users can always map a drive letter to the desired share to work around it?

solr/core/src/test-files/solr/solr-50-all.xml
```diff
@@ -24,6 +24,7 @@
   <str name="configSetsHandler">testConfigSetsHandler</str>
   <str name="managementPath">testManagementPath</str>
   <str name="sharedLib">testSharedLib</str>
+  <str name="allowPaths">/tmp,/home/john</str>
```
dsmiley, line 27: Since this file is a NamedList, shouldn't we be using `<arr` for an array of strings?

janhoy: It could, but to make it convenient to set it as a Java Sysprop, I kept it as a simple string. De  default `solr.xml` defines it as `${solr.allowPaths:}` so that `-Dsolr.allowPaths=foo,bar,baz` will work. I'm not aware of any way to specify a native array as Java Sysprop?

solr/core/src/java/org/apache/solr/core/CoreContainer.java
```diff
@@ -338,6 +339,19 @@ public CoreContainer(NodeConfig config, CoresLocator locator, boolean asyncSolrC
         ExecutorUtil.newMDCAwareCachedThreadPool(
             cfg.getReplayUpdatesThreads(),
             new SolrNamedThreadFactory("replayUpdatesExecutor")));
+
+    this.allowPaths = new java.util.HashSet<>();
+    this.allowPaths.add(cfg.getSolrHome().toAbsolutePath());
```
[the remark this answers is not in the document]
janhoy: Ok, simplified it.

solr/core/src/test-files/solr/solr-50-all.xml
dsmiley: ok... I suppose both could be supported provided that the consumer checked with the aid of some new utility method that doesn't exist like `Utils.toListOfString(Object, char separator) : String` but that might be a bit much.  I suspect we have other places where we trivially use a delimited string and fail to leverage NamedList.  Your call; no strong opinion here.

solr/core/src/java/org/apache/solr/core/SolrPaths.java
janhoy: Anyone who have a Windows box to test this on?

janhoy:
Ran the whole test suite and uncovered various tests that use "illegal" temp test folders, that now fail. That was expected. So the last commit ba0b544 addresses these tests:
* Give a way to whitelist all paths by setting `-Dsolr.allowPaths=*`
* Add a `CoreContainer.getAllowPaths()` method that tests use to allow individual folders (I like that better than letting tests set global sysprops)
* This also led to a small change in the path comparison - we now convert Path -> String -> Path to make sure paths are comparable, even Lucene's `FilterPath` class used in tests

To review, the easiest is probably to just load the last commit ba0b544

janhoy:
The tests still don't pass on Windows, and I have found the reason. Will push a few more changes on friday.

solr/core/src/java/org/apache/solr/core/SolrPaths.java
janhoy: I have tested on Windows, validated that UNC is blocked, and modified the tests with separate ones running in Windows and non-Windows environments.

janhoy:
All tests now passing on macOS and hopefully Windows (running tests now in a slow VirtualBox).

solr/core/src/java/org/apache/solr/core/CoreContainer.java
```diff
@@ -1259,6 +1277,28 @@ public SolrCore create(String coreName, Path instancePath, Map<String, String> p
...
+   * specified in solr.xml's allowPaths element. Delegates to {@link SolrPaths#assertPathAllowed(Path, Set)}
+   * @param pathToAssert path to check
+   * @throws SolrException if path is outside allowed paths
+   */
+  public void assertPathAllowed(Path pathToAssert) throws SolrException {
+    SolrPaths.assertPathAllowed(pathToAssert, allowPaths);
+  }
+
+  /**
+   * <p>Return the file system paths that should be allowed for various API requests.
```
janhoy, line 1291: @dsmiley see JavaDoc. I was hoping to keep this method private. Don't we have a special annotation that will allow access from test scope even if the method is not public?

solr/core/src/java/org/apache/solr/core/SolrPaths.java
```diff
@@ -128,4 +130,35 @@ private static void logOnceInfo(String key, String msg) {
...
+          "Path " + pathToAssert + " disallowed. UNC paths not supported. Please use drive letter instead.");
+    }
+    // Conversion Path -> String -> Path is to be able to compare against org.apache.lucene.mockfile.FilterPath instances
+    final Path path = Path.of(pathToAssert.toString()).normalize();
+    if (path.startsWith("..")) {
+      throw new SolrException(SolrException.ErrorCode.BAD_REQUEST,
+          "Path " + pathToAssert + " disallowed due to path traversal..");
+    }
+    if (!path.isAbsolute()) return; // All relative paths are accepted
+    if (allowPaths.contains(Paths.get("_ALL_"))) return; // Catch-all path "*"/"_ALL_" will allow all other paths
```
janhoy, line 158: This is the workaround I did after realizing that Windows `Path` class is not happy with `*` as a path. When parsing the value from solr.xml/sysprop, we detect `*` and store it as a Path `_ALL_`. Then in the assert method we check for that special path and skip further testing.

Exception is UNC paths and `..` paths which are still rejected (should they?)

solr/core/src/java/org/apache/solr/core/CoreContainer.java
dsmiley: Javadoc is good; thanks.
`com.google.common.annotations.VisibleForTesting`

janhoy:
I think this is approaching committable state. Appreciate if someone with a good Windows box would run the full test suite on Windows. But I think I'll anyway merge to master and let Jenkins work on it for a few rounds. Then I'll backport to 8.x branch in good time before 8.6 branch cut.

dsmiley:
@mkhludnev ?  I recall you use Windows.

PR #1574, opened 2020-06-12
SOLR-14566: Add request-ID to all distrib-search requests

gerlowskija:
# Description

When reading through logs, associating the log message from the coordinator node with the messages for per-shard requests downstream can be difficult.  Nothing unique (or semi-uniquely) identifies shard requests as being part of an upstream request. 

# Solution

This PR moves some logic out of DebugComponent that generates a guarantee-ably unique request ID.  On coordinator nodes this value is logged out as a special field in the request log message.  On downstream shard requests it appears as "rid" in the parameter list.

# Tests

Manual testing to ensure the log field is added in appropriate situations.  Some changes to DebugComponent to account for its new role as a consumer of "rid" only.  Looked to update SearchHandlerTest but couldn't think of a non-hacky way to test this.  Anyone has any ideas please lmk.

# Checklist


dsmiley:
+1 clever and useful

solr/core/src/java/org/apache/solr/handler/component/SearchHandler.java
```diff
@@ -500,6 +509,29 @@ public void handleRequestBody(SolrQueryRequest req, SolrQueryResponse rsp) throw
...
 
+  private void tagRequestWithRequestId(ResponseBuilder rb) {
+    String rid = getRequestId(rb.req);
+    if (StringUtils.isBlank(rb.req.getParams().get(CommonParams.REQUEST_ID))) {
+      ModifiableSolrParams params = new ModifiableSolrParams(rb.req.getParams());
+      params.add(CommonParams.REQUEST_ID, rid);//add rid to the request so that shards see it
+      rb.req.setParams(params);
+    }
+    if (rb.isDistrib) {
+      rb.rsp.addToLog(CommonParams.REQUEST_ID, rid); //to see it in the logs of the landing core
```
tflobbe, line 520: Do we now want it also in the cordinator node?

solr/core/src/java/org/apache/solr/handler/component/SearchHandler.java
```diff
@@ -500,6 +509,29 @@ public void handleRequestBody(SolrQueryRequest req, SolrQueryResponse rsp) throw
...
+
+  public static String getRequestId(SolrQueryRequest req) {
+    String rid = req.getParams().get(CommonParams.REQUEST_ID);
+    return StringUtils.isNotBlank(rid) ? rid : generateRid(req);
+  }
+
+  @SuppressForbidden(reason = "Need currentTimeMillis, only used for naming")
+  private static String generateRid(SolrQueryRequest req) {
+    String hostName = req.getCore().getCoreContainer().getHostName();
+    return hostName + "-" + req.getCore().getName() + "-" + System.currentTimeMillis() + "-" + ridCounter.getAndIncrement();
```
dsmiley, line 532: Wow, this could be a long string, and it'll show up as a parameter and is thus kinda in-your-face when looking at logs.  And this, you propose, is by default. Isn't that a concern, albeit not huge?

solr/core/src/java/org/apache/solr/handler/component/SearchHandler.java
```diff
@@ -70,6 +73,11 @@
 
   private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
 
+  /**
+   * A counter to ensure that no RID is equal, even if they fall in the same millisecond
+   */
+  private static final AtomicLong ridCounter = new AtomicLong();
```
dsmiley, line 79: LongAdder is probably a better choice

solr/core/src/java/org/apache/solr/handler/component/SearchHandler.java
janhoy: Does the string need to be human readable? A 32 bit hash based on node-id+counter would be unique enough?

janhoy:
Perhaps a stupid question, but how does this compare with the jaegertracing / opentracing support? Would not that already provide some request id to follow a request across nodes? I have absolutely no in-depth knowledge of that module so have me excused if the question is far out :)

solr/core/src/java/org/apache/solr/handler/component/SearchHandler.java
gerlowskija: I'm a bit torn.

I agree that the length makes this a bit "in-your-face", and that a hash would be a good way to condense things.  I don't love the verbose-ness as things are now.  But, technically speaking, that loses us our uniqueness guarantee.  And if we're not shooting for strict-uniqueness here, then aren't we just as good with the NOW timestamp that was originally proposed but set aside for lack of uniqueness?

To zoom out, we've got three options for what the identifier is here:

1. **NOW timestamp** - concise with some semantic meaning, but not guaranteed to be unique
2. **RequestID** - very semantically meaningful and guaranteed to be unique, but "in your face"
3. **Hashed RequestID** - concise, but no semantic meaning and not guaranteed to be unique (though slightly more unique than a NOW timestamp?)

I originally implemented (1), and then moved to (2) based on feedback in JIRA.  But if there's no consensus there then maybe I'll go back to (1).  In any case I don't see much advantage in (3) over (1)

solr/core/src/java/org/apache/solr/handler/component/SearchHandler.java
gerlowskija: 👍 I'll look into it.  Tbh I moved this from DebugComponent and didn't consider alternatives much.

solr/core/src/java/org/apache/solr/handler/component/SearchHandler.java
gerlowskija: We do!  That's actually what this line does.  L520 (this line) adds it on the coordinator, and L516 adds it to sub-requests, so that they also get the request-id as a query-param.

gerlowskija:
> Perhaps a stupid question, but how does this compare with the jaegertracing / opentracing support? Would not that already provide some request id to follow a request across nodes? I have absolutely no in-depth knowledge of that module so have me excused if the question is far out :)

@janhoy

It's not a dumb question.  The way I look at it, there's two questions to ask about OpenTracing as it relates to this PR: (1) Why do we need this PR if we've got OpenTracing?, and (2) Does OpenTracing have id-generation code we can reuse here?

For the first question - the value of this PR is that it gets us very lightweight request correlation out of the box in Solr's logs on all distributed queries.  OpenTracing is vastly more powerful, but it requires more setup and more resources: it requires enabling of a contrib (or use of a custom plugin), it requires an external Jaeger server to ship data to (and consumes network bandwidth to do so) , and because it's heavier-weight OpenTracing is typically only run on a sample percentage of requests. It's not reliably present for all requests, unless you crank the sample % up to 100 and take a performance hit.

So in practice OpenTracing and the request-id stuff I'm adding here are good for different audiences and use-cases.  OpenTracing is going to be primarily useful for teams which want to have ongoing monitoring/tracing on requests coming through their cluster.  And of course it's going to appeal more to teams that have the Ops personnel to manage a Jaeger deployment, etc.  OTOH, Request-id logging is primarily useful for a Solr expert debugging a small time-window of logs.

For the second question - if you can believe it, Solr's OpenTracing integration doesn't have any explicit ID-generation code.  Instead the ID-generation is done inside of some Jaeger code (see `JaegerSpanContext.convertTraceId`).  Technically we could reuse this code, but it'd require creating a small graph of Jaeger objects.  Since it'd couple us directly to Jaeger and would involve extra object creation, I didn't think this was a good way to go.

solr/core/src/java/org/apache/solr/handler/component/SearchHandler.java
gerlowskija: Lmk if anyone has a strong opinion here.  I've gotten feedback from both directions (make it unique (but verbose) vs make it concise (but not unique)).  I'm going to just choose one in the next day or two to avoid dragging things out.

dsmiley: I think the uniqueness goal is important (thus don't need hash), but we could cut back on the verbose-ness.  node ID + counter is, I think, about right.

dsmiley:
I agree with Jason's characterization relating to multiple use-cases, but there is overlap, and it'd be nice to see shared code for overlapping concerns instead of completely separate.  For example, tracing sets `MDCLoggingContext.setTracerId` which populates the 't' in every log -- cool!  Yet adding some new request ID parameter would not populate this but it would populate a request parameter.  So if a query triggers additional log lines for whatever reason, those log lines wouldn't have the request ID visible.  Furthermore, this new request ID being added is only for distributed search; it's not comprehensive (not distributed indexing or administrative requests).  To me, that's a shame.  But I'm not going to block what you're doing... progress not perfection, I suppose.

What I would love to see is a simple Tracer implementation that does no distributed tracing (no tracing server).  It would merely generate an ID and propagate it via HTTP headers.  Thanks to Dat's existing open tracing integration work, this ID would show up in logs and would get propagated.

solr/core/src/java/org/apache/solr/handler/component/SearchHandler.java
gerlowskija: I don't think LongAdder is appropriate here - it doesn't have an incrementAndGet or anything similar that would ensure threads get unique counter values.

The "atomicity" that it offers, afaict, is that calls to `add(long)` can't be clobbered across threads and will eventually be incorporated into the sum().  Nothing guarantees that two threads can't get the same `sum()` value at a given time, without additional synchronization.  I found this [SO post](https://stackoverflow.com/questions/45269587/does-java-longadders-increment-sum-prevent-getting-the-same-value-twice) helpful.

Now that I look at the synchronization generally though it reinforces my impression that I need to benchmark this before I commit, to ensure the lock contention isn't noticeable.  It's a shame LongAdder doesn't look like a good fit.

dsmiley: It's really not a big deal and I'd be shocked if it noticeably impacted performance.

solr/core/src/java/org/apache/solr/handler/component/SearchHandler.java
```diff
@@ -510,14 +510,18 @@ public void handleRequestBody(SolrQueryRequest req, SolrQueryResponse rsp) throw
...
   private void tagRequestWithRequestId(ResponseBuilder rb) {
-    String rid = getRequestId(rb.req);
-    if (StringUtils.isBlank(rb.req.getParams().get(CommonParams.REQUEST_ID))) {
-      ModifiableSolrParams params = new ModifiableSolrParams(rb.req.getParams());
-      params.add(CommonParams.REQUEST_ID, rid);//add rid to the request so that shards see it
-      rb.req.setParams(params);
-    }
-    if (rb.isDistrib) {
-      rb.rsp.addToLog(CommonParams.REQUEST_ID, rid); //to see it in the logs of the landing core
+    final String disableFlag = rb.req.getParams().get(CommonParams.DISABLE_REQUEST_ID);
```
dsmiley, line 513: I don't think this flag should be a request parameter; I think it should be a "Cluster Property".

Secondly, if you do need to get this from a SolrParams, see `getBool` method, accepting a default.

solr/core/src/java/org/apache/solr/handler/component/SearchHandler.java
gerlowskija: I agree, but I wanted an excuse to try out the perf [framework](https://github.com/thesearchstack/solr-bench) that Ishan has been teasing off and on for awhile, so I tested it anyways.  With 30 threads submitting queries in parallel it added ~1.5ms to the p95 times.

Sounds pretty insignificant to me.  I added a flag query param to disable it in any case.

solr/core/src/java/org/apache/solr/handler/component/SearchHandler.java
gerlowskija: Man you're quick on the reviews.

I like a SolrParam better than a cluster property.  Users might well have reasons to enable/disable it differently based on the source or type of query.  A Cluster Property doesn't offer that sort of granularity (at least, as I understand them).

dsmiley: Having to disable one parameter showing up by adding another parameter kinda defeats the point.  To some people, this new parameter is needless noise.  To them, your solution is to tell them to add more noise (another param) to make that one go away.  I don't understand why users would toggle this on a query-by-query basis.  I suspect people holistically like this new ID on all their requests, or they don't.  But maybe you have something in mind?  I suppose it *could* be both -- you set the default as a cluster property but allow an override for the request.  Hmm; yeah I rather like that.

solr/core/src/java/org/apache/solr/handler/component/SearchHandler.java
```diff
@@ -500,6 +508,31 @@ public void handleRequestBody(SolrQueryRequest req, SolrQueryResponse rsp) throw
     }
   }
 
+  private void tagRequestWithRequestId(ResponseBuilder rb) {
+    final boolean ridTaggingDisabled = rb.req.getParams().getBool(CommonParams.DISABLE_REQUEST_ID, false);
+    if (! ridTaggingDisabled) {
+      String rid = getRequestId(rb.req);
+      if (StringUtils.isBlank(rb.req.getParams().get(CommonParams.REQUEST_ID))) {
+        ModifiableSolrParams params = new ModifiableSolrParams(rb.req.getParams());
```
dsmiley, line 516: Maybe we don't need a flag parameter to decide what to do with rid.  What if Solr sees rid=\* (or true?) then it knows to generate a rid, and if it sees a blank/empty string (or false?) then it does nothing.  If it's something else than it *is* the rid (either from the client or generated).  One parameter to look at for this feature seems nicer than two.  Also, relating to my concern about wholesale disabling, it could be set in the solrconfig as a param default.

solr/core/src/java/org/apache/solr/handler/component/SearchHandler.java
```diff
@@ -298,7 +305,8 @@ public void handleRequestBody(SolrQueryRequest req, SolrQueryResponse rsp) throw
     final RTimerTree timer = rb.isDebug() ? req.getRequestTimer() : null;
 
     final ShardHandler shardHandler1 = getAndPrepShardHandler(req, rb); // creates a ShardHandler object only if it's needed
-    
+
+    tagRequestWithRequestId(rb);
```
dsmiley, line 309: super minor but please add a blank line after this because it's not related to timer.

shalinmangar:
I think this is a good idea in general. All I would request is for users to be able to provide their own request-id which should disable the auto-generation of request id inside solr if present. This will help users track requests in their entire stack including solr.

gerlowskija:
> All I would request is for users to be able to provide their own request-id which should disable the auto-generation of request id inside solr if present

This is actually how it works currently- if there's already an 'rid' param present, SearchHandler will use that value and propagate it, in lieu of generating its own.

solr/core/src/java/org/apache/solr/handler/component/SearchHandler.java
janhoy: > I like a SolrParam better than a cluster property. Users might well have reasons to enable/disable it differently based on the source or type of query.

If users want full per-query control, they don't need a Solr feature at all, they can just start including `whatever=1234` in their requests and track it in their logs. :)

solr/core/src/java/org/apache/solr/handler/component/SearchHandler.java
gerlowskija: I like the simplicity of having the flag be its own separate parameter over packing in multiple different "flag" values to the existing param.

Simplicity and intuitiveness are admittedly subjective though, I'm open to your opinion if you can convince me why you see having only a single param is easier for users to digest.

dsmiley: Parameter / cognitive overload / excessive.

If a user passes `rid=whatever` then Solr will pass this along (thanks to you); right?  Then in a sense it's self-disableable by a user with a blank rid and putting it in solrconfig.xml init param for `/*`.

gerlowskija: I guess I just see that as being more complex.  Imagine describing that behavior to a newcomer in one sentence: "Well if you don't specify it at all we do X, if you do specify it we do Y, and if you specify it as blank we do Z".  Versus: "It either autogenerates a request ID or uses whatever you pass in.  There's also a flag to disable it if you want".

I'm the first to admit these sort of takes are subjective though.  If someone else chimes in here and I get outnumbered I'll change it for sure.  But I'll stick to my guns for now.

solr/core/src/java/org/apache/solr/handler/component/SearchHandler.java
```diff
@@ -500,6 +508,31 @@ public void handleRequestBody(SolrQueryRequest req, SolrQueryResponse rsp) throw
...
+        params.add(CommonParams.REQUEST_ID, rid);//add rid to the request so that shards see it
+        rb.req.setParams(params);
+      }
+      if (rb.isDistrib) {
+        rb.rsp.addToLog(CommonParams.REQUEST_ID, rid); //to see it in the logs of the landing core
+      }
+    }
+  }
+
+  public static String getRequestId(SolrQueryRequest req) {
```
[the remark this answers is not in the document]
gerlowskija: To documenting: sure +1.

To your other point, I'm more on the fence.  If we always stored the RID in params, this would be a no-brainer.  But on the coordinator node we don't because it's displayed separately in coordinator log messages.

I spent some time fiddling with the composition of the helper methods here before the long weekend but couldn't come up with an arrangement that made sense, without throwing everything in one larger method (that's hard for other places, like DebugComponent, to reuse.).  I did come up with some name clarifications and Javadocs here though - and I hope those address your underlying concern.

solr/core/src/java/org/apache/solr/handler/component/SearchHandler.java
dsmiley: > "It either autogenerates a request ID or uses whatever you pass in. There's also a flag to disable it if you want"

That's also an X, Y, and Z :-)

In my suggestion, the consequence of a blank rid needn't be a special case -- treat it as valid if you like.  It's super-short "rid=" and so adds negligible noise when propagated.

dsmiley:
Shouldn't this RID be saved to MDC like a tracer ID does?

If there is a tracer ID active, maybe this code should either (a) do nothing, or (b) use the tracer ID instead of generate one?

gerlowskija:
> Shouldn't this RID be saved to MDC like a tracer ID does?

**TL;DR** It _could_, but it _shouldn't_ for practical reasons.

It could.  Also, you could do this on all requests, not just distributed ones.  And you could get it working on indexing requests too, and admin requests which store chunks of work on the overseer queue.  Ideally you could do this as some new OpenTracing implementation that would serve as an alternative to the JaegerTracing implementation that's documented right now.  And you could update Joel's SolrLogPostTool (which is really cool btw - just to plug it) to parse out the rid's as their own field.

There's a lot of really cool and powerful improvements to make here.  They're just not what I'm setting out to do right now.  SOLR-8274 was created 5 years ago to add MDC-based tracing, and I'd prefer that approach stay on that JIRA and not spread to SOLR-14566.  I created SOLR-14566 as a trivial fix to get users some very-performant, configured ootb tracing while others pursue a more comprehensive MDC approach in parallel.  If someone ever does do SOLR-8274, they can rip out this logic here.  But I want to get users (and really, the debuggers that support them behind the scenes) _something_ in the meantime.  To use the perfect vs good analogy - I'd rather not tie the "good" of getting this in now to the "perfect" of SOLR-8274 that I hope gets in someday but never generated much momentum.

PR #1576, opened 2020-06-13
Alternative approach to LUCENE-8962

s1monw:
this is just a prototype - only intended for discussion

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
+++ b/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
@@ -33,6 +33,7 @@ import java.util.Queue;
 import java.util.Set;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.Semaphore;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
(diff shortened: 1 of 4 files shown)


s1monw:
see #1552 for reference

lucene/core/src/java/org/apache/lucene/index/MergePolicy.java
```diff
@@ -399,8 +423,19 @@ public String segString(Directory dir) {
       }
       return b.toString();
     }
+
+    boolean await(long timeout, TimeUnit unit) {
+      for (OneMerge merge : merges) {
+        if (merge.await(timeout, unit) == false) {
```
dsmiley, line 429: This looks suspicious when there is more than one merge.  Shouldn't the timeout decrease as time is used up by earlier merges?

In practice, when is there more than one?  I've been confused on this matter when I developed a custom MP/MS.

s1monw: in a real change that's correct. in a prototype as this is it's really just there to visualize the idea. I didn't do this on purpose to not discuss impl details. that's not the point of this it's really just a PR to make commenting simpler.

msfroh:
This approach makes sense to me. 

I like how much simpler the addition of MergeSpecification.await() makes things, versus the CountDownLatch hackery of the previous approach. Also, updatePendingMerges returning the MergeSpecification is much cleaner than explicitly computing a merge from within prepareCommitInternal.

s1monw:
@msfroh I opened https://github.com/apache/lucene-solr/pull/1585 to make it easier to do this.

lucene/core/src/java/org/apache/lucene/index/MergePolicy.java
s1monw: I fixed this in the followup PR https://github.com/apache/lucene-solr/pull/1585

s1monw:
superseded by #1585

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
```diff
@@ -2144,39 +2145,43 @@ private synchronized boolean updatePendingMerges(MergePolicy mergePolicy, MergeT
...
         final int numMerges = spec.merges.size();
         for(int i=0;i<numMerges;i++) {
           final MergePolicy.OneMerge merge = spec.merges.get(i);
           merge.maxNumSegments = maxNumSegments;
         }
       }
     } else {
-      spec = mergePolicy.findMerges(trigger, segmentInfos, this);
+      switch (trigger) {
+        case COMMIT:
```
dsmiley, line 2170: There is an inconsistency here that suggests something is wrong, or at least confusing enough to deserve a comment.  For case COMMIT, we call a findFullFlushMerges.  Shouldn't it be on case FULL_FLUSH to be consistent with the method we are calling?  Or should findFullFlushMerges be called findCommitMerges?

mikemccand: Full flush happens for refresh and commit.

But we have only implemented "merge on commit" so far.

I would love to also add "merge on refresh", but until we do so, I think it's correct for `IndexWriter` to separate out the `COMMIT` case here like this.

PR #1579, opened 2020-06-15
SOLR-14541: Ensure classes that implement equals implement hashCode or suppress warnings

ErickErickson:
I've created new hashCode methods for all of the classes that implement equals but not hashCode and removed associated SuppressWarnings. I've marked certain of them with TODOs where I'm really uncertain what the right thing to do is to draw attention, but I'd appreciate people looking at the others.

gw check succeeds, but that's all I'm guaranteeing at present. I need to let this bake a while and come back and re-visit them in detail if people who know the particular code better don't give a thumbs-up. hashCode implementations can be tricky, and these are definitely straw-man.

I'll add some more comments on the JIRA


solr/core/src/java/org/apache/solr/cloud/rule/Rule.java
```diff
@@ -365,7 +365,7 @@ public boolean equals(Object obj) {
 
     @Override
     public int hashCode() {
-      throw new UnsupportedOperationException("TODO unimplemented");
+      return Objects.hash(name, fuzzy);
```
murblanc, line 368: Given that `fuzzy` is not included in the `equals()` implementation just above, we could have two instances that are `equals()` but that have different values for `hashCode()`. I believe this is not correct.

solr/core/src/java/org/apache/solr/pkg/PackageAPI.java
```diff
@@ -211,7 +211,7 @@ public boolean equals(Object obj) {
 
     @Override
     public int hashCode() {
-      throw new UnsupportedOperationException("TODO unimplemented");
+      return Objects.hash(version, manifestSHA512);
```
murblanc, line 214: `equals()` above does not compare `manifestSHA512`.
Unless `manifestSHA512` are equal if `version` are (in which case no need to add it here?), we could have two objets that are `equals()` yet have different `hashCode()`, which would be against the spec (see `java.lang.Object.hashCode()` javadoc).

solr/solrj/src/java/org/apache/solr/client/solrj/cloud/autoscaling/AutoScalingConfig.java
```diff
@@ -305,10 +303,10 @@ public boolean equals(Object o) {
       return properties.equals(that.properties);
     }
 
-//    @Override
-//    public int hashCode() {
-//      throw new UnsupportedOperationException("TODO unimplemented");
-//    }
+    @Override
+    public int hashCode() {
+      return Objects.hash(name, actionClass);
```
murblanc, line 308: `equals()` is based on comparing `properties` and `hashCode()` on comparing `name` and `actionClass`. This can't guarantee two objets that are `equals()` are going to have the same `hashCode()`, as required.

solr/solrj/src/java/org/apache/solr/client/solrj/cloud/autoscaling/AutoScalingConfig.java
```diff
@@ -587,11 +585,11 @@ public boolean equals(Object o) {
...
     return getProperties().equals(that.getProperties());
   }
-//  @Override
-//  public int hashCode() {
-//    throw new UnsupportedOperationException("TODO unimplemented");
-//  }
 
+  @Override
+  public int hashCode() {
+    return Objects.hash(policy, zkVersion);
```
murblanc, line 591: `equals()` does not compare `zkVersion`, so `hashCode()` should not be based on it either.

solr/solrj/src/java/org/apache/solr/client/solrj/cloud/autoscaling/Policy.java
```diff
@@ -288,10 +287,11 @@ public boolean equals(Object o) {
...
   }
 
-//  @Override
-//  public int hashCode() {
-//    throw new UnsupportedOperationException("TODO unimplemented");
-//  }
+  //TODO: Uncertain about this one
+  @Override
+  public int hashCode() {
+    return Objects.hash(zkVersion);
```
murblanc, line 293: `equals()` compares `policies`, `clusterPolicy` and `clusterPreferences`, but does not compare `zkVersion`.
`hashCode()` should not use `zkVersion` in hash computation.

solr/solrj/src/java/org/apache/solr/common/cloud/DocCollection.java
```diff
@@ -383,10 +382,12 @@ public boolean equals(Object that) {
...
 
-//  @Override
-//  public int hashCode() {
-//    throw new UnsupportedOperationException("TODO unimplemented DocCollection.hashCode");
-//  }
+  @Override
+  public int hashCode() {
+    return Objects.hash(super.hashCode(), znodeVersion, name, replicationFactor,
+            numNrtReplicas, numTlogReplicas, numPullReplicas, maxShardsPerNode,
+            autoAddReplicas, policy, readOnly);
```
murblanc, line 389: The superclass (`ZkNodeProps`) `hashCode()` hashes the `props` map, so already takes care of producing a hash value based on `replicationFactor`, `numNrtReplicas`, `numTlogReplicas`, `numPullReplicas`, `maxShardsPerNode`, `autoAddReplicas`, `policy` and `readOnly` that were extracted from the map. These do not have to be added here again but it's **not** incorrect to have them.

solr/solrj/src/java/org/apache/solr/common/cloud/Replica.java
```diff
@@ -153,10 +152,12 @@ public boolean equals(Object o) {
...
     return name.equals(replica.name);
   }
-//  @Override
-//  public int hashCode() {
-//    throw new UnsupportedOperationException("TODO unimplemented Replica.hashCode()");
-//  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(name, nodeName, collection);
```
murblanc, line 158: `equals()` does not compare `nodeName` and `collection`, so `hashCode()` should not be based on these values.

solr/solrj/src/java/org/apache/solr/common/cloud/ZkNodeProps.java
```diff
@@ -171,8 +167,10 @@ public boolean getBool(String key, boolean b) {
...
   }
-//  @Override
-//  public int hashCode() {
-//    throw new UnsupportedOperationException("TODO unimplemented ZkNodeProps.hashCode");
-//  }
+
+  //TODO: I'm very uncertain about this
+  @Override
+  public int hashCode() {
+    return 0;
```
murblanc, line 174: A constant `hashCode()` return value is correct but does not _spread_ objects into buckets in any useful way. Here we could return `propMap.hashCode()` instead.

solr/solrj/src/java/org/apache/solr/common/util/ValidatingJsonMap.java
```diff
@@ -348,10 +348,12 @@ public boolean equals(Object that) {
     return that instanceof Map && this.delegate.equals(that);
   }
 
-//  @Override
-//  public int hashCode() {
-//    throw new UnsupportedOperationException("TODO unimplemented ValidatingJsonMap.hashCode");
-//  }
+  //TODO: Really uncertain about this. Hashing the map itself seems
+  // about as expensive as resolving with equals.
```
murblanc, line 352: `hashCode()` and `equals()` are not used for the same type of computations, it's not one _or_ the other, so their relative costs are not necessarily an implementation criteria.

solr/solrj/src/java/org/apache/solr/client/solrj/io/stream/CloudSolrStream.java
```diff
@@ -491,6 +490,11 @@ public boolean equals(Object o) {
       return this == o;
     }
 
+    @Override
```
gerlowskija, line 493: As someone who has at least a passing understanding of this code, this and the other Streaming class implementations below gets my +1.

solr/core/src/java/org/apache/solr/pkg/PackageAPI.java
ErickErickson: WDYT about just version here? If we include files, computing the hashcode seems like it could get unreasonably expensive. It's not clear to me how long that list could be.

Comparing two different packages that have the same version would break equals too so this seems safe.

solr/core/src/java/org/apache/solr/cloud/rule/Rule.java
ErickErickson: Just using name here.

solr/solrj/src/java/org/apache/solr/client/solrj/cloud/autoscaling/AutoScalingConfig.java
ErickErickson: equals doesn't even include name, so I guess it'll just have to be properties only to conform to the contract.

solr/solrj/src/java/org/apache/solr/client/solrj/cloud/autoscaling/AutoScalingConfig.java
ErickErickson: Looks like this just needs to be a hash on getPolicy(). I don't think it can even be the member var policy since getPolicy can change it.

solr/solrj/src/java/org/apache/solr/client/solrj/cloud/autoscaling/Policy.java
ErickErickson: Done. In the interests of speed, would just returning just the hash of getPolicies(). Even though in this case that's just the member var, future-proofing if that changes seems wise.

solr/solrj/src/java/org/apache/solr/common/cloud/DocCollection.java
ErickErickson: Fixed zkNodeProps to return the hash of the propmap. I think for this, the name and znodeversion are enough (not even adding returning super.hashCode) in the interests of speed.

solr/solrj/src/java/org/apache/solr/common/cloud/Replica.java
ErickErickson: I'll use just name.

solr/solrj/src/java/org/apache/solr/common/cloud/ZkNodeProps.java
ErickErickson: Yeah, it was late that night ;). Hashed propMap.

solr/solrj/src/java/org/apache/solr/common/util/ValidatingJsonMap.java
ErickErickson: Right. Still, every time a hashcode is computed, it iterates the entire map which could be arbitrarily large. I don't see much other choice though.

solr/solrj/src/java/org/apache/solr/client/solrj/io/stream/CloudSolrStream.java
ErickErickson: OK, thanks.

solr/core/src/java/org/apache/solr/pkg/PackageAPI.java
murblanc: I believe that when `hashCode()` is called (when using a collection) it is followed by a call to `equals()` if there's an element in that slot. More efficient (better spreading) `hashCode()` would save calls to `equals()`. Therefore, I'm not sure saving on execution time on `hashCode()` makes sense, as we will have to pay it in the cost of computing more `equals()`.
It might pay in cases where the collections will contain very few elements, in which case `hashCode()` is not really important, but in these cases its performance is likely not an issue either.

The approach I would take here is do the right thing in implementing `hashCode()`, then if we have performance issues with some objects, understand what those are and solve them (_insert Donald Knuth quote here_).

Therefore I would base `hashCode()` on version and files like `equals()`.

solr/solrj/src/java/org/apache/solr/client/solrj/cloud/autoscaling/AutoScalingConfig.java
murblanc: `name` comes from `properties` so would be ok to add it to the `hashCode()` computation, but it would not be useful if `properties` is already used.

solr/core/src/java/org/apache/solr/pkg/PackageAPI.java
murblanc: What I wrote above might not hold. Given that `equals()` does:

`        return Objects.equals(this.version, that.version)
            && Objects.equals(this.files, that.files);
`
it is not computing the equality of `files` when `version` is different, so indeed it might be better to not use `files` in `hashCode()`.

solr/solrj/src/java/org/apache/solr/client/solrj/cloud/autoscaling/Policy.java
murblanc: Agreed. Given that `equals()` calls the getter, wise to do the same in `hashCode()`.

solr/solrj/src/java/org/apache/solr/client/solrj/cloud/autoscaling/AutoScalingConfig.java
ErickErickson: Yeah, thought about including the other props, but I think we're OK with just getPolicy(). 

Like I mentioned before, though, I'm trying to limit scope here so I'm going to leave the getPolicy stuff as it is.

solr/core/src/java/org/apache/solr/pkg/PackageAPI.java
ErickErickson: Thanks,

ErickErickson:
I'll merge this in a bit via another methodl

PR #1582, opened 2020-06-15
Remove some needless toAbsolutePath calls

dsmiley:
Continuation of #1546
Here, we avoid calling toAbsolutePath when not needed because we know it's already absolute.  Look carefully; there's a few other changes.

I liken this to pointless null checks in code when you know the thing isn't null.


dsmiley:
Yes; tests pass.  I re-acquainted myself with these specific changes now and they all make sense to me.

PR #1585, opened 2020-06-16
LUCENE-8962: Allow waiting for all merges in a merge spec

s1monw:
This change adds infrastructure to allow straight forward waiting
on one or more merges or an entire merge specification. This is
a basis for LUCENE-8962.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
+++ b/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
@@ -2129,12 +2129,12 @@ public class IndexWriter implements Closeable, TwoPhaseCommit, Accountable,
 
   private final void maybeMerge(MergePolicy mergePolicy, MergeTrigger trigger, int maxNumSegments) throws IOException {
     ensureOpen(false);
-    if (updatePendingMerges(mergePolicy, trigger, maxNumSegments)) {
+    if (updatePendingMerges(mergePolicy, trigger, maxNumSegments) != null) {
       mergeScheduler.merge(mergeSource, trigger);
     }
   }
 
-  private synchronized boolean updatePendingMerges(MergePolicy mergePolicy, MergeTrigger trigger, int maxNumSegments)
+  private synchronized MergePolicy.MergeSpecification updatePendingMerges(MergePolicy mergePolicy, MergeTrigger trigger, int maxNumSegments)
     throws IOException {
 
     // In case infoStream was disabled on init, but then enabled at some
@@ -2144,22 +2144,21 @@ public class IndexWriter implements Closeable, TwoPhaseCommit, Accountable,
     assert maxNumSegments == UNBOUNDED_MAX_MERGE_SEGMENTS || maxNumSegments > 0;
     assert trigger != null;
     if (stopMerges) {
-      return false;
+      return null;
     }
 
     // Do not start new merges if disaster struck
     if (tragedy.get() != null) {
-      return false;
+      return null;
     }
-    boolean newMergesFound = false;
+
     final MergePolicy.MergeSpecification spec;
     if (maxNumSegments != UNBOUNDED_MAX_MERGE_SEGMENTS) {
       assert trigger == MergeTrigger.EXPLICIT || trigger == MergeTrigger.MERGE_FINISHED :
       "Expected EXPLICT or MERGE_FINISHED as trigger even with maxNumSegments set but was: " + trigger.name();
 
       spec = mergePolicy.findForcedMerges(segmentInfos, maxNumSegments, Collections.unmodifiableMap(segmentsToMerge), this);
-      newMergesFound = spec != null;
-      if (newMergesFound) {
+      if (spec != null) {
         final int numMerges = spec.merges.size();
         for(int i=0;i<numMerges;i++) {
           final MergePolicy.OneMerge merge = spec.merges.get(i);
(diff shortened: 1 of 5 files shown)


s1monw:
@mikemccand @msokolov @msfroh I prepared some of my changes so we can pull them in and incorporate them into the overall change.

lucene/core/src/java/org/apache/lucene/index/MergePolicy.java
```diff
@@ -399,6 +427,23 @@ public String segString(Directory dir) {
       }
       return b.toString();
     }
+
+    /**
+     * Waits if necessary for at most the given time for all merges.
+     */
+    boolean await(long timeout, TimeUnit unit) {
+      try {
+        CompletableFuture<Void> future = CompletableFuture.allOf(merges.stream()
```
dsmiley, line 436: What I learned today: `CompletableFuture.allOf(....)` -- cool, thanks!  So elegant.

Let me try and return the favor:
You can go directly to an array, avoiding toList:
```
CompletableFuture<Void> future = CompletableFuture.allOf(merges.stream()
            .map(m -> m.completable).toArray(CompletableFuture<?>[]::new));
```
BTW IntelliJ pointed this out and had an automatic replacement when I hovered my cursor over "collect".

lucene/core/src/java/org/apache/lucene/index/MergePolicy.java
```diff
@@ -399,6 +427,23 @@ public String segString(Directory dir) {
...
+            .map(m -> m.completable).collect(Collectors.toList()).toArray(new CompletableFuture<?>[0]));
+        future.get(timeout, unit);
+        return true;
+      } catch (InterruptedException e) {
+        Thread.interrupted();
```
msfroh, line 441: In my change, I remember I originally tried handling the `InterruptedException` while waiting for the `CountDownLatch` in a similar way, but it caused an intermittent test failure (in `TestIndexWriterWithThreads`, maybe?) You might run into the same test failure once you start calling this `await` method from within `IndexWriter`.

I was able to get the test passing by rethrowing it wrapped in a `ThreadInterruptedException`.

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
```diff
@@ -4289,7 +4287,7 @@ private synchronized void mergeFinish(MergePolicy.OneMerge merge) {
   @SuppressWarnings("try")
   private synchronized void closeMergeReaders(MergePolicy.OneMerge merge, boolean suppressExceptions) throws IOException {
     final boolean drop = suppressExceptions == false;
-    try (Closeable finalizer = merge::mergeFinished) {
+    try (Closeable finalizer = () -> merge.mergeFinished(suppressExceptions==false)) {
```
msokolov, line 4290: how about some spaces around the operator==?

lucene/core/src/java/org/apache/lucene/index/MergePolicy.java
```diff
@@ -362,6 +366,30 @@ public void checkAborted() throws MergeAbortedException {
...
+        return false;
+      } catch (ExecutionException | TimeoutException e) {
+        return false;
+      }
+    }
+
+    boolean isDone() {
```
mikemccand, line 386: Maybe add javadoc about the lack of thread safety here?  I.e. this could return `false` and shortly thereafter it becomes `true`.

lucene/core/src/java/org/apache/lucene/index/MergePolicy.java
```diff
@@ -362,6 +366,30 @@ public void checkAborted() throws MergeAbortedException {
...
+      return completable.isDone();
+    }
+
+    boolean isCommitted() {
+      return completable.getNow(Boolean.FALSE);
```
mikemccand, line 391: Javadoc here too?

lucene/core/src/java/org/apache/lucene/index/MergePolicy.java
```diff
@@ -196,6 +200,7 @@ final void setMergeThread(Thread owner) {
    *
    * @lucene.experimental */
   public static class OneMerge {
+    private final CompletableFuture<Boolean> completable = new CompletableFuture<>();
```
mikemccand, line 203: Could we rename the variable to `completed` or maybe `mergeCompleted`?

lucene/core/src/java/org/apache/lucene/index/MergePolicy.java
```diff
@@ -196,6 +200,7 @@ final void setMergeThread(Thread owner) {
    *
    * @lucene.experimental */
   public static class OneMerge {
```
mikemccand, line 202: Maybe we should (later, in dedicated PR) pull this out into its own java source?

lucene/core/src/java/org/apache/lucene/index/MergePolicy.java
```diff
@@ -362,6 +366,30 @@ public void checkAborted() throws MergeAbortedException {
...
+      } catch (ExecutionException | TimeoutException e) {
+        return false;
+      }
+    }
+
+    boolean isDone() {
+      return completable.isDone();
+    }
+
+    boolean isCommitted() {
```
mikemccand, line 390: Maybe rename to `isCompleted`?  `committed` is overloaded term -- could try to mean its files got `fsync'`d :)

lucene/core/src/java/org/apache/lucene/index/MergePolicy.java
s1monw: yeah I want to do that anyway

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
s1monw: done

lucene/core/src/java/org/apache/lucene/index/MergePolicy.java
s1monw: done

lucene/core/src/java/org/apache/lucene/index/MergePolicy.java
s1monw: done thanks

s1monw:
@mikemccand @msokolov @msfroh @dsmiley I pushed a new commit to address your comments. thanks you!

PR #1590, opened 2020-06-17
LUCENE-9408: Ensure OneMerge#mergeFinished is only called once

s1monw:
in the case of an exception it's possible that some OneMerge instances
will be closed multiple times. This commit ensures that mergeFinished is
really just called once instead of multiple times.


lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
```diff
@@ -4286,25 +4287,30 @@ private synchronized void mergeFinish(MergePolicy.OneMerge merge) {
...
-        }
-        rld.release(sr);
-        release(rld);
-        if (drop) {
-          readerPool.drop(rld.info);
-        }
-      });
-    } finally {
-      Collections.fill(merge.readers, null);
+    if (merge.isDone() == false) {
```
mikemccand, line 4290: It is a little weird that we don't then have a `merge.setDone()` or something inside this `if` to make the idempotency really clear.  But I understand this code is pre-existingly already hairy so it is not simple to fix that, so progress not perfection!

s1monw: I can call it `hasFinished()` that would make it more clear? WDYT

s1monw: @mikemccand I pushed a new commit to address this

PR #1592, opened 2020-06-18
SOLR-14579 First pass at dismantling Utils

megancarey:
# Description

Eliminating warnings due to static Functions in the Utils class within Solr. 

# Solution

Removed a few Function variables from the Utils class which were adding little value and causing warnings throughout the code. Removed a handful of unused variables, etc. to reduce warnings as well.

# Tests

I ran `ant test` from the solr directory, as all of the changes in this PR are limited to Solr.

# Checklist

### DIFF
--- a/solr/core/src/java/org/apache/solr/cloud/autoscaling/AutoScalingHandler.java
+++ b/solr/core/src/java/org/apache/solr/cloud/autoscaling/AutoScalingHandler.java
@@ -323,8 +323,6 @@ public class AutoScalingHandler extends RequestHandlerBase implements Permission
     String policyName = (String) op.getVal("");
 
     if (op.hasError()) return currentConfig;
-    String eventTypeStr = op.getStr(EVENT);
-    TriggerEventType.valueOf(eventTypeStr.trim().toUpperCase(Locale.ROOT));
 
     Map<String, List<Clause>> policies = currentConfig.getPolicy().getPolicies();
     if (policies == null || !policies.containsKey(policyName)) {
@@ -544,6 +542,8 @@ public class AutoScalingHandler extends RequestHandlerBase implements Permission
 
     if (op.hasError()) return currentConfig;
     String waitForStr = op.getStr(WAIT_FOR, null);
+    String eventTypeStr = op.getStr(EVENT);
+    TriggerEventType.valueOf(eventTypeStr.trim().toUpperCase(Locale.ROOT));
 
     CommandOperation opCopy = new CommandOperation(op.name, Utils.getDeepCopy((Map) op.getCommandData(), 10));
 
--- a/solr/core/src/java/org/apache/solr/cloud/autoscaling/AutoScalingHandler.java
+++ b/solr/core/src/java/org/apache/solr/cloud/autoscaling/AutoScalingHandler.java
@@ -323,6 +323,8 @@ public class AutoScalingHandler extends RequestHandlerBase implements Permission
     String policyName = (String) op.getVal("");
 
     if (op.hasError()) return currentConfig;
+    String eventTypeStr = op.getStr(EVENT);
+    TriggerEventType.valueOf(eventTypeStr.trim().toUpperCase(Locale.ROOT));
 
     Map<String, List<Clause>> policies = currentConfig.getPolicy().getPolicies();
     if (policies == null || !policies.containsKey(policyName)) {
@@ -539,11 +541,8 @@ public class AutoScalingHandler extends RequestHandlerBase implements Permission
                                              AutoScalingConfig currentConfig) throws KeeperException, InterruptedException {
     // we're going to modify the op - use a copy
     String triggerName = op.getStr(NAME);
-    String eventTypeStr = op.getStr(EVENT);
 
     if (op.hasError()) return currentConfig;
-    TriggerEventType eventType = TriggerEventType.valueOf(eventTypeStr.trim().toUpperCase(Locale.ROOT));
-
     String waitForStr = op.getStr(WAIT_FOR, null);
 
     CommandOperation opCopy = new CommandOperation(op.name, Utils.getDeepCopy((Map) op.getCommandData(), 10));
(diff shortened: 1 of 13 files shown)


dsmiley:
+1
Can you see who added these and get their attention here for their opinion?.

megancarey:
cc @noblepaul - It looks like you authored the Utils class. Can you please take a look at this change when you get a chance?

solr/core/src/java/org/apache/solr/cloud/autoscaling/sim/SimClusterStateProvider.java
```diff
@@ -351,10 +340,10 @@ public void simSetClusterState(ClusterState initialState) throws Exception {
         // DocCollection will be created later
         collectionsStatesRef.put(dc.getName(), new CachedCollectionRef(dc.getName(), dc.getZNodeVersion()));
         collProperties.computeIfAbsent(dc.getName(), name -> new ConcurrentHashMap<>()).putAll(dc.getProperties());
-        opDelays.computeIfAbsent(dc.getName(), Utils.NEW_HASHMAP_FUN).putAll(defaultOpDelays);
+        opDelays.computeIfAbsent(dc.getName(), o -> new HashMap<String,Long>()).putAll(defaultOpDelays);
```
noblepaul, line 343: This was added for a purpose.
`o -> new HashMap<String,Long>()` 

this creates a new lambda class. 
OTOH `Utils.NEW_HASHMAP_FUN` is created once and reused everywhere

noblepaul:
-1

don't commit this

noblepaul:
`o -> new HashMap<String,Long>()`

this creates a new lambda class. 
OTOH `Utils.NEW_HASHMAP_FUN` is created once and reused everywhere.

solr/core/src/java/org/apache/solr/cloud/autoscaling/sim/SimClusterStateProvider.java
msokolov: Is this some kind of tight inner loop where we need to be concerned about squeezing out every last gram of garbage? I'm not super familiar with this code, but it rather looks like a one-time initialization that will run once for each collection, so really it's a handful of objects we're talking about I think. Seems like a case of premature optimization, not really worth a warning from the compiler, IMO

megancarey: Should we be using Map.putIfAbsent instead of compute? The value mapped to the key doesn't actually use the value, so I don't believe we need to do computeIfAbsent, and if lambda functions are the issue, putIfAbsent will remove that altogether.

I'll try this out and run tests to see if that's an acceptable option, and report back in a bit.

noblepaul: `putIfAbsent()` is wore than `computeIfAbsent()` 

the former actually create an object even before the method is called. The latter does it only after if the key has no value

solr/core/src/java/org/apache/solr/cloud/autoscaling/AutoScalingHandler.java
```diff
@@ -542,7 +542,7 @@ private AutoScalingConfig handleSetTrigger(SolrQueryRequest req, SolrQueryRespon
     String eventTypeStr = op.getStr(EVENT);
 
     if (op.hasError()) return currentConfig;
-    TriggerEventType eventType = TriggerEventType.valueOf(eventTypeStr.trim().toUpperCase(Locale.ROOT));
+    TriggerEventType.valueOf(eventTypeStr.trim().toUpperCase(Locale.ROOT));
```
madrob, line 545: I think you can remove the whole line.

solr/core/src/java/org/apache/solr/cloud/autoscaling/sim/SimClusterStateProvider.java
```diff
@@ -191,7 +190,7 @@ public DocCollection getColl() throws InterruptedException, IOException {
                 }
                 Map<String, Object> props;
                 synchronized (ri) {
-                  props = new HashMap<>(ri.getVariables());
+                  props = new HashMap<String,Object>(ri.getVariables());
```
madrob, line 193: I don't think we need this?

solr/core/src/java/org/apache/solr/cloud/autoscaling/sim/SimClusterStateProvider.java
```diff
@@ -2293,15 +2274,14 @@ public void simSetShardValue(String collection, String shard, String key, Object
...
-  @SuppressWarnings({"unchecked"})
   public void simSetReplicaValues(String node, Map<String, Map<String, List<ReplicaInfo>>> source, boolean overwrite) {
     List<ReplicaInfo> infos = nodeReplicaMap.get(node);
     if (infos == null) {
       throw new RuntimeException("Node not present: " + node);
     }
     // core_node_name is not unique across collections
-    Map<String, Map<String, ReplicaInfo>> infoMap = new HashMap<>();
-    infos.forEach(ri -> infoMap.computeIfAbsent(ri.getCollection(), Utils.NEW_HASHMAP_FUN).put(ri.getName(), ri));
+    Map<String, Map<String, ReplicaInfo>> infoMap = new HashMap<String, Map<String, ReplicaInfo>>();
```
madrob, line 2283: `<>` is fine here.

solr/core/src/java/org/apache/solr/cloud/autoscaling/AutoScalingHandler.java
megancarey: If the eventTypeStr is an invalid event type, it will throw an exception that there are no matches in TriggerEventType enum. I'll add a comment so the reason for this line is more obvious.

ErickErickson:
See the JIRA, wound up only commenting the static lambdas in Utils.java

uschindler:
This pull request is fine, please apply it. The comments by @noblepaul are not true, there's no instances or classes generated on every call. See https://issues.apache.org/jira/browse/SOLR-14579?focusedCommentId=17188582&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17188582

ErickErickson:
Fixed up as per Uwe's criticisms.

PR #1593, opened 2020-06-18
LUCENE-9409: Check file lengths before creating slices.

jpountz:
This changes terms and points to check the length of the index/data
files before creating slices in these files. A side-effect of this is
that we can no longer verify checksums of the meta file before checking
the length of other files, but this shouldn't be a problem. On the other
hand it helps make sure that we would return a clear exception in case
of truncation instead of a confusing OutOfBoundsException that isn't
clear whether it's due to index corruption or a bug in Lucene.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/codecs/lucene86/Lucene86PointsReader.java
+++ b/lucene/core/src/java/org/apache/lucene/codecs/lucene86/Lucene86PointsReader.java
@@ -72,7 +72,6 @@ public class Lucene86PointsReader extends PointsReader implements Closeable {
           readState.segmentInfo.getId(),
           readState.segmentSuffix);
 
-      long indexLength = -1, dataLength = -1;
       try (ChecksumIndexInput metaIn = readState.directory.openChecksumInput(metaFileName, readState.context)) {
         Throwable priorE = null;
         try {
@@ -82,6 +81,11 @@ public class Lucene86PointsReader extends PointsReader implements Closeable {
               Lucene86PointsFormat.VERSION_CURRENT,
               readState.segmentInfo.getId(),
               readState.segmentSuffix);
+
+          final long indexLength = metaIn.readLong();
+          CodecUtil.retrieveChecksum(indexIn, indexLength);
+          final long dataLength = metaIn.readLong();
+          CodecUtil.retrieveChecksum(dataIn, dataLength);
 
           while (true) {
             int fieldNumber = metaIn.readInt();
@@ -93,18 +97,12 @@ public class Lucene86PointsReader extends PointsReader implements Closeable {
             BKDReader reader = new BKDReader(metaIn, indexIn, dataIn);
             readers.put(fieldNumber, reader);
           }
-          indexLength = metaIn.readLong();
-          dataLength = metaIn.readLong();
         } catch (Throwable t) {
           priorE = t;
         } finally {
           CodecUtil.checkFooter(metaIn, priorE);
         }
       }
-      // At this point, checksums of the meta file have been validated so we
-      // know that indexLength and dataLength are very likely correct.
-      CodecUtil.retrieveChecksum(indexIn, indexLength);
-      CodecUtil.retrieveChecksum(dataIn, dataLength);
       success = true;
     } finally {
       if (success == false) {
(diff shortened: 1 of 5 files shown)


lucene/core/src/java/org/apache/lucene/codecs/lucene86/Lucene86PointsWriter.java
```diff
@@ -68,11 +71,9 @@ public Lucene86PointsWriter(SegmentWriteState writeState, int maxPointsInLeafNod
                                  writeState.segmentInfo.getId(),
                                  writeState.segmentSuffix);
 
-      String metaFileName = IndexFileNames.segmentFileName(writeState.segmentInfo.name,
-          writeState.segmentSuffix,
-          Lucene86PointsFormat.META_EXTENSION);
-      metaOut = writeState.directory.createOutput(metaFileName, writeState.context);
-      CodecUtil.writeIndexHeader(metaOut,
+      tempMetaOut = writeState.directory.createTempOutput(
```
mikemccand, line 74: Why are we switching to a temp file and copying to the real file after closing?  Maybe add a comment explaining?

lucene/core/src/java/org/apache/lucene/codecs/lucene86/Lucene86PointsReader.java
```diff
@@ -93,18 +97,12 @@ public Lucene86PointsReader(SegmentReadState readState) throws IOException {
...
           }
-          indexLength = metaIn.readLong();
-          dataLength = metaIn.readLong();
         } catch (Throwable t) {
           priorE = t;
         } finally {
           CodecUtil.checkFooter(metaIn, priorE);
         }
       }
-      // At this point, checksums of the meta file have been validated so we
```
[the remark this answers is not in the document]
jpountz: we don't lose safety, but in case of a corrupt meta file, it might be slightly more confusing in the sense that the suppressed exception will complain about a truncated index/data file

lucene/core/src/java/org/apache/lucene/codecs/lucene86/Lucene86PointsWriter.java
jpountz: This is because we need to write file lengths of the index/data files before any offsets/lengths of slices into these files. But since these index/data files have not been written yet, we don't know the length yet. So I wrote into a temp file, and only then write the final metadata file that includes first the lengths of the index/data files and then metadata about the KD trees that includes offsets into these index/data files. I'll add a comment.

jpountz: As an alternative, I could buffer the metadata in memory like we do for terms. It will require changing some APIs to replace IndexOutput with DataOutputs but other than that it shouldn't be too hard.

rmuir:
Sorry, I'm against this change. The test is broken. It looks like we are willing to make bad tradeoffs in order to deliver CorruptIndexException and only CorruptIndexException if anything goes wrong. Fix the test instead!

> A side-effect of this is that we can no longer verify checksums of the meta file before checking the length of other files

This is seriously the wrong tradeoff: let's fix the test instead. If we unexpectedly hit EOF, EOFException is the correct exception. If an index is out of bounds, IndexOutOfBoundsException is the correct exception.

jpountz:
I like the CorruptIndexException because it tells me that the problem is that the file got altered after being written, while I would otherwise wonder if there is a bug in Lucene. As an alternative, would it work better for you if we called `retrieveChecksum(IndexInput)` before the try block, and then again with the length (`retrieveChecksum(IndexInput, long)`) after the try block once the checksum of the meta file has been validated?

lucene/core/src/java/org/apache/lucene/codecs/lucene86/Lucene86PointsWriter.java
mikemccand: OK thanks for the explanation @jpountz.  I am OK with using temp files for this ...

jpountz:
I repurposed this PR to instead make the test expect out-of-bounds exceptions. Does it look better to you @rmuir @uschindler ?

uschindler:
I am fine to fix the test. Sure you have to first figure out why the index is out of bounds, and the exact exception may be misleading, but that's actually what's happening here. If you want other exceptions, another fix would be to enforce the IO layer to have a meaningful exception and implement it for all directory implementations.

PR #1594, opened 2020-06-18
Replace DWPT.DocState with simple method parameters

s1monw:
DWPT.DocState had some history value but today in a little bit more
cleaned up DWPT and IndexingChain there is little to no value in having
this class. It also requires explicit cleanup which is not not necessary
anymore.


s1monw:
@dweiss maybe you have a moment to look at this

mikemccand:
+1, thanks for cleaning things up @s1monw.

PR #1600, opened 2020-06-22
SOLR Move "userfiles" stuff from SolrPaths to CoreContainer.

dsmiley:
Convert String and File paths to Path API.

CoreContainer seems to me the proper place for the node to expose key directories at a node level.  It's already doing this.  It looked a bit clumsy to me to have the String constant and extra methods elsewhere for userfiles.  And I'd rather not see SolrDispatchFilter touched whatsoever for this feature, even if it was one line.

Once I moved that, and declared as type Path, I couldn't un-see all the code that did heavy back & forth conversion between String paths and File paths.  Not only is String terrible because it's untyped, but Path is the modern replacement, not File.  I wish we could do forbiddenAPI on File but there's lots of old code.  This userfiles code was only added last year though.

I experimented with using Files.walkTree in findReadableFiles() but backed off because I already extended my scope very far and I'd rather eventually see all of the Files streamed instead of needing to materialize _all_ of them into a list. Also, findReadableFiles is pretty clear and wasn't any clearer when I played with a Files.walkTree considering all the rules, particularly ignoring files that aren't readable.  I wonder how intentional that is... maybe it actually _should_ be an error instead of silently ignoring those files.

@gerlowskija

### DIFF
--- a/solr/core/src/java/org/apache/solr/core/SolrPaths.java
+++ b/solr/core/src/java/org/apache/solr/core/SolrPaths.java
@@ -39,13 +39,6 @@ import org.slf4j.LoggerFactory;
 public final class SolrPaths {
   private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
 
-  /**
-   * Solr allows users to store arbitrary files in a special directory located directly under SOLR_HOME.
-   * <p>
-   * This directory is generally created by each node on startup.  Files located in this directory can then be
-   * manipulated using select Solr features (e.g. streaming expressions).
-   */
-  public static final String USER_FILES_DIRECTORY = "userfiles";
   private static final Set<String> loggedOnce = new ConcurrentSkipListSet<>();
 
   private SolrPaths() {} // don't create this
@@ -92,26 +85,6 @@ public final class SolrPaths {
       logOnceInfo("home_default", "solr home defaulted to '" + home + "' (could not find system property or JNDI)");
     }
     return Paths.get(home).toAbsolutePath().normalize();
-  }
-
-  public static void ensureUserFilesDataDir(Path solrHome) {
-    final Path userFilesPath = getUserFilesPath(solrHome);
-    final File userFilesDirectory = new File(userFilesPath.toString());
-    if (!userFilesDirectory.exists()) {
-      try {
-        final boolean created = userFilesDirectory.mkdir();
-        if (!created) {
-          log.warn("Unable to create [{}] directory in SOLR_HOME [{}].  Features requiring this directory may fail.", USER_FILES_DIRECTORY, solrHome);
-        }
-      } catch (Exception e) {
-        log.warn("Unable to create [{}] directory in SOLR_HOME [{}].  Features requiring this directory may fail.",
-            USER_FILES_DIRECTORY, solrHome, e);
-      }
-    }
-  }
-
-  public static Path getUserFilesPath(Path solrHome) {
-    return Paths.get(solrHome.toAbsolutePath().toString(), USER_FILES_DIRECTORY).toAbsolutePath();
   }
 
   /**
--- a/solr/core/src/java/org/apache/solr/handler/CatStream.java
+++ b/solr/core/src/java/org/apache/solr/handler/CatStream.java
@@ -17,15 +17,14 @@
 
 package org.apache.solr.handler;
 
-import java.io.File;
 import java.io.IOException;
 import java.lang.invoke.MethodHandles;
 import java.nio.file.Files;
-import java.nio.file.Paths;
+import java.nio.file.Path;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Iterator;
 import java.util.List;
+import java.util.stream.Stream;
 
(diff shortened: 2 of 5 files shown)


gerlowskija:
I don't mind this PR at all, I just have some questions.  I don't really understand the aversion to `SolrPaths` and `File`.  More detailed questions below.  Treat this comment as a "+1", as long as you get back to me on the questions below.

> CoreContainer seems to me the proper place for the node to expose key directories at a node level. It's already doing this. It looked a bit clumsy to me to have the String constant and extra methods elsewhere for userfiles.

I don't have a strong opinion on whether this logic is invoked by SolrDispatchFilter or CoreContainer, but I'm surprised at your objection about where the logic lives.  Why is `SolrPaths` a bad place for constants/logic around locating/creating elements of Solr's directory structure?  AFAIK this is exactly what `SolrPaths` is for.  Why clutter up CoreContainer with path logic when we have classes specifically meant to hold that?

> Path is the modern replacement, not File. I wish we could do forbiddenAPI on File but there's lots of old code.
If true, this is a good argument for the work Erick's doing to strengthen our compile-time enforcement of these opinions.  I have no objection to `Path`, but I also had no idea that there was anything wrong with using `File`.  In the diff here, the APIs look relatively identical. What reason is there to prefer Path?

dsmiley:
> Why clutter up CoreContainer with path logic when we have classes specifically meant to hold that?

`CoreContainer.getSolrHome()` and `CoreContainer.getCoreRootDirectory` are two other node level paths attached to CC, so why not this one as well?  I could easily forsee an improvement to "userfiles" where it's location is configurable just as the previous two are.  My only concern, which seems to be shared by you, is that CoreContainer is pretty big as it is.  But I think this path and it's initialization is in scope of its responsibility.  I could forsee reducing the length of CC in other ways.

> but I also had no idea that there was anything wrong with using File

Lucene fully converted in https://issues.apache.org/jira/browse/LUCENE-5945 (see opening description).  See official information from Oracle https://docs.oracle.com/javase/tutorial/essential/io/legacy.html 

The main thing "wrong" with the code I couldn't stand was the excessive and needless use of Strings for fields/parameters/variables that were paths.  Imagine if you looked at someone's calculator app only to see the numeric value held as a String always!  Here, *if* I saw File used pervasively, I wouldn't of bothered doing the conversion to Path because at least File is a respectable choice but not String.

Hey @ErickErickson if you are up for more cleanup tasks, maybe you might be interested in a full conversion of File -> Path cutover.  At least for anything non-private and 9.0.

gerlowskija:
> My only concern, which seems to be shared by you, is that CoreContainer is pretty big as it is. But I think this path and it's initialization is in scope of its responsibility. I could forsee reducing the length of CC in other ways

Yeah, that was my concern for sure.  I don't think `getSolrHome()` and `getCoreRootDirectory` are terribly placed in CoreContainer - but I think there's a good SRP argument to make that they'd be better belong in somewhere more specifically targeted to that functionality like `SolrPaths`.

But I hear your point that there's already precedent for this functionality in CoreContainer, and that ultimately none of this is going to do much to slim down CoreContainer.  So LGTM.

solr/core/src/java/org/apache/solr/core/SolrPaths.java
```diff
@@ -94,26 +87,6 @@ public static Path locateSolrHome() {
     return Paths.get(home).toAbsolutePath().normalize();
   }
 
-  public static void ensureUserFilesDataDir(Path solrHome) {
-    final Path userFilesPath = getUserFilesPath(solrHome);
-    final File userFilesDirectory = new File(userFilesPath.toString());
-    if (!userFilesDirectory.exists()) {
```
dsmiley, line 100: The delta here with Paths API isn't line-for-line; it's much better.  `Files.createDirectories` works if it exists and simply throws if something goes wrong.

solr/core/src/java/org/apache/solr/handler/CatStream.java
```diff
@@ -158,28 +157,20 @@ public Explanation toExplanation(StreamFactory factory) throws IOException {
...
-    for (String crawlRoot : relativePathRoots) {
-      final File crawlRootFile = new File(crawlRoot);
-      if (crawlRootFile.isAbsolute()) {
-        throw new SolrException(SolrException.ErrorCode.BAD_REQUEST,
-            "file/directory to stream must be provided as an absolute path: " + crawlRoot);
-      }
-      if ( crawlRoot.contains("..")) {
+    for (String crawlRootStr : commaDelimitedFilepaths.split(",")) {
+      Path crawlRootPath = chroot.resolve(crawlRootStr).normalize();
+      if (! crawlRootPath.startsWith(chroot)) {
```
dsmiley, line 163: Path "startsWith" API makes what you were trying to do simpler (less code).

joel-bernstein: I hope you're right about this. In my testing it does prevent the use of "..", but there is nothing explicit that I've read that says the change made here will prevent the use of "..". This is not a small issue because of the security ramifications.

This PR also exposes the chroot in the error message which needs to be changed.

PR #1601, opened 2020-06-22
LUCENE-8962: Ensure we don't include fully deleted segments in a commit

s1monw:
IW might drop segments that are merged into a fully deleted segment on the floor
and deletes the newly created files right away. We should not include these segments
in a commit since we can't guarantee valid ref-counts on these files.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
+++ b/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
@@ -3234,15 +3234,23 @@ public class IndexWriter implements Closeable, TwoPhaseCommit, Accountable,
               // corresponding add from an updateDocument) can
               // sneak into the commit point:
               toCommit = segmentInfos.clone();
-
+              pendingCommitChangeCount = changeCount.get();
+              // This protects the segmentInfos we are now going
+              // to commit.  This is important in case, eg, while
+              // we are trying to sync all referenced files, a
+              // merge completes which would otherwise have
+              // removed the files we are now syncing.
+              deleter.incRef(toCommit.files(false));
               if (anyChanges && maxCommitMergeWaitSeconds > 0) {
                 SegmentInfos committingSegmentInfos = toCommit;
                 onCommitMerges = updatePendingMerges(new OneMergeWrappingMergePolicy(config.getMergePolicy(), toWrap ->
                     new MergePolicy.OneMerge(toWrap.segments) {
                       @Override
-                      public void mergeFinished(boolean committed) throws IOException {
+                      public void mergeFinished(boolean committed, boolean segmentDropped) throws IOException {
                         assert Thread.holdsLock(IndexWriter.this);
-                        if (committed && includeInCommit.get()) {
+                        if (segmentDropped == false
+                            && committed
+                            && includeInCommit.get()) {
                           deleter.incRef(info.files());
                           Set<String> mergedSegmentNames = new HashSet<>();
                           for (SegmentCommitInfo sci : segments) {
@@ -3262,8 +3270,8 @@ public class IndexWriter implements Closeable, TwoPhaseCommit, Accountable,
                           committingSegmentInfos.counter = Math.max(committingSegmentInfos.counter, segmentCounter + 1);
                           committingSegmentInfos.applyMergeChanges(applicableMerge, false);
                         }
-                        toWrap.mergeFinished(committed);
-                        super.mergeFinished(committed);
+                        toWrap.mergeFinished(committed, false);
+                        super.mergeFinished(committed, segmentDropped);
                       }
 
                       @Override
@@ -3274,14 +3282,7 @@ public class IndexWriter implements Closeable, TwoPhaseCommit, Accountable,
                 ), MergeTrigger.COMMIT, UNBOUNDED_MAX_MERGE_SEGMENTS);
               }
 
-              pendingCommitChangeCount = changeCount.get();
 
-              // This protects the segmentInfos we are now going
-              // to commit.  This is important in case, eg, while
-              // we are trying to sync all referenced files, a
-              // merge completes which would otherwise have
-              // removed the files we are now syncing.    
-              deleter.incRef(toCommit.files(false));
             }
             success = true;
           } finally {
@@ -4019,7 +4020,7 @@ public class IndexWriter implements Closeable, TwoPhaseCommit, Accountable,
       // Must close before checkpoint, otherwise IFD won't be
       // able to delete the held-open files from the merge
       // readers:
-      closeMergeReaders(merge, false);
+      closeMergeReaders(merge, false, dropSegment);
     }
 
     if (infoStream.isEnabled("IW")) {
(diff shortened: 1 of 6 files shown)


lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
```diff
@@ -3234,15 +3234,23 @@ private long prepareCommitInternal() throws IOException {
               // corresponding add from an updateDocument) can
               // sneak into the commit point:
               toCommit = segmentInfos.clone();
-
+              pendingCommitChangeCount = changeCount.get();
```
s1monw, line 3237: I moved these above the updatePendingMerges. not sure it's a problem but looks better from a readers perspective to incRef before we do anything to it potentially.

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
```diff
@@ -4341,27 +4342,33 @@ private synchronized void mergeFinish(MergePolicy.OneMerge merge) {
...
-        }
-        rld.release(sr);
-        release(rld);
-        if (drop) {
-          readerPool.drop(rld.info);
-        }
-      });
+    try {
+      // first call mergeFinished before we potentially drop the reader and the last reference.
+      merge.mergeFinished(suppressExceptions == false, droppedSegment);
```
s1monw, line 4349: this adds clear semantics when we do what rather than relying on implicit ordering of things in a finalizer.

s1monw:

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
mikemccand: And this also reversed the order of these operations, right?  So now we first call `mergeFinished`, and then all the rld release/drop logic.

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
```diff
@@ -3234,15 +3234,23 @@ private long prepareCommitInternal() throws IOException {
...
               if (anyChanges && maxCommitMergeWaitSeconds > 0) {
                 SegmentInfos committingSegmentInfos = toCommit;
                 onCommitMerges = updatePendingMerges(new OneMergeWrappingMergePolicy(config.getMergePolicy(), toWrap ->
                     new MergePolicy.OneMerge(toWrap.segments) {
                       @Override
-                      public void mergeFinished(boolean committed) throws IOException {
+                      public void mergeFinished(boolean committed, boolean segmentDropped) throws IOException {
                         assert Thread.holdsLock(IndexWriter.this);
-                        if (committed && includeInCommit.get()) {
+                        if (segmentDropped == false
```
mikemccand, line 3251: This is the functional part of the fix, right?  If we will drop the segment (because it is 100% deleted), now we will not attempt to apply it to the commit?

But, does this now mean that the attempted merged (for commit) segments, that are 100% deleted, will remain in the commit?  Shouldn't we also drop them from the commit point?

lucene/core/src/test/org/apache/lucene/index/TestIndexWriterMergePolicy.java
```diff
@@ -396,10 +396,12 @@ protected void merge(MergePolicy.OneMerge merge) throws IOException {
             waitForMerge.await();
             writer.updateDocument(new Term("id", "2"), d2);
             writer.flush();
-            waitForUpdate.countDown();
           } catch (Exception e) {
             throw new AssertionError(e);
+          } finally {
+            waitForUpdate.countDown();
```
mikemccand, line 402: Ahh is this so the test does not hang on an (unexpected) exception in the `try` block?

lucene/core/src/java/org/apache/lucene/index/MergePolicy.java
```diff
@@ -403,6 +409,22 @@ boolean isDone() {
     Optional<Boolean> hasCompletedSuccessfully() {
       return Optional.ofNullable(mergeCompleted.getNow(null));
     }
+
+    void onMergeCommit() {
+    }
+
+    void setMergeReaders(IOContext mergeContext, ReaderPool readerPool) throws IOException {
+      for (final SegmentCommitInfo info : segments) {
```
mikemccand, line 417: Maybe `assert` that `readers` is empty at the start here?

lucene/core/src/java/org/apache/lucene/index/MergePolicy.java
```diff
@@ -403,6 +409,22 @@ boolean isDone() {
...
+        // Hold onto the "live" reader; we will use this to
+        // commit merged deletes
+        final ReadersAndUpdates rld = readerPool.get(info, true);
```
mikemccand, line 420: This used to be `getPooledInstance`, but I think this is functionally identical.

lucene/core/src/test/org/apache/lucene/index/TestIndexWriterMergePolicy.java
```diff
@@ -396,10 +396,12 @@ protected void merge(MergePolicy.OneMerge merge) throws IOException {
             waitForMerge.await();
             writer.updateDocument(new Term("id", "2"), d2);
```
mikemccand, line 397: This test fully deleted the 2nd segment (since it only had `d2`, which we are deleting here).  Maybe also add a similar case that leaves one not deleted document in the segment too?  I think we optimize 100% deleted merge source segments too, so this should tickle slightly different path.

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
```diff
@@ -3907,6 +3927,7 @@ private static void carryOverHardDeletes(ReadersAndUpdates mergedReadersAndUpdat
   @SuppressWarnings("try")
   private synchronized boolean commitMerge(MergePolicy.OneMerge merge, MergeState mergeState) throws IOException {
```
mikemccand, line 3928: We should really rename this method to not use the word `commit` since that already means something (different!) in `IndexWriter`.  Maybe `finishMerge`, or `applyMergedSegments` or something.  Naming is the hardest part!  We don't have to do that in this PR...

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
```diff
@@ -3265,24 +3269,40 @@ public void mergeFinished(boolean committed, boolean segmentDropped) throws IOEx
...
                           committingSegmentInfos.counter = Math.max(committingSegmentInfos.counter, segmentCounter + 1);
                           committingSegmentInfos.applyMergeChanges(applicableMerge, false);
                         }
                         toWrap.mergeFinished(committed, false);
                         super.mergeFinished(committed, segmentDropped);
                       }
 
+                      @Override
+                      void onMergeCommit() {
+                        origInfo = this.info.clone();
```
mikemccand, line 3283: Whoa!  This is the functional part of the fix?  Because we snapshot the `SegmentCommitInfo` *before* carrying over deletes that happened while the merge was running?  Very cool.

But, how do we ensure that the prior live docs (without concurrent deletions carried over) are written for this merged segment, and matching `origInfo`'s `delGen`?  Normally `commitMergedDeletesAndUpdates` would do that, except it is carrying over the concurrent deletions, which we need to avoid.  Oh, wait, sorry: such deletions are of course merged away!  So, we want 100% live docs in the newly merged segment.  Tricky.

We probably also need a test case confirm that (concurrent) DV updates are also transactional?  I think this same approach should work for DV updates too?  We will keep only the DV updates that got merged away, and any new ones that we carry over will not be visible in the merged segment because you are cloning the `SegmentCommitInfo` before `commitMerge` runs?

lucene/core/src/test/org/apache/lucene/index/TestIndexWriterMergePolicy.java
s1monw: yes

s1monw:
I pushed all the commits and fixes for your comments to this branch https://github.com/apache/lucene-solr/tree/jira/lucene-8962

PR #1602, opened 2020-06-22
SOLR-14582: Expose IWC.setMaxCommitMergeWaitMillis in Solr's index config

tflobbe:
LUCENE-8962 added the ability to merge segments synchronously on commit. This isn't done by default and the default MergePolicy won't do it, but custom merge policies can take advantage of this. Solr allows plugging in custom merge policies, so if someone wants to make use of this feature they could, however, they need to set IndexWriterConfig.maxCommitMergeWaitSeconds to something greater than 0.
Since this is an expert feature, I plan to document it only in javadoc and not the ref guide.

### DIFF
--- a/solr/core/src/java/org/apache/solr/update/SolrIndexConfig.java
+++ b/solr/core/src/java/org/apache/solr/update/SolrIndexConfig.java
@@ -68,6 +68,13 @@ public class SolrIndexConfig implements MapSerializable {
 
   public final double ramBufferSizeMB;
   public final int ramPerThreadHardLimitMB;
+  /**
+   * When using a custom merge policy that allows triggering synchronous merges on commit
+   * (see {@link MergePolicy#findFullFlushMerges(org.apache.lucene.index.MergeTrigger, org.apache.lucene.index.SegmentInfos, org.apache.lucene.index.MergePolicy.MergeContext)}),
+   * a timeout (in seconds) can be set for those merges to finish. Use {@code <maxCommitMergeWaitSeconds>10</maxCommitMergeWaitSeconds>} in the {@code <indexConfig>} section.
+   * See {@link IndexWriterConfig#setMaxCommitMergeWaitSeconds(long)}.
+   */
+  public final int maxCommitMergeWaitSeconds;
 
   public final int writeLockTimeout;
   public final String lockType;
@@ -87,6 +94,7 @@ public class SolrIndexConfig implements MapSerializable {
     maxBufferedDocs = -1;
     ramBufferSizeMB = 100;
     ramPerThreadHardLimitMB = -1;
+    maxCommitMergeWaitSeconds = -1;
     writeLockTimeout = -1;
     lockType = DirectoryFactory.LOCK_TYPE_NATIVE;
     mergePolicyFactoryInfo = null;
@@ -129,8 +137,9 @@ public class SolrIndexConfig implements MapSerializable {
         true);
 
     useCompoundFile = solrConfig.getBool(prefix+"/useCompoundFile", def.useCompoundFile);
-    maxBufferedDocs=solrConfig.getInt(prefix+"/maxBufferedDocs",def.maxBufferedDocs);
+    maxBufferedDocs = solrConfig.getInt(prefix+"/maxBufferedDocs", def.maxBufferedDocs);
     ramBufferSizeMB = solrConfig.getDouble(prefix+"/ramBufferSizeMB", def.ramBufferSizeMB);
+    maxCommitMergeWaitSeconds = solrConfig.getInt(prefix+"/maxCommitMergeWaitSeconds", def.maxCommitMergeWaitSeconds);
 
     // how do we validate the value??
     ramPerThreadHardLimitMB = solrConfig.getInt(prefix+"/ramPerThreadHardLimitMB", def.ramPerThreadHardLimitMB);
@@ -185,6 +194,7 @@ public class SolrIndexConfig implements MapSerializable {
         "maxBufferedDocs", maxBufferedDocs,
         "ramBufferSizeMB", ramBufferSizeMB,
         "ramPerThreadHardLimitMB", ramPerThreadHardLimitMB,
+        "maxCommitMergeWaitSeconds", maxCommitMergeWaitSeconds,
         "writeLockTimeout", writeLockTimeout,
         "lockType", lockType,
         "infoStreamEnabled", infoStream != InfoStream.NO_OUTPUT);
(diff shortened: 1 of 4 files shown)


solr/core/src/test/org/apache/solr/update/SolrIndexConfigTest.java
```diff
@@ -208,4 +210,16 @@ public void testToMap() throws Exception {
...
+  
+  public void testMaxCommitMergeWaitSeconds() throws Exception {
+    SolrConfig sc = new SolrConfig(TEST_PATH().resolve("collection1"), "solrconfig-test-misc.xml");
+    assertEquals(-1, sc.indexConfig.maxCommitMergeWaitMillis);
+    assertEquals(IndexWriterConfig.DEFAULT_MAX_COMMIT_MERGE_WAIT_MILLIS, sc.indexConfig.toIndexWriterConfig(h.getCore()).getMaxCommitMergeWaitMillis());
+    System.setProperty("solr.tests.maxCommitMergeWait", "10");
+    sc = new SolrConfig(TEST_PATH().resolve("collection1"), "solrconfig-test-misc.xml");
+    assertEquals(10, sc.indexConfig.maxCommitMergeWaitMillis);
+    assertEquals(10, sc.indexConfig.toIndexWriterConfig(h.getCore()).getMaxCommitMergeWaitMillis());
+    System.clearProperty("solr.tests.maxCommitMergeWait");
```
madrob, line 222: I don't think we need to clear this, the test runner takes care of it and will reset it to whatever it was previously if it had been set?

solr/core/src/java/org/apache/solr/update/SolrIndexConfig.java
```diff
@@ -87,6 +100,7 @@ private SolrIndexConfig(SolrConfig solrConfig) {
     maxBufferedDocs = -1;
     ramBufferSizeMB = 100;
     ramPerThreadHardLimitMB = -1;
+    maxCommitMergeWaitMillis = -1;
```
dsmiley, line 103: Lets use `org.apache.lucene.index.IndexWriterConfig#DEFAULT_MAX_COMMIT_MERGE_WAIT_MILLIS`.  In the future, this constant might change, and that's good.

solr/core/src/java/org/apache/solr/update/SolrIndexConfig.java
```diff
@@ -68,6 +68,19 @@
 
   public final double ramBufferSizeMB;
   public final int ramPerThreadHardLimitMB;
+  /**
```
dsmiley, line 71: I appreciate documentation but _redundant_ documentation -- not so much.  Can't you do a @see or @link to one place -- `org.apache.lucene.index.IndexWriterConfig#setMaxCommitMergeWaitMillis`

solr/core/src/java/org/apache/solr/update/SolrIndexConfig.java
```diff
@@ -129,8 +143,9 @@ public SolrIndexConfig(SolrConfig solrConfig, String prefix, SolrIndexConfig def
         true);
 
     useCompoundFile = solrConfig.getBool(prefix+"/useCompoundFile", def.useCompoundFile);
-    maxBufferedDocs=solrConfig.getInt(prefix+"/maxBufferedDocs",def.maxBufferedDocs);
+    maxBufferedDocs = solrConfig.getInt(prefix+"/maxBufferedDocs", def.maxBufferedDocs);
     ramBufferSizeMB = solrConfig.getDouble(prefix+"/ramBufferSizeMB", def.ramBufferSizeMB);
+    maxCommitMergeWaitMillis = solrConfig.getInt(prefix+"/maxCommitMergeWait", def.maxCommitMergeWaitMillis);
```
dsmiley, line 148: I noticed you omitted a "Millis" setting.  I think either you should add this for clarity, or for consistency with some other times I see in solrconfig (e.g. maxTime), use suffix "Time".

solr/core/src/java/org/apache/solr/update/SolrIndexConfig.java
tflobbe: I don't think that's a good idea. "-1" from the Solr perspective means "don't set the value".  If someone has a custom merge policy where they have a different value set by code, we would be changing it to `org.apache.lucene.index.IndexWriterConfig#DEFAULT_MAX_COMMIT_MERGE_WAIT_MILLIS` without them having set anything on `solrconfig.xml`. 
> In the future, this constant might change, and that's good.

That's fine, since we aren't calling `iwc.setMaxCommitMergeWaitMillis(...)` in the default case, we'll be taking the new default for all cores where no alternative configuration has been provided.

solr/core/src/java/org/apache/solr/update/SolrIndexConfig.java
tflobbe: I had it with `MIllis` and I removed that because I didn't see anything else in the solrconfig that included the unit for time (it's always milliseconds). You are suggesting `maxCommitMergeWait` or `maxCommitMergeTime`?

dsmiley: The "Wait" part is important too too.  "maxCommitMergeWaitTime" is my preference

solr/core/src/test/org/apache/solr/update/SolrIndexConfigTest.java
dsmiley: Please verify if this is so before removing.  Perhaps there may be a difference in behavior between gradle & IntelliJ; so try both.  I thought clearing wasn't necessary as well but I recall hearing from someone (@dweiss ?) that the auto-clearing wasn't working.

tflobbe: Yes, I checked. I believe the cleanup is after the class. I discussed with @madrob, and I'll move the cleanup to a `tearDown()` method

solr/core/src/java/org/apache/solr/update/SolrIndexConfig.java
tflobbe: I didn't want to drop the link/see by itself, so I added a single line introducing the functionality. After that I added specifics about Solr (solrconfig configuration and the warning setting). I honestly don't see the problem here.

solr/core/src/java/org/apache/solr/update/SolrIndexConfig.java
dsmiley: ok, good point.

tflobbe:
@dsmiley, looks good? should I merge?

gus-asf:
I think this breaks the build... I'm seeing a failure in org.apache.solr.core.TestConfig#testDefaults which doesn't seem to account for maxCommitMergeWaitTime also lots of failures on http://fucit.org/solr-jenkins-reports/failure-report.html in last 24h for this test

PR #1606, opened 2020-06-23
SOLR-14588: Implement Circuit Breakers

atris:
This commit consists of two parts: initial circuit breakers infrastructure and real JVM memory based circuit breaker which monitors incoming search requests and rejects them with SERVICE_TOO_BUSY error if the defined threshold is breached, thus giving headroom to existing indexing and search requests to complete.


solr/core/src/test-files/solr/collection1/conf/solrconfig-cache-enable-disable.xml
```diff
@@ -70,6 +70,10 @@
 
     <queryResultWindowSize>10</queryResultWindowSize>
 
+    <useCircuitBreakers>false</useCircuitBreakers>
+
+    <memoryCircuitBreakerThreshold>100</memoryCircuitBreakerThreshold>
```
madrob, line 75: Does this have to be set when useCircuitBreakers is false? I would like our configs to be more tolerant.

solr/core/src/java/org/apache/solr/util/circuitbreaker/MemoryCircuitBreaker.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    seenMemory = calculateLiveMemoryUsage();
+
+    return (seenMemory >= allowedMemory);
+  }
+
+  @Override
+  public String printDebugInfo() {
+    return "seen memory " + seenMemory + " allowed memory " + allowedMemory;
```
madrob, line 59: this would be better as seenMemory=X allowedMemory=Y - a little bit less human readable but a lot easier to grep for or search in something like Splunk.

solr/core/src/java/org/apache/solr/core/SolrConfig.java
```diff
@@ -224,6 +224,9 @@ private SolrConfig(SolrResourceLoader loader, String name, boolean isConfigsetTr
     queryResultWindowSize = Math.max(1, getInt("query/queryResultWindowSize", 1));
     queryResultMaxDocsCached = getInt("query/queryResultMaxDocsCached", Integer.MAX_VALUE);
     enableLazyFieldLoading = getBool("query/enableLazyFieldLoading", false);
+
+    useCircuitBreakers = getBool("query/useCircuitBreakers", false);
+    memoryCircuitBreakerThreshold = getInt("query/memoryCircuitBreakerThreshold", 100);
```
madrob, line 229: Should we validate that this is between 0 and 100?

solr/core/src/java/org/apache/solr/util/circuitbreaker/MemoryCircuitBreaker.java
```diff
@@ -0,0 +1,88 @@
...
+
+    if (currentMaxHeap <= 0) {
+      return Long.MIN_VALUE;
+    }
+
+    double thresholdInFraction = (double) thresholdValueInPercentage / 100;
+    double actualLimit = currentMaxHeap * thresholdInFraction;
+
+    if (actualLimit <= 0) {
+      throw new IllegalStateException("Memory limit cannot be less than or equal to zero");
```
madrob, line 74: Is this checking for overflow? How does the error condition occur otherwise?

solr/core/src/java/org/apache/solr/util/circuitbreaker/MemoryCircuitBreaker.java
```diff
@@ -0,0 +1,88 @@
...
+  private double getCurrentMemoryThreshold() {
+    int thresholdValueInPercentage = solrCore.getSolrConfig().memoryCircuitBreakerThreshold;
+    long currentMaxHeap = MEMORY_MX_BEAN.getHeapMemoryUsage().getMax();
+
+    if (currentMaxHeap <= 0) {
+      return Long.MIN_VALUE;
+    }
+
+    double thresholdInFraction = (double) thresholdValueInPercentage / 100;
+    double actualLimit = currentMaxHeap * thresholdInFraction;
```
madrob, line 71: I think this makes more sense to be a long. Returning fractional bytes is meaningless.

solr/core/src/test/org/apache/solr/util/TestCircuitBreaker.java
```diff
@@ -0,0 +1,191 @@
...
+    initCore("solrconfig-memory-circuitbreaker.xml", "schema.xml");
+    for (int i = 0 ; i < NUM_DOCS ; i ++) {
+      assertU(adoc("name", "john smith", "id", "1"));
+      assertU(adoc("name", "johathon smith", "id", "2"));
+      assertU(adoc("name", "john percival smith", "id", "3"));
+      assertU(commit());
+      assertU(optimize());
+
+      //commit inside the loop to get multiple segments to make search as realistic as possible
+      assertU(commit());
```
madrob, line 62: commit-optimize-commit in a loop won't get you multiple segments

solr/core/src/test/org/apache/solr/util/TestCircuitBreaker.java
```diff
@@ -0,0 +1,191 @@
...
+    }
+  }
+
+  @Override
+  public void tearDown() throws Exception {
+    executor = null;
```
madrob, line 68: Probably want to also call executor.shutdown in addition to null?

solr/core/src/test/org/apache/solr/util/TestCircuitBreaker.java
```diff
@@ -0,0 +1,191 @@
...
+    args.put(CommonParams.FL, "id");
+
+    AtomicInteger failureCount = new AtomicInteger();
+
+    CircuitBreaker circuitBreaker = new BuildingUpMemoryPressureCircuitBreaker(h.getCore());
+
+    h.getCore().getCircuitBreakerManager().registerCircuitBreaker(CircuitBreakerType.MEMORY, circuitBreaker);
+
+    for (int i = 0; i < 5; i++) {
+      System.out.println("i is " + i);
```
madrob, line 122: Use logger instead of system.out

solr/core/src/test/org/apache/solr/util/TestCircuitBreaker.java
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    executor.shutdown();
+    try {
+      executor.awaitTermination(Long.MAX_VALUE, TimeUnit.NANOSECONDS);
+    } catch (InterruptedException e) {
+      throw new RuntimeException(e.getMessage());
+    }
+
+    assert failureCount.get() == 1;
```
madrob, line 141: use unit asserts instead of java asserts please

solr/core/src/java/org/apache/solr/util/circuitbreaker/MemoryCircuitBreaker.java
```diff
@@ -0,0 +1,88 @@
...
+    return actualLimit;
+  }
+
+  /**
+   * Calculate the live memory usage for the system. This method has package visibility
+   * to allow using for testing
+   * @return Memory usage in bytes
+   */
+  protected long calculateLiveMemoryUsage() {
+    return MEMORY_MX_BEAN.getHeapMemoryUsage().getUsed();
```
madrob, line 86: Is this expensive to get? If we're doing it for every query, I'd want to make sure that it is an efficient call.

solr/core/src/java/org/apache/solr/util/circuitbreaker/MemoryCircuitBreaker.java
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  @Override
+  public String printDebugInfo() {
+    return "seen memory " + seenMemory + " allowed memory " + allowedMemory;
+  }
+
+  private double getCurrentMemoryThreshold() {
+    int thresholdValueInPercentage = solrCore.getSolrConfig().memoryCircuitBreakerThreshold;
+    long currentMaxHeap = MEMORY_MX_BEAN.getHeapMemoryUsage().getMax();
```
madrob, line 64: This should be constant, right? Get it once at constructor?

madrob:
This has been something that I've wanted for a long time, so I'm really excited to see somebody picking up the work! Thank you @atris !

solr/core/src/test-files/solr/collection1/conf/solrconfig-cache-enable-disable.xml
atris: I dont think so, since it is checked when circuit breakers are enabled. I put this config here for completion.

solr/core/src/java/org/apache/solr/core/SolrConfig.java
atris: Good catch, thanks!

solr/core/src/test/org/apache/solr/util/TestCircuitBreaker.java
atris: This should be shutdown in the test itself. I have added a check to check if it is shut down before making the reference null.

solr/core/src/test/org/apache/solr/util/TestCircuitBreaker.java
atris: Stray debugging output, removed, thanks

solr/core/src/java/org/apache/solr/util/circuitbreaker/MemoryCircuitBreaker.java
atris: Good point, thanks

solr/core/src/java/org/apache/solr/util/circuitbreaker/MemoryCircuitBreaker.java
atris: I did not see any regression -- if we note anything otherwise, we can probably cache these values with a TTL. I would be surprised if it causes anything noticeable, though.

atris:
To provide some context, I ran this against large facets with the memory circuit breaker tripping in the right place.

@madrob Thanks for reviewing, I have raised another iteration fixing your comments. Please see and let me know your thoughts and comments.

solr/core/src/java/org/apache/solr/util/circuitbreaker/MemoryCircuitBreaker.java
madrob: This article claims that each call is about 100ns, https://www.inoneo.com/en/blog/8/java/monitoring-jvm-memory-usage-inside-a-java-application and suggest using a listener instead. I'm not sure how true this is for more modern JVMs anyway.

If you look into drop wizard metrics that we're reporting in JMX, MemoryUsageGuageSet reports memory as the combination of heap and non-heap. I'm not sure we need to consider non-heap here, but at least a comment highlighting that would be useful.

solr/core/src/java/org/apache/solr/core/SolrConfig.java
```diff
@@ -224,6 +224,13 @@ private SolrConfig(SolrResourceLoader loader, String name, boolean isConfigsetTr
...
+
+    if (memoryCircuitBreakerThreshold > 100 || memoryCircuitBreakerThreshold < 0) {
```
madrob, line 231: We only want to do this validation if circuitBreakers are enabled, right?

solr/core/src/java/org/apache/solr/util/circuitbreaker/MemoryCircuitBreaker.java
```diff
@@ -0,0 +1,89 @@
...
+  // overhead of calculating the condition parameters but can result in false positives.
+  @Override
+  public boolean isCircuitBreakerGauntletTripped() {
+    if (!isCircuitBreakerEnabled()) {
+      return false;
+    }
+
+    allowedMemory = getCurrentMemoryThreshold();
+
+    if (allowedMemory < 0) {
```
madrob, line 49: This will never be true.

solr/core/src/java/org/apache/solr/core/SolrConfig.java
atris: I was being overtly cautious here -- it seems wrong if the parameter is incorrectly specified regardless of whether it is used or not?

solr/core/src/java/org/apache/solr/util/circuitbreaker/MemoryCircuitBreaker.java
atris: Another of my overtly cautious safety nets :) Removed, thanks.

solr/core/src/java/org/apache/solr/util/circuitbreaker/MemoryCircuitBreaker.java
atris: Yes, this is an overflow check.

solr/core/src/test/org/apache/solr/util/TestCircuitBreaker.java
atris: Removed the commit optimize.

solr/core/src/java/org/apache/solr/util/circuitbreaker/MemoryCircuitBreaker.java
atris: Yes, I considered MemoryUsageGaugeSet and did not use it for the precise reason that you described. To be honest, the listener based implementation is harder to read than the current, and I was not sure if it is worth going that route for shaving off < 100 ns per query. However, if you have concerns or if we ever hear slowness reports due to this feature, I am happy to take that route. For now, I have added the comment as suggested, thanks.

atris:
@madrob Updated, please see and share your thoughts.

solr/core/src/java/org/apache/solr/core/SolrConfig.java
anshumg: Perhaps you want to change this to early saying "Circuit breaker disabled, but incorrectly configured" or <current message> conditionally?

solr/core/src/java/org/apache/solr/core/SolrCore.java
```diff
@@ -1164,6 +1171,16 @@ private SolrCoreMetricManager initCoreMetricManager(SolrConfig config) {
     return coreMetricManager;
   }
 
+  private CircuitBreakerManager initCircuitBreakerManager() {
+    CircuitBreakerManager circuitBreakerManager = new CircuitBreakerManager();
+
+    // Install the default circuit breakers
+    CircuitBreaker memoryCircuitBreaker = new MemoryCircuitBreaker(this);
+    circuitBreakerManager.registerCircuitBreaker(CircuitBreakerType.MEMORY, memoryCircuitBreaker);
```
anshumg, line 1179: I was thinking of moving this outside of SolrCore, so custom Circuit breakers could be implemented and plugged without any change to SolrCore? What do you think?

solr/core/src/java/org/apache/solr/core/SolrConfig.java
atris: I have changed the check to be only performed when circuit breakers are enabled.

solr/core/src/java/org/apache/solr/core/SolrCore.java
atris: Agreed. I have moved this to a method in CircuitBreakerManager that can be delegated to for handling this scenario.

atris:
@madrob @anshumg Thanks for reviewing -- I have updated per comments. Please see and let me know if it looks fine.

solr/core/src/java/org/apache/solr/util/circuitbreaker/MemoryCircuitBreaker.java
atris: @madrob Just verified that Elasticsearch also uses getHeapMemoryUsage() to trigger similar checks per call. Also, I am not sure if NotificationListener will allow us to be real time since it triggers a notification when the threshold is breached. but I could not find a way to know when the threshold has gone back to normal. I am proposing sticking to this for now, if that sounds fine?

madrob: Yea, add some comments to this effect to the code so that if people run into issues with it they can more easily figure out what's up without having to find the comments from this PR.

solr/core/src/java/org/apache/solr/util/circuitbreaker/MemoryCircuitBreaker.java
```diff
@@ -0,0 +1,87 @@
...
+
+    allowedMemory = getCurrentMemoryThreshold();
+
+    seenMemory = calculateLiveMemoryUsage();
+
+    return (seenMemory >= allowedMemory);
+  }
+
+  @Override
+  public String printDebugInfo() {
```
madrob, line 55: Is this thread safe? If there are multiple queries happening, concurrent with GC, would these values be stale?

solr/core/src/java/org/apache/solr/util/circuitbreaker/MemoryCircuitBreaker.java
```diff
@@ -0,0 +1,87 @@
...
+    return "seen memory=" + seenMemory + " allowed memory=" + allowedMemory;
+  }
+
+  private long getCurrentMemoryThreshold() {
+    int thresholdValueInPercentage = solrCore.getSolrConfig().memoryCircuitBreakerThreshold;
+
+    if (currentMaxHeap <= 0) {
+      return Long.MIN_VALUE;
```
madrob, line 63: Does this make sense? It indicates an error condition in the JVM, right? Can move this check to the constructor and throw an Exception since we'd effectively be blocking all queries at that point anyway?

solr/core/src/test/org/apache/solr/util/TestCircuitBreaker.java
```diff
@@ -0,0 +1,184 @@
...
+
+    h.getCore().getCircuitBreakerManager().registerCircuitBreaker(CircuitBreakerType.MEMORY, circuitBreaker);
+
+    expectThrows(SolrException.class, () -> {
+      h.query(req("name:\"john smith\""));
+    });
+  }
+
+  public void testBuildingMemoryPressure() {
+    ExecutorService executor = ExecutorUtil.newMDCAwareCachedThreadPool(
```
madrob, line 102: Use a try-finally to make sure this executor is shut down.

solr/core/src/java/org/apache/solr/util/circuitbreaker/CircuitBreakerManager.java
```diff
@@ -0,0 +1,104 @@
...
+ * of whether a circuit breaker has tripped or not.
+ *
+ * There are two typical ways of using this class's instance:
+ * 1. Check if any circuit breaker has triggered -- and know which circuit breaker has triggered.
+ * 2. Get an instance of a specific circuit breaker and perform checks.
+ *
+ * It is a good practise to register new circuit breakers here if you want them checked for every
+ * request.
+ */
+public class CircuitBreakerManager {
```
madrob, line 36: A question about the design of this class... We have methods for registering a custom circuit breaker, and some notion of default circuit breakers, but I don't see an easy way to add additional ones. Ideally this would be through the XML or through some kind of schema API or could be very powerful combined with the Package API. I don't have a full idea of what this needs to look like, and this is fine as is to commit now, but I'd really like us to come back to this and develop the idea further. Possibly as a SIP?

solr/core/src/java/org/apache/solr/handler/component/SearchHandler.java
```diff
@@ -289,6 +295,19 @@ public void handleRequestBody(SolrQueryRequest req, SolrQueryResponse rsp) throw
       rb.requestInfo.setResponseBuilder(rb);
     }
 
+    //TODO: Should this be for indexing requests as well?
```
madrob, line 298: I think we would want to have two separate controls for this.

solr/core/src/java/org/apache/solr/handler/component/SearchHandler.java
```diff
@@ -289,6 +295,19 @@ public void handleRequestBody(SolrQueryRequest req, SolrQueryResponse rsp) throw
...
+    CircuitBreakerManager circuitBreakerManager = req.getCore().getCircuitBreakerManager();
+    Map<CircuitBreakerType, CircuitBreaker> trippedCircuitBreakers = circuitBreakerManager.checkAllCircuitBreakers();
```
madrob, line 300: Do we want to time this and include that in the response (useful for both passed and failed breakers)

solr/core/src/java/org/apache/solr/util/circuitbreaker/MemoryCircuitBreaker.java
atris: Added a comment, thanks

solr/core/src/java/org/apache/solr/util/circuitbreaker/CircuitBreakerManager.java
```diff
@@ -0,0 +1,104 @@
...
+
+  private final Map<CircuitBreakerType, CircuitBreaker> circuitBreakerMap = new HashMap<>();
+
+  // Allows replacing of existing circuit breaker
+  public void registerCircuitBreaker(CircuitBreakerType circuitBreakerType, CircuitBreaker circuitBreaker) {
+    assert circuitBreakerType != null && circuitBreaker != null;
```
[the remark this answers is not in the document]
atris: A paranoid security check. Removed, thanks

atris:
> How does this work for distributed requests? When the coordinator node is under memory pressure? When one of the fan out nodes in under memory pressure?

I believe the behaviour would be same -- if the node is not able to handle the request, it will error back with the transient error code. If it is coordinator node, the result will be sent back to the user and if it is a fan out node, the coordinator needs to deal with the regular way of a node erroring out?

solr/core/src/java/org/apache/solr/util/circuitbreaker/MemoryCircuitBreaker.java
atris: My idea was to avoid blocking errors as much as possible but I agree with you. Fixed, thanks

solr/core/src/java/org/apache/solr/handler/component/SearchHandler.java
atris: ++

solr/core/src/java/org/apache/solr/util/circuitbreaker/CircuitBreakerManager.java
atris: That is a good point. I was thinking of a schema API since I dont trust XMLs for class level detailing much. Your idea to combine with the package API is a great one. I will open a JIRA once this is committed and follow up with a SIP. For now, adding a comment for the same.

solr/core/src/java/org/apache/solr/handler/component/SearchHandler.java
atris: Not sure if I understood. Would you mean the epoch timestamp when the circuit breaker triggered?

madrob: I mean in response we send timing info with `debug=true`(see `RTimerTree subt = timer.sub( "prepare" );`), maybe this is worthwhile to include there as well since otherwise the times may not add up

solr/core/src/java/org/apache/solr/util/circuitbreaker/MemoryCircuitBreaker.java
atris: Yes, these values can be stale. I was debating if we need a synchronization here for exact values. Would it make sense to move these to thread local state inside the class instead?

atris: Moved to thread local

solr/core/src/java/org/apache/solr/handler/component/SearchHandler.java
atris: Added, thanks

atris:
@madrob Updated, please see and let me know your thoughts and comments.

solr/core/src/java/org/apache/solr/util/circuitbreaker/MemoryCircuitBreaker.java
```diff
@@ -44,25 +48,20 @@ public boolean isCircuitBreakerGauntletTripped() {
...
+    seenMemory.set(calculateLiveMemoryUsage());
 
-    return (seenMemory >= allowedMemory);
+    return (seenMemory.get() >= allowedMemory.get());
   }
 
   @Override
   public String printDebugInfo() {
-    return "seen memory=" + seenMemory + " allowed memory=" + allowedMemory;
+    return "seenMemory=" + seenMemory + " allowedMemory=" + allowedMemory;
```
madrob, line 60: Need `.get()` here.

solr/core/src/java/org/apache/solr/handler/component/SearchHandler.java
```diff
@@ -289,6 +295,23 @@ public void handleRequestBody(SolrQueryRequest req, SolrQueryResponse rsp) throw
...
+    if (trippedCircuitBreakers != null) {
+      final RTimerTree timer = rb.isDebug() ? req.getRequestTimer() : null;
+
+      if (timer != null) {
+        RTimerTree subt = timer.sub("circuitbreaker");
+        rb.setTimer(subt.sub("circuitbreaker"));
+      }
+      String errorMessage = CircuitBreakerManager.constructFinalErrorMessageString(trippedCircuitBreakers);
+      rsp.add(STATUS, FAILURE);
+      rsp.setException(new SolrException(SolrException.ErrorCode.SERVICE_UNAVAILABLE, "Circuit Breakers tripped " + errorMessage));
```
madrob, line 311: I'm not sure this is the right error code. A 5xx code usually indicates a server error - and I'm not sure how we effectively convey to clients that this error is something that is ok to retry. They might log the message, but retry logic will typically look at the code returned as a first branch in the decision tree. Need to think about this and maybe look at some examples.

solr/core/src/java/org/apache/solr/core/SolrConfig.java
```diff
@@ -224,6 +224,11 @@ private SolrConfig(SolrResourceLoader loader, String name, boolean isConfigsetTr
     queryResultWindowSize = Math.max(1, getInt("query/queryResultWindowSize", 1));
     queryResultMaxDocsCached = getInt("query/queryResultMaxDocsCached", Integer.MAX_VALUE);
     enableLazyFieldLoading = getBool("query/enableLazyFieldLoading", false);
+
+    useCircuitBreakers = getBool("query/useCircuitBreakers", false);
```
madrob, line 228: I'd like for all of this to be dynamically configurable at some point, but it doesn't have to be in this PR. Can add it to the future SIP or create a separate JIRA for it, as you think would be appropriate.

solr/core/src/java/org/apache/solr/handler/component/SearchHandler.java
```diff
@@ -289,6 +295,23 @@ public void handleRequestBody(SolrQueryRequest req, SolrQueryResponse rsp) throw
       rb.requestInfo.setResponseBuilder(rb);
     }
 
+    //TODO: Should this be for indexing requests as well?
+    CircuitBreakerManager circuitBreakerManager = req.getCore().getCircuitBreakerManager();
+    Map<CircuitBreakerType, CircuitBreaker> trippedCircuitBreakers = circuitBreakerManager.checkAllCircuitBreakers();
+
+    if (trippedCircuitBreakers != null) {
+      final RTimerTree timer = rb.isDebug() ? req.getRequestTimer() : null;
```
madrob, line 303: I'd combine this with the later call to req.getRequestTimer so that we're not doing that twice. You're also checking this before we call rb.setDebug(), so it probably is always false at this point.

We should be timing the circuitBreakerCheck as well (Lines 299-300).

solr/core/src/java/org/apache/solr/handler/component/SearchHandler.java
```diff
@@ -289,6 +295,23 @@ public void handleRequestBody(SolrQueryRequest req, SolrQueryResponse rsp) throw
...
+
+      if (timer != null) {
+        RTimerTree subt = timer.sub("circuitbreaker");
```
madrob, line 306: We never stop this subtimer.

solr/core/src/java/org/apache/solr/util/circuitbreaker/MemoryCircuitBreaker.java
atris: Interesting, I thought the default toString delegate should handle the right debug output. Added the get, thanks

solr/core/src/java/org/apache/solr/handler/component/SearchHandler.java
atris: I looked at a couple of HTTP codes and SERVICE_UNAVAILABLE seems to be the closest one matching the scenario (https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/503). Did you have anything else in mind?

solr/core/src/java/org/apache/solr/core/SolrConfig.java
atris: I would ideally want this to be a part of the API we were discussing upstream so will add it to the SIP.

atris:
@madrob I aim to close this PR out today if possible -- so happy to stay awake late tonight if you want me to iterate :)

solr/core/src/java/org/apache/solr/handler/component/SearchHandler.java
madrob: Coming back to this, yea, I think 503 is best. I was thinking maybe 429, but this is probably good.

atris:
@madrob @anshumg Thank you for the review!

PR #1608, opened 2020-06-24
LUCENE-9379: Encrypting Directory - functional but to optimize

bruno-roustant:
Performance issue because of too many new Ciphers when slicing IndexInput.
javax.crypto.Cipher is heavy weight to create and is stateful. I tried a CipherPool, but actually there are many cases where we need to get lots of slices of the IndexInput so we have to create lots of new stateful Cipher. The pool turns out to be a no-go, there are too many Cipher in it.

TODO:
- find a lighter alternative to Cipher if it exists.
- a couple of tests still fail because of unclosed IndexOutput.


lucene/core/src/java/org/apache/lucene/store/EncryptingDirectory.java
```diff
@@ -0,0 +1,92 @@
...
+
+import org.apache.lucene.index.SegmentInfo;
+
+public class EncryptingDirectory extends FilterDirectory {
+
+  private final KeySupplier keySupplier;
+  private final SegmentKeySupplier segmentKeySupplier;
+  private final SegmentInfo segmentInfo;
+
+  public EncryptingDirectory(Directory directory, KeySupplier keySupplier) {
```
juanka588, line 30: you can create two objects: one with the keySupplier and the directory, the second one with the segment key supplier and segment info and instead of using the constructors use an static factory method

lucene/core/src/java/org/apache/lucene/store/EncryptingDirectory.java
```diff
@@ -0,0 +1,92 @@
...
+    return segmentInfo == null ? null : segmentInfo.getId();
+  }
+
+  public interface KeySupplier {
+
+    /**
+     * Gets the encryption key for the provided file name.
+     * @return The key; or null if none, in this case the data is not encrypted. It must be either 128, 192 or 256 bits long.
+     */
+    byte[] getKey(String fileName);
```
juanka588, line 81: is it expected to callers to modify this byte[] key? Otherwise it would be preferable to return an immutable object giving a view of the bytes

lucene/core/src/java/org/apache/lucene/store/EncryptingDirectory.java
bruno-roustant: Right, I'll split it into FileEncryptingDirectory and SegmentEncryptingDirectory.

lucene/core/src/java/org/apache/lucene/store/EncryptingDirectory.java
bruno-roustant: A byte[] is the way we configure a Cipher (it is cloned in the process). I'll update the doc to say it's not modified.

lucene/core/src/java/org/apache/lucene/store/EncryptingIndexInput.java
```diff
@@ -0,0 +1,299 @@
...
+  private ByteBuffer inBuffer;
+  private ByteBuffer outBuffer;
+  private byte[] inArray;
+  private byte[] oneByteBuf;
+  private int padding;
+  private boolean closed;
+
+  public EncryptingIndexInput(IndexInput indexInput, byte[] key) throws IOException {
+    this("Decrypting " + indexInput.toString(),
+        HEADER_IV_LENGTH, indexInput.length() - HEADER_IV_LENGTH - CodecUtil.footerLength(), false,
```
juanka588, line 67: method getInnerLength()?

lucene/core/src/java/org/apache/lucene/store/EncryptingIndexInput.java
```diff
@@ -0,0 +1,299 @@
...
+      throw new IllegalArgumentException("Invalid position=" + position);
+    }
+    if (position > length()) {
+      throw new EOFException("Seek beyond EOF (position=" + position + ", length=" + length() + ") in " + this);
+    }
+    long targetPosition = position + sliceOffset;
+    long delegatePosition = indexInput.getFilePointer();
+    long currentPosition = delegatePosition - outBuffer.remaining();
+    if (targetPosition >= currentPosition && targetPosition <= delegatePosition) {
+      outBuffer.position(outBuffer.position() + (int) (targetPosition - currentPosition));
```
juanka588, line 152: Math.toIntExact()?

lucene/core/src/java/org/apache/lucene/store/EncryptingDirectory.java
```diff
@@ -0,0 +1,60 @@
...
+    IndexInput indexInput = in.openInput(name, context);
+    byte[] key = getKey(name);
+    return key == null ? indexInput : createEncryptingIndexInput(indexInput, key);
+  }
+
+  /**
+   * Gets the encryption key for the provided file name.
+   * @return The key; or null if none, in this case the data is not encrypted.
+   */
+  protected abstract byte[] getKey(String fileName);
```
juanka588, line 55: I still prefer the wrapping object avoiding undesirable array references which can be modified.
this also could improve performance if the key is immutable and the clone method is called on the wrapping object.

lucene/core/src/java/org/apache/lucene/store/EncryptingIndexInput.java
bruno-roustant: The target position is within the buffered output, which has a small integer size. I'll add a comment here.

lucene/core/src/java/org/apache/lucene/store/EncryptingDirectory.java
bruno-roustant: javax.crypto.Cipher clones the key. But another Encrypter implementation could read the key bytes to create another AES key, without needing to clone.
I'll let it as a byte[] and I'll be clear in the javadoc it is not modified and not referenced.

lucene/core/src/java/org/apache/lucene/util/crypto/AesCtrEncrypter.java
```diff
@@ -0,0 +1,65 @@
...
+  /**
+   * Initializes this encrypter at the provided CTR block counter (counter of blocks of size {@link EncryptionUtil#AES_BLOCK_SIZE}).
+   * <p>For example, the data byte at index i is inside the block at counter = i / {@link EncryptionUtil#AES_BLOCK_SIZE}.
+   * CTR mode computes an IV for this block based on the initial IV (at counter 0) and the provided counter. This allows
+   * efficient random access to encrypted data. Only the target block needs to be decrypted.</p>
+   * <p>This method must be called first. Then the next call to {@link #process(ByteBuffer, ByteBuffer)} will start at the
+   * beginning of the block: the first byte of data at input buffer {@link ByteBuffer#position()} must be the first byte
+   * of the block.</p>
+   */
+  void init(long counter);
```
juanka588, line 43: setPosition(), setBlockPosoition(), moveToBlock()?

rmuir:
I'm -1 to this approach because it breaks the use of the OS cache (I stated this on the JIRA issue, it was ignored, I don't want anyone to forget).

Because we use the os filesystem cache, any encryption must also be done by os (filesystem, block level, etc) so that page caching works correctly.

dsmiley:
Lets use JIRA ( LUCENE-9379 ) to discuss higher level concerns and GitHub code reviews for code details.

PR #1610, opened 2020-06-24
LUCENE-9384: Backport for field sort optimization

mayya-sharipova:
Backport for: LUCENE-9280: Collectors to skip noncompetitive documents (#1351)

Similar how scorers can update their iterators to skip non-competitive
documents, collectors and comparators should also provide and update
iterators that allow them to skip non-competive documents.

To enable sort optimization for numeric sort fields,
the following needs to be done:
1) the field should be indexed with both doc_values and points, that
must have the same field name and same data
2) SortField#setCanSkipNonCompetitiveDocs must be set
3) totalHitsThreshold should not be set to max value.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/SortField.java
+++ b/lucene/core/src/java/org/apache/lucene/search/SortField.java
@@ -104,6 +104,11 @@ public class SortField {
   // Used for 'sortMissingFirst/Last'
   protected Object missingValue = null;
 
+  // For numeric sort fields, if true, indicates that the same data has been
+  // indexed with doc values and points, which allows to use sort optimization
+  // and skip non-competitive documents.
+  private boolean skipNonCompetitiveDocs = false;
+
   /** Creates a sort by terms in the given field with the type of term
    * values explicitly given.
    * @param field  Name of field to sort by.  Can be <code>null</code> if
@@ -394,6 +399,21 @@ public class SortField {
     return buffer.toString();
   }
 
+
+  /**
+   * For numeric sort fields, setting this field, indicates that
+   * the same numeric data has been indexed with two fields: doc values and points and
+   * that these fields have the same name.
+   * This allows to use sort optimization and skip non-competitive documents.
+   */
+  public void setSkipNonCompetitiveDocs() {
+    this.skipNonCompetitiveDocs = true;
+  }
+
+  public boolean getSkipNonCompetitiveDocs() {
+    return skipNonCompetitiveDocs;
+  }
+
   /** Returns true if <code>o</code> is equal to this.  If a
    *  {@link FieldComparatorSource} was provided, it must properly
    *  implement equals (unless a singleton is always used). */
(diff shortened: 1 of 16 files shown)


lucene/core/src/java/org/apache/lucene/search/SortField.java
```diff
@@ -394,6 +399,21 @@ public String toString() {
...
   }
 
+
+  /**
+   * For numeric sort fields, setting this field, indicates that
+   * the same numeric data has been indexed with two fields: doc values and points and
+   * that these fields have the same name.
+   * This allows to use sort optimization and skip non-competitive documents.
+   */
+  public void setSkipNonCompetitiveDocs() {
```
jimczi, line 409: I wonder if we should make the option less ambiguous by naming it something like `setCanUsePoints` or something along those lines ? I am afraid that this could be misleading if we set this option on a sort field that use the _score or a keyword field for instance. We should maybe at least validate that the field is not `null` ?

mayya-sharipova: How about `setCanUsePointsForSort`? I am also fine with `setCanUsePoints`.

mayya-sharipova: @jpountz  Do you have suggestions how we should expose sort optimization in 8.x? Are you ok with a proposal to have  a method for `SortField::setCanUsePointsForSort` that when set `true` indicates that the same data is used for indexing points and docValues?

jpountz: IMO `setCanUsePoints` is good enough since this is a "SortField" so it should be obvious that this is for sorting?

lucene/core/src/java/org/apache/lucene/search/SortField.java
```diff
@@ -408,6 +428,7 @@ public boolean equals(Object o) {
       && other.reverse == this.reverse
       && Objects.equals(this.comparatorSource, other.comparatorSource)
       && Objects.equals(this.missingValue, other.missingValue)
+      && other.skipNonCompetitiveDocs == this.skipNonCompetitiveDocs
```
jpountz, line 431: I wonder if we should actually take it into account for equals/hashcode as this could disable index sorting optimizations, which check for equality of the index-time and search-time SortField instances.

mayya-sharipova: @jpountz Thank you for your feedback. Alternatively we can add `setCanUsePoints` parameter to `Sort` object instead of `SortField`?  What do you think?

mayya-sharipova:
@jimczi  @jpountz Thanks for your feedback, I have addressed it:
- renamed `SortField#setSkipNonCompetitiveDocs` to `SortField#setCanUsePoints `
- remove `canUsePoints` field from `SortField#equals` and `hashCode`.

lucene/core/src/java/org/apache/lucene/search/SortField.java
```diff
@@ -428,7 +428,6 @@ public boolean equals(Object o) {
       && other.reverse == this.reverse
       && Objects.equals(this.comparatorSource, other.comparatorSource)
       && Objects.equals(this.missingValue, other.missingValue)
-      && other.skipNonCompetitiveDocs == this.skipNonCompetitiveDocs
```
jpountz: Can you add a comment to say that we're not taking it into account on purpose? I foresee that someone will file a PR to add it after running a code analysis tool that would complain that not all fields are taken into account for equals otherwise.

PR #1611, opened 2020-06-24
Change JoinQuery class's visibility to package again

atris:


atris:
@dsmiley Updated as per discussion

atris:
Cool, thanks. Will merge post the CI completion, just to be sure :)

PR #1613, opened 2020-06-24
LUCENE-8574 Cache ExpressionFunctionValues

zhaih:
# Description

 * Add a new DoubleValuesSource which will enforce only one value per name
in dependencies
 * Add few fields to `ExpressionFunctionValues` to ensure for a given doc, it is only computed once
 * Added a unit test that need this mechanism to pass, otherwise would cost
huge memory and time

`ant test` and `ant precommit` passed

# Solution

In `CachingExpressionValuesSource`, a modified version of `getValues` is used to get `DoubleValues`, it will be called with additional `valueCache` passed in and update `valueCache` along the creation process to ensure only one value per name is created.

# Tests

A new test method `testFibonacciExpr` is added to `TestExpressionValueSource` to show the difference, normally it takes 1.x sec on my laptop with `n=30`, and much longer when `n=40` because of the exponential growth.

# Checklist

### DIFF
--- a/lucene/expressions/src/java/org/apache/lucene/expressions/CachingExpressionValueSource.java
+++ b/lucene/expressions/src/java/org/apache/lucene/expressions/CachingExpressionValueSource.java
@@ -62,7 +62,8 @@ final class CachingExpressionValueSource extends ExpressionValueSource {
           values = variables[i].getValues(readerContext, scores);
         }
         if (values == null) {
-          throw new RuntimeException("Internal error. External (" + externalName + ") does not exist.");
+          throw new RuntimeException("Unrecognized variable (" + externalName + ") referenced in expression (" +
+              expression.sourceText + ").");
         }
         valuesCache.put(externalName, values);
       }
--- a/lucene/expressions/src/java/org/apache/lucene/expressions/CachingExpressionValueSource.java
+++ b/lucene/expressions/src/java/org/apache/lucene/expressions/CachingExpressionValueSource.java
@@ -1,0 +1,70 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.lucene.expressions;
+
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+
+import org.apache.lucene.index.LeafReaderContext;
+import org.apache.lucene.search.DoubleValues;
+import org.apache.lucene.search.DoubleValuesSource;
+
+final class CachingExpressionValueSource extends ExpressionValueSource {
+
+  CachingExpressionValueSource(Bindings bindings, Expression expression) {
+    super(bindings, expression);
+  }
+
+  CachingExpressionValueSource(DoubleValuesSource[] variables, Expression expression, boolean needsScores) {
+    super(variables, expression, needsScores);
+  }
+
+  public CachingExpressionValueSource(ExpressionValueSource expressionValueSource) {
+    super(expressionValueSource.variables, expressionValueSource.expression, expressionValueSource.needsScores);
+  }
+
+  @Override
+  public DoubleValues getValues(LeafReaderContext readerContext, DoubleValues scores) throws IOException {
+    return getValuesWithCache(readerContext, scores, new HashMap<>());
+  }
+
+  public DoubleValues getValuesWithCache(LeafReaderContext readerContext, DoubleValues scores,
+                                                  Map<String, DoubleValues> valuesCache) throws IOException {
+    DoubleValues[] externalValues = new DoubleValues[expression.variables.length];
+
+    for (int i = 0; i < variables.length; ++i) {
+      String externalName = expression.variables[i];
+      DoubleValues values = valuesCache.get(externalName);
+      if (values == null) {
+        if (variables[i] instanceof CachingExpressionValueSource) {
+          values = ((CachingExpressionValueSource)variables[i]).getValuesWithCache(readerContext, scores, valuesCache);
+        } else {
+          values = variables[i].getValues(readerContext, scores);
+        }
+        if (values == null) {
+          throw new RuntimeException("Internal error. External (" + externalName + ") does not exist.");
+        }
+        valuesCache.put(externalName, values);
+      }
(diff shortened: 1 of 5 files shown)


lucene/expressions/src/test/org/apache/lucene/expressions/TestExpressionValueSource.java
```diff
@@ -128,6 +128,34 @@ public void testDoubleValuesSourceEquals() throws Exception {
     assertFalse(vs1.equals(vs4));
   }
 
+  public void testFibonacciExpr() throws Exception {
+    int n = 40;
+    SimpleBindings bindings = new SimpleBindings();
+    bindings.add("f0", DoubleValuesSource.constant(0));
+    bindings.add("f1", DoubleValuesSource.constant(1));
+    for (int i = 2; i < n + 1; i++) {
```
mikemccand, line 136: Maybe adds comments to explain that before the fix, this test takes XXX hours to finish, and after the fix it's nearly instantaneous as judged by casual human observer?

lucene/expressions/src/java/org/apache/lucene/expressions/CachingExpressionValueSource.java
```diff
@@ -0,0 +1,70 @@
...
+
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+
+import org.apache.lucene.index.LeafReaderContext;
+import org.apache.lucene.search.DoubleValues;
+import org.apache.lucene.search.DoubleValuesSource;
+
+final class CachingExpressionValueSource extends ExpressionValueSource {
```
mikemccand, line 28: Maybe add some javadocs explaining why this is needed, and explain that the cache is only per-hit?

lucene/expressions/src/java/org/apache/lucene/expressions/CachingExpressionValueSource.java
```diff
@@ -0,0 +1,70 @@
...
+  public CachingExpressionValueSource(ExpressionValueSource expressionValueSource) {
+    super(expressionValueSource.variables, expressionValueSource.expression, expressionValueSource.needsScores);
+  }
+
+  @Override
+  public DoubleValues getValues(LeafReaderContext readerContext, DoubleValues scores) throws IOException {
+    return getValuesWithCache(readerContext, scores, new HashMap<>());
+  }
+
+  public DoubleValues getValuesWithCache(LeafReaderContext readerContext, DoubleValues scores,
```
mikemccand, line 47: Can this be `private`?

lucene/expressions/src/java/org/apache/lucene/expressions/CachingExpressionValueSource.java
```diff
@@ -0,0 +1,70 @@
...
+                                                  Map<String, DoubleValues> valuesCache) throws IOException {
+    DoubleValues[] externalValues = new DoubleValues[expression.variables.length];
+
+    for (int i = 0; i < variables.length; ++i) {
+      String externalName = expression.variables[i];
+      DoubleValues values = valuesCache.get(externalName);
+      if (values == null) {
+        if (variables[i] instanceof CachingExpressionValueSource) {
+          values = ((CachingExpressionValueSource)variables[i]).getValuesWithCache(readerContext, scores, valuesCache);
```
mikemccand, line 56: Add space between the cast and `variables[i]`?

lucene/expressions/src/java/org/apache/lucene/expressions/CachingExpressionValueSource.java
```diff
@@ -0,0 +1,70 @@
...
+        } else {
+          values = variables[i].getValues(readerContext, scores);
+        }
+        if (values == null) {
+          throw new RuntimeException("Internal error. External (" + externalName + ") does not exist.");
```
mikemccand, line 61: Hmm when does this happen?  Is it if the expression references a dependent variable that was not bound?  Can we change the message to something like: `Unrecognized variable XXX referenced in expression YYY`?

lucene/expressions/src/java/org/apache/lucene/expressions/ExpressionFunctionValues.java
```diff
@@ -38,14 +41,23 @@
...
     }
+    currentDoc = doc;
+    computed = false;
     return true;
   }
   
   @Override
   public double doubleValue() {
-    return expression.evaluate(functionValues);
+    if (!computed) {
```
mikemccand, line 57: Can we use `computed == false` instead to improve readability by possibly aged eyes and reduce risk of future bugs?

lucene/expressions/src/java/org/apache/lucene/expressions/CachingExpressionValueSource.java
zhaih: Yeah I think this error is happened because of a dependent variable was not bound. The message is actually inherited from `ExpressionValueSource#getValue` so I believe we'd better not change it or change both?

lucene/expressions/src/java/org/apache/lucene/expressions/CachingExpressionValueSource.java
zhaih: Good catch! Yes I think it should be `private`

lucene/expressions/src/test/org/apache/lucene/expressions/TestExpressionValueSource.java
zhaih: Sure.

lucene/expressions/src/java/org/apache/lucene/expressions/CachingExpressionValueSource.java
mikemccand: Oh yeah can we please change both?

zhaih: Changed!

zhaih:
Rebased to master and gradle precommit failed because of following warning (which becomes error since `-Werror specified`):
> /home/runner/work/lucene-solr/lucene-solr/solr/core/src/java/org/apache/solr/handler/export/ExportWriterStream.java:69: warning: [rawtypes] found raw type: List
        List lst = new ArrayList();

ErickErickson:
Patrick:

This is a relatively recent change (failing on unsuppressed warnings).

The preferred solution is to to type the list, e.g.

List<String> lst = new ArrayList<>();

If that’s too difficult, you can add

@SupppressWarnings({“rawtypes”})

in this case you can probably add it right above the declaration. But please try to solve it
correctly first by adding a type to the List.

Yeah, I know there are a zillion SuppressWarnings in the code; we’ve been accruing
this kind of cruft for years. I recently went through and added lots of SuppressWarnings so
we could get to the point of having a chance of not adding any new warnings because
things fail. BTW, compilation should have failed well in advance of precommit.

Second BTW, “the gradle way” is “gradlew check -x test”...

Oh, and if you are working on code anyway with SuppressWarnings and can fix a few
of those too on the way by that would be great!

Best,
Erick

zhaih:
Eric:
Thanks for clarifying it!
However I think it's an existing code lies in current mainline that cause this warning(but not code change in this PR):
https://github.com/apache/lucene-solr/blob/1b8fb70216030c19e0f1139065c56e0c273eb206/solr/core/src/java/org/apache/solr/handler/export/ExportWriterStream.java#L69
So I believe we should open another PR to fix it?

mikemccand:
> So I believe we should open another PR to fix it?

Looks like this is (separately) fixed now ... @zhaih could you rebase so tests re-run?

PR #1615, opened 2020-06-25
SOLR-14588: Follow Up Fixes and Documentation

atris:
This commit is a follow up to the original commit and adds more documentation and adds timing information for circuit breaker check only if circuit breakers are enabled


atris:
@ErickErickson This fixes the test failure. I ran the test around 20 times with and without the failing seed -- seems clean.

@sigram Updated the documentation -- let me know if it looks fine.

This is a minor PR -- I plan to merge in 24 (well, 22, so that I get proper sleep tomorrow :) ) hours unless objections.

PR #1616, opened 2020-06-25
SOLR-14590: Add RankQParserPlugin

tflobbe:
This is still WIP, just the query parsing so far. Currently working on the indexing side of things.

### DIFF
--- a/solr/core/src/java/org/apache/solr/schema/RankField.java
+++ b/solr/core/src/java/org/apache/solr/schema/RankField.java
@@ -1,0 +1,79 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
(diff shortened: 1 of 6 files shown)


solr/core/src/java/org/apache/solr/schema/RankField.java
```diff
@@ -0,0 +1,79 @@
...
+    properties &= ~(UNINVERTIBLE | STORED | DOC_VALUES);
+    
+  }
+
+  @Override
+  protected IndexableField createField(String name, String val, IndexableFieldType type) {
+    if (val == null || val.isEmpty()) {
+      return null;
+    }
+    String[] parts = val.split(":");
```
vthacker, line 55: Chatting with @tflobbe on this - Can we improve the API to not expose `fieldName/fieldValue` pair and just ask the user to enter one input?

For example could we take either of these two approaches
1. Use an `_internalFeature` field. The `fieldName` that the user defines would be used as the `featureName` in the FeatureField ctor . So internally everything is stored in one field.
2. Every feature gets is own field - so the user only defines the `fieldName` and we use the same value as the `featureName` 

Approach 2 would definitely lead to lots of fields. But maybe it's worth exploring 1 and seeing if that helps improve the API without trading any performance.

tflobbe:
Superseded by https://github.com/apache/lucene-solr/pull/1620

PR #1617, opened 2020-06-26
LUCENE-8962: Merge small segments on commit

s1monw:
This PR revisits the merge-on-commit patch submitted by @msfroh a little while ago. The only change from that earlier PR is a fix for failures uncovered by TestIndexWriter.testRandomOperations, some whitespace cleanups, and a rebase on the current master branch. The problem was that updateSegmentInfosOnMergeFinish would incorrectly decRef a merged segments' files if that segment was modified by deletions (or updates) while it was being merged.

With this fix, I ran the failing test case several thousands of times with no failures, whereas before it would routinely fail after a few hundred test runs.

Co-authored-by: Michael Froh <msfroh@apache.org>
Co-authored-by: Michael Sokolov <sokolov@falutin.net>

This is the third try after the patch itself and #1552


s1monw:

lucene/core/src/java/org/apache/lucene/index/IndexWriterConfig.java
```diff
@@ -459,6 +463,27 @@ public IndexWriterConfig setCommitOnClose(boolean commitOnClose) {
...
+   * Note: This settings has no effect unless {@link MergePolicy#findFullFlushMerges(MergeTrigger, SegmentInfos, MergePolicy.MergeContext)}
+   * has an implementation that actually returns merges which by default doesn't return any merges.
+   */
+  public IndexWriterConfig setMaxCommitMergeWaitMillis(long maxCommitMergeWaitMillis) {
+    this.maxCommitMergeWaitMillis = maxCommitMergeWaitMillis;
+    return this;
+  }
+
+  /** We only allow sorting on these types */
+  private static final EnumSet<SortField.Type> ALLOWED_INDEX_SORT_TYPES = EnumSet.of(SortField.Type.STRING,
```
mikemccand, line 481: Hmm how did this sneak in?

msfroh:
Wow! I didn't realize that capturing/cloning live docs at a point in time would be so complicated. I learned a lot from this PR about how deletes are handled. Thanks, Simon!

lucene/core/src/java/org/apache/lucene/index/IndexWriterConfig.java
mikemccand: OK I see this was removed earlier in mainline for LUCENE-9330 -- it must've snuck back in as a merge conflict.

I'll re-remove now.

lucene/core/src/java/org/apache/lucene/index/MergePolicy.java
```diff
@@ -399,6 +416,40 @@ boolean hasFinished() {
...
+      ArrayList<MergeReader> readers = new ArrayList<>(segments.size());
+      try {
+        for (final SegmentCommitInfo info : segments) {
+          // Hold onto the "live" reader; we will use this to
+          // commit merged deletes
+          readers.add(readerFactory.apply(info));
+        }
+      } finally {
+        // ensure we assign this to close them in the case of an exception
+        this.mergeReaders = List.copyOf(readers);
```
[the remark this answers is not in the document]
s1monw: I left a comment. I use immutable lists everywhere here it would be fatal if we modify this list outside of OneMerge. I think that justifies the copy.

lucene/core/src/java/org/apache/lucene/index/IndexWriterConfig.java
s1monw: 👍

lucene/core/src/java/org/apache/lucene/index/IndexWriterConfig.java
```diff
@@ -459,6 +463,21 @@ public IndexWriterConfig setCommitOnClose(boolean commitOnClose) {
...
+   * Expert: sets the amount of time to wait for merges (during {@link IndexWriter#commit}) returned by
+   * MergePolicy.findFullFlushMerges(...).
+   * If this time is reached, we proceed with the commit based on segments merged up to that point.
+   * The merges are not cancelled, and will still run to completion independent of the commit,
+   * like natural segment merges. The default is <code>{@value IndexWriterConfig#DEFAULT_MAX_COMMIT_MERGE_WAIT_MILLIS}</code>.
+   *
+   * Note: This settings has no effect unless {@link MergePolicy#findFullFlushMerges(MergeTrigger, SegmentInfos, MergePolicy.MergeContext)}
+   * has an implementation that actually returns merges which by default doesn't return any merges.
+   */
+  public IndexWriterConfig setMaxCommitMergeWaitMillis(long maxCommitMergeWaitMillis) {
```
s1monw, line 476: I am torn if we'd use a double or long here. I think we used double in other places but I do like a discrete value better...

s1monw:
> Wow! I didn't realize that capturing/cloning live docs at a point in time would be so complicated. I learned a lot from this PR about how deletes are handled. Thanks, Simon!

@msfroh I didn't either :) it's been a while that I digged into merging code and I think the same is true for @mikemccand - I am impressed how far you got in this pretty complex class. I'd love to see you digging more and coming up with more stuff like this. Thank you for contributing this. I am always happy to help. This entire process from the first patch to the feature being integrated is a great example of how amazing OSS is.

s1monw:
merged into master. Thanks everybody involved!!

PR #1620, opened 2020-06-26
SOLR-14590 : Add support for Lucene's FeatureField in Solr

tflobbe:
This implements the "Approach 2"  suggested by Varun here: https://github.com/apache/lucene-solr/pull/1616#discussion_r446404384

### DIFF
--- a/solr/core/src/java/org/apache/solr/schema/RankField.java
+++ b/solr/core/src/java/org/apache/solr/schema/RankField.java
@@ -29,10 +29,43 @@ import org.apache.lucene.search.TermQuery;
 import org.apache.solr.common.SolrException;
 import org.apache.solr.response.TextResponseWriter;
 import org.apache.solr.search.QParser;
+import org.apache.solr.search.RankQParserPlugin;
 import org.apache.solr.uninverting.UninvertingReader.Type;
 
+/**
+ * <p>
+ * {@code RankField}s can be used to store scoring factors to improve document ranking. They should be used
+ * in combination with {@link RankQParserPlugin}. To use:
+ * </p>
+ * <p>
+ * Define the {@code RankField} {@code fieldType} in your schema:
+ * </p>
+ * <pre class="prettyprint">
+ * &lt;fieldType name="rank" class="solr.RankField" /&gt;
+ * </pre>
+ * <p>
+ * Add fields to the schema, i.e.:
+ * </p>
+ * <pre class="prettyprint">
+ * &lt;field name="rank_1" type="rank" /&gt;
+ * </pre>
+ * 
+ * Query using the {@link RankQParserPlugin}, for example
+ * <pre class="prettyprint">
+ * http://localhost:8983/solr/techproducts?q=memory _query_:{!rank f='rank_1', function='log' scalingFactor='1.2'}
+ * </pre>
+ * 
+ * @see RankQParserPlugin
+ * @lucene.experimental
+ * @since 8.6
+ */
 public class RankField extends FieldType {
   
+  /*
+   * While the user can create multiple RankFields, internally we use a single Lucene field,
+   * and we map the Solr field name to the "feature" in Lucene's FeatureField. This is mainly
+   * to simplify the user experience.
+   */
   public static final String INTERNAL_RANK_FIELD_NAME = "_internal_rank_field";
 
   @Override
@@ -75,6 +108,7 @@ public class RankField extends FieldType {
     } catch (NumberFormatException nfe) {
       throw new SolrException(SolrException.ErrorCode.BAD_REQUEST, "Error while creating field '" + name + "' from value '" + val + "'. Expecting float.", nfe);
     }
+    // Internally, we always use the same field
     return new FeatureField(INTERNAL_RANK_FIELD_NAME, name, featureValue);
   }
   
--- a/solr/core/src/java/org/apache/solr/schema/RankField.java
+++ b/solr/core/src/java/org/apache/solr/schema/RankField.java
@@ -1,0 +1,79 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.solr.schema;
+
+import java.io.IOException;
+import java.util.Map;
+
+import org.apache.lucene.document.FeatureField;
+import org.apache.lucene.index.IndexableField;
+import org.apache.lucene.index.IndexableFieldType;
+import org.apache.lucene.search.Query;
+import org.apache.lucene.search.SortField;
+import org.apache.solr.common.SolrException;
+import org.apache.solr.response.TextResponseWriter;
+import org.apache.solr.search.QParser;
+import org.apache.solr.uninverting.UninvertingReader.Type;
+
+public class RankField extends FieldType {
+
+  @Override
+  public Type getUninversionType(SchemaField sf) {
+    throw null;
+  }
+
+  @Override
+  public void write(TextResponseWriter writer, String name, IndexableField f) throws IOException {
+  }
+  
+  @Override
+  protected void init(IndexSchema schema, Map<String,String> args) {
+    super.init(schema, args);
+    properties &= ~(UNINVERTIBLE | STORED | DOC_VALUES);
+    
+  }
+
+  @Override
+  protected IndexableField createField(String name, String val, IndexableFieldType type) {
+    if (val == null || val.isEmpty()) {
+      return null;
+    }
+    String[] parts = val.split(":");
+    if (parts.length != 2) {
+      throw new SolrException(SolrException.ErrorCode.BAD_REQUEST, "Error while creating field '" + name + "' from value '" + val + "'. Expecting format to be 'feature:value'");
+    }
+    float featureValue;
+    try {
+      featureValue = Float.parseFloat(parts[1]);
+    } catch (NumberFormatException nfe) {
+      throw new SolrException(SolrException.ErrorCode.BAD_REQUEST, "Error while creating field '" + name + "' from value '" + val + "'. Expecting format to be 'feature:value'", nfe);
+    }
+    return new FeatureField(name, parts[0], featureValue);
+  }
+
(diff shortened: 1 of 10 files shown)


solr/core/src/java/org/apache/solr/schema/RankField.java
```diff
@@ -0,0 +1,92 @@
...
+    try {
+      featureValue = Float.parseFloat(val);
+    } catch (NumberFormatException nfe) {
+      throw new SolrException(SolrException.ErrorCode.BAD_REQUEST, "Error while creating field '" + name + "' from value '" + val + "'. Expecting float.", nfe);
+    }
+    return new FeatureField(INTERNAL_RANK_FIELD_NAME, name, featureValue);
+  }
+
+  @Override
+  public Query getFieldQuery(QParser parser, SchemaField field, String externalVal) {
```
vthacker, line 82: I think so? Essentially this is an exists query right?

We can be safe and not support this for now? 

Thinking aloud on another option - What if we rewrite ( like how you have it right now ) when the value is `*` and throw an exception otherwise?

tflobbe: Good idea. I'll add that

solr/core/src/java/org/apache/solr/schema/RankField.java
```diff
@@ -0,0 +1,140 @@
...
+ * @since 8.6
+ */
+public class RankField extends FieldType {
+  
+  /*
+   * While the user can create multiple RankFields, internally we use a single Lucene field,
+   * and we map the Solr field name to the "feature" in Lucene's FeatureField. This is mainly
+   * to simplify the user experience.
+   */
+  public static final String INTERNAL_RANK_FIELD_NAME = "_internal_rank_field";
```
dsmiley, line 69: should end with a final underscore as well like our other internal fields

dsmiley: might simply \_rank\_ be fine?  The "internal" aspect is implied by the leading/trailing underscore, and I don't think _any_ field needs to have "field" in its name ;-)

solr/core/src/java/org/apache/solr/schema/RankField.java
```diff
@@ -0,0 +1,140 @@
...
+ * Define the {@code RankField} {@code fieldType} in your schema:
+ * </p>
+ * <pre class="prettyprint">
+ * &lt;fieldType name="rank" class="solr.RankField" /&gt;
+ * </pre>
+ * <p>
+ * Add fields to the schema, i.e.:
+ * </p>
+ * <pre class="prettyprint">
+ * &lt;field name="rank_1" type="rank" /&gt;
```
vthacker, line 50: small nit: something like `document_length_boost` might be a better example to help a user realize how they can leverage this feature?

solr/core/src/java/org/apache/solr/schema/RankField.java
vthacker: > Thinking aloud on another option - What if we rewrite ( like how you have it right now ) when the value is * and throw an exception otherwise?

Nice! we added it to `getExistenceQuery` and in `getFieldQuery` we throw an exception!

solr/core/src/java/org/apache/solr/search/RankQParserPlugin.java
```diff
@@ -0,0 +1,158 @@
...
+ * 
+ *  @see RankField
+ * 
+ * @lucene.experimental
+ * @since 8.6
+ */
+public class RankQParserPlugin extends QParserPlugin {
+  
+  public static final String NAME = "rank";
+  public static final String FIELD = "f";
```
vthacker, line 42: 👍 All parsers in https://lucene.apache.org/solr/guide/8_5/other-parsers.html that expect a field use `f` as the key

solr/core/src/test/org/apache/solr/schema/RankFieldTest.java
```diff
@@ -0,0 +1,261 @@
...
+    
+    assertQ(req("q", "foo",
+        "defType", "dismax",
+        "qf", "str_field^10",
+        "bq", "{!rank f='" + RANK_2 + "' function='log' scalingFactor='1'}"
+        ),
+        "//*[@numFound='2']",
+        "//result/doc[1]/str[@name='id'][.='1']",
+        "//result/doc[2]/str[@name='id'][.='2']");
+  }
```
vthacker, line 259: I manually checked the 'score' from these two runs to ensure that the score was changing! Would it make sense to add this as a test?

```
String rsp = h.query(req("q", "foo",
        "defType", "dismax",
        "qf", "str_field^10",
        "fl", "*,score"
    ));

    rsp = h.query(req("q", "foo",
        "defType", "dismax",
        "qf", "str_field^10",
        "fl", "*,score",
        "bq", "{!rank f='" + RANK_2 + "' function='log' scalingFactor='1'}"
    ));
```

vthacker: Also I was curious to see if `debugQuery=true` does the right thing ( it does! )

```


<?xml version="1.0" encoding="UTF-8"?>
<response>
   <lst name="responseHeader">
      <int name="status">0</int>
      <int name="QTime">161</int>
      <lst name="params">
         <str name="q">foo</str>
         <str name="defType">dismax</str>
         <str name="qf">str_field^10</str>
         <str name="fl">*,score,rank_1,rank_2</str>
         <str name="wt">xml</str>
         <str name="debugQuery">true</str>
         <str name="bq">{!rank f='rank_2' function='log' scalingFactor='1'}</str>
      </lst>
   </lst>
   <result name="response" numFound="2" start="0" maxScore="1.9273467" numFoundExact="true">
      <doc>
         <str name="id">1</str>
         <str name="str_field">foo</str>
         <float name="score">1.9273467</float>
      </doc>
      <doc>
         <str name="id">2</str>
         <str name="str_field">foo</str>
         <float name="score">1.5218816</float>
      </doc>
   </result>
   <lst name="debug">
      <str name="rawquerystring">foo</str>
      <str name="querystring">foo</str>
      <str name="parsedquery">+DisjunctionMaxQuery(((str_field:foo)^10.0)) () FeatureQuery(FeatureQuery(field=_rank_, feature=rank_2, function=LogFunction(scalingFactor=1.0)))</str>
      <str name="parsedquery_toString">+((str_field:foo)^10.0) () FeatureQuery(field=_rank_, feature=rank_2, function=LogFunction(scalingFactor=1.0))</str>
      <lst name="explain">
         <str>1.9273467 = sum of:
  0.82873434 = weight(str_field:foo in 0) [SchemaSimilarity], result of:
    0.82873434 = score(freq=1.0), computed as boost * idf * tf from:
      10.0 = boost
      0.18232156 = idf, computed as log(1 + (N - n + 0.5) / (n + 0.5)) from:
        2 = n, number of documents containing term
        2 = N, total number of documents with field
      0.45454544 = tf, computed as freq / (freq + k1 * (1 - b + b * dl / avgdl)) from:
        1.0 = freq, occurrences of term within document
        1.2 = k1, term saturation parameter
        0.75 = b, length normalization parameter
        1.0 = dl, length of field
        1.0 = avgdl, average length of field
  1.0986123 = Log function on the _rank_ field for the rank_2 feature, computed as w * log(a + S) from:
    1.0 = w, weight of this function
    1.0 = a, scaling factor
    2.0 = S, feature value</str>
         <str>1.5218816 = sum of:
  0.82873434 = weight(str_field:foo in 1) [SchemaSimilarity], result of:
    0.82873434 = score(freq=1.0), computed as boost * idf * tf from:
      10.0 = boost
      0.18232156 = idf, computed as log(1 + (N - n + 0.5) / (n + 0.5)) from:
        2 = n, number of documents containing term
        2 = N, total number of documents with field
      0.45454544 = tf, computed as freq / (freq + k1 * (1 - b + b * dl / avgdl)) from:
        1.0 = freq, occurrences of term within document
        1.2 = k1, term saturation parameter
        0.75 = b, length normalization parameter
        1.0 = dl, length of field
        1.0 = avgdl, average length of field
  0.6931472 = Log function on the _rank_ field for the rank_2 feature, computed as w * log(a + S) from:
    1.0 = w, weight of this function
    1.0 = a, scaling factor
    1.0 = S, feature value</str>
      </lst>
      <str name="QParser">DisMaxQParser</str>
      <null name="altquerystring" />
      <arr name="boost_queries">
         <str>{!rank f='rank_2' function='log' scalingFactor='1'}</str>
      </arr>
      <arr name="parsed_boost_queries">
         <str>FeatureQuery(FeatureQuery(field=_rank_, feature=rank_2, function=LogFunction(scalingFactor=1.0)))</str>
      </arr>
      <null name="boostfuncs" />
      <lst name="timing">
         <double name="time">160.0</double>
         <lst name="prepare">
            <double name="time">1.0</double>
            <lst name="query">
               <double name="time">1.0</double>
            </lst>
            <lst name="facet">
               <double name="time">0.0</double>
            </lst>
            <lst name="facet_module">
               <double name="time">0.0</double>
            </lst>
            <lst name="mlt">
               <double name="time">0.0</double>
            </lst>
            <lst name="highlight">
               <double name="time">0.0</double>
            </lst>
            <lst name="stats">
               <double name="time">0.0</double>
            </lst>
            <lst name="expand">
               <double name="time">0.0</double>
            </lst>
            <lst name="terms">
               <double name="time">0.0</double>
            </lst>
            <lst name="debug">
               <double name="time">0.0</double>
            </lst>
         </lst>
         <lst name="process">
            <double name="time">155.0</double>
            <lst name="query">
               <double name="time">3.0</double>
            </lst>
            <lst name="facet">
               <double name="time">0.0</double>
            </lst>
            <lst name="facet_module">
               <double name="time">0.0</double>
            </lst>
            <lst name="mlt">
               <double name="time">0.0</double>
            </lst>
            <lst name="highlight">
               <double name="time">0.0</double>
            </lst>
            <lst name="stats">
               <double name="time">0.0</double>
            </lst>
            <lst name="expand">
               <double name="time">0.0</double>
            </lst>
            <lst name="terms">
               <double name="time">0.0</double>
            </lst>
            <lst name="debug">
               <double name="time">152.0</double>
            </lst>
         </lst>
      </lst>
   </lst>
</response>
```

solr/core/src/test/org/apache/solr/schema/RankFieldTest.java
```diff
@@ -0,0 +1,261 @@
...
+
```
vthacker, line 260: If I add `"fl", "*,score,rank_1,rank_2",` we won't retrieve the value back. I think it's fine not to throw an error or anything right?

solr/core/src/java/org/apache/solr/schema/RankField.java
tflobbe: I've switched to `pagerank`, since that's also in Jira and tests.

solr/core/src/test/org/apache/solr/schema/RankFieldTest.java
tflobbe: Added the "score change" test

solr/core/src/test/org/apache/solr/schema/RankFieldTest.java
tflobbe: I think it should be fine? I believe this is the behavior when trying to retrieve non-stored (or dv) fields?

PR #1624, opened 2020-06-28
use MethodHandles in AnnotatedAPI

noblepaul:
@uschindler Please review


solr/core/src/java/org/apache/solr/api/AnnotatedApi.java
```diff
@@ -306,7 +313,7 @@ void invoke(SolrQueryRequest req, SolrQueryResponse rsp, CommandOperation cmd) {
       } catch (InvocationTargetException ite) {
         log.error("Error executing command ", ite);
         throw new SolrException(SolrException.ErrorCode.SERVER_ERROR, ite.getCause());
-      } catch (Exception e) {
+      } catch (Throwable e) {
```
dsmiley, line 316: Catching Throwable is generally something to avoid.  Why here do you think it makes sense?

noblepaul: `methodhandle.invoke()` throws `Throwable`. What choice do we have?

chatman:
With this PR, the security exceptions are gone. +1 to merge.

solr/core/src/java/org/apache/solr/api/AnnotatedApi.java
```diff
@@ -87,16 +88,18 @@ public EndPoint getEndPoint() {
...
   }
-  public static List<Api> getApis(Class<? extends Object> klas , Object obj) {
-    if (!Modifier.isPublic(klas.getModifiers())) {
-      throw new RuntimeException(klas.getName() + " is not public");
+  public static List<Api> getApis(Class<? extends Object> theClass , Object obj)  {
+    Class<?> klas = null;
+    try {
+      klas = MethodHandles.publicLookup().accessClass(theClass);
+    } catch (IllegalAccessException e) {
+      throw new RuntimeException(klas.getName() + " is not public", e);
```
murblanc, line 96: If the exception is thrown, `klas` will be null and `klass.getName()` will throw NPE

solr/core/src/java/org/apache/solr/api/AnnotatedApi.java
```diff
@@ -225,7 +228,11 @@ public void call(SolrQueryRequest req, SolrQueryResponse rsp) {
       if (Modifier.isPublic(method.getModifiers())) {
         this.command = command;
         this.obj = obj;
-        this.method = method;
+        try {
+          this.method = MethodHandles.publicLookup().unreflect(method);
+        } catch (IllegalAccessException e) {
+          throw new RuntimeException("Unable to unlookup method");
```
chatman, line 234: Lets throw SolrException here.

murblanc: Unrelated to to this PR (i.e. comment holds for previous version of the code), maybe `InterruptedException` should be caught separately and `Thread.currentThread().interrupt()` be called.

murblanc: unlookup? (did you mean unreflect?)

solr/core/src/java/org/apache/solr/api/AnnotatedApi.java
murblanc: I think a comment referencing the `MethodHandle` Javadoc would be helpful here, as the Javadoc explains `Throwable` should be caught and the exception wrapped.

> But in Java source code, methods which perform method handle calls must either explicitly throw Throwable, or else must catch all throwables locally, rethrowing only those which are legal in the context, and wrapping ones which are illegal.

solr/core/src/java/org/apache/solr/api/AnnotatedApi.java
```diff
@@ -113,7 +116,7 @@ public EndPoint getEndPoint() {
       return Collections.singletonList(new AnnotatedApi(specProvider, endPoint, commands, null));
     } else {
       List<Api> apis = new ArrayList<>();
-      for (Method m : klas.getDeclaredMethods()) {
+      for (Method m : klas.getMethods()) {
         EndPoint endPoint = m.getAnnotation(EndPoint.class);
         if (endPoint == null) continue;
         if (!Modifier.isPublic(m.getModifiers())) {
```
murblanc, line 122: Is this still needed? Wouldn't `MethodHandles.publicLookup().accessClass(theClass)` only return `public` methods?

uschindler: Yes, we still need some way to check public methods. The accessClass() method above returns itsself, but the check for public is not needed if we use getMethods() instead of getDeclaredMethods().
The accessClass() check can't be backported to 8.x as it is new in Java 9. So we need to somehow do the check for accessibility manually.

murblanc: So you agree @uschindler that the `if` can be removed in the `master` branch we're looking at, did I understand you correctly?

solr/core/src/java/org/apache/solr/api/AnnotatedApi.java
uschindler: There are two things to do:
- have they try catch only around the invoke and nothing else, so we don't accidentally catch any unwanted exception in the surrounding code
- only checked exceptions should be wrapped, everything else rethrown. This keeps stack traces clean.

The reason for the declaration of "Throwable" comes from the fact that the method handle always throws the original Exception and it's now wrapped like for reflection. As checked exceptions is a compiler thing, normal code using method handles in bytecode or invokedynamic never needs to catch Throwable. It's only the stupid Java compiler that enforces us to catch everything, because it can't know if the handle might possibly throw a checked exception. When writing scripting languages you just ignore the Throwable when you produce bytecode. 🤓

Here we should catch all possible exceptions that are commonly thrown by plugins: SolrException, Runtime exception, Error and rethrown those. The last catch block would catch Throwable and log it separately as some fatal error.

solr/core/src/java/org/apache/solr/api/AnnotatedApi.java
uschindler: Good catch. The code here is a mess, as you declare the variable as null. You should declare it final before catch block and not assign null. Then compiler would have been complained.

The current code is a horrible antipattern. Always declare variables as final before try and assign once.

solr/core/src/java/org/apache/solr/api/AnnotatedApi.java
uschindler: Yes. For both branches.

In 8.x we have to fix the class accessibility check (unrelated to the if statement).

solr/core/src/java/org/apache/solr/api/AnnotatedApi.java
uschindler: The exception text should not mention internal details. It should just say that the method cannot be called, as it's not accessible (non public, class not visible, different classloader,...)

solr/core/src/java/org/apache/solr/api/AnnotatedApi.java
uschindler: The if statement is not needed, because the public checks are done by unreflect().

This code is from reflection times to catch error early. By using method handles everything is checked and linked early, so you can't get any surprises later. So the public checks is complete obsolete, all this is handled by unreflect.

solr/core/src/java/org/apache/solr/api/AnnotatedApi.java
```diff
@@ -306,7 +313,7 @@ void invoke(SolrQueryRequest req, SolrQueryResponse rsp, CommandOperation cmd) {
       } catch (InvocationTargetException ite) {
```
uschindler, line 313: This catch block can go away!

solr/core/src/java/org/apache/solr/api/AnnotatedApi.java
noblepaul: Thanks, everyone. I wanted to get a quick review for the implementation. Basically, I was not too sure about the idioms we normally use for `MethodHandle` based reflection code.

solr/core/src/java/org/apache/solr/api/AnnotatedApi.java
```diff
@@ -87,16 +83,18 @@ public EndPoint getEndPoint() {
   public static List<Api> getApis(Object obj) {
     return getApis(obj.getClass(), obj);
   }
-  public static List<Api> getApis(Class<? extends Object> klas , Object obj) {
-    if (!Modifier.isPublic(klas.getModifiers())) {
-      throw new RuntimeException(klas.getName() + " is not public");
+  public static List<Api> getApis(Class<? extends Object> theClass , Object obj)  {
+    Class<?> klas = null;
```
uschindler, line 87: this should be declared `final` and not nulled.

noblepaul: how is that possible?

PR #1634, opened 2020-06-30
SOLR-14592: Upgrade Zookeeper to 3.6.1

ErickErickson:
Resolving some of the Zookeeper dependencies was interesting. Zookeeper depends on org.xerial.snappy, commons-lang 2.6 and netty 4.1.48

I had to add org.xerial.snappy as a runtime dependency to get past NoClassDefFound errors that only appear when _running_ either embedded Zookeeper or instantiating a Zookeeper server when running tests. Which we do a lot.

For Ant I needed to do similar, and in addition commons-lang 2.6. Why Gradle apparently uses that automagically is a mystery, but when trying Ant, I'd get a NoClassDefFound error for StringUtils unless I included commons-lang 2.6 explicitly.

Netty was at 4.1.47 originally. I ran into thrashing for checksums until I upgraded netty to at least 4.1.48. Gradle would go ahead and pull this down automatically and update all the relevant checksums, but Ant didn't know about it and would delete all the 4.1.48 checksums and create 4.1.47 ones, which Gradle would delete and add back the ones for 4.1.48 next time updateLicenses was run. So I upgraded it to 4.1.50 on the theory that we might as well get the most recent and now ant and gradle play nice together

Finally, I claim that getMyServerId is unnecessary. It's in there from a long time ago and apparently tries to guess the Zookeeper id in the absence of a myid file, and nobody chimed in when I suggested removing it.


dweiss:
bq. but Ant didn't know about it and would delete all the 4.1.48 checksums and create 4.1.47 ones, which Gradle would delete and add back the ones for 4.1.48 next time updateLicenses was run

gradle build uses palantir's version plugin which tries to figure out the "top" dependency version for a given library across all dependencies. You can ask it which version it picked and why (see help/dependencies.txt).

solr/core/build.gradle
```diff
@@ -138,6 +138,8 @@ dependencies {
 
   testImplementation ('com.sun.jersey:jersey-servlet') { transitive = false }
 
+  // required for instantiating a Zookeeper server in tests or embedded
+  runtimeOnly ('org.xerial.snappy:snappy-java') { transitive = false }
```
dweiss, line 142: I would exclude any dependencies this has explicitly (not by just saying all of transitive dependencies should be omitted).

versions.lock
```diff
@@ -60,14 +61,14 @@ io.dropwizard.metrics:metrics-jmx:4.1.5 (1 constraints: 0c050736)
...
 io.jaegertracing:jaeger-thrift:1.1.0 (1 constraints: 0405f335)
-io.netty:netty-buffer:4.1.47.Final (6 constraints: 29558e4c)
-io.netty:netty-codec:4.1.47.Final (2 constraints: fc135b82)
-io.netty:netty-common:4.1.47.Final (8 constraints: 9d6ed3cf)
-io.netty:netty-handler:4.1.47.Final (2 constraints: 2f176f42)
-io.netty:netty-resolver:4.1.47.Final (3 constraints: ad213fd0)
-io.netty:netty-transport:4.1.47.Final (5 constraints: 7847722a)
-io.netty:netty-transport-native-epoll:4.1.47.Final (2 constraints: 2f176f42)
-io.netty:netty-transport-native-unix-common:4.1.47.Final (2 constraints: 081ad205)
+io.netty:netty-buffer:4.1.50.Final (6 constraints: 0555463a)
```
dweiss, line 64: See here? There are 6 other dependencies that include this one. That's where version consolidation conflicted with ant, probably.

ErickErickson: OK, thanks. Nothing else to do here, though I take it.

solr/core/build.gradle
ErickErickson: I'm not sure there are any. There's probably a nifty way to make Gradle _tell_ me what the dependencies are, and "gradlew dependencies" seems to work for most other packages, but even if I rebuild with transitive = true, then run "gradlew dependencies", there are no dependencies shown for snappy. Hints?

dweiss: gradlew dependencies is the way to go as it shows you exactly which dependencies it includes (for each configuration). This means that compilation-time dependencies may be different than runtime dependencies, for example. For a given configuration it will show you the subtree of dependencies for a given entry. If snappy doesn't have any sub-dependencies then the subtree is empty.

Alternatively, you can take a look at the pom file directly. :)

https://repo1.maven.org/maven2/org/xerial/snappy/snappy-java/1.1.7.6/snappy-java-1.1.7.6.pom

I only see test dependencies and "provided" which isn't included transitively. So you can remove that "transitive = false" block as it doesn't do anything for snappy. It is weird to me that this dependency isn't included by zookeeper itself (it should be!). Perhaps it's because zookeeper is declared with transitive=false. If so then a much better way to solve this would be to remove the explicit snappy dependency and just include it transitively with zookeeper.

PR #1635, opened 2020-07-01
SOLR-14610: ReflectMapWriter to use VarHandle instead of old legacy reflection

noblepaul:


noblepaul:
Pl review @uschindler

PR #1638, opened 2020-07-01
SOLR-14066: Deprecate DIH

chatman:
Deprecating DIH. See SOLR-14066 for details. This is based on @janhoy's previous PR, with minor changes here and there.


madrob:
@chatman Can we close this? It looks like DIH has already been removed in master branch.

PR #1639, opened 2020-07-02
SOLR-14022: Deprecate CDCR

chatman:
See JIRA for more details.


chatman:
It was merged with 8.6. Thanks.

PR #1643, opened 2020-07-02
SOLR-14619 avoid SolrIndexSearcher.getDocSet(q)

dsmiley:
There are places where we get a DocSet from a query when we don't need to.
Also, improved DocSet.getTopFilter to implement DocSetProducer so as to avoid possible re-construction of the DocSet.
Deprecated SolrPluginUtils.numDocs. Nobody calls it.
Maybe @yonik or @mkhludnev might be interested in code reviewing?


dsmiley:
As mentioned in the associated JIRA issue https://issues.apache.org/jira/browse/SOLR-14619 , this PR is blocked up a bit on the Filter score matter.

PR #1647, opened 2020-07-03
SOLR-14626: General improvements to coreContainer shutdown

chatman:
* Removing unnecessary "preClose()"
* Other changes


shalinmangar:
Please be specific about the changes. Why do you think preClose is unnecessary? I see that invocation of `removeEphemeralLiveNode` method is removed entirely. Why do you think that is a safe change to make? With this PR, the live node is not removed until the zk client is closed which happens after pretty much everything is closed. This will cause other clients/replicas to keep sending requests to this node and erroring out.

chatman:
Closing this, all of these improvements are covered in SOLR-14636.

PR #1658, opened 2020-07-07
LUCENE-9423: Handle exception in NIOFSDirectory#openInput

dnhatn:
If we fail to get the size of a file in the constructor of NIOFSIndexInput, then we will leak a FileChannel opened in NIOFSDirectory#openInput.


dnhatn:
@jpountz @s1monw Thanks for reviewing.

PR #1662, opened 2020-07-10
Harden TestBuildingUpMemoryPressure

atris:
1. Add specific checks for exception message expectation.
2. Ensure that a non triggering value is returned when the fake circuit breaker should not be tripping.

Also adds temporary logging for debugging, to be removed.


atris:
Minimally invasive change -- should solve the test failure. I beasted this around 150 times with no failures and the counts do seem to match. Unless objections, I plan to merge this.

PR #1668, opened 2020-07-13
SOLR-13939: Extract any non-gradle related patches (deprecations, URL…

ErickErickson:
These are the thread leak changes from some work Mark Miller did quite a while ago. "gradlew check" passes, BUT I haven't examined the changes at all, I wanted to make it available ASAP for people to look at.

I don't even know if the approach _should_ be pushed. I've just mechanically grouped all the related changes in order to get a well-defined unit to examine.


ErickErickson:
@risdenk  Thanks for looking. This is more about adding thread leak detection to a bunch of classes. When Mark was working on Gradle, a lot of other stuff was changed that weren't strictly related to changing the build system and this JIRA is just about taking any of those changes and making them their own.

Looking over some recent changes, I doubt this'll really cut down on the number of failures due to thread leak detection. OTOH, I don't see any harm in it.

On the other other hand, I have to check inheritance. SolrIgnoredThreadsFilter is included in SolrTestCase, so IDK whether it's necessary to include it in anything that inherits from it... Siiiggghhh.

PR #1669, opened 2020-07-13
SOLR-14151 Make schema components load from packages

noblepaul:


solr/core/src/java/org/apache/solr/pkg/PackageListeners.java
```diff
@@ -96,15 +97,42 @@ private synchronized void invokeListeners(PackageLoader.Package pkg) {
...
+      public void runLater(String name,  Runnable runnable  ) {
+        if(runLater == null) runLater = new LinkedHashMap<>();
+        if(name == null) {
+          name = runnable.getClass().getSimpleName() + "@" + runnable.hashCode();
+        }
+        runLater.put(name, runnable);
+      }
+      private void runLaterTasks(){
+        if(runLater == null) return;
+        new Thread(() -> runLater.forEach((s, runnable) -> {
```
dsmiley, line 127: Could use find a Solr based ExecutorService for this instead?  It sets up MDC and we ensure it gets shut down.

solr/core/src/java/org/apache/solr/pkg/SchemaPluginsLoader.java
```diff
@@ -0,0 +1,160 @@
...
+/**
+ * A {@link SolrClassLoader} that is specifically designed to load schema plugins from packages.
+ * This class would register a listener for any package that is used in a schema and reload the schema
+ * if any of those packages are updated
+ * */
+public class SchemaPluginsLoader implements SolrClassLoader {
+    final CoreContainer coreContainer;
+    final SolrResourceLoader loader;
+    final Function<String, String> pkgVersionSupplier;
+    private Map<String ,PackageAPI.PkgVersion> packageVersions =  new HashMap<>(1);
```
dsmiley, line 42: nit: spacing issue.
Also, please use 'final' for all fields that are final and not just some.  Likewise for 'private' for fields than can be private.

solr/core/src/java/org/apache/solr/core/SolrClassLoader.java
```diff
@@ -0,0 +1,31 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.solr.core;
+
+
+/**A generic interface to load plugin classes
```
dsmiley, line 20: nit: see my longer comment on formatting javadoc.  This case right here really gets to me.

solr/core/src/java/org/apache/solr/schema/IndexSchema.java
```diff
@@ -188,6 +190,7 @@ public IndexSchema(String name, InputSource is, Version luceneVersion, SolrResou
   protected IndexSchema(Version luceneVersion, SolrResourceLoader loader, Properties substitutableProperties) {
     this.luceneVersion = Objects.requireNonNull(luceneVersion);
     this.loader = loader;
+    this.solrClassLoader = loader.getCore() == null? loader: loader.getCore().getSchemaPluginsLoader();
```
dsmiley, line 193: This seems problematic.  The code involved in index schema loading has access to two fields that both implement SolrClassLoader:  `loader` and `solrClassLoader` which you just added.  And then you changed many lines to use SolrClassLoader which just as well could have been as it was before -- `loader` (SRL).  I can see that you're doing this so that a new SchemaPluginsLoader thing could be used.  What if SchemaPluginsLoader was an SRL itself, and delegated the resource-loading methods to the "real" SRL?

Put differently, we could create a synthetic SRL whose methods delegate to an appropriate plugin enabled ClassLoader and a real SRL for the configSet to find resources.  WDYT?

dsmiley: Continuing my idea... Imagine a "ZkConfigSetResourceProvider" and "FileSystemConfigSetResourceProvider" pair of classes implementing an interface ConfigSetResourceProvider that is only for serving a resource (InputStream), not a class.  Just method openResource ideally.  Then imagine a SRL that demands a ConfigSetResourceProvider, and a ClassLoader (or SolrClassLoader as you prefer).  See where this is going?

dsmiley: With that idea, there would only need to be one SRL class (no subclasses), and it'd be easy to create new instances based on those two primary components.  

I'm sure there is some tech debt entanglements in SRL relating to tracking instancePath (there's a TODO I added in there, initLibs() to remove that one) and harder are waitingForCore, infoMBeans, waitingForResources, and of course managedResourceRegistry.  If those get moved off somehow, then I hope the picture I propose becomes more clear.

solr/core/src/java/org/apache/solr/core/SolrResourceLoader.java
```diff
@@ -752,6 +765,9 @@ public void close() throws IOException {
   }
 
 
+  public CoreContainer getCoreContainer(){
```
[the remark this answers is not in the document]
noblepaul: This is necessary for #1666

solr/core/src/java/org/apache/solr/handler/StreamHandler.java
```diff
@@ -158,8 +158,8 @@ public Class getClazz() {
     }
 
     @Override
-    protected void initNewInstance(PackageLoader.Package.Version newest) {
-      clazz = newest.getLoader().findClass(pluginInfo.className, Expressible.class);
+    protected Object initNewInstance(PackageLoader.Package.Version newest) {
+      return clazz = newest.getLoader().findClass(pluginInfo.className, Expressible.class);
```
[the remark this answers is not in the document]
noblepaul: Actually no. it can be anything what the listener wants it to be. In this case it's a class

solr/core/src/java/org/apache/solr/pkg/PackageListeners.java
noblepaul: True. This needs to be done more gracefully.

solr/core/src/java/org/apache/solr/core/SolrCore.java
```diff
@@ -191,6 +191,11 @@
 
   private String name;
   private String logid; // used to show what name is set
+  /**
+   * A unique id to differentiate multiple instances of the same core
+   * If we reload a core, the name remains same , but the id will be new
+   */
+  public final UUID uniqueId = UUID.randomUUID();
```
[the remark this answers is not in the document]
noblepaul: `AtomicLong` is mutable. We need an immutable object

solr/core/src/java/org/apache/solr/core/SolrResourceLoader.java
```diff
@@ -198,6 +206,11 @@ synchronized void reloadLuceneSPI() {
     TokenizerFactory.reloadTokenizers(this.classLoader);
   }
 
+  public SolrCore getCore(){
```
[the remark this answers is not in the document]
noblepaul: This is used in #1666

solr/core/src/java/org/apache/solr/schema/IndexSchema.java
noblepaul: >  What if SchemaPluginsLoader was an SRL itself, and delegated the resource-loading methods to the "real" SRL?

Well, technically it's possible. The current SRL is a mess. At some point in the future we may end up making it clean and usable. Today it's not. We should clearly differentiate between places where we need to load resources and places where we need to load classes. A Minimal interface should be enough for loading classes. SRL is a heavy concrete class.

solr/core/src/java/org/apache/solr/core/SolrCore.java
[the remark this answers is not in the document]
dsmiley: The primitive long instance field would be final (immutable).  A static final AtomicLong (which is mutable) would only be used to generate a new core UID.

solr/core/src/java/org/apache/solr/schema/IndexSchema.java
dsmiley: In the design I propose here, all the debt I mentioned in SRL except instancePath/getConfigDir could move to the new SolrClassLoader because they are class-loading related and not resource-loading related.  InstancePath/getConfigDir is debt I could erase quickly.  Then I think we're in good shape to move forward with a debt-free SRL that is a simple combination of it's two components.  If that sounds nice to you, I could work on a POC immediately.

solr/core/src/java/org/apache/solr/core/ConfigSet.java
```diff
@@ -30,15 +32,18 @@
 
   private final SolrConfig solrconfig;
 
-  private final IndexSchema indexSchema;
+  /**Provide a Schema object on demand
+   * The first Boolean is to signify a a forcefetch
+   */
+  private final Function<Boolean, IndexSchema> indexSchema;
```
dsmiley, line 38: I think using the generic interface Function here is pushing the bounds of readability/intuitiveness.  Lets just provide an interface/class instead.  Perhaps get() and getForce() with no boolean arg needed.

solr/core/src/java/org/apache/solr/core/CoreContainer.java
```diff
@@ -1588,20 +1590,32 @@ private CoreDescriptor reloadCoreDescriptor(CoreDescriptor oldDesc) {
...
+  public void reload(String name) {
+    reload(name, null);
+  }
   /**
    * Recreates a SolrCore.
    * While the new core is loading, requests will continue to be dispatched to
    * and processed by the old core
    *
    * @param name the name of the SolrCore to reload
+   * @param coreId The unique Id of the core
```
dsmiley, line 1606: plus mention it's optional, and then if it's specified and the coreId doesn't match, then the reload is a no-op.  Maybe rename coreId to ifCoreIdMatches to reflect not only what it is but how it's used.

solr/core/src/java/org/apache/solr/handler/SchemaHandler.java
```diff
@@ -202,6 +202,38 @@ private void handleGET(SolrQueryRequest req, SolrQueryResponse rsp) {
     }
   }
 
+  /**If a plugin is loaded from a package, the version of the package being used should be added
```
dsmiley, line 205: nit: another example of oddly formatted javadoc.

solr/core/src/java/org/apache/solr/pkg/MultiPackageListener.java
```diff
@@ -0,0 +1,151 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+import java.util.function.Function;
+/**
+ * A {@link SolrClassLoader} that is designed to listen to a set of packages.
+ * This class would register a listener each package that is loaded through this
+ * if any of those packages are updated , the onReload runnable is executed
+ * */
+public class MultiPackageListener implements SolrClassLoader , PackageListeners.Listener {
```
dsmiley, line 38: I find this confusing... it could just be the name.  Something that implements SolrClassLoader should probably have a name that makes that aspect more pronounced and less so the listener aspect.

solr/core/src/java/org/apache/solr/util/plugin/AbstractPluginLoader.java
```diff
@@ -86,7 +86,7 @@ public AbstractPluginLoader(String type, Class<T> pluginClassType)
    * @param node - the XML node defining this plugin
    */
   @SuppressWarnings("unchecked")
-  protected T create( SolrResourceLoader loader, String name, String className, Node node ) throws Exception
+  protected T create(SolrClassLoader loader, String name, String className, Node node ) throws Exception
```
dsmiley, line 89: BTW I do like these sorts of changes where you've found SRL used where the caller really just needs a class loader.

solr/core/src/java/org/apache/solr/pkg/MultiPackageListener.java
noblepaul: name suggestions are welcome

solr/core/src/java/org/apache/solr/schema/IndexSchema.java
noblepaul: At this point I want this to get out of my way. I would fix this one instead of spending more time on SRL refactoring.

I believe this is in a reasonably good shape. We can do more refactoring later.

Let's do things that add value to users

TBH, I don't care about any particular class here. If possible, I would just nuke SRL altogether. It's just a kitchen sink and reeks of bad designs all over

solr/core/src/java/org/apache/solr/core/SolrResourceLoader.java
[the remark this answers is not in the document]
noblepaul: I may remove it altogether now

solr/core/src/java/org/apache/solr/schema/IndexSchema.java
dsmiley: Okay; lets get this done.

solr/core/src/java/org/apache/solr/core/SolrCore.java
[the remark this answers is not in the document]
noblepaul: UUID is cheap and there is no advantage in using something else. Cores are not created so frequently that we should care too much about it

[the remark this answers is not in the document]
dsmiley: Okay... but the AtomicLong incrementAndGet technique is _very_ common for for an internal-id use case (I could list others in Lucene/Solr).  I have not seen a UUID used yet for this.  Up to you.

solr/core/src/java/org/apache/solr/core/ConfigSet.java
```diff
@@ -79,4 +75,14 @@ public NamedList getProperties() {
   public boolean isTrusted() {
     return trusted;
   }
+
+  /**Provide a Schema object on demand
```
dsmiley, line 79: https://google.github.io/styleguide/javaguide.html#s7-javadoc

Glad to see this supplier though.

solr/core/src/java/org/apache/solr/pkg/PackageListeningClassLoader.java
```diff
@@ -35,7 +35,7 @@
  * This class would register a listener each package that is loaded through this
  * if any of those packages are updated , the onReload runnable is executed
  * */
-public class MultiPackageListener implements SolrClassLoader , PackageListeners.Listener {
+public class PackageListeningClassLoader implements SolrClassLoader , PackageListeners.Listener {
```
dsmiley, line 38: Good name.

PR #1671, opened 2020-07-14
LUCENE-9427: Ensure unified highlighter considers all terms in fuzzy query.

jtibshirani:
This fixes a regression where if a fuzzy query corresponds to an exact match
(for example it has maxEdits: 0), then the unified highlighter doesn't produce
highlights for the matching terms.

The proposed fix is to always pass back an automaton when visiting a fuzzy
query. This seems to match the previous behavior before the regression was
introduced.

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -136,6 +136,9 @@ Bug fixes
 * LUCENE-9365: FuzzyQuery was missing matches when prefix length was equal to the term length
   (Mark Harwood, Mike Drob)
 
+* LUCENE-9427: Fix a regression where the unified highlighter didn't produce
+  highlights on fuzzy queries that correspond to exact matches. (Julie Tibshirani)
+
 Other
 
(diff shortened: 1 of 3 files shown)


jtibshirani:
It looks like this behavior changed during the `QueryVisitor` refactor: https://github.com/apache/lucene-solr/pull/581/files#diff-449ea2758bc22783ca3f819096e1b638R97. Previously we always extracted an automaton from fuzzy queries.

romseygeek:
Related: https://issues.apache.org/jira/browse/LUCENE-9274

I think ideally the fix would be https://issues.apache.org/jira/browse/LUCENE-9277, but this works as a stopgap.

jtibshirani:
Thanks @romseygeek for taking a look. What do you see as the next steps? Would it make sense to merge this, and I could look into LUCENE-9277 in a follow-up?

jtibshirani:
I just pushed a changelog update.

romseygeek:
Merged to master in 688583fc2d01c39bba63d19cf57bb5720eda1afd

PR #1672, opened 2020-07-15
SOLR-14651: Metrics History could disable better

dsmiley:
* SolrRrdBackendFactory should not be created if history is disabled
* Disable MetricsHistoryHandler by default in tests
* Await shutdown of all executors

Interestingly, all tests passed with the history handler disabled.  Isn't that a problem @sigram ?

Somewhat related: I'm also inclined to make a little refactor in CoreContainer.shutdown such that `IOUtils.closeQuietly(metricsHistoryHandler.getSolrClient());` can move _into_ metricsHistoryHandler.close()


dsmiley:
You might like this one @markrmiller

dsmiley:
I see @markrmiller  did something similar at about the same time: https://github.com/apache/lucene-solr/commit/a7b9847a7e98c27d74b13ac0a46d3e36945fbba3 but instead more completely disabled it, not just the RRD aspect.  And it's purely for tests, as opposed to the PR here that has a production aspect if you choose to configure to disable the RRD.

dsmiley:
@sigram can I get a review please?

solr/solrj/src/java/org/apache/solr/common/util/ExecutorUtil.java
```diff
@@ -77,6 +77,12 @@ public static void shutdownAndAwaitTermination(ExecutorService pool) {
     awaitTermination(pool);
   }
 
+  public static void shutdownNowAndAwaitTermination(ExecutorService pool) {
+    if(pool == null) return;
```
sigram, line 81: missing space after `if`

PR #1678, opened 2020-07-16
SOLR-14613: Initial POC based on Ilan's proposal.

sigram:
See Jira for background.

This PR is based on Ilan's proposal for a minimal API that is totally independent of Solr internals. I renamed some of the interfaces for convenience, and implemented a few concrete classes (POJOs) to pass data around.

So far I implemented a round-trip shim between SolrCloudManager and AssignerClusterState and its components - see TestPolicy8x for details. This is sufficient to support the Policy engine calculations using the new API + shim. 

I didn't start converting the Solr side of things to actually provide the new APIs to the Assigner implementations, and I didn't hook up the creation of Assigner yet.

### DIFF
--- a/solr/core/src/java/org/apache/solr/cloud/api/collections/assign/AddReplicaRequest.java
+++ b/solr/core/src/java/org/apache/solr/cloud/api/collections/assign/AddReplicaRequest.java
@@ -1,0 +1,72 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.solr.cloud.api.collections.assign;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+
+import org.apache.solr.common.cloud.Replica;
+
+/**
+ *
+ */
+public class AddReplicaRequest extends BaseAssignRequest {
+  private final Replica.Type type;
+  private final Map<String, Object> params = new HashMap<>();
+  private final String targetNode;
+  private final String coreName;
+  private final Set<String> nodeSet;
+
+  public AddReplicaRequest(String collection, String shard, Replica.Type type, Map<String, Object> params,
+                           String targetNode, String coreName, Set<String> nodeSet) {
+    super(collection, shard);
+    this.type = type;
+    if (params != null) {
+      this.params.putAll(params);
+    }
+    this.targetNode = targetNode;
+    this.coreName = coreName;
+    this.nodeSet = nodeSet;
+    Objects.requireNonNull(this.type, "'type' must not be null");
+    Objects.requireNonNull(this.coreName, "'coreName' must not be null");
+  }
+
+  public Replica.Type getType() {
+    return type;
+  }
+
+  public Map<String, Object> getParams() {
+    return params;
+  }
+
+  // impls may request a specific target node
+  public String getTargetNode() {
+    return targetNode;
+  }
+
+  public String getCoreName() {
+    return coreName;
+  }
+
+  // subset of live nodes to consider as valid targets, or null
+  public Set<String> getNodeSet() {
+    return nodeSet;
+  }
+}
--- a/solr/core/src/java/org/apache/solr/cloud/api/collections/assign/AddReplicaRequest.java
+++ b/solr/core/src/java/org/apache/solr/cloud/api/collections/assign/AddReplicaRequest.java
@@ -22,18 +22,16 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
 
-import org.apache.solr.common.cloud.Replica;
-
 /**
  *
  */
 public class AddReplicaRequest extends BaseAssignRequest {
-  private final Replica.Type type;
+  private final ReplicaType type;
   private final Map<String, Object> params = new HashMap<>();
   private final String targetNode;
   private final Set<String> nodeSet;
 
-  public AddReplicaRequest(String collection, String shard, Replica.Type type, Map<String, Object> params,
+  public AddReplicaRequest(String collection, String shard, ReplicaType type, Map<String, Object> params,
                            String targetNode, Set<String> nodeSet) {
     super(collection, shard);
     this.type = type;
@@ -45,7 +43,7 @@ public class AddReplicaRequest extends BaseAssignRequest {
     Objects.requireNonNull(this.type, "'type' must not be null");
   }
 
-  public Replica.Type getType() {
+  public ReplicaType getType() {
     return type;
   }
 
--- a/solr/core/src/java/org/apache/solr/cloud/api/collections/assign/AddReplicaRequest.java
+++ b/solr/core/src/java/org/apache/solr/cloud/api/collections/assign/AddReplicaRequest.java
@@ -31,21 +31,18 @@ public class AddReplicaRequest extends BaseAssignRequest {
   private final Replica.Type type;
   private final Map<String, Object> params = new HashMap<>();
   private final String targetNode;
-  private final String coreName;
   private final Set<String> nodeSet;
 
   public AddReplicaRequest(String collection, String shard, Replica.Type type, Map<String, Object> params,
-                           String targetNode, String coreName, Set<String> nodeSet) {
+                           String targetNode, Set<String> nodeSet) {
     super(collection, shard);
     this.type = type;
     if (params != null) {
       this.params.putAll(params);
     }
     this.targetNode = targetNode;
-    this.coreName = coreName;
     this.nodeSet = nodeSet;
     Objects.requireNonNull(this.type, "'type' must not be null");
-    Objects.requireNonNull(this.coreName, "'coreName' must not be null");
   }
 
   public Replica.Type getType() {
@@ -59,10 +56,6 @@ public class AddReplicaRequest extends BaseAssignRequest {
   // impls may request a specific target node
   public String getTargetNode() {
     return targetNode;
-  }
-
-  public String getCoreName() {
-    return coreName;
   }
 
   // subset of live nodes to consider as valid targets, or null
--- a/solr/core/src/java/org/apache/solr/cloud/api/collections/assign/AssignDecision.java
+++ b/solr/core/src/java/org/apache/solr/cloud/api/collections/assign/AssignDecision.java
@@ -1,0 +1,26 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.solr.cloud.api.collections.assign;
+
+/**
+ *
+ */
+public interface AssignDecision {
+
+  AssignRequest getAssignRequest();
+
+}
--- a/solr/core/src/java/org/apache/solr/cloud/api/collections/assign/AssignDecisions.java
+++ b/solr/core/src/java/org/apache/solr/cloud/api/collections/assign/AssignDecisions.java
@@ -1,0 +1,41 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.solr.cloud.api.collections.assign;
+
+import java.util.Collection;
+
+/**
+ *
+ */
+public class AssignDecisions {
+
+  private final AssignerClusterState outputClusterState;
+  private final Collection<AssignDecision> decisions;
+
+  public AssignDecisions(AssignerClusterState outputClusterState, Collection<AssignDecision> decisions) {
+    this.outputClusterState = outputClusterState;
+    this.decisions = decisions;
+  }
+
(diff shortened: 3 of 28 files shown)


solr/core/src/java/org/apache/solr/cloud/api/collections/assign/AddReplicaRequest.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+
+import org.apache.solr.common.cloud.Replica;
+
+/**
+ *
+ */
+public class AddReplicaRequest extends BaseAssignRequest {
```
murblanc, line 30: Requests for work such as this one (passed from Solr to the plug-in implementing the assignment decisions) should likely be interfaces, not classes.
Also I'd prefer `Replica.Type` to be made a toplevel class (enum) so we don't have to expose `Replica` to external code (as `Replica` brings many other classes with it)

solr/core/src/java/org/apache/solr/cloud/api/collections/assign/Assigner.java
```diff
@@ -0,0 +1,30 @@
...
+
+import java.util.List;
+
+/**
+ * Interface implemented by assign strategies.
+ */
+public interface Assigner {
+
+  AssignDecisions computeAssignments(AssignerClusterState initialState,
+                                     List<AssignRequest> requests) throws InterruptedException, AssignerException;
```
murblanc, line 28: Nice touch here to spare the plugin from having to deal with interruptions!

solr/core/src/java/org/apache/solr/cloud/api/collections/assign/AssignDecision.java
```diff
@@ -0,0 +1,26 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.solr.cloud.api.collections.assign;
+
+/**
+ *
+ */
+public interface AssignDecision {
```
murblanc, line 22: Added value of this interface unclear to me given we have a base class and concrete implementations.
Would make more sense IMO if we had interfaces for all possible decision types, and provide actual concrete implementations as an optional convenience to the plugin writer.

solr/core/src/java/org/apache/solr/cloud/api/collections/assign/AssignerNodeState.java
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ *
+ */
+public interface AssignerNodeState {
+  Collection<AssignerReplica> getReplicas();
+
+  long getTotalDiskGB();
+
+  long getFreeDiskGB();
```
murblanc, line 31: Maybe make number of cores an explicit metric as well?
Note in current Autoscaling framework it is always requested, as is free disk space.

Unclear to me how total disk space would be used by the plugin.

solr/core/src/java/org/apache/solr/cloud/api/collections/assign/AssignerReplica.java
```diff
@@ -0,0 +1,86 @@
...
+ * Represents basic information about a replica.
+ */
+public class AssignerReplica {
+  private final String name;
+  private final String node;
+  private final String collection;
+  private final String shard;
+  private final String core;
+  private final Replica.Type type;
+  private final Replica.State state;
```
murblanc, line 34: I'd prefer Type and State to be moved outside of an implementation class so won't bring any dependencies or leak any internal class to the plugin.

solr/core/src/java/org/apache/solr/cloud/api/collections/assign/AssignDecisions.java
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ *
+ */
+public class AssignDecisions {
+
+  private final AssignerClusterState outputClusterState;
+  private final Collection<AssignDecision> decisions;
+
+  public AssignDecisions(AssignerClusterState outputClusterState, Collection<AssignDecision> decisions) {
```
murblanc, line 29: Why should the plugin pass an `outputClusterState`?

murblanc: The Solr side support framework (didn't read all classes yet) might want to apply the decisions to another cluster state maintained centrally, used as a "master copy" including all changes from all computations and passed (in an immutable form so the plugin doesn't become crazy) to the next computation. Internally we don't have to make copies of everything, we can share data but give consistent point in time reads.

solr/core/src/java/org/apache/solr/cloud/api/collections/assign/AssignerReplica.java
murblanc: (or wrap them if we don't want to touch Replica at this stage)

solr/core/src/java/org/apache/solr/cloud/api/collections/assign/DeleteReplicaDecision.java
```diff
@@ -0,0 +1,26 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.solr.cloud.api.collections.assign;
+
+/**
+ *
+ */
+public class DeleteReplicaDecision extends BaseAssignDecision {
```
murblanc, line 22: Missing details on what to delete here, no? Otherwise what's the plugin role in the delete?

solr/core/src/java/org/apache/solr/cloud/api/collections/assign/DeleteReplicaRequest.java
```diff
@@ -0,0 +1,33 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.solr.cloud.api.collections.assign;
+
+/**
+ *
+ */
+public class DeleteReplicaRequest extends BaseAssignRequest {
```
murblanc, line 22: Likely a delete request should only specify the shard and let the plugin pick which replica to kill...
Otherwise no need to ask the plugin, just delete the replica directly.

solr/core/src/java/org/apache/solr/cloud/api/collections/assign/MoveReplicaRequest.java
```diff
@@ -0,0 +1,28 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.solr.cloud.api.collections.assign;
+
+/**
+ *
+ */
+public interface MoveReplicaRequest extends AssignRequest {
```
murblanc, line 22: What is the plugin asked to decide/compute here?

solr/core/src/java/org/apache/solr/cloud/api/collections/assign/policy8x/AssignerCloudManager.java
```diff
@@ -0,0 +1,81 @@
...
+import org.apache.solr.client.solrj.cloud.SolrCloudManager;
+import org.apache.solr.client.solrj.impl.ClusterStateProvider;
+import org.apache.solr.cloud.api.collections.assign.AssignerClusterState;
+import org.apache.solr.common.util.ObjectCache;
+import org.apache.solr.common.util.TimeSource;
+
+/**
+ *
+ */
+public class AssignerCloudManager implements SolrCloudManager {
```
murblanc, line 20: This is an internal class, right? The plugins do not see it?

solr/core/src/java/org/apache/solr/cloud/api/collections/assign/policy8x/AssignerCloudManager.java
```diff
@@ -0,0 +1,81 @@
...
+  private final ObjectCache objectCache = new ObjectCache();
+  private final TimeSource timeSource;
+  private final AssignerClusterState assignerClusterState;
+  private final AssignerClusterStateProvider clusterStateProvider;
+  private final AssignerNodeStateProvider nodeStateProvider;
+  private final AssignerDistribStateManager distribStateManager;
+
+  public AssignerCloudManager(AssignerClusterState assignerClusterState,
```
murblanc, line 28: I don't get the logic. I'm assuming package `*.policy8x` is internal. It should not depend on state as seen by the plugins but should **build** that state based on internal SolrCloud state.
Here and in other classes in this package I have the impression we do things the wrong way?

murblanc:
You lost me when I got to the policy8x package classes.
My understanding is that the assigner driving the plugins would build the state as seen by the plugins from SolrCloud internal state, would build the requests as seen by the plugins from the internal requests (Hint thing most likely?), then the plugin will be called, then the decisions will be unpacked into SolrCloud actions.

Internal code should not have to unpack the state in the format used by the plugins... Am I missing something?

solr/core/src/java/org/apache/solr/cloud/api/collections/assign/AddReplicaRequest.java
```diff
@@ -0,0 +1,72 @@
...
+  private final Replica.Type type;
+  private final Map<String, Object> params = new HashMap<>();
+  private final String targetNode;
+  private final String coreName;
```
noblepaul, line 34: do you mean the replica Name? The core name and the replica name are not required. The system always assign a unique name

solr/core/src/java/org/apache/solr/cloud/api/collections/assign/AssignDecision.java
noblepaul: I would prefer interfaces.

solr/core/src/java/org/apache/solr/cloud/api/collections/assign/AssignerNodeState.java
sigram: Yes, we should add cores here. Freedisk and totaldisk are very much needed to decide where to move replicas - you want to move them to a node that has enough free disk, and one that is not too "full".

solr/core/src/java/org/apache/solr/cloud/api/collections/assign/AssignerReplica.java
sigram: I'll wrap them. Moving Type and State would cause massive diffs across the whole code base.

solr/core/src/java/org/apache/solr/cloud/api/collections/assign/DeleteReplicaDecision.java
sigram: All necessary details are already in the enclosed DeleteReplicaRequest.

solr/core/src/java/org/apache/solr/cloud/api/collections/assign/DeleteReplicaRequest.java
sigram: That's not always the case - when you want to decommission a node you know exactly what replicas to delete.

solr/core/src/java/org/apache/solr/cloud/api/collections/assign/MoveReplicaDecision.java
```diff
@@ -0,0 +1,36 @@
...
+ *
+ */
+public interface MoveReplicaDecision extends AssignDecision {
+  String getSourceNodeName();
+
+  String getTargetNodeName();
+
+  String getCollectionName();
+
+  String getSliceName();
```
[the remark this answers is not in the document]
sigram: Yes :)

solr/core/src/java/org/apache/solr/cloud/api/collections/assign/MoveReplicaRequest.java
sigram: In some cases you know exactly where you want to move a replica. In other cases you're asking for the best place other than the current place. This supports both scenarios - if `targetNode` is specified then this translates directly to a MOVEREPLICA cmd without any computation. If `targetNode` is null then the plugin needs to compute the best new placement.

solr/core/src/java/org/apache/solr/cloud/api/collections/assign/policy8x/AssignerCloudManager.java
sigram: The current Policy engine relies on `SolrCloudManager` API. This class is a shim - it provides a synthetic `SolrCloudManager` that uses data from `AssignerClusterState`, and nothing else.

solr/core/src/java/org/apache/solr/cloud/api/collections/assign/policy8x/AssignerCloudManager.java
sigram: This package is a proof of concept that the Policy engine can fully function as a plugin when it only has the information provided by the new API (`AssignerClusterState`).

solr/core/src/java/org/apache/solr/cloud/api/collections/assign/DeleteReplicaDecision.java
murblanc: So why are we talking to the plugin for this?

solr/core/src/java/org/apache/solr/cloud/api/collections/assign/DeleteReplicaRequest.java
murblanc: Again, if it's just a delete and we already know what to delete, the plugin doesn't add much value.
Solr calls plugins to ask a question and the plugins provide answers. What's the question and what are the possible answers?

solr/core/src/java/org/apache/solr/cloud/api/collections/assign/AddReplicaRequest.java
sigram: Right, I'll fix this.

solr/core/src/java/org/apache/solr/cloud/api/collections/assign/policy8x/AssignerCloudManager.java
murblanc: I finally got that following our Slack exchange... I'm not sure that's a proof this will really work in a non trivial case, because we're missing the logic that goes around properties/variables and their types.

solr/core/src/java/org/apache/solr/cloud/api/collections/assign/AssignerNodeState.java
```diff
@@ -0,0 +1,35 @@
...
+
+  int getNumCores();
+
+  Map<String, Object> getProperties();
```
noblepaul, line 34: What are these properties? javadocs missing?

solr/core/src/java/org/apache/solr/cloud/api/collections/assign/AssignerShardState.java
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ *
+ */
+public interface AssignerShardState {
+  String getName();
+
+  Collection<AssignerReplica> getReplicas();
+
+  Map<String, Object> getProperties();
```
noblepaul, line 30: what are shard properties? Examples?

solr/core/src/java/org/apache/solr/cloud/api/collections/assign/ReplicaState.java
```diff
@@ -0,0 +1,62 @@
...
+ * limitations under the License.
+ */
+package org.apache.solr.cloud.api.collections.assign;
+
+import java.util.Locale;
+
+/**
+ *
+ */
+public enum ReplicaState {
```
noblepaul, line 24: How is it any different from `Replia.State` ?

solr/core/src/java/org/apache/solr/cloud/api/collections/assign/ReplicaType.java
```diff
@@ -0,0 +1,47 @@
...
+ * limitations under the License.
+ */
+package org.apache.solr.cloud.api.collections.assign;
+
+import java.util.Locale;
+
+/**
+ *
+ */
+public enum ReplicaType {
```
noblepaul, line 24: How is this different from `Replica.Type` ?

solr/core/src/java/org/apache/solr/cloud/api/collections/assign/ReplicaState.java
sigram: No difference. This is just to separate this api from Replica class. I considered moving the inner Replica.State and Replica.Type to top level but this causes massive diffs all over the place. Eventually we may want to do this but now it would just obscure the important changes.

sigram:
Another variant of this has already been merged.

PR #1682, opened 2020-07-20
Use FileSystem.newInstance instead of get in HDFSBackupRepository

atris:
FileSystem.get can cause FileSystem closed exceptions, especially with S3. Using FileSystem.newInstance should help that case.


atris:
@mkhludnev Let me know if this looks fine

atris:
Tests pass -- no regressions. @mkhludnev Any thoughts?

chatman:
I think we should have corresponding JIRA issues for all changes. This helps keep track of all improvements in a release.

PR #1686, opened 2020-07-21
SOLR-13528: Implement Request Rate Limiters

atris:
This commit introduces two functionalities: request rate limiting and ability to identify requests based on type (indexing, search, admin). The default rate limiter rate limits query requests based on configurable parameters which can be set in web.xml. Note that this rate limiting works at a JVM level, not a core/collection level.


atris:
@anshumg @madrob @sigram @ErickErickson Please take a look

solr/core/src/java/org/apache/solr/servlet/RateLimitManager.java
```diff
@@ -0,0 +1,153 @@
...
+ */
+public class RateLimitManager {
+  public final static int DEFAULT_CONCURRENT_REQUESTS= 10;
+  public final static long DEFAULT_EXPIRATION_TIME_INMS = 300;
+  public final static long DEFAULT_SLOT_ACQUISITION_TIMEOUT_MS = -1;
+
+  private final Map<String, RequestRateLimiter> requestRateLimiterMap;
+
+  public RateLimitManager() {
+    this.requestRateLimiterMap = new HashMap<String, RequestRateLimiter>();
```
madrob, line 47: nit: use `<>` instead of explicit types.

solr/core/src/java/org/apache/solr/servlet/RequestRateLimiter.java
```diff
@@ -0,0 +1,199 @@
...
+      @Override
+      public void onTimeout(AsyncEvent asyncEvent) throws IOException {
+        AsyncContext asyncContext = asyncEvent.getAsyncContext();
+
+        if (!waitQueue.remove(asyncContext)) {
+          return;
+        }
+
+        HttpServletResponse servletResponse = ((HttpServletResponse)asyncEvent.getSuppliedResponse());
+        String responseMessage = "Too many requests for this request type." +
```
madrob, line 134: Unused.

solr/core/src/java/org/apache/solr/servlet/RequestRateLimiter.java
```diff
@@ -0,0 +1,199 @@
...
+  private Semaphore allowedConcurrentRequests;
+  private RateLimiterConfig rateLimiterConfig;
+  private Queue<AsyncContext> waitQueue;
+  private Queue<AsyncListener> listenerQueue;
+
+  public RequestRateLimiter(RateLimiterConfig rateLimiterConfig) {
+    this.rateLimiterConfig = rateLimiterConfig;
+    this.allowedConcurrentRequests = new Semaphore(rateLimiterConfig.allowedRequests);
+    this.waitQueue = new ConcurrentLinkedQueue<>();
+    this.listenerQueue = new ConcurrentLinkedQueue<>();
```
madrob, line 58: We never poll this queue, what is it used for?

solr/core/src/java/org/apache/solr/servlet/SolrDispatchFilter.java
```diff
@@ -331,6 +338,10 @@ public void close() {
           metricManager = null;
         }
       }
+
+      if (rateLimitManager != null) {
```
madrob, line 342: should this be in the finally block instead?

solr/core/src/java/org/apache/solr/servlet/SolrDispatchFilter.java
```diff
@@ -377,6 +389,16 @@ public void doFilter(ServletRequest _request, ServletResponse _response, FilterC
         }
       }
 
+      try {
+        accepted = rateLimitManager.handleRequest(request);
+      } catch (InterruptedException e) {
+        throw new RuntimeException(e.getMessage());
```
madrob, line 395: SolrException or SolrServerException probably

solr/core/src/java/org/apache/solr/servlet/RateLimitManager.java
```diff
@@ -0,0 +1,153 @@
...
+      // No rate limiter for this request type
+      return;
+    }
+
+    // Give preference to the previous request's rate limiter
+    if (previousRequestRateLimiter.resumePendingOperation()) {
+      return;
+    }
+
+    for (Map.Entry<String, RequestRateLimiter> currentEntry : requestRateLimiterMap.entrySet()) {
```
madrob, line 99: I'm confused as to why this exists. If a query request completes, why does that mean that we have room for an update request now?

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrClient.java
```diff
@@ -358,7 +358,11 @@ protected HttpRequestBase createMethod(@SuppressWarnings({"rawtypes"})SolrReques
     if (parser == null) {
       parser = this.parser;
     }
-    
+
+    Header[] contextHeaders = new Header[2];
+    contextHeaders[0] = new BasicHeader(CommonParams.SOLR_REQUEST_CONTEXT_PARAM, getContext().toString());
```
madrob, line 363: This seems trivial to forge. Can we tie this into the authentication code paths somehow so that we can really trust that the request is coming from a server instead of just some application claiming to be a server?

solr/solr-ref-guide/src/rate-limiters.adoc
```diff
@@ -0,0 +1,101 @@
...
+If a request exceeds the request quota, further incoming requests are automatically queued asynchronously with
+a configurable timeout.
+
+== When To Use Rate Limiters
+Rate limiters should be used when the user wishes to allocate a guaranteed capacity of the request threadpool to a specific
+request type. Indexing and search requests are mostly competing with each other for CPU resources. This becomes especially
+pronounced under high stress in production workloads.
+
+== Rate Limiter Configurations
+The default rate limiter is search rate limiter. Accordingly, it can be configured in web.xml under initParams for
```
madrob, line 31: Is there anywhere else that we can configure this? I thought the trend was to move away from putting things in `web.xml` as much as possible.

solr/solr-ref-guide/src/rate-limiters.adoc
```diff
@@ -0,0 +1,101 @@
...
+----
+<param-name>isQueryRateLimiterEnabled</param-name>
+----
+[source,xml]
+----
+<param-value>true</param-value>
+----
+
+==== Maximum Number Of Concurrent Requests
+Allows setting maximum concurrent search requests at a given point in time. Default value is 10.
```
madrob, line 51: This is a maximum number for the JVM? Not collection or core specific? 10 seems like a very low default.

solr/solr-ref-guide/src/rate-limiters.adoc
```diff
@@ -0,0 +1,101 @@
...
+[source,xml]
+----
+<param-value>15</param-value>
+----
+
+==== Request Slot Allocation Wait Time
+Wait time in ms for which a request will wait for a slot to be available when all slots are full,
+before the request is put into the wait queue. This allows requests to have a chance to proceed if
+the unavailability of the request slots for this rate limiter is a transient phenomenon. Default value
+is -1, indicating no wait.
```
madrob, line 65: What does a value of 0 mean?

solr/solr-ref-guide/src/rate-limiters.adoc
```diff
@@ -0,0 +1,101 @@
...
+----
+<param-name>queryWaitForSlotAllocationInMS</param-name>
+----
+[source,xml]
+----
+<param-value>100</param-value>
+----
+
+==== Request Expiration Time
+Time in ms after which a request will expire in the wait queue. Default value is 200 ms.
```
madrob, line 76: What does 0 mean here? Do negative values mean no expiration?

solr/solr-ref-guide/src/rate-limiters.adoc
```diff
@@ -0,0 +1,101 @@
...
+These are some of the things to keep in mind when using rate limiters
+
+=== Over Subscribing
+It is possible to define a size of quota for a request type which exceeds the size
+of the available threadpool. Solr does not enforce rules on the size of a quota that
+can be define for a request type. This is intentionally done to allow users full
+control on their quota allocation. However, if the quota exceeds the available threadpool's
+size, the standard queuing policies of the threadpool will kick in.
+
+=== Work Stealing
```
madrob, line 97: This should be configurable to enable/disable

solr/core/src/java/org/apache/solr/servlet/SolrDispatchFilter.java
```diff
@@ -441,6 +463,13 @@ public void doFilter(ServletRequest _request, ServletResponse _response, FilterC
       consumeInputFully(request, response);
       SolrRequestInfo.reset();
       SolrRequestParsers.cleanupMultipartFiles(request);
+
+      if (accepted) {
+        rateLimitManager.decrementActiveRequests(request);
```
madrob, line 468: There is a race condition here where new incoming requests can get in before resume of an existing request.

solr/core/src/java/org/apache/solr/servlet/RequestRateLimiter.java
```diff
@@ -0,0 +1,199 @@
...
+  public boolean handleRequest(HttpServletRequest request) throws InterruptedException {
+
+    if (!rateLimiterConfig.isEnabled) {
+      return true;
+    }
+
+    boolean accepted = allowedConcurrentRequests.tryAcquire(rateLimiterConfig.waitForSlotAcquisition, TimeUnit.MILLISECONDS);
+
+    if (!accepted) {
+      AsyncContext asyncContext = request.startAsync();
```
madrob, line 70: Is this different from admin requests sent with `async=[async-id]`?

solr/core/src/java/org/apache/solr/servlet/RequestRateLimiter.java
atris: This exists to store the listeners that are registered for the given async contexts. Removed the same, thanks

solr/core/src/java/org/apache/solr/servlet/SolrDispatchFilter.java
atris: Would that not defy the purpose of the destroy method?

solr/core/src/java/org/apache/solr/servlet/RateLimitManager.java
atris: This is work stealing -- a thread first tries to get work from its own request rate limiter and if not present, let the thread be used for a different request type. This is to allow smoothening out of skews. Added a flag to enable or disable this behaviour.

solr/solr-ref-guide/src/rate-limiters.adoc
atris: Unfortunately, I do not think that SolrDispatchFilter has access to solr-config.xml.

solr/solr-ref-guide/src/rate-limiters.adoc
atris: Changed to number of cores * 3.

solr/solr-ref-guide/src/rate-limiters.adoc
atris: Yes. Updated the docs to state the same.

solr/solr-ref-guide/src/rate-limiters.adoc
atris: Yes, updated to say the same.

solr/core/src/java/org/apache/solr/servlet/RequestRateLimiter.java
atris: I dont think so -- an async request should just return its current context here.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrClient.java
atris: I am honestly not sure about that -- Can you suggest a mechanism to achieve that? Ideas are more than welcome!

solr/core/src/java/org/apache/solr/servlet/SolrDispatchFilter.java
atris: Redundant now, thanks

atris:
@madrob I spent some time thinking and decided that we should not be blocking rejected requests but instead returning them immediately with 429 (Too Many Requests) HTTP code. Here is why:

1. User should have the ability to control the behaviour in case of a quota breach.
2. Adding queuing logic to Solr seems as an anti pattern -- are we over stretching the boundaries of our responsibility?
3. This might create a false impression that Solr has slowed down since request time will asymptotically increase before timeouts begin to happen. Rejecting upfront will keep the users with reality.
4. In many scenarios, it is important for the user to know the status of their request usage so as to adjust accordingly.

Accordingly, I have updated the PR. Apologies for the delay. Please take a look and let me know your thoughts and comments.

solr/core/src/java/org/apache/solr/servlet/RateLimitManager.java
```diff
@@ -0,0 +1,148 @@
...
+    String requestContext = request.getHeader(SOLR_REQUEST_CONTEXT_PARAM);
+    String typeOfRequest = request.getHeader(SOLR_REQUEST_TYPE_PARAM);
+
+    if (typeOfRequest == null) {
+      // Cannot determine if this request should be throttled
+      return true;
+    }
+
+    // Do not throttle internal requests
+    if (requestContext != null && requestContext.equals(SolrRequest.SolrClientContext.SERVER.toString())) {
```
anshumg, line 65: This is really clean!

solr/core/src/java/org/apache/solr/servlet/SolrDispatchFilter.java
```diff
@@ -377,6 +385,20 @@ public void doFilter(ServletRequest _request, ServletResponse _response, FilterC
...
+      } catch (InterruptedException e) {
+        throw new SolrException(ErrorCode.SERVER_ERROR, e.getMessage());
+      }
+
+      if (!accepted) {
+        String errorMessage = "Too many requests for this request type." +
+            "Please try after some time or increase the quota for this request type";
+
+        response.sendError(429, errorMessage);
+        //throw new SolrException(ErrorCode.TOO_MANY_REQUESTS, "FOOFOOOFOO");
```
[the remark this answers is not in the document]
atris: Good catch, thanks!

atris:
@anshumg Thank you so much for the review. I have updated per your comments -- please see and let me know your thoughts and comments.

solr/core/src/test/org/apache/solr/servlet/TestRequestRateLimiter.java
```diff
@@ -0,0 +1,185 @@
...
+    CloudSolrClient client = cluster.getSolrClient();
+    client.setDefaultCollection(COLLECTION);
+
+    CollectionAdminRequest.createCollection(COLLECTION, 1, 1).process(client);
+    cluster.waitForActiveCollection(COLLECTION, 1, 1);
+
+    SolrDispatchFilter solrDispatchFilter = cluster.getJettySolrRunner(0).getSolrDispatchFilter();
+
+    RequestRateLimiter.RateLimiterConfig rateLimiterConfig = new RequestRateLimiter.RateLimiterConfig(SolrRequest.SolrRequestType.QUERY,
+        true, 1, DEFAULT_SLOT_ACQUISITION_TIMEOUT_MS, 5 /* allowedRequests */, true /* isWorkStealingEnabled */);
```
anshumg, line 59: s/stealing/borrowing

atris:
Thanks @anshumg . Fixed the same.

solr/solr-ref-guide/src/rate-limiters.adoc
```diff
@@ -0,0 +1,116 @@
...
+[source,xml]
+----
+<param-value>15</param-value>
+----
+
+==== Request Slot Allocation Wait Time
+Wait time in ms for which a request will wait for a slot to be available when all slots are full,
+before the request is put into the wait queue. This allows requests to have a chance to proceed if
+the unavailability of the request slots for this rate limiter is a transient phenomenon. Default value
+is -1, indicating no wait. 0 will represent the same -- no wait.
```
madrob, line 66: Add a indication here that setting higher slot allocation wait times may be reflected as higher QTime. It's probably apparent to us now, but somebody debugging why their queries are getting slow will likely appreciate the hint.

solr/solr-ref-guide/src/rate-limiters.adoc
```diff
@@ -0,0 +1,116 @@
...
+=== Over Subscribing
+It is possible to define a size of quota for a request type which exceeds the size
+of the available threadpool. Solr does not enforce rules on the size of a quota that
+can be define for a request type. This is intentionally done to allow users full
+control on their quota allocation. However, if the quota exceeds the available threadpool's
+size, the standard queuing policies of the threadpool will kick in.
+
+=== Slot Borrowing
+If a quota does not have backlog but other quotas do, then the relatively less busier quota can
+"borrow" slot from the busier quotas. This is done on a round robin basis today with a futuristic
```
madrob, line 114: Is there a JIRA for this? Would be good to link to it if it exists (or create it if it does not)

solr/solr-ref-guide/src/rate-limiters.adoc
```diff
@@ -0,0 +1,116 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+Solr allows rate limiting per request type. Each request type can be allocated a maximum allowed number of concurrent requests
+that can be active. The default rate limiting is implemented for updates and searches.
+
+If a request exceeds the request quota, further incoming requests are rejected with HTTP error code 429 (Too Many Requests).
+
+Note that rate limiting works at an instance (JVM) level, not at a core or collection level. Consider that when planning capacity.
```
madrob, line 24: Can we file a JIRA for finer grained request rate limiting?

solr/solr-ref-guide/src/rate-limiters.adoc
```diff
@@ -0,0 +1,116 @@
...
+
+== When To Use Rate Limiters
+Rate limiters should be used when the user wishes to allocate a guaranteed capacity of the request threadpool to a specific
+request type. Indexing and search requests are mostly competing with each other for CPU resources. This becomes especially
```
madrob, line 28: This talks about indexing and searching requests, but I only see configuration for query rate limiters below. If update requests are not considered here, we should look at that as follow-on work, and the docs should reflect the state of things.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrClient.java
```diff
@@ -387,7 +394,13 @@ protected HttpRequestBase createMethod(@SuppressWarnings({"rawtypes"})SolrReques
         throw new SolrException(SolrException.ErrorCode.BAD_REQUEST, "GET can't send streams!");
       }
 
-      return new HttpGet(basePath + path + wparams.toQueryString());
+      //return new HttpGet(basePath + path + wparams.toQueryString());
```
madrob, line 397: unneeded?

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrClient.java
```diff
@@ -387,7 +394,13 @@ protected HttpRequestBase createMethod(@SuppressWarnings({"rawtypes"})SolrReques
...
+
+      HttpGet result = new HttpGet(basePath + path + wparams.toQueryString());
+
+      result.addHeader(contextHeaders[0]);
```
madrob, line 401: let's extract this to a method, along with lines 445 and 366

solr/solrj/src/java/org/apache/solr/client/solrj/request/SolrPing.java
```diff
@@ -54,6 +54,11 @@ protected SolrPingResponse createResponse(SolrClient client) {
   public ModifiableSolrParams getParams() {
     return params;
   }
+
+  @Override
+  public String getRequestType() {
+    return SolrRequestType.QUERY.toString();
```
madrob, line 60: This is maybe more of an admin request? WDYT?

solr/solrj/src/java/org/apache/solr/client/solrj/request/QueryRequest.java
```diff
@@ -76,5 +76,9 @@ public SolrParams getParams() {
     return query;
   }
 
+  @Override
+  public String getRequestType() {
+    return SolrRequestType.QUERY.toString();
```
madrob, line 81: Something about this that I think bothers me is that we don't make a differentiation between distinct handlers, so we can't use this to build QoS levels. Like I can't say that RTG should always go through while `/select` might be acceptable to throttle and `/spellcheck` is even lower priority.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrClient.java
madrob: https://lucene.apache.org/solr/guide/8_5/authentication-and-authorization-plugins.html#securing-inter-node-requests has some ideas.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrClient.java
```diff
@@ -98,6 +99,8 @@
   private static final Charset FALLBACK_CHARSET = StandardCharsets.UTF_8;
   private static final String DEFAULT_PATH = "/select";
   private static final long serialVersionUID = -946812319974801896L;
+
+  protected static final Set<Integer> UNMATCHED_ACCEPTED_ERROR_CODES = new HashSet<>(Arrays.asList(429));
```
madrob, line 103: nit: `Collections.singleton`

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrClient.java
```diff
@@ -624,6 +641,12 @@ private HttpEntityEnclosingRequestBase fillContentStream(
       if (procCt != null) {
         String procMimeType = ContentType.parse(procCt).getMimeType().trim().toLowerCase(Locale.ROOT);
         if (!procMimeType.equals(mimeType)) {
+          if (isUnmatchedErrorCode(mimeType, httpStatus)) {
```
madrob, line 644: I don't understand when this condition occurs. Can you add some comments explaining it?

solr/core/src/java/org/apache/solr/servlet/RateLimitManager.java
```diff
@@ -0,0 +1,148 @@
...
+ *
+ * The actual rate limiting and the limits should be implemented in the corresponding RequestRateLimiter
+ * implementation. RateLimitManager is responsible for the orchestration but not the specifics of how the
+ * rate limiting is being done for a specific request type.
+ */
+public class RateLimitManager {
+  public final static int DEFAULT_CONCURRENT_REQUESTS= (Runtime.getRuntime().availableProcessors()) * 3;
+  public final static long DEFAULT_SLOT_ACQUISITION_TIMEOUT_MS = -1;
+  private final Map<String, RequestRateLimiter> requestRateLimiterMap;
+  private final Map<HttpServletRequest, RequestRateLimiter> activeRequestsMap;
```
madrob, line 44: Please add some comments here clarifying that callers must acquire the semaphore lease before adding requests to the map (and must delete from the map before returning the lease).

solr/core/src/java/org/apache/solr/servlet/RateLimitManager.java
```diff
@@ -0,0 +1,148 @@
...
+  public void decrementActiveRequests(HttpServletRequest request) {
+    RequestRateLimiter requestRateLimiter = activeRequestsMap.get(request);
+
+    if (requestRateLimiter == null) {
+      // No rate limiter for this request type
+      return;
+    }
+
+    requestRateLimiter.decrementConcurrentRequests();
+    activeRequestsMap.remove(request);
```
madrob, line 122: Potential deadlock - reorder this. Possibly consider using `computeIfPresent`.

solr/core/src/java/org/apache/solr/servlet/RateLimitManager.java
```diff
@@ -0,0 +1,148 @@
...
+      if (requestRateLimiter.getRateLimiterConfig().isSlotBorrowingEnabled && requestRateLimiter.allowSlotBorrowing()) {
+        return requestRateLimiter;
+      }
+    }
+
+    return null;
+  }
+
+  // Decrement the active requests in the rate limiter for the corresponding request type.
+  public void decrementActiveRequests(HttpServletRequest request) {
```
madrob, line 113: I feel like this is insufficient. If this request is part of its own queue, but there is another request of this same type that is currently borrowing a slot from another queue, then we shout prioritize returning the borrowed slot instead of our own.

This might have potential performance implications and can lead to livelock/request starvation.

solr/core/src/java/org/apache/solr/servlet/SolrDispatchFilter.java
```diff
@@ -184,6 +187,11 @@ public void init(FilterConfig config) throws ServletException
       coresInit = createCoreContainer(solrHomePath, extraProperties);
       this.httpClient = coresInit.getUpdateShardHandler().getDefaultHttpClient();
       setupJvmMetrics(coresInit);
+      RateLimitManager.Builder builder = new RateLimitManager.Builder();
+
+      builder.setConfig(config);
```
madrob, line 192: This is required, make it a constructor parameter on the builder?

solr/core/src/java/org/apache/solr/servlet/SolrDispatchFilter.java
```diff
@@ -377,6 +386,19 @@ public void doFilter(ServletRequest _request, ServletResponse _response, FilterC
         }
       }
 
+      try {
+        accepted = rateLimitManager.handleRequest(request);
+      } catch (InterruptedException e) {
+        throw new SolrException(ErrorCode.SERVER_ERROR, e.getMessage());
```
madrob, line 392: Thread.currentThread.interrupt

solr/core/src/test/org/apache/solr/servlet/TestRequestRateLimiter.java
```diff
@@ -0,0 +1,185 @@
...
+        try {
+          future.get();
+        } catch (Exception e) {
+          assertTrue("Not true " + e.getMessage(), e.getMessage().contains("non ok status: 429, message:Too Many Requests"));
+        }
+      }
+
+      MockRequestRateLimiter mockQueryRateLimiter = (MockRequestRateLimiter) rateLimitManager.getRequestRateLimiter(SolrRequest.SolrRequestType.QUERY);
+
+      assertTrue("Incoming request count did not match. Expected == 25  incoming " + mockQueryRateLimiter.incomingRequestCount.get(),
```
madrob, line 111: assertEquals

solr/core/src/test/org/apache/solr/servlet/TestRequestRateLimiter.java
```diff
@@ -0,0 +1,185 @@
...
+
+    client.commit();
+
+    ExecutorService executor = ExecutorUtil.newMDCAwareCachedThreadPool("threadpool");
+    List<Callable<Boolean>> callableList = new ArrayList<>();
+    List<Future<Boolean>> futures;
+
+    try {
+      for (int i = 0; i < 25; i++) {
+        callableList.add(new Callable<Boolean>() {
```
madrob, line 81: nit: use a lambda?

solr/core/src/test/org/apache/solr/servlet/TestRequestRateLimiter.java
```diff
@@ -0,0 +1,185 @@
...
+        });
+      }
+
+      futures = executor.invokeAll(callableList);
+
+      for (Future<?> future : futures) {
+        try {
+          future.get();
+        } catch (Exception e) {
+          assertTrue("Not true " + e.getMessage(), e.getMessage().contains("non ok status: 429, message:Too Many Requests"));
```
madrob, line 105: assertThat(e.getMessage(), contains(...))

solr/core/src/test/org/apache/solr/servlet/TestRequestRateLimiter.java
```diff
@@ -0,0 +1,185 @@
...
+            return true;
+          }
+        });
+      }
+
+      futures = executor.invokeAll(callableList);
+
+      for (Future<?> future : futures) {
+        try {
+          future.get();
```
madrob, line 103: Should we assert anything about the result of this?

solr/core/src/test/org/apache/solr/servlet/TestRequestRateLimiter.java
```diff
@@ -0,0 +1,185 @@
...
+
+    try {
+      for (int i = 0; i < 25; i++) {
+        callableList.add(new Callable<Boolean>() {
+          @Override
+          public Boolean call() throws Exception {
+            try {
+              QueryResponse response = client.query(new SolrQuery("*:*"));
+
+              if (response.getResults().getNumFound() > 0) {
```
madrob, line 87: should this ever be false?

solr/core/src/test/org/apache/solr/servlet/TestRequestRateLimiter.java
```diff
@@ -0,0 +1,185 @@
...
+import org.apache.solr.client.solrj.response.QueryResponse;
+import org.apache.solr.cloud.SolrCloudTestCase;
+import org.apache.solr.common.SolrInputDocument;
+import org.apache.solr.common.util.ExecutorUtil;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+import static org.apache.solr.servlet.RateLimitManager.DEFAULT_SLOT_ACQUISITION_TIMEOUT_MS;
+
+public class TestRequestRateLimiter extends SolrCloudTestCase {
```
madrob, line 40: I'd like to see some testing around the slot borrowing?

solr/core/src/java/org/apache/solr/servlet/RateLimitManager.java
```diff
@@ -0,0 +1,148 @@
...
+  /* For a rejected request type, do the following:
+   * For each request rate limiter whose type that is not of the type of the request which got rejected,
+   * check if slot borrowing is enabled. If enabled, try to acquire a slot.
+   * If allotted, return else try next request type.
+   */
+  private RequestRateLimiter trySlotBorrowing(String requestType) {
+    for (Map.Entry<String, RequestRateLimiter> currentEntry : requestRateLimiterMap.entrySet()) {
+      RequestRateLimiter requestRateLimiter = currentEntry.getValue();
+
+      if (requestRateLimiter.getRateLimiterConfig().requestType.equals(requestType)) {
```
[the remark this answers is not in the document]
atris: Good catch, thanks. Fixed.

solr/core/src/java/org/apache/solr/servlet/RateLimitManager.java
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    if (requestRateLimiter.handleRequest()) {
+      activeRequestsMap.put(request, requestRateLimiter);
+      return true;
+    }
+
+    requestRateLimiter = trySlotBorrowing(typeOfRequest);
+
+    if (requestRateLimiter != null) {
```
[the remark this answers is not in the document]
atris: No, since the slot would already have been acquired when the slot borrowing succeeded.

solr/core/src/test/org/apache/solr/servlet/TestRequestRateLimiter.java
atris: We assert on the response count in the actual query response in the executable.

solr/core/src/test/org/apache/solr/servlet/TestRequestRateLimiter.java
atris: In case of an exception (TOO_MANY_REQUESTS?)

solr/core/src/java/org/apache/solr/servlet/RequestRateLimiter.java
```diff
@@ -0,0 +1,131 @@
...
+
+/**
+ * Handles rate limiting for a specific request type.
+ *
+ * The control flow is as follows:
+ * Handle request -- Check if slot is available -- If available, acquire slot and proceed --
+ * else reject the same.
+ */
+public class RequestRateLimiter {
+  private Semaphore allowedConcurrentRequests;
```
anshumg, line 34: Might want to declare both of these as final ?

solr/core/src/test/org/apache/solr/servlet/TestRequestRateLimiter.java
```diff
@@ -0,0 +1,185 @@
...
+      assertTrue("Incoming total processed requests count did not match. Expected " + mockQueryRateLimiter.incomingRequestCount.get() + " incoming "
+              + (mockQueryRateLimiter.acceptedNewRequestCount.get() + mockQueryRateLimiter.rejectedRequestCount.get()),
+          (mockQueryRateLimiter.acceptedNewRequestCount.get() + mockQueryRateLimiter.rejectedRequestCount.get()) == mockQueryRateLimiter.incomingRequestCount.get());
+    } finally {
+      executor.shutdown();
+    }
+  }
+
+  private static class MockRequestRateLimiter extends RequestRateLimiter {
+    AtomicInteger incomingRequestCount;
```
anshumg, line 126: You could declare all of these as final

solr/core/src/test/org/apache/solr/servlet/TestRequestRateLimiter.java
```diff
@@ -0,0 +1,185 @@
...
+
+    ExecutorService executor = ExecutorUtil.newMDCAwareCachedThreadPool("threadpool");
+    List<Callable<Boolean>> callableList = new ArrayList<>();
+    List<Future<Boolean>> futures;
+
+    try {
+      for (int i = 0; i < 25; i++) {
+        callableList.add(new Callable<Boolean>() {
+          @Override
+          public Boolean call() throws Exception {
```
anshumg, line 83: this throws is redundant here considering we always throw a RuntimeException

solr/core/src/test/org/apache/solr/servlet/TestRequestRateLimiter.java
```diff
@@ -0,0 +1,185 @@
...
+        } catch (Exception e) {
+          assertTrue("Not true " + e.getMessage(), e.getMessage().contains("non ok status: 429, message:Too Many Requests"));
+        }
+      }
+
+      MockRequestRateLimiter mockQueryRateLimiter = (MockRequestRateLimiter) rateLimitManager.getRequestRateLimiter(SolrRequest.SolrRequestType.QUERY);
+
+      assertTrue("Incoming request count did not match. Expected == 25  incoming " + mockQueryRateLimiter.incomingRequestCount.get(),
+          mockQueryRateLimiter.incomingRequestCount.get() == 25);
+      assertTrue("Incoming accepted new request count did not match. Expected 5 incoming " + mockQueryRateLimiter.acceptedNewRequestCount.get(),
```
anshumg, line 113: Let's change these assertions to `assertEquals`? It's going to help debug or get more information from failure. `assertTrue` only prints true/false, but equals provides information like `expected X but got Y`

solr/solrj/src/java/org/apache/solr/client/solrj/request/QueryRequest.java
atris: This seems like a query prioritization model IMO -- a good idea. I will open a follow up JIRA once this goes in.

solr/core/src/test/org/apache/solr/servlet/TestRequestRateLimiter.java
atris: Converted to lambda

solr/core/src/java/org/apache/solr/servlet/RateLimitManager.java
atris: Discussed offline. Marking experimental and will follow up with a JIRA for rethinking the same.

solr/core/src/test/org/apache/solr/servlet/TestRequestRateLimiter.java
atris: Some of them cannot be converted since they are expressions. I have converted the remaining to assertEquals, thanks.

atris:
@anshumg @madrob Updated. Please see and let me know if this looks ready to go.

solr/solr-ref-guide/src/index.adoc
```diff
@@ -124,6 +125,8 @@ The *<<getting-started.adoc#getting-started,Getting Started>>* section guides yo
 *<<legacy-scaling-and-distribution.adoc#legacy-scaling-and-distribution,Legacy Scaling and Distribution>>*: This section tells you how to grow a Solr distribution by dividing a large index into sections called shards, which are then distributed across multiple servers, or by replicating a single index across multiple services.
 
 *<<circuit-breakers.adoc#circuit-breakers,Circuit Breakers>>*: This section talks about circuit breakers, a way of allowing a higher stability of Solr nodes and increased service level guarantees of requests that are accepted by Solr.
+
+*<<rate-limiters.adoc#rate-limiters,Request Rate Limiters>>*: This section talks about request rate limiters, a way of guaranteeing throughput per request type and dedicating resource quotas by resource type.
```
anshumg, line 129: can we specify that this is per JVM/Solr instance?

atris: I had added that in rate-limiters.adoc, added here as well.

anshumg:
LGTM. If there's nothing else pending w.r.t. other reviewers, feel free to merge this in.

anshumg:
The precommit checks seem to be failing though. please make sure the precommit and the tests pass as expected before merging.

chatman:
Can you please wait until tomorrow (say 36 hours more) for me to take a look? Extremely busy with some client work, but I would love to take a look, if you don't mind. If I don't show up here in 48 hours from now, please feel free to ignore me.

atris:
Thanks @anshumg !

@chatman Sure. I will leave the PR open until Friday morning IST post which, barring objections, I plan to merge.

solr/core/src/java/org/apache/solr/servlet/RateLimitManager.java
```diff
@@ -0,0 +1,167 @@
...
+    this.requestRateLimiterMap = new HashMap<>();
+    this.activeRequestsMap = new ConcurrentHashMap<>();
+  }
+
+  // Handles an incoming request. The main orchestration code path, this method will
+  // identify which (if any) rate limiter can handle this request. Internal requests will not be
+  // rate limited
+  // Returns true if request is accepted for processing, false if it should be rejected
+  public boolean handleRequest(HttpServletRequest request) throws InterruptedException {
+    String requestContext = request.getHeader(SOLR_REQUEST_CONTEXT_PARAM);
```
madrob, line 64: Did we give up on using the authentication plugin info to determine if this is an internal request or not?

solr/core/src/test/org/apache/solr/servlet/TestRequestRateLimiter.java
```diff
@@ -102,31 +103,101 @@ public Boolean call() throws Exception {
...
+
+          return true;
+        });
+      }
+
+      futures = executor.invokeAll(callableList);
+
+      for (Future<?> future : futures) {
+        try {
+          future.get();
```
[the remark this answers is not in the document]
atris: assertTrue(future.get() != null); instead?

solr/core/src/java/org/apache/solr/servlet/RateLimitManager.java
```diff
@@ -38,9 +41,14 @@
...
 public class RateLimitManager {
+  private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+
   public final static int DEFAULT_CONCURRENT_REQUESTS= (Runtime.getRuntime().availableProcessors()) * 3;
   public final static long DEFAULT_SLOT_ACQUISITION_TIMEOUT_MS = -1;
   private final Map<String, RequestRateLimiter> requestRateLimiterMap;
+
+  // IMPORTANT: The slot from the corresponding rate limiter should be acquired before adding the request
+  // to this map. Subsequently, the request should be deleted from the map before the slot is released.
   private final Map<HttpServletRequest, RequestRateLimiter> activeRequestsMap;
```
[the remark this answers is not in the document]
atris: This is already a ConcurrentHashMap. The comment is redundant, removing.

solr/core/src/java/org/apache/solr/servlet/RequestRateLimiter.java
```diff
@@ -53,6 +53,8 @@ public boolean handleRequest() throws InterruptedException {
    * Whether to allow another request type to borrow a slot from this request rate limiter. Typically works fine
    * if there is a relatively lesser load on this request rate limiter's type compared to the others (think of skew).
    * @return true if allow, false otherwise
+   *
+   * @lucene.experimental -- Can cause slots to be blocked if a request borrows a slot and is itself long lived.
    */
   public boolean allowSlotBorrowing() {
     synchronized (this) {
```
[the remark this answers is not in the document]
atris: Good point. I was not sure around using availablePermits for this use but decided for using it. IMO the worst that can happen is that a request gets an indication from availablePermits as to the availability of a slot but fails to acquire it since another concurrent request has acquired the same. The approach that you suggest is more comprehensive, so changed to that, thanks.

solr/core/src/java/org/apache/solr/servlet/RateLimitManager.java
atris: https://issues.apache.org/jira/browse/SOLR-14720

solr/core/src/test/org/apache/solr/servlet/TestRequestRateLimiter.java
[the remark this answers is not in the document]
madrob: I’m just going by memory here, but I thought future.get gives you the return value of the callable which should always be true in this case for success?

solr/core/src/java/org/apache/solr/servlet/RequestRateLimiter.java
```diff
@@ -58,18 +58,18 @@ public RequestRateLimiter(RateLimiterConfig rateLimiterConfig) {
...
     }
 
     if (borrowableSlotsPool.tryAcquire(rateLimiterConfig.waitForSlotAcquisition, TimeUnit.MILLISECONDS)) {
-      return new Pair<Boolean, AcquiredSlotMetadata>(true, new AcquiredSlotMetadata(this, true));
+      return new Pair<Boolean, AcquiredSlotMetadata>(true, new AcquiredSlotMetadata(this, borrowableSlotsPool));
     }
 
-    return new Pair<Boolean, AcquiredSlotMetadata>(false, null);
+    return new Pair<Boolean, AcquiredSlotMetadata>(false, new AcquiredSlotMetadata(null, null));
   }
```
madrob: I would write this as something like:

```
class SemaphoreWrapper() {
  Semaphore wrapped;

  release() {
    if (wrapped != null) wrapped.release();
  }
}

  public SempahoreWrapper handleRequest() throws InterruptedException {
      if (!rateLimiterConfig.isEnabled) {
       return nopPool; // = new SemaphoreWrapper(null);
     }

    if (guaranteedSlotsPool.tryAcquire(rateLimiterConfig.waitForSlotAcquisition, TimeUnit.MILLISECONDS)) {
       return new SemaphoreWrapper(guaranteedSlotsPool);
     }

    if (borrowableSlotsPool.tryAcquire(rateLimiterConfig.waitForSlotAcquisition, TimeUnit.MILLISECONDS)) {
       return new SemaphoreWrapper(borrowableSlotsPool);
     }
 
      return null;
   }
```

And probably have the limiter own the wrappers and return the right thing each time instead of creating a new wrapper.

atris: We need the rate limiter to be a part of the returned value here so that the decrement can be invoked on the right RequestRateLimiter?

solr/core/src/java/org/apache/solr/servlet/RequestRateLimiter.java
```diff
@@ -55,21 +54,21 @@ public RequestRateLimiter(RateLimiterConfig rateLimiterConfig) {
    * NOTE: Always check for a null metadata object even if this method returns a true -- this will be the scenario when
    * rate limiters are not enabled.
    * */
-  public Pair<Boolean, AcquiredSlotMetadata> handleRequest() throws InterruptedException {
+  public SlotMetadata handleRequest() throws InterruptedException {
 
     if (!rateLimiterConfig.isEnabled) {
-      return new Pair<Boolean, AcquiredSlotMetadata>(true, new AcquiredSlotMetadata(null, null));
+      return new SlotMetadata(null);
```
madrob, line 60: I'd prefer to see these as pre declared members so that we're not allocating a new metadata on each request.

atris:
@madrob Updated, please see and let me know your thoughts.

solr/core/src/java/org/apache/solr/servlet/RequestRateLimiter.java
```diff
@@ -152,14 +143,22 @@ public RateLimiterConfig(SolrRequest.SolrRequestType requestType, boolean isEnab
...
       this.usedPool = usedPool;
     }
+
+    public void decrementRequest() {
+      if (usedPool != null) {
+        usedPool.release();
+      }
+    }
+
+    public boolean isUsedPoolNull() {
```
madrob, line 160: This seems like an implementation detail that we don't need to expose?

solr/core/src/java/org/apache/solr/servlet/RateLimitManager.java
```diff
@@ -147,19 +145,16 @@ public boolean handleRequest(HttpServletRequest request) throws InterruptedExcep
 
   // Decrement the active requests in the rate limiter for the corresponding request type.
   public void decrementActiveRequests(HttpServletRequest request) {
-    RequestRateLimiter.AcquiredSlotMetadata acquiredSlotMetadata = activeRequestsMap.get(request);
+    RequestRateLimiter.SlotMetadata slotMetadata = activeRequestsMap.get(request);
 
-    if (acquiredSlotMetadata == null) {
+    if (slotMetadata == null) {
```
madrob, line 150: I think I would flip this condition so that we only have one return point from the method.

solr/core/src/java/org/apache/solr/servlet/RateLimitManager.java
```diff
@@ -132,12 +130,12 @@ public boolean handleRequest(HttpServletRequest request) throws InterruptedExcep
           Thread.currentThread().interrupt();
         }
 
-        if (result != null && result.first()) {
-          if (result.second() == null) {
-            throw new IllegalStateException("AcquiredSlotMetadata object null even when slot is acquired and rate limiters are enabled");
-          }
+        if (result == null) {
```
madrob, line 133: This should never happen, right?

solr/core/src/java/org/apache/solr/servlet/RequestRateLimiter.java
```diff
@@ -80,20 +79,12 @@ public RequestRateLimiter(RateLimiterConfig rateLimiterConfig) {
    *
    * @lucene.experimental -- Can cause slots to be blocked if a request borrows a slot and is itself long lived.
    */
-  public Pair<Boolean, AcquiredSlotMetadata> allowSlotBorrowing() throws InterruptedException {
+  public SlotMetadata allowSlotBorrowing() throws InterruptedException {
```
madrob, line 82: javadoc on this is out of date, still refers to booleans

solr/core/src/java/org/apache/solr/servlet/RateLimitManager.java
```diff
@@ -0,0 +1,183 @@
...
+
+  public RateLimitManager() {
+    this.requestRateLimiterMap = new HashMap<>();
+    this.activeRequestsMap = new ConcurrentHashMap<>();
+  }
+
+  // Handles an incoming request. The main orchestration code path, this method will
+  // identify which (if any) rate limiter can handle this request. Internal requests will not be
+  // rate limited
+  // Returns true if request is accepted for processing, false if it should be rejected
```
madrob, line 60: nit: use proper javadoc

solr/core/src/java/org/apache/solr/servlet/RequestRateLimiter.java
```diff
@@ -152,14 +143,22 @@ public RateLimiterConfig(SolrRequest.SolrRequestType requestType, boolean isEnab
     }
   }
 
-  // Represents the metadata for an acquired slot
-  static class AcquiredSlotMetadata {
-    public RequestRateLimiter requestRateLimiter;
-    public Semaphore usedPool;
+  // Represents the metadata for a slot
+  static class SlotMetadata {
```
madrob, line 147: 👍

solr/core/src/java/org/apache/solr/servlet/RequestRateLimiter.java
atris: We need this to avoid caching a request when request rate limiter is disabled. Renaming the method.

solr/core/src/java/org/apache/solr/servlet/RateLimitManager.java
atris: Yes, thats why the check.

solr/core/src/java/org/apache/solr/servlet/RequestRateLimiter.java
```diff
@@ -157,7 +162,7 @@ public void decrementRequest() {
       }
     }
 
-    public boolean isUsedPoolNull() {
+    public boolean isReleasable() {
       return usedPool == null;
```
madrob, line 166: Should be != null, right? Check the condition wherever this is called too, might need to invert those.

solr/core/src/java/org/apache/solr/servlet/RequestRateLimiter.java
```diff
@@ -40,32 +42,35 @@
   private final Semaphore borrowableSlotsPool;
 
   private final RateLimiterConfig rateLimiterConfig;
+  private final SlotMetadata guaranteedSlotMetadata;
+  private final SlotMetadata borrowedSlotMetadata;
+  private final SlotMetadata nullSlotMetadata;
```
madrob, line 47: can this be static? pretty minor but since it doesn't do anything, it can be shared by all the instances and save some memory

solr/core/src/java/org/apache/solr/servlet/RequestRateLimiter.java
atris: Fixed

atris:
Thank you @madrob and @anshumg !

PR #1689, opened 2020-07-23
LUCENE-9439: Matches API should enumerate hit fields that have no positions (support empty iterator)

dweiss:
A minimally-intrusive patch that seems to work just fine.


lucene/join/src/java/org/apache/lucene/search/join/ToParentBlockJoinQuery.java
```diff
@@ -195,7 +195,9 @@ public Matches matches(LeafReaderContext context, int doc) throws IOException {
           return null;
         }
       }
-      return MatchesUtils.MATCH_WITH_NO_TERMS;
+
+      // TODO: which fields should be here?
```
dweiss, line 199: This bit I wasn't really sure about.

PR #1694, opened 2020-07-26
SOLR-14680: Provide simple interfaces to our cloud classes  (only API)

noblepaul:
A few notes before anyone who starts reviewing this

-  This was created after I saw similar attempt as a part #1684 . I believe this has to receive a more wider input and review irrespective of whether devs are interested in autoscaling or not
- This is a WIP PR
- The concrete implementations are for demo purposes. Can be omitted, if required. Anything outside the `o.a.s.cluster.api` package is optional and will be removed 
- The interfaces are designed to be minimal to avoid overload. We can and will add more methods later. Let's not add a lot


murblanc:
Can we start simple by defining the external interface without the required changes to internal classes first? Will make a smaller PR easier to discuss.

murblanc:
I also feel this PR has a lot of overlap with the one I started to define an API for Autoscaling plugins (https://github.com/apache/lucene-solr/pull/1684).

Unless we want to pursue two distinct options for these interfaces and decide later (perfectly OK if that’s the idea) I suggest we use the other PR that has also examples of how these interfaces are used (makes it easier to reason about them and decide what goes where).

gus-asf:
These interfaces look like a good way to begin. Thoughts: This is an excellent point at which to consider and then stick to a preferred set of terminology. My favorites are:

- **Machines** are the operating system level container or hardware and have
  - **Nodes** which are running solr processes on machines which have
    - **Collections** which are logical indexes associated with a particular schema and have
      - **Shards** which split a collection on a routing value to allow scaling and have
        - **Replicas** which are duplicate copies implementing a shard often hosted on separate nodes.

We should strive to apply only ONE word for each level throughout the interfaces to make them as simple and easy to understand as possible. 

These 5 terms all have nice clear english words that communicate their function to some extent. Core however only really means "central thing" and I have always thought it was a very confusing word to use here since it has almost no memnonic value. Slice is also good for the shard level, but I think shard is no worse and more commonly used in documentation. Although core has been used many times in the past, I don't think anyone is going to have trouble finding what they want via the terms I'm suggesting.

solr/solrj/src/java/org/apache/solr/common/util/LinkedSimpleHashMap.java
```diff
@@ -0,0 +1,27 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.solr.common.util;
+
+import java.util.LinkedHashMap;
+
+public class LinkedSimpleHashMap<T> extends LinkedHashMap<CharSequence, T>  implements SimpleMap<T> {
```
gus-asf, line 21: this class appears unused...

solr/solrj/src/java/org/apache/solr/common/cloud/ClusterState.java
```diff
@@ -38,7 +43,7 @@
  * {@link ZkStateReader#getClusterState()}.
  * @lucene.experimental
  */
-public class ClusterState implements JSONWriter.Writable {
+public class ClusterState implements JSONWriter.Writable , SolrCluster {
```
gus-asf, line 46: Little worried that this design allows for casting of the SolrCluster reference...

murblanc:
Snitches have a target (at least node or replica, but one could imagine Cluster, collection or shard if such snitches do not already exist).
Would be helpful for all these classes to implement a common interface so they can be passed when a notch target is needed in Autoscaling plugins.

chatman:
As I suggested in the other PR on autoscaling, I feel we should change the package name:
https://github.com/apache/lucene-solr/pull/1684#issuecomment-664051353

Also, WDYT about marking the concrete classes DocCollection, Replica etc. with @lucene.internal to make it clear that they shouldn't be used directly from the plugins? Alternatively, we can mark these new interfaces in this PR with a new annotation (e.g. @solr.external) that signals to plugin writers that these will be stable for their use across versions.

gus-asf:
While marking things internal and external is good, a stronger possibility to code more defensively and make "new plugins" get instantiated/injected with implementations of a facade that return instances of the interfaces. If those implementations can't just be cast to internal types we then have a much higher barrier to abuse (reflection). 

As for package names I agree we need to think carefully about them. It is a question what the goal is too. If we're reworking the internals, sdk won't sound right at all. If we are building something only meant to be consumed by an outer layer of plugins to insulate the plugins from internal refactoring, something in that vein sounds fine (or maybe o.a.s.cloud.plugin.api?) Settling on what our goals are will help us decide what name is best.

solr/solrj/src/java/org/apache/solr/common/util/LinkedSimpleHashMap.java
noblepaul: yes, this was to demonstrate how to implement a default impl of `SimpleMap`

solr/solrj/src/java/org/apache/solr/common/cloud/ClusterState.java
noblepaul: We can have multiple implementations of `SolrCluster` .The idea of existing classes implementing these interfaces is to have readily available implementations

noblepaul:
> Can we start simple by defining the external interface without the required changes to internal classes first? Will make a smaller PR easier to discuss.

Feel free to ignore the implementing classes. That was to demonstrate one way of implementing these interfaces. I'm happy to focus on the interfaces inside the `sdk` package

noblepaul:
>I also feel this PR has a lot of overlap with the one I started to define an API for Autoscaling plugins #1684 

Yes, totally.That ticket is trying to do something more than the scope of that ticket. I want to limit the scope of that ticket and have wider discussion on how a simple set of interfaces to represent the existing Solr cluster/cloud

solr/solrj/src/java/org/apache/solr/common/cloud/ClusterState.java
chatman: Let's rename Solr luster to Cluster.

solr/solrj/src/java/org/apache/solr/common/cloud/sdk/SolrNode.java
```diff
@@ -0,0 +1,33 @@
...
+ * A read only view of a Solr node
+ */
+public interface SolrNode {
+
+  /** The node name */
+  String name();
+
+  String baseUrl(boolean isV2);
+
+  SimpleMap<ShardReplica> cores();
```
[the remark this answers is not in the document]
noblepaul: I would say, that is something that can be easily computed from the `cores()` method and can be provided as a helper method elsewhere

noblepaul:
>These interfaces look like a good way to begin. Thoughts: This is an excellent point at which to consider and then stick to a preferred set of terminology. My favorites are:

There is no way to have  a "machine" abstraction in Solr. We can start of a `Node` as an atomic unit. I wanted to carefully avoid the class names already taken up so that existing code can be left as it is

solr/solrj/src/java/org/apache/solr/common/cloud/sdk/SolrCluster.java
```diff
@@ -0,0 +1,46 @@
...
+ */
+
+package org.apache.solr.common.cloud.sdk;
+
+import org.apache.solr.common.util.SimpleMap;
+
+/**
+ * Represents a Solr cluster
+ */
+public interface SolrCluster {
```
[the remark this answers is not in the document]
noblepaul: I know. Imagine this will also be used outside of Solr as a part of SolrJ. So if you read some client code, 

`Cluster cluster;`

is less readable compared to

`SolrCluster solrCluster`;

Eventually, I would wish to replace a lot of SolrJ code/API with a standard set of interfaces.

solr/solrj/src/java/org/apache/solr/common/cloud/sdk/SolrCluster.java
```diff
@@ -0,0 +1,46 @@
...
+  /** collections in the cluster */
+  SimpleMap<SolrCollection> collections();
```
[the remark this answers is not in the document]
noblepaul: Yes, I wanted to include aliases. Where do you propose it to be there. 

I would say , There should be no `Alias` class/interface. 

It should be as simple as 
`SolrCluster#collections#get(String name , include alias) `

and 

` SolrCollection#aliases() returns List<String>`

solr/solrj/src/java/org/apache/solr/common/cloud/sdk/SolrCollection.java
```diff
@@ -0,0 +1,33 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.solr.common.cloud.sdk;
+
+import org.apache.solr.common.util.SimpleMap;
+
+/** Represents a collection in Solr */
+public interface SolrCollection {
```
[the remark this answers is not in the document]
noblepaul: That should be a utility method outside

gus-asf:
> > These interfaces look like a good way to begin. Thoughts: This is an excellent point at which to consider and then stick to a preferred set of terminology. My favorites are:
> 
> There is no way to have a "machine" abstraction in Solr. We can start of a `Node` as an atomic unit. I wanted to carefully avoid the class names already taken up so that existing code can be left as it is

Yeah machine is a difficult notion as it's a real world thing entirely outside java. It probably has to be optional and provided by sysprop or env var. and even so might be reducible to a property on node. Certainly detecting it would be fraught with peril. But machine designations (and rack designations) might be important to some folks.

murblanc:
Many methods returning other objects are returning names of other abstractions being defined here (Shard, Collection, Node etc.).
I’d think an instance of the abstraction should be returned instead, and that instance would have a getName() method (i.e have SolrCollection getCollection() rather than String getCollection() on a Shard for example).
What’s the rationale for returning names rather than objects?

murblanc:
I see two cases where we need interfaces such as defined here:
- internal code. Coding to interfaces rather than the actual implementation makes for better structured code usually with less implementation leaks,
- external “plugins”.

What’s the intention? If only the later, then there’s really no need to have the internal classes implement these interfaces, a wrapper is ok.
I believe addressing both points with a single interface is complicated (and to a point counterproductive as it ties internal and external views).

noblepaul:
>I’d think an instance of the abstraction should be returned instead,

I though about it and decided against it. We should always try to keep the methods light and avoid creating unnecessary Objects.
for instance, the `Shard#leader()` returns the name of the leader replica. A String is lightweight object and we do not need to create a `ShardReplica` object unnecessarily.

> I see two cases where we need interfaces such as defined here:

Both. over time , we will move as much of our code to use these interfaces instead of concrete classes.

I do not see a reason why external plugins cannot use these interfaces. This is not to say that external plugins may need some more methods. In that case, just compose your own interfaces using these as building blocks.

murblanc:
Thanks Noble.
So what’s your guidance regarding the Autoscaling PR? Shall we assume the Autoscaling plugins need to use these interfaces or shall I continue building cluster/node/collection/shard/replica abstractions that make plugin development easier?

If the first option, there will be a few requirements on these interfaces (and their implementations), for example allowing them to be keys in maps (variable values or snitches are per node or collection or replica etc, they need to be stored somewhere so the plugin can access them). Ideally also as stated elsewhere, all these interfaces implement a common one (could be empty) so that snitch/variable target can be passed easily.

The Autoscaling plugin work can also define similar abstractions on top of the interfaces defined here.
I do see value for all plugins of all types to share a single set of abstractions but obviously this puts more constraints on these abstractions...

murblanc:
We also need to guarantee some immutability of the instances passed to plugins. Computing placement if the underlying data (cluster state) can change during the computation would be tricky. So these abstractions are read only so plugins can’t use them to change Solr state but the instances passed to plugins are immutable. That would make it possibly a bit harder to only have the internal classes implement these interfaces.

noblepaul:
> So what’s your guidance regarding the Autoscaling PR? Shall we assume the Autoscaling plugins need to use these interfaces or shall I continue building cluster/node/collection/shard/replica abstractions that make plugin development easier?

I'm sure a lot of your autoscaling needs can be directly met by some of these interfaces. I'm also sure there are some needs not met by these. You can compose other interfaces using these as building blocks. 

In essence I wish we don't create multiple interfaces to represent the same concepts in our cluster.

>  Computing placement if the underlying data (cluster state) can change during the computation would be tricky. 

The assign implementation should read the cluster state before the computation starts. However there is no way we can guarantee immutability of everything between the time the plugin is invoked and the data is fetched. that's a lot of data. And we do not even know what kind of plugin are we writing and what are the metrics it is going to use.

gus-asf:
> We also need to guarantee some immutability of the instances passed to plugins. Computing placement if the underlying data (cluster state) can change during the computation would be tricky. So these abstractions are read only so plugins can’t use them to change Solr state but the instances passed to plugins are immutable. That would make it possibly a bit harder to only have the internal classes implement these interfaces.

Currently, the overseer work queue ensures orderly processing and locking during command execution, but there is some dissatisfaction with that, and in-order processing is not quite the same thing as it will still accept commands that are stale or that will become stale as the queue is processed. **IF** we want to attempt to prevent operations on stale data, perhaps something along the lines of optimistic locking. HOW to achieve such a thing would be an interesting question. Would need some notion of a "version" for each object that might be mutated and every mutation would increment it, some admin commands would then need to check the version if a version is supplied with the command. Such a thing would need use versions that live in zookeeper to ensure all nodes can see them. We would want to be careful about updating the version only when needed to avoid spamming watchers with trivial changes (since they watch entire blobs of json), so there is a possible cost to this. This part of the discussion may be straying afield from the current PR however, and maybe should be carried back to the list.

Also, before we build such a thing, what specific scenarios are you thinking of as problematic?

noblepaul:
@murblanc 

I understand your desire to have all data prefetched in the beginning. However, it is totally possible to implement that with the current set of interfaces. 

We can have an implementation of `SolrCluster` & it's associated interfaces to fetch this info

- just in time (say `LazilyFetchedSolrCluster` ?). This is what Solr would use in most places. The APIs are designed to fetch information as lazily as possible. 
- or prefetched the moment we construct the class (say `EagerlyFetchedSolrCluster` ?). This class can take in flags that tell it to what level do we need to prefetch information . "state.json" , "all-leaders", "core details-of all nodes" etc etc

noblepaul:
@gus-asf 

We may have  a concept of znode versions of each objects assuming that it is coming from some file in ZK. However it's totally possible to construct a `SolrCluster` object from a persisted state in a file/database. So, we may want to tread in that direction with caution. We may even start constructing arbitrary `SolrCluster` objects from in-line JSON etc etc for JUnit tests.

murblanc:
I don’t really mind when data is fetched, but I think that once cluster state is returned to a plugin code and that plugin accessed the data, the view should be consistent. In other words, if a plugin calls getNodes() twice and the number of nodes changes between the calls, plugins have to be a lot more cautious (or make copies...). I’d rather serve stale but consistent data (staleness is something that has to be dealt with anyway) than have data changing while computations are taking place.

chatman:
bq. I don’t really mind when data is fetched, but I think that once cluster state is returned to a plugin code and that plugin accessed the data, the view should be consistent. In other words, if a plugin calls getNodes() twice and the number of nodes changes between the calls, plugins have to be a lot more cautious (or make copies...). I’d rather serve stale but consistent data (staleness is something that has to be dealt with anyway) than have data changing while computations are taking place.

To ensure that, I think Noble's `EagerlyFetchedSolrCluster` will be useful. Whoever gives the plugin a Cluster object will pre-fetch all the leader info, nodes info, possibly even metrics info etc. and give out an immutable object (deeply nested).

murblanc:
EagerlyFetchedSolrCluster would then guarantee immutability? Must also be careful not too fetch too much. Autoscaling for example might not always need all details on everything.

chatman:
> EagerlyFetchedSolrCluster would then guarantee immutability?

Yes, that's the idea.

> Must also be careful not too fetch too much. Autoscaling for example might not always need all details on everything.

That's true. I think that's why Noble proposes: "This class can take in flags that tell it to what level do we need to prefetch information." Autoscaling plugin can specify the exact level of details it needs.

chatman:
+1 to `o.a.s.cluster.api` package name. +1 to getting rid of "cloud" here.

solr/solrj/src/java/org/apache/solr/cluster/api/SolrNode.java
```diff
@@ -0,0 +1,39 @@
...
+public interface SolrNode {
+
+  /** The node name */
+  String name();
+
+  /**Base http url for this node
+   *
+   * @param isV2 if true gives the /api endpoint , else /solr endpoint
+   */
+  String baseUrl(boolean isV2);
```
janhoy, line 32: Instead of boolean isV2, what about an Enum, that could then evolve?

solr/solrj/src/java/org/apache/solr/cluster/api/Config.java
```diff
@@ -0,0 +1,40 @@
...
+import java.util.Set;
+import java.util.function.Consumer;
+
+import org.apache.solr.common.SolrException;
+
+public interface Config {
+
+  String name();
+
+  /**set of files in the config */
```
janhoy, line 30: General : Please format Javadocs properly with text starting on line 2 and capital letter, even if it takes more lines.

solr/solrj/src/java/org/apache/solr/cluster/api/Config.java
```diff
@@ -0,0 +1,40 @@
...
+
+package org.apache.solr.cluster.api;
+
+import java.io.InputStream;
+import java.util.Set;
+import java.util.function.Consumer;
+
+import org.apache.solr.common.SolrException;
+
+public interface Config {
```
janhoy, line 26: Use `ConfigSet` instead? `Config` is soooo overloaded.

solr/solrj/src/java/org/apache/solr/cluster/api/Config.java
dsmiley: A one-line variant is acceptable: https://google.github.io/styleguide/javaguide.html#s7.1.1-javadoc-multi-line
The only problem I see with the javadoc here is that it's missing a leading space, and it should capitalize the first letter.

janhoy: Ok, agree

solr/solrj/src/java/org/apache/solr/cluster/api/Config.java
noblepaul: I don't like both names , `Config` & `ConfigSet`

`configset` suggests it is a set of configs? in reality it's a single configuration which has multiple files

murblanc:
Machine vs Node: environment variables (rack, availability zone etc) are "machine" abstractions but can be considered at the node level. Free disk is clearly a machine abstraction (two nodes on same machine might be sharing the same free disk and skewing decisions based on it) but I don't see how this can be solved (even a single node might be sharing space with other processes).
Therefore I don't see how a machine abstraction helps.

murblanc:
Would love to see sample plug-in code accessing cluster config. How the instances are obtained and how the cluster is explored (nodes, collections, shards, replicas).

noblepaul:
@murblanc I shall try to provide some sample implementation in another 2-3 days

noblepaul:
Opening another PR

solr/solrj/src/java/org/apache/solr/cluster/api/SolrNode.java
noblepaul: done

noblepaul:
I intend to merge this soon

PR #1695, opened 2020-07-26
LUCENE-9321: Fix Javadoc offline link base url for snapshot build

mocobeta:
JIRA: https://issues.apache.org/jira/browse/LUCENE-9321

This includes two changes in documentation task.

1. Fix Javadoc offline link base url to make snapshot build pass `checkBrokenLinks`.
2. Explicitly disable cross-project link for `:solr:test-framework`, otherwise broken links are generated due to split package issue.
See: https://github.com/apache/lucene-solr/blob/master/solr/test-framework/build.xml#L59

By applying those changes, `./gradlew :solr:checkBrokenLinks` succeeds on snapshot build.

### DIFF
--- a/gradle/documentation/documentation.gradle
+++ b/gradle/documentation/documentation.gradle
@@ -22,16 +22,20 @@ configure(rootProject) {
   ext {
     luceneDocUrl = project.propertyOrDefault('lucene.javadoc.url', {
       if (project.version != project.baseVersion) {
+        // non-release build
         project('lucene').file('build/documentation').toURI().toASCIIString()
       } else {
+        // release build
         "https://lucene.apache.org/core/${urlVersion}"
       }
     }())
 
     solrDocUrl = project.propertyOrDefault('solr.javadoc.url', {
       if (project.version != project.baseVersion) {
+        // non-release build
         project('solr').file('build/documentation').toURI().toASCIIString()
       } else {
+        // release build
         "https://lucene.apache.org/solr/${urlVersion}"
       }
     }())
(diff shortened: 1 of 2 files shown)


dweiss:
Look all right to me.

mocobeta:
@uschindler would you check the changes?

uschindler:
Hi looks fine. I recognized the strange this with the Solr test framework already (it's also part of the Ant build). I don't fully understand the issue, but as a workaround it's ok for now.

Otherwise looks good!

mocobeta:
Thanks for confirming, I merged it.

p.s.

> I don't fully understand the issue, but as a workaround it's ok for now.

Here is a bit more detailed explanation for the second change I omitted for laziness...
Without the change, several invalid links are generated (in solr-test-framework javadocs). e.g.:
```
BROKEN LINK: file:///mnt/hdd/repo/lucene-solr/lucene/build/documentation/core/org/apache/lucene/util/LuceneTestCase.html
BROKEN LINK: file:///mnt/hdd/repo/lucene-solr/solr/build/documentation/solrj/org/apache/solr/client/solrj/embedded/JettySolrRunner.html
```

They are invalid because `LuceneTestCase` actually belongs to `lucene-test-framework` module, and `JettySolrRunner` belongs to `solr-core` module. Seems like javadoc tool sees only package names in `element-list` and don't check if the classes really exist in the package when making links to referred classes (maybe it's an expected behaviour since the tool cannot retrieve the actual HTML files when `--linkoffline` option is used).
(Of course if we have no shared package names between modules, this problem should not occur.)

PR #1697, opened 2020-07-27
LUCENE-9292: BKDWriter refactor: Group point configuration in its own class

iverase:
see https://issues.apache.org/jira/browse/LUCENE-9292

This changes introduces a new class called BKDConfig which contains the point configuration. It just makes things a bit more tidy and it encapsulates all the bytes calculations for number of dimension and bytes per dimension.

Relates #1376


lucene/core/src/java/org/apache/lucene/util/bkd/BKDConfig.java
```diff
@@ -0,0 +1,90 @@
...
+ */
+
+package org.apache.lucene.util.bkd;
+
+import org.apache.lucene.util.ArrayUtil;
+
+/**
+ * Basic parameters for indexing points on the BKD tree.
+ */
+public class BKDConfig {
```
jpountz, line 25: final?

jpountz (approved):

lucene/core/src/java/org/apache/lucene/util/bkd/BKDConfig.java
iverase: of course, thanks!

PR #1700, opened 2020-07-27
Found that asciidoctor 1.5.6.2 cause build-site to fail on long anchors

gus-asf:
Documenting need to have ascii doctor version up to date.

### DIFF
--- a/solr/solr-ref-guide/README.adoc
+++ b/solr/solr-ref-guide/README.adoc
@@ -21,13 +21,11 @@ This is the source for the Solr Reference Guide.
 Raw content is stored in Asciidoc (`.adoc`) formatted files in the `src/` directory.
 
 == Prerequisites for Building
-These files are processed with AsciiDoctor in 2 different ways:
+These files are processed with AsciiDoctor.
 
-* HTML version, using Jekyll:
+* HTML is generated using Jekyll with these key dependencies:
 ** `Ruby` (v2.3 or higher)
 ** The following gems must be installed:
-*** `asciidoctor`:v2.0.10 or higher.
-Use `gem install asciidoctor` to install.
 *** `jekyll`: v3.5, not v4.x.
 Use `gem install jekyll --force --version 3.5.0` to force install of v3.5.0.
 *** `jekyll-asciidoc`: v2.1 or higher; latest version (3.0.0) is fine.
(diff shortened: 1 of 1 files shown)


ctargett:
This is only a problem if you already had it installed before it was upgraded several months ago. At that time, I sent mail to the Dev list explaining the need for people to specifically upgrade, and it’s explained in the Jira issue I used at that time. I’m on vacation, so I’ll leave searching for those things up to you. For people who never installed any of the dependencies, I believe the right version of the gem gets installed with jekyll-asciidoc gem, it just doesn’t upgrade it if you already have it, which is why I’d removed it from the README. And as soon as we move to Gradle, the need for any dependencies goes away entirely.

So what you choose to do with this PR with all that info is up to you, but I’d prefer it be clear and accurate about who needs the extra step. As I mentioned, I’m on vacation, so won’t be reviewing this again until next week at the earliest.

dweiss:
Please use gradle to compile the refguide. It'll keep dependencies up to date and consistent.

gus-asf:
> Please use gradle to compile the refguide. It'll keep dependencies up to date and consistent.

This is a pull for 8x only. I am aware that gradle is managing deps for 9x :) a big improvment!

gus-asf:
> This is only a problem if you already had it installed before it was upgraded several months ago. At that time, I sent mail to the Dev list explaining the need for people to specifically upgrade, and it’s explained in the Jira issue I used at that time. I’m on vacation, so I’ll leave searching for those things up to you. For people who never installed any of the dependencies, I believe the right version of the gem gets installed with jekyll-asciidoc gem, it just doesn’t upgrade it if you already have it, which is why I’d removed it from the README. And as soon as we move to Gradle, the need for any dependencies goes away entirely.
> 
> So what you choose to do with this PR with all that info is up to you, but I’d prefer it be clear and accurate about who needs the extra step. As I mentioned, I’m on vacation, so won’t be reviewing this again until next week at the earliest.

Ah, it seems I missed that email. It seems to have been in October last year and I was crazy busy with client work at that point with two projects trying to wrap up up simultaneously before Thanksgiving. I think it's probably good to have some note in the README for others who missed the email, or who perhaps only followed through on one computer, or on the odd chance someone used asciidoctor for something else previously. I will be clear that it's only necessary if asciidoctor is previously installed. I suspected you would have something to add here, and I appreciate the review, especially during your vacation. Sorry if this disrupted it, Enjoy!

PR #1701, opened 2020-07-27
SOLR-14671: Parsing dynamic ZK config sometimes cause NuberFormatException

janhoy:
See https://issues.apache.org/jira/browse/SOLR-14671

### DIFF
--- a/solr/CHANGES.txt
+++ b/solr/CHANGES.txt
@@ -164,6 +164,13 @@ Other Changes
 
 * SOLR-11868: Deprecate CloudSolrClient.setIdField, use information from Zookeeper (Erick Erickson)
 
+==================  8.6.1 ==================
+
+Bug Fixes
+---------------------
+
+* SOLR-14671: Parsing dynamic ZK config sometimes cause NuberFormatException (janhoy)
+
 ==================  8.6.0 ==================
 
 Consult the lucene/CHANGES.txt file for additional, low level, changes in this release.
--- a/solr/core/src/java/org/apache/solr/handler/admin/ZookeeperStatusHandler.java
+++ b/solr/core/src/java/org/apache/solr/handler/admin/ZookeeperStatusHandler.java
@@ -112,7 +112,7 @@ public class ZookeeperStatusHandler extends RequestHandlerBase {
           .map(h -> h.resolveClientPortAddress() + ":" + h.clientPort)
           .sorted().collect(Collectors.toList());
       List<String> dynamicHosts = zkDynamicConfig.getServers().stream()
-          .map(h -> h.resolveClientPortAddress() + ":" + h.clientPort)
+          .map(h -> h.resolveClientPortAddress() + ":" + (h.clientPort != null ? h.clientPort : "2181"))
           .sorted().collect(Collectors.toList());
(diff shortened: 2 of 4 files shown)


henrik242:
By the way, should the rest of the integer parsers (leaderPort etc) have similar null safety checks?

janhoy:
> By the way, should the rest of the integer parsers (leaderPort etc) have similar null safety checks?

Should not be needed, as the other ports are always required. See the JIRA for why clientPort is not always required even if it is documented as such in the normal case.

solr/core/src/java/org/apache/solr/handler/admin/ZookeeperStatusHandler.java
```diff
@@ -112,15 +112,21 @@ public void handleRequestBody(SolrQueryRequest req, SolrQueryResponse rsp) throw
           .map(h -> h.resolveClientPortAddress() + ":" + h.clientPort)
           .sorted().collect(Collectors.toList());
       List<String> dynamicHosts = zkDynamicConfig.getServers().stream()
-          .map(h -> h.resolveClientPortAddress() + ":" + h.clientPort)
+          .map(h -> h.resolveClientPortAddress() + ":" + (h.clientPort != null ? h.clientPort : "2181"))
```
janhoy, line 115: This is not bullet proof if e.g. user has `clientPort=1234` in `zoo.cfg` and in zkHost connection string. Then we'll add a warning that dynamic config differs from zkHost, which is not entirely true since we just lack the port part. We have no way from client to read the `clientPort` from server except from connecting to the server with 4LW *ont the clientPort* which is a chicken and egg. This hack will make the comparison work for default port, which is a compromise.

janhoy:
Updated the PR with some logic to handle the missing port numers. We now fall back to iterating the zk hosts from static zkHost connection string in this special case. This may mean that if a reconfiguration has happened, we will display the list of static hosts in Solr Admin instead of the list of dynamic hosts. That should be fine since Solr does not fully support dynamic reconfiguration of ZK anyway.

solr/core/src/java/org/apache/solr/handler/admin/ZookeeperStatusHandler.java
HoustonPutman: Maybe instead of defaulting to `2181`, we default to `hostsFromConnectionString.getServers().get(0).clientPort`? I imagine that would match a vast majority of scenarios.

janhoy: Agree, I'll do that change and merge

solr/CHANGES.txt
```diff
@@ -164,6 +164,13 @@ Other Changes
 
 * SOLR-11868: Deprecate CloudSolrClient.setIdField, use information from Zookeeper (Erick Erickson)
 
+==================  8.6.1 ==================
+
+Bug Fixes
+---------------------
+
+* SOLR-14671: Parsing dynamic ZK config sometimes cause NuberFormatException (janhoy)
```
henrik242, line 172: @janhoy Typo in Nu[m]berFormatException

PR #1704, opened 2020-07-28
Found that asciidoctor 1.5.6.2 cause build-site to fail on long anchors (#1700)

gus-asf:
* Found that asciidoctor 1.5.6.2 cause build-site to fail on long anchors, 2.0.10 definitely works.
* Clarity on when an upgrade is required, also fix obsolete wording from when PDFs were generated.


HoustonPutman:
Looks good to me!

PR #1706, opened 2020-07-29
LUCENE-9443: UnifiedHighlighter shouldn't close reader

dsmiley:
A regression from 8.6.  Don't close the underlying IndexReader.

https://issues.apache.org/jira/browse/LUCENE-9443

Instead of reverting the commit, there's something to be said for closing the wrapped reader since, after all, it's something you're supposed to close.  Since this is a wrapped reader created by the UH, it's easy to make its doClose do nothing.  Furthermore I tweaked a test that would fail prior to this change.  I checked precommit to ensure there is still no warning.

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -196,6 +196,13 @@ Other
 ---------------------
 (No changes)
 
+======================= Lucene 8.6.1 =======================
+
+Bug Fixes
+---------------------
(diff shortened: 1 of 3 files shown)


dsmiley:
Yup, and branch_8_6 -- the placement of the CHANGES.txt entry into 8.6.1 indicates this.

HoustonPutman:
Sorry, misread the branch this was going to. Understood.

ErickErickson:
LGTM, I really like the do-nothing close and the comment where it's actually used is very clear.

Thanks!

PR #1707, opened 2020-07-30
SOLR-14692: Allow 'method' specification on JSON Facet join domain transforms

gerlowskija:
# Description

Solr offers several different join implementations which can be switched off providing the "method" local-param on JoinQuery's.  Each of these implementations has different performance characteristics and can behave very differently depending on a user's data and use case.

When joins are used internally as a part of JSON Faceting's "join" domain-transform though, users have no way to specify which implementation they would like to use.  We should correct this by adding a "method" property to the join domain-transform.  This will let user's choose the join that's most performant for their use case during JSON Facet requests.

# Solution

Adds parsing to FacetRequest for a 'method' value, which is then forwarded to JoinQParserPlugin.createJoinQuery to create the appropriate join-type.

# Tests

None, yet.

# Checklist


solr/core/src/java/org/apache/solr/search/JoinQParserPlugin.java
```diff
@@ -182,9 +203,34 @@ public Query parse() throws SyntaxError {
    * @param subQuery the query to define the starting set of documents on the "left side" of the join
    * @param fromField "left side" field name to use in the join
    * @param toField "right side" field name to use in the join
+   * @param method indicates which implementation should be used to process the join.  Currently only 'index',
+   *               'dvWithScore', and 'topLevelDV' are supported.
    */
-  public static Query createJoinQuery(Query subQuery, String fromField, String toField) {
-    return new JoinQuery(fromField, toField, null, subQuery);
+  public static Query createJoinQuery(Query subQuery, String fromField, String toField, String method) {
+    final EnumSet<Method> methodWhitelist = EnumSet.of(Method.index, Method.topLevelDV, Method.dvWithScore);
```
munendrasn, line 210: I think this could be reused by creating static enumSet

solr/core/src/java/org/apache/solr/search/facet/FacetRequest.java
```diff
@@ -163,15 +166,23 @@ public boolean canBecomeNonEmpty() {
 
     /** Are we doing a query time join across other documents */
     public static class JoinField {
+      public static final String FROM_PARAM = "from";
+      public static final String TO_PARAM = "to";
+      public static final String METHOD_PARAM = "method";
+      public static final Set<String> SUPPORTED_JOIN_PROPERTIES = Sets.newHashSet(FROM_PARAM, TO_PARAM, METHOD_PARAM);
```
munendrasn, line 172: these variables could be package-private or preferably private

gerlowskija:
Thanks for the review Munendra; I made the changes you suggested.  Merging now.

PR #1709, opened 2020-07-31
LUCENE-9446: In boolean rewrite, remove MatchAllDocsQuery filter clauses

jtibshirani:
Previously, we only removed 'match all' FILTER clauses if there was at least one
MUST clause. Now they're also removed if there is another distinct FILTER clause.

This lets boolean queries like `#field:value #*:*` be written to `#field:value`.

### DIFF
--- a/lucene/core/src/test/org/apache/lucene/search/TestBooleanRewrites.java
+++ b/lucene/core/src/test/org/apache/lucene/search/TestBooleanRewrites.java
@@ -312,11 +312,19 @@ public class TestBooleanRewrites extends LuceneTestCase {
         .add(new TermQuery(new Term("foo", "baz")), Occur.MUST)
         .add(new MatchAllDocsQuery(), Occur.FILTER)
         .build();
-    BooleanQuery expected = new BooleanQuery.Builder()
+    Query expected = new BooleanQuery.Builder()
         .setMinimumNumberShouldMatch(bq.getMinimumNumberShouldMatch())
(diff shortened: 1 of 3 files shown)


lucene/core/src/test/org/apache/lucene/search/TestBooleanRewrites.java
```diff
@@ -312,12 +312,20 @@ public void testRemoveMatchAllFilter() throws IOException {
...
         .build();
     assertEquals(expected, searcher.rewrite(bq));
+
+    bq = new BooleanQuery.Builder()
+            .add(new TermQuery(new Term("foo", "bar")), Occur.FILTER)
+            .add(new MatchAllDocsQuery(), Occur.FILTER)
+            .build();
+    expected = new BoostQuery(new ConstantScoreQuery(
+            new TermQuery(new Term("foo", "bar"))), 0.0f);
+    assertEquals(expected, searcher.rewrite(bq));
```
jpountz, line 328: maybe also add a test for the case when there are two `MatchAllDocsQuery` queries as filters with no `MUST` clauses?

jtibshirani: 👍

jtibshirani:
Thanks for the review, I added another test and a changelog entry.

jpountz:
FYI I backported to 8.7 and moved the CHANGES entry from 9.0 to 8.7.

PR #1715, opened 2020-08-03
SOLR-14704 add download option to cloud.sh

gus-asf:
Initial cut at this, only tested on mac osx so far.


noblepaul:
@gus-asf is it possible to have some more information on the PR title?

gus-asf:
Yeah, good point commit message that it came from could have been better too. 

Basic idea here is to allow this script to curl a tarball from a URL rather than building a copy. This needs to be tested on linux too, and probably should sort out/document the interplay/precedence between this new option and the -r (recompile) option. This script also needs to be upgraded to use gradle on master, but last time I checked gradle wasn't building tarballs yet.

PR #1716, opened 2020-08-04
SOLR-14706: Fix support for default autoscaling policy

HoustonPutman:
https://issues.apache.org/jira/browse/SOLR-14706

This is to introduce backwards compatibility for the default autoscaling policy that was removed in 8.6.1 (and 8.7).

The upgrade notes now include the steps to remove the default autoscaling policy via the autoscaling API.

I have tested the following upgrade patterns, and am happy with the results. I'll do more tests if y'all think it is necessary.

8.6.0 -> 8.6.1 RC1 (fail)
8.6.0 -> this patch (success)
8.6.0 -> 8.6.1 RCI -> this patch (success)

I will also be porting this with a few documentation tweaks to 8x

### DIFF
--- a/solr/core/src/java/org/apache/solr/cloud/autoscaling/sim/SimScenario.java
+++ b/solr/core/src/java/org/apache/solr/cloud/autoscaling/sim/SimScenario.java
@@ -814,7 +814,9 @@ public class SimScenario implements AutoCloseable {
         values.put(key, val);
       }
       for (String node : nodes) {
-        scenario.cluster.getSimNodeStateProvider().simSetNodeValues(node, values);
+        Map<String, Object> newValues = new HashMap<>(scenario.cluster.getSimNodeStateProvider().simGetNodeValues(node));
+        newValues.putAll(values);
+        scenario.cluster.getSimNodeStateProvider().simSetNodeValues(node, newValues);
       }
     }
   }
--- a/solr/solr-ref-guide/src/solr-upgrade-notes.adoc
+++ b/solr/solr-ref-guide/src/solr-upgrade-notes.adoc
@@ -36,22 +36,27 @@ If you are upgrading from 7.x, see the section <<Upgrading from 7.x Releases>> b
 
 === Solr 8.6.1 (Upgrading from 8.6.0 only)
 
+See the https://cwiki.apache.org/confluence/display/SOLR/ReleaseNote861[8.6.1 Release Notes^]
+for an overview of the fixes included in Solr 8.6.1.
+
+When upgrading to 8.6.1 users should be aware of the following major changes from 8.6.0.
+
 *Autoscaling*
 
 * As mentioned in the 8.6 upgrade notes, a default autoscaling policy was provided starting in 8.6.0.
 This default autoscaling policy resulted in increasingly slow collection creation calls in large clusters (50+ collections).
 +
 In 8.6.1 the default autoscaling policy has been removed, and clouds will not use autoscaling unless a policy has explicitly been created.
-If your cloud has been upgraded to 8.6.0 and you wish to fix this performance degredations, upgrade to 8.6.1 and remove the default cluster policy via the following command.
+If your cloud is running 8.6.0 and **not using an explicit autoscaling policy**, upgrade to 8.6.1 and remove the default cluster policy and preferences via the following command.
 Replace `localhost:8983` with your Solr endpoint.
 +
 ```
-curl -X POST -H 'Content-type:application/json'  -d '{set-cluster-policy : []}' http://localhost:8983/api/cluster/autoscaling
+curl -X POST -H 'Content-type:application/json'  -d '{set-cluster-policy : [], set-cluster-preferences : []}' http://localhost:8983/api/cluster/autoscaling
 ```
 +
 For more details on autoscaling policies, see the section <<solrcloud-autoscaling-overview.adoc#cluster-policy,Cluster Policy>>.
 +
-This information is only relevant for users upgrading from 8.6.0. If upgrading from an earlier version, this warning can be ignored.
+This information is only relevant for users upgrading from 8.6.0. If upgrading from an earlier version to 8.6.1+, this warning can be ignored.
 
 === Solr 8.6
 
--- a/solr/solr-ref-guide/src/solr-upgrade-notes.adoc
+++ b/solr/solr-ref-guide/src/solr-upgrade-notes.adoc
@@ -42,7 +42,12 @@ If you are upgrading from 7.x, see the section <<Upgrading from 7.x Releases>> b
 This default autoscaling policy resulted in increasingly slow collection creation calls in large clusters (50+ collections).
 +
 In 8.6.1 the default autoscaling policy has been removed, and clouds will not use autoscaling unless a policy has explicitly been created.
-In order to fix the performance degradations introduced in 8.6.0, merely upgrade to 8.6.1.
+In order to fix the performance degradations introduced in 8.6.0, upgrade to 8.6.1 and remove the default cluster policy via the following command.
+Replace `localhost:8983` with your Solr endpoint.
++
+```
+curl -X POST -H 'Content-type:application/json'  -d '{set-cluster-policy : []}' http://localhost:8983/api/cluster/autoscaling
+```
 +
 For more details on autoscaling policies, see the section <<solrcloud-autoscaling-overview.adoc#cluster-policy,Cluster Policy>>.
 
--- a/solr/solr-ref-guide/src/solr-upgrade-notes.adoc
+++ b/solr/solr-ref-guide/src/solr-upgrade-notes.adoc
@@ -34,7 +34,7 @@ Detailed steps for upgrading a Solr cluster are in the section <<upgrading-a-sol
 
 If you are upgrading from 7.x, see the section <<Upgrading from 7.x Releases>> below.
 
-=== Solr 8.6.1
+=== Solr 8.6.1 (Upgrading from 8.6.0 only)
 
 *Autoscaling*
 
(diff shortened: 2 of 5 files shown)


solr/solrj/src/java/org/apache/solr/client/solrj/cloud/autoscaling/Clause.java
```diff
@@ -686,7 +686,7 @@ boolean isShardAbsent() {
       for (Row r : session.matrix) {
         computedValueEvaluator.node = r.node;
         SealedClause sealedClause = getSealedClause(computedValueEvaluator);
-        if (!sealedClause.getGlobalTag().isPass(r)) {
+        if (r.isLive() && !sealedClause.getGlobalTag().isPass(r)) {
```
HoustonPutman, line 689: I brought this back in from the revert because it looks like an improvement that was unrelated to the default policy.

I'm happy to remove it if y'all think I should.

solr/solr-ref-guide/src/solr-upgrade-notes.adoc
```diff
@@ -34,17 +34,24 @@ Detailed steps for upgrading a Solr cluster are in the section <<upgrading-a-sol
...
 * As mentioned in the 8.6 upgrade notes, a default autoscaling policy was provided starting in 8.6.0.
 This default autoscaling policy resulted in increasingly slow collection creation calls in large clusters (50+ collections).
 +
 In 8.6.1 the default autoscaling policy has been removed, and clouds will not use autoscaling unless a policy has explicitly been created.
-In order to fix the performance degradations introduced in 8.6.0, merely upgrade to 8.6.1.
+If your cloud has been upgraded to 8.6.0 and you wish to fix this performance degredations, upgrade to 8.6.1 and remove the default cluster policy via the following command.
+Replace `localhost:8983` with your Solr endpoint.
++
+```
+curl -X POST -H 'Content-type:application/json'  -d '{set-cluster-policy : []}' http://localhost:8983/api/cluster/autoscaling
```
gus-asf, line 49: The other caveat is probably that they shouldn't run this command if they have been actively using autoscaling before they upgraded to 8.6.0

HoustonPutman: I tried to make this more clear. Let me know if you think my wording is off

solr/solrj/src/java/org/apache/solr/client/solrj/cloud/autoscaling/Clause.java
```diff
@@ -117,10 +117,10 @@ private Clause(Map<String, Object> m) {
     strict = Boolean.parseBoolean(String.valueOf(m.getOrDefault("strict", "true")));
     Optional<String> globalTagName = m.keySet().stream().filter(Policy.GLOBAL_ONLY_TAGS::contains).findFirst();
     if (globalTagName.isPresent()) {
-      globalTag = parse(globalTagName.get(), m);
-      if (m.size() > 2) {
-        throw new RuntimeException("Only one extra tag supported for the tag " + globalTagName.get() + " in " + toJSONString(m));
+      if (m.size() > 3) {
```
gus-asf, line 120: from my very limited understanding of this code it appears that something like
```
    if (globalTagName.isPresent()) {
      validateGlobalTag(m, globalTagName.get());
      globalTag = parse(globalTagName.get(), m);
 ...
and 
...
  private void validateGlobalTag(Map<String, Object> m, String tagName) {
    if (m.size() > 2) {
      if (!m.containsKey("strict") || m.size() > 3) {
        throw new RuntimeException("Only, 'strict' and one extra tag supported for the tag " + tagName + " in " + toJSONString(m));
      }
    }
  }
```
would be the real intention here... but it would be great if @sigram could confirm that.

solr/solr-ref-guide/src/solr-upgrade-notes.adoc
gus-asf: Yeah that looks better :)

solr/solr-ref-guide/src/solr-upgrade-notes.adoc
```diff
@@ -34,17 +34,29 @@ Detailed steps for upgrading a Solr cluster are in the section <<upgrading-a-sol
...
 * As mentioned in the 8.6 upgrade notes, a default autoscaling policy was provided starting in 8.6.0.
 This default autoscaling policy resulted in increasingly slow collection creation calls in large clusters (50+ collections).
 +
 In 8.6.1 the default autoscaling policy has been removed, and clouds will not use autoscaling unless a policy has explicitly been created.
-In order to fix the performance degradations introduced in 8.6.0, merely upgrade to 8.6.1.
+If your cloud is running 8.6.0 and **not using an explicit autoscaling policy**, upgrade to 8.6.1 and remove the default cluster policy and preferences via the following command.
+Replace `localhost:8983` with your Solr endpoint.
++
+```
+curl -X POST -H 'Content-type:application/json'  -d '{set-cluster-policy : [], set-cluster-preferences : []}' http://localhost:8983/api/cluster/autoscaling
```
sigram, line 54: Setting preferences to `[]` actually removes the default preferences that have always been implicitly present, so this changes the behavior as compared with 8.5 and earlier. We should only reset the cluster_policy to `[]`.

solr/solr-ref-guide/src/solr-upgrade-notes.adoc
```diff
@@ -72,7 +84,9 @@ More information about this new feature is available in the section <<common-que
 
 *Autoscaling*
 
-* Solr now includes a default autoscaling policy.
+* **NOTE: The default autoscaling policy has been removed as of 8.6.1**
++
+Solr now includes a default autoscaling policy.
```
sigram, line 89: This statement conflicts with the statement above. 8.5.x and 8.6.1 includes default preferences but not the default policy.

solr/solrj/src/java/org/apache/solr/client/solrj/cloud/autoscaling/Clause.java
sigram: @gus-asf yes, that's the intent, though your pseudo-code is still incorrect - if there's no `strict` tag then `m.size() > 2` is already an error. In other words, for global tags we expect exactly 2 keys (tag and operand), with optional third key `strict`.

solr/solrj/src/java/org/apache/solr/client/solrj/cloud/autoscaling/Clause.java
sigram, line 689: +1, this was a genuine bug in 8.5.

solr/core/src/java/org/apache/solr/cloud/autoscaling/sim/SimScenario.java
```diff
@@ -814,7 +814,9 @@ public void execute(SimScenario scenario) throws Exception {
         values.put(key, val);
       }
       for (String node : nodes) {
-        scenario.cluster.getSimNodeStateProvider().simSetNodeValues(node, values);
+        Map<String, Object> newValues = new HashMap<>(scenario.cluster.getSimNodeStateProvider().simGetNodeValues(node));
```
sigram, line 817: +1, this fixed a genuine bug in < 8.6.0

solr/solr-ref-guide/src/solr-upgrade-notes.adoc
HoustonPutman: Hmmm, I'm a bit confused about that. So if I spin up a new cloud with 8.5, I don't see a `cluster-preferences`. It basically looks the same as a 8.6.1 cloud that has the above command run. So if we reverted the change of the defaulting patch, shouldn't a 8.5 cloud and 8.6.1 cloud with identical `autoscaling.json` ZNodes behave the same?

solr/solrj/src/java/org/apache/solr/client/solrj/cloud/autoscaling/Clause.java
HoustonPutman: I think that's the logic Gus used. It's equivalent to:

```java
private void validateGlobalTag(Map<String, Object> m, String tagName) {
    if (m.size() > 2) {
      if (!(m.containsKey("strict") && m.size() == 3)) {
        throw new RuntimeException("Only, 'strict' and one extra tag supported for the tag " + tagName + " in " + toJSONString(m));
      }
    }
  }
```

This will error unless there are exactly 3 keys and once of them is `strict`.

solr/solr-ref-guide/src/solr-upgrade-notes.adoc
gus-asf: Fresh install of 8.5.2 gives me:
```
{
  "triggers":{
    ".auto_add_replicas":{
      "name":".auto_add_replicas",
      "event":"nodeLost",
      "waitFor":120,
      "enabled":true,
      "actions":[
        {
          "name":"auto_add_replicas_plan",
          "class":"solr.AutoAddReplicasPlanAction"},
        {
          "name":"execute_plan",
          "class":"solr.ExecutePlanAction"}]},
    ".scheduled_maintenance":{
      "name":".scheduled_maintenance",
      "event":"scheduled",
      "startTime":"NOW",
      "every":"+1DAY",
      "enabled":true,
      "actions":[
        {
          "name":"inactive_shard_plan",
          "class":"solr.InactiveShardPlanAction"},
        {
          "name":"inactive_markers_plan",
          "class":"solr.InactiveMarkersPlanAction"},
        {
          "name":"execute_plan",
          "class":"solr.ExecutePlanAction"}]}},
  "listeners":{
    ".auto_add_replicas.system":{
      "beforeAction":[],
      "afterAction":[],
      "stage":[
        "STARTED",
        "ABORTED",
        "SUCCEEDED",
        "FAILED",
        "BEFORE_ACTION",
        "AFTER_ACTION",
        "IGNORED"],
      "trigger":".auto_add_replicas",
      "class":"org.apache.solr.cloud.autoscaling.SystemLogListener"},
    ".scheduled_maintenance.system":{
      "beforeAction":[],
      "afterAction":[],
      "stage":[
        "STARTED",
        "ABORTED",
        "SUCCEEDED",
        "FAILED",
        "BEFORE_ACTION",
        "AFTER_ACTION",
        "IGNORED"],
      "trigger":".scheduled_maintenance",
      "class":"org.apache.solr.cloud.autoscaling.SystemLogListener"}},
  "properties":{}}

```

HoustonPutman:
Tested this out locally upgrading 8.6.0 -> 8.6.1 RC1 (failure) -> this PR (success)

Also added you to the CHANGES.txt Gus, thanks for all the invaluable help!

PR #1718, opened 2020-08-05
SOLR-14702: Add Upgrade Notes and CHANGES entry

tflobbe:
This is just the CHANGES.txt entries, including upgrade notes
CC @MarcusSorealheis


HoustonPutman:
I think adding something in the ref guide upgrade notes too would be worthwile.

tflobbe:
Ah, good point, I seemed to remember some change with the release notes recently but I couldn't remember what it was

atris:
I took a look late -- but +1

PR #1711, opened 2020-08-02
SOLR-14702: Remove oppressive language (part1)

MarcusSorealheis:
# Description

This PR removes all references of "master slave" from the Solr code base. this should've happened years ago as it can result in alienating many people working with the project.

# Solution

Replaces master with leader and slave with follower.

# Tests

Run standard tests.

# Checklist

Please review the following and check all that apply:

- [x] I have reviewed the guidelines for [How to Contribute](https://wiki.apache.org/solr/HowToContribute) and my code conforms to the standards described there to the best of my ability.
- [x] I have created a Jira issue and added the issue ID to my pull request title.
- [x] I have given Solr maintainers [access](https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork) to contribute to my PR branch. (optional but recommended)
- [x] I have developed this patch against the `master` branch.
- [x] I have run `ant precommit` and the appropriate test suite.
- [x] I have added tests for my changes.
- [x] I have added documentation for the [Ref Guide](https://github.com/apache/lucene-solr/tree/master/solr/solr-ref-guide) (for Solr changes only).

PR #1719, opened 2020-08-06
SOLR-14702: Remove Master and Slave from Code Base and Docs (8.x)

tflobbe:
This PR brings the changes from https://github.com/apache/lucene-solr/pull/1711 into 8.x and add a commit to bring back compatibility. With this code change, a Solr instance running 8.7 should be able to replicate from a Solr instance running 8.x < 8.7, and a Solr instance running 9.0 should be able to replicate from an instance running 8.7
Some manual testing to check before merging this is listed here: https://issues.apache.org/jira/browse/SOLR-14708
I haven't validated the UI yet.

I kept the "bring back compatibility" changes in a single commit to make it easier to review the changes.

CC @MarcusSorealheis


solr/core/src/java/org/apache/solr/cloud/ReplicateFromLeader.java
```diff
@@ -76,12 +76,12 @@ public void startReplication(boolean switchTransactionLog) throws InterruptedExc
...
-      slaveConfig.add("fetchFromLeader", Boolean.TRUE);
-      slaveConfig.add(ReplicationHandler.SKIP_COMMIT_ON_MASTER_VERSION_ZERO, switchTransactionLog);
-      slaveConfig.add("pollInterval", pollIntervalStr);
+      NamedList<Object> followerConfig = new NamedList<>();
+      followerConfig.add("fetchFromLeader", Boolean.TRUE);
+      followerConfig.add(ReplicationHandler.SKIP_COMMIT_ON_LEADER_VERSION_ZERO, switchTransactionLog);
+      followerConfig.add("pollInterval", pollIntervalStr);
       NamedList<Object> replicationConfig = new NamedList<>();
-      replicationConfig.add("slave", slaveConfig);
+      replicationConfig.add("follower", followerConfig);
```
tflobbe, line 84: I left this intentionally using the new parameters because this configuration is only read locally

MarcusSorealheis: 🤨

MarcusSorealheis: left intentionally — what are you saying?

MarcusSorealheis: It should be left intentionally. It comes from the leader, right?

tflobbe: For 8.x branches, every time we set a parameter or configuration (for requests, etc) we use the legacy names. In this particular case, I'm using the new parameters even for setting, and the reason is that this is a configuration that is set and read locally only. The TLOG/PULL replicas create this to create their replication process.

tflobbe:
> I thought was changed in my PR...

They were, my point is that I didn't re-set these to legacy terms (after I backported your commit) because in this particular case we don't need it.

PR #1723, opened 2020-08-06
SOLR prometheus: simplify concurrent collection

dsmiley:
The intent of this is to simplify some concurrent code in the Prometheus exporter that I think is too confusing / contorted -- particularly Async.java.  Git blame points at @shalinmangar so I would love a review to see what you think.  I played with a few different approaches, and ultimately realized that we're working around using an Executor instead of an ExecutorService to benefit from invokeAll.  I wish Java didn't have a distinction between Executor & ExecutorService but there is and we should all just ignore plain Executor IMO.

I haven't run this in-the-field but I could do so locally.


dsmiley:
@ErickErickson the gradle validateLogCalls is complaining because I added a log statement without explicitly checking "log.isErrorEnabled" first.  validate-log-calls.gradle is *not* supposed to do that for an error level log.  Can you see what's going on please?

ErickErickson:
The first part of the line gives the most explicit statement: "cause: 'getMessage or getCause in log line’”, the latter part of the line is more generic.

This check was added as part of SOLR-14523, since these two calls were often used in places that provided too little information. If you think this is an exception, you can add //logok to the line and it will pass.

Erick

dsmiley:
Woops; sorry; I should have looked more clearly!  This is a "logok" situation I should do.

ErickErickson:
If you only knew how many times I’ve beaten my head against a wall for _hours_, then read something that was in front of my face all the time more carefully… ;)

PR #1724, opened 2020-08-07
SOLR-14684: CloudExitableDirectoryReaderTest failing about 25% of the time

CaoManhDat:


solr/solrj/src/java/org/apache/solr/client/solrj/impl/LBSolrClient.java
```diff
@@ -155,6 +159,7 @@ public ServerIterator(Req req, Map<String, ServerWrapper> zombieServers) {
       this.req = req;
       this.zombieServers = zombieServers;
       this.timeAllowedNano = getTimeAllowedInNanos(req.getRequest());
+      log.info("TimeAllowedNano:{}", this.timeAllowedNano);
```
dsmiley, line 162: Are you sure we should log at info level here?  This seems more like a debug situation.

CaoManhDat: Thank you, I totally forget that on creating this PR.

PR #1725, opened 2020-08-07
LUCENE-9449 Skip docs with _doc sort and "after"

mayya-sharipova:
Enhance DocComparator to provide an iterator over competitive documents
when searching with "after" FieldDoc.
This iterator can quickly position on the desired "after" document,
and skip all documents before "after" or even whole segments
that contain only documents before "after".

Related to LUCENE-9280


lucene/core/src/java/org/apache/lucene/search/FilteringDocLeafComparator.java
```diff
@@ -0,0 +1,157 @@
...
+ */
+public class FilteringDocLeafComparator implements FilteringLeafFieldComparator {
+    private final FieldComparator.DocComparator in;
+    private DocIdSetIterator topValueIterator; // iterator that starts from topValue if possible
+    private final int minDoc;
+    private final int maxDoc;
+    private final int docBase;
+    private boolean iteratorUpdated = false;
+
+    public FilteringDocLeafComparator(LeafFieldComparator in, LeafReaderContext context) {
```
jimczi, line 35: Can we force the `in` to be a `FieldComparator.DocComparator` ?

lucene/core/src/java/org/apache/lucene/search/FilteringFieldComparator.java
```diff
@@ -68,10 +68,12 @@ public int compareValues(T first, T second) {
    * @param comparator – comparator to wrap
    * @param reverse – if this sort is reverse
    * @param singleSort – true if this sort is based on a single field and there are no other sort fields for tie breaking
+   * @param hasAfter – true if this sort has after FieldDoc
    * @return comparator wrapped as a filtering comparator or the original comparator if the filtering functionality
    * is not implemented for it
    */
-  public static FieldComparator<?> wrapToFilteringComparator(FieldComparator<?> comparator, boolean reverse, boolean singleSort) {
+  public static FieldComparator<?> wrapToFilteringComparator(FieldComparator<?> comparator, boolean reverse, boolean singleSort,
+      boolean hasAfter) {
```
jimczi, line 76: Do we really need to add the `hasAfter` ? Can we check the if the `topValue` in the DocComparator is greater than 0 instead ?

lucene/core/src/java/org/apache/lucene/search/FieldValueHitQueue.java
```diff
@@ -160,18 +160,20 @@ private FieldValueHitQueue(SortField[] fields, int size, boolean filterNonCompet
    *          The number of hits to retain. Must be greater than zero.
    * @param filterNonCompetitiveDocs
    *    {@code true} If comparators should be allowed to filter non-competitive documents, {@code false} otherwise
+   * @param hasAfter
+   *    {@code true} If this sort has "after" FieldDoc
    */
   public static <T extends FieldValueHitQueue.Entry> FieldValueHitQueue<T> create(SortField[] fields, int size,
-      boolean filterNonCompetitiveDocs) {
+      boolean filterNonCompetitiveDocs, boolean hasAfter) {
```
jimczi, line 167: Can we avoid adding `hasAfter` here ? See my comment below.

lucene/core/src/java/org/apache/lucene/search/FieldValueHitQueue.java
```diff
@@ -121,7 +121,7 @@ protected boolean lessThan(final Entry hitA, final Entry hitB) {
   }
   
   // prevent instantiation and extension.
-  private FieldValueHitQueue(SortField[] fields, int size, boolean filterNonCompetitiveDocs) {
+  private FieldValueHitQueue(SortField[] fields, int size, boolean filterNonCompetitiveDocs, boolean hasAfter) {
```
jimczi, line 124: Not sure that `hasAfter` is really needed here.

lucene/core/src/java/org/apache/lucene/search/FieldValueHitQueue.java
```diff
@@ -95,8 +95,8 @@ protected boolean lessThan(final Entry hitA, final Entry hitB) {
    */
   private static final class MultiComparatorsFieldValueHitQueue<T extends FieldValueHitQueue.Entry> extends FieldValueHitQueue<T> {
 
-    public MultiComparatorsFieldValueHitQueue(SortField[] fields, int size, boolean filterNonCompetitiveDocs) {
-      super(fields, size, filterNonCompetitiveDocs);
+    public MultiComparatorsFieldValueHitQueue(SortField[] fields, int size, boolean filterNonCompetitiveDocs, boolean hasAfter) {
```
jimczi, line 98: Not sure that `hasAfter` is really needed here.

lucene/core/src/java/org/apache/lucene/search/FilteringDocLeafComparator.java
```diff
@@ -0,0 +1,157 @@
...
+
+import org.apache.lucene.index.LeafReaderContext;
+
+import java.io.IOException;
+
+/**
+ * This comparator is used when there is sort by _doc asc together with "after" FieldDoc.
+ * The comparator provides an iterator that can quickly skip to the desired "after" document.
+ */
+public class FilteringDocLeafComparator implements FilteringLeafFieldComparator {
```
jimczi, line 27: Maybe rename to `AfterDocLeafComparator` ?

lucene/core/src/java/org/apache/lucene/search/FilteringFieldComparator.java
mayya-sharipova: At that moment topValue is not set yet, it will be set later in the constructor of `PagingFieldCollector`.

lucene/core/src/java/org/apache/lucene/search/FieldValueHitQueue.java
mayya-sharipova: At this point of time, topValue for comparators is not set yet, that's why we need `hasAfter`.
As an alternative to this implementation, we can pass `FieldDoc after` to `FieldValueHitQueue.create` and setTopValue during `FieldValueHitQueue` creation.

lucene/core/src/java/org/apache/lucene/search/FilteringDocLeafComparator.java
mayya-sharipova: I like `AfterDocLeafComparator`, but I renamed to `FilteringAfterDocLeafComparator` for consistency with all other filtering comparators. Please let me know if you still like it to be renamed

mayya-sharipova:
@jimczi Thank you for the initial feedback.  I tried to address it, can you please continue the review

mayya-sharipova:
@jimczi Thank you for the initial feedback. I have tried to address it. Please continue to review, when you have time.

 I have caught up with @jimczi offline, and his main comment was that we need to redesign all comparators is such a way that they all provide skipping functionality by default (without wrapping them into FilteringComparator and FilteringLeafComparator).   While this be a target of following PRs, the goal of this PR is do this only for `DocComparator`.

So  addressing @jimczi's feedback,  in this PR:
-  `DocComparator` was redesigned to have skipping functionality be default.

lucene/core/src/java/org/apache/lucene/search/comparators/DocComparator.java
```diff
@@ -0,0 +1,185 @@
...
+            } else {
+                this.minDoc = -1;
+                this.maxDoc = -1;
+                this.topValueIterator = null;
+            }
+        }
+
+        @Override
+        public void setBottom(int slot) {
+            bottom = docIDs[slot];
```
jimczi, line 108: We should be able to early terminate here if the total hits threshold has been reached. If it's not reached yet, we can early terminate later in `setHitsThresholdReached`.

jimczi: Any query sorted by doc id can early terminate after N matches. That's an important aspect of the optimization since it can be handled by the hits threshold transparently. If there is no `after` value, the threshold should be an upper bound of the number of document that we will collect in the comparator.

lucene/core/src/java/org/apache/lucene/search/FilteringLeafFieldComparator.java
```diff
@@ -32,8 +33,22 @@
...
+   * hits threshold is enough to start updating their iterators, even when queue is not yet full.
    */
-  void setCanUpdateIterator() throws IOException;
+  void setHitsThresholdReached() throws IOException;
+
+  /**
+   * Informs this leaf comparator that queue has become full.
+   * This method is called from a collector when queue becomes full.
+   */
+  void setQueueFull() throws IOException;
```
jimczi, line 47: That seems redundant. I think we can add a single method: setHitsThresholdReached(int). We have all the other contexts needed in the LeafFieldComparator. The complexity is that `setBottom `and `setHitsThresholdReached(int)` are called on the leaves even though they are global events. Maybe forcing the hierarchy of the filtering fields to use a `SimpleFieldComparator` could simplify things ? This way you don't have to repeat redundant informations (hits threshold and queue full) on each leave.

lucene/core/src/java/org/apache/lucene/search/TopFieldCollector.java
```diff
@@ -150,11 +150,15 @@ public void setScorer(Scorable scorer) throws IOException {
...
-        totalHitsRelation = TotalHits.Relation.GREATER_THAN_OR_EQUAL_TO;
+
+      if (filteringLeafComparator != null && hitsThresholdChecker.isThresholdReached()) {
+        // hitsThreshold was reached in previous segments, notify this segment's leaf comparator about it
+        filteringLeafComparator.setHitsThresholdReached();
+        // if queue became full in previous segments, notify this segment's leaf comparator about it
+        if (queueFull) filteringLeafComparator.setQueueFull();
+        if (filteringLeafComparator.iteratorUpdated()) {
+          totalHitsRelation = TotalHits.Relation.GREATER_THAN_OR_EQUAL_TO;
+        }
```
jimczi, line 161: See my previous comment, I think we should call `setHitsThresholdReached` once, when the threshold is reached. 
Same for `setQueueFull` that is already call in `setBottom`. Said differently, this part could be entirely removed ;).

jimczi (changes requested):
Thanks @mayya-sharipova . I left two comments that I find important, one regarding the introduction of new functions in the filtering leaf comparator. I don't think they are needed. And the other regarding the early termination of queries sorted by doc id. We don't need to visit more documents if the queue is full and the the hits threshold has been reached.

lucene/core/src/java/org/apache/lucene/search/FieldComparator.java
```diff
@@ -136,6 +136,13 @@ public int compareValues(T first, T second) {
     }
   }
 
+  /**
+   * Informs the comparator that sorting is done in reverse.
+   * This is necessary only for skipping functionality.
+   */
+  public void setReverse() {
```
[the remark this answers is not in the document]
mayya-sharipova: addressed in 746c8fa

lucene/core/src/java/org/apache/lucene/search/FilteringLeafFieldComparator.java
mayya-sharipova: addressed in 746c8fa

lucene/core/src/java/org/apache/lucene/search/TopFieldCollector.java
mayya-sharipova: addressed in 746c8fa

lucene/core/src/java/org/apache/lucene/search/comparators/DocComparator.java
mayya-sharipova: @jimczi  Thank you for your comment. I would like to clarify something:
> can early terminate after N matches

What's the way to terminate after N matches here? Is it to update an iterator to an empty iterator?

Isn't this termination already handled in TopFieldCollector with the code around `canEarlyTerminate`?  
Is the plan to remove the code around `canEarlyTerminate` in `TopFieldCollector`? Should we do after we also handle the case on early termination with the same index and query sort?

mayya-sharipova:
@jimczi I have tried to address your comments in 746c8fa.  Can you please continue to review when you have time.

lucene/core/src/java/org/apache/lucene/search/comparators/DocComparator.java
jimczi: > What's the way to terminate after N matches here? Is it to update an iterator to an empty iterator?

I think so, yes. Updating to an empty iterator is what we do for constant score queries for instance. 

> Isn't this termination already handled in TopFieldCollector with the code around canEarlyTerminate?

The code is only for sorted index and while I think that we should move this code in the sort comparator, I agree that it's out of the scope of this PR. Early termination should be handled in the field comparator so that we don't need to add new logic in the main collector.

mayya-sharipova: @jimczi Thanks for the feedback. I have added an early termination after N matches for DocComparator in 21de242.

> The code is only for sorted index 

We also have an early termination in a collector based on [_doc order](https://github.com/apache/lucene-solr/blob/master/lucene/core/src/java/org/apache/lucene/search/TopFieldCollector.java#L177-L180).   I can remove this code later once I can make sure that all queries use `DefaultBulkScorer` that uses collector's iterator.

jimczi: Right, sorry I forgot that we added the early termination logic for _doc order in the collector. Maybe that's ok to leave it as it is then. We can revise after we ensure that all bulk scorer uses the collector's iterator ?

mayya-sharipova: @jimczi That sounds good to me.  Thank you for the feedback.  Do you have any further comments for this PR?

lucene/core/src/java/org/apache/lucene/search/comparators/DocComparator.java
```diff
@@ -0,0 +1,183 @@
...
+    private int bottom;
+    private int topValue;
+    private boolean topValueSet;
+    private boolean bottomValueSet;
+    private boolean hitsThresholdReached;
+
+    /** Creates a new comparator based on document ids for {@code numHits} */
+    public DocComparator(int numHits, boolean reverse) {
+        this.docIDs = new int[numHits];
+        // skipping functionality is enabled if we are sorting by _doc in asc order
```
jimczi, line 43: Should it be activated only if the comparator is used as the primary sort ?

lucene/core/src/java/org/apache/lucene/search/comparators/DocComparator.java
```diff
@@ -0,0 +1,183 @@
...
+        @Override
+        public void setHitsThresholdReached() {
+            hitsThresholdReached = true;
+            updateIterator();
+        }
+
+        private void updateIterator() {
+            if (enableSkipping == false || hitsThresholdReached == false) return;
+            if (bottomValueSet) {
+                // since we've collected top N matches, we can early terminate
```
jimczi, line 170: Can you add a comment explaining that early termination is already implemented in the collector but we'll remove in a follow up ?

jimczi (commented):
I left two questions but it's getting close. Thanks for iterating on this @mayya-sharipova !

lucene/core/src/java/org/apache/lucene/search/comparators/DocComparator.java
mayya-sharipova: addressed in 485fe4f

lucene/core/src/java/org/apache/lucene/search/comparators/DocComparator.java
mayya-sharipova: @jimczi Thanks Jim. We also had an implicit protection for a primary sort in [MultiLeafFieldComparator.java](https://github.com/apache/lucene-solr/pull/1725/files#diff-fbcae9439c314674b6d465e171fe535eR95) i, but I agree it is good to explicitly indicate a primary sort. 

Addressed in 485fe4f

jimczi: > We also had an implicit protection for a primary sort 

Ok I see so the expectation is that `setHitsThresholdReached` is only called on the primary sort and we use it as a signal to enable skipping. 

> Addressed in 485fe4f

I was thinking that we could use the information exposed in [`SortField#getComparator`](https://github.com/apache/lucene-solr/blob/master/lucene/core/src/java/org/apache/lucene/search/SortField.java#L448), we don't need to add a new callback. However, I think we can rely on what you describe above, `setHitsThresholdReached` and `competitiveIterator` are the callbacks that we need to enable the optimization. Sorry for the back and forth but I we probably don't need 485fe4f ;).

mayya-sharipova: @jimczi Thanks for the feedback.

> I was thinking that we could use the information exposed in SortField#getComparator, we don't need to add a new callback

Good point, I haven't noticed that before, I've refactored the code to use `sortPos` from this function.


> However, I think we can rely on what you describe above, setHitsThresholdReached and competitiveIterator are the callbacks that we need to enable the optimization.   ... we probably don't need 485fe4f

Doing sort optimization only on primary sort in `MultiLeafFieldComparator` is implicit and some refactoring of it may bring wrong result. So I think it is good to have extra protection

lucene/core/src/test/org/apache/lucene/search/TestFieldSortOptimizationSkipping.java
```diff
@@ -290,5 +299,114 @@ public void testFloatSortOptimization() throws IOException {
...
+    writer.close();
+    reader.close();
+    dir.close();
+  }
+
+
+  public void testDocSortOptimization() throws IOException {
+    final Directory dir = newDirectory();
+    final IndexWriter writer = new IndexWriter(dir, new IndexWriterConfig());
+    final int numDocs = atLeast(1500);
```
jimczi, line 379: why do you need that many documents ? `100` should be enough, no ?

lucene/core/src/test/org/apache/lucene/search/TestFieldSortOptimizationSkipping.java
```diff
@@ -290,5 +299,114 @@ public void testFloatSortOptimization() throws IOException {
...
+      searcher.search(new MatchAllDocsQuery(), collector);
+      TopDocs topDocs = collector.topDocs();
+      assertEquals(topDocs.scoreDocs.length, numHits);
+      for (int i = 0; i < numHits; i++) {
+        assertEquals(i, topDocs.scoreDocs[i].doc);
+      }
+      assertTrue(collector.isEarlyTerminated());
+      // check that very few hits were collected
+      assertTrue(topDocs.totalHits.value < 5);
+    }
```
jimczi, line 405: Can you add a test with a boolean query or a simple filter ?

lucene/core/src/test/org/apache/lucene/search/TestFieldSortOptimizationSkipping.java
```diff
@@ -290,5 +299,114 @@ public void testFloatSortOptimization() throws IOException {
     dir.close();
   }
 
+  public void testDocSortOptimizationWithAfter() throws IOException {
+    final Directory dir = newDirectory();
+    final IndexWriter writer = new IndexWriter(dir, new IndexWriterConfig());
+    final int numDocs = atLeast(1500);
```
jimczi, line 305: `1500` seems big for a unit test ;)

lucene/core/src/test/org/apache/lucene/search/TestFieldSortOptimizationSkipping.java
```diff
@@ -290,5 +299,114 @@ public void testFloatSortOptimization() throws IOException {
...
+        writer.commit();
+      }
+    }
+    final IndexReader reader = DirectoryReader.open(writer);
+    IndexSearcher searcher = new IndexSearcher(reader);
+    final int numHits = 3;
+    final int totalHitsThreshold = 3;
+    final int searchAfter = 1400;
+
+    // sort by _doc with search after should trigger optimization
```
jimczi, line 319: you can try different values of  search after in a loop to increase the coverage?

jimczi (approved):
I left some comments around testing but the changes  look good to me.

mayya-sharipova:
@jimczi Thanks for the review so far, I am wondering if you have any further comments?

lucene/core/src/test/org/apache/lucene/search/TestFieldSortOptimizationSkipping.java
mayya-sharipova: Addressed in 493d9eb

lucene/core/src/test/org/apache/lucene/search/TestFieldSortOptimizationSkipping.java
mayya-sharipova: Addressed in 493d9eb

lucene/core/src/test/org/apache/lucene/search/TestFieldSortOptimizationSkipping.java
mayya-sharipova: Addressed in 493d9eb

PR #1726, opened 2020-08-07
SOLR-14722: timeAllowed should track from req creation

dsmiley:
https://issues.apache.org/jira/browse/SOLR-14722


solr/core/src/java/org/apache/solr/search/SolrQueryTimeoutImpl.java
```diff
@@ -42,23 +44,23 @@ public static SolrQueryTimeoutImpl getInstance() {
   }
 
   /**
-   * Get the current value of timeoutAt.
+   * The time (nanoseconds) at which the request should be considered timed out.
    */
-  public static Long get() {
+  public static Long getTimeoutAtNs() {
```
dsmiley, line 49: For back-compat sake, on 8x I could retain the former method as Deprecated... but I'm unsure if anyone calls it.

solr/core/src/java/org/apache/solr/search/SolrQueryTimeoutImpl.java
```diff
@@ -67,8 +69,21 @@ public boolean shouldExit() {
...
+    } else {
+      reset();
+    }
+  }
+
+  /**
+   * Sets the time allowed (milliseconds), assuming we start a timer immediately.
+   * You should probably invoke {@link #set(SolrQueryRequest)} instead.
    */
   public static void set(Long timeAllowed) {
```
madrob, line 88: should this be a primitive instead of a boxed type?

solr/core/src/java/org/apache/solr/search/SolrQueryTimeoutImpl.java
```diff
@@ -67,8 +69,21 @@ public boolean shouldExit() {
...
 
   /**
-   * Method to set the time at which the timeOut should happen.
-   * @param timeAllowed set the time at which this thread should timeout.
+   * Sets or clears the time allowed based on how much time remains from the start of the request plus the configured
+   * {@link CommonParams#TIME_ALLOWED}.
+   */
+  public static void set(SolrQueryRequest req) {
+    long timeAllowed = req.getParams().getLong(CommonParams.TIME_ALLOWED, -1L);
+    if (timeAllowed >= 0L) {
```
madrob, line 77: Should be `>`, not `>=`. Doc on time allowed state that zero is no timeout, not immediate timeout. Looks like we were previously inconsistent about this.

solr/core/src/java/org/apache/solr/search/SolrQueryTimeoutImpl.java
dsmiley: Oh yeah; I forgot that -- indeed a primitive.  It's weird it was boxed.

solr/core/src/java/org/apache/solr/search/SolrQueryTimeoutImpl.java
atris: This seems inconsistent -- should we not be marking no timeout as -1?

dsmiley: `>` vs `>=` is debatable; there's an argument both ways.  I suspect there's a test for it this way but moreover, I don't think we should change it.  It's fine.

madrob: https://github.com/apache/lucene-solr/pull/1726/files#diff-d8beef800870f194d61993d701fd9cc2L77 has `>=`
https://github.com/apache/lucene-solr/pull/1726/files#diff-65e9f3712efc1ec962ea82a04a1d7aa1L104 has `>`

Either way, please update the docs at https://github.com/apache/lucene-solr/blob/master/solr/solrj/src/java/org/apache/solr/common/params/CommonParams.java#L162 because they are absolutely wrong (`>= 0` means no timeout???)

dsmiley: MLT Handler was the exception it's used *way* less often than SearchHandler and so I ignored that.  I agree that the docs are wrong; I will update them.

PR #1728, opened 2020-08-08
SOLR-14596: equals/hashCode for common SolrRequest classes

gerlowskija:
# Description

Currently, many SolrRequest classes (e.g. UpdateRequest) don't implement `equals()` or `hashCode()`.  This isn't a problem for Solr's normal operation, but it can be a barrier in unit testing SolrJ client code. `equals()` implementations would make it much easier to assert that client code is building the request that developers think it's building.

# Solution

This takes the first baby step towards remedying this by adding `equals` and `hashCode` implementations for a few SolrRequest implementations.  Specifically the base SolrRequest, UpdateRequest, and CollectionAdminRequest and all its subclasses.

In an effort to simplify, I've standardized on using EqualsBuilder/HashCodeBuilder for these classes.  It relies on apache-commons, but since that's already a dependency of SolrJ I don't think this is a problem.  It results in much simpler code than writing the required logic by hand.

# Tests

None yet but some forthcoming.

# Checklist

- [X] I have added documentation for the [Ref Guide](https://github.com/apache/lucene-solr/tree/master/solr/solr-ref-guide) (for Solr changes only).

### DIFF
--- a/solr/solrj/src/java/org/apache/solr/client/solrj/ResponseParser.java
+++ b/solr/solrj/src/java/org/apache/solr/client/solrj/ResponseParser.java
@@ -18,6 +18,9 @@ package org.apache.solr.client.solrj;
 
 import java.io.Reader;
 import java.io.InputStream;
+
+import org.apache.commons.lang.builder.EqualsBuilder;
+import org.apache.commons.lang.builder.HashCodeBuilder;
 import org.apache.solr.common.util.NamedList;
 
 /**
@@ -48,5 +51,32 @@ public abstract class ResponseParser
   public String getVersion()
   {
     return "2.2";
+  }
+
+  @Override
+  public int hashCode() {
+    return new HashCodeBuilder()
+        .append(getWriterType())
+        .append(getContentType())
+        .append(getVersion())
+        .toHashCode();
+  }
+
+  @Override
+  public boolean equals(Object rhs) {
+    if (rhs == null || getClass() != rhs.getClass()) {
+      return false;
+    } else if (this == rhs) {
+      return true;
+    } else if (hashCode() != rhs.hashCode()) {
+      return false;
+    }
+
+    final ResponseParser rhsCast = (ResponseParser) rhs;
+    return new EqualsBuilder()
+        .append(getWriterType(), rhsCast.getWriterType())
+        .append(getContentType(), rhsCast.getContentType())
+        .append(getVersion(), rhsCast.getVersion())
+        .isEquals();
   }
 }
--- a/solr/solrj/src/java/org/apache/solr/client/solrj/SolrRequest.java
+++ b/solr/solrj/src/java/org/apache/solr/client/solrj/SolrRequest.java
@@ -26,6 +26,8 @@ import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
 
+import org.apache.commons.lang.builder.EqualsBuilder;
+import org.apache.commons.lang.builder.HashCodeBuilder;
 import org.apache.solr.client.solrj.request.RequestWriter;
 import org.apache.solr.common.params.SolrParams;
 import org.apache.solr.common.util.ContentStream;
@@ -90,6 +92,7 @@ public abstract class SolrRequest<T extends SolrResponse> implements Serializabl
 
   private String basicAuthUser, basicAuthPwd;
 
+  // Can be used as an alternative to the base-url generally specified at the client-level.  Ignored by most clients.
   private String basePath;
 
   @SuppressWarnings({"rawtypes"})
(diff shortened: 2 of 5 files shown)


solr/solrj/src/java/org/apache/solr/client/solrj/ResponseParser.java
```diff
@@ -49,4 +52,31 @@ public String getVersion()
...
+  public boolean equals(Object rhs) {
+    if (rhs == null || getClass() != rhs.getClass()) {
+      return false;
+    } else if (this == rhs) {
+      return true;
+    } else if (hashCode() != rhs.hashCode()) {
+      return false;
+    }
+
+    final ResponseParser rhsCast = (ResponseParser) rhs;
```
ErickErickson, line 75: I'm curious about the comparison between Objects.hash(bunch of params) and the commons builder. I'm not advocating you change to Objects.hash, rather wondering whether there's a particular advantage to using one over the other.

solr/solrj/src/java/org/apache/solr/client/solrj/SolrRequest.java
```diff
@@ -244,6 +247,8 @@ public String getBasePath() {
   public void addHeader(String key, String value) {
     if (headers == null) {
       headers = new HashMap<>();
+      final HashMap<String, String> asdf = new HashMap<>();
```
madrob, line 250: what?

solr/solrj/src/test/org/apache/solr/client/solrj/request/TestUpdateRequest.java
```diff
@@ -17,52 +17,164 @@
...
+import static org.junit.Assert.assertNotEquals;
+
 public class TestUpdateRequest {
 
   @Rule
   public ExpectedException exception = ExpectedException.none();
 
   @Before
   public void expectException() {
-    exception.expect(NullPointerException.class);
```
madrob, line 34: remove the whole method?

solr/solrj/src/test/org/apache/solr/client/solrj/request/TestUpdateRequest.java
```diff
@@ -17,52 +17,164 @@
...
     UpdateRequest req = new UpdateRequest();
     req.add(Arrays.asList(new SolrInputDocument(), new SolrInputDocument(), null));
   }
 
+  @Test
+  public void testEqualsMethod() {
+    final SolrInputDocument doc1 = new SolrInputDocument("id", "1", "value_s", "foo");
+    final SolrInputDocument doc2 = new SolrInputDocument("id", "2", "value_s", "bar");
+    final SolrInputDocument doc3 = new SolrInputDocument("id", "3", "value_s", "baz");
+/*
```
madrob, line 92: left over from other testing?

solr/solrj/src/java/org/apache/solr/client/solrj/ResponseParser.java
madrob: I think I prefer Objects.hash, but I'm not sure why? Definitely willing to be convinced the other way if there's a reason or a difference or even if they're equivalent and there is already inertia here.

gerlowskija: I didn't make an intentional choice against Objects.hash - I'm just more familiar with HashCodeBuilder, so that's what I used.

Anyone have a concrete reason to prefer one vs the other?  If not, I'll do a little digging to see if one has advantages and report back.

Also relevant to this discussion is that @dsmiley made an argument in the [jira ticket](https://issues.apache.org/jira/browse/SOLR-14596) that hashCode impls aren't worth the maintenance burden and should be skipped here.  Since you guys showed some interest in the PR, either of you want to cast a vote there to settle things out more conclusively?  (I don't agree personally with his argument, but am happy to go with a clear consensus.)

solr/solrj/src/test/org/apache/solr/client/solrj/request/TestUpdateRequest.java
gerlowskija: Yes.  My bad - the tests are a bit WIP/experimental at this point.  I ran out of weekend-time partially into starting them.  I made a comment about this in the jira, but should've mentioned it here where it'd come more easily to the attention of reviewers.

gerlowskija:
> Were these autogenerated? By which IDE?

They weren't, no.  But in many cases they could be.  I skipped autogeneration because in a lot of cases I needed to dig into the instance variables in question to understand (1) whether they were relevant for equality, and (2) whether the instance variable types had adequate hashCode/equals methods themselves (e.g. SolrResponse.equals relies on ResponseParser.equals which didn't exist).

But I've got nothing against autogenerated methods if they're double-checked, and I might go that route in other parts of SolrJ. 🤷

risdenk:
@gerlowskija is this still valid? It looks like it might be modulo some merge conflicts in main.

PR #1730, opened 2020-08-10
SOLR-14680: Provide an implementation for the new SolrCluster API

noblepaul:
PR implementing #1694


noblepaul:
Planning to commit this soon

chatman:
I'm actively reviewing this. Please give me 2 days more before merging.

murblanc:
I could only look briefly on the phone screen (vacation) but two comments:

Separating out the new lazy implementations into another PR and keeping this one for adding interfaces to internal classes would have made reviewing easier.

Are there places in the code where currently the concrete classes are used and that could be changed to use the interfaces instead? In other words, how/where would these interfaces be used?

noblepaul:
> Separating out the new lazy implementations into another PR and keeping this one for adding interfaces to internal classes would have made reviewing easier.

Yeah, that was the other PR #1694 , I shall revive it. 

>Are there places in the code where currently the concrete classes are used and that could be changed to use the interfaces instead? In other words, how/where would these interfaces be used?

The current concrete classes do not use/implement these interfaces. These interfaces will only be a part of implementations. for instance, the `LazySolrCluster` is one of the impl. In the future we should add a couple more

noblepaul:
I realized that I had to tweak the APIs after writing an implementation. I have updated the API-only PR #1694 to reflect the latest

murblanc:
> The current concrete classes do not use/implement these interfaces. These interfaces will only be a part of implementations. for instance, the `LazySolrCluster` is one of the impl. In the future we should add a couple more

@noble what is the target use case of the interface and lazy implementation? I thought your aim was to create interfaces to existing internal classes therefore I expected them to implement these interfaces so these interfaces to be used in the code replacing the actual classes...
Maybe it's just me not understading your intention here.

noblepaul:
>what is the target use case of the interface and lazy implementation? 

The objectives are many

-  Totally refactor Solr code base to minimize dependencies on concrete classes. This enables us to do simulation and testing, make code more readable, and enable refactoring
- As we move to  a new mode for Solr with a lean core and packages/plugins, we want to have less API surface area against which the plugins are written. This enables us the plugins to work against a wider range of versions without rewriting/recompiling

The problem with the existing classes implementing the interfaces is that, users of the APIs will cast these objects to the underlying concrete classes, which defeats the purpose

PR #1731, opened 2020-08-10
LUCENE-9451 Sort.rewrite does not always return this when unchanged

madrob:
https://issues.apache.org/jira/browse/LUCENE-9451

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/DoubleValuesSource.java
+++ b/lucene/core/src/java/org/apache/lucene/search/DoubleValuesSource.java
@@ -275,12 +275,12 @@ public abstract class DoubleValuesSource implements SegmentCacheable {
       this.value = value;
       this.doubleValues = new DoubleValues() {
         @Override
-        public double doubleValue() throws IOException {
+        public double doubleValue() {
           return value;
         }
 
         @Override
-        public boolean advanceExact(int doc) throws IOException {
+        public boolean advanceExact(int doc) {
           return true;
         }
       };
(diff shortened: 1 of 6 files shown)


lucene/core/src/java/org/apache/lucene/search/DoubleValuesSource.java
```diff
@@ -456,13 +456,16 @@ public String toString() {
 
     @Override
     public SortField rewrite(IndexSearcher searcher) throws IOException {
-      DoubleValuesSortField rewritten = new DoubleValuesSortField(producer.rewrite(searcher), reverse);
+      DoubleValuesSource rewrittenSource = producer.rewrite(searcher);
+      if (rewrittenSource == producer) {
```
madrob, line 460: This might be better as an object equality check instead of reference equality, I'm not sure.

jpountz: Reference equality is the right check.

lucene/core/src/java/org/apache/lucene/search/Sort.java
```diff
@@ -177,12 +177,12 @@ public Sort rewrite(IndexSearcher searcher) throws IOException {
     SortField[] rewrittenSortFields = new SortField[fields.length];
     for (int i = 0; i < fields.length; i++) {
       rewrittenSortFields[i] = fields[i].rewrite(searcher);
-      if (fields[i] != rewrittenSortFields[i]) {
+      if (! fields[i].equals(rewrittenSortFields[i])) {
```
jpountz, line 180: this should remain a reference equality check

lucene/core/src/test/org/apache/lucene/search/TestSort.java
```diff
@@ -878,4 +878,22 @@ public void testMultiSort() throws IOException {
...
+      Sort sort = new Sort(longSource.getSortField(false));
+
+      assertSame(sort, sort.rewrite(searcher));
+
+      DoubleValuesSource doubleSource = DoubleValuesSource.constant(1.0);
+      sort = new Sort(doubleSource.getSortField(false));
+
+      assertSame(sort, sort.rewrite(searcher));
+    }
+  }
```
jpountz, line 898: This test looks like it belongs to TestDoubleValuesSource/TestLongValuesSource more than it belongs to TestSort?

PR #1732, opened 2020-08-10
Clean up many small fixes

madrob:
* Abstract classes don't need public constructors since they can only be
  called by subclasses
* Don't escape html characters in `@code` tags in javadoc
* Fixed a few int/long arithmetic
* Use Arrays.toString instead of implicit byte[].toString

### DIFF
--- a/lucene/codecs/src/java/org/apache/lucene/codecs/blocktreeords/OrdsBlockTreeTermsWriter.java
+++ b/lucene/codecs/src/java/org/apache/lucene/codecs/blocktreeords/OrdsBlockTreeTermsWriter.java
@@ -604,7 +604,7 @@ public final class OrdsBlockTreeTermsWriter extends FieldsConsumer {
           assert ent.isTerm: "i=" + i;
 
           PendingTerm term = (PendingTerm) ent;
-          assert StringHelper.startsWith(term.termBytes, prefix): "term.term=" + term.termBytes + " prefix=" + prefix;
+          assert StringHelper.startsWith(term.termBytes, prefix): term + " prefix=" + prefix;
           BlockTermState state = term.state;
           final int suffix = term.termBytes.length - prefixLength;
           /*
@@ -640,7 +640,7 @@ public final class OrdsBlockTreeTermsWriter extends FieldsConsumer {
           PendingEntry ent = pending.get(i);
           if (ent.isTerm) {
             PendingTerm term = (PendingTerm) ent;
-            assert StringHelper.startsWith(term.termBytes, prefix): "term.term=" + term.termBytes + " prefix=" + prefix;
+            assert StringHelper.startsWith(term.termBytes, prefix): term + " prefix=" + prefix;
             BlockTermState state = term.state;
             final int suffix = term.termBytes.length - prefixLength;
             /*
--- a/lucene/core/src/java/org/apache/lucene/analysis/Analyzer.java
+++ b/lucene/core/src/java/org/apache/lucene/analysis/Analyzer.java
@@ -94,7 +94,7 @@ public abstract class Analyzer implements Closeable {
    * Create a new Analyzer, reusing the same set of components per-thread
    * across calls to {@link #tokenStream(String, Reader)}. 
    */
-  public Analyzer() {
+  protected Analyzer() {
     this(GLOBAL_REUSE_STRATEGY);
   }
 
@@ -104,9 +104,9 @@ public abstract class Analyzer implements Closeable {
    * NOTE: if you just want to reuse on a per-field basis, it's easier to
    * use a subclass of {@link AnalyzerWrapper} such as 
    * <a href="{@docRoot}/../analysis/common/org/apache/lucene/analysis/miscellaneous/PerFieldAnalyzerWrapper.html">
-   * PerFieldAnalyerWrapper</a> instead.
+   * PerFieldAnalyzerWrapper</a> instead.
    */
-  public Analyzer(ReuseStrategy reuseStrategy) {
+  protected Analyzer(ReuseStrategy reuseStrategy) {
     this.reuseStrategy = reuseStrategy;
   }
 
@@ -440,9 +440,9 @@ public abstract class Analyzer implements Closeable {
    * {@link Analyzer#tokenStream(String, java.io.Reader)}.
    */
   public static abstract class ReuseStrategy {
-
     /** Sole constructor. (For invocation by subclass constructors, typically implicit.) */
-    public ReuseStrategy() {}
+    // Explicitly declared so that we have non-empty javadoc
+    protected ReuseStrategy() {}
 
     /**
      * Gets the reusable TokenStreamComponents for the field with the given name.
--- a/lucene/core/src/java/org/apache/lucene/codecs/blocktree/BlockTreeTermsWriter.java
+++ b/lucene/core/src/java/org/apache/lucene/codecs/blocktree/BlockTreeTermsWriter.java
@@ -709,7 +709,7 @@ public final class BlockTreeTermsWriter extends FieldsConsumer {
 
           PendingTerm term = (PendingTerm) ent;
 
-          assert StringHelper.startsWith(term.termBytes, prefix): "term.term=" + term.termBytes + " prefix=" + prefix;
+          assert StringHelper.startsWith(term.termBytes, prefix): term + " prefix=" + prefix;
           BlockTermState state = term.state;
           final int suffix = term.termBytes.length - prefixLength;
           //if (DEBUG2) {
@@ -741,7 +741,7 @@ public final class BlockTreeTermsWriter extends FieldsConsumer {
           if (ent.isTerm) {
             PendingTerm term = (PendingTerm) ent;
 
-            assert StringHelper.startsWith(term.termBytes, prefix): "term.term=" + term.termBytes + " prefix=" + prefix;
+            assert StringHelper.startsWith(term.termBytes, prefix): term + " prefix=" + prefix;
             BlockTermState state = term.state;
             final int suffix = term.termBytes.length - prefixLength;
             //if (DEBUG2) {
--- a/lucene/core/src/java/org/apache/lucene/index/DocValuesUpdate.java
+++ b/lucene/core/src/java/org/apache/lucene/index/DocValuesUpdate.java
@@ -45,7 +45,7 @@ abstract class DocValuesUpdate {
   final String field;
   // used in BufferedDeletes to apply this update only to a slice of docs. It's initialized to BufferedUpdates.MAX_INT
   // since it's safe and most often used this way we safe object creations.
-  final int docIDUpto;
+  final int docIDUpTo;
   final boolean hasValue;
 
   /**
@@ -54,19 +54,19 @@ abstract class DocValuesUpdate {
    * @param term the {@link Term} which determines the documents that will be updated
    * @param field the {@link NumericDocValuesField} to update
    */
-  protected DocValuesUpdate(DocValuesType type, Term term, String field, int docIDUpto, boolean hasValue) {
-    assert docIDUpto >= 0 : docIDUpto + "must be >= 0";
+  protected DocValuesUpdate(DocValuesType type, Term term, String field, int docIDUpTo, boolean hasValue) {
+    assert docIDUpTo >= 0 : docIDUpTo + "must be >= 0";
     this.type = type;
     this.term = term;
     this.field = field;
-    this.docIDUpto = docIDUpto;
+    this.docIDUpTo = docIDUpTo;
     this.hasValue = hasValue;
   }
 
   abstract long valueSizeInBytes();
   
-  final int sizeInBytes() {
-    int sizeInBytes = RAW_SIZE_IN_BYTES;
+  final long sizeInBytes() {
+    long sizeInBytes = RAW_SIZE_IN_BYTES;
     sizeInBytes += term.field.length() * Character.BYTES;
     sizeInBytes += term.bytes.bytes.length;
     sizeInBytes += field.length() * Character.BYTES;
@@ -85,7 +85,7 @@ abstract class DocValuesUpdate {
   
   @Override
   public String toString() {
-    return "term=" + term + ",field=" + field + ",value=" + valueToString() + ",docIDUpto=" + docIDUpto;
+    return "term=" + term + ",field=" + field + ",value=" + valueToString() + ",docIDUpTo=" + docIDUpTo;
   }
   
   /** An in-place update to a binary DocValues field */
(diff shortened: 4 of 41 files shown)


lucene/core/src/java/org/apache/lucene/analysis/Analyzer.java
```diff
@@ -94,7 +94,7 @@
    * Create a new Analyzer, reusing the same set of components per-thread
    * across calls to {@link #tokenStream(String, Reader)}. 
    */
-  public Analyzer() {
```
dweiss, line 97: Can you not change those scopes in public API classes? This applies here and in other places -- protected changed to package-scope for source is not really an API-compatible change.

lucene/core/src/java/org/apache/lucene/codecs/blocktree/BlockTreeTermsWriter.java
```diff
@@ -709,7 +709,7 @@ private PendingBlock writeBlock(int prefixLength, boolean isFloor, int floorLead
 
           PendingTerm term = (PendingTerm) ent;
 
-          assert StringHelper.startsWith(term.termBytes, prefix): "term.term=" + term.termBytes + " prefix=" + prefix;
+          assert StringHelper.startsWith(term.termBytes, prefix): "term.term=" + new String(term.termBytes) + " prefix=" + prefix;
```
dweiss, line 712: This is wrong, uses default locale.

lucene/core/src/java/org/apache/lucene/index/DocValuesUpdate.java
```diff
@@ -152,12 +152,12 @@ static BytesRef readFrom(DataInput in, BytesRef scratch) throws IOException {
     }
 
     NumericDocValuesUpdate(Term term, String field, Long value) {
-      this(term, field, value != null ? value.longValue() : -1, BufferedUpdates.MAX_INT, value != null);
+      this(term, field, value != null ? value : -1, BufferedUpdates.MAX_INT, value != null);
     }
 
 
-    private NumericDocValuesUpdate(Term term, String field, long value, int docIDUpTo, boolean hasValue) {
```
dweiss, line 159: previous version was correct camel case (upTo).

lucene/core/src/java/org/apache/lucene/index/DocumentsWriterFlushControl.java
```diff
@@ -324,12 +324,12 @@ synchronized void doOnAbort(DocumentsWriterPerThread perThread) {
     }
   }
 
-  private void checkoutAndBlock(DocumentsWriterPerThread perThread) {
+  private synchronized void checkoutAndBlock(DocumentsWriterPerThread perThread) {
```
dweiss, line 327: These are serious changes... you're adding synchronization on core classes. I don't think they should be piggybacked on top of trivial ones - I'm sure @s1monw would chip in whether this synchronization here makes sense but he'll probably overlook if it's a bulk of trivial changes on top.

lucene/core/src/java/org/apache/lucene/analysis/Analyzer.java
madrob: I understand that it's notionally an API change, but `abstract` classes have no reason for public constructors. We can make everything protected and the subclasses that people use will be able to pick it up. I was over-zealous in a couple places going to package instead of protected, I'll fix that up.

lucene/core/src/java/org/apache/lucene/codecs/blocktree/BlockTreeTermsWriter.java
madrob: Are these UTF-8? I wasn't sure, and hoped somebody would let me know during review.

lucene/core/src/java/org/apache/lucene/index/DocValuesUpdate.java
madrob: There were 16 instances of `Upto` and 4 of `UpTo` so I went with the more common one for consistency. Happy to switch the other way if it's more correct according to English. Looking it up now and looks like "upto" isn't a word?

lucene/core/src/java/org/apache/lucene/index/DocumentsWriterFlushControl.java
madrob: I'll split this out.

madrob: https://issues.apache.org/jira/browse/LUCENE-9453 I explain in that issue why I believe it is minor, but it will help to get more eyes on it

lucene/core/src/java/org/apache/lucene/analysis/Analyzer.java
dweiss: The constructors I understand, fine (although it's really a no-op change, as you indicated).

lucene/core/src/java/org/apache/lucene/codecs/blocktree/BlockTreeTermsWriter.java
dweiss: They should be. You can also dump it as a byte array for consistency with other changes you made.

lucene/core/src/java/org/apache/lucene/index/SegmentInfos.java
```diff
@@ -440,7 +440,7 @@ public static final SegmentInfos readCommit(Directory directory, ChecksumIndexIn
       if (format >= VERSION_70) { // oldest supported version
         CodecUtil.checkFooter(input, priorE);
       } else {
-        throw IOUtils.rethrowAlways(priorE);
```
dweiss, line 443: Leave this as it was (with throw ...) - don't know whether IntelliJ is smart enough to detect this method always throws an exception but other compilers are not (and this ensures the compiler sees it as a the only codepath leaving the clause).

lucene/core/src/java/org/apache/lucene/codecs/blocktree/BlockTreeTermsWriter.java
```diff
@@ -709,7 +710,8 @@ private PendingBlock writeBlock(int prefixLength, boolean isFloor, int floorLead
 
           PendingTerm term = (PendingTerm) ent;
 
-          assert StringHelper.startsWith(term.termBytes, prefix): "term.term=" + term.termBytes + " prefix=" + prefix;
```
dweiss, line 712: that 'term' class (PendingTerm) actually has a perfectly fine toString method... why not just remove termBytes and let it do its job?

lucene/core/src/java/org/apache/lucene/codecs/blocktree/BlockTreeTermsWriter.java
```diff
@@ -741,7 +743,8 @@ private PendingBlock writeBlock(int prefixLength, boolean isFloor, int floorLead
           if (ent.isTerm) {
             PendingTerm term = (PendingTerm) ent;
 
-            assert StringHelper.startsWith(term.termBytes, prefix): "term.term=" + term.termBytes + " prefix=" + prefix;
```
dweiss, line 744: Same here.

lucene/core/src/java/org/apache/lucene/analysis/Analyzer.java
```diff
@@ -367,12 +367,12 @@ public void close() {
     /**
      * Original source of the tokens.
      */
-    protected final Consumer<Reader> source;
```
dweiss, line 370: Don't change to package private scope here. It will prevent subclasses from outside of the package from accessing those fields (and there may be classes outside of Lucene code doing that).

lucene/core/src/java/org/apache/lucene/analysis/Analyzer.java
```diff
@@ -367,12 +367,12 @@ public void close() {
...
+    final Consumer<Reader> source;
     /**
      * Sink tokenstream, such as the outer tokenfilter decorating
      * the chain. This can be the source if there are no filters.
      */
-    protected final TokenStream sink;
```
dweiss, line 375: Same here.

lucene/codecs/src/java/org/apache/lucene/codecs/blocktreeords/OrdsBlockTreeTermsWriter.java
```diff
@@ -604,7 +605,7 @@ private PendingBlock writeBlock(int prefixLength, boolean isFloor, int floorLead
           assert ent.isTerm: "i=" + i;
 
           PendingTerm term = (PendingTerm) ent;
-          assert StringHelper.startsWith(term.termBytes, prefix): "term.term=" + term.termBytes + " prefix=" + prefix;
```
dweiss, line 607: term has a toString method - use it.

lucene/codecs/src/java/org/apache/lucene/codecs/blocktreeords/OrdsBlockTreeTermsWriter.java
```diff
@@ -640,7 +641,7 @@ private PendingBlock writeBlock(int prefixLength, boolean isFloor, int floorLead
           PendingEntry ent = pending.get(i);
           if (ent.isTerm) {
             PendingTerm term = (PendingTerm) ent;
-            assert StringHelper.startsWith(term.termBytes, prefix): "term.term=" + term.termBytes + " prefix=" + prefix;
```
dweiss, line 643: term has a toString method - use it.

lucene/core/src/java/org/apache/lucene/analysis/Analyzer.java
uschindler: Removing this would break many outside analyzers. I know it's seldom that analyzers access these fields, but this is a real backward breaking change. Don't do this.

I have no problem with the ctors, but here it's serious! Why did Intellij suggest a change like that? Looks like it was not so intelligent. 🤨

uschindler: There's also no risk somebody could do anything wrong. It's both (also next one) final field and the it's for consuming only.

madrob: I think it's because the field is final and there is a getter for it, so the code analyzer prefers encapsulation?

lucene/core/src/java/org/apache/lucene/index/SegmentInfos.java
madrob: The original compiler complaint was that the throw is inside the finally block. Could I replace the "Unreachable code" at the end with this rethrow? I believe the logic will be the same.

lucene/core/src/java/org/apache/lucene/analysis/Analyzer.java
dweiss: Maybe. It doesn't matter though - this changes the API of a class that's been there for ages. I bet there is a class out there somewhere (let's say A extends Analyzer) and another one (B extends A) where A overrides the getter but B reaches out for the original field. Do we want this to break just to hide a field that can be useful for subclasses just to silence an automatic code inspection? I don't think we should.

lucene/core/src/java/org/apache/lucene/index/SegmentInfos.java
dweiss: And what's wrong about a throw from within finally? A finally block is technically just a block of code, like any other. The compiler very likely assumes you're suppressing an exception if you throw from within finally but it's not the case here. 

I don't know if moving that throw will change the logic. Maybe not. Maybe yes. Given the two options, I wouldn't touch it. My concern was that you slipped such things as part of an otherwise "trivial" set of patches.

lucene/sandbox/src/java/org/apache/lucene/codecs/idversion/VersionBlockTreeTermsWriter.java
```diff
@@ -582,7 +583,7 @@ private PendingBlock writeBlock(int prefixLength, boolean isFloor, int floorLead
           assert ent.isTerm: "i=" + i;
 
           PendingTerm term = (PendingTerm) ent;
-          assert StringHelper.startsWith(term.termBytes, prefix): "term.term=" + term.termBytes + " prefix=" + prefix;
+          assert StringHelper.startsWith(term.termBytes, prefix): "term.term=" + Arrays.toString(term.termBytes) + " prefix=" + prefix;
```
dweiss, line 586: inconsistent with other calls (should use term.toString())?

lucene/sandbox/src/java/org/apache/lucene/codecs/idversion/VersionBlockTreeTermsWriter.java
```diff
@@ -610,7 +611,7 @@ private PendingBlock writeBlock(int prefixLength, boolean isFloor, int floorLead
           PendingEntry ent = pending.get(i);
           if (ent.isTerm) {
             PendingTerm term = (PendingTerm) ent;
-            assert StringHelper.startsWith(term.termBytes, prefix): "term.term=" + term.termBytes + " prefix=" + prefix;
+            assert StringHelper.startsWith(term.termBytes, prefix): "term.term=" + Arrays.toString(term.termBytes) + " prefix=" + prefix;
```
dweiss, line 614: term.toString() again?

lucene/test-framework/src/java/org/apache/lucene/index/BaseTermVectorsFormatTestCase.java
```diff
@@ -504,8 +504,8 @@ protected void assertEquals(RandomTokenStream tk, FieldType ft, Terms terms) thr
           }
           try {
```
dweiss, line 505: Since you're cleaning up I think it'd be better to use assertThrows() with lambda...

lucene/sandbox/src/java/org/apache/lucene/codecs/idversion/VersionBlockTreeTermsWriter.java
madrob: yikes, thank you.

lucene/test-framework/src/java/org/apache/lucene/index/BaseTermVectorsFormatTestCase.java
madrob: So it turns out be be less clean, since two of the implementations throw IllegalStateException and two throw AssertionError.

dweiss: @jpountz  may be interested in why this is inconsistent between implementations?

madrob:
@vthacker this is also some error-prone warnings, so may overlap with what you're looking at.

dweiss (approved):

PR #1734, opened 2020-08-10
LUCENE-9453 Add sync around volatile write

madrob:
checkoutAndBlock is not synchronized, but has a non-atomic write to numPending. Meanwhile, all of the other writes to numPending are in sync methods.

In this case it turns out to be ok because all of the code paths calling this method are already sync:

`synchronized doAfterDocument -> checkout -> checkoutAndBlock`
`checkoutLargestNonPendingWriter -> synchronized(this) -> checkout -> checkoutAndBlock`

If we make synchronized checkoutAndBlock that protects us against future changes, shouldn't cause any performance impact since the code paths will already be going through a sync block, and will make an IntelliJ warning go away.

Found via IntelliJ warnings. 

https://issues.apache.org/jira/browse/LUCENE-9453

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/DocumentsWriterFlushControl.java
+++ b/lucene/core/src/java/org/apache/lucene/index/DocumentsWriterFlushControl.java
@@ -324,12 +324,12 @@ final class DocumentsWriterFlushControl implements Accountable, Closeable {
     }
   }
 
-  private void checkoutAndBlock(DocumentsWriterPerThread perThread) {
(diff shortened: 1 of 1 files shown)


s1monw:
this lock is left out intentionally to reduce the overhead of acquiring a reentrant lock. We only call this method internally from a method that guarantees the lock is held. I don't think we should add this, for correctness it's not necessary at this point. Adding something like `assert Thread.holdsLock(this);` I'd be ok with.

PR #1736, opened 2020-08-11
Harden RequestRateLimiter Tests

atris:
This commit adds higher data size and load in the test path. Also improves
the asserts that are performed.

### DIFF
--- a/solr/core/src/test/org/apache/solr/servlet/TestRequestRateLimiter.java
+++ b/solr/core/src/test/org/apache/solr/servlet/TestRequestRateLimiter.java
@@ -73,7 +73,7 @@ public class TestRequestRateLimiter extends SolrCloudTestCase {
     assertEquals(350, mockQueryRateLimiter.incomingRequestCount.get());
 
     assertTrue((mockQueryRateLimiter.acceptedNewRequestCount.get() == mockQueryRateLimiter.incomingRequestCount.get()
-        || mockQueryRateLimiter.rejectedRequestCount.get() > 0));g
+        || mockQueryRateLimiter.rejectedRequestCount.get() > 0));
     assertEquals(mockQueryRateLimiter.incomingRequestCount.get(),
         mockQueryRateLimiter.acceptedNewRequestCount.get() + mockQueryRateLimiter.rejectedRequestCount.get());
   }
--- a/solr/core/src/test/org/apache/solr/servlet/TestRequestRateLimiter.java
+++ b/solr/core/src/test/org/apache/solr/servlet/TestRequestRateLimiter.java
@@ -65,12 +65,15 @@ public class TestRequestRateLimiter extends SolrCloudTestCase {
 
     solrDispatchFilter.replaceRateLimitManager(rateLimitManager);
 
-    processTest(client, 10000 /* number of documents */, 350 /* number of queries */);
+    int numDocs = TEST_NIGHTLY ? 10000 : 100;
+
+    processTest(client, numDocs, 350 /* number of queries */);
 
     MockRequestRateLimiter mockQueryRateLimiter = (MockRequestRateLimiter) rateLimitManager.getRequestRateLimiter(SolrRequest.SolrRequestType.QUERY);
 
     assertEquals(350, mockQueryRateLimiter.incomingRequestCount.get());
 
+    assertTrue(mockQueryRateLimiter.acceptedNewRequestCount.get() > 0);
     assertTrue((mockQueryRateLimiter.acceptedNewRequestCount.get() == mockQueryRateLimiter.incomingRequestCount.get()
         || mockQueryRateLimiter.rejectedRequestCount.get() > 0));
     assertEquals(mockQueryRateLimiter.incomingRequestCount.get(),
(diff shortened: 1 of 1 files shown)


solr/core/src/test/org/apache/solr/servlet/TestRequestRateLimiter.java
```diff
@@ -66,15 +66,11 @@ public void testConcurrentQueries() throws Exception {
 
     solrDispatchFilter.replaceRateLimitManager(rateLimitManager);
 
-    processTest(client);
+    processTest(client, 10000 /* number of documents */, 350 /* number of queries */);
```
madrob, line 69: Can we limit the higher footprint to Nightly?

solr/core/src/test/org/apache/solr/servlet/TestRequestRateLimiter.java
```diff
@@ -66,15 +66,11 @@ public void testConcurrentQueries() throws Exception {
...
 
     MockRequestRateLimiter mockQueryRateLimiter = (MockRequestRateLimiter) rateLimitManager.getRequestRateLimiter(SolrRequest.SolrRequestType.QUERY);
 
-    assertEquals(25, mockQueryRateLimiter.incomingRequestCount.get());
-    assertTrue("Incoming accepted new request count did not match. Expected 5 incoming " + mockQueryRateLimiter.acceptedNewRequestCount.get(),
```
madrob, line 74: Somewhat concerning that the fix to the test is to relax the assertion conditions

solr/core/src/test/org/apache/solr/servlet/TestRequestRateLimiter.java
```diff
@@ -124,12 +120,12 @@ private void processTest(CloudSolrClient client) throws Exception {
...
 
     try {
-      for (int i = 0; i < 25; i++) {
+      for (int i = 0; i < numQueries; i++) {
         callableList.add(() -> {
           try {
             QueryResponse response = client.query(new SolrQuery("*:*"));
 
-            assertEquals(100, response.getResults().getNumFound());
+            assertEquals(10000, response.getResults().getNumFound());
```
madrob, line 128: should be numDocuments

solr/core/src/test/org/apache/solr/servlet/TestRequestRateLimiter.java
atris: It isnt really a relaxation -- the remaining assert should cover for all cases that can happen for rate limiting. The catch is that rate limiting is not a guaranteed phenomenon -- we create a high load and it should happen. I have added an additional assert -- let me know if it looks fine.

solr/core/src/test/org/apache/solr/servlet/TestRequestRateLimiter.java
```diff
@@ -48,7 +47,7 @@ public static void setupCluster() throws Exception {
     configureCluster(1).addConfig(FIRST_COLLECTION, configset("cloud-minimal")).configure();
   }
 
-  @Test
+  @Nightly
```
madrob, line 50: This isn't what I meant, sorry for being unclear. Keep this as `@Test` but when selecting the number of documents and queries do something like https://github.com/apache/lucene-solr/blob/master/lucene/core/src/test/org/apache/lucene/index/TestMultiDocValues.java#L52

solr/core/src/test/org/apache/solr/servlet/TestRequestRateLimiter.java
```diff
@@ -66,20 +65,19 @@ public void testConcurrentQueries() throws Exception {
...
     MockRequestRateLimiter mockQueryRateLimiter = (MockRequestRateLimiter) rateLimitManager.getRequestRateLimiter(SolrRequest.SolrRequestType.QUERY);
 
-    assertEquals(25, mockQueryRateLimiter.incomingRequestCount.get());
-    assertTrue("Incoming accepted new request count did not match. Expected 5 incoming " + mockQueryRateLimiter.acceptedNewRequestCount.get(),
-        mockQueryRateLimiter.acceptedNewRequestCount.get() < 25);
-    assertTrue("Incoming rejected new request count did not match. Expected 20 incoming " + mockQueryRateLimiter.rejectedRequestCount.get(),
-        mockQueryRateLimiter.rejectedRequestCount.get() > 0);
+    assertEquals(350, mockQueryRateLimiter.incomingRequestCount.get());
+
+    assertTrue((mockQueryRateLimiter.acceptedNewRequestCount.get() == mockQueryRateLimiter.incomingRequestCount.get()
```
madrob, line 74: Should we assert that accepted + rejected = total?
And that accepted > 0.

atris: That is what we do in this assert?
    assertEquals(mockQueryRateLimiter.incomingRequestCount.get(),
        mockQueryRateLimiter.acceptedNewRequestCount.get() + mockQueryRateLimiter.rejectedRequestCount.get());

solr/core/src/test/org/apache/solr/servlet/TestRequestRateLimiter.java
madrob: s/Nightly/Test

solr/core/src/test/org/apache/solr/servlet/TestRequestRateLimiter.java
madrob: yea I was looking at the wrong side of the diff

solr/core/src/test/org/apache/solr/servlet/TestRequestRateLimiter.java
atris: Updated

PR #1737, opened 2020-08-11
SOLR-14615: Implement CPU Utilization Based Circuit Breaker

atris:
This commit introduces CPU based circuit breaker. This circuit breaker
tracks the average CPU load per minute and triggers if the value exceeds
a configurable value.

This commit also adds a specific control flag for Memory Circuit Breaker
to allow enabling/disabling the same.

### DIFF
--- a/solr/core/src/java/org/apache/solr/core/SolrConfig.java
+++ b/solr/core/src/java/org/apache/solr/core/SolrConfig.java
@@ -233,7 +233,7 @@ public class SolrConfig extends XmlConfigFile implements MapSerializable {
     isMemoryCircuitBreakerEnabled = getBool("circuitBreaker/isMemoryCircuitBreakerEnabled", false);
     memoryCircuitBreakerThresholdPct = getInt("circuitBreaker/memoryCircuitBreakerThresholdPct", 95);
 
-    cpuCircuitBreakerThresholdPct = getInt("circuitBreaker/cpuCircuitBreakerThresholdPct", 95);
+    cpuCircuitBreakerThreshold = getInt("circuitBreaker/cpuCircuitBreakerThreshold", 95);
 
     validateCircuitBreakerThresholds();
     
@@ -537,7 +537,7 @@ public class SolrConfig extends XmlConfigFile implements MapSerializable {
   public final int memoryCircuitBreakerThresholdPct;
   public final boolean isMemoryCircuitBreakerEnabled;
   public final boolean isCpuCircuitBreakerEnabled;
-  public final int cpuCircuitBreakerThresholdPct;
+  public final int cpuCircuitBreakerThreshold;
 
   // IndexConfig settings
   public final SolrIndexConfig indexConfig;
@@ -825,12 +825,6 @@ public class SolrConfig extends XmlConfigFile implements MapSerializable {
           throw new IllegalArgumentException("Valid value range of memoryCircuitBreakerThresholdPct is 50 -  95");
         }
       }
-
-      if (isCpuCircuitBreakerEnabled) {
-        if (cpuCircuitBreakerThresholdPct > 95 || cpuCircuitBreakerThresholdPct < 40) {
-          throw new IllegalArgumentException("Valid value range for cpuCircuitBreakerThresholdPct is 40 - 95");
-        }
-      }
     }
   }
 
@@ -907,7 +901,7 @@ public class SolrConfig extends XmlConfigFile implements MapSerializable {
     m.put("isCpuCircuitBreakerEnabled", isCpuCircuitBreakerEnabled);
     m.put("isMemoryCircuitBreakerEnabled", isMemoryCircuitBreakerEnabled);
     m.put("memoryCircuitBreakerThresholdPct", memoryCircuitBreakerThresholdPct);
-    m.put("cpuCircuitBreakerThresholdPct", cpuCircuitBreakerThresholdPct);
+    m.put("cpuCircuitBreakerThreshold", cpuCircuitBreakerThreshold);
     for (SolrPluginInfo plugin : plugins) {
       List<PluginInfo> infos = getPluginInfos(plugin.clazz.getName());
       if (infos == null || infos.isEmpty()) continue;
--- a/solr/core/src/java/org/apache/solr/core/SolrConfig.java
+++ b/solr/core/src/java/org/apache/solr/core/SolrConfig.java
@@ -229,8 +229,8 @@ public class SolrConfig extends XmlConfigFile implements MapSerializable {
     enableLazyFieldLoading = getBool("query/enableLazyFieldLoading", false);
 
     useCircuitBreakers = getBool("circuitBreaker/useCircuitBreakers", false);
-    isCpuCircuitBreakerEnabled = getBool("circuitBreaker/isCpuCircuitBreakerEnabled", false);
-    isMemoryCircuitBreakerEnabled = getBool("circuitBreaker/isMemoryCircuitBreakerEnabled", false);
+    cpuCircuitBreakerEnabled = getBool("circuitBreaker/cpuCircuitBreakerEnabled", false);
+    memoryCircuitBreakerEnabled = getBool("circuitBreaker/memoryCircuitBreakerEnabled", false);
     memoryCircuitBreakerThresholdPct = getInt("circuitBreaker/memoryCircuitBreakerThresholdPct", 95);
 
     cpuCircuitBreakerThreshold = getInt("circuitBreaker/cpuCircuitBreakerThreshold", 95);
@@ -535,8 +535,8 @@ public class SolrConfig extends XmlConfigFile implements MapSerializable {
   // Circuit Breaker Configuration
   public final boolean useCircuitBreakers;
   public final int memoryCircuitBreakerThresholdPct;
-  public final boolean isMemoryCircuitBreakerEnabled;
-  public final boolean isCpuCircuitBreakerEnabled;
+  public final boolean memoryCircuitBreakerEnabled;
+  public final boolean cpuCircuitBreakerEnabled;
   public final int cpuCircuitBreakerThreshold;
 
   // IndexConfig settings
@@ -820,7 +820,7 @@ public class SolrConfig extends XmlConfigFile implements MapSerializable {
 
   private void validateCircuitBreakerThresholds() {
     if (useCircuitBreakers) {
-      if (isMemoryCircuitBreakerEnabled) {
+      if (memoryCircuitBreakerEnabled) {
         if (memoryCircuitBreakerThresholdPct > 95 || memoryCircuitBreakerThresholdPct < 50) {
           throw new IllegalArgumentException("Valid value range of memoryCircuitBreakerThresholdPct is 50 -  95");
         }
@@ -898,8 +898,8 @@ public class SolrConfig extends XmlConfigFile implements MapSerializable {
     m.put("enableLazyFieldLoading", enableLazyFieldLoading);
     m.put("maxBooleanClauses", booleanQueryMaxClauseCount);
     m.put("useCircuitBreakers", useCircuitBreakers);
-    m.put("isCpuCircuitBreakerEnabled", isCpuCircuitBreakerEnabled);
-    m.put("isMemoryCircuitBreakerEnabled", isMemoryCircuitBreakerEnabled);
+    m.put("cpuCircuitBreakerEnabled", cpuCircuitBreakerEnabled);
+    m.put("memoryCircuitBreakerEnabled", memoryCircuitBreakerEnabled);
     m.put("memoryCircuitBreakerThresholdPct", memoryCircuitBreakerThresholdPct);
     m.put("cpuCircuitBreakerThreshold", cpuCircuitBreakerThreshold);
     for (SolrPluginInfo plugin : plugins) {
--- a/solr/core/src/java/org/apache/solr/core/SolrConfig.java
+++ b/solr/core/src/java/org/apache/solr/core/SolrConfig.java
@@ -228,12 +228,11 @@ public class SolrConfig extends XmlConfigFile implements MapSerializable {
     queryResultMaxDocsCached = getInt("query/queryResultMaxDocsCached", Integer.MAX_VALUE);
     enableLazyFieldLoading = getBool("query/enableLazyFieldLoading", false);
 
-    useCircuitBreakers = getBool("circuitBreaker/useCircuitBreakers", false);
-    cpuCircuitBreakerEnabled = getBool("circuitBreaker/cpuCircuitBreakerEnabled", false);
-    memoryCircuitBreakerEnabled = getBool("circuitBreaker/memoryCircuitBreakerEnabled", false);
-    memoryCircuitBreakerThresholdPct = getInt("circuitBreaker/memoryCircuitBreakerThresholdPct", 95);
-
-    cpuCircuitBreakerThreshold = getInt("circuitBreaker/cpuCircuitBreakerThreshold", 95);
+    useCircuitBreakers = getBool("circuitBreakers/@enabled", false);
+    cpuCBEnabled = getBool("circuitBreakers/cpuBreaker/@enabled", false);
+    memCBEnabled = getBool("circuitBreakers/memBreaker/@enabled", false);
+    memCBThreshold = getInt("circuitBreakers/memBreaker/@threshold", 95);
+    cpuCBThreshold = getInt("circuitBreakers/cpuBreaker/@threshold", 95);
 
     validateCircuitBreakerThresholds();
     
@@ -534,10 +533,10 @@ public class SolrConfig extends XmlConfigFile implements MapSerializable {
 
   // Circuit Breaker Configuration
   public final boolean useCircuitBreakers;
-  public final int memoryCircuitBreakerThresholdPct;
-  public final boolean memoryCircuitBreakerEnabled;
-  public final boolean cpuCircuitBreakerEnabled;
-  public final int cpuCircuitBreakerThreshold;
+  public final int memCBThreshold;
+  public final boolean memCBEnabled;
+  public final boolean cpuCBEnabled;
+  public final int cpuCBThreshold;
 
   // IndexConfig settings
   public final SolrIndexConfig indexConfig;
@@ -820,8 +819,8 @@ public class SolrConfig extends XmlConfigFile implements MapSerializable {
 
   private void validateCircuitBreakerThresholds() {
     if (useCircuitBreakers) {
-      if (memoryCircuitBreakerEnabled) {
-        if (memoryCircuitBreakerThresholdPct > 95 || memoryCircuitBreakerThresholdPct < 50) {
+      if (memCBEnabled) {
+        if (memCBThreshold > 95 || memCBThreshold < 50) {
           throw new IllegalArgumentException("Valid value range of memoryCircuitBreakerThresholdPct is 50 -  95");
         }
       }
@@ -898,10 +897,10 @@ public class SolrConfig extends XmlConfigFile implements MapSerializable {
     m.put("enableLazyFieldLoading", enableLazyFieldLoading);
     m.put("maxBooleanClauses", booleanQueryMaxClauseCount);
     m.put("useCircuitBreakers", useCircuitBreakers);
-    m.put("cpuCircuitBreakerEnabled", cpuCircuitBreakerEnabled);
-    m.put("memoryCircuitBreakerEnabled", memoryCircuitBreakerEnabled);
-    m.put("memoryCircuitBreakerThresholdPct", memoryCircuitBreakerThresholdPct);
-    m.put("cpuCircuitBreakerThreshold", cpuCircuitBreakerThreshold);
+    m.put("cpuCircuitBreakerEnabled", cpuCBEnabled);
+    m.put("memoryCircuitBreakerEnabled", memCBEnabled);
+    m.put("memoryCircuitBreakerThresholdPct", memCBThreshold);
+    m.put("cpuCircuitBreakerThreshold", cpuCBThreshold);
     for (SolrPluginInfo plugin : plugins) {
       List<PluginInfo> infos = getPluginInfos(plugin.clazz.getName());
       if (infos == null || infos.isEmpty()) continue;
--- a/solr/core/src/java/org/apache/solr/core/SolrConfig.java
+++ b/solr/core/src/java/org/apache/solr/core/SolrConfig.java
@@ -538,8 +538,6 @@ public class SolrConfig extends XmlConfigFile implements MapSerializable {
   public final boolean isMemoryCircuitBreakerEnabled;
   public final boolean isCpuCircuitBreakerEnabled;
   public final int cpuCircuitBreakerThresholdPct;
-  
-  public final boolean useRangeVersionsForPeerSync;
 
   // IndexConfig settings
   public final SolrIndexConfig indexConfig;
--- a/solr/core/src/java/org/apache/solr/core/SolrConfig.java
+++ b/solr/core/src/java/org/apache/solr/core/SolrConfig.java
@@ -229,9 +229,13 @@ public class SolrConfig extends XmlConfigFile implements MapSerializable {
     enableLazyFieldLoading = getBool("query/enableLazyFieldLoading", false);
 
     useCircuitBreakers = getBool("circuitBreaker/useCircuitBreakers", false);
+    isCpuCircuitBreakerEnabled = getBool("circuitBreaker/isCpuCircuitBreakerEnabled", false);
+    isMemoryCircuitBreakerEnabled = getBool("circuitBreaker/isMemoryCircuitBreakerEnabled", false);
     memoryCircuitBreakerThresholdPct = getInt("circuitBreaker/memoryCircuitBreakerThresholdPct", 95);
 
-    validateMemoryBreakerThreshold();
+    cpuCircuitBreakerThresholdPct = getInt("circuitBreaker/cpuCircuitBreakerThresholdPct", 95);
+
+    validateCircuitBreakerThresholds();
     
     filterCacheConfig = CacheConfig.getConfig(this, "query/filterCache");
     queryResultCacheConfig = CacheConfig.getConfig(this, "query/queryResultCache");
@@ -531,6 +535,11 @@ public class SolrConfig extends XmlConfigFile implements MapSerializable {
   // Circuit Breaker Configuration
   public final boolean useCircuitBreakers;
   public final int memoryCircuitBreakerThresholdPct;
+  public final boolean isMemoryCircuitBreakerEnabled;
+  public final boolean isCpuCircuitBreakerEnabled;
+  public final int cpuCircuitBreakerThresholdPct;
+  
+  public final boolean useRangeVersionsForPeerSync;
 
   // IndexConfig settings
   public final SolrIndexConfig indexConfig;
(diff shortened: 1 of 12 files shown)


noblepaul:
The configuration can be as simple as

`<circuitreaker property="value" property2="value">`


This way you can just read all the attributes all at once from the `PluginInfo`  .

atris:
> The configuration can be as simple as
> 
> `<circuitBreaker memoryCircuitBreakerThresholdPct="" isCpuCircuitBreakerEnabled="true" memoryCircuitBreakerThresholdPct ="" cpuCircuitBreakerThresholdPct="">`
> 
> This way you can just read all the attributes all at once from the `PluginInfo` .
> CircuitBreaker should be a type of plugin. It should be an interface

As discussed offline, I will refactor circuit breaker infrastructure to use PluginInfo as a part of 8.7 (hence will leave this PR's JIRA open for that effort). No proceeding with that effort in this PR.

solr/core/src/java/org/apache/solr/util/circuitbreaker/CPUCircuitBreaker.java
```diff
@@ -0,0 +1,108 @@
...
+ */
+public class CPUCircuitBreaker extends CircuitBreaker {
+  private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+  private static final OperatingSystemMXBean operatingSystemMXBean = ManagementFactory.getOperatingSystemMXBean();
+
+  private final boolean isCpuCircuitBreakerEnabled;
+  private final double cpuUsageThreshold;
+
+  // Assumption -- the value of these parameters will be set correctly before invoking getDebugInfo()
+  private final ThreadLocal<Double> seenCPUUsage = new ThreadLocal<>();
```
madrob, line 52: thread locals should be static

solr/core/src/java/org/apache/solr/core/SolrConfig.java
```diff
@@ -811,10 +818,18 @@ private void initLibs(SolrResourceLoader loader, boolean isConfigsetTrusted) {
...
-        throw new IllegalArgumentException("Valid value range of memoryCircuitBreakerThresholdPct is 50 -  95");
+      if (isMemoryCircuitBreakerEnabled) {
+        if (memoryCircuitBreakerThresholdPct > 95 || memoryCircuitBreakerThresholdPct < 50) {
+          throw new IllegalArgumentException("Valid value range of memoryCircuitBreakerThresholdPct is 50 -  95");
+        }
+      }
+
+      if (isCpuCircuitBreakerEnabled) {
+        if (cpuCircuitBreakerThresholdPct > 95 || cpuCircuitBreakerThresholdPct < 40) {
+          throw new IllegalArgumentException("Valid value range for cpuCircuitBreakerThresholdPct is 40 - 95");
```
madrob, line 831: I don't think CPU load average is typically measured on a 0-100 scale. Can you confirm some sample values of what calculateLiveCPUUsage returns?

solr/core/src/java/org/apache/solr/util/circuitbreaker/CPUCircuitBreaker.java
```diff
@@ -0,0 +1,108 @@
...
+    double localAllowedCPUUsage = getCpuUsageThreshold();
+    double localSeenCPUUsage = calculateLiveCPUUsage();
+
+    if (localSeenCPUUsage < 0) {
+      if (log.isWarnEnabled()) {
+        String msg = "Unable to get CPU usage";
+
+        log.warn(msg);
+
+        return false;
```
madrob, line 81: return false if only if warn enabled seems odd.

atris: Good catch, thanks.

solr/core/src/java/org/apache/solr/core/SolrConfig.java
atris: I see values between 0 - 100. Ran stress locally and validated values.

atris: Scratch that, this is a value that can exceed 100 as well. Removed the top ceiling on the value here.

MarcusSorealheis: It typically doesn't go past 450 unless you have an issue. I would still keep it bounded at some measure @atris

atris:
@madrob Any further thoughts on this?

solr/solr-ref-guide/src/circuit-breakers.adoc
```diff
@@ -35,33 +35,67 @@ will be disabled globally. Per circuit breaker configurations are specified in t
...
+
+Note that this configuration will be overridden by the global circuit breaker flag -- if circuit breakers are disabled, this flag
+will not help you.
+
+To set the triggering threshold in units of CPU utilization, use the following parameter.
+[source,xml]
+----
+<cpuCircuitBreakerThreshold>75</cpuCircuitBreakerThreshold>
+----
+The range of valid values for this parameter is [40 to infinity].
```
madrob, line 92: Why is 40 a good lower bound?

solr/solr-ref-guide/src/circuit-breakers.adoc
```diff
@@ -35,33 +35,67 @@ will be disabled globally. Per circuit breaker configurations are specified in t
...
+
+Note that both these circuit breakers are checked for each incoming search request and consider the current heap usage of the node and average CPU usage over the last minute, respectively.
+The checks does not impact performance, but any performance regressions that are suspected to be caused by this feature should be reported to the dev list.
 
 == Performance Considerations
-It is worth noting that while JVM circuit breaker does not add any noticeable overhead per query, having too many
+It is worth noting that while JVM or CPU circuit breakers does not add any noticeable overhead per query, having too many
```
madrob, line 98: s/does not/do not

solr/solr-ref-guide/src/circuit-breakers.adoc
```diff
@@ -35,33 +35,67 @@ will be disabled globally. Per circuit breaker configurations are specified in t
...
+will not help you.
+
+To set the triggering threshold in units of CPU utilization, use the following parameter.
+[source,xml]
+----
+<cpuCircuitBreakerThreshold>75</cpuCircuitBreakerThreshold>
+----
+The range of valid values for this parameter is [40 to infinity].
+
+Note that both these circuit breakers are checked for each incoming search request and consider the current heap usage of the node and average CPU usage over the last minute, respectively.
```
madrob, line 94: Be more explicit about the CPU check using the "1 min load avg"

Since this is a relatively complicated metric (compared to straight memory usage) let's also provide a link - "For more information on how load avg is calculated refer to https://en.wikipedia.org/wiki/Load_(computing)"

Is this the same on Windows? Can we enlist one of our Windows install havers to check this out? @dweiss ?

solr/core/src/test/org/apache/solr/util/TestCircuitBreaker.java
```diff
@@ -179,7 +236,13 @@ public void testResponseWithCBTiming() {
...
   }
 
-  private class MockCircuitBreaker extends CircuitBreaker {
+  private void removeAllExistingCircuitBreakers() {
+    List<CircuitBreaker> registeredCircuitBreakers = h.getCore().getCircuitBreakerManager().getRegisteredCircuitBreakers();
+
+    registeredCircuitBreakers.clear();
+  }
+
+  private class MockCircuitBreaker extends MemoryCircuitBreaker {
```
madrob, line 245: static class

solr/core/src/test/org/apache/solr/util/TestCircuitBreaker.java
```diff
@@ -157,6 +163,57 @@ public void testBuildingMemoryPressure() {
...
+            throw new RuntimeException(e.getMessage());
+          }
+        });
+      }
+
+      executor.shutdown();
+      try {
+        executor.awaitTermination(Long.MAX_VALUE, TimeUnit.NANOSECONDS);
+      } catch (InterruptedException e) {
+        throw new RuntimeException(e.getMessage());
```
madrob, line 206: thread.interrupt

solr/core/src/test/org/apache/solr/util/TestCircuitBreaker.java
```diff
@@ -157,6 +163,57 @@ public void testBuildingMemoryPressure() {
...
+          try {
+            h.query(req("name:\"john smith\""));
+          } catch (SolrException e) {
+            if (!e.getMessage().startsWith("Circuit Breakers tripped")) {
+              if (log.isInfoEnabled()) {
+                String logMessage = "Expected error message for testFakeCPUCircuitBreaker was not received. Error message " + e.getMessage();
+                log.info(logMessage);
+              }
+              throw new RuntimeException("Expected error message was not received. Error message " + e.getMessage());
+            }
```
madrob, line 194: Can we simplify this whole block to `assertThat(e.getMessage(), containsString("Circuit Breakers tripped"))`

solr/core/src/test/org/apache/solr/util/TestCircuitBreaker.java
```diff
@@ -157,6 +163,57 @@ public void testBuildingMemoryPressure() {
...
+
+      CircuitBreaker circuitBreaker = new FakeCPUCircuitBreaker(h.getCore().getSolrConfig());
+
+      h.getCore().getCircuitBreakerManager().register(circuitBreaker);
+
+      for (int i = 0; i < 5; i++) {
+        executor.submit(() -> {
+          try {
+            h.query(req("name:\"john smith\""));
+          } catch (SolrException e) {
```
madrob, line 187: Does circuit breaker throw a more specific exception? It's kind of cumbersome from a client API perspective that they need to inspect the string message for knowing that they should retry from circuit breakers.

solr/core/src/test/org/apache/solr/util/TestCircuitBreaker.java
```diff
@@ -157,6 +163,57 @@ public void testBuildingMemoryPressure() {
...
+
+    try {
+      removeAllExistingCircuitBreakers();
+
+      CircuitBreaker circuitBreaker = new FakeCPUCircuitBreaker(h.getCore().getSolrConfig());
+
+      h.getCore().getCircuitBreakerManager().register(circuitBreaker);
+
+      for (int i = 0; i < 5; i++) {
+        executor.submit(() -> {
```
madrob, line 184: save this `Future` and assert that it did not throw an exception

solr/solr-ref-guide/src/circuit-breakers.adoc
atris: Removed.

solr/core/src/test/org/apache/solr/util/TestCircuitBreaker.java
atris: Shouldnt return code 500 help there?

solr/solr-ref-guide/src/circuit-breakers.adoc
atris: Updated the documentation, thanks.

solr/core/src/java/org/apache/solr/util/circuitbreaker/CPUCircuitBreaker.java
```diff
@@ -0,0 +1,108 @@
...
+
+/**
+ * <p>
+ * Tracks current CPU usage and triggers if the specified threshold is breached.
+ *
+ * This circuit breaker gets the average CPU load over the last minute and uses
+ * that data to take a decision. Ideally, we should be able to cache the value
+ * locally and only query once the minute has elapsed. However, that will introduce
+ * more complexity than the current structure and might not get us major performance
+ * wins. If this ever becomes a performance bottleneck, that can be considered.
```
sigram, line 36: Javadoc for OSMxBean says "This method is designed to provide a hint about the system load and may be queried frequently." Not sure what "frequently" means here, though...

It will be interesting to see the dynamic behavior of this breaker - I'm somewhat worried that the 1-min average may be too unstable and we end up flip-flopping between the states too often (eg. every dozen requests or so). Depending on the volume of momentary load (eg. an ongoing large merge or split shard) the 1-min average may exceed a threshold even though it doesn't properly reflect a sustained longer-term overload that we should worry about. Average load is a convenient lie ;)

OTOH that's the only method we have in the OS MXBean, so we would have to compute that longer-term average ourselves, which is messy... so I guess we'll see .

solr/core/src/java/org/apache/solr/util/circuitbreaker/CPUCircuitBreaker.java
```diff
@@ -0,0 +1,108 @@
...
+    allowedCPUUsage.set(localAllowedCPUUsage);
+
+    seenCPUUsage.set(localSeenCPUUsage);
+
+    return (localSeenCPUUsage >= localAllowedCPUUsage);
+  }
+
+  @Override
+  public String getDebugInfo() {
+    if (seenCPUUsage.get() == 0L || allowedCPUUsage.get() == 0L) {
```
sigram, line 94: I think that on `ThreadLocal.get()` you would get a null value and consequently an NPE. You would have to use `ThreadLocal.withInitial(supplier)` to get a `Double(0)` here.

solr/core/src/java/org/apache/solr/util/circuitbreaker/CPUCircuitBreaker.java
sigram: Also, caching doesn't have to be complicated .. you simply check the elapsed time since last request and if it's longer than timeout you refresh the value first, it's probably less than 10 lines of code.

solr/core/src/java/org/apache/solr/util/circuitbreaker/MemoryCircuitBreaker.java
```diff
@@ -43,6 +43,7 @@
   private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
   private static final MemoryMXBean MEMORY_MX_BEAN = ManagementFactory.getMemoryMXBean();
 
+  private boolean isMemoryCircuitBreakerEnabled;
```
sigram, line 46: We know it's a boolean and it's in the MemoryCircuitBreaker, why not simply call it `enabled` (like many other Solr plugins do)?

solr/core/src/java/org/apache/solr/core/SolrConfig.java
```diff
@@ -229,9 +229,13 @@ private SolrConfig(SolrResourceLoader loader, String name, boolean isConfigsetTr
     enableLazyFieldLoading = getBool("query/enableLazyFieldLoading", false);
 
     useCircuitBreakers = getBool("circuitBreaker/useCircuitBreakers", false);
+    isCpuCircuitBreakerEnabled = getBool("circuitBreaker/isCpuCircuitBreakerEnabled", false);
```
sigram, line 232: I'm not sure if the boolean flags should always contain `is`, also I generally hate too long names... ;) we already know this is a section for circuit breakers, so the name doesn't have to repeat all of it. How about `cpuBreakerEnabled`, `memoryBreakerEnabled` etc?

solr/core/src/test/org/apache/solr/util/TestCircuitBreaker.java
sigram: Currently the `CircuitBreakerManager.toErrorMessage` uses individual breaker's `getDebugInfo` to construct the error message. I think we should improve these individual messages to clearly say why a break was tripped. Currently it's not obvious - eg. when a mem breaker trips it just reports the current and total memory but not the thresholdPct.

Maybe we need a different method like `CircuitBreaker.getErrorMessage` ?

solr/solr-ref-guide/src/circuit-breakers.adoc
sigram: This is still not removed in the latest rev?

Also, we can be more helpful here: maybe add something like "typically 1-min average CPU load may often reach values of 10s for a busy machine, but values consistently above 20 and higher indicate progressively serious contention." - or something like that.

solr/core/src/java/org/apache/solr/util/circuitbreaker/CPUCircuitBreaker.java
madrob: Yea, I think this is a very good point. If load average only updates every 5 seconds, that seems like a huge interval where we're going to block query execution. Let's think about how we can get better granularity.

solr/core/src/java/org/apache/solr/util/circuitbreaker/MemoryCircuitBreaker.java
madrob: Agree with this. `is*` prefix typically makes more sense on the methods not fields.

solr/core/src/test/org/apache/solr/util/TestCircuitBreaker.java
madrob: return code 500 doesn't matter if the SolrJ client converts it into an exception every single time.

solr/core/src/java/org/apache/solr/util/circuitbreaker/CPUCircuitBreaker.java
atris: The Javadocs state that the average load is updated per one minute -- so I am not sure how we can get a granularity better than that. Its unfortunate that this unit is not configurable (and I believe it must be internally caching as well since it talks about being prepared for "querying frequently". Any thoughts on this?

atris: Thinking more about this, the load is an average of one minute. If there is a transient spike -- should it not be smoothened out over the minute so that the average value is reasonable?

solr/core/src/test/org/apache/solr/util/TestCircuitBreaker.java
atris: https://issues.apache.org/jira/browse/SOLR-14755

solr/core/src/java/org/apache/solr/util/circuitbreaker/CPUCircuitBreaker.java
```diff
@@ -91,11 +91,18 @@ public boolean isTripped() {
 
   @Override
   public String getDebugInfo() {
-    if (seenCPUUsage.get() == 0L || allowedCPUUsage.get() == 0L) {
+
+    if (seenCPUUsage.withInitial(supplier).get() == 0.0 || seenCPUUsage.withInitial(supplier).get() == 0.0) {
```
madrob, line 95: withInitial is supposed to be on the declaration, not on each use

solr/core/src/java/org/apache/solr/util/circuitbreaker/CPUCircuitBreaker.java
sigram: Uh ... I see there's still some misunderstanding about this. The call itself is directly passed to the native method that invokes stdlib `getloadavg`, which in turn reads these values from the /proc pseudo-fs. So, the cost is truly minimal and the call doesn't block - if it turns out that it's still too costly to call for every request then we can introduce some timeout-based caching.

These averages are so called exponentially weighted moving averages, so indeed a 1-min average has traces of past load values from outside the 1-min window, which helps in smoothing it. This may turn out to be sufficient to avoid false positives due to short-term spikes (such as large merges). Linux loadavg represents to some degree a combined CPU + disk IO load, so indeed intensive IO operations will affect it.

We always have an option to use Codahale Meter to easily calculate 5- and 15-min EWMAs if it turns out that we're getting too many false positives. Until then users can configure higher thresholds, thus reducing the number of false positives at the cost of higher contention.

solr/core/src/java/org/apache/solr/util/circuitbreaker/CPUCircuitBreaker.java
```diff
@@ -0,0 +1,116 @@
...
+    if (seenCPUUsage.get() == 0.0 || seenCPUUsage.get() == 0.0) {
+      log.warn("CPUCircuitBreaker's monitored values (seenCPUUSage, allowedCPUUsage) not set");
+    }
+
+    return "seenCPUUSage=" + seenCPUUsage.get() + " allowedCPUUsage=" + allowedCPUUsage.get();
+  }
+
+  @Override
+  public String getErrorMessage() {
+    return "CPU Circuit Breaker Triggered. Seen CPU usage " + seenCPUUsage.get() + " and allocated threshold " +
```
sigram, line 105: Maybe "greater than allocated threshold" ?

solr/core/src/java/org/apache/solr/util/circuitbreaker/MemoryCircuitBreaker.java
```diff
@@ -88,13 +95,19 @@ public boolean isTripped() {
...
+    if (seenMemory.get() == 0.0 || allowedMemory.get() == 0.0) {
       log.warn("MemoryCircuitBreaker's monitored values (seenMemory, allowedMemory) not set");
     }
 
     return "seenMemory=" + seenMemory.get() + " allowedMemory=" + allowedMemory.get();
   }
 
+  @Override
+  public String getErrorMessage() {
+    return "Memory Circuit Breaker Triggered. Seen JVM heap memory usage " + seenMemory.get() + " and allocated threshold " +
```
sigram, line 107: Similarly, "greater than allocated threshold" ?

solr/core/src/java/org/apache/solr/util/circuitbreaker/MemoryCircuitBreaker.java
```diff
@@ -76,6 +79,10 @@ public boolean isTripped() {
       return false;
     }
 
+    if (!enabled) {
+      return false;
+    }
+
     long localAllowedMemory = getCurrentMemoryThreshold();
```
sigram, line 86: The same comment applies as in the CPU breaker regarding caching vs. calling this for every request. The cost should be minimal, it's just making a fast native call to read a value.

solr/core/src/test-files/solr/collection1/conf/solrconfig-memory-circuitbreaker.xml
```diff
@@ -82,8 +82,14 @@
 
     <useCircuitBreakers>true</useCircuitBreakers>
```
sigram, line 83: I hate typing too much :)  and overly verbose names are often easier to mistype.

What do you think about this (note the plural `circuitBreakers` - we have more than one now!):
```
<circuitBreakers enabled="true">
 <cpuBreaker enabled="true" threshold="75"/>
 <memBreaker enabled="true" threshold="75"/>
...
</circuitBreakers>
```

solr/solr-ref-guide/src/circuit-breakers.adoc
```diff
@@ -35,33 +35,68 @@ will be disabled globally. Per circuit breaker configurations are specified in t
 <useCircuitBreakers>false</useCircuitBreakers>
 ----
 
+This flag acts as the highest authority and global controller of circuit breakers. For using specific circuit breakers, each one
+needs to be individually enabled in addition to this flag being enabled.
```
sigram, line 39: If we change the format of the config this section needs to be updated too.

solr/solr-ref-guide/src/circuit-breakers.adoc
```diff
@@ -35,33 +35,68 @@ will be disabled globally. Per circuit breaker configurations are specified in t
...
+<cpuCircuitBreakerEnabled>true</cpuCircuitBreakerEnabled>
+----
+
+Note that this configuration will be overridden by the global circuit breaker flag -- if circuit breakers are disabled, this flag
+will not help you.
+
+To set the triggering threshold in units of CPU utilization, use the following parameter.
+[source,xml]
+----
+<cpuCircuitBreakerThreshold>75</cpuCircuitBreakerThreshold>
```
sigram, line 92: The value of 75 would rarely make sense for a production system, maybe change this to 20?

solr/core/src/java/org/apache/solr/util/circuitbreaker/MemoryCircuitBreaker.java
atris: So we do not need to cache here.

solr/core/src/java/org/apache/solr/util/circuitbreaker/MemoryCircuitBreaker.java
```diff
@@ -88,13 +95,20 @@ public boolean isTripped() {
 
   @Override
   public String getDebugInfo() {
-    if (seenMemory.get() == 0L || allowedMemory.get() == 0L) {
+    if (seenMemory.get() == 0.0 || allowedMemory.get() == 0.0) {
```
madrob, line 98: I don't think this was supposed to change

solr/core/src/java/org/apache/solr/util/circuitbreaker/MemoryCircuitBreaker.java
```diff
@@ -43,22 +43,25 @@
   private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
   private static final MemoryMXBean MEMORY_MX_BEAN = ManagementFactory.getMemoryMXBean();
 
+  private boolean enabled;
   private final long heapMemoryThreshold;
 
   // Assumption -- the value of these parameters will be set correctly before invoking getDebugInfo()
-  private final ThreadLocal<Long> seenMemory = new ThreadLocal<>();
-  private final ThreadLocal<Long> allowedMemory = new ThreadLocal<>();
+  private final ThreadLocal<Long> seenMemory = ThreadLocal.withInitial(() -> 0L);
```
madrob, line 50: static

solr/core/src/test/org/apache/solr/util/TestCircuitBreaker.java
```diff
@@ -41,6 +45,9 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import static org.hamcrest.CoreMatchers.containsString;
+
+@SuppressWarnings({"rawtypes"})
```
madrob, line 50: Do we have to suppress here? I'd rather not have warnings in the new code to begin with

solr/core/src/test/org/apache/solr/util/TestCircuitBreaker.java
```diff
@@ -119,33 +130,42 @@ public void testBuildingMemoryPressure() {
...
+
       CircuitBreaker circuitBreaker = new BuildingUpMemoryPressureCircuitBreaker(h.getCore().getSolrConfig());
 
       h.getCore().getCircuitBreakerManager().register(circuitBreaker);
 
+      List<Future> futures = new ArrayList<>();
+
       for (int i = 0; i < 5; i++) {
-        executor.submit(() -> {
+        Future future = executor.submit(() -> {
```
madrob, line 142: `Future<?>`

solr/solr-ref-guide/src/circuit-breakers.adoc
```diff
@@ -32,36 +32,58 @@ will be disabled globally. Per circuit breaker configurations are specified in t
...
 Consider the following example:
 
 JVM has been allocated a maximum heap of 5GB (-Xmx) and memoryCircuitBreakerThresholdPct is set to 75. In this scenario, the heap usage
 at which the circuit breaker will trip is 3.75GB.
 
-Note that this circuit breaker is checked for each incoming search request and considers the current heap usage of the node, i.e every search
-request will get the live heap usage and compare it against the set memory threshold. The check does not impact performance,
-but any performance regressions that are suspected to be caused by this feature should be reported to the dev list.
 
+=== CPU Utilization Based Circuit Breaker
```
madrob, line 69: @sigram we could call this Load Avg Based Circuit Breaker, WDYT? Maybe that is too much implementation detail?

solr/server/solr/configsets/_default/conf/solrconfig.xml
```diff
@@ -615,10 +612,29 @@
...
+     and the average load over the last minute at which the circuit breaker should start rejecting queries.
+
+     Consider a scenario where the max heap allocated is 4 GB and memoryCircuitBreakerThreshold is
+     defined as 75. Threshold JVM usage will be 4 * 0.75 = 3 GB. Its generally a good idea to keep this value between 75 - 80% of maximum heap
+     allocated.
+
+     If, at any point, the current JVM heap usage goes above 3 GB, queries will be rejected until the heap usage goes below 3 GB again.
+     If you see queries getting rejected with 503 error code, check for "Circuit Breakers tripped"
+     in logs and the corresponding error message should tell you what transpired (if the failure
+     was caused by tripped circuit breakers).
```
madrob, line 630: This is not part of this CB, right?

solr/core/src/test/org/apache/solr/util/TestCircuitBreaker.java
```diff
@@ -206,11 +286,12 @@ public FakeMemoryPressureCircuitBreaker(SolrConfig solrConfig) {
     @Override
     protected long calculateLiveMemoryUsage() {
       // Return a number large enough to trigger a pushback from the circuit breaker
+      System.out.println("I AM SENDING MAX VALUE");
```
madrob, line 289: log, not sys out.

solr/core/src/test/org/apache/solr/util/TestCircuitBreaker.java
```diff
@@ -157,6 +177,60 @@ public void testBuildingMemoryPressure() {
     }
   }
 
+  public void testFakeCPUCircuitBreaker() {
+    ExecutorService executor = ExecutorUtil.newMDCAwareCachedThreadPool(
```
madrob, line 181: Better practice to declare this right before the `try` block, without intervening declarations.

atris:
Thank you @madrob and @ab!

PR #1738, opened 2020-08-11
LUCENE-9454: upgrade hamcrest to version 2.2.

dweiss:
Precommit passes with ant and gradle. Running tests now.


lucene/test-framework/build.gradle
```diff
@@ -22,9 +22,13 @@ description = 'Framework for testing Lucene-based applications'
 dependencies {
   api project(':lucene:core')
 
-  api ("com.carrotsearch.randomizedtesting:randomizedtesting-runner")
-  api ('org.hamcrest:hamcrest-core')
-  api ("junit:junit")
+  api ("com.carrotsearch.randomizedtesting:randomizedtesting-runner", {
```
dweiss, line 25: Explanation: these exclusions are needed so that we don't pull in hamcrest-core which is a deprecated package just pointing at hamcrest.

PR #1740, opened 2020-08-11
LUCENE-9458: WDGF and WDF should tie-break by endOffset

dsmiley:
Can happen with catenateAll and not generating word xor number part when the input ends with the non-generated sub-token.

https://issues.apache.org/jira/browse/LUCENE-9458

CC @jimczi maybe you could review this please; I believe you reviewed the predecessor.


dsmiley:
> I wonder if this is something that we should enforce somewhere. It seems that you only want the original term to come first so maybe we can mark it with a special flag ?

Even if the test is changed to not involve the original term (thus don't expect it in output either), the test would reveal a problem with the "catenateAll" token (different from "preserveOriginal").

> Why does it matter that it appears first ? Is it to be compatible with other filters like the synonym filter ?

It "looks right" to me that it comes first intuitively.  Longest tokens first that start at the same position.  That's not a great reason, I realize.  We should tie break on something so that it's not arbitrary -- a better reason.  Also, at my company I have a delegating TokenFilter to WDGF that presumes this is the case, and this is how I found this inconsistency.

I want to try a "fuzz test" of WDGF to see if a pure start & end offset based ordering is sufficient, or is it truly necessary to also look at the position increment and position length.  My theory is that with what WDGF/WDF does, the offsets alone are fine to sort on because I don't think any later sub-token ("later" by offset) would have a token position happening earlier, and that likewise longer pos lengths come first.  The fuzz test would use a small dictionary of small sub-tokens and then it'd recombine them randomly to see if the graph it produces is identical to an alternate WDGF/WDF with tweaked token sort rules.  I don't think this'd be committable because the "alternate" would be temporary changes to the sorter compare implementation that looks at a system property.

lucene/analysis/common/src/java/org/apache/lucene/analysis/miscellaneous/WordDelimiterFilter.java
```diff
@@ -398,10 +399,20 @@ public void reset() throws IOException {
...
+      // earlier start offset
       int cmp = Integer.compare(startOff[i], startOff[j]);
-      if (cmp == 0) {
-        cmp = Integer.compare(posInc[j], posInc[i]);
+      if (cmp != 0) {
+        return cmp;
       }
+      
+      // earlier position
+      cmp = Integer.compare(posInc[j], posInc[i]);
```
dsmiley, line 409: @bruno-roustant pointed out to me that the existing comparison here looks faulty -- notice it's doing the *later* position due to the swap of "j" and "i".  This is how it was before.  I think it's wrong, and makes me suspect that it's likely this comparison yields a zero.  If true, this is more evidence that suggest the sort can merely look at start & end offsets.

dsmiley:
Sorry for the delay; I have had a solution for over a month locally but didn't share it yet.  I'm pretty comfortable with what I just pushed.

I did some "fuzz testing" to determine if there was a token ordering difference given the same input but varying the `compare` logic.  It revealed that I could simplify WDGF's `compare` logic to only look at the start and end offset.  However, WDF's `compare` failed this exercise; my change introduced a new position in some cases, although not the ones I explicitly tested for.  WDF is more complicated (to me any way), and furthermore WDF is deprecated so I'm not motivated to disturb the logic there.

I added the fuzz testing here as a comment.  It's not really committable live because it's requires modifying WDGF's `compare` to actually do two compares and assert the same result.  Without that, there is no assertion being checked -- no comparison.

Suggested CHANGES.txt would be improvement:
"WordDelimiterGraphFilter should tie-break by endOffset to emit longer tokens first. The same graph is produced."

dsmiley:
@jimczi I dunno if you want to look at this again but I plan to commit within the next day for 8.7.

PR #1742, opened 2020-08-12
Standalone distribution assembly for Luke

dweiss:


mocobeta:
Thanks, it works very well for me! I added dependencies on all analysis modules and suggester. 

When I run `./gradlew :lucene:luke:run`, I got this error and then figured out that `-Dant.home` property is needed (also Ant must be installed on the machine beforehand) to run the task.
```
Execution failed for task ':lucene:luke:run'.
> Execute failed: java.io.IOException: Cannot locate antRun script: Property 'ant.home' not found
```
We are dropping ant build, and the `assemble` task works as an alternative to `ant run` (to me), do we need this task at all...?

dweiss:
I'd add those dependencies to 'standalone' configuration, not implementation - so that the implementation configuration is indeed minimal? 

As for the launcher: this is caused by 'vmlauncher: false'. If you set it to true then it'll work but you'd need to know exact path to 'java' - which you can get from Gradle's VM API. Can you try to figure it out? We can leave that 'run' task for Uwe. :)

mocobeta:
> I'd add those dependencies to 'standalone' configuration, not implementation

I moved them. Thanks.

mocobeta:
> As for the launcher: this is caused by 'vmlauncher: false'. If you set it to true then it'll work but you'd need to know exact path to 'java' - which you can get from Gradle's VM API. Can you try to figure it out? 

With setting 'vmlauncher' to true, the launcher seems to work for me. I didn't need any other changes or configuration... am I missing something?

dweiss:
I think the VM launcher doesn't respect search path on all environments. You have to provide an explicit path to the binary you're willing to launch. This is how it used to work, at least. I'm sure it'll be more consistent and reliable to have an explicit path to java. Something like this should retrieve it:

     JavaInstallationRegistry registry = project.extensions.getByType(JavaInstallationRegistry)
     JavaInstallation currentJvm = registry.installationForCurrentVirtualMachine.get()
     return currentJvm.jdk.get().javadocExecutable.asFile

mocobeta:
Ah, I got it. Thanks again for your patience. https://github.com/apache/lucene-solr/pull/1742/commits/243979a3eb42ba027e1fa9c82745c65a88d386cd works for me.

PR #1745, opened 2020-08-12
Append MultiCollectorTest to TestMultiCollector.

cpoerschke:
As part of https://issues.apache.org/jira/browse/LUCENE-8626 I noticed that `MultiCollector` has two tests. This pull request proposes to unify them into one test.


cpoerschke:
> Was this a straight copy from one class to the other? ...

Yes, I did `cat $oneFile >> $otherFile` and then in the IDE stripped out the middle duplicate license and import bits, adding extra imports at the top.

PR #1746, opened 2020-08-12
Fix syntax warning in smokeTestRelease.py

munendrasn:
while verifying the 8.6.1 release with the latest python(3.8.5), observed
one SyntaxWarning.
https://adamj.eu/tech/2020/01/21/why-does-python-3-8-syntaxwarning-for-is-literal/


munendrasn:
@janhoy 
Thanks for the review

PR #1747, opened 2020-08-13
SOLR-14751: Zookeeper Admin screen not working for old ZK versions

janhoy:
See https://issues.apache.org/jira/browse/SOLR-14751

Solution is to treat `NoNodeException` not as an unrecoverable exception, but as expected for older ZK versions and return empty string.

A workaround for 8.6.x is to do a

    bin/solr zkroot /zookeeper/config

Then the Admin UI in 8.6 will start working with old ZK.


janhoy:
SOLR-14751

PR #1748, opened 2020-08-13
LUCENE-9447: Increase block sizes for stored fields.

jpountz:
BEST_SPEED and BEST_COMPRESSION keep the same compression algorithms but
increase their respective block sizes from 16kB and 60kB to 64kB and
256kB.

I disabled the per-thread caching of buffers, which could likely become
memory hogs with these greater block sizes.


jpountz:
I'm closing in favor of #1762 and I'll open a separate PR to disable the per-thread caching of buffers.

PR #1754, opened 2020-08-14
SOLR-14700: avoid NPEs in TupleStream.getShards() when streamContext … …is null.

cpoerschke:
https://issues.apache.org/jira/browse/SOLR-14700


solr/solrj/src/java/org/apache/solr/client/solrj/io/stream/TupleStream.java
```diff
@@ -141,19 +141,30 @@ public UUID getStreamNodeId(){
...
+      if (streamContext != null) {
+        solrParams = new ModifiableSolrParams(streamContext.getRequestParams());
+      } else {
+        solrParams = new ModifiableSolrParams();
+      }
       solrParams.add(requestParams);
 
       RequestReplicaListTransformerGenerator requestReplicaListTransformerGenerator =
-          Optional.ofNullable(streamContext.getRequestReplicaListTransformerGenerator()).orElseGet(RequestReplicaListTransformerGenerator::new);
+          Optional.ofNullable(streamContext != null ? streamContext.getRequestReplicaListTransformerGenerator() : null).orElseGet(RequestReplicaListTransformerGenerator::new);
```
madrob, line 167: There's a lot of complexity in this line, can we make it an if/else without wrapping and unwrapping an Optional? We are already doing a check on `streamContext != null` 6 lines up.

solr/solrj/src/java/org/apache/solr/client/solrj/io/stream/TupleStream.java
```diff
@@ -141,19 +141,30 @@ public UUID getStreamNodeId(){
       shards = shardsMap.get(collection);
     } else {
       //SolrCloud Sharding
-      CloudSolrClient cloudSolrClient =
-          Optional.ofNullable(streamContext.getSolrClientCache()).orElseGet(SolrClientCache::new).getCloudSolrClient(zkHost);
+      SolrClientCache solrClientCache = (streamContext != null ? streamContext.getSolrClientCache() : null);
+      final SolrClientCache localSolrClientCache;
+      if (solrClientCache == null) {
+        solrClientCache = localSolrClientCache = new SolrClientCache();
```
madrob, line 147: This is clever, add a comment that we're saving `localCache` because we need to close it later. :)

solr/solrj/src/java/org/apache/solr/client/solrj/io/stream/TupleStream.java
```diff
@@ -141,19 +141,30 @@ public UUID getStreamNodeId(){
       shards = shardsMap.get(collection);
     } else {
       //SolrCloud Sharding
-      CloudSolrClient cloudSolrClient =
-          Optional.ofNullable(streamContext.getSolrClientCache()).orElseGet(SolrClientCache::new).getCloudSolrClient(zkHost);
+      SolrClientCache solrClientCache = (streamContext != null ? streamContext.getSolrClientCache() : null);
```
madrob, line 144: I think this ternary folds into the conditional that we have:

```
      SolrClientCache scc, lscc;
      if (streamContext != null) {
        scc = streamContext.getSolrClientCache();
      } else {
        scc = lscc = new SolrClientCache();
      }
```

solr/solrj/src/test/org/apache/solr/client/solrj/io/stream/StreamingTest.java
```diff
@@ -2716,4 +2716,11 @@ private ParallelStream parallelStream(TupleStream stream, FieldComparator compar
     return pstream;
   }
 
+  public void testCloudSolrStreamWithoutStreamContext() throws Exception {
+    SolrParams sParams = StreamingTest.mapParams("q", "*:*", "fl", "id", "sort", "id asc");
+    try (CloudSolrStream stream = new CloudSolrStream(zkHost, COLLECTIONORALIAS, sParams)) {
+      stream.open();
```
madrob, line 2722: I really like how simple this test is.

solr/solrj/src/java/org/apache/solr/client/solrj/io/stream/TupleStream.java
cpoerschke: It's possible for `streamContext.getSolrClientCache()` to return `null` and in that case combining the ternary and the conditional would give a different result. Would it help to have a comment re: that subtlety perhaps?

solr/solrj/src/java/org/apache/solr/client/solrj/io/stream/TupleStream.java
cpoerschke: Good idea, will do.

solr/solrj/src/java/org/apache/solr/client/solrj/io/stream/TupleStream.java
cpoerschke: Good question. I just tried to fold the logic into the `streamContext != null` above but then the unit tests failed and (rightly so) identified that `streamContext.getRequestReplicaListTransformerGenerator()` can return `null`. We could still remove use of the `Optional` of course but it would require an additional `if (requestReplicaListTransformerGenerator == null)` block then. Thought?

madrob: Yea, I think an extra null check is going to be more clear than the use of Optional here.

solr/solrj/src/java/org/apache/solr/client/solrj/io/stream/TupleStream.java
madrob: Yea, add a comment. I definitely missed that subtlety.

madrob:
I just learned of `Objects.requireNonNullElse()` that can make some of this more readable. Up to you if you want to switch code to that or leave as is, @cpoerschke I'm happy either way!

PR #1755, opened 2020-08-14
SOLR-14753: Improve thread annotation name

MarcusSorealheis:
# Description

The thread safety annotation names are misleading/confusing for readers, yet the purpose is to clarify things for readers. Here is a fix proposed by @cpoerschke 

# Solution

Change `SolrSingleThreaded` to `SolrThreadUnsafe`

# Tests

No tests required. 

# Checklist


solr/solrj/src/java/org/apache/solr/common/annotation/SolrSingleThreaded.java
```diff
@@ -29,6 +29,6 @@
 @Documented
 @Retention(SOURCE)
 @Target(TYPE)
-public @interface SolrSingleThreaded {
+public @interface SolrThreadUnsafe {
```
anshumg, line 32: Please change the name of the file too.

MarcusSorealheis: oops

MarcusSorealheis: stepped away to get dinner. back now

MarcusSorealheis: @anshumg Should be good now. Mid-work I was starting to get headache from now eating.

solr/solrj/src/java/org/apache/solr/common/annotation/SolrThreadUnsafe.java
```diff
@@ -29,6 +29,6 @@
 @Documented
 @Retention(SOURCE)
 @Target(TYPE)
-public @interface SolrSingleThreaded {
```
cpoerschke, line 32: Via https://issues.apache.org/jira/browse/SOLR-13998 the `SolrSingleThreaded` public interface was included in releases from https://github.com/apache/lucene-solr/blob/releases/lucene-solr/8.4.0/solr/solrj/src/java/org/apache/solr/common/annotation/SolrSingleThreaded.java onwards. And so potentially folks could be using it in their custom code built against 8.4 onwards, though this might be unlikely in practice perhaps, it's difficult to know.

I wonder if we have or want to have some sort of backwards compatibility support for annotations? E.g. the scope of this pull request here could become _"Add SolrThreadUnsafe annotation, deprecate SolrSingleThreaded annotation."_ i.e. we keep `SolrSingleThreaded` around as deprecated for the remainder of the 8x releases and a separate commit for master branch only would remove the deprecated interface from 9.0 onwards.

Thoughts?

MarcusSorealheis: I think it's a good thought. The reason I think it is so prescient is the fact that this annotation originated in a private branch where I expect many pieces will land in open source. The original annotation name may be used in that code. Keeping this name for now will make it easier to bring that work into open source.

If it's in their custom code, I would hope the name change would be trivial.

chatman:
Thanks for the PR, Marcus. I committed this change as part of SOLR-14731. Can you please close both the PRs (as usual, I forgot to do it via commit message and my github credentials are still not in order, sorry)?

solr/solrj/src/java/org/apache/solr/common/annotation/SolrThreadUnsafe.java
anshumg: I thought I left a comment here but seems like I missed it.

I agree that it will make it easier for people to use the wrong annotation, but that's something that the person who writes/reviews/commits should notice. It's also something we could change, but removing this annotation is going to lead to a breaking change which we might have to just live with for back-compat reasons.

The private branch is no longer really in use, so we don't have to worry about anything from that specific branch making it into upstream.

Let's leave the old annotation as deprecated.

anshumg: Was late to the party here :)

Thanks for taking care of this.

PR #1764, opened 2020-08-19
Ensure we only rollback IW once

s1monw:
Today we might rollback IW more than once if we hit an exception during
the rollback code when we shutdown. This change moves the rollback code outside
of the try block to ensure we always roll back but never roll back twice.

### DIFF
--- a/lucene/core/src/test/org/apache/lucene/index/TestIndexWriterExceptions.java
+++ b/lucene/core/src/test/org/apache/lucene/index/TestIndexWriterExceptions.java
@@ -2023,4 +2023,44 @@ public class TestIndexWriterExceptions extends LuceneTestCase {
 
     dir.close();
   }
+
+
+  public void testOnlyRollbackOnceOnException() throws IOException {
+    AtomicBoolean once = new AtomicBoolean(false);
+    InfoStream stream = new InfoStream() {
+      @Override
+      public void message(String component, String message) {
+        if ("TP".equals(component) && "rollback before checkpoint".equals(message)) {
+          if (once.compareAndSet(false, true)) {
+            throw new RuntimeException("boom");
+          } else {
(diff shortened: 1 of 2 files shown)


lucene/core/src/test/org/apache/lucene/index/TestIndexWriterExceptions.java
```diff
@@ -2023,4 +2023,44 @@ public void eval(MockDirectoryWrapper dir) throws IOException {
...
+    };
+    try (Directory dir = newDirectory()) {
+      try (IndexWriter writer = new IndexWriter(dir, newIndexWriterConfig().setInfoStream(stream)){
+        @Override
+        protected boolean isEnableTestPoints() {
+          return true;
+        }
+      }) {
+      }
+    } catch (RuntimeException x) {
```
dnhatn, line 2060: why `x` ;)

s1monw: yeah why x.... :D

dweiss: I think I have to veto this. I liked x better.

s1monw: let the bikeshedding begin 👯 I call a vote...

dweiss: In the spirit peaceful atmosphere: let's call it e + x = "ex"...

msokolov: I'm really happy there were no votes for runtimeException (IDE generated code fancult vote)

PR #1766, opened 2020-08-20
SOLR-14765: optimize DocList creation for sort-irrelevant cases

magibney:


cpoerschke:
continued in https://github.com/apache/solr/pull/592

PR #1769, opened 2020-08-20
SOLR-14789: Absorb the docker-solr repo.

HoustonPutman:
WIP - https://issues.apache.org/jira/browse/SOLR-14789

**This is intended for 9.0 only, and will not be backported to 8.x.**  
We can continue using the docker-solr repo for all 8.x.y releases.

### Goals

The goal of this PR is to move all functionality for how current solr images are built to lucene solr. However there are clearly a lot of other things we want to accomplish too:

- Migrate functionality from the `solr/docker/include/scripts` into `solr/bin` or Solr itself
- Support the "official" docker releases (https://hub.docker.com/_/solr not just https://hub.docker.com/apache/solr)
- Make the solr docker file more "cloud native"

Since this docker image will not be officially used until 9.0 at the earliest, this gives us time to iterate and accomplish all of these goals. However merging in this initial functionality that works like the current docker-solr setup first lets us more easily make and test these changes.

### Setup of `/solr/docker`

`gradle assemble` will now create a docker image that has been created with contents of `:solr:packaging`.

The "Solr image" is broken up into 2 docker files.

1. The `package` dockerfile, which contains the solr release under `/opt/solr-build/solr-${VERSION}`. This image requires no other content at the end.
2. The `runtime` dockerfile, which is given the `package` image name as an input and generates everything currently setup in the Solr docker image, copying the package contents from the `package` image.

The `runtime` docker image is the one that will be pushed to Dockerhub eventually, and is run by users.

The idea is that there is also a task to generate the release docker image (which can be found under `/solr/docker/package/Dockerfile.release-package`) that is then passed to the runtime docker image. But this is a future goal and not necessary in order to merge this PR IMO.

I basically learned gradle to write this, so please give any advice on how I could make the build process better. I'm sure it is not optimal.

### Testing

Once you run `gradle assemble`, you can test the new docker image in two ways:

- `gradle :solr:docker:test`, which will run the docker-solr integration tests using the newly assembled solr image. This will eventually fail when it reaches the test that needs root permissions however, since gradle doesn't have root permissions itself. This is a TODO.
- `docker run -d -p 8983:8983 -e SOLR_JETTY_HOST="0.0.0.0" apache/solr:9.0.0-SNAPSHOT`, then visit the admin screen at http://localhost:8983 . 

TODO:
- [ ] Add ability to build with "release" package
- [ ] Remove ability to push the `package` image.
- [ ] Integrate `gradle :solr:docker:tests` into the overall `gradle tests`
- [x] Make all of the tests work with gradle.


HoustonPutman:
I should point out that the `tests`, `docs` and `runtime/include/scripts` are almost identical copies from the docker-solr repo. Some small changes have been made to the `scripts` files, but no actual runtime logic has changed.

I'd be happy to copy exactly from the repo and make the small changes in a different PR, if that would make y'all more comfortable.

madrob:
I was able to build this and did some very simple validation after launching containers using:

```
docker run --rm -d --name zk-for-solr -p 2181:2181 zookeeper
docker run --rm -p 8983:8983 --link zk-for-solr:zk-for-solr -e SOLR_JETTY_HOST="0.0.0.0" apache/solr:9.0.0-SNAPSHOT -c -z zk-for-solr:2181
docker run --rm -p 8984:8984 --link zk-for-solr:zk-for-solr -e SOLR_JETTY_HOST="0.0.0.0" apache/solr:9.0.0-SNAPSHOT -c -z zk-for-solr:2181 -p 8984
```

Can we make sure that user run instructions are included in the ref guide, or possibly the tutorial?

dsmiley:
While SOLR_JETTY_HOST is a nice security feature for someone running Solr outside of Docker, isn't it pointless (needless pain) for anyone using Docker?  I think so, thus the docker settings could set this so that you don't have to.

janhoy:
My first attempt at this gave some bash errors:

```
./gradlew assemble
> Configure project :solr:docker
readlink: illegal option -- f
usage: readlink [-n] [file ...]
/Users/janhoy/git/lucene-solr/solr/docker/tests/cases/create_core_exec/test.sh: line 18: ./../../shared.sh: No such file or directory

FAILURE: Build failed with an exception.
```

I recognize this from the docker-solr repo, the test scripts use some commands that do not work on MacOS. So I added the workaround with putting gnu variants of these tools in my path, but then I could still not make the assemble task run:

```
> Configure project :solr:docker
Test /Users/janhoy/git/lucene-solr/solr/docker/tests/cases/create_core_exec apache/solr:9.0.0-SNAPSHOT
Cleaning up left-over containers from previous runs
Running test_apache_solr_9.0.0_SNAPSHOT
Unable to find image 'apache/solr:9.0.0-SNAPSHOT' locally
docker: Error response from daemon: pull access denied for apache/solr, repository does not exist or may require 'docker login': denied: requested access to the resource is denied.
See 'docker run --help'.

FAILURE: Build failed with an exception.
```

I had to uncomment the `task test()` from `docker/build.gradle` and then my build succeeded.

HoustonPutman:
@janhoy The tests should hopefully work for you now, though I do have the gnu utils installed, so I'm not sure they will.

The tests are no longer run by default in the assemble, so that should work for you now regardless.

I have done some modification on the tests to simplify out some of the logic. I also changed the permissions of folders created during the tests, so that it doesn't require root permissions anymore to run them. This works for me, but it might not work for others. Not confident on that yet.

janhoy:
Ok, will give it another try soon.

Btw - should we set `SOLR_JETTY_HOST` as an `ENV` or `ARG` or something?

HoustonPutman:
Its currently set as an ENV, the last line of the large ENV section.

solr/docker/docs/docker-networking.md
```diff
@@ -0,0 +1,249 @@
...
+that specifically use the `--ip` option.
+
+```
+docker network create --driver=overlay --subnet 192.168.22.0/24 --ip-range=192.168.22.128/25 netzksolr
+```
+
+As a simple test, check the automatic assignment and specific assignment work:
+
+```
+mak@trinity10:~$ docker run -i --rm --net=netzksolr busybox ip -4 addr show eth0 | grep inet
```
thelabdude, line 30: let's remove `mak@trinity10:~$` from the commands

HoustonPutman:
Thanks for reviewing the docs as well. I plan on migrating a lot of the docs to the ref guide, but I think we should probably leave a bulk of that work to a separate PR.

The earlier we get this PR merged, the easier it will be to break up the work we want to do into separate JIRAs that can be attacked independently. (more extensive documentation, reduction of the scripts, make it solr more "cloud native", a plan for release images, etc)

solr/docker/build.gradle
```diff
@@ -0,0 +1,95 @@
...
+description = 'Solr Docker image'
+
+docker {
+  name = "apache/solr:${version}"
+  files file('include'), '.dockerignore'
+  buildArgs(['BASE_IMAGE' : 'openjdk:11-jre-slim', 'SOLR_PACKAGE_IMAGE' : 'apache/solr-build:local-package', 'SOLR_VERSION': "${version}"])
+}
+
+// Create the docker image as a part of the assemble task
+assemble.dependsOn tasks.docker
```
uschindler, line 39: Hi,
as noted on Slack already. Like for the refguide, don't yet execute docker by default on assemble.
I would make it optional for now, until we find a way how to assemble docker without docker or make it print a warning if no docker is enabled.

dsmiley: +1 100% to what Uwe said

solr/docker/docs/docker-compose.yml
```diff
@@ -0,0 +1,13 @@
+version: '3'
+services:
+  solr:
+    image: solr
```
dsmiley, line 4: Here and probably some README type places, the assumption is that the image is named "solr".  Yet it appears that you're deciding that the name "apache/solr" is the new default name.  I think pick one or or the other.

solr/docker/tests/shared.sh
```diff
@@ -90,18 +90,44 @@ function prepare_dir_to_mount {
...
-    sudo chmod -R 777 "$folder"
-    sudo chown -R "$userid" "$folder"
-  else
+  if command -v setfacl &> /dev/null; then
     setfacl -m "u:$userid:rwx" "$folder"
   fi
 }
+
+
+# Shared setup
```
dsmiley, line 106: Are you making general improvements to the test scripts, or do they relate to necessities of the port to lucene-solr?  I suggest suppressing your natural urge to improve code in a porting job to limit to just do the port.  Future work items can refactor common code out of bash scripts or whatever else.

HoustonPutman: These changes were necessary for separating out the test output from the source directory from the tests as well as having unique container names for the tests. This was needed in the migration, because the tests assumed that you were running in a particular directory layout before (the layout in docker-solr). Moving to gradle, this needed to be a lot more flexible.

HoustonPutman: But I am happy to leave the testing stuff to a different PR if that will be easier to review.

solr/docker/docs/docker-compose.yml
HoustonPutman: I will pick apache/solr. It will be easier for us to find and replace if we decide to do something else later.

HoustonPutman: I'm also going to go with the assumption that if we go with apache/solr, we will backfill all of the docker-solr releases to apache/solr. (So all of the previous version tags will be available under the new name.)

solr/docker/tests/shared.sh
dsmiley: Your justification makes sense.  The cause of those changes weren't clear to me before.

dsmiley:
CC @dweiss you might want to briefly review this new module for the gradle aspect

dsmiley:
FYI related conversation on the dev list about higher level integration tests: https://lists.apache.org/thread.html/r27d742357f9e1342cce1c9aaa215a10d94f64b43b2e681f19fa8d150%40%3Cdev.lucene.apache.org%3E

solr/docker/build.gradle
```diff
@@ -0,0 +1,94 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+plugins {
+  id "base"
+  id "com.palantir.docker" version "0.25.0"
```
madrob, line 20: I think our convention is to declare the plugin version in `build.gradle` with `apply false`

solr/docker/package/Dockerfile.local-package
```diff
@@ -0,0 +1,5 @@
+ARG BASE_IMAGE=alpine:3.12
```
madrob, line 1: I'm confused why we need this as an intermediate docker image containing just the tgz? Also, why make this configurable and why not `FROM scratch`?

madrob:
Overall looks good, just a pair of minor questions/comments.

solr/docker/package/Dockerfile.local-package
HoustonPutman: Yeah, the idea was to allow for people to build in environments that don't have access to docker hub.

Didn't know about scratch, but it's perfect. Thanks for the suggestion.

solr/docker/Docker-FAQ.md
```diff
@@ -0,0 +1,320 @@
...
+-----------------------------------------------------
+
+See the [docker compose example](docs/docker-compose.yml).
+
+
+I'm confused about the different invocations of solr -- help?
+-------------------------------------------------------------
+
+The different invocations of the Solr docker image can look confusing, because the name of the
+image is "solr" and the Solr command is also "solr", and the image interprets various arguments in
```
madrob, line 227: should be `apache/solr`

HoustonPutman:
Thanks everyone for the help! Now we should be able to iterate more easily on the next tasks 😄

dsmiley:
I'm looking at the docker image this produces closely, and I see it's more than a bit bloated thanks to this line:
https://github.com/apache/lucene-solr/blob/3ae0ca23d937bef2865689748ac9e556b40aff38/solr/docker/Dockerfile#L45

    COPY --from=solr_package "/opt/solr-$SOLR_VERSION.tgz" "/opt/solr-$SOLR_VERSION.tgz"

Which adds an entire layer of waste that is all of Solr 😱 

I'm also a confused why we need an entire gradle module `:solr:docker:package` just for providing input to the Dockerfile.  I read the PR opening description which sort of says why it exists... but I think that could be provided all in one module using different tasks for the different steps.  I'm also not sure we need to support Dockerfile.release-package does since the image production process is merged with the project; it needn't download a .tgz from anywhere.  Right?  I don't think direct production of the image prohibits the goals stated above about producing an official image.

HoustonPutman:
The extra step exists because there was no consensus around how to do official release images.

If we want to decide that the official image should be built the same way as it currently is in the project (via the local build), then we can get rid of the sub-module and the extra step. However if we want to have the official image use the official release binaries, as it does in `docker-solr`, then we will need to keep the submodule.

I would have preferred to have all of this done in one module, but the gradle docker plugin only supports building one image per-module. So if we want to build multiple images (which is necessary for supporting the two image types, local and release), we need two modules.

I am all for not adding support for official binary release strategy, and consolidating into one docker file. I just don't want to make that decision unanimously.

dsmiley:
I'm glad your are amenable to changes, and that the complexity & Docker image weight I see will melt away if we only produce an image from the Solr assembly.  That is identical to the "official release" except packaging -- plain dir vs tgz of the same.  I can appreciate there were unknowns causing you to add this extra baggage because it might be useful but I prefer to follow a KISS/YAGNI philosophy so that we don't pay complexity debt on something not yet needed.

janhoy:
+1 to lighter weight. However, our users should somehow be able to verify that a Docker image pulled from Docker Hub (or downloaded from elsewhere) is indeed the officially voted-upon binaries that they find in the release repo. Downloads from mirrors are easy to verify as we provide `.sha512` and `.asc` files for them. Likewise [artifacts from maven](https://repo1.maven.org/maven2/org/apache/solr/solr-core/8.7.0/) also have `.asc` and `.sha1` files for every jar. Current docker-solr Dockerfile can be inspected in that it downloads the official tarball and validates GPG signature. The lightweight Dockerfile performs no such checks and cannot be validated the same way.

So here is my proposal. We build the docker image from folder instead of tgz, but also add documentation to our [download page](https://lucene.apache.org/solr/downloads.html) on how to verify the solr binaries inside the image. Could even script it:

    curl -o verify-docker.sh https://lucene.apache.org/solr/verify-docker.sh
    docker run --rm -v ./verify-docker.sh:/verify-docker.sh apache/solr:9.0.0 sh /verify-docker.sh

HoustonPutman:
I completely agree with everything you've said @janhoy, and really like that solution! I am by no means an expert on this verification stuff, so what do you imagine `verify-docker.sh` would be doing under the hood?

dsmiley:
Wouldn't it be simpler for the release manager to build the docker image, examine the sha256 hash of the image, and publish that to the download location, making it official?  Someone who wants to use the official Solr docker image who is ultra-paranoid can reference the image by hash like so:

    docker run --rm solr@sha256:02fe5f1ac04c28291fba23a18cd8765dd62c7a98538f07f2f7d8504ba217284d

That runs Solr 8.7, the official one.  It's compact and can even be broadcasted easily in the release announcement for future Solr releases for people to get _and run_ the latest release immediately, and be assured it's the correct one.

I wonder what other major Apache projects do.

janhoy:
> what do you imagine `verify-docker.sh` would be doing under the hood

It would verify the sha512 sum of each jar with the corresponding checksum published by the project, and also allow the user to verify the PGP signature of each jar to assert it was signed by a committer.

> Wouldn't it be simpler for the release manager to build the docker image, examine the sha256 hash of the image, and publish that to the download location, making it official

That's a great idea David, that the RM records the image SHA when pushing, and publishes that sha. People can then either pull with SHA or verify later with `docker images --digests solr`. That should be sufficient for most users. For those who want to futher assert that they use the very same binaries signed by the committer, we could still document how to perform PGP checks on the jars.

dsmiley:
> we could still document how to perform PGP checks on the jars

Just confirming... we agree that ./verify-docker.sh is needless, and we need to just _document_ (e.g. in the ref guide) how to verify individual JARs at the CLI.

janhoy:
Yea, add the image hash to download page and document how to check each jar file.

PR #1770, opened 2020-08-20
SOLR-14763 SolrJ HTTP/2 Async API using CompletableFuture

rishisankar:
# Description

This change adds a standard `CompletableFuture<NamedList<Object>> requestAsync(SolrRequest<?> request)` API to SolrClient, implemented for Http2SolrClient and CloudHttp2SolrClient (Http2SolrClient already had an async method implemented recently, but this was not standard/not extended to the cloud client). Additionally, this change removes custom AsyncListener/Cancellable interfaces, instead using CompletableFuture as a standard for async request methods.

# Solution

* An async method was already added to Http2SolrClient in [SOLR-14354](https://github.com/apache/lucene-solr/pull/1470/files). This change refactors the already existing method to fit the new method signature
* The LBHttp2SolrClient has a separate async method using its internal Req/Rsp classes, also added in SOLR-14354. This change also refactors this method to use CompletableFuture (although this doesn't override the SolrClient API, as it still uses Req/Rsp).
* Http2SolrCloudClient uses an LBHttp2SolrClient internally, so it can utilize LBHttp2SolrClient's async request method to make async calls as well. This refactors BaseCloudSolrClient to add CompletableFuture support so that Http2SolrClient can make async requests.
* For now, SolrClient will throw an UnsupportedOperationException if requestAsync method is called on a client besides Http2SolrClient or CloudHttp2SolrClient.

# Tests

Added test cases to Http2SolrClient to test async requests in the same way synchronous requests are tested. Also added a test case to test that async requests and normal requests made by CloudHttp2SolrClient index/query in the same way.

# Checklist


solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseCloudSolrClient.java
```diff
@@ -476,8 +478,11 @@ public void registerDocCollectionWatcher(String collection, DocCollectionWatcher
     assertZKStateProvider().zkStateReader.registerDocCollectionWatcher(collection, watcher);
   }
 
+  // TODO: async direct updates
```
rishisankar, line 481: Before going further with this implementation, I wanted to get another opinion on the general implementation approach I'm taking here -- i.e. changing internal methods to return a CompletableFuture even for synchronous requests, except that they are already completed unless the request is async. A perhaps more hacky alternative is wrapping the CompletableFuture within a NamedList (similar to what I did [here](https://github.com/apache/lucene-solr/pull/1688)).

solr/solrj/src/java/org/apache/solr/client/solrj/SolrClient.java
```diff
@@ -1290,6 +1295,39 @@ public SolrDocumentList getById(Collection<String> ids, SolrParams params) throw
...
+   *
+   * @param request the request to execute
+   *
+   * @return a {@link CompletableFuture} that tracks the progress of the async request. Supports cancelling requests via
+   * {@link CompletableFuture#cancel(boolean)}, adding callbacks/error handling using {@link CompletableFuture#whenComplete(BiConsumer)}
+   * and {@link CompletableFuture#exceptionally(Function)} methods, and other CompletableFuture functionality. Will
+   * complete exceptionally in case of either an {@link IOException} or {@link SolrServerException} during the request.
+   * Once completed, the CompletableFuture will contain a {@link NamedList} with the response from the server.
+   */
+  public CompletableFuture<NamedList<Object>> requestAsync(final SolrRequest<?> request) {
```
CaoManhDat, line 1327: I wonder should we put this method here? Since only Http2SolrClient gonna support it?

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
```diff
@@ -394,26 +393,26 @@ public void onHeaders(Response response) {
...
-            if (failure != CANCELLED_EXCEPTION) {
-              asyncListener.onFailure(new SolrServerException(failure.getMessage(), failure));
-            }
+            future.completeExceptionally(new SolrServerException(failure.getMessage(), failure));
           }
         });
-    return () -> req.abort(CANCELLED_EXCEPTION);
+    future.exceptionally((error) -> {
+      if (error instanceof CancellationException) {
+        req.abort(new Exception());
```
CaoManhDat, line 411: I feel the way cancellation is implemented is little bit wrong. When we abort a request the given exception will be called on listeners, in that case we don't want to cascade that exception further. By doing this way, the `new Exception()` will be throw to upper level.

rishisankar: Although the exception would be called by listeners, wouldn't the `future.completeExceptionally` that is used to handle the exception be a noop as the future has already been cancelled?

solr/solrj/src/java/org/apache/solr/client/solrj/SolrClient.java
rishisankar: I was wondering about that as well - my reason for putting it here was that if other client subclasses later implement it as well, I think having the API be standardized for all SolrClient subclasses would be good. But it would probably be better if a default implementation is added rather than throwing an error (perhaps by adding a separate `setExecutor` method that a user could call and have the default impl use that user-provided executor. Or initializing an executor in this class).

dsmiley: If possible, I'd prefer that requestAsync simply not take a collection param.  Instead, both the request itself might have it, or the SolrClient might have it as a default.  I know lots of methods on SolrClient take it but I'd prefer to cease that trend for simplicity.

RE having requestAsync on SolrClient at all -- I think throwing UnsupportedOperationException is acceptable.  We could implement a default implementation here last or in another issue.

solr/core/src/java/org/apache/solr/handler/component/HttpShardHandler.java
```diff
@@ -156,36 +155,31 @@ public void submit(final ShardRequest sreq, final String shard, final Modifiable
...
-      @Override
-      public void onStart() {
-        if (tracer != null && span != null) {
-          tracer.inject(span.context(), Format.Builtin.HTTP_HEADERS, new SolrRequestCarrier(req));
-        }
-        SolrRequestInfo requestInfo = SolrRequestInfo.getRequestInfo();
-        if (requestInfo != null) req.setUserPrincipal(requestInfo.getReq().getUserPrincipal());
-      }
+    long startTime = System.nanoTime();
+    if (tracer != null && span != null) {
```
dsmiley, line 159: I suspect that this code may be misplaced.  Formerly it was executed by the thread that called onStart.  Is that the same thread calling submit() here?

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseCloudSolrClient.java
dsmiley: Since we're merely talking about using a CompletableFuture wrapper (right?), it seems fine to me.  This is internal too.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseCloudSolrClient.java
```diff
@@ -1052,10 +1074,12 @@ public RouteException(ErrorCode errorCode, NamedList<Throwable> throwables, Map<
       }
     }
 
-    return resp;
+    return respFuture;
```
dsmiley, line 1077: very minor:  I'd prefer "rspFuture".  I see "response" abbreviated to "rsp" in general, much less to "resp".

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseCloudSolrClient.java
```diff
@@ -1167,8 +1191,38 @@ public RouteException(ErrorCode errorCode, NamedList<Throwable> throwables, Map<
     }
 
     LBSolrClient.Req req = new LBSolrClient.Req(request, theUrlList);
-    LBSolrClient.Rsp rsp = getLbClient().request(req);
-    return rsp.getResponse();
+    if (isAsyncRequest) {
+      CompletableFuture<LBSolrClient.Rsp> lbFuture = ((LBHttp2SolrClient) getLbClient()).requestAsync(req);
+      CompletableFuture<NamedList<Object>> apiFuture = new CompletableFuture<>();
+      lbFuture.whenComplete((result, throwable) -> {
+        if (lbFuture.isCompletedExceptionally()) {
```
dsmiley, line 1198: simply check if throwable is not null instead?

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
dsmiley: I'm a little curious about the code being referenced here:
* Why do an instanceof check on "error" at all; why not simply always req.abort(error)?
* Why create a new Exception when we already have a suitable Throwable type in variable "error"?

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseCloudSolrClient.java
```diff
@@ -1167,8 +1191,38 @@ public RouteException(ErrorCode errorCode, NamedList<Throwable> throwables, Map<
     }
 
     LBSolrClient.Req req = new LBSolrClient.Req(request, theUrlList);
-    LBSolrClient.Rsp rsp = getLbClient().request(req);
-    return rsp.getResponse();
+    if (isAsyncRequest) {
+      CompletableFuture<LBSolrClient.Rsp> lbFuture = ((LBHttp2SolrClient) getLbClient()).requestAsync(req);
```
tflobbe, line 1195: Maybe we want to document somewhere that in order to support async requests, the internal lb client needs to extend `LBHttp2SolrClient`?

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
CaoManhDat: > Although the exception would be called by listeners, wouldn't the future.completeExceptionally that is used to handle the exception be a noop as the future has already been cancelled?

I see your point, but I in this case from CompletableFuture perspective, it get called with `cancel()` first then `completeExceptionally` will that mess up with internal state of the future instance?

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseCloudSolrClient.java
```diff
@@ -853,6 +858,18 @@ public RouteException(ErrorCode errorCode, NamedList<Throwable> throwables, Map<
...
+      // synchronous requests should return an already completed future
+      return getNowOrException(makeRequest(request, collection, false));
+  }
+
+  CompletableFuture<NamedList<Object>> makeRequest(SolrRequest<?> request,
+                                                   String collection,
+                                                   boolean isAsyncRequest) throws SolrServerException, IOException {
+    if (isAsyncRequest && !(getLbClient() instanceof LBHttp2SolrClient)) {
+      log.warn("Asynchronous requests require HTTP/2 SolrJ client, defaulting to synchronous request.");
+      isAsyncRequest = false;
```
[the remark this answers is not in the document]
dsmiley: What you propose sounds good Rishi.

solr/core/src/java/org/apache/solr/handler/component/HttpShardHandler.java
rishisankar: I think they're the same thread? In the previous implementation, the same thread that calls `submit()` [calls the async request method](https://github.com/apache/lucene-solr/blob/master/solr/core/src/java/org/apache/solr/handler/component/HttpShardHandler.java#L160) and then within the async request method [calls `onStart()`](https://github.com/apache/lucene-solr/blob/master/solr/solrj/src/java/org/apache/solr/client/solrj/impl/LBHttp2SolrClient.java#L89) - so it seems to me like previously the thread that called `submit()` and `onStart()` are the same.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseCloudSolrClient.java
```diff
@@ -537,40 +541,100 @@ public void registerDocCollectionWatcher(String collection, DocCollectionWatcher
...
+        } else {
+          threadPool.submit(() -> {
+            try {
+              future.complete(getLbClient().request(lbRequest).getResponse());
+            } catch (Exception e) {
+              future.completeExceptionally(e);
+            }
+          });
+
+          responseFutures.put(url, future.whenComplete((result, error) -> {
```
dsmiley, line 622: Here I see you populate shardResponses & exceptions only for the synchronous code path.  I don't see why we don't do something with the async code path.  The CF that's created leaves scope and doesn't save its data anywhere.  I suspect the async code path with parallel updates is not tested.

dsmiley: Also, there needs to be synchronization to avoid concurrent population of shardResponses & exceptions structures.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseCloudSolrClient.java
```diff
@@ -580,23 +644,71 @@ public void registerDocCollectionWatcher(String collection, DocCollectionWatcher
...
+        for (CompletableFuture<NamedList<Object>> cf : futuresArray) {
+          cf.cancel(true);
+        }
+      }
+      return null;
+    });
+    return updateFuture;
+  }
+
+  private CompletableFuture<Void> doUpdatesWithoutExecutor(final Map<String, ? extends LBSolrClient.Req> routes,
```
dsmiley, line 660: Since the asyc & sync code paths are rather different inside here, I think you should structure this method as a big if-else.  The only part in common is like 2 lines.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseCloudSolrClient.java
```diff
@@ -476,8 +479,9 @@ public void registerDocCollectionWatcher(String collection, DocCollectionWatcher
     assertZKStateProvider().zkStateReader.registerDocCollectionWatcher(collection, watcher);
   }
 
-  @SuppressWarnings({"unchecked"})
-  private NamedList<Object> directUpdate(AbstractUpdateRequest request, String collection) throws SolrServerException {
+  private CompletableFuture<NamedList<Object>> directUpdate(AbstractUpdateRequest request,
```
dsmiley, line 482: I have to wonder if it's worthwhile to have directUpdate support Http2 Async callback given the complexity here?  I *do* like it internally for parallel updates -- doUpdatesWithExecutor.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseCloudSolrClient.java
```diff
@@ -862,15 +1015,39 @@ public RouteException(ErrorCode errorCode, NamedList<Throwable> throwables, Map<
...
     List<String> inputCollections =
         collection == null ? Collections.emptyList() : StrUtils.splitSmart(collection, ",", true);
-    return requestWithRetryOnStaleState(request, 0, inputCollections);
+
+    CompletableFuture<NamedList<Object>> apiFuture = new CompletableFuture<>();
+    final AtomicBoolean cancelled = new AtomicBoolean(false);
+    final AtomicReference<CompletableFuture<NamedList<Object>>> currentFuture = new AtomicReference<>();
+    apiFuture.exceptionally((error) -> {
+      if (apiFuture.isCancelled()) {
+        synchronized (cancelled) {
```
dsmiley, line 1024: This code is similar to the LB class where I provided feedback.  The same feedback may apply here -- I don't see the point of synchronization on cancelled or why cancelled exists if we have a CF which can hold that cancelled state.

dsmiley: I'll at-mention @CaoManhDat again whom may know as I think he authored this mechanism.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseCloudSolrClient.java
```diff
@@ -853,6 +993,19 @@ public RouteException(ErrorCode errorCode, NamedList<Throwable> throwables, Map<
...
+      return getNowOrException(makeRequest(request, collection, false));
+  }
+
+  /**
+   * Makes a request either synchronously or asynchronously depending on the isAsyncRequest parameter. The returned
+   * CompletableFuture will already be completed in the case of a sync request.
+   *
+   * For async requests, the internal LB Solr client given by getLbClient() must be an LBHttp2SolrClient.
+   */
+  CompletableFuture<NamedList<Object>> makeRequest(SolrRequest<?> request,
```
dsmiley, line 1006: Define as protected?

solr/solrj/src/java/org/apache/solr/client/solrj/SolrRequest.java
```diff
@@ -246,6 +248,42 @@ public final T process(SolrClient client) throws SolrServerException, IOExceptio
...
+   *
+   * @return a {@link CompletableFuture} that tracks the progress of the async request.
+   * Once completed, the CompletableFuture will contain the response.
+   */
+  public final CompletableFuture<T> processAsynchronously(SolrClient client, String collection) {
+    final long startNanos = System.nanoTime();
+    final CompletableFuture<NamedList<Object>> internalFuture = client.requestAsync(this, collection);
+    final CompletableFuture<T> apiFuture = new CompletableFuture<>();
+
+    internalFuture.whenComplete((result, error) -> {
```
dsmiley, line 265: I think this can be adapted to use `internalFuture.thenApply(...)`

solr/solrj/src/java/org/apache/solr/client/solrj/impl/LBHttp2SolrClient.java
```diff
@@ -112,95 +112,106 @@ public void onFailure(Exception e, boolean retryReq) {
...
     } catch (SolrServerException e) {
-      asyncListener.onFailure(e);
+      apiFuture.completeExceptionally(e);
     }
-    return () -> {
-      synchronized (cancelled) {
-        cancelled.set(true);
-        if (currentCancellable.get() != null) {
-          currentCancellable.get().cancel();
+    apiFuture.exceptionally((error) -> {
```
CaoManhDat, line 132: seems need to add `return` in front on this call, since `exceptionally` create a new future instance.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/LBHttp2SolrClient.java
```diff
@@ -112,95 +112,106 @@ public void onFailure(Exception e, boolean retryReq) {
...
+      if (apiFuture.isCancelled()) {
+        synchronized (cancelled) {
+          cancelled.set(true);
+          if (currentFuture.get() != null) {
```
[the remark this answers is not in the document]
CaoManhDat: Should we need to synchronized by `cancelled` here? The worst case here is cancel() will be called on previous future. 
Another point here is should we remove RetryListener here? With the `CompletableFuture` features, `RetryListener` just making things more complex.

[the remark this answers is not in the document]
dsmiley: With CompletableFuture, there's no harm in attempting to complete a future redundantly -- it's a no-op.  Thus if currentFuture is already cancelled, no extra work.

Removing RetryListener might mean not doing retry-s?  How do you propose retry-s be done without RL?  *Maybe* we don't need RL but we just need a static utility method somewhere like `boolean shouldRetry(Exception e)`?

dsmiley:
It's not obvious but if we return the result of CF.exceptionaly() then
*somehow* we need to ensure that the original CF can get completed for the
handler we add here to even get called.  It's not obvious; took some
playing with simple Java scratch files.  If we do nothing with the result
of CF.exceptionally then the handler added there *will* get called
(assuming there's an exception of course).  You can think of it like a
chain or tree.  We need to keep the head/root of the chain and use that to
trigger the actions along the chain.  If we pass the tail of the chain,
then it's useless.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/LBHttp2SolrClient.java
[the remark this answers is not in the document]
CaoManhDat: > Maybe we don't need RL but we just need a static utility method somewhere like boolean shouldRetry(Exception e)
Yeah something like that, so the first CF will have an `exceptionally` handler which create a new CF on exception, etc.

CaoManhDat:
Right, thank you for your clarification.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseCloudSolrClient.java
CaoManhDat: I will provide feedback on the above LBSolrClient.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseCloudSolrClient.java
```diff
@@ -537,40 +541,100 @@ public void registerDocCollectionWatcher(String collection, DocCollectionWatcher
...
+        if (isAsyncRequest) {
+          CompletableFuture<LBSolrClient.Rsp> reqFuture = getLbClient().requestAsync(lbRequest);
+          reqFuture.whenComplete((result, error) -> {
+            if (!reqFuture.isCompletedExceptionally()) {
+              future.complete(result.getResponse());
+            } else {
+              future.completeExceptionally(error);
+            }
+          });
+        } else {
```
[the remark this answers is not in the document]
rishisankar: I've added an implementation (in the SolrClient class) for requestAsync with an executor (so that the request can be canceled). Currently, it only takes an `ExecutorService`, although I can switch `ExecutorService` to `Executor` with the drawback of the async request not being cancellable - curious which you think would be better?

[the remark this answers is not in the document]
dsmiley: In my experience, everyone has and wants an ExecutorService.

solr/solrj/src/java/org/apache/solr/client/solrj/SolrRequest.java
```diff
@@ -260,18 +260,13 @@ public final T process(SolrClient client) throws SolrServerException, IOExceptio
...
-        T res = createResponse(client);
-        res.setResponse(result);
-        long endNanos = System.nanoTime();
-        res.setElapsedTime(TimeUnit.NANOSECONDS.toMillis(endNanos - startNanos));
-        apiFuture.complete(res);
-      } else {
-        apiFuture.completeExceptionally(error);
-      }
+
+    final CompletableFuture<T> apiFuture = internalFuture.thenApply((result) -> {
```
dsmiley, line 264: What if internalFuture completes exceptionally?

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseCloudSolrClient.java
```diff
@@ -926,136 +1092,191 @@ public RouteException(ErrorCode errorCode, NamedList<Throwable> throwables, Map<
...
-        for (Object invalidEntries : invalidStates.entrySet()) {
-          @SuppressWarnings({"rawtypes"})
-          Map.Entry e = (Map.Entry) invalidEntries;
-          getDocCollection((String) e.getKey(), (Integer) e.getValue());
-        }
+    requestWithRetryOnStaleStateHelper(request, retryCount, inputCollections, requestedCollections, isAdmin,
+        isAsyncRequest, apiFuture, currentFuture);
+  }
 
+  private void requestWithRetryOnStaleStateHelper(final SolrRequest<?> request,
```
dsmiley, line 1099: it's not clear to me what is gained by refactoring out this method.

dsmiley (changes requested):

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseCloudSolrClient.java
```diff
@@ -862,15 +1011,32 @@ public RouteException(ErrorCode errorCode, NamedList<Throwable> throwables, Map<
...
+    return apiFuture;
   }
 
   /**
    * As this class doesn't watch external collections on the client side,
    * there's a chance that the request will fail due to cached stale state,
    * which means the state must be refreshed from ZK and retried.
    */
-  protected NamedList<Object> requestWithRetryOnStaleState(@SuppressWarnings({"rawtypes"})SolrRequest request, int retryCount, List<String> inputCollections)
+  protected void requestWithRetryOnStaleState(SolrRequest<?> request,
```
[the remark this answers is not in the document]
rishisankar: > This method signature has become too hard to understand. Taking both a CompletableFuture "apiFuture" and AtomicReference to a CF "currentFuture" is too complicated. Can you please devise some other scheme? Just one CompletableFuture, or perhaps return the CF if that's sensible. I'm sure you had your reasons, but the complexity factor for me is just too high.

Hmm the issue is that the user should receive a future before the request and subsequent retries are processed. In that case, the apiFuture needs to be created before the recursive method -- but the recursive method also needs to have access to the apiFuture so that it can set the response/error once requests/retries are complete. So I think that at the very least, apiFuture probably needs to be an arg in the method signature.

I can think of a few other options for the currentFuture AtomicReference though --
* The AtomicReference is only used for cancelling,  i.e. from makeRequest
```
    apiFuture.exceptionally((error) -> {
      if (apiFuture.isCancelled()) {
        currentFuture.get().cancel(true);
      }
      return null;
    });
```
So perhaps, it may be easier to understand if it's changed to a `AtomicReference<Runnable>` which stores a lambda exp that cancels the current request's future.
* Another option is to ditch currentFuture altogether, and add `apiFuture.exceptionally` statements each time a request retry is sent out to take care of cancelling that specific request's future. Only difference is there will be one callback per request made instead of just one overall; however, all but the most recent callback will be no-ops as their corresponding request futures will have already failed (and spawned a retry request). And I don't imagine there'd be too many retries so the number of callbacks shouldn't cause any noticeable slowdown during cancellation.


Do you think either of these would help with the code readability?

<br/>

> You might also consider changing the retry approach to switch from recursion, to a special exception with a retryCount so the caller can do the retry. I'm just throwing this out there as an idea that _might_ make the overall code easier to understand -- not sure.

Might be misunderstanding your suggestion, but assuming a request is made asynchronously, the (api) future would be returned to the user before the request (and potential request retries) are completed. Thus, I don't think the caller can do the retry (as the caller will have finished execution before a request fails and a retry is decided upon).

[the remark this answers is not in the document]
dsmiley: > Another option is to ditch currentFuture altogether, and add apiFuture.exceptionally statements each time a request retry is sent out to take care of cancelling that specific request's future. Only difference is there will be one callback per request made instead of just one overall; however, all but the most recent callback will be no-ops as their corresponding request futures will have already failed (and spawned a retry request). And I don't imagine there'd be too many retries so the number of callbacks shouldn't cause any noticeable slowdown during cancellation.

That sounds like an acceptable trade-off.  A comment could point out the imperfection.

> Might be misunderstanding your suggestion, but assuming a request is made asynchronously, the (api) future would be returned to the user before the request (and potential request retries) are completed. Thus, I don't think the caller can do the retry (as the caller will have finished execution before a request fails and a retry is decided upon).

By "the caller", I still meant internally.  Put differently, imagine if the implementation was iterative instead of recursive before you came along to add async.  Maybe that would lead to something easier to understand/maintain? If not then nevermind.  It seems you've ruled it out so okay.

solr/solrj/src/java/org/apache/solr/client/solrj/SolrRequest.java
rishisankar: If internalFuture completes exceptionally, apiFuture will complete exceptionally with a `CompletionException` (with internalFuture's exception as its cause). That being said, I don't really like the `CompletionException` wrapper around the actual exception, so I'll revert to using whenComplete instead of thenApply.

rishisankar: If internalFuture completes exceptionally, apiFuture will complete exceptionally with a `CompletionException` (with internalFuture's exception as its cause). That being said, I think it's better to not have a CompletionException wrapper around the actual exception, so I'll revert to using whenComplete instead of thenApply.

solr/core/src/java/org/apache/solr/handler/component/HttpShardHandler.java
```diff
@@ -156,36 +155,31 @@ public void submit(final ShardRequest sreq, final String shard, final Modifiable
...
-      }
-
-      public void onFailure(Throwable throwable) {
+      } else if (!future.isCancelled()) {
         ssr.elapsedTime = TimeUnit.MILLISECONDS.convert(System.nanoTime() - startTime, TimeUnit.NANOSECONDS);
         srsp.setException(throwable);
         if (throwable instanceof SolrException) {
           srsp.setResponseCode(((SolrException) throwable).code());
         }
         responses.add(srsp);
```
dsmiley, line 178: I was going to say:  It should now be straight-forward to call responses.add(srsp) and ssr.elapsedTime = ... in one spot instead of duplicated for both successful and unsuccessful.  BUT... the other scenario is that the future was cancelled, which won't satisfy either the success/failure conditions.  So we do nothing.  Hmmm.  I guess it's okay; not sure if you thought of this.

dsmiley (changes requested):
Your last update was better Rishi; it addressed my concern about complexity of that method.

I think there's a relatively significant bit of feedback from a previous review still pending -- BaseCloudSolrClient.directUpdate.  I'll ask it in a different way:  Do we even need a boolean isAsyncRequest in directUpdate?  Just always do async; one code-path.  If we have an underlying HttpClient that support async, then it will work.  If we don't, then we have an executor we can pass.  If we don't have that, we can pass/use a trivial "direct Executor".  The goal being reduce the multiple code paths, which add complexity.

jdyer1:
Superseded by PR 2402

itygh:

PR #1773, opened 2020-08-21
SOLR-14773 Add LukeRequestHandler to the Solr Ref Guide

epugh:
To replace the old https://cwiki.apache.org/confluence/display/solr/LukeRequestHandler page, and eventually provide a link to the Luke application's getting started page!

### DIFF
--- a/solr/solr-ref-guide/src/luke-request-handler.adoc
+++ b/solr/solr-ref-guide/src/luke-request-handler.adoc
@@ -1,0 +1,77 @@
+= Luke Request Handler
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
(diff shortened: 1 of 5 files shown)


ctargett:
Eric, the content is good for the new page (I've long wanted to make a page for each request handler), but can you discuss your idea to put the new page under the UI section? While the handler is used by the UI, I'm not following how it fits in that overall section since users would use it more like an API.

When I've thought about making a page for each handler, I thought most of them would probably end up under the Implicit RequestHandlers page itself. (I'm also working on an extensive re-org of the entire Guide, so that's feeding some of my doubt on this too).

epugh:
Thanks for looking at this @ctargett (and I hope your vacation was good).   

I modeled the `luke-request-handler.adoc` page on the `mbean-request-handler.adoc` page, which is also linked from the UI page (maybe?)...  

However, now I'm thinking that maybe it should be under `monitoring-solr.adoc` instead?

ctargett:
The MBean Request Handler page is under Monitoring Solr. That makes sense - you look at MBeans as a way of monitoring stats. I quickly searched the source files and don't see any links from any of the UI pages to the MBean Handler page, FWIW.

Since one uses the Luke handler to inspect the schema, should it perhaps go somewhere under `documents-fields-and-schema-design.adoc`? WDYT? A link from the page about the Analysis screen (`analysis-screen.adoc`), which uses the Luke handler, to the new page might also be helpful if you want.

solr/solr-ref-guide/src/luke-request-handler.adoc
```diff
@@ -0,0 +1,77 @@
...
+[source,bash]
+----
+bin/solr start -e techproducts
+----
+
+To return summary information about the index:
+
+[source,text]
+ttp://localhost:8983/solr/techproducts/admin/luke?numTerms=0
+
```
ctargett, line 56: Missing starting 'h' for the URL.

epugh:
@ctargett do you think this is ready to go?

PR #1775, opened 2020-08-23
SOLR-14767 : fix long field parsing from string

apoorvprecisely:
# Description

Int and Long fields when passed as a string should consider all possible value formats and not throw NumberFormatException

# Solution

Use Double.parse as a fallback when it fails

# Tests

added a test to index with values as `42.2` which should not cause NumberFormatException

# Checklist

### DIFF
--- a/solr/CHANGES.txt
+++ b/solr/CHANGES.txt
@@ -251,6 +251,9 @@ Bug Fixes
 * SOLR-14850: Fix ExactStatsCache NullPointerException when shards.tolerant=true.
   (Eugene Tenkaev via ab)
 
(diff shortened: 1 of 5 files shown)


yonik:
These changes aren't correct...  parsing a long as a double (and then converting back to a long) will lose data.  A double cannot represent all integral values that a long can.  Parsing as a double should only be done as a backup if long parsing fails first.

apoorvprecisely:
will make changes, will use double parsing only as a fallback

munendrasn:
I have included changes for TrieFields and modified to parse as Float for Int fields. Probably, we could call toNativeType(val) from createField to avoid duplication but will defer that change to the future (unless someone else prefers that approach too)

PR #1776, opened 2020-08-24
LUCENE-9470: make TestXYMultiPolygonShapeQueries more resilient for CONTAINS queries

iverase:
This test sporadically fail due to numerical errors in the GeoUtils.orient method. This PR changes the way we check for CONTAINS relationship to prevent those errors.

@nknize can you take a look?


nknize (approved):
LGTM!  Thx @iverase

PR #1777, opened 2020-08-24
LUCENE-9477: Don't leave potentially broken segments file behind

s1monw:
If we fail to rollback an already renamed pending segments file during
commit due to a failure in directory syncing we might not fully roll back
to a proper state if we hit a failure during rollback which leaves the index
in a broken state. This is a best effort approach to remove the renamed file
in the case of a failure during sync.


lucene/core/src/java/org/apache/lucene/index/SegmentInfos.java
```diff
@@ -836,16 +835,24 @@ final String finishCommit(Directory dir) throws IOException {
...
+    boolean successRenameAndSync = false;
     final String dest;
     try {
       final String src = IndexFileNames.fileNameFromGeneration(IndexFileNames.PENDING_SEGMENTS, "", generation);
       dest = IndexFileNames.fileNameFromGeneration(IndexFileNames.SEGMENTS, "", generation);
       dir.rename(src, dest);
-      dir.syncMetaData();
-      success = true;
+      try {
+        dir.syncMetaData();
```
mikemccand, line 845: Our wild randomized testing infra was throwing an exception, rarely, in `syncMetaData`?

PR #1779, opened 2020-08-24
LUCENE-9478: Prevent DWPTDeleteQueue from referencing itself and leaking memory

s1monw:
In LUCENE-9304 we introduced some fixes that unfortunately hold on to the previous
DWPTDeleteQueue which is essentially leaking IW memory and cause applications to fail.
This fixes the memory leak and adds a test to ensure its not leaking memory.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/DocumentsWriterDeleteQueue.java
+++ b/lucene/core/src/java/org/apache/lucene/index/DocumentsWriterDeleteQueue.java
@@ -573,6 +573,8 @@ final class DocumentsWriterDeleteQueue implements Accountable, Closeable {
     }
   }
 
+  // we use a static method to get this lambda since we previously introduced a memory leak since it would
+  // implicitly reference this.nextSeqNo which holds on to this del queue. see LUCENE-9478 for reference
   private static LongSupplier getPrevMaxSeqIdSupplier(AtomicLong nextSeqNo) {
     return () -> nextSeqNo.get() - 1;
   }
--- a/lucene/core/src/java/org/apache/lucene/index/DocumentsWriterDeleteQueue.java
+++ b/lucene/core/src/java/org/apache/lucene/index/DocumentsWriterDeleteQueue.java
@@ -573,6 +573,10 @@ final class DocumentsWriterDeleteQueue implements Accountable, Closeable {
     }
   }
 
+  private static LongSupplier getPrevMaxSeqIdSupplier(AtomicLong nextSeqNo) {
+    return () -> nextSeqNo.get() - 1;
+  }
+
   /**
    * Advances the queue to the next queue on flush. This carries over the the generation to the next queue and
    * set the {@link #getMaxSeqNo()} based on the given maxNumPendingOps. This method can only be called once, subsequently
@@ -593,7 +597,7 @@ final class DocumentsWriterDeleteQueue implements Accountable, Closeable {
     return new DocumentsWriterDeleteQueue(infoStream, generation + 1, seqNo + 1,
         // don't pass ::getMaxCompletedSeqNo here b/c otherwise we keep an reference to this queue
         // and this will be a memory leak since the queues can't be GCed
-        () -> nextSeqNo.get() - 1);
+        getPrevMaxSeqIdSupplier(nextSeqNo));
 
   }
 
--- a/lucene/core/src/test/org/apache/lucene/index/TestDocumentsWriterDeleteQueue.java
+++ b/lucene/core/src/test/org/apache/lucene/index/TestDocumentsWriterDeleteQueue.java
@@ -16,6 +16,7 @@
  */
 package org.apache.lucene.index;
 
+import java.lang.ref.WeakReference;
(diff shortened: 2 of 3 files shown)


s1monw:
@uschindler can you please take a look at the test @mikemccand you should also look please

lucene/core/src/test/org/apache/lucene/index/TestDocumentsWriterDeleteQueue.java
```diff
@@ -36,6 +37,26 @@
  */
 public class TestDocumentsWriterDeleteQueue extends LuceneTestCase {
 
+
+  public void testAdvanceReferencesOriginal() {
+    WeakAndNext weakAndNext = new WeakAndNext();
+    DocumentsWriterDeleteQueue next = weakAndNext.next;
+    assertNotNull(next);
+    System.gc();
```
s1monw, line 45: this is terrible but I don't know of a better way of doing this. this at least works on my machine :) and would have found the issue

s1monw: @uschindler @dweiss maybe you have better ideas

lucene/core/src/java/org/apache/lucene/index/DocumentsWriterDeleteQueue.java
```diff
@@ -573,6 +573,10 @@ long getMaxCompletedSeqNo() {
     }
   }
 
+  private static LongSupplier getPrevMaxSeqIdSupplier(AtomicLong nextSeqNo) {
```
dweiss, line 576: Maybe add a comment to this method referencing the issue, Simon?

lucene/core/src/test/org/apache/lucene/index/TestDocumentsWriterDeleteQueue.java
dweiss, line 45: I checked openjdk sources and they use System.gc() with a similar expectation that it will always try to flush all weak refs before returning, so this seems fine. The test itself could maybe be made more explicit by utilizing an explicit ReferenceQueue rather than asserting the reference has been cleared but this is a matter of taste - if this works, it will in both cases.

lucene/core/src/java/org/apache/lucene/index/DocumentsWriterDeleteQueue.java
s1monw: done

lucene/core/src/test/org/apache/lucene/index/TestDocumentsWriterDeleteQueue.java
s1monw: cool! I think the simpler the better

uschindler: I would also use a ReferenceQueue to do the check. The weak.get() has a bad taste to me.
I did not check that System.gc() tries to cleanup weak refs, but if it does and we can rely on this, I am fine.

dnhatn:
> Can we bound the expected size of the leak? It is only the tail of the last delete queue on full flush? So you must have a number of deletes pending at full flush (commit, refresh) for this to matter?

Each delete queue would leak around 500 bytes.

uschindler (approved):
I am OK with this code.

I would really like to see the lambda methods generated by javac. I will do a javap call before and after your patch.

IMHO, the test is fine, but it only tests if there's a reference to `this` in order to access `nextSeqNo`.

uschindler:
This is a limitation of the Java Compiler javac. Whenever you access an instance field from a class (final or not doe snot matter), it created a lamda bount to this. The lambda method then does a getfield and invokes method. Here is the Bytecode of the lambda before this change:

```
  private long lambda$advanceQueue$1();
    Code:
       0: aload_0
       1: getfield      #20                 // Field nextSeqNo:Ljava/util/concurrent/atomic/AtomicLong;
       4: invokevirtual #75                 // Method java/util/concurrent/atomic/AtomicLong.get:()J
       7: lconst_1
       8: lsub
       9: lreturn
```

As you see it's a non-static method. The generated lambda therefor has a reference to this (methodhandle bound to `this`).

After your patch, the lambda is static:

```
  private static long lambda$getPrevMaxSeqIdSupplier$1(java.util.concurrent.atomic.AtomicLong);
    Code:
       0: aload_0
       1: invokevirtual #75                 // Method java/util/concurrent/atomic/AtomicLong.get:()J
       4: lconst_1
       5: lsub
       6: lreturn
```

The generated anonymous class will have just bw bound tothe AtomicLong and not `this`.

By analyzing the code we see that issue is fixed, test is not necessarily needed, but it's hard to check the bytecode :-)

+1 for the fix!

wojtek2kdev:
Hello, I'm newbie in Apache Lucene-Solr. Can someone explain me this bug as for kid? How can I reproduce it on my computer? I want to start participate by bug fixing, but at first I wanna check how previous issues was patched. Thanks in advance!

PR #1789, opened 2020-08-26
LUCENE-9484: Allow sorting an index after the fact

s1monw:
Today we need to decide on an index sorting before we create the index.
In some situations it might make a lot of sense to sort an index afterwards
when the index doesn't change anymore or to compress older indices.
This commit adds the ability to wrap readers from an unsorted index and merge it
into a sorted index by using IW#addIndices.


msokolov:
It's nice that this change is so straightforward. It makes me realize I don't know what happens today if we specify an index Sort and then open an existing index that is not sorted. Do we throw an error? Should we instead provide a sorted view on the index that can be used to rewrite it?

s1monw:
> It's nice that this change is so straightforward. It makes me realize I don't know what happens today if we specify an index Sort and then open an existing index that is not sorted. Do we throw an error? Should we instead provide a sorted view on the index that can be used to rewrite it?

yes we fail if you do that. I don't think we should do any magic here. rewriting can be very costly and taking lots of space. I think failing is the right thing to do.

jpountz (approved):
I didn't remember we had kept this SortingLeafReader, which was the old way we implemented index sorting by wrapping readers with it at merge time. I can think of a couple enhancements like making it a SortingCodecReader rather than a SortingLeafReader but this looks like a nice incremental step and looks simple enough. So +1.

s1monw:
@jpountz I pushed some improvements and inherited from CodecReader, can you take another look

lucene/core/src/java/org/apache/lucene/index/SortingCodecReader.java
```diff
@@ -1090,20 +1150,64 @@ public Bits getLiveDocs() {
...
+  public PointsReader getPointsReader() {
+    return new PointsReader() {
+      @Override
+      public void checkIntegrity() throws IOException {
+        in.getPointsReader().checkIntegrity();
+      }
+
+      @Override
+      public PointValues getValues(String field) throws IOException {
+        final PointValues inPointValues = in.getPointValues(field);
```
jpountz, line 1162: IMO we should delegate to `in#getPointsReader()` as the contract is slightly different on the reader and the points reader (the reader returns null if the field was not indexed with points while the points reader fails).

lucene/core/src/java/org/apache/lucene/index/SortingCodecReader.java
```diff
@@ -1090,20 +1150,64 @@ public Bits getLiveDocs() {
...
+        if (inPointValues == null) {
```
jpountz, line 1163: it can never be null since getValues may only be called on fields that were indexed with points on points readers

lucene/core/src/java/org/apache/lucene/index/SortingCodecReader.java
```diff
@@ -1090,20 +1150,64 @@ public Bits getLiveDocs() {
...
   private final Map<String,CachedNumericDVs> cachedNorms = new HashMap<>();
 
   @Override
-  public NumericDocValues getNormValues(String field) throws IOException {
-    final NumericDocValues oldNorms = in.getNormValues(field);
+  public NormsProducer getNormsReader() {
+    return new NormsProducer() {
+      @Override
+      public NumericDocValues getNorms(FieldInfo field) throws IOException {
+        final NumericDocValues oldNorms = in.getNormValues(field.name);
```
jpountz, line 1189: likewise, delegate to in#getNormsReader here?

jpountz (changes requested):
I think we should make sure to delegate to the points/dv/fields/... reader of the wrapped reader rather than to the wrapped reader itself (see some inline comments, I didn't tag all instances of this).

s1monw:
@jpountz I applied your comments and moved the inner classes from the reader to the classes where they are mainly used. We should have done that earlier then we would have not kept this unused class. Now that reader only holds private inner classes.

lucene/core/src/java/org/apache/lucene/index/SortingCodecReader.java
```diff
@@ -0,0 +1,514 @@
...
+  public StoredFieldsReader getFieldsReader() {
+    StoredFieldsReader delegate = in.getFieldsReader();
+    return newStoredFieldsReader(delegate);
+  }
+
+  private StoredFieldsReader newStoredFieldsReader(StoredFieldsReader delegate) {
+    return new StoredFieldsReader() {
+      @Override
+      public void visitDocument(int docID, StoredFieldVisitor visitor) throws IOException {
+        in.document(docMap.newToOld(docID), visitor);
```
jpountz, line 255: can you call `in.visitDocument` instead?

lucene/core/src/java/org/apache/lucene/index/NumericDocValuesWriter.java
```diff
@@ -165,4 +166,62 @@ public long longValue() {
...
+    }
+
+    @Override
+    public long longValue() {
+      return dvs.values[docID];
+    }
+
+    @Override
+    public long cost() {
+      return dvs.docsWithField.cardinality();
```
jpountz, line 214: and cache the cost?

lucene/core/src/java/org/apache/lucene/index/BinaryDocValuesWriter.java
```diff
@@ -202,4 +203,63 @@ public BytesRef binaryValue() {
...
+    }
+
+    @Override
+    public int advance(int target) {
+      throw new UnsupportedOperationException("use nextDoc instead");
+    }
+
+    @Override
+    public boolean advanceExact(int target) throws IOException {
+      throw new UnsupportedOperationException("use nextDoc instead");
```
jpountz, line 238: I just realized reading other comments that this method might be required when configuring index sorting on a binary field?

jpountz (approved):

lucene/core/src/java/org/apache/lucene/index/BinaryDocValuesWriter.java
s1monw: many others are needed but afaik if you wanna sort on a binary field it needs to be `SortedDocValues` and not `BinaryDocValues` - at least I wasn't able to make any test fail here..

jpountz: :+1:

PR #1791, opened 2020-08-27
LUCENE-9482: Fix deletion count error message

linwenyao:
`throw new CorruptIndexException("invalid deletion count: " + softDelCount + delCount + " vs maxDoc=" + info.maxDoc(), input);`
It needs to be changed to:
`throw new CorruptIndexException("invalid deletion count: " + (softDelCount + delCount) + " vs maxDoc=" + info.maxDoc(), input);`

The value of "softDelCount + delCount" must be calculated first.


dnhatn (approved):
LGTM. Thanks @linwenyao.

PR #1798, opened 2020-08-28
Remove ant Github Action

tflobbe:
`ant` no longer in master branch


tflobbe:
We have an action for gradle that's running `./gradlew precommit`. See https://github.com/apache/lucene-solr/blob/master/.github/workflows/gradle-precommit.yml#L24

dweiss:
Thank you.

PR #1804, opened 2020-08-30
SOLR-14151: Bug fixes

noblepaul:


solr/core/src/java/org/apache/solr/schema/IndexSchema.java
```diff
@@ -190,10 +197,41 @@ public IndexSchema(String name, InputSource is, Version luceneVersion, SolrResou
...
+    if(coreContainer != null) {
+     cores.forEach((coreName, id) -> {
+       try(SolrCore core = coreContainer.getCore(coreName, id)){
+         if(core != null) {
+           core.fetchLatestSchema();
+         }
+       }
+     });
+    }
+  }
```
tflobbe, line 223: Each IndexSchema object gets a `PackageListeningClassLoader` with a listener that, on change, triggers a reload of the schema in all cores?

noblepaul: This happens only if you share the schema. If schema is not shared it only results in reloading schema in that one core

noblepaul:
please refer to #1815

PR #1805, opened 2020-08-30
SOLR-14540: example of drill sideway block join facets

mkhludnev:
Contributing example of block join drill-seideway facets to Ref Guide


ctargett:
This PR is out of date since the repo change but I've got a new PR up in apache/solr#262 so I'll close this in favor of the other one.

PR #1806, opened 2020-08-30
LUCENE-9475: Switch validateSourcePatterns away from Ant legacy

uschindler:
This PR changes the legacy setp of the checkSourcePatterns task to directly execute the checker code in the running Gradle VM.

Because of Ant compatibility we preserved the old code using Ant's Groovy task. So it was executing Groovy shell inside Groovy. The code changes are minimal, further improvements are coming later.

This task has no outputs, so it's declared to run always. We may improve this, if we make the checked files and patterns a task input. Quick tests on different branches showed that task works.

This PR fixes the stupid warning caused by the groovy-inside-groovy:

```
> Task :validateSourcePatterns
Trying to override old definition of task fileScanner
```


dweiss (approved):
Thanks Uwe.

dweiss:
Mike Drob did some work on inputs/ outputs here, it was too early though (ant compatibility).
https://github.com/apache/lucene-solr/pull/1441

The outputs can be empty (but declared). The inputs as per usual - file collection. Should work out of the box.

uschindler:
Oh haven't seen that PR. I was just annoyed by the warning.

uschindler:
I wanted to change the task in the same way after all. 👍🏻

uschindler:
Hi I converted this PR to a draft, because I am incorporating changes from #1441 (which I closed, as outdated).

uschindler:
I think it's now ready. I did not change the inner closures to be real class members. The most important thing to me was to make the inputs declared and the outputs always uptodate, so the task only runs when inputs change. The temporary file by @madrob was not needed (the trick is `outputs.uptodateWhen{true}` (with no outputs, gradle always executes task - well known to the forbiddenapis policeman, trick is userd there, too).

uschindler:
In case you ask: Making is a class extends DefaultTask allows to better declare inputs/outputs. I am not a fan of manually adding them. This code is much more readable (because of annotations).

gradle/validation/validate-source-patterns.gradle
```diff
@@ -15,27 +15,233 @@
...
+  }
+}
+
+configure(rootProject) {
+  task("validateSourcePatterns", type: ValidateSourcePatternsTask) {
+    group = 'Verification'
+    description = 'Validate Source Patterns'
+    
+    // this task has no outputs, so it's always uptodate (if inputs don't change).
+    outputs.upToDateWhen { true }
```
dweiss, line 38: I would move it to the class's constructor? Then inputs and outputs are consistently in the same code.

gradle/validation/validate-source-patterns.gradle
```diff
@@ -15,27 +15,233 @@
...
   }
+}
+
+class ValidateSourcePatternsTask extends DefaultTask {
+
+  @InputFiles
+  final ConfigurableFileTree sourceFiles = project.fileTree(project.rootDir) {
+    [
```
dweiss, line 46: And this I'd move to the task declaration block (so that it's not hardcoded but configured :)

dweiss (approved):
I'm fine with it as it is - just left some hints that I think would make the separation between task class vs. task configuration better?

uschindler:
Thanks Dawid,
I will merge this soon. Robert's issue about `var` should be decided later. In #1802 I will add an exclude just for him.

uschindler:
By the way: Due to my changes, also `gradle/**`` is now checked by validateSourcePatterns. I also added the `.gradle` file extension (you might have noticed).

dweiss (approved):
Thanks Uwe!

PR #1808, opened 2020-08-31
LUCENE-9492: Fix beasting to accept also task names in form ":project…

uschindler:
…:beast"; fix bug with default value rejected by Integer.parseInt


dweiss (commented):
Thanks Uwe.

PR #1813, opened 2020-09-01
SOLR-14613: No new APIs. use the existing APIs

noblepaul:
WIP , do not commit. Untested

We are spending too much time to create and manage a new set of APIs instead of solving the real problem at hand

What does this PR do

- No new interfaces. Uses the existing interface
- Supports configuring `assign-strategy` in `clusterprops.json` . 
- Supports packages


tflobbe:
> Supports configuring assign-strategy in clusterprops.json .

@murblanc recently started [a thread](http://mail-archives.apache.org/mod_mbox/lucene-dev/202008.mbox/%3cCAGhHkK4MCrja3xS25FQOwMzmLQWNm-EN0PgLTvRR0ahkGqip7Q@mail.gmail.com%3e) in the dev list to discuss about configuration location. Can we try to resolve that first before we start adding configuration in different places?

sigram:
This approach was something that I suggested for branch_8x only, and only because the Policy-based placements didn't work and I needed a critical fix. This approach has minimal impact on the existing API and configs so it still may be a good workaround to implement in 8x. However, for master/trunk we need a better design anyway.

noblepaul:
> However, for master/trunk we need a better design anyway.

I would say that most users are actually happy if they do not have to implement a bunch of java interfaces, package a jar add it to Solr to just add a replica in the right place. Honestly nobody will do that. People would rather happily do the computation in client code and send that information as a input. If a DSL is available, they may use it. If not this will be of no interest to most users

sigram:
> Honestly nobody will do that. People would rather happily do the computation in client code

Really? This is like saying that we only ever need collection admin APIs and we don't need any autoscaling. That may be the case for some people, but not for others. Also, not everybody would go and implement their own plugin, instead they would likely use someone else's plugin that is known to work for a particular scenario.

I think the opposite of what you said is true - the expectation is that Solr will behave in a sane way and automatically do what's needed to "properly" use available resources, without too much manual handholding. The problem is of course that the definition of "properly" depends on the point of view and the scale of the problem - that's why we're having the discussion about plugins at all.

Take Salesforce's case for example: they had to hack around the existing AssignStrategy to implement their own, because the default ones didn't work for them. In the process they had to learn and mess around with a lot of Solr internals. It would have been clearly better if they could just implement it as an installable plugin, using an abstract API that doesn't expose too many Solr internals.

noblepaul:
>Really? This is like saying that we only ever need collection admin APIs and we don't need any autoscaling. 

I don't think I made myself clear.

Users would definitely like Solr to place the replicas correctly. But if it means implementing a some plugin in java and packaging it in a jar & deploying it in their cluster, they would rather not do it.
If it's as easy as writing down some DSL, they may use it

solr/core/src/java/org/apache/solr/cloud/api/collections/Assign.java
```diff
@@ -551,11 +555,29 @@ public AssignStrategyFactory(SolrCloudManager solrCloudManager) {
...
     }
 
+    @SuppressWarnings("unchecked")
     public AssignStrategy create(ClusterState clusterState, CloudConfig cloudConfig, DocCollection collection) throws IOException, InterruptedException {
       @SuppressWarnings({"unchecked", "rawtypes"})
       List<Map> ruleMaps = (List<Map>) collection.get("rule");
       @SuppressWarnings({"rawtypes"})
       List snitches = (List) collection.get(SNITCH);
+      Map<String, Object> props = solrCloudManager.getClusterStateProvider().getClusterProperties();
+      Map<String,Object> assignInfo = (Map<String, Object>) props.get("assign-strategy");
```
murblanc, line 565: Do you have code (and curl) examples of how the `assign-strategy` property or similar ones are stored in `clusterprops.json`?

As I'm switching my PR to not use `solr.xml` but instead use `clusterprops.json`, I not only need to retrieve the plugin class name if any, but also associated configuration values whose names/keys and types are unknown since they are plugin specific.

If I copy the code as written here, I assume an iteration over `assignInfo` would reveal all configured values, want to make sure this makes sense.

noblepaul: We already have an endpoint called  `/api/cluster/plugin`  GET/POST

Please refer to [SOLR-14404](https://issues.apache.org/jira/browse/SOLR-14404) for details

it can look as follows

```
curl -X POST -H 'Content-type:application/json' --data-binary '
{
  "set-assign-strategy": {
  "class": "full.ClassName", 
  "key1" : "val1"
  }
}' http://localhost:8983/api/cluster/plugins
```

noblepaul:
This was created as a PoC. It has outlived its purpose, closing....

PR #1815, opened 2020-09-01
SOLR-14151: Bug fixes

noblepaul:


solr/core/src/java/org/apache/solr/core/ConfigSetService.java
```diff
@@ -81,8 +81,7 @@ public final ConfigSet loadConfigSet(CoreDescriptor dcore) {
               ) ? false: true;
 
       SolrConfig solrConfig = createSolrConfig(dcore, coreLoader, trusted);
-      IndexSchema indexSchema = createIndexSchema(dcore, solrConfig, false);
-      return new ConfigSet(configSetName(dcore), solrConfig, force -> indexSchema, properties, trusted);
+      return new ConfigSet(configSetName(dcore), solrConfig, force -> createIndexSchema(dcore, solrConfig, force), properties, trusted);
```
tflobbe, line 84: Does this mean that every call to `configSet.getIndexSchema()` will create a new schema object?

tflobbe (commented):
Noble, could you be more specific about the bugs that this is fixing? Looking at PR, looks like the schema creation now happens asynchronously on reloads, but looking at the code, it'll still be created as part of the reload, right? The reload now also can be async, but you never use that?
This move to async doesn't change the schema/configset API contract, right (because the creation is still part of the same call, just delayed)? i.e. are there tests that used chnange-and-validate, those should still be valid IIUC.

solr/core/src/java/org/apache/solr/core/ConfigSetService.java
noblepaul: Yes.

solr/core/src/java/org/apache/solr/core/CoreContainer.java
```diff
@@ -1582,6 +1582,13 @@ private CoreDescriptor reloadCoreDescriptor(CoreDescriptor oldDesc) {
   public void reload(String name) {
     reload(name, null);
   }
+  public void reload(String name, UUID coreId, boolean async) {
+    if(async) {
+      runAsync(() -> reload(name, coreId));
+    } else {
+      reload(name, coreId);
+    }
+  }
```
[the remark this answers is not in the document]
noblepaul: It's a generic method. I thought I would use it. But there are bugs in our core reloading. So, if I use asyn reload, some tests  fail

noblepaul:
> Noble, could you be more specific about the bugs that this is fixing? 

I'm not sure exactly what the bug is `TestBulkSchemaConcurrent` was failing consistently after the original commit.So, I ensured that it passes consistently

PR #1817, opened 2020-09-01
Improve how Asserting* classes handle singleton doc values.

jtibshirani:
Some queries use DocValues.unwrapSingleton to execute different logic for
single-valued doc values. When tests use AssertingCodec or AssertingLeafReader,
unwrapSingleton will never unwrap the doc values, as they don't have the
expected class. So some queries have code paths that are never exercised with
these asserting wrappers.

This change makes sure to preserve the expected classes when creating asserting
doc values.

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -196,6 +196,8 @@ Improvements
 * LUCENE-9446: In BooleanQuery rewrite, always remove MatchAllDocsQuery filter clauses
   when possible. (Julie Tibshirani)
 
+* LUCENE-9501: Improve how Asserting* test classes handle singleton doc values.
+
 Optimizations
 ---------------------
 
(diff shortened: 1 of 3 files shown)


jtibshirani:
I noticed this when debugging a bad bug in `IndexSortSortedNumericDocValuesRangeQuery` that wasn't caught by tests. As a follow-up I plan to make at least one more testing fix, and also fix `IndexSortSortedNumericDocValuesRangeQuery` itself.

jpountz (approved):
This looks good to me.

jtibshirani:
Thanks for the review. Let me know if I should add a changelog entry or issue number (I wasn't sure what to do for smaller test fixes like this).

jpountz:
Good question. The test framework is used by our users so I think it makes sense to mention this change in the changelog.

PR #1819, opened 2020-09-02
SOLR-14782: Document how to unescape for the QueryElevationComponent.

bruno-roustant:


solr/solr-ref-guide/src/the-query-elevation-component.adoc
```diff
@@ -61,7 +61,18 @@ Optionally, in the Query Elevation Component configuration you can also specify
...
 
 `queryFieldType`::
-Specifies which fieldType should be used to analyze the incoming text. For example, it may be appropriate to use a fieldType with a LowerCaseFilter.
+Specifies which fieldType should be used to analyze the incoming text. For example, it may be appropriate to use a fieldType with a LowerCaseFilter. Other example, if you need to unescape backslash-escaped queries, then you can define the fieldType to preprocess with a PatternReplaceCharFilter. Here is the corresponding example of fieldType (traditionally in `schema.xml`):
+
+[source,xml]
+----
+<fieldType name="unescapelowercase" class="solr.TextField">
+  <analyzer>
+    <charFilter class="solr.PatternReplaceCharFilterFactory" pattern="\\(.)" replacement="$1"/>
```
dsmiley, line 70: I think there's a syntax to use such that we only un-escape non-alphanumeric?  If they type a\b (which might be in some identifier) then they probably meant that.

bruno-roustant: Indeed, but do we want to go into this level of detail?
The Lucene StandardQueryParser uses EscapeQuerySyntaxImpl.discardEscapeChar() which unescape everything without distinction except it detects unicode \uxxxx and converts it.
I suggest that we give a basic example of unescaping every backslash. If the user wants more sophisticated logic, then she will design the right pattern or the custom CharFilter.

bruno-roustant: The regex would be "\\([^\p{Alnum}])", but it applies to ASCII alphanum, not internationalized.

bruno-roustant: Actually it is possible to localize the pattern with
"\\\\([^\\p{javaLowerCase}\\p{javaUpperCase}\\p{Digit}])"

bruno-roustant: I added this other pattern as an example.

PR #1820, opened 2020-09-02
LUCENE-9464: Add high(er)-level hit highlighter example that demonstrates and uses low-level components

dweiss:


dweiss:
Ping, @romseygeek ? Let me know if you don't have time to take a look.

lucene/highlighter/src/test/org/apache/lucene/search/matchhighlight/TestMatchHighlighter.java
```diff
@@ -0,0 +1,466 @@
...
+          // possibly highlighted if they matched the query.
+          //
+          // This configures the highlighter so that the FLD_ID field is always returned verbatim,
+          // and FLD_TEXT1 is returned *only if it contained a query match*.
+          MatchHighlighter highlighter =
+              new MatchHighlighter(searcher, analyzer)
+                .appendFieldHighlighter(FieldValueHighlighters.verbatimValue(FLD_ID))
+                .appendFieldHighlighter(FieldValueHighlighters.highlighted(
+                    80 * 3, 1, new PassageFormatter("...", ">", "<"), FLD_TEXT1::equals))
+                .appendFieldHighlighter(FieldValueHighlighters.skipRemaining());
```
romseygeek, line 151: Just out of interest, what happens if you don't add a `skipRemaining()` highlighter to the end?  If it's necessary, maybe we should convert to a Builder pattern and automatically add it to the end of the highlighters list before returning the MatchHighlighter?

romseygeek (approved):
This is great Dawid! I left one question, about always having to add `skipRemaining()`.  

I think it would be interesting to use this to try and build equivalents for the other existing highlighter tests, and in particular to do some benchmarking - I suspect that we can get at least parity on performance with the other highlighters by doing some query rewriting.  I don't have time immediately but will try and get to it before long.

lucene/highlighter/src/test/org/apache/lucene/search/matchhighlight/TestMatchHighlighter.java
dweiss: These field highlighters can apply to more than one field and the highlighter may receive fields it wasn't prepared to process (because they were part of the query). I opted to be explicit - if no field highlighter "accepts" a field, the entire highlighting process fails. The "skip remaining" is essentially a rule that does nothing for any field that wasn't consumed before.

Performance-wise I admit I wasn't too keen on optimizing things too early. This thing, even if slower, just works for me in 
so many more scenarios than any of the approaches I've tried with other highlighters in Lucene. That "two queries at once" example isn't entirely out of thin air... we actually do use something like this to highlight queries originating from different sources.

dweiss:
I've merged this in. If we decide to change it or improve the API, let it be another issue.

PR #1823, opened 2020-09-03
SOLR-14510: Remove deprecations added with BlockMax WAND support

tflobbe:
With BMW support, we added a new method to Text response writers to include in the response if the numFound is exact or not. This PR is to remove the previously existing and deprecated methods and makes the new method abstract to force it's implementation in subclasses. This is intended for master branch only.

### DIFF
--- a/solr/core/src/java/org/apache/solr/response/GeoJSONResponseWriter.java
+++ b/solr/core/src/java/org/apache/solr/response/GeoJSONResponseWriter.java
@@ -294,13 +294,6 @@ class GeoJSONWriter extends JSONWriter {
     }
   }
 
-  @Deprecated
-  @Override
-  public void writeStartDocumentList(String name, 
-      long start, int size, long numFound, Float maxScore) throws IOException {
-    throw new UnsupportedOperationException();
-  }
-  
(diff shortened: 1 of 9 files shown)


anshumg (approved):
Thanks @tflobbe ! LGTM.

It would be useful to add this deprecation information to the Solr Upgrade Notes (CHANGES.txt).

cpoerschke:
> Remove deprecations added with BMW support 

> ... It would be useful to add this deprecation information ...

+1 to make it a clearer what is being deprecated. "BMW support" at first glance made me think of the car manufacturer but no, it's not that but "BlockMax WAND support" instead.

To the reader of the deprecation information, does it matter why the thing that is being removed was deprecated, I wonder? If not then something like _"Remove deprecated writeStartDocumentList variant in TextResponseWriter and its sub-classes."_ could work perhaps, though it's rather long, hmm.

tflobbe:
Added a note to the upgrade section of the reference guide. Let me know what you think

cpoerschke (approved):
LGTM

anshumg (approved):

PR #1824, opened 2020-09-03
LUCENE-9500: Separate the Deflater hack from the Lucene code to a subclass of java.util.zip.Deflater

uschindler:
A factory method allows to return a different instance depending on Java version (for now everything after Java 11 is disallowed).

We can improve this when we exactly know which versions are affected. Once the bug is fixed and Lucene uses a later minimum version, we can easily remove the class.

TODO: We should add a forbiddenapis to disallow direct use of deflater by disallowing its constructor.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/codecs/lucene87/BugfixDeflater_JDK8252739.java
+++ b/lucene/core/src/java/org/apache/lucene/codecs/lucene87/BugfixDeflater_JDK8252739.java
@@ -27,6 +27,26 @@ import java.util.zip.Inflater;
  */
 final class BugfixDeflater_JDK8252739 extends Deflater {
   
+  public static final boolean IS_BUGGY_JDK = detectBuggyJDK();
+
+  /**
+   * Creates a {@link Deflater} instance, which works around JDK-8252739.
+   * <p>
+   * Use this whenever you intend to call {@link #setDictionary(byte[], int, int)} or
+   * {@link #setDictionary(java.nio.ByteBuffer)} on a {@code Deflater}.
+   * */
+  public static Deflater createDeflaterInstance(int level, boolean nowrap, int dictLength) {
+    if (dictLength < 0) {
+      throw new IllegalArgumentException("dictLength must be >= 0");
+    }
+    if (IS_BUGGY_JDK) {
+      return new BugfixDeflater_JDK8252739(level, nowrap, dictLength);
+    } else {
+      return new Deflater(level, nowrap);
+    }
+  }
+  
+  
   private final byte[] dictBytesScratch;
 
   private BugfixDeflater_JDK8252739(int level, boolean nowrap, int dictLength) {
@@ -44,17 +64,6 @@ final class BugfixDeflater_JDK8252739 extends Deflater {
     }
   }
 
-  public static Deflater createDeflaterInstance(int level, boolean nowrap, int dictLength) {
-    if (dictLength < 0) {
-      throw new IllegalArgumentException("dictLength must be >= 0");
-    }
-    if (IS_BUGGY_JDK) {
-      return new BugfixDeflater_JDK8252739(level, nowrap, dictLength);
-    } else {
-      return new Deflater(level, nowrap);
-    }
-  }
-  
   private static boolean detectBuggyJDK() {
     final byte[] testData = new byte[] { 1, 2, 3, 4, 5, 6, 7, 8 };
     final byte[] compressed = new byte[32]; // way enough space
@@ -100,6 +109,4 @@ final class BugfixDeflater_JDK8252739 extends Deflater {
     return false;
   }
   
-  public static final boolean IS_BUGGY_JDK = detectBuggyJDK();
-
 }
--- a/lucene/core/src/java/org/apache/lucene/codecs/lucene87/BugfixDeflater_JDK8252739.java
+++ b/lucene/core/src/java/org/apache/lucene/codecs/lucene87/BugfixDeflater_JDK8252739.java
@@ -22,7 +22,7 @@ import org.apache.lucene.util.Constants;
 
 /**
  * This class is a workaround for JDK bug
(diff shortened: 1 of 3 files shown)


uschindler:
This effectively  reverts @jpountz's hotfix commit, but as he moved classes around in the meantime, I had to do it manually.

lucene/core/src/java/org/apache/lucene/codecs/lucene87/DeflateWithPresetDictCompressionMode.java
```diff
@@ -155,16 +155,15 @@ public Decompressor clone() {
...
 
-    final byte[] dictBytes;
-    final int blockLength;
+    private final int dictLength, blockLength;
     final Deflater compressor;
     byte[] compressed;
     boolean closed;
 
     DeflateWithPresetDictCompressor(int level, int dictLength, int blockLength) {
-      compressor = new Deflater(level, true);
```
dweiss, line 165: Just a thought, really.

If it's a bug that can be probed for (and it can be - see Adrian's repro) then it could as well be a static initialization of a supplier of Deflater instances; if we probe for a buggy JVM, we return the wrapper. If we don't we return the Deflater. This way on non-affected JVMs nothing happens and if we do use the wrapper, we know the JVM is broken.

uschindler: that's also my idea! I am working on probing at the moment. like 128 bytes of dict and 128 bytes of data. If result is correct the factory method i implemented uses the plain Deflater.

uschindler: I added detection in latest commit. Works with 8 bytes :-)

uschindler:
Hi, I added detectction in my latest commit:

```
C:\Users\Uwe Schindler\Projects\lucene\trunk-lusolr1>java -version
openjdk version "11.0.2" 2019-01-15
OpenJDK Runtime Environment 18.9 (build 11.0.2+9)
OpenJDK 64-Bit Server VM 18.9 (build 11.0.2+9, mixed mode)

C:\Users\Uwe Schindler\Projects\lucene\trunk-lusolr1>gradlew -Ptests.verbose=true :lucene:core:test --tests TestLucene87StoredFieldsFormatHighCompression | grep buggy
  2> JDK is buggy: false

```

vs.

```
C:\Users\Uwe Schindler\Projects\lucene\trunk-lusolr1>java -version
openjdk version "13" 2019-09-17
OpenJDK Runtime Environment (build 13+33)
OpenJDK 64-Bit Server VM (build 13+33, mixed mode, sharing)

C:\Users\Uwe Schindler\Projects\lucene\trunk-lusolr1>gradlew -Ptests.verbose=true :lucene:core:test --tests TestLucene87StoredFieldsFormatHighCompression | grep buggy
  2> JDK is buggy: true
```

The `System.err.println()` needs of course be removed before merge :-)

uschindler:
I converted this to a draf, because checks failed with precommit. But this should stay in for doing quick checks by @jpountz like commented before!

lucene/core/src/java/org/apache/lucene/codecs/lucene87/BugfixDeflater_JDK8252739.java
```diff
@@ -0,0 +1,102 @@
...
+    }
+    
+    // all fine
+    return false;
+  }
+  
+  public static final boolean IS_BUGGY_JDK = detectBuggyJDK();
+  
+  static {
+    System.err.println("JDK is buggy: " + IS_BUGGY_JDK);
```
jpountz, line 99: Did you only mean to use this for debugging?

lucene/core/src/java/org/apache/lucene/codecs/lucene87/BugfixDeflater_JDK8252739.java
```diff
@@ -0,0 +1,102 @@
...
+      inflater.end();
+    }
+
+    if (Arrays.equals(testData, restored) == false) {
+      return true;
+    }
+    
+    // all fine
+    return false;
+  }
```
jpountz, line 94: If you confirmed that this fails on at least one buggy JVM, this detection code LGTM.

jpountz (approved):
Thanks @uschindler, very simple, I like it.

lucene/core/src/java/org/apache/lucene/codecs/lucene87/BugfixDeflater_JDK8252739.java
jpountz: oops nevermind I just read your other comment

lucene/core/src/java/org/apache/lucene/codecs/lucene87/BugfixDeflater_JDK8252739.java
jpountz: likewise, nevermind

dweiss (approved):

uschindler:
Hi,
I did further tests (except Java 8, I need to merge first), Results looking fine:
- OpenJDK JDK 11.0.2 reports **non-buggy**
- AdoptJDK 11.0.5 reports **buggy**
- Oracle JDK 11.0.8 reports **buggy**
- JDK 12 reports **non-buggy**
- JDK 13 reports **buggy**
- JDK 14 reports **buggy**

I keep the class for now as package private. Thoughts:
- If Elasticsearch or Solr want to report on startup that the JDK they are using is buggy and they should upgrade (once they have a fix available, a fix for JDK 11 looks promising as this was a regression!), we can make the class public, so startup code has access to the boolean flag
- We may print a Info in the test, so when running tests you get informed about the issue (useful for quick testing a new JDK release)

rmuir (approved):
thanks for adding the detection!

uschindler:
Short info: After backporting to 8.x, of course JDK 8u192 was detected as **non-buggy**.

PR #1825, opened 2020-09-03
SOLR-14828: reduce 'error' logging noise in BaseCloudSolrClient.requestWithRetryOnStaleState

cpoerschke:
https://issues.apache.org/jira/browse/SOLR-14828

### DIFF
--- a/solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseCloudSolrClient.java
+++ b/solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseCloudSolrClient.java
@@ -970,9 +970,6 @@ public abstract class BaseCloudSolrClient extends SolrClient {
               rootCause instanceof SocketException ||
               wasCommError(rootCause));
 
-      log.error("Request to collection {} failed due to ({}) {}, retry={} commError={} errorCode={} ",
-          inputCollections, errorCode, rootCause, retryCount, wasCommError, errorCode);
-
       if (wasCommError
           || (exc instanceof RouteException && (errorCode == 503)) // 404 because the core does not exist 503 service unavailable
         //TODO there are other reasons for 404. We need to change the solr response format from HTML to structured data to know that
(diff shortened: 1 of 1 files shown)


solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseCloudSolrClient.java
```diff
@@ -994,12 +991,15 @@ public RouteException(ErrorCode errorCode, NamedList<Throwable> throwables, Map<
...
           // the state would not have been updated
-          log.info("trying request again");
+          log.info("Request to collection {} failed due to ({}) {}, retry={} commError={} errorCode={} - retrying",
+              inputCollections, errorCode, rootCause, retryCount, wasCommError, errorCode);
           return requestWithRetryOnStaleState(request, retryCount + 1, inputCollections);
         }
       } else {
         log.info("request was not communication error it seems");
       }
+      log.error("Request to collection {} failed due to ({}) {}, retry={} commError={} errorCode={} ",
```
HoustonPutman, line 1001: Since you add ` - retrying` to the info logging above, maybe add something like ` - max retries hit` here

cpoerschke: `max retries hit` is a little tricky to determine exactly since e.g. `commError` and other things influence the logic. But inclusion of `MAX_STALE_RETRIES` in the logging could make it clear(er) when "max retries hit" was likely the case.

HoustonPutman (approved):

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseCloudSolrClient.java
```diff
@@ -994,12 +991,15 @@ public RouteException(ErrorCode errorCode, NamedList<Throwable> throwables, Map<
...
           // the state would not have been updated
-          log.info("trying request again");
+          log.info("Request to collection {} failed due to ({}) {}, retry={} maxRetries={} commError={} errorCode={} - retrying",
+              inputCollections, errorCode, rootCause, retryCount, MAX_STALE_RETRIES, wasCommError, errorCode);
           return requestWithRetryOnStaleState(request, retryCount + 1, inputCollections);
         }
       } else {
         log.info("request was not communication error it seems");
       }
+      log.error("Request to collection {} failed due to ({}) {}, retry={} maxRetries={} commError={} errorCode={} ",
```
cpoerschke, line 1001:
```suggestion
      log.info("Request to collection {} failed due to ({}) {}, retry={} maxRetries={} commError={} errorCode={} ",
```

@noblepaul - is this what you had in mind when you wrote "switch this to INFO" on the https://issues.apache.org/jira/browse/SOLR-14828 ticket? Or would you prefer keeping the logging at line 973 only but as `INFO` instead of `ERROR`?

PR #1828, opened 2020-09-04
LUCENE-9497: add google error prone checks

dweiss:


gradle/validation/error-prone.gradle
```diff
@@ -0,0 +1,148 @@
...
+allprojects { prj ->
+  plugins.withType(JavaPlugin) {
+    prj.apply plugin: 'net.ltgt.errorprone'
+
+    dependencies {
+      errorprone("com.google.errorprone:error_prone_core")
+    }
+
+    tasks.withType(JavaCompile) { task ->
+      options.errorprone.errorproneArgs = [
```
madrob, line 27: I would also like to see `options.errorprone.disableWarningsInGeneratedCode = true`

gradle/validation/error-prone.gradle
```diff
@@ -0,0 +1,148 @@
...
+          // test
+          '-Xep:ExtendingJUnitAssert:OFF',
```
madrob, line 29: Personal style, but I think ```options.errorprone {
disable 'ExtendingJUnitAssert'
}```
is more clear than using `-Xep`?

gradle/validation/error-prone.gradle
dweiss: I suggested the same to Varun.

vthacker: Let's add this ? I probably missed it in my PR

lucene/core/src/java/org/apache/lucene/analysis/CharArrayMap.java
```diff
@@ -523,6 +523,7 @@ public void clear() {
    * @throws NullPointerException
    *           if the given map is <code>null</code>.
    */
+  @SuppressWarnings("ReferenceEquality")
```
vthacker, line 526: This was the example you wanted to try out on how to suppress legitimate warnings of ReferenceEquality ( or any other warnings ) when we start enabling the checks ?

gradle/validation/error-prone.gradle
vthacker: I am okay with either of the two styles. Ideally we'd want this list to get much shorter soon :)

vthacker (approved):
LGTM!

lucene/core/src/java/org/apache/lucene/analysis/CharArrayMap.java
dweiss: Yes, exactly. I wanted the pr to include an example of how this can be done.

gradle/validation/error-prone.gradle
dweiss: Varun - feel free to take this branch (or patch) and roll it out on yours. I didn't intend it to be committed, I just wanted to show what's needed for it to compile and work.

vthacker: sounds good! I'll take the current branch and add two things and then commit the code
1. `options.errorprone.disableWarningsInGeneratedCode = true`
2. CHANGES entry

PR #1830, opened 2020-09-04
LUCENE-9506: Gradle: split validateSourcePatterns into per-project an…

dweiss:


dweiss:
@uschindler you know these checks better - to me I refactored this stuff in the same way as before (and they're running much faster now since they can run in parallel). If you have a spare minute to eyeball though it'd be good.

gradle/validation/validate-source-patterns.gradle
```diff
@@ -29,50 +33,117 @@ buildscript {
...
-        include "lucene/**/*.${it}"
-        include "solr/**/*.${it}"
-        include "dev-tools/**/*.${it}"
-        include "gradle/**/*.${it}"
+    sourceFiles = fileTree(projectDir) {
+      extensions.each{
+        include "*.${it}"
+      }
+
+      // default excludes.
```
madrob, line 76: should the excludes be an input property so that we don't have to repeat them later on root project?

uschindler:
I will check over the weekend. Was too busy today!

gradle/validation/validate-source-patterns.gradle
dweiss: It could be. I didn't have time to clean up everything. The speedup was significant for me anyway (order of magnitude).

dweiss:
No worries. Not very urgent.

gradle/validation/validate-source-patterns.gradle
uschindler: I agree with this. Maybe make a closure out of it, so it can be applied everywhere. I can do this, if you like.

gradle/validation/validate-source-patterns.gradle
```diff
@@ -29,50 +33,117 @@ buildscript {
...
+      exclude "$buildDir/**"
+      exclude '**/dist/**'
+      exclude 'dev-tools/missing-doclet/src/**/*.java' // <-- TODO: remove once we allow "var" on master
+      exclude 'lucene/benchmark/work/**'
```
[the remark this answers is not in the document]
uschindler: But if the code does not fail, mabe remove the exclude (outdated?).

gradle/validation/validate-source-patterns.gradle
```diff
@@ -181,8 +252,13 @@ class ValidateSourcePatternsTask extends DefaultTask {
       }
     }
 
+    ProgressLogger progress = progressLoggerFactory.newOperation(this.class)
+    progress.start(this.name, this.name)
+
     sourceFiles.each{ f ->
-      logger.debug('Scanning source file: {}', f);
+      progress.progress("Scanning ${f.name}")
```
uschindler, line 259: Cool!

gradle/validation/validate-source-patterns.gradle
```diff
@@ -181,8 +252,13 @@ class ValidateSourcePatternsTask extends DefaultTask {
...
+      logger.info('Scanning source file: {}', f);
```
uschindler, line 260: I would keep this as debug or remove after adding the progress meter.
I often try to run gradlew with `--info`, but theis is way too much information.

Do you know how to silence the palantir version-lock checker? I hate that it prints so much information during configuration phase to info.

I'd like to enable `--info` on jenkins, but that's way to much.

uschindler (commented):
In addition:
This occurrence <https://github.com/apache/lucene-solr/blob/bb1afca9686b621ffb763675e514b48fd66670a7/gradle/validation/validate-source-patterns.gradle#L300> should instead test on `project.path.startsWith(':solr')` or similar.

uschindler (changes requested):
Looks fine, left some TODOs.

gradle/validation/validate-source-patterns.gradle
dweiss: Please, go ahead and take over!

gradle/validation/validate-source-patterns.gradle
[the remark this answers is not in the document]
dweiss: Yup.

gradle/validation/validate-source-patterns.gradle
dweiss: The --info switch is verbose. I don't think there's much you can do about it. I was thinking about it myself - perhaps we should devise our own internal logging verbosity level (and log to lifecycle when it's enabled).

uschindler: Info is like ant always was. Except the shitty palantir logging. Can we suppress this? I have all Gradle builds with info logging in jenkins, except Lucene.

dweiss: I don't know how to do that, Uwe.

uschindler: Nevertheless, I'd change ours back to debug (it's really only for debugging). If a violation happens, filename is printed anyways.

But we should change this now to not use relative paths anymore, because it's hard to figure out where file is located. I can take care of that.

dweiss: Hmm... What do you mean by "not use relative paths"?

uschindler: When it reports a violation it's relative to basedir of file collection. This was ok we with root level task, but should be changed now to just print full path.

dweiss: Doesn't the logger carry the information which project this is? This should be sufficient to locate the file in question? I don't mind changing it to an absolute path, but I think project + path should be enough to locate something?

uschindler:
Not sure how this is logged. I will check tomorrow.

uschindler:
Hi, I am reviewing: Actually the whole task does not scan any files, as the "extensions" are wrongly expanded. Should be `**/*.${it}`. I will commit a fix together with many other changes.

uschindler:
Actually for me the splitted tasks take muuuuuch longer to complete: almost 30 seconds. Maybe it's caused by the progress meter.

uschindler:
The other problem is that for ":solr" and ":lucene" it scans files twice, as the fileTree includes again the subprojects!

dweiss:
bq. Actually the whole task does not scan any files

Ah, so this would explain the speedup. :)

Let me reiterate, sorry about the noise.

dweiss:
I'll take a look, but not now - working.

dweiss:
You're right, Uwe - the subproject scan inclusion sucks. I don't know how to avoid it in an elegant way. This could be done with worker API as well but it's not the same - per project tasks are more cache friendly and run in parallel without any special hackery.

I'll give it another thought, thanks for the review and fixes!

uschindler:
Hi,
I was able to prevent gradle from scanning duplicates:

```
// Don't go into subprojects (scanned separately).
childProjects.keySet().each{
  exclude "${it}/**"
}
```

uschindler (approved):

uschindler (approved):

uschindler:
Hi,

this now passes build, but speed is on my laptop (Windows 10) only sligthly faster: 32s before, 24s with this PR (including configuration and without gradle daemon).

Another approach to fix this would be to only put some dirs on the include list (like `src/**`) and don't have so many excludes. The reason why it's the other way round ("scan everything, except..."): The original Ant task was setup like this to just scan all text files and source files splattered around in the ant build for invalid source patterns. It should detect any violation, anywhere (in build files, xml files, source code, resources,... everywhere). So the idea is to scan as much as possible and put outliers on a exclude list.

uschindler:
The reason why the speed difference is so small may be related in the fact that "project configuration" (collecting those file trees, one for each project) take much longer than the actual scanning.

uschindler:
I refactored the config a bit (project-specific excludes moved to separate configure-blocks). I also added `*.md` to the list of scanned files.

Looks ready now.

dweiss:
bq.  32s before, 24s with this PR 

Well, that's 25% faster. The biggest gain should come from running incrementally? If you modify a file in just one of the projects, only that project's task should be re-run.

gradle/validation/validate-source-patterns.gradle
```diff
@@ -29,50 +33,127 @@ buildscript {
...
       }
-      // TODO: For now we don't scan txt / md files, so we
-      // check licenses in top-level folders separately:
-      include '*.txt'
-      include '*/*.txt'
-      include '*.md'
-      include '*/*.md'
-      // excludes:
+      
+      // Don't go into child projects (scanned separately).
```
dweiss, line 77: Yep, that should do it.

gradle/validation/precommit.gradle
```diff
@@ -26,7 +26,6 @@ configure(rootProject) {
     dependsOn ":verifyLocks"
     dependsOn ":versionsPropsAreSorted"
     dependsOn ":checkWorkingCopyClean"
-    dependsOn ":validateSourcePatterns"
```
dweiss, line 29: I think we can leave the dependency on the root task as it depends on other source pattern validations. Or remove the snippet that connects them later.

gradle/validation/validate-source-patterns.gradle
```diff
@@ -29,50 +33,127 @@ buildscript {
...
+      // ourselves :-)
+      exclude 'gradle/validation/validate-source-patterns.gradle'
     }
   }
+
+  task validateSourcePatterns() {
+    group = 'Verification'
+    description = 'Validate Source Patterns'
+
+    // Make it depend on all so-named tasks from subprojects.
```
dweiss, line 137: (here).

dweiss:
I have a suspicion those filetree filters (exclusion filters) are fairly dumb and scan paths that can't be traversed anyway. I'd have to confirm to be sure (which I can't do now). Anyway, I think it's better than before anyway, so let's leave it for later improvements, if necessary.

gradle/validation/precommit.gradle
uschindler: I wanted to make it consistent with the others. Now all are handled the same.

uschindler: Removing the connection in subprojects is not good, as now I can run `gradlew check` in a subproject and it does everything right.

dweiss: Fair enough. But then the top-level one doesn't need to depend on children?

dweiss: Ah, ok. See it now.

gradle/validation/validate-source-patterns.gradle
uschindler: I also want this to be able to run it from root separately. If that's not needed and works automatically, remove it here.

I just made it consistent with other validation tasks

dweiss:
I'm committing it then, ok Uwe?

gradle/validation/validate-source-patterns.gradle
dweiss: Yes, it does work. Like I said - a task without a project qualifier is in fact an invocation of any such task in any subproject. So:
```
> gradlew -p lucene/analysis validateSourcePatterns  --console=plain
...
> Task :lucene:analysis:validateSourcePatterns UP-TO-DATE
> Task :lucene:analysis:phonetic:validateSourcePatterns UP-TO-DATE
> Task :lucene:analysis:opennlp:validateSourcePatterns UP-TO-DATE
> Task :lucene:analysis:stempel:validateSourcePatterns UP-TO-DATE
> Task :lucene:analysis:morfologik:validateSourcePatterns UP-TO-DATE
> Task :lucene:analysis:smartcn:validateSourcePatterns UP-TO-DATE
> Task :lucene:analysis:nori:validateSourcePatterns UP-TO-DATE
> Task :lucene:analysis:kuromoji:validateSourcePatterns UP-TO-DATE
> Task :lucene:analysis:icu:validateSourcePatterns UP-TO-DATE
> Task :lucene:analysis:common:validateSourcePatterns UP-TO-DATE
```

uschindler:
I don't think that's the case. Running the whole thing a second time is quite fast. I remember that all those filters in Maven and Ant are based on the same code pattern (DirectoryScanner from Ant or Maven's Plexus Utils), which has two scanning modes, GM fast and slow. The slow one is dumb but also collects information about files filtered out. At least for includes those are intelligent.

uschindler:
Ok merge it. Thanks.

dweiss:
> I don't think that's the case. Running the whole thing a second time is quite fast. 

Because it's cached in the daemon? No way of telling without actually digging at gradle's code (which I've done before and it's not so scary). Let's leave it. If it's annoying then we'll think of a workaround.

uschindler:
Of course, running without daemon. Configure time is same. Execution is done very fast.

gradle/validation/validate-source-patterns.gradle
uschindler: I just wanted to add this to this dicsussion: https://github.com/gradle/gradle/issues/1010
I am working on that!

(I also figured out that we miss our lockfactory integration test in `:lucene:core`, in ANT it's still there (many JVMs hammering a single directory with lockfactories). We need a task in gradle for this, too. I will open another issue)

uschindler:
I checked source code: Actually Gradle does not use the Ant/Plexus DirectoryScanner, it has a new Java 7 based one. This one handles inclusions and exclusions in the same way and skips exclusions. If it is a directory then it passes `SKIP_SUBTREE`: https://github.com/gradle/gradle/blob/aeea780a34d61620ded70b2ebe9e68dd9bc82158/subprojects/file-collections/src/main/java/org/gradle/api/internal/file/collections/DefaultDirectoryWalker.java#L88-L99

uschindler:
I just pushed one more small change, ready to go!

uschindler:
Thanks!

PR #1831, opened 2020-09-04
SOLR-14749 the scheduler part

sigram:
See PR 1758 for the background on this.


solr/core/src/java/org/apache/solr/cloud/ClusterSingleton.java
```diff
@@ -0,0 +1,55 @@
...
+ * limitations under the License.
+ */
+package org.apache.solr.cloud;
+
+/**
+ * Intended for {@link org.apache.solr.core.CoreContainer} plugins that should be
+ * enabled only one instance per cluster.
+ * <p>Components that implement this interface are always in one of two states:
+ * <ul>
+ *   <li>STOPPED - the default state. The component is idle and does not perform
```
murblanc, line 24: Should we add `STARTING` and `STOPPING` states? Assuming the call to `start()` waits until the plugin has completed its startup (and similarly the call to `stop()` waiting for it to stop) might be expensive, and as implemented `start()` delays the `Overseer` starting in general.

Possibly waiting for `stop()` to complete makes sense (in order to guarantee no two `ClusterSingleton` plugins are running concurrently on the cluster, so maybe state `STOPPING` is not needed), but I'd think we don't need to wait for `start()` to have completed.

solr/core/src/java/org/apache/solr/cloud/Overseer.java
```diff
@@ -775,6 +779,42 @@ private void doCompatCheck(BiConsumer<String, Object> consumer) {
...
+        }
+      }
+    });
+  }
+
+  /**
+   * Stop {@link ClusterSingleton} plugins when we lose leadership.
+   */
+  private void stopClusterSingletons() {
+    PluginBag<SolrRequestHandler> handlers = getCoreContainer().getRequestHandlers();
```
murblanc, line 806: Should we stop currently configured `ClusterSingleton` handlers or rather stop all those we've started? Is the configuration immutable?

solr/core/src/java/org/apache/solr/cluster/scheduler/impl/CompiledSchedule.java
```diff
@@ -0,0 +1,115 @@
...
+    } else {
+      return false;
+    }
+  }
+
+  /**
+   * This setter MUST be invoked after each run.
+   * @param lastRunAt time when the schedule was last run.
+   */
+  void setLastRunAt(Instant lastRunAt) {
```
murblanc, line 112: Unclear if `lastRunAt` is the time the schedule last started or last completed. If completed, can we simplify by not passing an instant and building it here?

solr/core/src/java/org/apache/solr/cluster/scheduler/impl/CompiledSchedule.java
```diff
@@ -0,0 +1,115 @@
...
+
+import org.apache.solr.cluster.scheduler.Schedule;
+import org.apache.solr.common.SolrException;
+import org.apache.solr.util.DateMathParser;
+import org.apache.solr.util.TimeZoneUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * A version of {@link Schedule} where some of the fields are already resolved.
```
murblanc, line 37: Given this class does not implement `Schedule` (didn't get yet to the point where it's used) maybe its name or this comment should be clarified?

solr/core/src/java/org/apache/solr/cluster/scheduler/impl/SolrSchedulerImpl.java
```diff
@@ -0,0 +1,121 @@
...
+ * <p>Scheduled executions are triggered at most with {@link #SCHEDULE_INTERVAL_SEC} interval.
+ * Each registered {@link Schedulable} is processed sequentially and if its next execution time
+ * is in the past its {@link Schedulable#run()} method will be invoked.
+ * <p>NOTE: If the total time of execution of all registered Schedulable-s exceeds any schedule
+ * interval then exact execution times will be silently missed.</p>
+ */
+public class SolrSchedulerImpl implements SolrScheduler, ClusterSingleton {
+  private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+
+  public static final int SCHEDULE_INTERVAL_SEC = 10;
```
murblanc, line 46: Instead of polling for tasks to run every 10 seconds could we be smart and set the next execution time of the scheduler to when the next job needs to be run? Is it worth the additional investment?

solr/core/src/java/org/apache/solr/cluster/scheduler/SolrScheduler.java
```diff
@@ -0,0 +1,27 @@
...
+ * limitations under the License.
+ */
+package org.apache.solr.cluster.scheduler;
+
+/**
+ *
+ */
+public interface SolrScheduler {
+
+  void registerSchedulable(Schedulable schedulable);
```
murblanc, line 24: How is this method going to be called in practice? Do we assume each node (in the SolrCloud cluster) will register all tasks locally (those that are not `ClusterSingleton`) and that multiple instances are going to run in parallel? Or will a `ClusterSingleton` task be registering other scheduled tasks that as a consequence will only have a single instance running on the cluster?

solr/core/src/java/org/apache/solr/cluster/scheduler/Schedulable.java
```diff
@@ -0,0 +1,32 @@
...
+ * Component to be scheduled and executed according to the schedule.
+ */
+public interface Schedulable {
+
+  Schedule getSchedule();
+
+  /**
+   * Execute the component.
+   * <p>NOTE: this should be a lightweight method that executes quickly, to avoid blocking the
+   * execution of other schedules. If it requires more work it should do this in a separate thread.</p>
```
murblanc, line 29: If a scheduled component starts a new thread to do its work, the schedule is going to get skewed pretty quickly and we might have multiple "copies" of the scheduled component being started in parallel. We'd be delegating the responsibility of insuring a single executing instance of the component (on a given node where it was registered) to the component itself.

noblepaul:
Can we just commit the `ClusterSingleton` as one PR please? `Scheduler` and `ClusterSingleton` doesn't have to go in together

sigram:
@noblepaul in this PR the Scheduler depends on ClusterSingleton, that's why I imported it into this PR. Whatever the ClusterSingleton becomes it will be the version from PR-1758.

solr/core/src/java/org/apache/solr/cluster/scheduler/SolrScheduler.java
sigram: Huh, it looks like I lost some part of this PR during the split ... :) `SolrSchedulerImpl` is a `ClusterSingleton`, at some point I had a section in `CoreContainer` that would instantiate it and add to the list of singletons to be managed. And another section that would scan all registered handlers to see if any one of them was `Schedulable`.

(A side-note: I feel like we keep adding these small hacks just to compensate for the lack of a proper dependency-injection framework... sigh..)

solr/core/src/java/org/apache/solr/cloud/ClusterSingleton.java
sigram: I didn't want to complicate the API to include support for more complex state and async operations. Besides, there's only one parent that is supposed to call the `start()` method, and no other actor is supposed to do that, so from the POV of other components the singleton is either running or not, and the STARTING state is beyond their control anyway - from their POV the singleton is in "not running" state during startup.

solr/core/src/java/org/apache/solr/cloud/Overseer.java
sigram: I don't think it matters what was the configuration - the contract for `ClusterSingleton` requires that we stop all of them.

solr/core/src/java/org/apache/solr/cluster/scheduler/Schedulable.java
sigram: As for the schedule being skewed ... yeah, we're not building a Quartz replacement here :) The javadoc clearly states that scheduled runs may be missed if execution of schedulables takes too long.

Having said that, we should probably add some protection in the implementation to avoid situations where one buggy `Schedulable` starves all others.

solr/core/src/java/org/apache/solr/cluster/scheduler/impl/CompiledSchedule.java
sigram: This is an implementation detail. I can add more javadocs, sure, but this class is package-private.

solr/core/src/java/org/apache/solr/cluster/scheduler/impl/CompiledSchedule.java
sigram: Currently it's set at the start of execution. This reduces the schedule drift due to execution times.

solr/core/src/java/org/apache/solr/cluster/scheduler/impl/SolrSchedulerImpl.java
sigram: Not sure. This implementation is very simple on purpose, we can always complicate things later ;)

PR #1832, opened 2020-09-04
SOLR-14831: remove deprecated-and-unused "facet.distrib.mco" constant

cpoerschke:
https://issues.apache.org/jira/browse/SOLR-14831


HoustonPutman (approved):
Thanks for finding this! Not sure why it wasn't removed from master (then 8.0) initially...

PR #1833, opened 2020-09-04
LUCENE-9501: Fix invariant violation in IndexSortSortedNumericDocValuesRangeQuery.

jtibshirani:
Previously the DocIdSetIterator returned an old value for docID even after
advance returned NO_MORE_DOCS. This violates the DocIdSetIterator contract and
made it possible for the iterator's advance method to be called even after it
was already exhausted.


lucene/sandbox/src/test/org/apache/lucene/search/TestIndexSortSortedNumericDocValuesRangeQuery.java
```diff
@@ -65,7 +65,7 @@ public void testSameHitsAsPointRangeQuery() throws IOException {
         iw.deleteDocuments(LongPoint.newRangeQuery("idx", 0L, 10L));
       }
       final IndexReader reader = iw.getReader();
-      final IndexSearcher searcher = newSearcher(reader, false);
+      final IndexSearcher searcher = newSearcher(reader);
```
jtibshirani, line 68: This isn't critical for test coverage, but it seemed off that we had disabled wrapping the reader.

jpountz (approved):

PR #1834, opened 2020-09-04
Make sure to test normal scorers with asserting wrappers.

jtibshirani:
When a query is run at the top-level, the searcher uses `Weight#bulkScorer`.
Many queries don't implement this explicitly and instead rely on the default
implementation which delegates to `Weight#scorer`.

Previously `AssertingWeight` would always wrap the delegate's bulk scorer. So
for queries that rely on `Weight#scorer`, we weren't wrapping the scorer or
iterator to run checks. This change proposes that `AssertingWeight#bulkScorer`
sometimes use the default implementation to make sure we also test normal
scorers.

This change would have caught the bug in LUCENE-9501.


lucene/test-framework/src/java/org/apache/lucene/search/AssertingWeight.java
```diff
@@ -85,11 +85,18 @@ public long cost() {
...
   public BulkScorer bulkScorer(LeafReaderContext context) throws IOException {
-    BulkScorer inScorer = in.bulkScorer(context);
+    BulkScorer inScorer;
+    // We explicitly test both the delegate's bulk scorer, and also the normal scorer.
+    // This ensures that normal scorers are sometimes tested with an asserting wrapper.
+    if (random.nextBoolean()) {
+      inScorer = in.bulkScorer(context);
+    } else {
+      inScorer = super.bulkScorer(context);
+    }
```
jpountz, line 95: I know of several tests that test first with a top-level query and then with a filter to make sure that both the bulk scorer and the scorer are tested. With this change, 50% of the time, we wouldn't be testing the bulk scorer, so I wonder if we should replace `random.nextBoolean()` with `frequently()` or something like that.

jtibshirani: That makes sense to me, I can reduce the frequency with which we take the `Scorer` path.

I didn't know about the practice of testing both with a top-level query and a filter. To me this change is still valuable, it ensures we check the scorer/ iterator contract sometimes in a clear + solid way.

lucene/CHANGES.txt
```diff
@@ -204,7 +204,8 @@ Improvements
 * LUCENE-9446: In BooleanQuery rewrite, always remove MatchAllDocsQuery filter clauses
   when possible. (Julie Tibshirani)
 
-* LUCENE-9501: Improve how Asserting* test classes handle singleton doc values.
+* LUCENE-9501: Improve coverage for Asserting* test classes: make sure to handle singleton doc
+  values, and sometimes wrap Scorers by default. (Julie Tibshirani)
```
jpountz, line 208: I'm not sure I'd understand what "sometimes wrap scorers" would mean as a user reading the changelog. Maybe say something along the lines of "sometimes exercise the `Weight#scorer` API rather than `Weight#bulkScorer` for top-level queries".

jpountz (approved):
I left a minor comment about the changelog but this LGTM.

lucene/CHANGES.txt
jtibshirani: This is a good point, I updated the changelog entry.

PR #1835, opened 2020-09-06
LUCENE-9509: Refine lucene/BUILD.md and top-level README (for newdevs)

mocobeta:
Current lucene/BUILD.md is somehow mixed-up with information for Solr developers, this is wrong place for them. Solr related information/tips should be moved to the top-level README (or somewhere else).

Also the content could be elaborated for newdevs (required JDK version, and so on).


mocobeta:
You can see how the modified README/BUILD look like on GitHub at here:
https://github.com/mocobeta/lucene-solr/tree/jira/LUCENE-9509

PR #1837, opened 2020-09-06
LUCENE-7882: First idea of using Java 15 hidden anonymous classes for Lucene expressions

uschindler:
This PR is a first (but already working) idea of using Java 15 hidden classes (see https://openjdk.java.net/jeps/371) to implement the Lucene expressions. The big advantages:
- No classloader for every expression is needed, because the class is completely anonymous and has no strong reference to a classloader. Actually the class has a classloader to lookup any referenced other class, but it is NOT loaded by any classloader
- The class can easily be unloaded
- The performance of loading that class is better, as no locks can occur (classloaders have locks when looking up classes), see https://issues.apache.org/jira/browse/LUCENE-7882

Backside:
- The class and its methods do not appear in any stack trace. This also fails one test in the test suite. When using this for Lucene expressions, we have to think about a better way how to make the source code and method calls visible in stack traces. Like lambda frames the hidden class is not visible (unless enabled in JVM to show hidden frames). I have a question open on openjdk mailing list: https://mail.openjdk.java.net/pipermail/core-libs-dev/2020-September/068542.html
- It currently does not work if you pass a different classloader than Lucene's to the expressions module. To allow this we need to change APIs a bit (Classloader -> Lookup).

@mikemccand can you test this with JDK 15 (release candidate) and your test. You should not see any locks anymore, speed should be higher, and the created anonymous classes should be unloaded very fast.

### DIFF
--- a/lucene/expressions/src/java/org/apache/lucene/expressions/js/JavascriptCompiler.java
+++ b/lucene/expressions/src/java/org/apache/lucene/expressions/js/JavascriptCompiler.java
@@ -239,6 +239,8 @@ public final class JavascriptCompiler {
   }
 
   private Class<? extends Expression> defineClass(ClassLoader parent, byte[] bc) throws IllegalAccessException {
+    // if we are on Java 15+ and the classloader is our own classloader, we can create an
+    // anonymous hidden class (JEP-371):
     if (MH_defineHiddenClass != null && parent == getClass().getClassLoader()) {
       final Lookup hiddenLookup;
       try {
--- a/lucene/expressions/src/java/org/apache/lucene/expressions/js/JavascriptCompiler.java
+++ b/lucene/expressions/src/java/org/apache/lucene/expressions/js/JavascriptCompiler.java
@@ -18,8 +18,11 @@ package org.apache.lucene.expressions.js;
 
 import java.io.IOException;
 import java.io.Reader;
+import java.lang.invoke.MethodHandle;
 import java.lang.invoke.MethodHandles;
+import java.lang.invoke.MethodHandles.Lookup;
 import java.lang.invoke.MethodType;
+import java.lang.reflect.Array;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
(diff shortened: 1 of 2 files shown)


lucene/expressions/src/java/org/apache/lucene/expressions/js/JavascriptCompiler.java
```diff
@@ -85,7 +88,33 @@
...
+          MethodType.methodType(Lookup.class, byte[].class, boolean.class, emptyOptions.getClass()));
+      mh = mh.bindTo(MethodHandles.lookup()); // private lookup of JavascriptCompiler!
+      mh = MethodHandles.insertArguments(mh.asFixedArity(), 1, true, emptyOptions);
+    } catch (ReflectiveOperationException e) {
+      mh = null;
+    }
+    MH_defineHiddenClass = mh;
+  }
+  
+  private static final int CLASSFILE_VERSION = Opcodes.V11;
```
uschindler, line 117: Actually this should have been done on master already! We are on Java 11, so classfile format should be Java 11, too.

lucene/expressions/src/java/org/apache/lucene/expressions/js/JavascriptCompiler.java
```diff
@@ -206,6 +247,12 @@ private Expression compileExpression(ClassLoader parent) throws ParseException {
       throw new IllegalStateException("An internal error occurred attempting to compile the expression (" + sourceText + ").", exception);
     }
   }
+
```
dweiss, line 250: I wouldn't do this, to be honest. Rethrow Error and RuntimeException subclasses as they were, wrap anything else in a runtime exception with an appropriate caused-by delegate. Makes it easier to reason about the code should something happen.

lucene/expressions/src/test/org/apache/lucene/expressions/js/TestCustomFunctions.java
```diff
@@ -263,7 +263,7 @@ public void testThrowingException() throws Exception {
     PrintWriter pw = new PrintWriter(sw);
```
dweiss, line 263: Perhaps we should import assertj for tests. These assertions are so much cleaner with assertj. Don't know whether hamcrest equivalent exist (maybe it does).

lucene/expressions/src/java/org/apache/lucene/expressions/js/JavascriptCompiler.java
```diff
@@ -191,9 +220,21 @@ private Expression compileExpression(ClassLoader parent) throws ParseException {
 
     try {
       generateClass(getAntlrParseTree(), classWriter, externalsMap);
-
-      final Class<? extends Expression> evaluatorClass = new Loader(parent)
-        .define(COMPILED_EXPRESSION_CLASS, classWriter.toByteArray());
+      
+      final Class<? extends Expression> evaluatorClass;
```
dweiss, line 224: Just a suggestion. My personal preference is to set up a supplier (or a function) statically, depending on conditions, then just use that later forever. In this case you could have something like: 
````
static Function<byte[], Class<? extends Expression>> classLoader;
static {
  if (MH_...) {
    classLoader = (bytes) -> {};
  } else {
    classLoader = (bytes) -> {};
  }
}
````
which would be defined once and used without the knowledge of its internal implementation details later on, forever. Just a thought.

dweiss (approved):

lucene/expressions/src/test/org/apache/lucene/expressions/js/TestCustomFunctions.java
uschindler: Yeah, I added this only for my personal debugging, because this test fails with my change (as the expression is hidden from the stack trace, so the "contains" does not work. But when it fails, you don't see what should be there.

Assertj is really better as it automatically prints and also shortens the printout to only relevant part.

lucene/expressions/src/java/org/apache/lucene/expressions/js/JavascriptCompiler.java
uschindler: I changed it. The way how to do this is correct as it only intends to work around a problem of type inference in javac. For a methodhandle we know for sure, which exceptions are thrown (it's different to core reflection, where everything is wrapped in InvocationTargetException).

I moved the whole thing to a separate method and I catched and rethrow all "known" Exceptions (IllegalAccessException, RuntimeException, Error), all others can't happen, so it wraps with AssertionError.

lucene/expressions/src/test/org/apache/lucene/expressions/js/TestCustomFunctions.java
dweiss: I love assertj - it really prints out stuff in more elegant fashion than pure junit+hamcrest.

lucene/expressions/src/java/org/apache/lucene/expressions/js/JavascriptCompiler.java
uschindler: I was thinking about that, but as it throws a checked Exception it would need a new functional interface. I moved the whole stuff to a separate method `defineClass()`.

rmuir:
Nice improvement here. I think we should not let the stacktrace stuff be a blocker for this optimization. Out of box, the user should never even see exceptions anyway: the expression stuff all works on doubles. So all those math functions just return NaN/Inf for special cases because that's how java Math functions want to work.

mikemccand:
Whoa, thanks @uschindler -- this looks awesome.

> @mikemccand can you test this with JDK 15 (release candidate) and your test. You should not see any locks anymore, speed should be higher, and the created anonymous classes should be unloaded very fast.

We (Amazon customer facing product search team) make heavy use of Lucene's (powerful!) expressions module -- I will try to figure out a way to test this (we are on JDK 11 now for production).

uschindler:
Replaced by https://github.com/apache/lucene/pull/12873

PR #1838, opened 2020-09-07
SOLR-14768: Fix multipart POST to Solr.

dsmiley:
https://issues.apache.org/jira/browse/SOLR-14768

Regression from 8.6

Multipart POST would fail due to a NoClassDefFoundError of Jetty MultiPart.  Solr cannot access many Jetty classes, which is not noticeable in our tests.

I tested this manually:  I started the "techproducts" schema via `bin/solr start -e techproducts` then I did this:

````
curl http://localhost:8983/solr/techproducts/update/extract \
 -F literal.id=doc2 -F stream.body=@example/exampledocs/sample.html
````
It should work.  It used to dump a stacktrace from Solr.

I would be awesome if we could have an automated test for this given the test infrastructure we have, but AFAIK I don't see anything.  A suitable test would need to run bin/solr.  If we had a SolrExampleTests subclass that required a URL input then we could run tests against a Solr instance that is being released by the smoketester and also by a future Docker release process.  I'll raise this particular matter in the dev list for discussion.


dsmiley:
I'll commit soon; I tested and a user tested it's fixed with this change.

PR #1839, opened 2020-09-07
LUCENE-9511: Include StoredFieldsWriter in DWPT accounting

s1monw:
StoredFieldsWriter might consume some heap space memory that
can have a significant impact on decisions made in the IW if
writers should be stalled or DWPTs should be flushed if memory
settings are small in IWC and flushes are frequent. This change adds
RAM accounting to the StoredFieldsWriter since it's part of the
DWPT lifecycle and not just present during flush.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/DefaultIndexingChain.java
+++ b/lucene/core/src/java/org/apache/lucene/index/DefaultIndexingChain.java
@@ -52,7 +52,7 @@ import org.apache.lucene.util.RamUsageEstimator;
 /** Default general purpose indexing chain, which handles
  *  indexing all types of fields. */
 final class DefaultIndexingChain extends DocConsumer {
-  final Counter bytesUsed;
+  final Counter bytesUsed = Counter.newCounter();
   final DocumentsWriterPerThread docWriter;
   final FieldInfos.Builder fieldInfos;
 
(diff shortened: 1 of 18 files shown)


lucene/core/src/java/org/apache/lucene/index/DocumentsWriterPerThread.java
```diff
@@ -351,7 +354,7 @@ FlushedSegment flush(DocumentsWriter.FlushNotifications flushNotifications) thro
         flushState.liveDocs.clear(deleteDocIDs[i]);
       }
       flushState.delCountOnFlush = numDeletedDocIds;
-      bytesUsed.addAndGet(-(deleteDocIDs.length * Integer.SIZE));
+      bytesUsed.addAndGet(-(deleteDocIDs.length * Integer.BYTES));
```
mikemccand, line 357: Whoa, this was a latent (over-counting) bug?  Maybe `grep` for other places we are possibly incorrectly using `Integer.SIZE`!

Though, I think this array is typically tiny (usually 0 length), since it holds docs that hit non-aborting exception during indexing?

lucene/core/src/java/org/apache/lucene/index/DocumentsWriterPerThread.java
```diff
@@ -292,7 +295,7 @@ private void deleteLastDocs(int docCount) {
     for (int docId = from; docId < to; docId++) {
       deleteDocIDs[numDeletedDocIds++] = docId;
     }
-    bytesUsed.addAndGet((deleteDocIDs.length - size) * Integer.SIZE);
+    bytesUsed.addAndGet((deleteDocIDs.length - size) * Integer.BYTES);
```
mikemccand, line 298: Also a latent over-counting bug?

lucene/core/src/java/org/apache/lucene/index/DefaultIndexingChain.java
```diff
@@ -52,7 +52,7 @@
 /** Default general purpose indexing chain, which handles
  *  indexing all types of fields. */
 final class DefaultIndexingChain extends DocConsumer {
-  final Counter bytesUsed;
+  final Counter bytesUsed = Counter.newCounter();
```
mikemccand, line 55: So we are breaking out a separate `Counter` for `DefaultIndexingChain`, from `DWPT`, when previously they shared the same counter?

mikemccand (approved):
Change looks great!  Was there a real user use case that triggered this?  Or a randomized test failure?  I'm curious if the amount of RAM `StoredFieldsWriter` uses is sometimes substantial ...

This will cause IW to flush (if triggered by RAM) more often?  But hopefully not THAT much more often?  Let's watch [indexing throughput in our nightly benchmarks](https://home.apache.org/~mikemccand/lucenebench/indexing.html) and confirm it didn't impact indexing much.

lucene/core/src/java/org/apache/lucene/index/DocumentsWriterPerThread.java
s1monw: yeah I checked all instanced in our codebase after I found this.

lucene/core/src/java/org/apache/lucene/index/DocumentsWriterPerThread.java
s1monw: yeah same thing

lucene/core/src/java/org/apache/lucene/index/DefaultIndexingChain.java
s1monw: exactly that's the main change here otherwise we need to do back and forth accounting inside the StoredFieldsWriter

s1monw:
> Change looks great! Was there a real user use case that triggered this? Or a randomized test failure? I'm curious if the amount of RAM StoredFieldsWriter uses is sometimes substantial ...

This was triggered by a test failure that hit and OOM due to huge amount of flushing (maxBufferedDocs=3) and blocked flushes. The SFW uses about 650kb memory per DWPT instance and that might add quickly if flushing can't keep up with indexing. I don't think it will have a big impact on performance but these settings are valid and should never cause OOM. Now that it's accounted for IW stall protection will kick in and safes us from going OOM

PR #1841, opened 2020-09-08
SOLR 13528: Move Rate Limiter Configuration To Be A Plugin

atris:
This commit refactors rate limiters configuration to now be modelled as a plugin and move its configuration from web.xml to a dedicated section in solr.xml


tflobbe (approved):
LGTM

PR #1842, opened 2020-09-08
LUCENE-9512: Move LockFactory stress test to be a unit/integration test

uschindler:
This PR moves the lock factory stress test to be a :lucene:core unit/integration test. In verbose mode the output of clients is inherited, in standard test mode its written to files (to allow debugging).

Unfortunately, as the server uses TCP/IP socket to communicate with the childs, I have to add listen/accept to ephemeral ports to lucene's policy file (@rmuir looks ok?).

If this gets backported to 8.x, we can get rid of the rather complicated Ant Target, too.

See https://issues.apache.org/jira/browse/LUCENE-9379 for background information.


lucene/core/src/test/org/apache/lucene/store/TestLockFactoriesMultiJVM.java
```diff
@@ -0,0 +1,115 @@
...
+            Integer.toString(delay),
+            Integer.toString(rounds)
+          ), i, dir).start());
+      }
+      
+      // wait for all processes to exit
+      int exited = 0;
+      while (exited < clients) {
+        for (Process p : processes) {
+          if (p.waitFor(1, TimeUnit.SECONDS)) {
```
dweiss, line 93: 1 second may not be enough on heavy loaded machines?

gradle/testing/randomization/policies/tests.policy
```diff
@@ -64,6 +64,9 @@ grant {
   permission java.lang.RuntimePermission "getClassLoader";
   permission java.lang.RuntimePermission "setContextClassLoader";
 
+  // TestLockFactoriesMultiJVM opens a random port on 127.0.0.1 (port 0 = ephemeral port range):
```
dweiss, line 67: This property wouldn't be needed if you forked the "server: side as well as clients... :)

lucene/core/src/test/org/apache/lucene/store/TestLockFactoriesMultiJVM.java
```diff
@@ -0,0 +1,115 @@
...
+      int port;
+      while ((port = LockVerifyServer.PORT.get()) == -1) {
+        Thread.sleep(100L);
+      }
+      
+      // spawn clients as separate Java processes
+      final List<Process> processes = new ArrayList<>();
+      for (int i = 0; i < clients; i++) {
+        processes.add(applyRedirection(new ProcessBuilder(
+            Paths.get(System.getProperty("java.home"), "bin", "java").toString(),
```
dweiss, line 75: Does it make sense to limit the default heap of these subprocesses here?

lucene/core/src/test/org/apache/lucene/store/TestLockFactoriesMultiJVM.java
```diff
@@ -0,0 +1,115 @@
...
+import java.util.List;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
+
+import org.apache.lucene.util.LuceneTestCase;
+import org.apache.lucene.util.NamedThreadFactory;
+import org.apache.lucene.util.SuppressForbidden;
+
+public class TestLockFactoriesMultiJVM extends LuceneTestCase {
```
dweiss, line 32: I would maybe mark this as belonging to Slow group? Just a thought.

lucene/core/src/test/org/apache/lucene/store/TestLockFactoriesMultiJVM.java
```diff
@@ -0,0 +1,115 @@
...
+            dir.toString(),
+            Integer.toString(delay),
+            Integer.toString(rounds)
+          ), i, dir).start());
+      }
+      
+      // wait for all processes to exit
+      int exited = 0;
+      while (exited < clients) {
+        for (Process p : processes) {
```
dweiss, line 92: I think we should absolutely try to clean up forked processes in case something goes wrong (even if it's an internal suite timeout!) - regardless of the outcome of the test. I'd do a run over the process list in finally and if it's still in a live state, kill it. 

https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Process.html#destroyForcibly()

Another thought is maybe to use a list of completable futures to wait for all forked processes instead of counting them?
https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Process.html#onExit()

These are details; the cleanup is kind of important.

lucene/core/src/test/org/apache/lucene/store/TestLockFactoriesMultiJVM.java
rmuir: this is inside a loop though: so it will effectively wait forever?

lucene/core/src/test/org/apache/lucene/store/TestLockFactoriesMultiJVM.java
rmuir: FYI I ran it locally on a 2-core laptop. The tests are a bit slow but not the slowest in lucene-core.

The slowest tests (exceeding 500 ms) during this run:
   2.76s TestIndexWriter.testMaxCompletedSequenceNumber (:lucene:core)
   1.96s TestNRTCachingDirectory.testNRTAndCommit (:lucene:core)
   1.69s TestIndexWriter.testThreadInterruptDeadlock (:lucene:core)
   1.60s TestShardSearching.testSimple (:lucene:core)
   1.55s TestStandardAnalyzer.testLargePartiallyMatchingToken (:lucene:core)
   1.46s TestTieredMergePolicy.testSimulateAppendOnly (:lucene:core)
   **1.32s TestLockFactoriesMultiJVM.testNativeFSLockFactory (:lucene:core)**
   **1.23s TestLockFactoriesMultiJVM.testSimpleFSLockFactory (:lucene:core)**
   1.17s TestLevenshteinAutomata.testLev2 (:lucene:core)
   1.15s TestIndexSorting.testConcurrentUpdates (:lucene:core)

lucene/core/src/test/org/apache/lucene/store/TestLockFactoriesMultiJVM.java
dweiss: Ah, correct. A completable future wait-for would be more elegant here I think but fine as it is too.

gradle/testing/randomization/policies/tests.policy
uschindler: This does not work, as there needs to be a communication, so the test known when to start the clients. This was not different in Ant (the main process was running fork=false, the clients fork=true)

lucene/core/src/test/org/apache/lucene/store/TestLockFactoriesMultiJVM.java
uschindler: i tested 32M, coming in next push

lucene/core/src/test/org/apache/lucene/store/TestLockFactoriesMultiJVM.java
uschindler: It will be sligtly faster soon, but can be tuned. But marking it as `@Slow` is a bad idea because this is a very important test! Especially to collect information from users, if it really works everywhere.

To really do extensive testing do `-Dtests.nightly=true`, then it runs about 2 minutes.

lucene/core/src/test/org/apache/lucene/store/TestLockFactoriesMultiJVM.java
uschindler: I copied this code from another test doing similar stuff (crushing JVM). I will work change it a bit.

dweiss: onExit is from Java 9 (so wasn't available). I vaguely remember destroyForcibly didn't work as expected if it forked processes of its own (only killed the forked process, not its entire hierarchy) -- we have wrappers that take care of that (because we use process forking extensively in integration tests). From Java 11 on you can enumerate the hierarchy and kill it manually.

https://github.com/carrotsearch/procfork/blob/master/src/main/java/com/carrotsearch/procfork/ForkedProcess.java#L92-L93

uschindler:
I pushed some fixes:
- Removed the AtomicInteger for the port number. As the server is running in test, we just call the server directly and it uses a Consumer(InetAddr) as callback.
- Because of the callback, we can start the child processes without using a thread(pool).
- Cleanup of processes improved.

Keep in mind: The waitFor of 15s is is now way too much. Once the server ended and we run into the cleanup, all processes dies already, because the server shuts down when all clients dicsonnected. The cleanup is only needed if anything goes wrong on startup (like client can't connect to server due to network error on lo interface).

rmuir:
I don't want to block the changes here, but just think we should make a followup to implement the same test, but with clients in the same JVM.

There are two separate cases for NIO locks (held by same VM vs held by another process), so we should stress test them both. Past experience has shown that whatever lock testing we have, it is somehow never enough.

dweiss:
Let me fix that import error and look into it, Uwe. Don't commit it yet.

uschindler:
The import error was fixed already. It was not caused by this PR, but a change in master.

uschindler:
I merged from master to retrigger the check. Clicking on "Rerun checks" did not work, as it merged into same master commit as before. Makes no sense to me (maybe because of reproducibility).

dweiss:
I've tried to clean up things a bit and ended up implementing both forked and in-JVM mode. It's not super beautiful (those intermediate interfaces could probably be simplified) but it works and allows for some interesting combinations of mixed forked-and-local workers. 

I committed this in to your branch, Uwe, but it's really subject to criticism, modifications and/or reversal - your decision.

uschindler:
> I don't want to block the changes here, but just think we should make a followup to implement the same test, but with clients in the same JVM.
>
> There are two separate cases for NIO locks (held by same VM vs held by another process), so we should stress test them both. > Past experience has shown that whatever lock testing we have, it is somehow never enough.

I think this could be a separate issue. Basically, this just "restores" the test in master, which was done in Ant only. There are no functional changes.

But this new code is easy to adapt, there are several possibilities:
- We can clone the whole test and just replaces `List<Process> processes` by a thread pool and instead of spawning a JVM we just call main() in a thread. Cleanup is more or less the same.
- We can modify the current test to not only spawn multiple JVMs, but aditionally also add a thread. We can even randomize the number of clients/threads (currently fixed to 2 clients, but it's just a change in test config).

In both cases, the policy also needs the "connect" permission. An alternative would be to allow client/server to alternatively communicate through a PipedInput/OutputStream (if its in same JVM).

Some questions before I commit this:
- Should I rename the test to TestStressLockFactories?

uschindler:
Hi @dweiss 
indeed this looks now very complicated (before the test was so simple after my changes from last night). Indeed I want to remove some abstraction, the factory is unneeded. I always try to use the default functional interfaces provided by Java 8, and just creating one for exactly one methods is code bloat. A simple if/else in the factory method would be as fine.

We should maybe also change the number of client to 3 instead of 2. Also during nightly tests we can hammer with even more clients.

What I committed in: I disabled all custom Mock Filesystems, as this is important to run on native OS without any wrapping.

uschindler:
I'd also change the test name now to `TestStressLockFactories`.

uschindler:
Hi, I reordered the methods a bit in the class and applied some changes with number of clients:
- Default is now 3. Actually this does not change the test runtime (on my machine)
- On nightly we create 8 clients

I would still like to get rid of the Supplier. IMHO, we should remove one randomization level. As we now have 3 clients (or 8), a randomization could always have a 2:1 chance to produce a forked client. What do you think, @dweiss ?

uschindler:
FYI, the limit to 2 clients was caused by ANT's xml. Every client used another XML copypaste. But actually the number of clients is not really a performance factor.

dweiss:
I don't mind. Reshape any way you like.

uschindler:
Hi,
I refactored the code further and removed the randomization:
- If uses `(clientId % 4)==0` to produce a thread client and otherwise a process client
- In default mode, it creates 3 clients, so one of them is always in-process
- In nightly mode, it creates 8 clients, so two of them are in-process

I think that covers all modes.

I think it's ready to commit.

Question: Also backport to 8.x and remove the additional ANT task?

uschindler:
Hi again,
after reading the code of the LockStressTest: It tries to re-obtain a lock in the same JVM based on randomization. So actually the original test did everything right. @rmuir Can you verify this?

So IMHO, we can remove the inprocess stuff and rever back to the state from this morning.

uschindler:
Code is here:
```java
        try (final Lock l = verifyLF.obtainLock(lockDir, LOCK_FILE_NAME)) {
          if (rnd.nextInt(10) == 0) {
            if (rnd.nextBoolean()) {
              verifyLF = new VerifyingLockFactory(getNewLockFactory(lockFactoryClassName), in, out);
            }
            try (final Lock secondLock = verifyLF.obtainLock(lockDir, LOCK_FILE_NAME)) {
              throw new IOException("Double obtain");
            } catch (LockObtainFailedException loe) {
              // pass
            }
          }
          Thread.sleep(sleepTimeMS);
        } catch (LockObtainFailedException loe) {
          // obtain failed
        }
```

uschindler:
I was talking with Robert. Let's revert the in-process stuff. Makes not much difference. We actually have a very good test for in-process logs, part of BaseLockFactoryTestCase.

uschindler:
OK, reverted most of the changes in the stress tester. I kept the changes done by @dweiss about the magic numbers like starting gun "43" or access/release in the wire protocol.

uschindler:
@dweiss Are you fine with this? I would like to commit and backport this soon.

Unless you have some minor comments about the cleanup process, as I had to remove your abstraction on top of that.

dweiss:
It's fine, Uwe. Go ahead.

uschindler:
I backported this also to 8.x, removing the Ant targets related to this. Much cleaner now!

PR #1843, opened 2020-09-08
SOLR-14846 Remove Optional.ofNullable.orElse pattern

madrob:
https://issues.apache.org/jira/browse/SOLR-14846

### DIFF
--- a/solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpClientBuilderFactory.java
+++ b/solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpClientBuilderFactory.java
@@ -17,7 +17,6 @@
 package org.apache.solr.client.solrj.impl;
 
 import java.io.Closeable;
-import java.util.Optional;
 
 /**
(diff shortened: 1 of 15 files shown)


solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpClientBuilderFactory.java
```diff
@@ -36,7 +35,7 @@
    *                by configured (optional).
    * @return the {@linkplain SolrHttpClientBuilder}
    */
-  public SolrHttpClientBuilder getHttpClientBuilder(Optional<SolrHttpClientBuilder> builder);
+  public SolrHttpClientBuilder getHttpClientBuilder(SolrHttpClientBuilder builder);
```
vthacker, line 38: I'm wondering if this is a change we want to do in 9.0 only?

madrob: Yea, since we're changing APIs here, it would have to probably be 9 only

HoustonPutman (approved):
Looks good to me

PR #1847, opened 2020-09-09
LUCENE-9514: Include TermVectorsWriter in DWPT accounting

s1monw:
TermVectorsWriter might consume some heap space memory that
can have a significant impact on decisions made in the IW if
writers should be stalled or DWPTs should be flushed if memory
settings are small in IWC and flushes are frequent. This change adds
RAM accounting to the TermVectorsWriter since it's part of the
DWPT lifecycle and not just present during flush.


lucene/codecs/src/java/org/apache/lucene/codecs/simpletext/SimpleTextTermVectorsWriter.java
```diff
@@ -193,4 +189,9 @@ private void write(BytesRef bytes) throws IOException {
   private void newLine() throws IOException {
     SimpleTextUtil.writeNewline(out);
   }
+
+  @Override
+  public long ramBytesUsed() {
+    return scratch.get().bytes.length;
```
mikemccand, line 195: Ha.

lucene/core/src/java/org/apache/lucene/index/StoredFieldsConsumer.java
```diff
@@ -42,6 +44,7 @@
   protected void initStoredFieldsWriter() throws IOException {
     if (writer == null) { // TODO can we allocate this in the ctor? we call start document for every doc anyway
       this.writer = codec.storedFieldsFormat().fieldsWriter(directory, info, IOContext.DEFAULT);
+      accountable = writer;
```
mikemccand, line 47: Hmm what is happening here?  Are we using this new `accountable` member somewhere?  I don't see it in the diffs?  Oh, I see, it is returned in `DefaultIndexingChain.getChildResources()`, ok!  Maybe add comment above its declaration explaining that our parent/owning `DefaultIndexingChain` returns/uses it?

lucene/core/src/java/org/apache/lucene/util/Accountable.java
```diff
@@ -41,4 +41,8 @@
     return Collections.emptyList();
   }
 
+  /**
+   * An accountable that always returns 0
+   */
+  Accountable NULL_ACCOUNTABLE = () -> 0;
```
mikemccand, line 47: Interface members are always `static final` right?

mikemccand (approved):
Thanks @s1monw.  Surprising `TermVectorsWriter` might use non-trivial RAM!

lucene/core/src/java/org/apache/lucene/util/Accountable.java
s1monw: yes

uschindler: To avoid such questions, you can be explicit.

s1monw: sure, but do we then initialize all booleans with `= false`? I thinks it's totally obsolet.

PR #1849, opened 2020-09-09
LUCENE-9517: Add doPrivileged block when creating BugfixDeflater_JDK8252739 objects

iverase:
see https://issues.apache.org/jira/browse/LUCENE-9517

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/codecs/lucene87/BugfixDeflater_JDK8252739.java
+++ b/lucene/core/src/java/org/apache/lucene/codecs/lucene87/BugfixDeflater_JDK8252739.java
@@ -16,6 +16,8 @@
  */
 package org.apache.lucene.codecs.lucene87;
 
+import java.security.AccessController;
+import java.security.PrivilegedAction;
 import java.util.Arrays;
 import java.util.zip.DataFormatException;
 import java.util.zip.Deflater;
@@ -40,7 +42,9 @@ final class BugfixDeflater_JDK8252739 extends Deflater {
       throw new IllegalArgumentException("dictLength must be >= 0");
     }
     if (IS_BUGGY_JDK) {
-      return new BugfixDeflater_JDK8252739(level, nowrap, dictLength);
+      // Some JDKs need "accessDeclaredMembers" privileges to create a Deflater object
+      return AccessController.doPrivileged(
+              (PrivilegedAction<BugfixDeflater_JDK8252739>) () -> new BugfixDeflater_JDK8252739(level, nowrap, dictLength));
     } else {
       return new Deflater(level, nowrap);
     }


iverase:
A couple of notes:

* Not sure how this change can be tested.
* We know this is only an issue for JDK11, should be add a check and only perform privilege action in that case?

dweiss:
This is a hideous implementation detail inside the JDK but I think the privileged action should just be there - I think this is the correct way to dodge the problem.

s1monw:
change LGTM

uschindler:
Sorry, this is bullshit. Fix your CI and don't apply such stuff in Lucene.

uschindler (changes requested):
-1

uschindler:
Actually the bug is in JDK and not in our code.

uschindler:
Actually I remember that this same issue happened also at other places in the JDK, luckily it was fixed on their side, because I complained on the mailing list.

uschindler:
If this an issue at your CI, just use StackWalker and let this go through.

uschindler:
The problem with the fix here is:
- I have no idea how the stack trace looks like. And under which circumstances you see this bug. Can you post a link to your CI, to see if it is causeed by the CI infrastrcuture? Does it also break with production ready software? How?
- Have you opened a bug report in JDK 10, 11 (it's caused by this change: https://bugs.openjdk.java.net/browse/JDK-8185582)? Should I do?

Are there any other ways how to workaround this bug?

Why does this PR help for this bug (I don't know the stack trace)?

uschindler:
Does it help, because this code is only called from Deflater's ctor? http://hg.openjdk.java.net/jdk/jdk11/file/1ddf9a99e4ad/src/java.base/share/classes/java/util/zip/Deflater.java#l207

lucene/core/src/java/org/apache/lucene/codecs/lucene87/BugfixDeflater_JDK8252739.java
```diff
@@ -40,7 +42,9 @@ public static Deflater createDeflaterInstance(int level, boolean nowrap, int dic
       throw new IllegalArgumentException("dictLength must be >= 0");
     }
     if (IS_BUGGY_JDK) {
-      return new BugfixDeflater_JDK8252739(level, nowrap, dictLength);
+      // Some JDKs need "accessDeclaredMembers" privileges to create a Deflater object
+      return AccessController.doPrivileged(
+              (PrivilegedAction<BugfixDeflater_JDK8252739>) () -> new BugfixDeflater_JDK8252739(level, nowrap, dictLength));
```
uschindler, line 47: At least this should be `PrivilegedAction<Deflater>`, because that's what we return.

uschindler:
Maybe we should do another fix for the original bug: As only setDictionary is affected, my proposal would be to not subclass `Deflater` at all and instead just return an interface only having the Deflater#setDictionary() method.

In buggy jdks it returns a interface pointing to the original Deflater's method, otherwise our own one. I can create a new PR for that. I like this more.

This may also speedup Java 11, because it's not doing crazy reflection on every new Deflater instance.

dweiss:
> Sorry, this is bullshit. Fix your CI and don't apply such stuff in Lucene.

Mhmm... I don't think it's in Lucene. Lucene subclasses the JDK and the problem is in the JDK code itself?

iverase:
Thanks Uwe for taking the time reviewing this issue. 

>I have no idea how the stack trace looks like. And under which circumstances you see this bug. Can you post a link to your CI, to see if it is causeed by the CI infrastrcuture? Does it also break with production ready software? How?

Stack-trace looks like:

```
java.security.AccessControlException: access denied ("java.lang.RuntimePermission" "accessDeclaredMembers") |  

at __randomizedtesting.SeedInfo.seed([7E21C06936079E14:4A7EB7FED693B2BB]:0) |  
-- | --
  |   | at java.security.AccessControlContext.checkPermission(AccessControlContext.java:472) |  
  |   | at java.security.AccessController.checkPermission(AccessController.java:897) |  
  |   | at java.lang.SecurityManager.checkPermission(SecurityManager.java:322) |  
  |   | at java.lang.Class.checkMemberAccess(Class.java:2847) |  
  |   | at java.lang.Class.getDeclaredMethod(Class.java:2471) |  
  |   | at java.util.zip.Deflater$DeflaterZStreamRef.get(Deflater.java:991) |  
  |   | at java.util.zip.Deflater.<init>(Deflater.java:207) |  
  |   | at org.apache.lucene.codecs.lucene87.BugfixDeflater_JDK8252739.<init>(BugfixDeflater_JDK8252739.java:53) |  
  |   | at org.apache.lucene.codecs.lucene87.BugfixDeflater_JDK8252739.createDeflaterInstance(BugfixDeflater_JDK8252739.java:43) |  
```

I believe It happens every time  the runtime environment is set to JDK11 na dit is a affected by the Deflator bug. Here is a link to one of the logs:

https://elasticsearch-ci.elastic.co/job/elastic+elasticsearch+7.x+matrix-java-periodic-fips/ES_RUNTIME_JAVA=adoptopenjdk11,nodes=general-purpose/212/consoleFull

This is un-released code so it is not production ready.

> Have you opened a bug report in JDK 10, 11 (it's caused by this change: https://bugs.openjdk.java.net/browse/JDK-8185582)? Should I do?

It would be great if you can report it. 

> The workaround may only work for Elasticsearch, because lucene-core.jar has the accessDeclaredMembers permission given due to the MMapDirectory hack. 

This is a good point and I agree this fix only works if you give special permissions to the jar.

> Maybe we should do another fix for the original bug: As only setDictionary is affected, my proposal would be to not subclass Deflater at all and instead just return an interface only having the Deflater#setDictionary() method.

That is another option I considered and it now makes more sense than this approach. +1 not to subclass Deflater.

uschindler:
I have a better fix without hacks, just not subclassing Deflater!

See PR #1850. I will close this one.

uschindler:
> It would be great if you can report it.

Will do later! I will post the link on the main issue. I will refer to your name.

uschindler:
FYI, here is the JDK bug: https://bugs.openjdk.java.net/browse/JDK-8252976

iverase:
Thanks for filing the JDK bug, looks great!

s1monw:
@uschindler why are you so grumpy, you have so much to add to this issue, why don’t you do it in a positive way?

You could have reacted with something like „oh man what a pity, that’s a bug in the JDK. Let’s fix it in a more elegant and general way than adding a doProviledged block... I have some ideas“ - it’s as important as not adding hacks to lucene. Community over code buddy. You are around long enough, folks reading this are discouraged.

uschindler:
Nevertheless, I think we should work the new workaround, because I don't want JDK to do crazy reflection on every constructor call. It would still be interesting to know, if making `BugfixDeflater_JDK8252739` public.

dweiss:
Simon is right that language matters. And stays on the Web forever. 
We all have dark sides. And we're all scared of yours, Uwe! :)

+1 to the new workaround. Adding forbidden-apis should keep it under control.

uschindler:
Thanks Dawid! All is fine :-)

One problem that was a bit confusing when @iverase opened the original issue was this sentence: "We are running into issues when running Elasticsearch CI with java security turned on." This looks like an issue only with CI and the PR is just to workaround some bug in the CI system, which should definitely not done. I have seen many issues opened by ES people using that terminology. It should be different: If Elastic's CI discovers a bug, a person should verify and understand it, and after that open an issue with "running tests in our Elasticsearch codebase discovered a bug". The terminology that was used is confusing as it implies that only CI is not happy.

Actually this bug should be visible to anybody running ES tests? 100% of all Tests should fail.

uschindler:
> I have a go locally and It does not seem to work.

Tht's what I expected. If they would have used `MethodHandles.lookup()` it would have worked better without `AccessController`. Thanks!!!

> Final question, I guess it is not possible in forbidden-apis to prevent subclassing? e.g prevent subclassing Deflator class.

No, you can't. There are some way to do this for abstract classes, if you put the constructor on forbiddenapis (as it's called by subclasses). For the use case here it won't work, as we have to call ctor to create the instance.

Thanks for the response to all other comments. Sorry for the issue, I was a bit harsh! 

I will now merge the other PR.

PR #1850, opened 2020-09-09
LUCENE-9517: Don't subclass Deflater and instead create a patch for setDictionary() using a functional interface

uschindler:
Bugfix for LUCENE-9517: Instead of subclassing Deflater like in LUCENE-9500, we just provide a method reference (using a functional interface), so code can call Deflater#setDictionary.

The issue was caused by a missing AccessController#setPrivileged() in JDK's code when Deflater was subclassed. This workaround does not subclass, so the missing doPrivileged is not needed.

This also adds the faulty methods to forbiddenapis.


dweiss (approved):
I liked the previous approach better - it is a pity it doesn't work, sigh. This is ok too, of course.

iverase (approved):
LGTM

jpountz:
Sorry I was on vacation last week so I couldn't review it in time. Thanks @uschindler.

PR #1860, opened 2020-09-11
LUCENE-9488 Update release process to work with gradle

madrob:
https://issues.apache.org/jira/browse/LUCENE-9488

### DIFF
--- a/dev-tools/scripts/buildAndPushRelease.py
+++ b/dev-tools/scripts/buildAndPushRelease.py
@@ -112,32 +112,21 @@ def prepare(root, version, gpgKeyID, gpgPassword):
 
   open('rev.txt', mode='wb').write(rev.encode('UTF-8'))
   
-  print('  lucene prepare-release')
-  os.chdir('lucene')
-  cmd = './gradlew -Dversion=%s' % version
+  print('  prepare-release')
+  cmd = './gradlew -Dversion=%s clean' % version
+  # TODO cd lucene ; ant dist-all?
+  # TODO cd solr ; ant package?
   if gpgKeyID is not None:
-    cmd += ' -Psigning.keyId=%s prepare-release' % gpgKeyID
-  else:
-    cmd += ' prepare-release-no-sign'
+    # TODO sign
+    # cmd += ' -Psigning.keyId=%s publishSignedPublicationToMavenLocal' % gpgKeyID
+    pass
+  cmd += ' mavenLocal'
 
   if gpgPassword is not None:
     runAndSendGPGPassword(cmd, gpgPassword)
   else:
     run(cmd)
   
-  print('  solr prepare-release')
-  os.chdir('../solr')
-  cmd = './gradlew -Dversion=%s' % version
-  if gpgKeyID is not None:
-    cmd += ' -Dsigning.keyId=%s prepare-release' % gpgKeyID
-  else:
-    cmd += ' prepare-release-no-sign'
-
-  if gpgPassword is not None:
-    runAndSendGPGPassword(cmd, gpgPassword)
-  else:
-    run(cmd)
-    
   print('  done!')
   print()
   return rev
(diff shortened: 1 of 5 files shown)


dev-tools/scripts/releaseWizard.yaml
```diff
@@ -411,7 +410,7 @@ groups:
         cmd: git pull
         stdout: true
       - !Command
-        cmd: ant clean precommit
+        cmd: ./gradlew clean check -x test
```
janhoy, line 413: Will the `./` syntax work on Windows?

janhoy (commented):
Did not test the changes, have only one comment regarding Windows compatibility

dev-tools/scripts/releaseWizard.yaml
madrob: It works in PowerShell but not Command Prompt. I don't know how important this distinction is.

HoustonPutman:
Should the smokeTester be in the scope of this PR?

madrob:
@janhoy I'm trying to run this, but the `create_stable_branch` step tries to create a new `branch_8x` for me, I think that logic needs some updating, but I'm not completely sure what needs to change. If you're able, please take a look, otherwise I'll skip that part and come back to it later.

dev-tools/scripts/releaseWizard.yaml
janhoy: You can definere a variable like we do for rename command, that used `gradlew.bat` for win?

janhoy:
There may be bugs related to major release since that has never been done with the wizard. I guess it should have created a `branch_9x`?

dev-tools/scripts/releaseWizard.yaml
madrob: Works great!

madrob:
@janhoy This isn't actually ready yet, but I'd like to merge what I have so far. The reason is that the way the script pulls from upstream to prep the RC, I kind of need changes on the actual master branch in order to proceed with further testing without resorting to some super hacky tricks locally to make it think that my fork is the place to pull from.

janhoy (approved):

janhoy:
Sure, all progress is good progress

lucene/version.properties
```diff
@@ -0,0 +1,10 @@
+# This file contains some version properties as used by various build files.
```
uschindler, line 1: Ülease remove that file, it wont be used.

If release scripts need to edit the version number after a release, <root>/build.gradle is your friend. Theres also a version.base and that's the source of information.

https://github.com/apache/lucene-solr/blob/master/build.gradle#L38

There is a reason why this file was removed, so dont restore it!

uschindler (commented):
Please revert the file addition.

dev-tools/scripts/buildAndPushRelease.py
```diff
@@ -108,37 +107,26 @@ def prepare(root, version, gpgKeyID, gpgPassword):
...
+  print('  ./gradlew -Dtests.badapples=false clean check')
+  run('./gradlew -Dtests.badapples=false clean check')
 
   open('rev.txt', mode='wb').write(rev.encode('UTF-8'))
   
-  print('  lucene prepare-release')
-  os.chdir('lucene')
-  cmd = 'ant -Dversion=%s' % version
+  print('  prepare-release')
+  cmd = './gradlew -Dversion=%s clean' % version
```
uschindler, line 116: This must be `-Dversion.release=%s`, as gradle has some reserved meaning of `version`.

dev-tools/scripts/releaseWizard.yaml
```diff
@@ -1059,7 +1056,7 @@ groups:
         logfile: checkout-release-tag.log
         tee: true
       - !Command
-        cmd: ant documentation -Dversion={{ release_version }}
+        cmd: "{{ gradle_cmd }} documentation -Dversion={{ release_version }}"
```
uschindler, line 1059: This must be `-Dversion.release=%s`, as gradle has some reserved meaning of `version`.

uschindler (commented):
Please change the property names, too.

PR #1861, opened 2020-09-12
SOLR-10391: Add overwrite option to UPLOAD ConfigSet action

tflobbe:
When set to `true`, Solr will overwrite an existing configset in ZooKeeper if an UPLOAD action happens on an existing configset. 
A new `cleanup` parameter can also be passed to let Solr know what to do with the files that existed in the old configset, but no longer exist in the new configset (remove or keep)

### DIFF
--- a/solr/core/src/java/org/apache/solr/handler/admin/ConfigSetsHandler.java
+++ b/solr/core/src/java/org/apache/solr/handler/admin/ConfigSetsHandler.java
@@ -20,9 +20,12 @@ import java.io.InputStream;
 import java.lang.invoke.MethodHandles;
 import java.nio.charset.StandardCharsets;
 import java.util.Collection;
+import java.util.Collections;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipInputStream;
@@ -38,6 +41,7 @@ import org.apache.solr.common.SolrException;
 import org.apache.solr.common.SolrException.ErrorCode;
 import org.apache.solr.common.cloud.SolrZkClient;
 import org.apache.solr.common.cloud.ZkConfigManager;
+import org.apache.solr.common.cloud.ZkMaintenanceUtils;
 import org.apache.solr.common.cloud.ZkNodeProps;
 import org.apache.solr.common.params.ConfigSetParams;
 import org.apache.solr.common.params.ConfigSetParams.ConfigSetAction;
@@ -172,9 +176,13 @@ public class ConfigSetsHandler extends RequestHandlerBase implements PermissionN
 
     // Create a node for the configuration in zookeeper
     boolean trusted = getTrusted(req);
+    Set<String> filesToDelete = Collections.emptySet();
     if (overwritesExisting) {
       if (!trusted) {
         ensureOverwritingUntrustedConfigSet(zkClient, configPathInZk);
+      }
+      if (req.getParams().getBool(ConfigSetParams.CLEANUP, false)) {
+        filesToDelete = getAllConfigsetFiles(zkClient, configPathInZk);
       }
     } else {
       zkClient.makePath(configPathInZk, ("{\"trusted\": " + Boolean.toString(trusted) + "}").
@@ -185,6 +193,11 @@ public class ConfigSetsHandler extends RequestHandlerBase implements PermissionN
     ZipEntry zipEntry = null;
     while ((zipEntry = zis.getNextEntry()) != null) {
       String filePathInZk = configPathInZk + "/" + zipEntry.getName();
+      if (filePathInZk.endsWith("/")) {
+        filesToDelete.remove(filePathInZk.substring(0, filePathInZk.length() -1));
+      } else {
+        filesToDelete.remove(filePathInZk);
+      }
       if (zipEntry.isDirectory()) {
         zkClient.makePath(filePathInZk, false,  true);
       } else {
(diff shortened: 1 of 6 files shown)


epugh:
I wish we followed REST approaches for so many of our API's...  Each one has a slightly different semantics for basic CRUD.  Wish we had:

```
CRUD       HTTP
Create      POST
Read         GET
Update     PUT
Delete      DELETE
```

HoustonPutman:
I agree with @epugh on the whole REST thing. 

I also think that the default behavior should be that files that aren't present in the new configSet should be removed. I can see the use case for keeping old files around, but I think most people would want to update all files together. I could be wrong about that though.

tflobbe:
> I wish we followed REST approaches for so many of our API's

I wish that too, but Solr APIs aren't REST as of today ([not even V2 API is fully REST, and this was intentional AFAIK](https://issues.apache.org/jira/browse/SOLR-8029?focusedCommentId=14740121&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-14740121), things like `MODIFYCOLLECTION` in V2 API are still POSTs requests). I think I'd rather keep consistency with other APIs than make this particular one better?

> I also think that the default behavior should be that files that aren't present in the new configSet should be removed. 

I was planning to add that as a different PR but also making it not-default, mostly because it can be pretty destructive (i.e. you created the zip file incorrectly, or created it with the "conf" directory on top or something like that it could remove all files) but I'm fine if people think it should be the default. Also, are you OK with doing that as a different PR or want me to do the change here? My intention was to keep the PR small for easier reviews, but if we think this feature is incomplete without that I can do it here too, I'm fine either way.

solr/core/src/java/org/apache/solr/handler/admin/ConfigSetsHandler.java
```diff
@@ -170,21 +176,90 @@ private void handleConfigUploadRequest(SolrQueryRequest req, SolrQueryResponse r
...
-        getBytes(StandardCharsets.UTF_8), true);
+    Set<String> filesToDelete = Collections.emptySet();
+    if (overwritesExisting) {
+      if (!trusted) {
+        ensureOverwritingUntrustedConfigSet(zkClient, configPathInZk);
+      }
+      if (req.getParams().getBool(ConfigSetParams.CLEANUP, false)) {
+        filesToDelete = getAllConfigsetFiles(zkClient, configPathInZk);
+      }
+    } else {
```
HoustonPutman, line 187: If you are overwriting an untrusted configSet with a trusted configSet, I din't think that will get updated here. I guess you would only want it if CLEANUP is true, since only then are you sure that all of the content is trusted.

So doing it in that if statement would probably work.

solr/core/src/java/org/apache/solr/handler/admin/ConfigSetsHandler.java
```diff
@@ -170,21 +176,90 @@ private void handleConfigUploadRequest(SolrQueryRequest req, SolrQueryResponse r
...
     while ((zipEntry = zis.getNextEntry()) != null) {
       String filePathInZk = configPathInZk + "/" + zipEntry.getName();
+      if (filePathInZk.endsWith("/")) {
+        filesToDelete.remove(filePathInZk.substring(0, filePathInZk.length() -1));
+      } else {
+        filesToDelete.remove(filePathInZk);
+      }
       if (zipEntry.isDirectory()) {
-        zkClient.makePath(filePathInZk, true);
+        zkClient.makePath(filePathInZk, false,  true);
```
HoustonPutman, line 202: We might want to clear data that the znode has if it exists and has data.

solr/core/src/java/org/apache/solr/handler/admin/ConfigSetsHandler.java
```diff
@@ -170,21 +176,90 @@ private void handleConfigUploadRequest(SolrQueryRequest req, SolrQueryResponse r
...
+    final Set<String> files = new HashSet<>();
+    if (!configPathInZk.startsWith(ZkConfigManager.CONFIGS_ZKNODE + "/")) {
+      throw new IllegalArgumentException("\"" + configPathInZk + "\" not recognized as a configset path");
+    }
+    ZkMaintenanceUtils.traverseZkTree(zkClient, configPathInZk, ZkMaintenanceUtils.VISIT_ORDER.VISIT_POST, new ZkMaintenanceUtils.ZkVisitor() {
+      @Override
+      public void visit(String path) throws InterruptedException, KeeperException {
+        files.add(path);
+      }
+    });
```
HoustonPutman, line 239: the `new ZKMaintenanceUtils.ZKVisitory() { ... }` can be replaced with `files::add`

solr/core/src/java/org/apache/solr/handler/admin/ConfigSetsHandler.java
```diff
@@ -170,21 +176,90 @@ private void handleConfigUploadRequest(SolrQueryRequest req, SolrQueryResponse r
...
+      configSetNodeContent = zkClient.getData(configSetZkPath, null, null, true);
+    } catch (KeeperException e) {
+      throw new SolrException(ErrorCode.SERVER_ERROR, "Exception while fetching current configSet at " + configSetZkPath, e);
+    } catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
+      throw new SolrException(ErrorCode.SERVER_ERROR, "Interrupted while fetching current configSet at " + configSetZkPath, e);
+    }
+    @SuppressWarnings("unchecked")
+    Map<Object, Object> contentMap = (Map<Object, Object>) Utils.fromJSON(configSetNodeContent);
+    Boolean isCurrentlyTrusted = (Boolean) contentMap.get("trusted");
```
HoustonPutman, line 259: using `getOrDefault("trusted", false)` would remove the need for the null check.

solr/core/src/java/org/apache/solr/handler/admin/ConfigSetsHandler.java
tflobbe: As it currently stands, an overwrite never updates the "trusted" flag, if it wasn't trusted before, it'll continue to be untrusted after the upgrade. If it was trusted, only an authenticated request would be allowed. I see your point that if "cleanup" is set to "true", an authenticated request could turn the configset to trusted, is that what you are suggesting?

solr/core/src/java/org/apache/solr/handler/admin/ConfigSetsHandler.java
tflobbe: Can that ever happen in a configset?

solr/core/src/java/org/apache/solr/handler/admin/ConfigSetsHandler.java
HoustonPutman: Yeah, for example if you want to use plugins or something similar, I could see someone not using a trusted configSet at first, but then wanting to overwrite their configSet to be "trusted" to be able to use the plugins.

I don't see why that shouldn't be possible, given that there's little difference security-wise between creating a new config set and overwriting w/ cleanup.

solr/core/src/java/org/apache/solr/handler/admin/ConfigSetsHandler.java
HoustonPutman: Not very likely, was just thinking about possibilities. It's probably not an issue 99.9% of times, so it likely doesn't need to be addressed.

tflobbe: Yes, I don't think this would be needed at this point. No API would set this, and no Solr component would read it. I guess the only case would be a custom component that writes and reads directly to ZooKeeper.

tflobbe:
I plan to merge this tomorrow

PR #1869, opened 2020-09-14
SOLR-14866 autowidth tables in ref guide

epugh:


HoustonPutman:
This isn't an issue anymore because we removed PDF support?

HoustonPutman:
Built the site and compared it to what's currently published, and the table looks better to me!

epugh:
So, @HoustonPutman are you voting for `[%autowidth.spread,options="header"]` being the approach to use?   I can start with that on all the TODO marked tables.

Do you think we should do it on ALL tables?

HoustonPutman:
Didn't know it was a vote, but I like the result of the new format. Are there other options or is it just autowidth vs. what is already there?

And we should probably look each of the tables after altering them to make sure they still look good, but yeah I think it's a good idea unless Cassandra disagrees.

epugh:
You may not know this, but the original norse meaning of "Eric" (or as some people like to misspell it, "Erick") is "Likes to refactor the most tedious open source code".   ;-)

I'll start with the TODO's, and glad you like the edit.  Doesn't appear to be a global style approach that would work.

solr/solr-ref-guide/src/other-parsers.adoc
```diff
@@ -237,7 +237,7 @@ Comma separated list of tags for excluding queries from parameters above. See ex
 {!bool filter=foo should=bar}
 ----
 
-Parameters might also be multivalue references. The former example above is equivlent to
```
arafalov, line 240: This and next empty space seems not needed. But 'equivlent' spelling could be fixed.

solr/solr-ref-guide/src/performance-statistics-reference.adoc
```diff
@@ -191,9 +187,7 @@ You can get the statistics shown in the table below with an API request such as
 
 The following statistics are available for each of the caches mentioned above:
 
-// TODO: Change column width to %autowidth.spread when https://github.com/asciidoctor/asciidoctor-pdf/issues/599 is fixed
-
-[cols="25,75",options="header"]
+[%autowidth.spread,width="100%",options="header"]
```
arafalov, line 190: This one gets parameter wrapped-up in weird way. I wish there was a way to mark column no-wrap specifically for this one. But it is ok otherwise.

arafalov (changes requested):
Need to resolve the merge with my PR (already submitted), remove extra spaces (2 places) and ideally fix spelling.

Overall, it looks like an improvement in majority of the cases and none examples where it got noticeably worse. Ready to go after the minor suggestions.

solr/solr-ref-guide/src/performance-statistics-reference.adoc
epugh: There was another example where I decided the autowidth was worse than the manual spaced one, so I removed the TODO but left it as a`[cols=]` style.   If you think it's better to keep the `[cols=]` let me know and I'll revert that change!

solr/solr-ref-guide/src/other-parsers.adoc
epugh: okay, fixed the typo, but left spacing to match the other spacing.

solr/solr-ref-guide/src/performance-statistics-reference.adoc
epugh: Should we just keep the old format?

ctargett:
So, the changes look fine enough to me, but I don't think they're actually needed.

The `%autowidth.spread` works by setting a `spread` class on the <table> element. This is the default, though, so it's set even without defining that attribute. The way it has been with `cols="30,70"`, or whatever it was for each table, overrode that by setting the `<colgroup>` element on each table to set the width of each column.

I suspect if you just remove the `cols` parameter and TODO note, the table will look pretty close the way it does with the new `autospread` attribute.

I figured this out by inspecting the HTML for tables that had been created after the conversion from Confluence (when these TODOs were introduced), such as the one on `aliases.adoc`, which don't have the TODO. I noticed that it gets `<table class="spread">` (plus some other classes) and the `<colgroup>` is set to an equal % for each column. Which is basically which happens with the change here.

It's also possible to avoid setting an explicit `options="header"` if all tables include an empty line between the first row and the subsequent rows. I'll leave that up to individual preference, though - the table in `aliases.adoc` does that, but others may not know to do that so an example of explicitly setting the header row isn't a negative.

epugh:
Thanks for the sleuthing @ctargett, I never thought about the case of no command.   I tested it this morning, and you are correct.

So, I think one pattern or the other for tables headings might be the way to go.  I know that I learn by looking at other examples in the ref guide, so if I see other tables using the options="header", then I'll do the same.  Likewise if I looked instead at the example in `aliases.adoc` first. 

I updated `the-standard-query-parser.adoc` to remove the autowidth.spread, and I think it is cleaner...


I'm happy to rework this PR to both remove all the TODO's and

arafalov:
Having consistent best practices encoded in the document, would be a Win from my point of view. I also look a lot at other examples, just in case our use-case is different from generic adoc advice.

ctargett:
bq. Having consistent best practices encoded in the document

The place that codifies our asciidoc best practices is in `meta-docs/asciidoc-syntax.adoc`, which is published with every Ref Guide as part of `how-to-contribute.adoc`. It in fact says to use the `options="header"` syntax for header rows, but to be honest I don't see any problem with someone using any valid syntax that ends up with the same HTML style.

epugh:
Thanks for pointing that out.  So, should we change it up in the asciidoc-syntax.adoc to remove the options="header", and then start progressivley going with the simpler style?   I like how the simpler style removes a lot of thinking, and maybe we highlight to only use the more complex style when you need "specific control over the column width in tables".   There are a couple of places where the autowidth doesnt' look good, places where, if we had a nowrap type setting we would use that in conjuction with autowidth.

Again, happy to do the leg work, I know it grows the scope of this PR beyond removing those TODO's to revamping all tables.

ctargett:
I don't view the syntax of "leave a blank line after the first line to make it a header row" as simpler (which is what I think you mean), at least it isn't for me. Until I looked it up yesterday, I completely forgot it was an option.

To be clear, I don't think it's worth the effort to try to standardize all tables to use the same syntax nor to spend time policing what form of the syntax people choose to use. The flexibility of the syntax is one of its strengths, and we could bikeshed all kinds of options but we all have better things to do. Like if you have that much time for Ref Guide edits, I could suggest several other things to work on instead. We should encourage people to review our guidelines and use them, but at the same time we should support those who are already comfortable doing things their particular way.

Re: the wrapping. It appears if we add `white-space: nowrap` (or `white-space: pre`, both prevent wrapping) to the CSS for the `<td>` (and maybe also `<th>`) element (`ref-guide.css` L#996 or thereabouts), that forces text without spaces not to break.

epugh:
Okay, so I think I'm done, but would love one more set of eyes!  I did tweak the `asciidoc-syntax.adoc` and put in the autowidth approach.  What we had listed actually wasn't ever used in the docs ;-)   Please let me know what you think.

ctargett:
I'm a little confused what's going on here at this point. It seems the latest commits increase the mixing of styles, which as I said is fine with me in general, but sort of counter to what you said you wanted to do.

There are 3 tables in `charfilterfactories.adoc` for example, and two are now changed to remove the `[options="header"]` and add a blank line between header row and data rows, but the header param remains on the 3rd table which also includes the `%autowidth.spread,width="100%"` parameters which I thought I showed aren't necessary and are already the default.

Despite all that, the `asciidoc-syntax.adoc` is updated to imply the best practice is defining the autospread, width, and header params explicitly, even though other tables in this PR were changed to definitely not do that. It stops short of saying it's a best practice, though, it just points out that one might see it (when before it was more direct in what it recommended writers to do).

So, I'm no longer sure what's really being achieved here. Removing TODOs that aren't needed anymore because they're holdovers from PDF days is a laudable goal and helpful cleanup, but is it any more clear now what both the internal docs and examples within the content would recommend someone do to define tables? I know that wasn't the original intent, but it got turned into a goal and now it's not doing that. At this point I'm not sure if I should approve the PR (since I don't think it's ultimately that big a deal) or if I should look to hold the PR to some level of internal consistency?

PR #1870, opened 2020-09-14
SOLR-14865: 'Index Merge Metrics' documentation correction

cpoerschke:
https://issues.apache.org/jira/browse/SOLR-14865

### DIFF
--- a/solr/solr-ref-guide/src/metrics-reporting.adoc
+++ b/solr/solr-ref-guide/src/metrics-reporting.adoc
@@ -534,7 +534,26 @@ These metrics are available only on a per-core basis. Metrics can be aggregated 
 
 These metrics are collected in respective registries for each core (e.g., `solr.core.collection1....`), under the `INDEX` category.
 
-Basic metrics are always collected - collection of additional metrics can be turned on using boolean parameters in the `/config/indexConfig/metrics` section of `solrconfig.xml`:
+Metrics collection is controlled by boolean parameters in the `/config/indexConfig/metrics` section of `solrconfig.xml`:
+
+Basic metrics:
+
+[source,xml]
+----
+<config>
+  ...
+  <indexConfig>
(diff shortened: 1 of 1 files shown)


cpoerschke:
@ctargett and/or @sigram would you have any thoughts on this documentation change, with a view towards it perhaps being included in the upcoming 8.7 Solr Reference Guide? Thanks!

solr/solr-ref-guide/src/metrics-reporting.adoc
```diff
@@ -534,15 +534,34 @@ These metrics are available only on a per-core basis. Metrics can be aggregated
 
 These metrics are collected in respective registries for each core (e.g., `solr.core.collection1....`), under the `INDEX` category.
 
-Basic metrics are always collected - collection of additional metrics can be turned on using boolean parameters in the `/config/indexConfig/metrics` section of `solrconfig.xml`:
+Metrics collection is controlled by boolean parameters in the `/config/indexConfig/metrics` section of `solrconfig.xml`:
+
```
ctargett, line 538: I was a little confused by the fully qualified XML path here, we don't do that in other places in the docs so it jumped out at me. If we think maybe we should do that more often, this would be fine and I can make an issue to update other places in the Guide to someday be consistent but otherwise I only wonder if it would be jarring for others also (I might be alone in my impression).

cpoerschke: Interesting find! Yes, I can see that above the _"The metrics available in your system can be customized by modifying the `<metrics>` element in `solr.xml`."_ wording is used and `<metrics>` is more pleasant to read than `/config/indexConfig/metrics` though that's subjective of course. The context helps differentiate between the different `<metrics>` elements and so yes, I agree it would be less jarring (and shorter!) to not use the fully qualified XML path here.

PR #1872, opened 2020-09-14
LUCENE-9510: Don't pull a merge instance when flushing stored fields out-of-order.

jpountz:
With recent changes to stored fields that split blocks into several sub
blocks, the merge instance has become much slower at random access since
it would decompress all sub blocks when accessing a document. Since
stored fields likely get accessed in random order at flush time when
index sorting is enabled, it's better not to use the merge instance.

On a synthetic benchmark that has one stored field and one numeric
doc-value field that is used for sorting and fed with random values,
this made indexing more than 4x faster.


jimczi (approved):
lgtm

PR #1874, opened 2020-09-15
LUCENE-9510: Don't compress temporary stored fields and term vectors when index sorting is enabled.

jpountz:
When index sorting is enabled, stored fields and term vectors can't be
written on the fly like in the normal case, so they are written into
temporary files that then get resorted. For these temporary files,
disabling compression speeds up indexing significantly.

On a synthetic test that indexes stored fields and a doc value field
populated with random values that is used for index sorting, this
resulted in a 3x indexing speedup.


jimczi (approved):
Nice speedup! LGTM

PR #1877, opened 2020-09-16
SOLR-13181: param macro expansion could throw

dsmiley:
StringIndexOutOfBoundsException on bad syntax

TODO run tests etc.

https://issues.apache.org/jira/browse/SOLR-13181

### DIFF
--- a/solr/core/src/java/org/apache/solr/request/macro/MacroExpander.java
+++ b/solr/core/src/java/org/apache/solr/request/macro/MacroExpander.java
@@ -144,7 +144,7 @@ public class MacroExpander {
       final int matchedStart = idx;
       idx += macroStart.length();
 
-      int rbrace = val.indexOf('}', idx);
+      int rbrace = val.indexOf('}', matchedStart + macroStart.length());
       if (rbrace == -1) {
         // no matching close brace...
         continue;
--- a/solr/core/src/java/org/apache/solr/request/macro/MacroExpander.java
+++ b/solr/core/src/java/org/apache/solr/request/macro/MacroExpander.java
@@ -142,7 +142,6 @@ public class MacroExpander {
 
       // found unescaped "${"
       final int matchedStart = idx;
-      idx += macroStart.length();
 
       int rbrace = val.indexOf('}', matchedStart + macroStart.length());
       if (rbrace == -1) {
@@ -161,7 +160,7 @@ public class MacroExpander {
       // update "start" to be at the end of ${...}
       idx = start = rbrace + 1;
 
-      // String in-between = val.substring(idx, rbrace);
+      // String in-between braces
       StrParser parser = new StrParser(val, matchedStart + macroStart.length(), rbrace);
       try {
         String paramName = parser.getId();
--- a/solr/core/src/java/org/apache/solr/request/macro/MacroExpander.java
+++ b/solr/core/src/java/org/apache/solr/request/macro/MacroExpander.java
@@ -119,8 +119,8 @@ public class MacroExpander {
     int start = 0;  // start of the unprocessed part of the string
     StringBuilder sb = null;
     for (;;) {
+      assert idx >= start;
       idx = val.indexOf(macroStart, idx);
-      int matchedStart = idx;
 
       // check if escaped
       if (idx > 0) {
@@ -136,11 +136,12 @@ public class MacroExpander {
       }
       else if (idx < 0) {
         if (sb == null) return val;
-        sb.append(val.substring(start));
+        sb.append(val, start, val.length());
         return sb.toString();
       }
 
       // found unescaped "${"
+      int matchedStart = idx;
       idx += macroStart.length();
 
       int rbrace = val.indexOf('}', idx);
@@ -154,14 +155,14 @@ public class MacroExpander {
       }
 
       if (matchedStart > 0) {
-        sb.append(val.substring(start, matchedStart));
+        sb.append(val, start, matchedStart);
       }
 
       // update "start" to be at the end of ${...}
-      start = rbrace + 1;
+      idx = start = rbrace + 1;
 
-      // String inbetween = val.substring(idx, rbrace);
-      StrParser parser = new StrParser(val, idx, rbrace);
+      // String in-between = val.substring(idx, rbrace);
+      StrParser parser = new StrParser(val, matchedStart + macroStart.length(), rbrace);
       try {
         String paramName = parser.getId();
         String defVal = null;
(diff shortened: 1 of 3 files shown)


solr/core/src/java/org/apache/solr/request/macro/MacroExpander.java
```diff
@@ -136,11 +136,12 @@ private String _expand(String val) {
       }
       else if (idx < 0) {
         if (sb == null) return val;
-        sb.append(val.substring(start));
+        sb.append(val, start, val.length());
         return sb.toString();
       }
 
       // found unescaped "${"
+      int matchedStart = idx;
```
cpoerschke, line 144:
```suggestion
      final int matchedStart = idx;
```

solr/core/src/java/org/apache/solr/request/macro/MacroExpander.java
```diff
@@ -154,14 +155,14 @@ else if (idx < 0) {
...
+        sb.append(val, start, matchedStart);
       }
 
       // update "start" to be at the end of ${...}
-      start = rbrace + 1;
+      idx = start = rbrace + 1;
 
-      // String inbetween = val.substring(idx, rbrace);
-      StrParser parser = new StrParser(val, idx, rbrace);
+      // String in-between = val.substring(idx, rbrace);
```
cpoerschke, line 164: How about removing rather than amending the `// String inbetween = val.substring(idx, rbrace);` commented out code line?

solr/core/src/java/org/apache/solr/request/macro/MacroExpander.java
```diff
@@ -136,11 +136,12 @@ private String _expand(String val) {
...
-        sb.append(val.substring(start));
+        sb.append(val, start, val.length());
         return sb.toString();
       }
 
       // found unescaped "${"
+      int matchedStart = idx;
       idx += macroStart.length();
 
       int rbrace = val.indexOf('}', idx);
```
cpoerschke, line 147:
```suggestion
      int rbrace = val.indexOf('}', matchedStart + macroStart.length());
```

similar to line 165 below and then the `idx += macroStart.length();` assignment would not be needed since there would be no use of `idx` between that incrementing assigment and the `idx = start = rbrace + 1;` resetting assignment below.

solr/core/src/java/org/apache/solr/request/macro/MacroExpander.java
```diff
@@ -188,7 +189,7 @@ else if (failOnMissingParams) {
 
       } catch (SyntaxError syntaxError) {
         // append the part we would have skipped
-        sb.append( val.substring(matchedStart, start) );
+        sb.append(val, matchedStart, start);
```
cpoerschke, line 192: observation: `foo.append(bar.substring(x,y));` is also found in other places in the code base, not sure how it might work implementation wise but it would be lovely if tooling would flag up that there's a more efficient alternative

dsmiley: IntelliJ pointed this out to me :-)  It didn't know about the val.substring(start) (implied val.length()) so I did those manually.  Yes, it'd be nice if everywhere we could do this in one go.  Feel free to file an issue to do so if you have time.  IntelliJ can make quick work of that.

solr/core/src/java/org/apache/solr/request/macro/MacroExpander.java
dsmiley, line 164:
```suggestion
      // String in-between braces
```

solr/core/src/java/org/apache/solr/request/macro/MacroExpander.java
```diff
@@ -136,14 +136,15 @@ private String _expand(String val) {
...
       else if (idx < 0) {
         if (sb == null) return val;
-        sb.append(val.substring(start));
+        sb.append(val, start, val.length());
         return sb.toString();
       }
 
       // found unescaped "${"
+      final int matchedStart = idx;
       idx += macroStart.length();
```
dsmiley, line 145: Can now remove idx manipulation until later when idx & start are changed in one line
```suggestion
```

cpoerschke (approved):

solr/core/src/test/org/apache/solr/request/macro/TestMacroExpander.java
```diff
@@ -143,15 +142,31 @@ public void testMapExprExpandOn() {
...
     } finally {
       System.setProperty("StreamingExpressionMacros", oldVal);
     }
   }
+
+  @Test
+  public void testUnbalanced() { // SOLR-13181
+    final MacroExpander meSkipOnMissingParams = new MacroExpander(Collections.emptyMap());
+    final MacroExpander meFailOnMissingParams = new MacroExpander(Collections.emptyMap(), true);
+    assertEquals("${noClose", meSkipOnMissingParams.expand("${noClose"));
```
cpoerschke, line 160: How about adding (say) `${goodAnswer} ${noClose` as another case and it turning into `42 ${noClose` if there is a `goodAnswer=42` mapping i.e. to demonstrate that absence of closing curly bracket partially affects expansion (as opposed to no expansion at all happening)?

dsmiley: I'm not 100% clear what you suggest; let me try to guess.  Let's say the parameter map has exactly one param == goodAnswer with value 42.
````
    assertEquals("42", meSkipOnMissingParams.expand("${goodAnswer}"));
    assertEquals("42", meFailOnMissingParams.expand("${goodAnswer}"));
````
Is that what you suggest?

cpoerschke: `${goodAnswer} ${noClose` is what I had in mind. Pushed a corresponding commit to the PR's branch, hope you don't mind.

gus-asf:
Did see your review request, but my life is doing that whole "it never rains but it pours" thing right now and I keep not getting to this. My primary concern wrt to this class is the need to ensure that default installations do not expand streaming expression parameters (but folks who need that for existing installs can turn it on with knowledge of the risks). I saw your comment on the ticket that fixed that originally, and yeah calling out a single parameter is yucky, but it was the least back compat breaking thing I could come up with. A more sensible, less backwards compatible solution for 9x is certainly a possibility. I may not have time to do any deep review here soon enough, so don't wait on me.

solr/core/src/test/org/apache/solr/request/macro/TestMacroExpander.java
dsmiley: Ah; thanks!

dsmiley:
@gus-asf (in reference to you're fix in SOLR-12891):  I can appreciate you didn't want to break back-compat.  In this issue/PR, I'm not getting at that matter at all, it's a separate bug fix when the syntax is given wrong.

PR #1881, opened 2020-09-17
SOLR-14876: Upgrade to zookeeper 3.6.2 

odidev:
The update netty ver 4.1.50 includes both security fixes and AArch64 performance improvements
Refer release notes for detail: https://netty.io/news/2020/05/13/4-1-50-Final.html

Signed-off-by: odidev <odidev@puresoftware.com>


ErickErickson:
Does this have an associated Solr JIRA? I couldn't find one...

ErickErickson:
Hmmm, All the jars for netty that ship with Solr are already 4.1.50.Final, it's an explicit dependency in versions.props.

Of course if you run an external ZK, you can download any version you choose.

That said, there's no reason _not_ to upgrade ZK, I just doubt getting 4.1.50 is germane.

ErickErickson:
This has been fixed in Solr 8.7...

PR #1882, opened 2020-09-17
LUCENE-9529: Track dirtiness of stored fields via a number of docs, not chunks.

jpountz:
The problem of tracking dirtiness via numbers of chunks is that larger
chunks make stored fields readers more likely to be considered dirty, so
I'm trying to work around it by tracking numbers of docs instead.


rmuir:
we may want to apply new logic to the CompressingTermVectorsWriter as well for consistency?

jpountz:
Agreed, I just pushed a change to take care of term vectors as well that I had locally.

rmuir:
+1

We can always try to improve this thing later in subsequent issues. This still keeps a hard safety net on the number of dirty chunks, so I can't see it introducing any new worst-case behavior

PR #1889, opened 2020-09-18
LUCENE-9534: Ensure DWPT#ramBytesUsed is only called unter lock

s1monw:
Consumers of the used RAM of a DWPT should use it's committed bytesUsed
value that's threadsafe.


mikemccand (approved):

jpountz (approved):
Thanks!

PR #1890, opened 2020-09-18
Rename ConfigSetsAPITest to TestConfigSetsAPIShareSchema

cpoerschke:
As part of https://issues.apache.org/jira/browse/LUCENE-8626 I noticed that we have both

* https://github.com/apache/lucene-solr/blob/releases/lucene-solr/8.6.2/solr/core/src/test/org/apache/solr/cloud/ConfigSetsAPITest.java and
* https://github.com/apache/lucene-solr/blob/releases/lucene-solr/8.6.2/solr/core/src/test/org/apache/solr/cloud/TestConfigSetsAPI.java which is mildly confusing.

I also note that
* https://github.com/apache/lucene-solr/blob/releases/lucene-solr/8.6.2/solr/core/src/test/org/apache/solr/cloud/TestConfigSetsAPIExclusivity.java and
* https://github.com/apache/lucene-solr/blob/releases/lucene-solr/8.6.2/solr/core/src/test/org/apache/solr/cloud/TestConfigSetsAPIZkFailure.java exist in the same package i.e. this part of the code base favours `Test*.java` naming.

Something like "TestConfigSetsAPIShareSchema" more accurately describes what the "ConfigSetsAPITest" test does.


tflobbe:
Christine, if you could give me a couple days, I'm doing some refactor on `TestConfigSetsAPI` to improve and add some more tests

tflobbe:
I refer to https://github.com/apache/lucene-solr/pull/1892

cpoerschke:
> ... if you could give me a couple days ... #1892

Sure, no problem at all.

Now that `TestConfigSetsAPI` extends `SolrCloudTestCase` too, I wonder
* if `TestConfigSetsAPISolrCloud` would still be a good replacement for the `ConfigSetsAPITest` still, or
* if adding of `ConfigSetsAPITest` functionality to `TestConfigSetsAPI` might be better?

(Haven't looked at details as yet.)

cpoerschke:
> ... if adding of `ConfigSetsAPITest` functionality to `TestConfigSetsAPI` might be better? ...

Looking more closely, there seem to be sufficient differences between the tests i.e. them remaining separate they would be clearer. However, `TestConfigSetsAPIShareSchema` could be a better alternative name than `TestConfigSetsAPISolrCloud`.

PR #1891, opened 2020-09-18
SOLR-14877: Add github action for running SolrJ tests.

HoustonPutman:
https://issues.apache.org/jira/browse/SOLR-14877

### DIFF
--- a/.github/workflows/solrj-test.yml
+++ b/.github/workflows/solrj-test.yml
@@ -1,0 +1,27 @@
+name: SolrJ Tests
+
+on:
+  pull_request:
+    branches:
+      - 'master'
+    paths:
+      - '.github/workflows/solrj-test.yml'
+      - 'solr/solrj/**'
+
+jobs:
+  test:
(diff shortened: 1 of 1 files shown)


.github/workflows/solrj-test.yml
```diff
@@ -0,0 +1,27 @@
...
+    # Setup
+    - uses: actions/checkout@v2
+    - name: Set up JDK 11
+      uses: actions/setup-java@v1
+      with:
+        java-version: 11
+    - name: Grant execute permission for gradlew
+      run: chmod +x gradlew
+    - name: Test the SolrJ Package
+      run: ./gradlew solr:solrj:test
```
madrob, line 27: Do we need to consider that this test run won't have our generated gradle properties?

HoustonPutman: I copied the same format as the precommit action, is there something we need to add to use the right properties?

madrob (approved):

PR #1893, opened 2020-09-19
LUCENE-9444 Utility class to get facet labels from taxonomy for a fac…

goankur:
…et field. This is useful if a facet field is requested in the list of return fields for each hit.

# Description

This pull request adds a convenience class that allows an application to get facet labels for a facet field without knowing the internals of faceting implementation.

# Solution

Simple class with APIs to get facet labels for a facet field and (optionally) a facet dimension

# Tests

Tests were added to demonstrate expected API usage and ensure correctness of implementation

# Checklist

### DIFF
--- a/lucene/facet/src/java/org/apache/lucene/facet/taxonomy/TaxonomyFacetLabels.java
+++ b/lucene/facet/src/java/org/apache/lucene/facet/taxonomy/TaxonomyFacetLabels.java
@@ -16,7 +16,6 @@
  */
 package org.apache.lucene.facet.taxonomy;
 
-import org.apache.lucene.facet.FacetsConfig;
 import org.apache.lucene.index.LeafReaderContext;
 import org.apache.lucene.util.IntsRef;
 
@@ -43,10 +42,6 @@ public class TaxonomyFacetLabels {
    */
   private final TaxonomyReader taxoReader;
 
-  /**
-   * {@code FacetsConfig} provided to the constructor
-   */
-  private final FacetsConfig config;
 
   /**
    * {@code OrdinalsReader} to decode ordinals previously indexed into the {@code BinaryDocValues} facet field
@@ -56,9 +51,8 @@ public class TaxonomyFacetLabels {
   /**
    * Sole constructor.  Do not close the provided {@link TaxonomyReader} while still using this instance!
    */
-  public TaxonomyFacetLabels(TaxonomyReader taxoReader, FacetsConfig config, String indexFieldName) throws IOException {
+  public TaxonomyFacetLabels(TaxonomyReader taxoReader, String indexFieldName) throws IOException {
     this.taxoReader = taxoReader;
-    this.config = config;
     this.indexFieldName = indexFieldName;
     this.ordsReader = new DocValuesOrdinalsReader(indexFieldName);
   }
--- a/lucene/facet/src/java/org/apache/lucene/facet/taxonomy/TaxonomyFacetLabels.java
+++ b/lucene/facet/src/java/org/apache/lucene/facet/taxonomy/TaxonomyFacetLabels.java
@@ -1,0 +1,184 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.lucene.facet.taxonomy;
+
+import org.apache.lucene.facet.FacetsConfig;
+import org.apache.lucene.index.LeafReaderContext;
+import org.apache.lucene.util.IntsRef;
+
+import java.io.IOException;
+
+import static org.apache.lucene.facet.taxonomy.TaxonomyReader.INVALID_ORDINAL;
+import static org.apache.lucene.facet.taxonomy.TaxonomyReader.ROOT_ORDINAL;
+
+/**
+ * Utility class to easily retrieve previously indexed facet labels, allowing you to skip also adding stored fields for these values,
+ * reducing your index size.
+ *
+ * @lucene.experimental
+ **/
+public class TaxonomyFacetLabels {
+
+  /**
+   * Index field name provided to the constructor
+   */
+  private final String indexFieldName;
+
+  /**
+   * {@code TaxonomyReader} provided to the constructor
+   */
+  private final TaxonomyReader taxoReader;
+
+  /**
+   * {@code FacetsConfig} provided to the constructor
+   */
+  private final FacetsConfig config;
+
+  /**
+   * {@code OrdinalsReader} to decode ordinals previously indexed into the {@code BinaryDocValues} facet field
+   */
+  private final OrdinalsReader ordsReader;
+
+  /**
+   * Sole constructor.  Do not close the provided {@link TaxonomyReader} while still using this instance!
+   */
+  public TaxonomyFacetLabels(TaxonomyReader taxoReader, FacetsConfig config, String indexFieldName) throws IOException {
+    this.taxoReader = taxoReader;
+    this.config = config;
+    this.indexFieldName = indexFieldName;
+    this.ordsReader = new DocValuesOrdinalsReader(indexFieldName);
+  }
+
+  /**
+   * Create and return an instance of {@link FacetLabelReader} to retrieve facet labels for
+   * multiple documents and (optionally) for a specific dimension.  You must create this per-segment,
+   * and then step through all hits, in order, for that segment.
+   *
+   * <p><b>NOTE</b>: This class is not thread-safe, so you must use a new instance of this
+   * class for each thread.</p>
+   *
+   * @param readerContext LeafReaderContext used to access the {@code BinaryDocValues} facet field
+   * @return an instance of {@link FacetLabelReader}
+   * @throws IOException when a low-level IO issue occurs
+   */
+  public FacetLabelReader getFacetLabelReader(LeafReaderContext readerContext) throws IOException {
+    return new FacetLabelReader(ordsReader, readerContext);
+  }
+
+  /**
+   * Utility class to retrieve facet labels for multiple documents.
+   *
+   * @lucene.experimental
+   */
+  public class FacetLabelReader {
+    private final OrdinalsReader.OrdinalsSegmentReader ordinalsSegmentReader;
+    private final IntsRef decodedOrds = new IntsRef();
+    private int currentDocId = -1;
+    private int currentPos = -1;
+
+    // Lazily set when nextFacetLabel(int docId, String facetDimension) is first called
+    private int[] parents;
+
+    /**
+     * Sole constructor.
+     */
+    public FacetLabelReader(OrdinalsReader ordsReader, LeafReaderContext readerContext) throws IOException {
+      ordinalsSegmentReader = ordsReader.getReader(readerContext);
+    }
+
+    /**
+     * Retrieves the next {@link FacetLabel} for the specified {@code docId}, or {@code null} if there are no more.
+     * This method has state: if the provided {@code docId} is the same as the previous invocation, it returns the
+     * next {@link FacetLabel} for that document.  Otherwise, it advances to the new {@code docId} and provides the
+     * first {@link FacetLabel} for that document, or {@code null} if that document has no indexed facets.  Each
+     * new {@code docId} must be in strictly monotonic (increasing) order.
+     *
+     * @param docId input docId provided in monotonic (non-decreasing) order
+     * @return the first or next {@link FacetLabel}, or {@code null} if there are no more
+     * @throws IOException when a low-level IO issue occurs
+     */
+    public FacetLabel nextFacetLabel(int docId) throws IOException {
+      if (currentDocId != docId) {
+        assert docId > currentDocId: "docs out of order!  currentDocId=" + currentDocId + " docId=" + docId;
+        ordinalsSegmentReader.get(docId, decodedOrds);
+        currentDocId = docId;
+        currentPos = decodedOrds.offset;
+      }
+
+      int endPos = decodedOrds.offset + decodedOrds.length;
+      assert currentPos <= endPos;
+
+      if (currentPos == endPos) {
+        // no more FacetLabels
+        return null;
+      }
+
+      int ord = decodedOrds.ints[currentPos++];
+      return taxoReader.getPath(ord);
+    }
+
+    private boolean isDescendant(int ord, int ancestorOrd) {
+      while (ord != INVALID_ORDINAL && ord != ROOT_ORDINAL) {
+        if (parents[ord] == ancestorOrd) {
+          return true;
+        }
+        ord = parents[ord];
+      }
+      return false;
+    }
+
+    /**
+     * Retrieves the next {@link FacetLabel} for the specified {@code docId} under the requested {@code facetDimension},
+     * or {@code null} if there are no more. This method has state: if the provided {@code docId} is the same as the
+     * previous invocation, it returns the next {@link FacetLabel} for that document.  Otherwise, it advances to
+     * the new {@code docId} and provides the first {@link FacetLabel} for that document, or {@code null} if that document
+     * has no indexed facets.  Each new {@code docId} must be in strictly monotonic (increasing) order.
+     *
+     * <p><b>NOTE</b>: this method loads the {@code int[] parents} array from the taxonomy index</p>
+     *
+     * @param docId input docId provided in non-decreasing order
+     * @return the first or next {@link FacetLabel}, or {@code null} if there are no more
+     * @throws IOException if {@link TaxonomyReader} has problems getting path for an ordinal
+     */
+    public FacetLabel nextFacetLabel(int docId, String facetDimension) throws IOException {
+      final int parentOrd = taxoReader.getOrdinal(new FacetLabel(facetDimension));
+      assert parentOrd != INVALID_ORDINAL : "Category ordinal not found for facet dimension: " + facetDimension;
+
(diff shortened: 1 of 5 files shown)


lucene/facet/src/java/org/apache/lucene/facet/taxonomy/TaxonomyFacetLabels.java
```diff
@@ -0,0 +1,184 @@
...
+     * first {@link FacetLabel} for that document, or {@code null} if that document has no indexed facets.  Each
+     * new {@code docId} must be in strictly monotonic (increasing) order.
+     *
+     * @param docId input docId provided in monotonic (non-decreasing) order
+     * @return the first or next {@link FacetLabel}, or {@code null} if there are no more
+     * @throws IOException when a low-level IO issue occurs
+     */
+    public FacetLabel nextFacetLabel(int docId) throws IOException {
+      if (currentDocId != docId) {
+        assert docId > currentDocId: "docs out of order!  currentDocId=" + currentDocId + " docId=" + docId;
```
mikemccand, line 116: Hmm let's upgrade this to a real `if` not an `assert` (that requires that assertions are enabled), and change to throw `IllegalArgumentException`?  In general if it is a problem that a user could legitimately trip up on, it should be a real `if`, as long as performance impact of such micro-policing is acceptable (which should be the case here).

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/TaxonomyFacetLabels.java
```diff
@@ -0,0 +1,184 @@
...
+        ordinalsSegmentReader.get(docId, decodedOrds);
+        currentDocId = docId;
+        currentPos = decodedOrds.offset;
+      }
+
+      int endPos = decodedOrds.offset + decodedOrds.length;
+      assert currentPos <= endPos;
```
mikemccand, line 123: This one is a good usage of `assert` because this should only be violated if there is a bug in this code or our default `Codec`, etc.  No accidental abuse by users could cause this.

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/TaxonomyFacetLabels.java
```diff
@@ -0,0 +1,184 @@
...
+    }
+
+    /**
+     * Retrieves the next {@link FacetLabel} for the specified {@code docId}, or {@code null} if there are no more.
+     * This method has state: if the provided {@code docId} is the same as the previous invocation, it returns the
+     * next {@link FacetLabel} for that document.  Otherwise, it advances to the new {@code docId} and provides the
+     * first {@link FacetLabel} for that document, or {@code null} if that document has no indexed facets.  Each
+     * new {@code docId} must be in strictly monotonic (increasing) order.
+     *
+     * @param docId input docId provided in monotonic (non-decreasing) order
```
mikemccand, line 110: Can we add a **Note** here, that the returned `FacetLabel` are not necessarily in the same order in which they were indexed?

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/TaxonomyFacetLabels.java
```diff
@@ -0,0 +1,184 @@
...
+    }
+
+    /**
+     * Retrieves the next {@link FacetLabel} for the specified {@code docId} under the requested {@code facetDimension},
+     * or {@code null} if there are no more. This method has state: if the provided {@code docId} is the same as the
+     * previous invocation, it returns the next {@link FacetLabel} for that document.  Otherwise, it advances to
+     * the new {@code docId} and provides the first {@link FacetLabel} for that document, or {@code null} if that document
+     * has no indexed facets.  Each new {@code docId} must be in strictly monotonic (increasing) order.
+     *
+     * <p><b>NOTE</b>: this method loads the {@code int[] parents} array from the taxonomy index</p>
```
mikemccand, line 151: Add same **Note** here as above comment?

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/TaxonomyFacetLabels.java
```diff
@@ -0,0 +1,184 @@
...
+     *
+     * @param docId input docId provided in non-decreasing order
+     * @return the first or next {@link FacetLabel}, or {@code null} if there are no more
+     * @throws IOException if {@link TaxonomyReader} has problems getting path for an ordinal
+     */
+    public FacetLabel nextFacetLabel(int docId, String facetDimension) throws IOException {
+      final int parentOrd = taxoReader.getOrdinal(new FacetLabel(facetDimension));
+      assert parentOrd != INVALID_ORDINAL : "Category ordinal not found for facet dimension: " + facetDimension;
```
mikemccand, line 159: Can we upgrade this to a real `if`?  Caller might legitimately pass in a non-existing `facetDimension`?

lucene/facet/src/test/org/apache/lucene/facet/taxonomy/TestTaxonomyLabels.java
```diff
@@ -0,0 +1,192 @@
...
+    assertTrue(facetLabels.stream()
+        .filter(l -> "Publish Date".equals(l.components[0]))
+        .map(l -> String.join("/", l.components[1], l.components[2], l.components[3]))
+        .collect(Collectors.toSet())
+        .equals(Set.of("2010/10/15", "2010/10/20", "2012/1/1", "2012/1/7", "1999/5/5")));
+
+    try {
+      facetLabels = lookupFacetLabels(taxoLabels, fc.getMatchingDocs(), null, true);
+      fail("Assertion error was not thrown for using docIds supplied in decreasing order");
+    } catch (AssertionError ae) {
```
mikemccand, line 178: Ahh, be careful here!  Lucene's `test-framework` sometimes randomly runs without assertions enabled, to help us catch bugs where we accidentally create code relying on assertions (it has happened, do not ask who).

So when tests run with assertions disabled then this test would false-fail?  But, if we upgrade to real `if` that would also fix this.

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/TaxonomyFacetLabels.java
```diff
@@ -0,0 +1,184 @@
...
+        parents = taxoReader.getParallelTaxonomyArrays().parents();
+      }
+
+      int endPos = decodedOrds.offset + decodedOrds.length;
+      assert currentPos <= endPos;
+
+      for (; currentPos < endPos; ) {
+        int ord = decodedOrds.ints[currentPos++];
+        if (isDescendant(ord, parentOrd) == true) {
+          return taxoReader.getPath(ord);
```
mikemccand, line 178: Hmm when we switch to `BinaryDocValues` for this lookup, we could fix this class to persistently hold a single `BinaryDocValues`, and improve performance for caller that need to look up multiple ordinals, which is the common case.

lucene/facet/src/test/org/apache/lucene/facet/taxonomy/TestTaxonomyLabels.java
```diff
@@ -0,0 +1,192 @@
...
+          facetLabels.add(facetLabel);
+        }
+      }
+    }
+
+    return facetLabels;
+  }
+
+
+  public void testBasic() throws Exception {
```
mikemccand, line 127: Could we make a small change to an existing randomized test, maybe `TestTaxonomyFacetCounts.testRandom`, to use this API to randomly retrieve N facet labels?  (So we exercise the class beyond the simplish `testBasic` ... maybe we uncover a rare bug, somewhere).

mikemccand (commented):
I think this is really close!  Thanks @goankur!

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/TaxonomyFacetLabels.java
goankur: Changed in the next revision.

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/TaxonomyFacetLabels.java
goankur: Got it thanks for explaining appropriate usage of assert and `IlegalArgumentException`

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/TaxonomyFacetLabels.java
goankur: done in next revision

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/TaxonomyFacetLabels.java
goankur: done in next revision

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/TaxonomyFacetLabels.java
goankur: done in next revision.

lucene/facet/src/test/org/apache/lucene/facet/taxonomy/TestTaxonomyLabels.java
goankur: Nice catch, I changed the implementation to throw `IllegalArgumentExceeption` after ugrade to `if` :-)

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/TaxonomyFacetLabels.java
goankur: By this class you mean `TaxonomyReader` because that is where `BinaryDocValues` will be cached as part of [LUCENE-9476](https://issues.apache.org/jira/browse/LUCENE-9476) improving performance for callers that need to lookup multiple ordinals ?

lucene/facet/src/test/org/apache/lucene/facet/taxonomy/TestTaxonomyLabels.java
goankur: Done in this revision

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/TaxonomyFacetLabels.java
mikemccand: Hmm, I was thinking this new per-segment class (`FacetLabelReader`) would hold onto a single `BinaryDocValues` instance, but you're right, that won't work.

The facet ordinals are stored / compacted in sorted order in a single `BinaryDocValues` field which we decode here into `decodedOrds`.  Resolving those ordinals to `FacetLabel` is costly, currently retrieving one stored document per ord.  After LUCENE-9450 (switching from stored fields to BDV), we could pull a new BDV for each unique `docId` passed to `nextFacetLabel`, and then bulk resolve all the ordinals?

Anyway, we can pursue all of this later -- this PR already looks awesome -- progress not perfection!

lucene/facet/src/test/org/apache/lucene/facet/FacetTestCase.java
```diff
@@ -56,6 +60,28 @@ public Facets getTaxonomyFacetCounts(TaxonomyReader taxoReader, FacetsConfig con
     return facets;
   }
 
+  public List<List<FacetLabel>> getTaxonomyFacetLabels(TaxonomyReader taxoReader, FacetsConfig config, FacetsCollector fc) throws IOException {
```
mikemccand, line 63: Thank you for adding this utility method so tests can easily use the new utility class!

Can we rename this to `getAllTaxonomyFacetLabels`, and add javadoc explaining that the outer list is one entry per matched hit, and the inner list is one entry per `FacetLabel` belonging to that hit?

lucene/facet/src/test/org/apache/lucene/facet/taxonomy/TestTaxonomyFacetCounts.java
```diff
@@ -726,6 +743,39 @@ public void testRandom() throws Exception {
     IOUtils.close(tw, searcher.getIndexReader(), tr, indexDir, taxoDir);
   }
 
+  private static List<List<FacetLabel>> sortedFacetLabels(List<List<FacetLabel>> allfacetLabels) {
+    for (List<FacetLabel> facetLabels : allfacetLabels) {
+      Collections.sort(facetLabels);
+    }
+
+    Collections.sort(allfacetLabels, (o1, o2) -> {
+      if (o1 == null) {
```
mikemccand, line 752: Hmm why are these `null` checks necessary?  Are we really seeing `null` in the argument?  Oh, I guess this legitimately happens when the hit had no facets?  Maybe add a comment?  Hmm, actually, looking at how actual and expected are populated, neither of them seems to add `null`?  One of them filters out empty list but the other does not?

lucene/facet/src/test/org/apache/lucene/facet/taxonomy/TestTaxonomyFacetCounts.java
```diff
@@ -726,6 +743,39 @@ public void testRandom() throws Exception {
     IOUtils.close(tw, searcher.getIndexReader(), tr, indexDir, taxoDir);
   }
 
+  private static List<List<FacetLabel>> sortedFacetLabels(List<List<FacetLabel>> allfacetLabels) {
+    for (List<FacetLabel> facetLabels : allfacetLabels) {
+      Collections.sort(facetLabels);
+    }
+
+    Collections.sort(allfacetLabels, (o1, o2) -> {
```
mikemccand, line 751: I'm confused why we are sorting the top list?  Isn't the top list in order of the hits?  And we want to confirm, for a given `docId` hit, that expected and actual labels match?

OK, I think I understand: this test does not index anything allowing you to track which original doc mapped to which `FacetLabel`, so then you cannot know, per segment, which docs ended up where :)

Given that, I think it's OK to do the top-level sort of all `List<FacetLabel>` across all hits.

lucene/facet/src/test/org/apache/lucene/facet/taxonomy/TestTaxonomyFacetCounts.java
```diff
@@ -711,6 +723,11 @@ public void testRandom() throws Exception {
         }
       }
 
+      // Test facet labels for each matching test doc
+      List<List<FacetLabel>> actualLabels = getTaxonomyFacetLabels(tr, config, fc);
+      assertEquals(expectedLabels.size(), actualLabels.size());
```
mikemccand, line 728: Hmm I think `expectedLabels` filters out empty `List<FacetLabel>` but `actualLabels` does not, so this might false trip?

goankur: Nice catch, thanks.  I fixed `actualLabels` generation in `FacetTestCase.getAllTaxonomyFacetLabels()` method to filter out empty `List<facetLabels>`.

lucene/facet/src/test/org/apache/lucene/facet/FacetTestCase.java
goankur: done in the next revision.

lucene/facet/src/test/org/apache/lucene/facet/taxonomy/TestTaxonomyFacetCounts.java
goankur: Thanks for catching this @mikemccand. I fixed the `actualLabels` to exclude empty lists. The null checks were just me being extra cautious. I realized they were unnecessary and removed them :-)

lucene/facet/src/test/org/apache/lucene/facet/taxonomy/TestTaxonomyFacetCounts.java
goankur: Yes, a document with `N`<sup>th</sup> position in the input sequence might end up with `K`<sup>th</sup> docId in a random segment making it harder to compare actual and expected labels.

Thanks for confirming that the approach is acceptable.

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/TaxonomyFacetLabels.java
goankur: The new per-segment class (`FacetLabelReader`) holds an instance of `DocValuesOrdinalsReader` which internally holds a reusable instance of `BinaryDocValues`.

In the `nextFacetLabel()` method `ordinalsSegmentReader.get(docId, decodedOrds)` that does the actual decoding of the ordinals is only called **once** if the input docId is the same as the one supplied in prior invocation of `nextFacetLabel()`.  

Subsequent invocations fetch the ordinals from `FacetLabelReader.decodedOrds` and use the ordinal to lookup FacetLabels using `taxoReader.getPath(ord)`. 

Yes, after LUCENE-9450, we can bulk resolve all the ordinals using a different API, something like -> `List<FacetLabels> taxoReader.getPaths(IntRef decodedOrds)` and add a new API - `FacetLabelReader.allFacetLabels(int docId)`

mikemccand: Yeah, or we could bulk map all ord -> FacetLabel as soon as caller goes to the next `docId`, and then iterate them one by one with this API.

Anyway, we don't need to solve that here.

lucene/facet/src/test/org/apache/lucene/facet/taxonomy/TestTaxonomyFacetCounts.java
mikemccand: Hmm I don't like filtering out "no facet fields" case from `FacetTestCase.getAllTaxonomyFacetLabels()` -- that means callers cannot safely map hits to the `List<FacetLabel>`?

Could we instead keep the empty lists in both expected and actual?

mikemccand (commented):
Thanks @goankur!

lucene/facet/src/test/org/apache/lucene/facet/taxonomy/TestTaxonomyFacetCounts.java
goankur: Done. I made changes to keep empty lists in both `expectedLabels` and `actualLabels` and also updated javadoc for `FacetTestCase.getAllTaxonomyFacetLabels()` to reflect this for `actualLabels`

mikemccand:
Thanks @goankur I think this is ready!  I'll confirm all tests pass and squash/merge soon.

PR #1894, opened 2020-09-20
SOLR-14878: Expose solr.xml's coreRootDirectory property via the System Settings API

arafalov:
# Description

Even if coreRootDirectory was set in solr.xml, external applications could not access it through the API. It was not exposed. This means it is not possible to run a command such as *bin/solr create_core -c test* as the code that pre-populates the core directory is doing it in the wrong place (if at all).

# Solution
Expose the property if it is defined and is different from solr_home

# Checklist

### DIFF
--- a/solr/core/src/java/org/apache/solr/handler/admin/SystemInfoHandler.java
+++ b/solr/core/src/java/org/apache/solr/handler/admin/SystemInfoHandler.java
@@ -23,6 +23,7 @@ import java.lang.management.ManagementFactory;
 import java.lang.management.OperatingSystemMXBean;
 import java.lang.management.RuntimeMXBean;
 import java.net.InetAddress;
+import java.nio.file.Path;
 import java.text.DecimalFormat;
 import java.text.DecimalFormatSymbols;
 import java.util.Date;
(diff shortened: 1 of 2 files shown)


solr/core/src/java/org/apache/solr/handler/admin/SystemInfoHandler.java
```diff
@@ -139,8 +140,19 @@ public void handleRequestBody(SolrQueryRequest req, SolrQueryResponse rsp) throw
...
       rsp.add("zkHost", getCoreContainer(req, core).getZkController().getZkServerAddress());
     }
-    if (cc != null)
-      rsp.add( "solr_home", cc.getSolrHome());
+    if (cc != null) {
+      String solrHome = cc.getSolrHome();
+      rsp.add("solr_home", solrHome);
+
+      Path coreRootDirectory = cc.getCoreRootDirectory();
+      if (coreRootDirectory != null) {
```
dsmiley, line 148: coreRootDirectory is always non-null.  I don't like null checks that are unnecessary; it may feel "safe" but slightly obscures code clarity and suggests possibilities that are impossibilities.

solr/core/src/java/org/apache/solr/handler/admin/SystemInfoHandler.java
```diff
@@ -139,8 +140,19 @@ public void handleRequestBody(SolrQueryRequest req, SolrQueryResponse rsp) throw
...
+        String coreRootDirectoryString = coreRootDirectory.toString();
+        if (!coreRootDirectoryString.equals(solrHome)) {
```
dsmiley, line 150: I disagree with this condition.  I think this should always emitted no matter where it is.  It's more helpful on consumers to depend on its existence rather than having it vary in some cases but not others.

arafalov: Ok, I was on borderline myself. I felt that because this option is barely used, having the same long string sent twice was bit of a waste of bandwidth. But, I guess it does not happen too often.

I can correct both generation and invocation of this option to make it a guaranteed setup. Let's just hope they don't run version 9 tool against version 8 core.

Also, if you don't like the name (the other open point), this is a good time to mention it.

dsmiley: The "solr_" part should be skipped IMO.  Everything here is Solr :-)

> Let's just hope they don't run version 9 tool against version 8 core.

Well it's up to them to have a tolerant client.  For the SolrCLI I think it makes sense to check for null in case of a cross-version situation.

PR #1899, opened 2020-09-21
Adding dev-docs around the use of Git Worktree.

HoustonPutman:
First attempt, very open to suggestions on making things more clear.


madrob (approved):

dsmiley (approved):
Fantastic

dsmiley:
Oh wait... this is confusing -- help/ and dev-docs/.  I went looking for this doc just now and looked in the wrong place (help/).  Any opinions @dweiss

dweiss:
help/ are plain text files that are also converted to gradle tasks. I wouldn't want them to be converted to adoc because then we'd have to parse them somehow. To me help/ is really low-level stuff related to build and tooling/ dev-docs/ seem to be higher-level process related stuff? This is the way I see it.

PR #1900, opened 2020-09-21
SOLR-14036: Remove explicit distrib=false from /terms handler

munendrasn:
* This removes shards, shards.qt, and shard whitelisting checks from TermsComponent. Similar to other components, this check will be done in HttpShardHandler


munendrasn:
@tflobbe I have removed CustomTermsComponentTest as none of the components currently have custom handling of shardWhitelist. Let me know if that should be added back

dsmiley:
I'm glad to see this :-)

I think the change of default behavior for users using /terms should be master-only and have a note in `solr-upgrade-notes.adoc`.  If you like, you could backport the underlying fix but not the change in default.

munendrasn:
@joel-bernstein Please review(not able to tags as a reviewer so the ping)

munendrasn:
>I think the change of default behavior for users using /terms should be master-only and have a note in solr-upgrade-notes.adoc. If you like, you could backport the underlying fix but not the change in default.

At this point, I'm planning to keep this master only but if there is a need will backport the necessary changes to 8x

arafalov:
I can't comment on the code, as I did not understand the shard part. I am happy if the extra parameter (distrib=false) is not needed. I did feel that maybe a test was missing somehow. Is there one that tests that distributed terms work? Especially, since terms handler is explicitly defining the Component Chain to be terms only and nothing else (not even sure if that's relevant, though).

munendrasn:
@arafalov 
There is one already which I have modified(DistributedTermsComponentTest) - https://github.com/apache/lucene-solr/pull/1900/files#diff-9b9ccbcf271c6320902d92479615f4fbR73

arafalov:
> @arafalov
> There is one already which I have modified(DistributedTermsComponentTest) - https://github.com/apache/lucene-solr/pull/1900/files#diff-9b9ccbcf271c6320902d92479615f4fbR73

Thank you for clarification. I saw that code change, but did not realize it answered my question.

dsmiley (changes requested):
Once I see the solr-upgrade-notes.adoc change, I'll approve.

munendrasn:
I have included the changes and upgrade entry. Instead of adding upgrade entry to `solr-upgrade-notes.adoc`, I have added to `major-changes-in-solr-9.adoc` as mentioned the former doc

solr/solr-ref-guide/src/major-changes-in-solr-9.adoc
```diff
@@ -128,6 +128,8 @@ _(raw; not yet edited)_
 * SOLR-14510: The `writeStartDocumentList` in `TextResponseWriter` now receives an extra boolean parameter representing the "exactness" of the numFound value (exact vs approximation).
   Any custom response writer extending `TextResponseWriter` will need to implement this abstract method now (instead previous with the same name but without the new boolean parameter).
 
+* SOLR-14036: Implicit /terms handler now supports distributed search by default, when running in cloud mode.
```
dsmiley, line 131: Reworded to help a user think through upgrading:
```suggestion
* SOLR-14036: Implicit /terms handler now returns terms across all shards in SolrCloud instead of only the local core.  Users/apps may be assuming the old behavior.  A request can be modified via the standard distrib=false param to only use the local core receiving the request.
```

dsmiley (approved):

PR #1901, opened 2020-09-21
SOLR-14883 Add a Muse (Continuous assurance platform) configuration

TomMD:
# Description

The Apache infrastructure team has installed the Muse GitHub application.  In order for Lucene-Solr to benefit the application must understand how to build the project.  While it has build heuristics, it does not guess between JDK8 or JDK11 (two JDKs most supported by several static analysis tools).  Thus, this configuration explicitly instructs use of JDK11.

# Solution

Add a config.toml with `jdk11 = true`.

# Tests

I've run Muse with this configuration before.  If, for some reason, the configuration is incorrect of has a typo then it at least won't negatively impact the rest of the project.

# Checklist


.muse/config.toml
```diff
@@ -0,0 +1 @@
+jdk11 = true
```
tflobbe, line 1: Could you add a comment or two pointing to the documentation how to add/change/skip checks?

TomMD: @tflobbe Sure, I'll put that here and amend the commit.

The full documentation is docs.muse.dev.  Most interesting is the repository configuration reference https://docs.muse.dev/docs/repository-configuration/ which details how to change which tools runs, add custom tools (other than those built into the platform by default), filter for or against certain bug types, etc.

For example, an explicit `.muse/config.toml` file might be:

```
jdk11 = true
build = "./gradlew assemble"
tools = [ "infer", "errorprone", "findsecbugs" ]
customTools = [ ]
```

tflobbe: Ah, I noticed you added the comments on the commit. Thanks for that, but I was thinking we could put some comment in the file itself, something like:
```
# This file can configure Muse PR static analysis. 
# Docs for adding, removing or changing checks can be found here: https://docs.muse.dev/docs/repository-configuration/
```
or something of that sort, mostly to save some googling. WDYT?
@dsmiley, do you have any concerns with this? (since you commented on the Jira issue). I seems like a great thing to try, and if Apache infra installed the app for us, I see no reason not to do it.

TomMD: I've added the comment to the file as well now.

tflobbe (approved):
LGTM. I'll merge tomorrow if no other concerns

madrob (approved):

PR #1903, opened 2020-09-21
Fix bug in sort optimization

mayya-sharipova:
Fix bug how iterator with skipping functionality
advances and produces docs

Relates to #1725

### DIFF
--- a/lucene/core/src/test/org/apache/lucene/search/TestFieldSortOptimizationSkipping.java
+++ b/lucene/core/src/test/org/apache/lucene/search/TestFieldSortOptimizationSkipping.java
@@ -53,6 +53,7 @@ public class TestFieldSortOptimizationSkipping extends LuceneTestCase {
       if (i == 7000) writer.flush(); // two segments
     }
     final IndexReader reader = DirectoryReader.open(writer);
+    writer.close();
     IndexSearcher searcher = new IndexSearcher(reader);
     final SortField sortField = new SortField("my_field", SortField.Type.LONG);
     final Sort sort = new Sort(sortField);
@@ -110,7 +111,6 @@ public class TestFieldSortOptimizationSkipping extends LuceneTestCase {
       assertEquals(topDocs.totalHits.value, numDocs); // assert that all documents were collected => optimization was not run
     }
 
-    writer.close();
     reader.close();
     dir.close();
   }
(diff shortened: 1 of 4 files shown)


mayya-sharipova:
@jimczi  I could not reproduce an issue of duplicate documents in Lucene, and I am still not sure how it actually happens.

 Even as we were [incorrectly advancing](https://github.com/apache/lucene-solr/blob/master/lucene/core/src/java/org/apache/lucene/search/Weight.java#L216) only `scorerIterator`, as `filteredIterator`  is an intersection of `collectorIterator` and `scorerIterator`, a next document from `filteredIterator` must be a document > the current document of `scorerIterator`? No?

There was also a question why we did not detect duplicate documents in the tests. First, they don't happen in the tests, and before I was not using `AssertingIndexSearcher`.  I've changed the tests on _doc sort to use `AssertingIndexSearcher`

mayya-sharipova:
@jimczi I was able to reproduce a bug, and in aaab92a added a test that reproducibly fails for me without this patch.  It is a very particular combination of docs and queries to have a `BitSetIterator` as a `scorerIterator` and  ensure that we go through `DefaultBulkSorter::scoreRange` section for scoring.

lucene/core/src/test/org/apache/lucene/search/TestFieldSortOptimizationSkipping.java
```diff
@@ -432,7 +439,48 @@ public void testDocSortOptimization() throws IOException {
...
+    reader.close();
+    dir.close();
+  }
+
+  /**
+   * Test that sorting on _doc works correctly.
+   * This test goes through DefaultBulkSorter::scoreRange, where scorerIterator is BitSetIterator.
+   * As a conjunction of this BitSetIterator with DocComparator's iterator, we get BitSetConjunctionDISI.
+   * BitSetConjuctionDISI advances based on the DocComparator's iterator, and doesn't consider
+   * that its BitSetIterator may have advanced passed a certain doc.
```
jimczi, line 451: Should we consider this a bug in `BitSetIterator` ?

jimczi (approved):
The fix looks good to me. Thanks for adding the test.

lucene/core/src/test/org/apache/lucene/search/TestFieldSortOptimizationSkipping.java
mayya-sharipova: I will create an issue for this.

mayya-sharipova: Issue created: https://issues.apache.org/jira/browse/LUCENE-9541

PR #1905, opened 2020-09-21
LUCENE-9488 Release with Gradle Part 2

madrob:
Deleted lucene/version.properties again


lucene/build.gradle
```diff
@@ -15,8 +15,56 @@
  * limitations under the License.
  */
 
+// Should we do this as :lucene:packaging similar to how Solr does it?
```
dweiss, line 18: Yes, please move it to :lucene:packaging. It makes things much cleaner in the long run.

lucene/build.gradle
```diff
@@ -15,8 +15,56 @@
...
 
 subprojects {
   group "org.apache.lucene"
-}
\ No newline at end of file
+}
+
+distributions {
+  main {
+      // This is empirically wrong, but it is mostly a copy from `ant package-zip`
```
dweiss, line 33: This is wrong. It should rely on published artifacts from other subprojects (set up proper dependencies) and assemble them in the right way into the distribution layout.

madrob: I tried looking at the solr packaging for figuring this out, but kept getting duplicate entries in the archive. I left the TODOs for what still remains to be added to the zip, can you give me a little more direction on what the implementation would look like?

dweiss: Sure but you'll have to be patient - can't do it right away.

dweiss: Haven't forgotten about it, just busy with work. Those release scripts will have to be adjusted to Solr and Lucene released independently in the future. Which requires independent builds, which requires repo split. Will have to get to it, eventually. Sigh.

madrob: My goal here with getting things releasable is to also turn the smoke tester back on so that we can hopefully catch issues before we actually go to do the release. I understand there’s going to be more split related work, but that shouldn’t stop us from working on the pieces that we can work on before that.

janhoy (approved):
LGTM. Mostly skimmed wizard parts

lucene/build.gradle
dweiss: Ok, fair enough. But I wouldn't want to add things to gradle that are not right - these are hard to get rid of later on. I'm really busy this week but I will correct those assembly bits and commit them to this PR. Please give me some time to work on this, thank you.

dweiss:
I put together binary distribution assembly the way I see this should eventually work, Mike. @uschindler may also take a look. It may look complicated on the surface - with the individual configurations for subprojects and dependency handlers declared programmatically - but I think this is the way to go as to get *consistency* in the output.

When you compare the assembled distribution against 8.x ZIP you'll see what I mean - some of the JARs in 8.x are from test scopes, we don't include some of the required JARs under lib/ in many artifacts... It's a mess. 

When you think of it I don't know if it makes sense to add copies of third-party JARs under lib/ at all... I think we could just generate a plain text file with all the dependencies listed. I don't think many people assemble projects manually these days anyway, so binary distributions just ship the "official" artifacts of a project.

dweiss:
Just to be clear - I don't know if it's everything concerning binary distribution (please eyeball diffing against 8.x), just a step forward. Have to dash to work now.

madrob:
Dawid, it took me a few tries of going through your comments and other examples in the code but I think it finally dawned on me what you meant as the proper way to do it. Each task can export its own outputs as an artifact, that's the piece that was missing in my mental model. Please take another look?

There's another piece where the Luke log4j dependency was causing a duplicate entry error in the archives... not sure if you were getting the same error or not. Otherwise the output zip looks pretty close to the real thing from 8x

dweiss:
Hi Mike. Well, "proper" in my opinion - there's certainly a million ways to do it. Thank you for being patient with me; I'll review as soon as I can. A separate thing is repository (and build) separation which this is closely related to since source releases should be buildable. I've extracted Lucene-only part of the build already and will proceed to Solr when we have Lucene snapshot builds up on Apache repositories.

dweiss:
I think it looks good now! If there's anything else we can follow-up.

lucene/packaging/build.gradle
```diff
@@ -0,0 +1,160 @@
...
+
+description = 'Lucene distribution packaging'
+
+// Declare all subprojects that should be included in binary distribution.
+// By default everything is included, unless explicitly excluded.
+def includeInBinaries = project(":lucene").subprojects.findAll {subproject ->
+    return !(subproject.path in [
+        ":lucene:packaging",
+        ":lucene:analysis",
+        ":lucene:luke", // nocommit - Encountered duplicate path "luke/lib/log4j-core-2.13.2.jar"
```
madrob, line 33: @dweiss this is still a problem, not sure why it comes up or how to resolve it.

dweiss:
I'll take a look later today.

lucene/packaging/build.gradle
dweiss: This wasn't obvious. Here's what happened.

We create a configuration with dependency on "full project", it looked like this:
```
handler.add(confFull, project(path: includedProject.path), {...
```

This in fact is a dependency on the configuration aptly named 'default', which includes all of project binaries and transitive dependencies (archives in general).

Luke declared a separate configuration for dependencies called 'standalone' which added a few non-project dependencies (the 'implementation' extended from it so they were visible on runtime classpath). The core of the problem was in that luke also exported an extra artifact that belonged to this 'standalone' configuration  - this was a set of JARs assembled under a folder. So when the packaging process was collecting files for inclusion, it encountered log4j twice: a copy in that 'standalone' folder and a copy from transitive project dependencies.

I recall you asked why this 'fail' on double archive entries is needed. Well, it's useful to catch pearls like this one above...

dweiss: I've verified that it collects Luke properly and the launch script starts it fine (on Windows).

madrob:
> Are we sure we need dependency at all. I would have assumed that Gradle has it already.

Gradle has a shaded commons codec that we can use - `org.gradle.internal.impldep.org.apache.commons.codec.digest.DigestUtils` - but I assumed that would be frowned upon.

> This is also a forbiddenapis violation: the charset is missing. Please use: setText(..., "UTF-8")

dweiss:
Yup, don't use internal* - not a good idea. As for gradle API - I don't think it'll be easy to apply forbidden APIs here?

uschindler:
You can apply forbiddenapis to groovy or gradle code. But for that you need to have everything compiled to build-src./build/classes With adhoc scripts it wont work.

dweiss:
That's what I though.

uschindler:
Nevertheless, there's still a forbiddenapis issue: When writing the SHA512 file, we don't apply charset, it's platform dependent.

gradle/releasing.gradle
```diff
@@ -0,0 +1,58 @@
...
+    dependencies {
+        classpath 'commons-codec:commons-codec:1.13'
+    }
+}
+
+allprojects {
+    plugins.withType(DistributionPlugin) {
+        def checksum = {
+            outputs.files.each { File file ->
+                new File(file.parent, file.name + ".sha512").text = new DigestUtils(MessageDigestAlgorithms.SHA_512).digestAsHex(file).trim() + "  " + file.name
```
uschindler, line 37: Use charset here. `File.text = ...` uses default charset. Use `File.setText(..., 'UTF-8')`

uschindler (changes requested):

gradle/releasing.gradle
```diff
@@ -0,0 +1,58 @@
+
```
danmuzi, line 1: minor comment but unnecessary blank line :)

solr/packaging/build.gradle
```diff
@@ -115,15 +121,15 @@ distributions {
         into "server"
       })
 
+      from(configurations.docs, {
+        into "docs"
+      })
       // docs/   - TODO: this is assembled via XSLT... leaving out for now.
+      // -- Is there more to do here?
```
uschindler, line 128: I don't think so. Documentation in Solr's Artifacts on Jenkins already contain everything perfectly. If its the same for Lucene, we're fine. Should land in the ZIP's `docs` folder.

uschindler (approved):
I can only check the Gradle stuff, not the Python scripts. From Gradle side, looks fine. So please merge this, as I'd like to enable the Lucene Artifact builds on Jenkins, so everybody can review and compare to 8.x.

solr/packaging/build.gradle
uschindler: Oh, I am not sure if this works. At least in current master, I see NO 'docs/' foder in the binary releaese!

lucene/packaging/build.gradle
```diff
@@ -0,0 +1,171 @@
...
+            })
+
+            // A couple more missing README files
+            from(project(':lucene:analysis').projectDir) {
+                include "README.txt"
+                into 'analysis'
+            }
+
+            // Copy files from documentation output to 'docs'
+            from(configurations.docs, {
```
uschindler, line 106: where is this coming from?

solr/packaging/build.gradle
```diff
@@ -115,15 +121,15 @@ distributions {
         into "server"
       })
 
+      from(configurations.docs, {
```
uschindler, line 124: where is this coming from?

solr/packaging/build.gradle
```diff
@@ -62,12 +63,17 @@ dependencies {
 
   example project(path: ":solr:example", configuration: "packaging")
   server project(path: ":solr:server", configuration: "packaging")
+
+  // Copy files from documentation output
+  docs project(path: ':solr', configuration: 'docs')
```
uschindler, line 68: I have no idea where this configuration is defined. The `:solr` project only has a task called `documentation` that needs to be executed before, but the outputs are not defined as config. So how should this work (same for Lucene)?

uschindler (changes requested):
I have some qeustions and don't understand how the 'docs' config should work. @dweiss can you help?

solr/packaging/build.gradle
uschindler: There are some `ext` properties prointing to the documentation after it's generated. For the Solr ZIP/TGZ its just some minimal folder: `project(':solr').docrootOnline` (packaged in ZIP)

For full lucene/solr docs, refer to `project(':lucene').docroot` (packaged in TGZ) and `project(':solr').docroot` (not packaged, only published on website).

solr/packaging/build.gradle
uschindler: See comments below.

lucene/packaging/build.gradle
dweiss: This is part of the PR, Uwe - here.
https://github.com/apache/lucene-solr/pull/1905/files#diff-d55d7b9fe9d94d4e2e5554e3c1108943R60-R63

I suggested using configuration-based copying instead of pointing at an explicit folder because it has several advantages. First, an artifact exported via configuration is configurable in the place where it logically belongs to. So you can put together a set of documentation artifacts in documentation configuration snippet and just depend on it when you're assembling something that uses those artifacts.

The second benefit is that dependencies on tasks assembling those artifacts are automatically resolved - you don't need to know anything beyond the name of the project and configuration from which you're coping. The rest is done by gradle. 

I think Mike's patch is fine here.

solr/packaging/build.gradle
dweiss: I explained this above, I hope. I think artifact/configuration based approach to collecting packaging dependencies is superior to explicit properties and task-dependencies.

solr/packaging/build.gradle
dweiss: I think we can create a separate issue for collecting Solr docs? This patch is mostly about Lucene and it's pretty big already.

solr/packaging/build.gradle
uschindler: But where is the documentation config defined. I did not find it anywhere else. I understand and I like it, but this is incomplete. And currently it does not work (I see no docs in ZIP files).

uschindler: So as far as I understand: In the documentation.gradle file I would add a "configuration docs" for `:solr` and `:lucene` and then use `builtBy 'documentation'` there. That's what I am missing here. I was about to set this up yesterday, but was not sure if I miss something.

dweiss: Ah, sorry - yes, you are correct. It should declare artifacts, the configuration to attach them to and the outputs. An example is in solr/example/build.gradle:

artifacts {
  packaging packagingDir, {
    builtBy assemblePackaging
  }
}

uschindler: Ah cool. That is what I was missing. I will add this after lunch. I was not sure how to define the artifacts. The Gradle documentation was not very helpful to me.
Thanks and sorry for stupid questions. 🤣

dweiss: There are no stupid questions. This is the relevant documentation bit here, I believe.
https://docs.gradle.org/current/userguide/cross_project_publications.html

uschindler:
Hi @madrob: I will now commit in the changes to make the documentation of Lucene and also Solr (still missing) in the TGZ working. As dawid said, theres the configuration missing.

lucene/packaging/build.gradle
uschindler: I fixed it to shipthe correct documentation. There were also some usless dependencies that I removed.

uschindler:
I fixed the documentation:
- Lucene shipos with full docs including javadocs
- Solr ships with a stub pointing to web page (we don't ship javadocs in TGZ). I will only fix the 'documentationOnline' task in a lter commit, to generate the HTML changed. In 8.x those are missing, but that's a pity and now easy to do.

uschindler (approved):

dweiss:
LGTM, thanks Uwe.

uschindler:
I did some cleanup of the documentation task (there were some XSL files as relics from Ant build before everything was changed to markdown). I also added the HTML formatted changes.txt back.

This is now ready, IMHO. Thanks @madrob for taking the action and making this avialble. Once this is merged, I will enable the Artifact build on Jenkins. Just waiting for the merge :-)

uschindler:
Hi @madrob are you planning to further work on this. I have seen one TODO regarding a README. If you don't have time, I can merge this and enable Jenkins Artifact builds.

madrob:
Rebasing, squashing, and pushing later today

uschindler:
You can just click the button...

uschindler:
Thanks. I will soon activate jenkins jobs. 👍🏻😉

uschindler:
Lucene artifact builds work: https://ci-builds.apache.org/job/Lucene/job/Lucene-Artifacts-master/

What's missing ist src.tgz files,are they not yet handled or do I miss a task?

madrob:
They're not handled yet, I suspect we should just be doing `git archive` rather than defining a gradle task, but I'll have more time to follow up next week

dweiss:
This isn't so easy as the current gradle build can't be separated into Lucene and Solr-only parts (as it was the case with ant). We could distribute everything (Solr and Lucene sources) or wait until builds are separated (which I showed is not that much of a big deal).

uschindler:
Yeah. At the moment both artifacts in source form would be the same. So I agree, just zip too level.

At Ant times Solr always was also a combined zip. Lucene was just Lucene subfolder, but it was somehow incomplete. It was able to build, but no real releases.

IMHO we should have always released a combined src.tgz and separate binaries

PR #1907, opened 2020-09-22
LUCENE-9538: Detect polygon self-intersections in the Tessellator

iverase:
This PR adds a check in the tessellator to detect self-intersections so it can provide a more meaningful error to the user.

cc: @nknize


iverase:
Hi @nknize, I would like to add this for the upcoming release, do you think you will have time to review it?

nknize (commented):
The logic LGTM. Do we have an idea what kind of performance overhead this adds? We may want to start considering teasing some of the validation checks out into a separate validator class and making the validation step optional. Some users have polygons that already went through prior validation / cleaning and this step would just be adding superfluous processing. Orient and determinant calculations aren't cheap in the aggregate.

iverase:
Performance test shows a small impact in indexing performance:

```
Index time (sec)||Force merge time (sec)||Index size (GB)||Reader heap (MB)||
||Dev||Base||Diff ||Dev  ||Base  ||diff   ||Dev||Base||Diff||Dev||Base||Diff ||
|480.5s|467.0s| 3%|0.0s|0.0s| 0%|2.24|2.24| 0%|0.01|0.01| 0%|
```

iverase:
Thinking more about this, I am not sure if we can make it optional. There can be situations when the polygon is invalid but it actually does not fail when it is tessellated. We have an example in the tests, where the polygon looks like:

image

This polygon is invalid and the new check will throw an error but it does not fail if you try to tessellate it. The documentation of the 'Tessellator` explicitly says that:

```
Holes may only touch at one vertex
```
IMHO I think we should fail in this case.

iverase:
superseded by https://github.com/apache/lucene/pull/428

PR #1908, opened 2020-09-22
LUCENE-9539: Use more compact datastructures for sorting doc-values

s1monw:
This change cuts over from object based datastructures to primitive / compressed datastructures.


jimczi (approved):
Sparse sorting, nice! The change looks great to me.
I was wondering if we could also compact the offsets array but that would require an extra sort per field. Although, that's probably not needed since we can also remove the caching of the doc values in order to ensure that we load only one field at a time during a merge.

lucene/core/src/java/org/apache/lucene/index/SortedSetDocValuesWriter.java
```diff
@@ -379,4 +352,31 @@ public long getValueCount() {
...
+        int newDocID = sortMap.oldToNew(docID);
+        long startOffset = ordOffset;
+        long ord;
+        while ((ord = oldValues.nextOrd()) != NO_MORE_ORDS) {
+          builder.add(ord + 1);
+          ordOffset++;
+        }
+        if (startOffset != ordOffset) { // do we have any values?
+          offsets[newDocID] = startOffset;
+          builder.add(0); // 0 ord marks next value
```
dweiss, line 375: Technically you could also have a sign switch as boundary (~value marks first value) but I don't think it'd result in any savings and would only complicate the value stream.

dweiss:
Nice. +1.

lucene/core/src/java/org/apache/lucene/index/SortedSetDocValuesWriter.java
s1monw: agreed, I didn't do it since it would have impacted readability too much IMO

PR #1910, opened 2020-09-22
Use github actions cache

madrob:
Cache our gradle downloads when dependancies haven't changed


HoustonPutman (approved):
Nice! Just re-ran both tests, and the caching took ~5 minutes off each.

+1

PR #1912, opened 2020-09-23
LUCENE-9535: Try to do larger flushes.

jpountz:
DWPTPool currently always returns the last DWPT that was added to the
pool. By returning the largest DWPT instead, we could try to do larger
flushes by finishing DWPTs that are close to being full instead of the
last one that was added to the pool, which might be close to being
empty.

When indexing wikimediumall, this change did not seem to improve the
indexing rate significantly, but it didn't slow things down either and
the number of flushes went from 224-226 to 216, about 4% less.

My expectation is that our nightly benchmarks are a best-case scenario
for DWPTPool as the same number of threads is dedicated to indexing over
time, but in the case when you have e.g. a single fixed threadpool that
is responsible for indexing into several indices, the number of indexing
threads that contribute to a given index might greatly vary over time.


s1monw:
I think we came from this years ago and them moved to the last recently used. I am trying to remember if it was easier to implement since we also fixed that we always hit the same DWPT if we are single threaded. But this was before I removed the PerThread indirection. Now with DWPT directly pooled it seems like a reasonable thing to do. @mikemccand WDYT

s1monw (changes requested):
any chance we can get a simple test case for this?

mikemccand:
Oh my!

We are talking about assigning DWPT to incoming indexing thread, right?  (And not which DWPT to pick for flushing because RAM buffer is full, which I think is currently "the biggest one/s"?).

I think this is a good idea.  It will tend to make the biggest DWPTs even bigger, especially when there is high variance of how many threads are indexing at once over time, until a flush is triggered.  I do not remember why we switched to "last DWPT".  Long ago we did have thread affinity, so the same indexing thread would try to get the same DWPT, but we moved away from that quite a while ago.

jpountz:
> We are talking about assigning DWPT to incoming indexing thread, right?

Right.

The thing that makes me hesitate is that it would probably make fetching a DWPT slightly more costly, which could slow indexing down with very fast indexing chains like the 1kb wikipedia documents dataset we use for nightly benchmarks since this method is called under a lock.

mikemccand:
> > We are talking about assigning DWPT to incoming indexing thread, right?
> 
> Right.
> 
> The thing that makes me hesitate is that it would probably make fetching a DWPT slightly more costly, which could slow indexing down with very fast indexing chains like the 1kb wikipedia documents dataset we use for nightly benchmarks since this method is called under a lock.

Let's see if benchmarks really notice that?  Net/net I think the change is a good idea.

jpountz:
I shot in the middle by still using a list rather than a true PQ but trying to keep it mostly ordered in order to make it more likely to poll one of the larger DWPTs. The number of flushes is a bit higher than the previous version of the patch (219 vs. 216) but still lower than current master (224-226).

lucene/core/src/java/org/apache/lucene/index/DocumentsWriterPerThreadPool.java
```diff
@@ -112,19 +110,12 @@ private synchronized DocumentsWriterPerThread newWriter() {
...
-      final Iterator<DocumentsWriterPerThread> descendingIterator = freeList.descendingIterator();
-      while (descendingIterator.hasNext()) {
-        DocumentsWriterPerThread perThread = descendingIterator.next();
-        if (perThread.tryLock()) {
-          descendingIterator.remove();
-          return perThread;
-        }
+      DocumentsWriterPerThread dwpt = freeList.poll(DocumentsWriterPerThread::tryLock);
+      if (dwpt == null) {
+        // DWPT is already locked before return by this method:
```
mikemccand, line 115: Hmm, can you put this comment back in its original place (just before the `return dwpt`?  It is confusing here, making me think the "allocate a new DWPT" case has something to do with the locking semantics.

lucene/core/src/java/org/apache/lucene/index/ApproximatePriorityQueue.java
```diff
@@ -0,0 +1,136 @@
...
+    for (int i = 0; i < Long.SIZE; ++i) {
+      slots.add(null);
+    }
+  }
+
+  /**
+   * Add an entry to this queue that has the provided weight.
+   */
+  void add(T entry, long weight) {
+    if (entry == null) {
```
mikemccand, line 50: Maybe switch to `assert`?

lucene/core/src/java/org/apache/lucene/index/ApproximatePriorityQueue.java
```diff
@@ -0,0 +1,136 @@
...
+      throw new NullPointerException();
+    }
+
+    final int expectedSlot = Long.numberOfLeadingZeros(weight);
+
+    // If the slot is already taken, we take the next one that is free.
+    final long freeSlots = ~usedSlots;
+    int destinationSlot = expectedSlot + Long.numberOfTrailingZeros(freeSlots >>> expectedSlot);
```
mikemccand, line 58: This magical math is hard to understand :)  Can you add some comments, e.g. "this is a faster way to do this more obvious thing: ..."?  I think the gist is that you could have used a `for` loop to (linearly) walk, looking for the next free slot, but instead we are doing these bitwise operations to make it O(1) cost?

Maybe `assert destinationSlot >= expectedSlot`?

lucene/core/src/java/org/apache/lucene/index/ApproximatePriorityQueue.java
```diff
@@ -0,0 +1,136 @@
...
+    if (destinationSlot < Long.SIZE) {
+      usedSlots |= 1L << destinationSlot;
+      slots.set(destinationSlot, entry);
```
mikemccand, line 61: Maybe `assert slots.get(destinationSlot) == null`?

lucene/core/src/java/org/apache/lucene/index/ApproximatePriorityQueue.java
```diff
@@ -0,0 +1,136 @@
...
+    } else {
+      slots.add(entry);
```
mikemccand, line 63: My box might be the only thing that exercises this line!!

Oh hmm, actually, `slots` can be used even when there are fewer than 64 concurrent indexing threads, since the dense 0 .. 63 case is based on `log(ramBytesUsed)`.

mikemccand (commented):
I think this solution looks really cool!

I am a bit worried about perf impact of this new "best effort" priority queue.  But if benchmarks are happy with it, I am too!  I'm curious to see impact on nightly indexing.

lucene/core/src/java/org/apache/lucene/index/ApproximatePriorityQueue.java
jpountz: you understood it correctly, I'll add the suggested comment and assert

lucene/core/src/java/org/apache/lucene/index/ApproximatePriorityQueue.java
jpountz: You're right that this code is unlikely to be exercised unless there are many indexing threads.

jpountz:
> I am a bit worried about perf impact of this new "best effort" priority queue.

Yeah I looked into it because I was worried of the performance impact of a proper priority queue which needs to do non-negligible work on every push/poll to keep the queue ordered. But it's not clear to me how much more costly it is compared to the current approach.

I was planning on running another benchmark with a lighter indexing chain to better see the impact of this change, e.g. with doc-value fields only, so that the amount of work that is performed between the time when the DWPT pool is polled and the time when the DWPT is added back to the queue is not much heavier than the action of polling from / pushing to the queue.

> I'm curious to see impact on nightly indexing.

So am I. :)

jpountz:
I ran the benchmark I mentioned above:
 - wikimediumall
 - no body field, only `title` indexed as a keyword and doc-value fields to make the indexing chain cheaper
 - 18 indexing threads
 - 3 merge threads

| Branch | Indexing time (s) | Number of flushes |
| ------------- | ------------- | -- |
| master | 134 | 48 |
| master | 125 | 48 |
| master | 154 | 47 |
| master | 139 | 47 |
| patch | 139 | 36 |
| patch | 123 | 39 |
| patch | 141 | 38 |
| patch | 142 | 36 |

Indexing times are a bit noisy so it's tricky to say whether it makes things faster or slower, but the total number of flushes is consistently smaller.

lucene/core/src/java/org/apache/lucene/index/DocumentsWriterPerThreadPool.java
jpountz: > making me think the "allocate a new DWPT" case has something to do with the locking semantics.

Hmm, this is exactly what my understanding is. :) To me the comment wat about highlighting that `newWriter()` implicitly takes the lock on the DWPT it creates?

jpountz:
I'm planning to merge this change to see how it plays with nightly benchmarks, especially now that it moved to a ThreadRipper 3990X. I'll revert if it makes things slower.

dweiss:
bq. especially now that it moved to a ThreadRipper 3990X. I'll revert if it makes things slower.

Who's 'it'? :)

I've been playing with TR 3970X and I can cause internal JVM warnings on GC not being able to catch up while all the threads are busy... it's fun to watch.

jpountz:
Woops I meant the nightly bencmarks, I edited my above message.

jpountz:
Merged manually.

PR #1916, opened 2020-09-23
Fix minor typo

dalbani:
Ignoring the default issue template given that this PR is about a tiny fix for a typo. Right?


madrob:
Thank you for finding and correcting this!

PR #1917, opened 2020-09-23
LUCENE-9535: Make ByteBuffersDataOutput#ramBytesUsed run in constant-time.

jpountz:
This is called transitively from `DocumentsWriterFlushControl#doAfterDocument` which is synchronized and appears to be a point of contention.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/store/ByteBuffersDataOutput.java
+++ b/lucene/core/src/java/org/apache/lucene/store/ByteBuffersDataOutput.java
@@ -33,6 +33,7 @@ import java.util.function.IntFunction;
 import org.apache.lucene.util.Accountable;
 import org.apache.lucene.util.BitUtil;
 import org.apache.lucene.util.BytesRef;
+import org.apache.lucene.util.RamUsageEstimator;
 import org.apache.lucene.util.UnicodeUtil;
 
 /**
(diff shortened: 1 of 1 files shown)


s1monw (approved):
LGTM

lucene/core/src/java/org/apache/lucene/store/ByteBuffersDataOutput.java
```diff
@@ -400,8 +400,13 @@ public void writeSetOfStrings(Set<String> set) {
   public long ramBytesUsed() {
     // Return a rough estimation for allocated blocks. Note that we do not make
     // any special distinction for direct memory buffers.
-    return RamUsageEstimator.NUM_BYTES_OBJECT_REF * blocks.size() + 
-           blocks.stream().mapToLong(buf -> buf.capacity()).sum();
+    ByteBuffer first = blocks.peek();
+    if (first == null) {
+      return 0L;
+    } else {
+      // All blocks have the same capacity.
```
dweiss, line 407: Hmmm... Do they? I don't think this is the case, in general, since you can plug in an arbitrary block provider/ recycler?

dweiss: I'm not sure I like this assumption (that all blocks are equal). Maybe we could maintain a separate ram usage on block addition/ removal and thus make the sum-time constant but also accurate?

jpountz: Oops thanks for catching. I thought they did because we make the assumption that all buffers have the same length in toDataInput, which is different from capacity.

jpountz: Yep, I'll do that.

dweiss: Thanks Adrien!

PR #1918, opened 2020-09-23
LUCENE-9535: Commit DWPT bytes used before locking indexing

s1monw:
Currently we calcualte the ramBytesUsed by the DWPT under the flushControl
lock. We can do this caculation safely outside of the lock without any downside.
The FlushControl lock should be used with care since it's a central part of indexing
and might block all indexing.


jpountz (approved):

s1monw:
@jpountz I had to change some stuff to make it work. Down the road I want to clean this up more so we don't need the extra step but I want to do this after we cut 8.7

jpountz (approved):
This makes sense to me and the change still LGTM.

PR #1919, opened 2020-09-24
Compute RAM usage ByteBuffersDataOutput on the fly.

jpountz:
This helps remove the assumption that all blocks have the same size.


jpountz:
@dweiss FYI I could not find a way to have blocks of different capacities as we have an assertion that the allocator creates blocks of the expected capacity, not larger.

dweiss:
I will take another look. I can't remember forcing block capacity but it's been a while!

uschindler:
I like this approach more to just sum up the size when new blocks are allocated and added to Deque.

+1

dweiss:
I like the explicit field too, actually - even if you're right about capacity of internal blocks, Adrien. - this assertion there may actually be my mistake and the capacity of a new block should actually be its limit (remaining free space)... Or maybe I did have capacity in mind (can't remember, to be honest).
```
currentBlock = blockAllocate.apply(requiredBlockSize);
    assert currentBlock.capacity() == requiredBlockSize;
```

PR #1923, opened 2020-09-27
SOLR-14900: Reference Guide build cleanup/module upgrade

arafalov:
# Description
buildSite task was generating some scary warnings on first build as it was getting various gems as well as a warning on each build. It was version locked before due to ant and PDF, but that's no longer part of Solr 9.

# Solution
Upgraded all explicit versions to the latest available.
Everything seems to build the same, the only changes were for the source block styles. We were supposed to be using thankful-eyes theme but the correct name of that theme is actually thankful_eyes and it is a dark theme, so we were falling back to 'github' theme. Made that more clear and explicit.
Switched to using stylesheets as well instead of embedded styles, as upgrade module was generating more markup for better dark-themes support (I guess).

There is still an old warning during build about _Illegal reflective access_ and a new one about _Improper use of Lexer#lex_ . It may be worth repeating the upgrade process in another couple of months to see if that had been fixed in relevant libraries.

# Tests
There were no code changes, but I visually and 'diff' checked the old and new html output.

# Checklist

- [X I have reviewed the guidelines for [How to Contribute](https://wiki.apache.org/solr/HowToContribute) and my code conforms to the standards described there to the best of my ability.


solr/solr-ref-guide/build.gradle
```diff
@@ -44,13 +44,13 @@ buildscript {
...
 
     dependencies {
-        classpath "org.asciidoctor:asciidoctorj:1.6.2"
+        classpath "org.asciidoctor:asciidoctorj:2.4.1"
     }
 }
 
 plugins {
     id 'java'
-    id 'com.github.jruby-gradle.base' version '2.0.0-alpha.7'
```
dweiss, line 53: This will conflict with the cleanup Chris is doing - I modified the same thing there.

solr/solr-ref-guide/src/_config.yml.template
```diff
@@ -7,7 +7,7 @@
 #
 
 # Gems that are included for building the site. jekyll-asciidoc allows Jekyll to use Asciidoctor for variables and settings
-gems: [jekyll-asciidoc]
+plugins: [jekyll-asciidoc]
```
dweiss, line 11: Likewise.

arafalov:
@dweiss Can we add all my changes to that PR/JIRA (which one?) and then we can obsolete this one. Or I can wait for it to be done and do the difference. Whatever works.

dweiss:
It's SOLR-14889. I think it may automatically merge (because those lines are identical) but it's hard to tell in advance. Just wanted you to know Chris is working on some cleanups independently.

arafalov:
Ok, I see it. Looks like Chris's work should go in first. I will sit on mine for a week and then adjust to whatever still needs to be done. There is no urgency, it was just a side-effect of another research.

Thank you for the heads up.

solr/solr-ref-guide/build.gradle
```diff
@@ -90,14 +90,14 @@ dependencies {
     depVer('org.apache.zookeeper:zookeeper')
 
     // jekyll dependencies
-    gems 'rubygems:jekyll:3.5.2'
+    gems 'rubygems:jekyll:4.1.1'
```
ctargett, line 93: There are significant changes between Jekyll 3 and 4 and we haven't upgraded to it yet because it needs some careful analysis and fixing. In the past attempts to build with Jekyll 4 have failed. I don't see evidence in this PR that you have done sufficient testing for me to feel confident that this is OK.

solr/solr-ref-guide/src/_config.yml.template
```diff
@@ -91,6 +91,11 @@ asciidoctor:
...
-    source-highlighter: "rouge"
-    rouge-theme: "thankful-eyes"
     stem:
+    source-highlighter: "rouge"
+    # Options from: https://github.com/jirutka/asciidoctor-rouge
+    rouge-css: "class"
+    # class option requires css file generated by build/.gems/gems/rouge-vx.y.z/bin/rougify style style_name
+    # other option: style for inline styles, which is larger in size
+    # rouge-css: "style"
+    # rouge-style: "github"
```
ctargett, line 101: I'm not sure I understand the point of the commented sections here?

solr/solr-ref-guide/src/_includes/head.html
```diff
@@ -11,6 +11,7 @@
 <link rel="stylesheet" href="{{ "css/ref-guide.css" }}">
 <link rel="stylesheet" href="{{ "css/decoration.css" }}">
 <link rel="stylesheet" href="{{ "css/search.css" }}">
+<link rel="stylesheet" href="{{"css/rouge-style-github.css" }}"
```
ctargett, line 14: Is there no way to do this without explicitly adding the CSS? It adds to the items that we must manage and will require us to maintain this CSS over time.

ctargett:
The Jekyll 3 to 4 upgrade is a major one and has not been done to date because it has caused failures in the past when attempted. I don't understand from this so far what testing has been done to ensure it is 100% working properly - every single page was compared visually and with 'diff'? And they were **exactly** the same?

solr/solr-ref-guide/build.gradle
arafalov: Do you remember any specific examples I could double-check on? I saw some past discussions, but it was about ant and PDF issues. I can double check. However, from the analysis I did do, only source blocks were actually affected.

solr/solr-ref-guide/src/_config.yml.template
arafalov: The styles generated for the source blocks (highlights) can be inlined or in a separate css, which needs to be generated. Due to the support for - I guess - dark mode - the inline styles make the documents somewhat larger. So, CSS looked like a better option.

Also, the previous style was spelled incorrectly and therefore falling-back to default style. I've just made of this explicit after spending time trying to figure out the actual situation.

solr/solr-ref-guide/src/_includes/head.html
arafalov: As above, you can make inline styles, but the file sizes grow. But I can switch to inline if that's such a significant concern.

arafalov:
Thank you for the feedback. Was there a particular methodology you were going to utilize for the upgrade yourself? I did visual and diff spot checks on pages that seemed to have significant changes and only found the highlighter differences (styles and additional markup around white-spaces). But, if there are expectations of significant discrepancies, I am happy to follow a different suggested approach.

ctargett:
Let me build with the patch and see if I want to revise my thoughts on it. It will take me until sometime tomorrow or Thursday to have time to look at it, just FYI.

ctargett:
> Not sure what to do about the license though, it is generated by the tool and does not seem to be licensed in any way itself.

The license and copyright would belong to the Rouge project, so you could add the copyright notice as in the license at https://github.com/rouge-ruby/rouge/blob/master/LICENSE and a link to the project. Then a note about how the CSS was generated (or link to docs on it) as helper tips for anyone who comes after you who wants to re-generate the CSS.

ctargett:
There are two new warnings I noticed being output with these changes: `Improper use of Lexer#lex - this method does not receive options. This will become an error in a future version.` This is a Rouge error.

We don't get verbose output with the Gradle build, so I ran Jekyll locally without Gradle to see if I could catch where these occur. They both happen when processing `response-writers.adoc`. If I had to guess, my first one would be that Rouge is balking at the Ruby code and misinterpreting it as inputs for itself. I have no idea if there are syntax shortcuts in that example or if it's really real Ruby...it's not making anything look any different than what it always has, though.

arafalov:
I thought this was kind of general error. Now that I know it is a specific file, I will dig in deeper and see if it can be eliminated.

The warning AFAIK is about the method that used to silently accept extra parameters and now is doing it loudly and threatens to quit in the future. So, the output may not change. But it is annoying to replace two old warnings with one new. So, if we can figure it out and remove it, that would be nice.

arafalov:
Ok, I reverted back to using direct styles. It only affects code-blocks, so the size changes are minimal and not needing to maintain the CSS file is a bonus. I agree.

Looking at generated code, the only differences I see is some additional highlighting of options and changing in order of code/pre html elements which seems to sometimes lead to removal of one level of pre/code (I suspect it was a bug before).

The warning about Lexer#lex is with source,php blocks in the file, due to asciidoctor/asciidoctor#3336, fixed in the master (not yet released). We can either ignore it or switch formatting for those two blocks to text. I think ignoring it is fine.

ctargett:
+1 this looks good. I think the `solr/solr-ref-guide/README.adoc` also needs to be updated, though. This just occurred to me today, sorry to add another thing. It's really out of date since the move to Gradle so if you prefer to go ahead with merging this I can pick up fixing the README separately. Up to you.

When I ran Jekyll locally in my earlier work with this, I needed to upgrade Ruby to 2.7 (the one that came with my OS was 2.3.x). Jekyll 3 does not run on Ruby 2.7, so if we don't make parallel changes in branch_8x for the Ant build, we'll get ourselves stuck when we need to do local builds for 8.x releases of the Ref Guide (it's been a pain today!). I'll take a stab at a new PR on this same SOLR-14900 for that.

PR #1925, opened 2020-09-28
Cleanup DWPT state handling

s1monw:
DWPT currently has no real notion of a state but it's lifecycle really
requires such a notion. We move DWPTs from active to flush pending to flushing
and execute certain actions like RAM accounting based on these states. To simplify
the transitions and the concurrency involved in it, it makes sense to formalize the
transitions and if it can happen under lock or not.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/DocumentsWriter.java
+++ b/lucene/core/src/java/org/apache/lucene/index/DocumentsWriter.java
@@ -429,10 +429,10 @@ final class DocumentsWriter implements Closeable, Accountable {
       final boolean isUpdate = delNode != null && delNode.isDelete();
       flushingDWPT = flushControl.doAfterDocument(dwpt, isUpdate);
     } finally {
-      if (dwpt.isFlushPending() || dwpt.isAborted()) {
-        dwpt.unlock();
-      } else {
+      if (dwpt.getState() == DocumentsWriterPerThread.State.ACTIVE && dwpt.isAborted() == false) {
         perThreadPool.marksAsFreeAndUnlock(dwpt);
+      } else {
+        dwpt.unlock();
       }
       assert dwpt.isHeldByCurrentThread() == false : "we didn't release the dwpt even on abort";
     }
@@ -446,7 +446,7 @@ final class DocumentsWriter implements Closeable, Accountable {
   private boolean doFlush(DocumentsWriterPerThread flushingDWPT) throws IOException {
     boolean hasEvents = false;
     while (flushingDWPT != null) {
-      assert flushingDWPT.hasFlushed() == false;
+      assert flushingDWPT.getState() == DocumentsWriterPerThread.State.FLUSHING : "expected FLUSHING but was: " + flushingDWPT.getState();
       hasEvents = true;
       boolean success = false;
       DocumentsWriterFlushQueue.FlushTicket ticket = null;
--- a/lucene/core/src/java/org/apache/lucene/index/DocumentsWriterFlushControl.java
+++ b/lucene/core/src/java/org/apache/lucene/index/DocumentsWriterFlushControl.java
@@ -157,25 +157,18 @@ final class DocumentsWriterFlushControl implements Accountable, Closeable {
   }
 
   DocumentsWriterPerThread doAfterDocument(DocumentsWriterPerThread perThread, boolean isUpdate) {
-    final long delta = perThread.getCommitLastBytesUsedDelta();
+    final long delta = perThread.commitLastBytesUsed();
     synchronized (this) {
-      // we need to commit this under lock but calculate it outside of the lock to minimize the time this lock is held
-      // per document. The reason we update this under lock is that we mark DWPTs as pending without acquiring it's
-      // lock in #setFlushPending and this also reads the committed bytes and modifies the flush/activeBytes.
-      // In the future we can clean this up to be more intuitive.
-      perThread.commitLastBytesUsed(delta);
       try {
         /*
          * We need to differentiate here if we are pending since setFlushPending
          * moves the perThread memory to the flushBytes and we could be set to
          * pending during a delete
          */
-        if (perThread.isFlushPending()) {
-          flushBytes += delta;
-          assert updatePeaks(delta);
-        } else {
-          activeBytes += delta;
-          assert updatePeaks(delta);
+        activeBytes += delta;
+        assert updatePeaks(delta);
+        if (perThread.isFlushPending() == false) {
+          assert perThread.getState() == DocumentsWriterPerThread.State.ACTIVE : "expected ACTIVE state but was: " + perThread.getState();
           if (isUpdate) {
             flushPolicy.onUpdate(this, perThread);
           } else {
@@ -235,7 +228,8 @@ final class DocumentsWriterFlushControl implements Accountable, Closeable {
     assert flushingWriters.contains(dwpt);
     try {
       flushingWriters.remove(dwpt);
-      flushBytes -= dwpt.getLastCommittedBytesUsed();
+      dwpt.transitionTo(DocumentsWriterPerThread.State.FLUSHED);
+      flushBytes -= dwpt.ramBytesUsed();
       assert assertMemory();
     } finally {
       try {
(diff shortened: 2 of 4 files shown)


s1monw:
@mikemccand can you take a look at this?

s1monw:
this is a followup from https://github.com/apache/lucene-solr/pull/1918

lucene/core/src/java/org/apache/lucene/index/DocumentsWriterPerThread.java
```diff
@@ -593,9 +568,61 @@ void unlock() {
...
+    FLUSHING(FLUSH_PENDING, true),
+    /**
+     * The DWPT has been flushed and is ready to be garbage collected.
+     */
+    FLUSHED(FLUSHING, false);
+
+    private final State previousState;
+    final boolean mustHoldLock; // only for asserts
+
+    State(State previousState, boolean mustHoldLock) {
```
dweiss, line 614: Now you only need a formal correctness solver for this Petri net and you'll be good. 👍

lucene/core/src/test/org/apache/lucene/index/TestIndexWriter.java
```diff
@@ -2933,7 +2933,7 @@ public void testFlushLargestWriter() throws IOException, InterruptedException {
     int numRamDocs = w.numRamDocs();
     int numDocsInDWPT = largestNonPendingWriter.getNumDocsInRAM();
     assertTrue(w.flushNextBuffer());
-    assertTrue(largestNonPendingWriter.hasFlushed());
+    assertEquals(DocumentsWriterPerThread.State.FLUSHED, largestNonPendingWriter.getState());
```
dweiss, line 2936: Wouldn't this be more pleasant to read?
assertTrue largestNonPendingWriter.inState(DocumentsWriterPerThread.State.FLUSHED)?

dweiss (approved):
I love state machines. +1 from me.

lucene/core/src/java/org/apache/lucene/index/DocumentsWriterPerThread.java
s1monw: I gotta solve the halting problem first many, I am on it.

lucene/core/src/test/org/apache/lucene/index/TestIndexWriter.java
s1monw: I guess it would. I am not sure if we should really add more methods just for tests?

dweiss: ok, fine it is then.

lucene/core/src/java/org/apache/lucene/index/DocumentsWriterPerThread.java
dweiss: You have time until the end of this week [George Dantzig]

jpountz (approved):
This reads much easier to me. I also like the move from ramBytesUsed/getLastCommitedBytesUsed to ramBytesUsedInternal/ramBytesUsed.

lucene/core/src/java/org/apache/lucene/index/DocumentsWriterFlushControl.java
```diff
@@ -157,25 +157,18 @@ private boolean updatePeaks(long delta) {
...
-        if (perThread.isFlushPending()) {
-          flushBytes += delta;
-          assert updatePeaks(delta);
-        } else {
-          activeBytes += delta;
-          assert updatePeaks(delta);
+        activeBytes += delta;
+        assert updatePeaks(delta);
+        if (perThread.isFlushPending() == false) {
+          assert perThread.getState() == DocumentsWriterPerThread.State.ACTIVE : "expected ACTIVE state but was: " + perThread.getState();
```
dnhatn, line 171: Do we still need `isFlushPending` method? Should we compare the state of `perThread` to ACTIVE or FLUSH_PENDING instead.

lucene/core/src/java/org/apache/lucene/index/DocumentsWriterPerThread.java
```diff
@@ -593,9 +568,61 @@ void unlock() {
...
+     * The DWPT should not receive any further documents and is current flushing or queued up for flushing.
+     */
+    FLUSHING(FLUSH_PENDING, true),
+    /**
+     * The DWPT has been flushed and is ready to be garbage collected.
+     */
+    FLUSHED(FLUSHING, false);
+
+    private final State previousState;
+    final boolean mustHoldLock; // only for asserts
```
dnhatn, line 612: This can be private?

dnhatn (approved):
I've left a few nits. Nice cleanup! Thanks, Simon.

lucene/core/src/java/org/apache/lucene/index/DocumentsWriterFlushControl.java
s1monw: that makes sense to me.

lucene/core/src/java/org/apache/lucene/index/DocumentsWriterFlushControl.java
```diff
@@ -157,49 +157,42 @@ private boolean updatePeaks(long delta) {
...
          * moves the perThread memory to the flushBytes and we could be set to
          * pending during a delete
          */
-        if (perThread.isFlushPending()) {
-          flushBytes += delta;
-          assert updatePeaks(delta);
-        } else {
-          activeBytes += delta;
-          assert updatePeaks(delta);
+        activeBytes += delta;
```
mikemccand, line 168: Hmm, why are we removing the conditional on "flush pending" here (and always updating `activeBytes`, never `flushBytes`)?  I guess `setFlushPending` will move all `activeBytes` over to `flushBytes`, so it is fine for us to always increment `activeBytes` here?

Edit: hmm it looks like we are also removing `setFlushPending`'s moving of `activeBytes` to `flushBytes` ;)  So now I don't quite understand how we are changing the RAM accounting here.

Edit again!: OK I see, we moved the RAM shifting into `checkoutFlushableWriter`, OK good!

Should we fix (or maybe just remove) the above comment now?

lucene/core/src/java/org/apache/lucene/index/DocumentsWriterPerThread.java
```diff
@@ -593,9 +568,61 @@ void unlock() {
...
+   * Returns the DWPTs current state.
    */
-  boolean hasFlushed() {
-    return hasFlushed.get() == Boolean.TRUE;
+  State getState() {
+    return state;
   }
+
+  /**
+   * Transitions the DWPT to the given state of fails if the transition is invalid.
```
mikemccand, line 578: s/`of`/`or`?

lucene/core/src/java/org/apache/lucene/index/DocumentsWriterPerThread.java
```diff
@@ -513,48 +515,21 @@ public String toString() {
...
-
-  /**
-   * Commits the current {@link #ramBytesUsed()} and stores it's value for later reuse.
-   * The last committed bytes used can be retrieved via {@link #getLastCommittedBytesUsed()}
-   */
-  void commitLastBytesUsed(long delta) {
-    assert isHeldByCurrentThread();
-    assert getCommitLastBytesUsedDelta() == delta : "delta has changed";
-    lastCommittedBytesUsed += delta;
+    return state == State.FLUSH_PENDING;
```
mikemccand, line 522: Maybe we should remove `isFlushPending` method entirely and expect callers to check `state == State.FLUSH_PENDING` themselves?

lucene/core/src/java/org/apache/lucene/index/DocumentsWriterFlushControl.java
```diff
@@ -295,27 +289,32 @@ public synchronized void waitForFlush() {
...
    */
-  public synchronized void setFlushPending(DocumentsWriterPerThread perThread) {
-    assert !perThread.isFlushPending();
+  synchronized void setFlushPending(DocumentsWriterPerThread perThread) {
+    assert !perThread.isFlushPending() : "state: " + perThread.getState();
     if (perThread.getNumDocsInRAM() > 0) {
-      perThread.setFlushPending(); // write access synced
-      final long bytes = perThread.getLastCommittedBytesUsed();
-      flushBytes += bytes;
-      activeBytes -= bytes;
```
mikemccand, line 304: Hmm we are no longer moving `bytes` from `activeBytes` to `flushedBytes` here?  Are we (the caller here?) doing that elsewhere?

lucene/core/src/java/org/apache/lucene/index/DocumentsWriterFlushControl.java
```diff
@@ -157,49 +157,42 @@ private boolean updatePeaks(long delta) {
   }
 
   DocumentsWriterPerThread doAfterDocument(DocumentsWriterPerThread perThread, boolean isUpdate) {
-    final long delta = perThread.getCommitLastBytesUsedDelta();
+    final long delta = perThread.commitLastBytesUsed();
     synchronized (this) {
-      // we need to commit this under lock but calculate it outside of the lock to minimize the time this lock is held
-      // per document. The reason we update this under lock is that we mark DWPTs as pending without acquiring it's
-      // lock in #setFlushPending and this also reads the committed bytes and modifies the flush/activeBytes.
-      // In the future we can clean this up to be more intuitive.
```
mikemccand, line 165: ;)

lucene/core/src/java/org/apache/lucene/index/DocumentsWriter.java
```diff
@@ -446,7 +446,7 @@ long updateDocuments(final Iterable<? extends Iterable<? extends IndexableField>
   private boolean doFlush(DocumentsWriterPerThread flushingDWPT) throws IOException {
     boolean hasEvents = false;
     while (flushingDWPT != null) {
-      assert flushingDWPT.hasFlushed() == false;
+      assert flushingDWPT.getState() == DocumentsWriterPerThread.State.FLUSHING : "expected FLUSHING but was: " + flushingDWPT.getState();
```
mikemccand, line 449: Maybe add a `boolean assertState(State)` to DWPT that does the above `assert` and then returns `true` so we can call it under `assert`?  Only downside is then we have double looking `assert`, i.e. `assert flushingDWPT.assertState(State.FLUSHING)`.  Looks like we are doing this in at least three places here ...

mikemccand (approved):
Change looks great!  I also love state machines!

I was a bit confused at first about what happened to the RAM accounting, but figured it out eventually :)  We have some asserts on the RAM accounting so I think if something were wrong, the tests would notice.

Thanks @s1monw

PR #1927, opened 2020-09-28
SOLR-12987: Deprecated plugins are logged once and with log category …

dsmiley:
…org.apache.solr.DEPRECATED

https://issues.apache.org/jira/browse/SOLR-12987


janhoy (approved):
Does cname include package name?

dsmiley:
"cname" is the class name as configured/given, which might be fully qualified (have the package) or might be abbreviated like "solr.DataImportHandler".

PR #1929, opened 2020-09-29
LUCENE-9548: Apache repository publishing

dweiss:


gradle/maven/defaults-maven.gradle
```diff
@@ -102,29 +123,74 @@ configure(subprojects.findAll { it.path in rootProject.published }) { prj ->
...
-          jars(MavenPublication) {
-            from components.java
-            groupId = project.group
-            artifactId = project.archivesBaseName
+          inceptionYear = "2000"
 
-            artifact sourcesJar
+          issueManagement {
+            system = "JIRA"
+            url = "https://issues.apache.org/jira/browse/LUCENE"
```
madrob, line 147: This should be split by project?

dweiss: All of it should be project-dependent, really... But yes -- feel free to commit a patch.

gradle/maven/defaults-maven.gradle
```diff
@@ -76,7 +78,26 @@ configure(subprojects.findAll { it.path in rootProject.published }) { prj ->
   apply plugin: 'signing'
 
   publishing {
-    // TODO: Add publishing repository details.
+    repositories {
+      maven {
+        name = "ApacheSnapshots"
+        url = "https://repository.apache.org/content/repositories/snapshots"
```
uschindler, line 84: Can't use the property defined before?: `apacheNexusUrl`

gradle/maven/defaults-maven.gradle
```diff
@@ -137,8 +254,17 @@ configure(subprojects.findAll { it.path in rootProject.published }) { prj ->
         }
       }
 
+      // Add aliases of convention tasks with shorter names.
+      task mavenToApacheSnapshots() {
```
uschindler, line 258: Maybe put this into separate file like maven-local.gradle?

uschindler (changes requested):

gradle/maven/defaults-maven.gradle
dweiss: Don't know if it makes sense. The "local" maven gradle config is there mostly so that I could review the artifacts without polluting my local Maven repo. It could be removed later. The rest of maven config does make sense as a single unit. Let's leave it for now and try to publish those snapshots. If needed, we can always change later.

PR #1931, opened 2020-09-29
SOLR-14597 Advanced Query Parser (WIP)

gus-asf:
still to do: 

 - move lucene classes to lucene ticket
 - Investigate why TestPackages doesn't fail now (did fail, I had fixed it, but new version doesn't need my fix, so need to understand why)
 - Docs!
 - Package?


risdenk:
I think this is being followed up here: https://github.com/apache/solr/pull/128

PR #1932, opened 2020-09-30
Reuse crypto keys in reference impl

noblepaul:
The key pair generation is expensive and instead of disabling public key handler We should just reuse a single key pair in the entire test suite.

### DIFF
--- a/solr/core/src/java/org/apache/solr/security/PublicKeyHandler.java
+++ b/solr/core/src/java/org/apache/solr/security/PublicKeyHandler.java
@@ -32,11 +32,13 @@ import java.security.spec.InvalidKeySpecException;
 public class PublicKeyHandler extends RequestHandlerBase {
   public static final String PATH = "/admin/info/key";
 
+  //This is an optimization for tests only
+  public static volatile CryptoKeys.RSAKeyPair REUSABLE_KEYPAIR ;
   final CryptoKeys.RSAKeyPair keyPair;
 
   @VisibleForTesting
-  public PublicKeyHandler() {
-    keyPair = new CryptoKeys.RSAKeyPair();
+  public PublicKeyHandler() throws IOException, InvalidKeySpecException {
+    keyPair = createKeyPair(null);
   }
 
   public PublicKeyHandler(CloudConfig config) throws IOException, InvalidKeySpecException {
@@ -44,6 +46,8 @@ public class PublicKeyHandler extends RequestHandlerBase {
   }
 
   private CryptoKeys.RSAKeyPair createKeyPair(CloudConfig config) throws IOException, InvalidKeySpecException {
+    CryptoKeys.RSAKeyPair reused = REUSABLE_KEYPAIR;
+    if(reused != null) return reused;
     if (config == null) {
       return new CryptoKeys.RSAKeyPair();
     }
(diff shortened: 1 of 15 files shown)


solr/core/src/java/org/apache/solr/security/PublicKeyHandler.java
```diff
@@ -32,18 +32,22 @@
 public class PublicKeyHandler extends RequestHandlerBase {
   public static final String PATH = "/admin/info/key";
 
+  //This is an optimization for tests only
+  public static volatile CryptoKeys.RSAKeyPair REUSABLE_KEYPAIR ;
```
madrob, line 36: Why not use the key-pair that's already on disk? See https://github.com/apache/lucene-solr/blob/master/solr/test-framework/src/java/org/apache/solr/SolrTestCaseJ4.java#L289-L290

solr/test-framework/src/java/org/apache/solr/SolrTestCase.java
```diff
@@ -124,6 +126,17 @@
 
   protected volatile static PerThreadExecService testExecutor;
 
+  private static final CryptoKeys.RSAKeyPair reusedKeys = new CryptoKeys.RSAKeyPair();
```
madrob, line 129: Why not use the key-pair that's already on disk? See https://github.com/apache/lucene-solr/blob/master/solr/test-framework/src/java/org/apache/solr/SolrTestCaseJ4.java#L289-L290

solr/test-framework/src/java/org/apache/solr/SolrTestCase.java
```diff
@@ -225,7 +238,7 @@ public static void setDefaultConfigDirSysPropIfNotSet() throws Exception {
     System.setProperty("solr.clustering.enabled", "false");
     System.setProperty("solr.peerSync.useRangeVersions", String.valueOf(random().nextBoolean()));
     System.setProperty("zookeeper.nio.directBufferBytes", Integer.toString(32 * 1024 * 2));
-    System.setProperty("solr.disablePublicKeyHandler", "true");
+    enableReuseOfCryptoKeys();
```
markrmiller, line 241: Looks like I had this reversed. We should likely enable reuse below in the non Nightly run and disable for the closer to prod Nightly run.

solr/test-framework/src/java/org/apache/solr/SolrTestCase.java
```diff
@@ -279,7 +292,6 @@ public static void setDefaultConfigDirSysPropIfNotSet() throws Exception {
       System.setProperty("solr.http2solrclient.maxpool.size", "12");
       System.setProperty("solr.http2solrclient.pool.keepalive", "1500");
```
markrmiller, line 294: Let's enable reuse here instead.

markrmiller (changes requested):

solr/test-framework/src/java/org/apache/solr/SolrTestCase.java
```diff
@@ -225,7 +238,7 @@ public static void setDefaultConfigDirSysPropIfNotSet() throws Exception {
     System.setProperty("solr.clustering.enabled", "false");
     System.setProperty("solr.peerSync.useRangeVersions", String.valueOf(random().nextBoolean()));
     System.setProperty("zookeeper.nio.directBufferBytes", Integer.toString(32 * 1024 * 2));
-    System.setProperty("solr.disablePublicKeyHandler", "true");
+    enableReuseOfCryptoKeys();
 
     if (!TEST_NIGHTLY) {
```
markrmiller, line 243: Let's enable reuse here instead.

solr/test-framework/src/java/org/apache/solr/SolrTestCase.java
noblepaul: This is for just testcase

PR #1937, opened 2020-10-01
LUCENE-9541 ConjunctionDISI sub-iterators check

mayya-sharipova:
Ensure sub-iterators of a conjunction iterator don't advance outside
of it.
Add assertions for checking that all sub-iterators are always on the
same document.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/ConjunctionDISI.java
+++ b/lucene/core/src/java/org/apache/lucene/search/ConjunctionDISI.java
@@ -142,11 +142,13 @@ public final class ConjunctionDISI extends DocIdSetIterator {
       List<DocIdSetIterator> allIterators,
       List<TwoPhaseIterator> twoPhaseIterators) {
 
-    // assert that all sub-iterators are on the same doc ID
+    // check that all sub-iterators are on the same doc ID
     int curDoc = allIterators.size() > 0 ? allIterators.get(0).docID() : twoPhaseIterators.get(0).approximation.docID();
     boolean iteratorsOnTheSameDoc = allIterators.stream().allMatch(it -> it.docID() == curDoc);
     iteratorsOnTheSameDoc = iteratorsOnTheSameDoc && twoPhaseIterators.stream().allMatch(it -> it.approximation().docID() == curDoc);
-    assert iteratorsOnTheSameDoc : "Sub-iterators of ConjunctionDISI are not the same document!";
+    if (iteratorsOnTheSameDoc == false) {
+      throw new IllegalArgumentException("Sub-iterators of ConjunctionDISI are not the same document!");
+    }
 
     long minCost = allIterators.stream().mapToLong(DocIdSetIterator::cost).min().getAsLong();
     List<BitSetIterator> bitSetIterators = new ArrayList<>();
@@ -236,7 +238,7 @@ public final class ConjunctionDISI extends DocIdSetIterator {
 
   @Override
   public int advance(int target) throws IOException {
-    assertItersOnSameDoc();
+    assert assertItersOnSameDoc() : "Sub-iterators of ConjunctionDISI are not the same document!";
     return doNext(lead1.advance(target));
   }
 
@@ -247,7 +249,7 @@ public final class ConjunctionDISI extends DocIdSetIterator {
 
   @Override
   public int nextDoc() throws IOException {
-    assertItersOnSameDoc();
+    assert assertItersOnSameDoc() : "Sub-iterators of ConjunctionDISI are not the same document!";
     return doNext(lead1.nextDoc());
   }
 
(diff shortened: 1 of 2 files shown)


lucene/core/src/java/org/apache/lucene/search/ConjunctionDISI.java
```diff
@@ -140,6 +141,13 @@ private static void addTwoPhaseIterator(TwoPhaseIterator twoPhaseIter, List<DocI
   private static DocIdSetIterator createConjunction(
       List<DocIdSetIterator> allIterators,
       List<TwoPhaseIterator> twoPhaseIterators) {
+
+    // assert that all sub-iterators are on the same doc ID
```
mayya-sharipova, line 145: @jpountz  Is this a correct assumption?  or could we start with iterators that are not on the same document?

lucene/core/src/java/org/apache/lucene/search/ConjunctionDISI.java
```diff
@@ -140,6 +141,13 @@ private static void addTwoPhaseIterator(TwoPhaseIterator twoPhaseIter, List<DocI
...
+    int curDoc = allIterators.size() > 0 ? allIterators.get(0).docID() : twoPhaseIterators.get(0).approximation.docID();
+    boolean iteratorsOnTheSameDoc = allIterators.stream().allMatch(it -> it.docID() == curDoc);
+    iteratorsOnTheSameDoc = iteratorsOnTheSameDoc && twoPhaseIterators.stream().allMatch(it -> it.approximation().docID() == curDoc);
+    assert iteratorsOnTheSameDoc : "Sub-iterators of ConjunctionDISI are not the same document!";
```
jpountz, line 149: let's make it an IllegalArgumentException?

lucene/core/src/java/org/apache/lucene/search/ConjunctionDISI.java
jpountz: this is the invariant between iterations indeed

lucene/core/src/java/org/apache/lucene/search/ConjunctionDISI.java
```diff
@@ -227,6 +236,7 @@ private int doNext(int doc) throws IOException {
 
   @Override
   public int advance(int target) throws IOException {
+    assertItersOnSameDoc();
```
jpountz, line 239: can we avoid iterating over all sub iterators unless assertions are enabled?

lucene/core/src/java/org/apache/lucene/search/ConjunctionDISI.java
mayya-sharipova: addressed in 74151e3

lucene/core/src/java/org/apache/lucene/search/ConjunctionDISI.java
mayya-sharipova: addressed in 74151e3

lucene/core/src/test/org/apache/lucene/search/TestConjunctionDISI.java
```diff
@@ -391,4 +391,20 @@ public void testCollapseSubConjunctionDISIs() throws IOException {
   public void testCollapseSubConjunctionScorers() throws IOException {
     testCollapseSubConjunctions(true);
   }
+
+  public void testIllegalAdvancementOfSubIteratorsTripsAssertion() throws IOException {
+    int maxDoc = 100;
```
jpountz, line 396: I think we need to call something like `assumeTrue(assertsAreEnabled)` since this test requires assertions to be enabled to succeed.

lucene/core/src/test/org/apache/lucene/search/TestConjunctionDISI.java
```diff
@@ -391,4 +391,20 @@ public void testCollapseSubConjunctionDISIs() throws IOException {
...
+
+    DocIdSetIterator[] iterators = new DocIdSetIterator[numIterators];
+    for (int i = 0; i < iterators.length; ++i) {
+      iterators[i] = new BitDocIdSet(set).iterator();
+    }
+    final DocIdSetIterator conjunction = ConjunctionDISI.intersectIterators(Arrays.asList(iterators));
+    int idx = TestUtil.nextInt(random() , 0, iterators.length-1);
+    iterators[idx].nextDoc(); // illegally advancing one of the sub-iterators outside of the conjunction iterator
+    AssertionError ex = expectThrows(AssertionError.class, () -> conjunction.nextDoc());
+    assertEquals(ex.getMessage(), "Sub-iterators of ConjunctionDISI are not the same document!");
```
jpountz, line 408: can you revert the two arguments to assertEquals? it expects the expected value first and the actual value second. While swapping arguments doesn't cause major issues, it creates confusing error messages when the assertion fails.

jpountz (approved):

lucene/core/src/test/org/apache/lucene/search/TestConjunctionDISI.java
mayya-sharipova: Thanks @jpountz, very good point, addressed in bf9a3e8

lucene/core/src/test/org/apache/lucene/search/TestConjunctionDISI.java
mayya-sharipova: Thanks @jpountz, my mistake, addressed in bf9a3e8

jpountz (approved):

PR #1940, opened 2020-10-02
LUCENE-9552: Adds a LatLonPoint query that accepts an array of LatLonGeometries

iverase:
New query that accepts an array of LatLonGeometries.


rmuir:
Why is this needed when the existing Polygon is a multipolygon and optimizes for that case? Instead of making an array of polygons, just use a single MultiPolygon?

iverase:
I have two motivations for this change:

1) Finding all points at a distance of a line can be approximated  by creating a polygon and two circles and finding all points that is inside any of those geometries. Currently three queries are need inside a boolean OR. With this approach it can be executed in a single query.

2) API-wise, it simplifies the implementation as you don't need to know which geometry you are dealing with to create the query.

rmuir:
Thanks for the explanation. For the first use-case, since its an approximation, why use circles instead of n-gons to approximate the circles? Then its just a simple multipolygon and should be fast?

iverase:
Yes, approximating the circle to a polygon would work. You will have a penalty in performance depending how many points you use for approximating the circle (which is something the user will have to decide).

But I am afraid if this solution scales well. The example I gave was the simple case, in general the input can be a path that can contains hundred of line segments. That mean you can end up with hundred of circles that you need to approximate using polygons with potentially thousands of edges. I would expect that at some point it can get very inefficient, even just building the query.

I would find a pity not to expose the capability accepting heterogeneous geometries as it is something that is already supported internally.

iverase:
@rmuir, I am trying to understand why adding this new API seems so problematic, I would be more incline to deprecate `newPolygonQuery` in favour of this new one. My feeling is that we should not force the user to decide which API to use but to provide the geometries and we internally decide the most optimal way to execute them?

nknize:
Circle approximation vs distance aside I think the real benefit of this PR is that it creates a single query for multiple geometry types whereas right now we have to create a Boolean query for handling GeometryCollections? I like the idea of handling the query logic in a single query vs having to build a Boolean query of many LatLon queries.

nknize:
Any opposition to the new query @rmuir?

PR #1941, opened 2020-10-02
LUCENE-9554: Expose IndexWriter#pendingNumDocs

dnhatn:
Some applications can use the pendingNumDocs from IndexWriter to estimate that the number of documents of an index is very close to the hard limit so that it can reject writes without constructing Lucene documents.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
+++ b/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
@@ -5465,6 +5465,7 @@ public class IndexWriter implements Closeable, TwoPhaseCommit, Accountable,
 
   /**
    * Returns the number of documents in the index including documents are being added (i.e., reserved).
+   * @lucene.experimental
    */
   public long getPendingNumDocs() {
     return pendingNumDocs.get();
--- a/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
+++ b/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
@@ -40,7 +40,6 @@ import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.BooleanSupplier;
-import java.util.function.IntConsumer;
 import java.util.function.IntPredicate;
 import java.util.stream.Collectors;
 import java.util.stream.StreamSupport;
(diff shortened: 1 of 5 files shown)


lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
```diff
@@ -1072,7 +1073,14 @@ public IndexWriter(Directory d, IndexWriterConfig conf) throws IOException {
 
       config.getFlushPolicy().init(config);
       bufferedUpdatesStream = new BufferedUpdatesStream(infoStream);
-      docWriter = new DocumentsWriter(flushNotifications, segmentInfos.getIndexCreatedVersionMajor(), pendingNumDocs,
+      final IntConsumer reserveDocs = numDocs -> {
+        if (numDocs > 0) {
+          reserveDocs(numDocs);
```
mikemccand, line 1078: I'm confused why we are calling separate methods when `numDocs` is negative or not?

mikemccand (commented):
+1 to make `pendingNumDocs` accessible, but I was a little confused by why we separate negative and non-negative cases in `IndexWriter`...

dnhatn:
> I was a little confused by why we separate negative and non-negative cases in IndexWriter...

@mikemccand Thank you for looking.  I reverted that change and will make it in a separate PR

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
```diff
@@ -5463,6 +5463,13 @@ private void tooManyDocs(long addedNumDocs) {
     throw new IllegalArgumentException("number of documents in the index cannot exceed " + actualMaxDocs + " (current document count is " + pendingNumDocs.get() + "; added numDocs is " + addedNumDocs + ")");
   }
 
+  /**
+   * Returns the number of documents in the index including documents are being added (i.e., reserved).
+   */
```
mikemccand, line 5468: Maybe add `@lucene.experimental`?  We are exposing (slightly) internal details about `IndexWriter` so maybe we need to reserve the right to change this API in the future ...

mikemccand (approved):
Thanks @dnhatn!  I left a small javadoc comment.

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
dnhatn: ++. I pushed https://github.com/apache/lucene-solr/pull/1941/commits/61d4fbc4a01cbfc50f386a8a88064e82d1027ee0

dnhatn:
Thanks Mike!

PR #1942, opened 2020-10-02
SOLR-14910: Use in-line tags for logger declarations in Gradle ValidateLogCalls that are non-standard, change //logok to //nowarn

ErickErickson:
Most of this is just changing //logok to //nowarn. 

The substantive changes are in validate-log-calls.gradle, taking out the special handling for several files and _not_ producing failures if there's a //nowarn on the line instead, plus adding a few //nowarn tags in the files that were handled specially.

### DIFF
--- a/solr/core/src/java/org/apache/solr/servlet/HttpSolrCall.java
+++ b/solr/core/src/java/org/apache/solr/servlet/HttpSolrCall.java
@@ -493,7 +493,7 @@ public class HttpSolrCall {
     }
     if (statusCode == AuthorizationResponse.FORBIDDEN.statusCode) {
       if (log.isDebugEnabled()) {
-        log.debug("UNAUTHORIZED auth header {} context : {}, msg: {}", req.getHeader("Authorization"), context, authResponse.getMessage()); // logOk
+        log.debug("UNAUTHORIZED auth header {} context : {}, msg: {}", req.getHeader("Authorization"), context, authResponse.getMessage()); // nowarn
       }
       sendError(statusCode,
           "Unauthorized request, Response code: " + statusCode);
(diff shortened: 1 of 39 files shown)


solr/core/src/java/org/apache/solr/servlet/HttpSolrCall.java
```diff
@@ -493,7 +493,7 @@ Action authorize() throws IOException {
     }
     if (statusCode == AuthorizationResponse.FORBIDDEN.statusCode) {
       if (log.isDebugEnabled()) {
-        log.debug("UNAUTHORIZED auth header {} context : {}, msg: {}", req.getHeader("Authorization"), context, authResponse.getMessage()); // logOk
+        log.debug("UNAUTHORIZED auth header {} context : {}, msg: {}", req.getHeader("Authorization"), context, authResponse.getMessage()); // nowarn
```
dsmiley, line 496: What does our source validation complain about here?

Many of the logok/nowarn places look fine to me at a glance but I'm no match for the logging policeman ;-)

ErickErickson:
It was decided (there’s a phrase I hate, used to avoid taking responsibility for something)…

Anyway, Andras Salamon and I decided that it was often an error to have logging calls with exception.getMessage() rather than the full stack trace, see: SOLR-14523 and the checking code isn’t sophisticated enough to figure out that this call to getMessage() has nothing to do with an exception. This is a case that doesn’t matter whether it’s in an "if (loglevelenabled)” clause or not…

WDYT about adding a reference to SOLR-14523 to the output message? That’s not obvious on a code review unfortunately...

dsmiley:
>  it was often an error to have logging calls with exception.getMessage() 

Ah; right, I agree with this very much and I thought I proposed this rule.

> WDYT about adding a reference to SOLR-14523 to the output message? 

to the validation rule error?  Yes; I think referring to JIRA issues is great -- gives people a place to go to learn/understand.

ErickErickson:
Pushed to master

PR #1943, opened 2020-10-02
LUCENE-9555: Advance conjuction Iterator for two phase iteration

mayya-sharipova:
PR #1351 introduced a sort optimization where
documents can be skipped.  
But there was a bug in case we were using two phase
approximation, as we would advance it without advancing
an overall conjunction iterator.

This patch fixed it.

Relates to #1351

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -142,9 +142,6 @@ Improvements
 
 Bug fixes
 
-* LUCENE-9555: For sort optimization, create conjunction between scorerIterator
-  and collectorIterator only if scorerIterator has not been advanced yet. (Mayya Sharipova)
-
 * LUCENE-8663: NRTCachingDirectory.slowFileExists may open a file while 
   it's inaccessible. (Dawid Weiss)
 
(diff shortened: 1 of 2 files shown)


mayya-sharipova:
This patch also addresses the failure of `TestUnifiedHighlighterStrictPhrases.testBasics`.
This test was using `ReqExclBulkScorer` that lead to `scorerIterator` being already on max document when we were trying to make a conjunction between `scorerIterator` and `collectorIterator`.

jpountz (commented):
It's a pity to disable this great optimization when scoring ranges of doc IDs versus scoring the entire doc ID space at once. Maybe we can apply the leap-frog logic without using ConjunctionDISI in that method, or alternatively wrap the scorer DISI in a way that makes usage of ConjunctionDISI legal?

```java
class RangeDISIWrapper {

  private final DocIdSetIterator in;
  private final int min, max;
  private int doc = -1;

  @Override
  public int advance(int target) {
    target = Math.max(min, target);
    if (target >= max) {
      return NO_MORE_DOCS;
    }
    return doc = in.advance(target);
  }

}
```

mayya-sharipova:
@jpountz Sorry for the noise, I have found the cause of this error, and the latest commit addresses it.

jpountz (approved):

PR #1945, opened 2020-10-04
SOLR-14680: revert from 8x

noblepaul:


noblepaul:
As requested by @sigram I'm reverting this from 8x

PR #1946, opened 2020-10-04
LUCENE-9558: Clean up package name conflicts for analyzers-icu.

mocobeta:
See https://issues.apache.org/jira/browse/LUCENE-9558

- move classes under "o.a.l.collation" to "o.a.l.a.icu"
- classes under "o.a.l.collation.tokenattributes" to "o.a.l.a.icu.tokenattributes"
- delete "o.a.l.collation" from analyzers-icu


msokolov:
I guess you got Robert's +1; the naming seems more sensible now to me, and we do it as part of major version upgrade.

BTW I stumbled on this split-package javadoc issue trying to add some utilities in lucene/misc/src/java/org/apache/lucene/util/hnsw - I guess I won't do that, but I wonder where benchmarking tools for hnsw belong? I guess we can put in core,  analogous to what we did with tools over in analysis/kuromoji (where we moved DictBuilder from src/tools into src/java)

mocobeta:
> BTW I stumbled on this split-package javadoc issue trying to add some utilities in lucene/misc/src/java/org/apache/lucene/util/hnsw - I guess I won't do that, but I wonder where benchmarking tools for hnsw belong?

Not pretty sure, but we could put benchmarking tools in lucene/benchmark?

PR #1948, opened 2020-10-05
LUCENE-9536: Optimize OrdinalMap when one segment contains all distinct values.

jtibshirani:
For doc values that are not too high cardinality, it is common for some large
segments to contain all distinct values. In this case, we can check if the first
segment ords map perfectly to global ords, and if so store the global ord deltas
and first segment indices as `LongValues.ZEROES`
to save some space.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/OrdinalMap.java
+++ b/lucene/core/src/java/org/apache/lucene/index/OrdinalMap.java
@@ -172,10 +172,12 @@ public class OrdinalMap implements Accountable {
 
   /** Cache key of whoever asked for this awful thing */
   public final IndexReader.CacheKey owner;
+  // number of global ordinals
+  final long valueCount;
   // globalOrd -> (globalOrd - segmentOrd) where segmentOrd is the the ordinal in the first segment that contains this term
-  final PackedLongValues globalOrdDeltas;
+  final LongValues globalOrdDeltas;
   // globalOrd -> first segment container
-  final PackedLongValues firstSegments;
+  final LongValues firstSegments;
   // for every segment, segmentOrd -> globalOrd
   final LongValues segmentToGlobalOrds[];
   // the map from/to segment ids
@@ -271,13 +273,26 @@ public class OrdinalMap implements Accountable {
       globalOrd++;
     }
 
-    this.firstSegments = firstSegments.build();
-    this.globalOrdDeltas = globalOrdDeltas.build();
+    long ramBytesUsed = BASE_RAM_BYTES_USED + segmentMap.ramBytesUsed();
+    this.valueCount = globalOrd;
+
+    // If the first segment contains all of the global ords, then we can apply a small optimization
+    // and hardcode the first segments and global ord deltas as all zeroes.
+    if (ordDeltaBits.length > 0 && ordDeltaBits[0] == 0L && ordDeltas[0].size() == this.valueCount) {
+      this.firstSegments = LongValues.ZEROES;
+      this.globalOrdDeltas = LongValues.ZEROES;
+      ramBytesUsed += RamUsageEstimator.shallowSizeOf(LongValues.ZEROES);
+    } else {
+      PackedLongValues packedFirstSegments = firstSegments.build();
+      PackedLongValues packedGlobalOrdDeltas = globalOrdDeltas.build();
+      this.firstSegments = packedFirstSegments;
+      this.globalOrdDeltas = packedGlobalOrdDeltas;
+      ramBytesUsed += packedFirstSegments.ramBytesUsed() + packedGlobalOrdDeltas.ramBytesUsed();
+    }
+
     // ordDeltas is typically the bottleneck, so let's see what we can do to make it faster
     segmentToGlobalOrds = new LongValues[subs.length];
-    long ramBytesUsed = BASE_RAM_BYTES_USED + this.globalOrdDeltas.ramBytesUsed()
-      + this.firstSegments.ramBytesUsed() + RamUsageEstimator.shallowSizeOf(segmentToGlobalOrds)
-      + segmentMap.ramBytesUsed();
+    ramBytesUsed += RamUsageEstimator.shallowSizeOf(segmentToGlobalOrds);
     for (int i = 0; i < ordDeltas.length; ++i) {
       final PackedLongValues deltas = ordDeltas[i].build();
       if (ordDeltaBits[i] == 0L) {
@@ -317,6 +332,7 @@ public class OrdinalMap implements Accountable {
         ramBytesUsed += RamUsageEstimator.shallowSizeOf(segmentToGlobalOrds[i]);
       }
     }
+
     this.ramBytesUsed = ramBytesUsed;
   }
 
@@ -348,7 +364,7 @@ public class OrdinalMap implements Accountable {
    * Returns the total number of unique terms in global ord space.
    */
   public long getValueCount() {
-    return globalOrdDeltas.size();
+    return valueCount;
   }
 
   @Override
@@ -359,10 +375,9 @@ public class OrdinalMap implements Accountable {
   @Override
   public Collection<Accountable> getChildResources() {
     List<Accountable> resources = new ArrayList<>();
-    resources.add(Accountables.namedAccountable("global ord deltas", globalOrdDeltas));
-    resources.add(Accountables.namedAccountable("first segments", firstSegments));
     resources.add(Accountables.namedAccountable("segment map", segmentMap));
-    // TODO: would be nice to return actual child segment deltas too, but the optimizations are confusing
+    // TODO: would be nice to return the ordinal and segment maps too, but it's not straightforward
+    //  because of optimizations.
     return resources;
   }
 }
--- a/lucene/core/src/java/org/apache/lucene/index/OrdinalMap.java
+++ b/lucene/core/src/java/org/apache/lucene/index/OrdinalMap.java
@@ -281,7 +281,6 @@ public class OrdinalMap implements Accountable {
     if (ordDeltaBits.length > 0 && ordDeltaBits[0] == 0L && ordDeltas[0].size() == this.valueCount) {
       this.firstSegments = LongValues.ZEROES;
       this.globalOrdDeltas = LongValues.ZEROES;
-      ramBytesUsed += RamUsageEstimator.shallowSizeOf(LongValues.ZEROES);
     } else {
       PackedLongValues packedFirstSegments = firstSegments.build();
       PackedLongValues packedGlobalOrdDeltas = globalOrdDeltas.build();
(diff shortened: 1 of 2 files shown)


jtibshirani:
I used `TestOrdinalMap` to test a map with 10,000 terms in a scenario and ~10 segments. In the scenario where one segment contains all ordinal values, it shows a small improvement:

```
baseline bytes used: 11184
new bytes used: 10536
```

lucene/core/src/java/org/apache/lucene/index/OrdinalMap.java
```diff
@@ -271,13 +273,26 @@ protected boolean lessThan(TermsEnumIndex a, TermsEnumIndex b) {
...
-    this.globalOrdDeltas = globalOrdDeltas.build();
+    long ramBytesUsed = BASE_RAM_BYTES_USED + segmentMap.ramBytesUsed();
+    this.valueCount = globalOrd;
+
+    // If the first segment contains all of the global ords, then we can apply a small optimization
+    // and hardcode the first segments and global ord deltas as all zeroes.
+    if (ordDeltaBits.length > 0 && ordDeltaBits[0] == 0L && ordDeltas[0].size() == this.valueCount) {
+      this.firstSegments = LongValues.ZEROES;
+      this.globalOrdDeltas = LongValues.ZEROES;
+      ramBytesUsed += RamUsageEstimator.shallowSizeOf(LongValues.ZEROES);
```
jpountz, line 284: We could ignore it completely from `ramBytesUsed`, since this singleton is allocated anyway, regardless of whether the optimization uses it.

lucene/core/src/java/org/apache/lucene/index/OrdinalMap.java
```diff
@@ -359,10 +375,9 @@ public long ramBytesUsed() {
   @Override
   public Collection<Accountable> getChildResources() {
     List<Accountable> resources = new ArrayList<>();
-    resources.add(Accountables.namedAccountable("global ord deltas", globalOrdDeltas));
-    resources.add(Accountables.namedAccountable("first segments", firstSegments));
     resources.add(Accountables.namedAccountable("segment map", segmentMap));
-    // TODO: would be nice to return actual child segment deltas too, but the optimizations are confusing
+    // TODO: would be nice to return the ordinal and segment maps too, but it's not straightforward
+    //  because of optimizations.
```
jpountz, line 380: could be do something like `if (firstSegments != LongValues.ZEROES) { resources.add(Accountables.namedAccountable("first segments", firstSegments)); }`?

jpountz (commented):
Thanks Julie, this looks good to me, I only left some minor comments. Do you know if we already have tests that exercise this optimization?

lucene/core/src/java/org/apache/lucene/index/OrdinalMap.java
```diff
@@ -271,13 +273,26 @@ protected boolean lessThan(TermsEnumIndex a, TermsEnumIndex b) {
...
     }
 
-    this.firstSegments = firstSegments.build();
-    this.globalOrdDeltas = globalOrdDeltas.build();
+    long ramBytesUsed = BASE_RAM_BYTES_USED + segmentMap.ramBytesUsed();
+    this.valueCount = globalOrd;
+
+    // If the first segment contains all of the global ords, then we can apply a small optimization
+    // and hardcode the first segments and global ord deltas as all zeroes.
+    if (ordDeltaBits.length > 0 && ordDeltaBits[0] == 0L && ordDeltas[0].size() == this.valueCount) {
```
mikemccand, line 281: Hmm why only the first segment?  Couldn't it be the 3rd segment, in addition, that matches the global ords?

Edit: ahh OK I understand now -- this opto is indeed specific to the first segment, so we can store `this.firstSegments` as all 0s.  Good!

Do we (somewhere, couldn't find it here) pre-sort all segments by the cardinality descending?  Then we could know all segments that meet this optimization are at the start of the segments list, and possibly building the ordinal map is faster (not sure).  But then we would need to un-sort in the end to return the final `OrdinalMap`.  But it might enable this opto to apply more often, except, I think we would then need an additional dereference on lookup, hmm.

Does our `PackedLongValues.monotonicBuilder` already optimize for the case where it is all 0s, for the case where another segment (not the first) has all the global values as well?

lucene/core/src/java/org/apache/lucene/index/OrdinalMap.java
```diff
@@ -271,13 +273,26 @@ protected boolean lessThan(TermsEnumIndex a, TermsEnumIndex b) {
       globalOrd++;
     }
 
-    this.firstSegments = firstSegments.build();
-    this.globalOrdDeltas = globalOrdDeltas.build();
+    long ramBytesUsed = BASE_RAM_BYTES_USED + segmentMap.ramBytesUsed();
+    this.valueCount = globalOrd;
+
+    // If the first segment contains all of the global ords, then we can apply a small optimization
+    // and hardcode the first segments and global ord deltas as all zeroes.
```
mikemccand, line 280: Insert possessive quote (`first segment's`)?

jtibshirani: I don't think the possessive quote gives the right meaning? Perhaps I could say 'first segment indices' here to be more clear.

lucene/core/src/java/org/apache/lucene/index/OrdinalMap.java
jtibshirani: > Do we (somewhere, couldn't find it here) pre-sort all segments by the cardinality descending?

We do in fact -- the segments are sorted by 'weight', which in all call sites corresponds to the number of unique terms. This was added in [LUCENE-5782](https://issues.apache.org/jira/browse/LUCENE-5782).

> Does our PackedLongValues.monotonicBuilder already optimize for the case where it is all 0s, for the case where another segment (not the first) has all the global values as well?

It does look like it -- when constructing the individual `PackedInts.Reader` instances, we identify the all 0s case and use the lightweight `PackedInts.NullReader`. It's great we optimize that case, but it does mean this PR doesn't make an enormous space difference.

lucene/core/src/java/org/apache/lucene/index/OrdinalMap.java
jtibshirani: I added this to address a failure in `TestOrdinalMap`. But now I see it makes more sense to modify the test !

lucene/core/src/java/org/apache/lucene/index/OrdinalMap.java
mikemccand: Oh I thought it was `the first segment's deltas as all zeros` and alos the `global ord deltas as all zeros`?  But I'm OK with just rewording it to make it less controversial, or even just leaving this wording!

lucene/core/src/java/org/apache/lucene/index/OrdinalMap.java
mikemccand: > > Do we (somewhere, couldn't find it here) pre-sort all segments by the cardinality descending?
> 
> We do in fact -- the segments are sorted by 'weight', which in all call sites corresponds to the number of unique terms. This was added in [LUCENE-5782](https://issues.apache.org/jira/browse/LUCENE-5782).

Ahh, so then we know the first segment will indeed have the most unique terms, and therefore the highest chance of having "all 0s" ord deltas.

I think 2nd and 3rd segments also might have all 0s ord deltas?  But we can try to optimize that in a followon issue ... progress not perfection!

> > Does our PackedLongValues.monotonicBuilder already optimize for the case where it is all 0s, for the case where another segment (not the first) has all the global values as well?
> 
> When constructing the individual `PackedInts.Reader` instances, we do identify the all 0s case and use the lightweight `PackedInts.NullReader`. It's great we optimize that case, but it does mean this PR doesn't make an enormous space difference.

Got it.  Well, it's great that all these layers optimize :)

jpountz: > I think 2nd and 3rd segments also might have all 0s ord deltas? But we can try to optimize that in a followon issue ... progress not perfection!

I wonder if you are confused here, the proposed changes optimize the mapping from global ordinals to the ordinals of one arbitrary segment. When a segment has all value, we can simplify by always picking this segment, but there is no need to optimize this for the 2nd or 3rd segments, since we only need to be able to translate global ordinals to the ordinal of a single segment. Or maybe I'm the one confused by what you were suggesting. :)

lucene/core/src/java/org/apache/lucene/index/OrdinalMap.java
jtibshirani: I think we'd need a cast here, since `LongValues` doesn't implement `Accountable`. Alternatively, we could consider a bigger change to have `LongValues` implement `Accountable`.

jtibshirani:
> Do you know if we already have tests that exercise this optimization?

I think several randomized tests will hit it, for example I noticed `TestLucene80DocValuesFormat` failing when I had a bug in the logic. I added a quick test case to `TestOrdinalMap` too for more solid coverage.

jtibshirani:
@jpountz @mikemccand no rush, but this is ready for another look.

lucene/core/src/java/org/apache/lucene/index/OrdinalMap.java
mikemccand: Aha!  Sorry, I was indeed confused ;)

This is to enable "retrieve `BytesRef` for this global ordinal" use-case, right?  For that, we first pick a segment to use (the first one also containing that `BytesRef`), then map to its segment-local ordinal, then retrieve the `BytesRef` for that using the existing doc values API for that segment.

We do not (need to, nor) expose an API today to "retrieve segment N's ordinal corresponding to global ordinal M".  Only the reverse direction (segment N's ordinal M maps to global ordinal O).

I think I understand now!

lucene/core/src/test/org/apache/lucene/index/TestOrdinalMap.java
```diff
@@ -32,6 +28,10 @@
 import org.apache.lucene.util.RamUsageTester;
 import org.apache.lucene.util.TestUtil;
 
+import java.io.IOException;
+import java.lang.reflect.Field;
+import java.util.HashMap;
+
```
mikemccand, line 34: Does (would) https://issues.apache.org/jira/browse/LUCENE-9564 enforce import ordering check?

mikemccand (approved):
Thank you @jtibshirani, this opto makes sense to me now and looks great!

lucene/core/src/test/org/apache/lucene/index/TestOrdinalMap.java
jtibshirani: I think it would avoid this sort of change (I use IntelliJ which autoformats  `java` imports at the end).

lucene/core/src/java/org/apache/lucene/index/OrdinalMap.java
jpountz: > This is to enable "retrieve BytesRef for this global ordinal" use-case, right?

Right!

> We do not (need to, nor) expose an API today to "retrieve segment N's ordinal corresponding to global ordinal M"

Correct.

PR #1950, opened 2020-10-06
LUCENE-9564: add spotless and gjf.

dweiss:


dsmiley:
What about declaring the import order?:
https://github.com/diffplug/spotless/tree/main/plugin-gradle#java
google-java-format alone does not normalize the import order, so it [says](https://github.com/google/google-java-format#intellij-android-studio-and-other-jetbrains-ides)

dweiss:
Not handled by the IDE plugin but handled by the formatter. So even if your IDE screws up things, pre-commit tidy pass will fix it.

dweiss:
You planted some doubt in me... but I do believe this is enforced:
https://google.github.io/styleguide/javaguide.html#s3.3.3-import-ordering-and-spacing

Try it. :) Mess it up on the branch and re-run gradlew tidy.  ;)

dsmiley:
I believe you :-).

The configuration I saw suggests in spotless suggests we have control over the import order.  Shouldn't we configure the order as we find to be most consistent in our codebase?

dweiss:
The less of the manual configuration we do, the better I think. If google
orders imports then we should just let it do it.  It hooks into the javac
compiler AST so I think it has all the necessary information not only to
sort the imports but also to resolve them properly (static imports, etc.).

gradle/validation/spotless.gradle
```diff
@@ -0,0 +1,31 @@
...
+      java {
+        licenseHeaderFile file("${resources}/asl-header.txt"), '^(\\s*package)'
+        lineEndings 'UNIX'
+        endWithNewline()
+        googleJavaFormat('1.9')
+
+        // Known problematic files.
+        targetExclude "**/HTMLStripCharFilter.java", "**/UAX29URLEmailTokenizerImpl.java", 
+           "**/PatternParser.java", "**/BuildNavDataFiles.java", "**/CheckLinksAndAnchors.java",
+           "**/TestSubQueryTransformer.java"
```
rmuir, line 18: can we exempt other generated files? Otherwise it will make a lot of regeneration tasks stuff no longer idempotent.

dweiss: I was thinking the other way around, actually always apply formatting after they run so that you get the same code in the output after regenerating these source files. This should be idempotent and have clean formatting too.  

I excluded these particular files because they break the formatter... Some of them have no package and others cause stack overflows - https://github.com/google/google-java-format/issues/528

rmuir: sorry, wrong button click. I still think this might be the wrong tradeoff. while it might make the gradle look simpler, it makes it awful for the one trying to work with the actual regeneration, it is really hard to keep in working correctly.

dweiss: Ok, fair enough. I also suggested elsewhere that we might want to just move those generated files into a different sourceset so that it's clear they are generated and they can be wiped easier upon regeneration. Or is there value in keeping them together, do you think?

rmuir: its worth thinking about yeah. if there was a consolidated list of them maybe it would make other tasks more intuitive: e.g. remove the crazy "Generated" matching that's in RAT and simply exclude them from license analysis too.

dweiss: Ok, I'll see what can be done when I get there.

PR #1951, opened 2020-10-06
SOLR-14691: Reduce object creation by using MapWriter / IteratorWriter.

sigram:
This replaces nearly all Map / List usage in `MetricUtils` with `MapWriter` / `IteratorWriter`. The PR includes also changes to `Utils.MapWriterJSONWriter` to similarly avoid creating Map / List instances and instead serialize `MapWriter` / `IteratorWriter` directly.

### DIFF
--- a/solr/core/src/java/org/apache/solr/handler/ReplicationHandler.java
+++ b/solr/core/src/java/org/apache/solr/handler/ReplicationHandler.java
@@ -1112,10 +1112,7 @@ public class ReplicationHandler extends RequestHandlerBase implements SolrCoreAw
   }
 
   private void addVal(MapWriter.EntryWriter ew, String key, Properties props, @SuppressWarnings({"rawtypes"})Class clzz) {
-    Object val = formatVal(key, props, clzz);
-    if (val != null) {
-      ew.putNoEx(key, val);
-    }
+    ew.putIfNotNullNoEx(key, formatVal(key, props, clzz));
   }
 
   private Object formatVal(String key, Properties props, @SuppressWarnings({"rawtypes"})Class clzz) {
--- a/solr/core/src/java/org/apache/solr/handler/ReplicationHandler.java
+++ b/solr/core/src/java/org/apache/solr/handler/ReplicationHandler.java
@@ -68,6 +68,7 @@ import org.apache.lucene.store.Directory;
 import org.apache.lucene.store.IOContext;
 import org.apache.lucene.store.IndexInput;
 import org.apache.lucene.store.RateLimiter;
+import org.apache.solr.common.MapWriter;
 import org.apache.solr.common.SolrException;
 import org.apache.solr.common.SolrException.ErrorCode;
 import org.apache.solr.common.params.CommonParams;
@@ -894,7 +895,7 @@ public class ReplicationHandler extends RequestHandlerBase implements SolrCoreAw
          true, "isLeader", getCategory().toString(), scope);
     solrMetricsContext.gauge(() -> isFollower,
          true, "isFollower", getCategory().toString(), scope);
-    final MetricsMap fetcherMap = new MetricsMap((detailed, map) -> {
+    final MetricsMap fetcherMap = new MetricsMap(map -> {
       IndexFetcher fetcher = currentIndexFetcher;
       if (fetcher != null) {
         map.put(LEADER_URL, fetcher.getLeaderUrl());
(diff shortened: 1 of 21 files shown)


solr/core/src/java/org/apache/solr/util/stats/MetricUtils.java
```diff
@@ -385,10 +406,10 @@ static double nsToMs(boolean convert, double value) {
   }
 
   // some snapshots represent time in ns, other snapshots represent raw values (eg. chunk size)
-  static void addSnapshot(Map<String, Object> response, Snapshot snapshot, PropertyFilter propertyFilter, boolean ms) {
+  static void addSnapshot(MapWriter.EntryWriter ew, Snapshot snapshot, PropertyFilter propertyFilter, boolean ms) {
     BiConsumer<String, Object> filter = (k, v) -> {
       if (propertyFilter.accept(k)) {
```
noblepaul, line 411: I see this pattern repeated in many places. I think a helper method can be added to avoid code duplication

noblepaul:
Can we get rid of the interface `PropertyFilter` and replace it with a `Predicate<String>`?

You can use the class `ConditionalKeyMapWriter` to filter

`MetricUtils.addMetrics()` : Why can't we just have an anonymous class that implements `MapWriter` and avoid creation of `NamedList`

solr/core/src/test/org/apache/solr/handler/admin/MetricsHandlerTest.java
```diff
@@ -308,19 +318,24 @@ public void testKeyMetrics() throws Exception {
...
+    val = Utils.getObjectByPath(map, true, "metrics/" + key3);
     assertNotNull(val);
 
     String key4 = "solr.core.collection1:QUERY./select.requestTimes:1minRate";
     resp = new SolrQueryResponse();
     handler.handleRequestBody(req(CommonParams.QT, "/admin/metrics", CommonParams.WT, "json",
         MetricsHandler.KEY_PARAM, key4), resp);
     values = resp.getValues();
-    val = values.findRecursive("metrics", key4);
+    map = new HashMap<>();
```
noblepaul, line 335: There is no need to convert this to `Map`.

`values._get()` can fetch nested values

noblepaul:
Is it possible to change  `MetricsMap implements Gauge<Map<String,Object>>` to 

` MetricsMap implements Gauge<MapWriter>`

sigram:
> Is it possible to change MetricsMap implements Gauge<Map<String,Object>>

Not in branch_8x - it's a part of the public API.

solr/core/src/java/org/apache/solr/util/stats/MetricUtils.java
sigram: Which pattern?

sigram:
> Can we get rid of the interface PropertyFilter and replace it with a Predicate<String>?
Again, we're limited by back-compat in 8x. We could add another set of methods in `MetricUtils` that use `Predicate<String>` and deprecate `PropertyFilter` ... but that would double the number of methods.

> MetricUtils.addMetrics()
Not sure what you're referring to - this method doesn't create a NamedList. It's used only in `OverseerStatusCmd`, which creates a NamedList anyway for other stuff.

noblepaul:
>Again, we're limited by back-compat in 8x. We could add another set of methods in MetricUtils that use Predicate<String> and deprecate PropertyFilter ... but that would double the number of methods.

That should be OK. We can double the no:of methods and deprecate the old one

noblepaul:
>Not in branch_8x - it's a part of the public API.

We can have both and mark the old as deprecated. The guages are creating too many Map objects

solr/core/src/java/org/apache/solr/handler/ReplicationHandler.java
```diff
@@ -1110,10 +1111,10 @@ private void addVal(NamedList<Object> nl, String key, Properties props, @Suppres
     }
   }
 
-  private void addVal(Map<String, Object> map, String key, Properties props, @SuppressWarnings({"rawtypes"})Class clzz) {
+  private void addVal(MapWriter.EntryWriter ew, String key, Properties props, @SuppressWarnings({"rawtypes"})Class clzz) {
     Object val = formatVal(key, props, clzz);
     if (val != null) {
-      map.put(key, val);
+      ew.putNoEx(key, val);
```
noblepaul, line 1117: there is a putIfNotNull()

noblepaul:
LGTM

sigram:
Merged - thanks!

PR #1952, opened 2020-10-06
LUCENE-9565 Fix competitive iteration

mayya-sharipova:
PR #1351 introduced a sort optimization where documents can be skipped.
But iteration over competitive iterators was not properly organized,
as they were not storing the current docID, and
when competitive iterator was updated, the current doc ID was lost.

This patch fixes it.

Relates to #1351

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/Weight.java
+++ b/lucene/core/src/java/org/apache/lucene/search/Weight.java
@@ -204,9 +204,14 @@ public abstract class Weight implements SegmentCacheable {
       collector.setScorer(scorer);
       DocIdSetIterator scorerIterator = twoPhase == null ? iterator : twoPhase.approximation();
       DocIdSetIterator collectorIterator = collector.competitiveIterator();
-      // if possible filter scorerIterator to keep only competitive docs as defined by collector
-      DocIdSetIterator filteredIterator = collectorIterator == null ? scorerIterator :
-          ConjunctionDISI.intersectIterators(Arrays.asList(scorerIterator, collectorIterator));
+      DocIdSetIterator filteredIterator = scorerIterator;
+      if (collectorIterator != null) {
+        if (scorerIterator.docID() != -1) {
+          collectorIterator.advance(scorerIterator.docID());
+        }
+        // filter scorerIterator to keep only competitive docs as defined by collector
+        filteredIterator = ConjunctionDISI.intersectIterators(Arrays.asList(scorerIterator, collectorIterator));
+      }
       if (filteredIterator.docID() == -1 && min == 0 && max == DocIdSetIterator.NO_MORE_DOCS) {
         scoreAll(collector, filteredIterator, twoPhase, acceptDocs);
         return DocIdSetIterator.NO_MORE_DOCS;
--- a/lucene/core/src/java/org/apache/lucene/search/Weight.java
+++ b/lucene/core/src/java/org/apache/lucene/search/Weight.java
@@ -210,7 +210,7 @@ public abstract class Weight implements SegmentCacheable {
       } else {
         if (scorerIterator.docID() != -1) {
           // Wrap ScorerIterator to start from -1 for conjunction 
-          scorerIterator = new RangeDISIWrapper(scorerIterator, scorerIterator.docID(), max);
+          scorerIterator = new RangeDISIWrapper(scorerIterator, max);
         }
         // filter scorerIterator to keep only competitive docs as defined by collector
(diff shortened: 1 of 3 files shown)


mayya-sharipova:
This patch works well with checks in  [ConjunctionDISI](https://github.com/apache/lucene-solr/pull/1937), and all tests pass as well.

After we merge this PR, we can reintroduce the checks in ConjunctionDISI.

lucene/core/src/java/org/apache/lucene/search/Weight.java
```diff
@@ -204,9 +204,14 @@ public int score(LeafCollector collector, Bits acceptDocs, int min, int max) thr
...
       DocIdSetIterator scorerIterator = twoPhase == null ? iterator : twoPhase.approximation();
       DocIdSetIterator collectorIterator = collector.competitiveIterator();
-      // if possible filter scorerIterator to keep only competitive docs as defined by collector
-      DocIdSetIterator filteredIterator = collectorIterator == null ? scorerIterator :
-          ConjunctionDISI.intersectIterators(Arrays.asList(scorerIterator, collectorIterator));
+      DocIdSetIterator filteredIterator = scorerIterator;
+      if (collectorIterator != null) {
+        if (scorerIterator.docID() != -1) {
+          collectorIterator.advance(scorerIterator.docID());
+        }
```
jpountz, line 211: I don't think that this is good enough as we might be advancing ahead of scorerIterator? This was why I thought that we should instead wrap scorerIterator in such a way that its initial docID would be -1.

jpountz (changes requested):

lucene/core/src/java/org/apache/lucene/search/Weight.java
mayya-sharipova: @jpountz  Addressed in cba6cf75f48a26cd48f09152c25518c91fd6660d

jpountz: oh, I had not considered setting `scorerIterator.docID()` as a min docID, maybe this means that we no longer need the `min` parameter of `RangeDISIWrapper`?

jpountz (approved):

lucene/core/src/java/org/apache/lucene/search/Weight.java
mayya-sharipova: Thanks @jpountz , addressed in d42c4649c81364f13c51c0b147dd600e57d7cccc

lucene/core/src/java/org/apache/lucene/search/Weight.java
```diff
@@ -266,4 +274,45 @@ static void scoreAll(LeafCollector collector, DocIdSetIterator iterator, TwoPhas
...
+      return advance(docID + 1);
+    }
+
+    @Override
+    public int advance(int target) throws IOException {
+      target = Math.max(min, target);
+      if (target >= max) {
+        return docID = NO_MORE_DOCS;
+      }
+      return docID = in.advance(target);
```
jpountz, line 308: it just occurred to me that this implementation is not correct in the case that the minimum bound of the range of doc IDs to score is less than the current doc ID of the scorer, have you seen any failures with your change? I wonder that we would need to do
```
if (target >= scorer.docID()) { return scorer.docID(); }
```
but we should create a test that fails without this

mayya-sharipova: Indeed the [recent failures](https://elasticsearch-ci.elastic.co/job/apache+lucene-solr+master/5689/console) on Lucene are because of this. `ReqExclBulkScorer` can `score` method several times and sometimes with the minimum bound that is <= scorer.docID().



I am thinking the logic is becoming more difficult. I am thinking to go back to the [initial commit](https://github.com/apache/lucene-solr/pull/1952/commits/574245c96db19291138b2eb7253451b567c69d89) and advance collectorIterator to the same doc as `scorerIterator`. At the beginning a  `collectorIterator` matches all docs, so it should precisely advance to the `scorerIterator.docID()).

mayya-sharipova: @jpountz  I've created a new PR: https://github.com/apache/lucene-solr/pull/1955.  So sorry for the mess.

mayya-sharipova: @jpountz  
> By this `if (target >= scorer.docID()) { return scorer.docID(); }`

Did you mean if (target <= scorer.docID()) { return scorer.docID(); } ?

PR #1955, opened 2020-10-06
LUCENE-9566 TestApproximationSearchEquivalence.testExclusion fix

mayya-sharipova:


lucene/core/src/java/org/apache/lucene/search/Weight.java
```diff
@@ -204,21 +204,21 @@ public int score(LeafCollector collector, Bits acceptDocs, int min, int max) thr
...
       if (filteredIterator.docID() == -1 && min == 0 && max == DocIdSetIterator.NO_MORE_DOCS) {
+        if (collectorIterator != null) {
+          filteredIterator = ConjunctionDISI.intersectIterators(Arrays.asList(scorerIterator, collectorIterator));
+        }
         scoreAll(collector, filteredIterator, twoPhase, acceptDocs);
         return DocIdSetIterator.NO_MORE_DOCS;
       } else {
+        if (collectorIterator != null) {
+          if (scorerIterator.docID() != -1) {
+            // here we assume that the initial version of collectorIterator matches all docs and can advance exactly to the desired docID
```
jpountz, line 217: but that's not true, it it?

mayya-sharipova: Thanks for the comment, I will think about this more

mayya-sharipova: @jpountz  My last [commit](4c5d8284941d315b7f15d819fbbc5d156421f02d) seems to address the test failures.  I think it may work. Sorry again for the trouble.

jpountz (approved):
LGTM

PR #1956, opened 2020-10-07
LUCENE-9563: Add .editorconfig

dsmiley:
https://issues.apache.org/jira/browse/LUCENE-9563


rmuir:
if we want to use https://issues.apache.org/jira/browse/LUCENE-9564 to auto-format code, why add David Smiley's personal code style and not the one supported by the formatter: https://github.com/google/google-java-format#intellij-android-studio-and-other-jetbrains-ides ?

dsmiley:
IntelliJ users (like myself) should indeed use that plug-in -- I installed it some weeks ago.  But users like @msokolov use Emacs, and there are lots of other simple editors.  ".editorconfig" is more standardized.  The particular editorconfig in the PR here is not suitable any more.  @msokolov would you care to submit a new PR with the "right" one, whatever that is?

PR #1961, opened 2020-10-07
LUCENE-9567: JPOSSFF loads built-in stop tags by default

msfroh:
If a user does not specify the "tags" argument for
JapanesePartOfSpeechStopFilterFactory, it is unlikely that their
intention was to omit the filter. (Why would they use
JapanesePartOfSpeechStopFilterFactory at all in that case?)

This change loads the stoptags.txt that ships in the Kuromoji
analysis jar if no tags argument is specified.

# Description

Loading default stop tags if "tags" argument is not specified in JapanesePartOfSpeechStopFilterFactory.

# Solution

If the "tags" argument is not specified. then we load the default stop tags (as defined by `JapaneseAnalyzer.getDefaultStopTags()`) in the constructor (rather than waiting for the call to `inform`).

This implementation was chosen to be consistent with KoreanPartOfSpeechStopFilterFactory.

I added a note to MIGRATE.md, in case anyone is currently using JapanesePartOfSpeechStopFilterFactory with no arguments, to let them know that it will begin having an effect due to this change.

# Tests

Added a unit test to verify the behavior without specifying the "tags" argument.

# Checklist

### DIFF
--- a/lucene/MIGRATE.md
+++ b/lucene/MIGRATE.md
@@ -1,5 +1,11 @@
 # Apache Lucene Migration Guide
 
+## JapanesePartOfSpeechStopFilterFactory loads default stop tags if "tags" argument not specified (LUCENE-9567)
+
+Previously, JapanesePartOfSpeechStopFilterFactory added no filter if `args` didn't include "tags". Now, it will load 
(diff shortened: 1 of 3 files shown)


msokolov:
looks good - I will merge soon if nobody objects

msokolov (approved):

rmuir:
+1

PR #1962, opened 2020-10-08
SOLR-14749 Provide a clean API for cluster-level event processing

sigram:
This is just the API part of the ticket, separated from PR-1758 to facilitate reviews of pure API.


sigram:
@noblepaul I added back the default implementation of `ClusterEventProducer`, and an example implementation of `CollectionsRepairEventListener` to illustrate how to re-implement the `autoAddReplicas` functionality. Please review.

(BTW, ignore the messed up merge history here... I made a mistake when merging, but the final diff with master is now correct).

solr/core/src/java/org/apache/solr/core/CoreContainer.java
```diff
@@ -252,6 +253,11 @@ public CoreLoadFailure(CoreDescriptor cd, Exception loadFailure) {
           getZkController().getOverseer() != null &&
           !getZkController().getOverseer().isClosed(),
       (r) -> this.runAsync(r));
+
+  private final ClusterEventProducerFactory clusterEventProducerFactory = new ClusterEventProducerFactory(this);
```
noblepaul, line 257: Does it have to be a field? It can be a local variable as well

solr/core/src/java/org/apache/solr/core/ClusterEventProducerFactory.java
```diff
@@ -0,0 +1,178 @@
...
+      throw new RuntimeException("this factory can be called only once!");
+    }
+    final ClusterEventProducer clusterEventProducer;
+    CustomContainerPlugins.ApiInfo clusterEventProducerInfo = plugins.getPlugin(ClusterEventProducer.PLUGIN_NAME);
+    if (clusterEventProducerInfo != null) {
+      // the listener in ClusterSingletons already registered it
+      clusterEventProducer = (ClusterEventProducer) clusterEventProducerInfo.getInstance();
+    } else {
+      // create the default impl
+      clusterEventProducer = new DefaultClusterEventProducer(cc);
```
noblepaul, line 87: There is no need to have a `ClusterEventProducer` if the user does not register it.

noblepaul (changes requested):

solr/core/src/java/org/apache/solr/core/CoreContainer.java
sigram: +1.

solr/core/src/java/org/apache/solr/core/ClusterEventProducerFactory.java
sigram: I'm not sure. If we treat this as a part of core functionality (the ability to produce and listen to cluster-level events) then we need a rudimentary default implementation to be always present.

noblepaul: This is a core , opt-in functionality

core = part of core
opt-in =. no traces if user does not explicitly add an event producer

sigram: AFAIK there is no mechanism in plugins to require the presence of another plugin. If we decide to make this functionality an opt-in then other plugins that depend on this may silently fail to work properly because the user forgot to specify this opt-in. This problem is not specific to this plugin, but rather it's a lack of functionality in our whole plugin ecosystem.

Until we have a working mechanism to specify plugin requirements / dependencies I think we have to provide sensible defaults, and a sensible default is a working implementation, plus the ability to opt-out or change the default impl. I'll add a mechanism to opt-out.

noblepaul: > AFAIK there is no mechanism in plugins to require the presence of another plugin.

There is no reason why we can't implement it. I think we should. Everything should be initialized as lazily as possible.

The add command for a `ClusterEventListener` has to do an extra check to see if a `ClusterEventProducer` is registered. If not, just register it

sigram: > There is no reason why we can't implement it. I think we should.

Sure, but it's not a part of this PR. If/when we add it then we can drop this initialization of the default impl.

noblepaul: NO , please. 

I am -1 on enabling this by default

noblepaul (approved):
LGTM

Please document examples

solr/core/src/java/org/apache/solr/cluster/events/impl/ClusterEventProducerFactory.java
```diff
@@ -0,0 +1,173 @@
...
+ */
+public class ClusterEventProducerFactory extends ClusterEventProducerBase {
+  private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+
+  private ContainerPluginsRegistry.PluginRegistryListener initialPluginListener;
+  private boolean created = false;
+
+  public ClusterEventProducerFactory(CoreContainer cc) {
+    super(cc);
+    initialPluginListener = new ContainerPluginsRegistry.PluginRegistryListener() {
```
murblanc, line 29: Wouldn't code be easier to read/maintain if named classes were defined rather than anonymous ones? They would have names that help understand their function, class javadoc etc... I feel the anonymous classes in this PR have enough content to justify a non anonymous implementation. That's likely just a personal preference of mine so feel free to ignore.

solr/core/src/java/org/apache/solr/cluster/events/impl/ClusterEventProducerFactory.java
```diff
@@ -0,0 +1,173 @@
...
+        added(replacement);
+        deleted(old);
+      }
+    };
+    plugins.registerListener(pluginListener);
+    created = true;
+    return clusterEventProducer;
+  }
+
+  private void transferListeners(ClusterEventProducer target, ContainerPluginsRegistry plugins) {
```
murblanc, line 154: Would there be a way to register listeners into an intermediate object (not tied to any specific producer), then have event producers feed events into this object (conceptually it would behave like a queue, but the implementation can be direct listener method calls). This would save having methods such as this one (and would not require having a producer in order to register listeners, if that helps...).

solr/core/src/java/org/apache/solr/cluster/events/impl/ClusterEventProducerFactory.java
```diff
@@ -0,0 +1,173 @@
...
+    cc.getClusterSingletons().getSingletons().put(ClusterEventProducer.PLUGIN_NAME +"_delegate", clusterEventProducer);
+    ContainerPluginsRegistry.ApiInfo clusterEventProducerInfo = plugins.getPlugin(ClusterEventProducer.PLUGIN_NAME);
+    if (clusterEventProducerInfo != null) {
+      // the listener in ClusterSingletons already registered it
+      clusterEventProducer.setDelegate((ClusterEventProducer) clusterEventProducerInfo.getInstance());
+    } else {
+      // use the default NoOp impl
+    }
+    // transfer those listeners that were already registered to the initial impl
+    transferListeners(clusterEventProducer, plugins);
```
murblanc, line 98: Any potential for race conditions here? Can events be produced while this code executes?

solr/core/src/java/org/apache/solr/cluster/events/impl/ClusterEventProducerFactory.java
```diff
@@ -0,0 +1,173 @@
...
+
+import java.lang.invoke.MethodHandles;
+import java.util.Set;
+
+/**
+ * This class helps in handling the initial registration of plugin-based listeners,
+ * when both the final {@link ClusterEventProducer} implementation and listeners
+ * are configured using plugins.
+ */
+public class ClusterEventProducerFactory extends ClusterEventProducerBase {
```
murblanc, line 21: This class is only used by a single thread, right?

solr/core/src/java/org/apache/solr/cluster/events/impl/ClusterEventProducerFactory.java
```diff
@@ -0,0 +1,173 @@
...
+  }
+
+  private void transferListeners(ClusterEventProducer target, ContainerPluginsRegistry plugins) {
+    // stop capturing listener plugins
+    plugins.unregisterListener(initialPluginListener);
+    // transfer listeners that are already registered
+    listeners.forEach((type, listenersSet) -> {
+      listenersSet.forEach(listener -> target.registerListener(listener, type));
+    });
+    listeners.clear();
```
murblanc, line 161: Can't this be causing issues with methods iterating over `listeners` in the base class? Maybe need to synchronize `transferListeners`?

solr/core/src/java/org/apache/solr/cluster/events/ClusterEventProducerBase.java
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  public CoreContainer getCoreContainer() {
+    return cc;
+  }
+
+  public abstract Set<ClusterEvent.EventType> getSupportedEventTypes();
+
+  protected void fireEvent(ClusterEvent event) {
+    listeners.getOrDefault(event.getType(), Collections.emptySet())
```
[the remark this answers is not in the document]
sigram: Good point, muse-dev :) Fixing.

solr/core/src/java/org/apache/solr/cluster/events/impl/ClusterEventProducerFactory.java
sigram: Yes.

solr/core/src/java/org/apache/solr/cluster/events/impl/ClusterEventProducerFactory.java
sigram: No. At this point in the initialization of CoreContainer any source of events will be ignored anyway (because the factory is basically a no-op implementation that doesn't generate any events), and listeners are being transferred before the final implementation is installed.

sigram:
Merged to master.

PR #1964, opened 2020-10-08
SOLR-14749: Cluster singleton part of PR-1785

sigram:
This is just the `ClusterSingleton` part, with an example implementation that loads the ClusterSingleton plugins via `CustomerContainerPlugins` API. See `TestContainerPlugin` to see how it works.


solr/core/src/java/org/apache/solr/handler/admin/ContainerPluginsApi.java
```diff
@@ -51,7 +51,7 @@
 public class ContainerPluginsApi {
   private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
 
-  public static final String PLUGIN = "plugin";
+  public static final String PLUGINS = "plugin";
```
noblepaul, line 54: why change the variable name at all?

solr/core/src/java/org/apache/solr/core/CoreContainer.java
```diff
@@ -168,6 +172,33 @@ public CoreLoadFailure(CoreDescriptor cd, Exception loadFailure) {
     }
   }
 
+  public static class ClusterSingletons {
```
noblepaul, line 175: There is no need for `CoreContainer` to be aware of `ClusterSingleton`

Only `ContainerPlugins` and `Overseer` need to be aware of it

noblepaul (changes requested):

solr/core/src/java/org/apache/solr/handler/admin/ContainerPluginsApi.java
sigram: it's a left-over from the earlier change that reverted but I missed this change.

Still, I think we should change the name of the path to "plugins" because the singular form doesn't make sense. Back-compat can be preserved by initially accepting both names.

solr/core/src/java/org/apache/solr/core/CoreContainer.java
sigram: That's the downside of splitting this PR into API & implementation ... The way we load CC it's nearly impossible to properly initialize singletons from within the load() method if we don't defer the initialization until the load() is completed. `ClusterSingletons` class is just a helper for deferred init of singletons once they are all discovered and loaded.

solr/core/src/java/org/apache/solr/api/AnnotatedApi.java
```diff
@@ -85,9 +85,18 @@ public EndPoint getEndPoint() {
...
+
+  /**
+   * Get a list of Api-s supported by this class.
+   * @param theClass class
+   * @param obj object of this class (may be null)
+   * @param required if true then an exception is thrown if no Api-s can be retrieved, if false
+   *                then absence of Api-s is silently ignored.
+   * @return list of discovered Api-s
+   */
+  public static List<Api> getApis(Class<? extends Object> theClass , Object obj, boolean required)  {
```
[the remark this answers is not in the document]
sigram: Do you mean the name is confusing? Either name is fine with me.

solr/core/src/java/org/apache/solr/core/CoreContainer.java
noblepaul: >That's the downside of splitting this PR into API & implementation ..

I thouh this Pr has both API and impl

`CC.load()` is a kitchen sink today. We should not need such large changes to CC everytime something is added

solr/core/src/java/org/apache/solr/handler/admin/ContainerPluginsApi.java
noblepaul: If you plan to change it, it should be a part of a different PR. We can remove that change from here

noblepaul:
I've added a listener functionality to `CustomContainerPlugins` . let's remove that `handleClusterSingletons()` from that class

solr/core/src/java/org/apache/solr/handler/admin/ContainerPluginsApi.java
noblepaul: I mean why these changes should be present in this PR?

solr/core/src/java/org/apache/solr/core/CoreContainer.java
sigram: Moved most of that into a utility class.

solr/core/src/java/org/apache/solr/handler/admin/ContainerPluginsApi.java
sigram: Removed.

noblepaul:
LGTM

noblepaul (approved):

sigram:
Merged.

PR #1965, opened 2020-10-08
LUCENE-9572 - TypeAsSynonymFilter gains selective flag transfer and an ignore list.

gus-asf:


gus-asf:
@sarowe do you have any comments?

sarowe:
> @sarowe do you have any comments?

I'll take a look later today.

lucene/analysis/common/src/java/org/apache/lucene/analysis/miscellaneous/TypeAsSynonymFilter.java
```diff
@@ -64,9 +82,15 @@ public boolean incrementToken() throws IOException {
       }
       termAtt.append(typeAtt.type());
       posIncrAtt.setPositionIncrement(0);
+      // control what flags transfer to synonym
+      int flags = getAttribute(FlagsAttribute.class).getFlags();
```
uschindler, line 86: This should not call getAttribute and instead use the instance `flagsAtt` above (instance field). In the next line it's correct.

lucene/CHANGES.txt
```diff
@@ -15,7 +15,7 @@ API Changes
 * LUCENE-8474: RAMDirectory and associated deprecated classes have been
   removed. (Dawid Weiss)
 
-* LUCENE-3041: The deprecated Weight#extractTerms() method has been 
+* LUCENE-3041: The deprecated Weight#extractTerms() method has been
```
uschindler, line 18: same issue like in previous PR!

uschindler (changes requested):
This looks fine to me except the "getAttribute". It should only use flagsAtt.
I don't fully understand why this flags handling is needed in the tokenfilter, because i have not yet reviewed the new Solr feature.
At least I cannot find any problematic parts. To me it's mostly important if the tokenfilter works the same like before if no exclusion set and no flags are passed, which seems guaranteed by `~0` (cool notation for all bits set)!

lucene/test-framework/src/java/org/apache/lucene/analysis/BaseTokenStreamTestCase.java
```diff
@@ -190,7 +196,8 @@ public static void assertTokenStreamContents(TokenStream ts, String[] output, in
       if (posLengthAtt != null) posLengthAtt.setPositionLength(45987653);
       if (keywordAtt != null) keywordAtt.setKeyword((i&1) == 0);
       if (payloadAtt != null) payloadAtt.setPayload(new BytesRef(new byte[] { 0x00, -0x21, 0x12, -0x43, 0x24 }));
-      
+      if (flagsAtt != null) flagsAtt.setFlags(Integer.MAX_VALUE); // all 1's
```
uschindler, line 199: that's all bits except the first 1 (`01111111111111111...`). To invert all bits use `-1` or better `~0`

lucene/test-framework/src/java/org/apache/lucene/analysis/BaseTokenStreamTestCase.java
```diff
@@ -294,6 +304,7 @@ public static void assertTokenStreamContents(TokenStream ts, String[] output, in
     if (posLengthAtt != null) posLengthAtt.setPositionLength(45987653);
     if (keywordAtt != null) keywordAtt.setKeyword(true);
     if (payloadAtt != null) payloadAtt.setPayload(new BytesRef(new byte[] { 0x00, -0x21, 0x12, -0x43, 0x24 }));
+    if (flagsAtt != null) flagsAtt.setFlags(Integer.MAX_VALUE); // all 1's
```
uschindler, line 307: same here

lucene/test-framework/src/java/org/apache/lucene/analysis/BaseTokenStreamTestCase.java
```diff
@@ -125,7 +125,7 @@ public void reflectWith(AttributeReflector reflector) {
   //     arriving to pos Y have the same endOffset)
   public static void assertTokenStreamContents(TokenStream ts, String[] output, int startOffsets[], int endOffsets[], String types[], int posIncrements[],
```
uschindler, line 126: maybe we should not change this method signature and just add another one, so we do not need to change unrelated files like MinHashFilter?

uschindler (changes requested):

lucene/CHANGES.txt
gus-asf: Well actually this PR predates other :), but yes same issue.

lucene/test-framework/src/java/org/apache/lucene/analysis/BaseTokenStreamTestCase.java
gus-asf: oops yeah realized that partway through (when I switched to ~0 for the other file), didn't come back and fix it here.

lucene/test-framework/src/java/org/apache/lucene/analysis/BaseTokenStreamTestCase.java
gus-asf: Done, I think what was originally submitted was a bit easier than sorting out all the parameters at the time, but yes nicer to minimize.

gus-asf:
> I don't fully understand why this flags handling is needed in the tokenfilter, because i have not yet reviewed the new Solr feature.

LUCENE-9575 (patch waiting for the test changes here) will show you where the flags are gained. The basic idea is that the flags are used to identify the source tokens that gave rise to synonym tokens and (sometimes) remove the source tokens *after* they participate in further (arbitrary) processing... the canonical example is to allow '401k' and '401(k)' to match the same documents about retirement plans and also NOT match documents regarding "401 Unauthorized" or "Vitamin K"

uschindler (approved):
Looks fine to me. I am waiting to see an example how this is used, maybe with the Solr example you bring later.

lucene/analysis/common/src/test/org/apache/lucene/analysis/miscellaneous/TestTypeAsSynonymFilter.java
```diff
@@ -0,0 +1,99 @@
...
+  public void testWithPrefixAndIgnore() throws Exception {
+    Token[] tokens = new Token[] {
+        new Token("foo", 1, 3),
+        new Token("foo", 5, 7),
+        new Token("foo", 9, 11),
+    } ;
+    tokens[0].setType("bar");
+    tokens[2].setType("ignoreme");
+    TokenStream ts = new CannedTokenStream(tokens);
+    ts = new TypeAsSynonymFilter(ts,"pfx_", Stream.of("word","ignoreme").collect(Collectors.toSet()), 0);
```
uschindler, line 68: Why not use `Set.of()` (since Java 11)?

uschindler: `Stream.of("word","ignoreme").collect(Collectors.toSet())` => `Set.of("word","ignoreme")`

uschindler:
I just recommend a small change in the test to use Set.of(), hwich is new in Java 11 (I think already in 9) for static sets.

lucene/analysis/common/src/test/org/apache/lucene/analysis/miscellaneous/TestTypeAsSynonymFilter.java
gus-asf: Ah yeah sure. Originally developed vs 8.4 but now I can use it :)

uschindler: In 8.x I'd use: `new HashSet(Arrays.asList("word","ignoreme"))`, but that's a matter of preference. I tend to not use streams if there's no transformations/mappings going on.

gus-asf:
Puzzling why RulesTest in solr is suddenly failing rat. It sure looks like it has a license just like all others and I didn't touch it.

gus-asf:
Well that was misleading. Rat is happy locally, not sure why github was blaming RulesTest's rat analysis. I did find a couple of unused imports, and realized that apparently `gradlew :lucene:check` doesn't run ecjLint as I had expected. Just for fun javadocs were running out of memory locally when I tried a full top level precommit which gives this lovely cryptic message: 
```
* What went wrong:
Execution failed for task ':lucene:demo:renderJavadoc'.
> java.io.IOException: Stream Closed
```
(had to edit line 543 of render-javadoc.gradle to get it to work)

lucene/analysis/common/src/java/org/apache/lucene/analysis/miscellaneous/TypeAsSynonymFilter.java
```diff
@@ -34,23 +36,39 @@
   private final CharTermAttribute termAtt = addAttribute(CharTermAttribute.class);
```
sarowe, line 36: Consider adding the new features to the class comment

lucene/analysis/common/src/java/org/apache/lucene/analysis/miscellaneous/TypeAsSynonymFilterFactory.java
```diff
@@ -18,12 +18,15 @@
...
 import java.util.Map;
+import java.util.Set;
 
 import org.apache.lucene.analysis.TokenStream;
 import org.apache.lucene.analysis.TokenFilterFactory;
 
 /**
  * Factory for {@link TypeAsSynonymFilter}.
+ *
+ * <p>In Solr this might be used as such
```
sarowe, line 29: Consider adding configuration for the new features to the example Solr configuration.

lucene/analysis/common/src/java/org/apache/lucene/analysis/miscellaneous/TypeAsSynonymFilterFactory.java
```diff
@@ -46,10 +49,17 @@
...
+  private final int synFlagMask;
 
   public TypeAsSynonymFilterFactory(Map<String,String> args) {
     super(args);
     prefix = get(args, "prefix");  // default value is null
+    String ignoreList = get(args, "ignore");
+    synFlagMask = getInt(args,"synFlagsMask", ~0);
+    if (ignoreList != null) {
+      ignore = Set.of(ignoreList.split(","));
+    }
```
sarowe, line 62: There is a method `getSet()` which you could use instead of `get()` for the ignoreList; it does the splitting for you, and also supports mixed space/comma separators.

lucene/analysis/common/src/test/org/apache/lucene/analysis/miscellaneous/TestTypeAsSynonymFilter.java
```diff
@@ -0,0 +1,98 @@
...
+import org.apache.lucene.analysis.BaseTokenStreamTestCase;
+import org.apache.lucene.analysis.CannedTokenStream;
+import org.apache.lucene.analysis.Token;
+import org.apache.lucene.analysis.TokenStream;
+
+/**
+ * Test that this filter moves the value in type to a synonym token with the same offsets. This is rarely
+ * useful by itself, but in combination with another filter that updates the type value with an appropriate
+ * synonym can be used to identify synonyms before tokens are modified by further analysis, and then
+ * add them at the end, ensuring that the synonym value has not ben subjected to the intervening analysis.
```
sarowe, line 31: typo: ben

lucene/test-framework/src/java/org/apache/lucene/analysis/BaseTokenStreamTestCase.java
```diff
@@ -390,19 +406,19 @@ public static void assertAnalyzesTo(Analyzer a, String input, String[] output, i
   }
 
   public static void assertAnalyzesTo(Analyzer a, String input, String[] output, int startOffsets[], int endOffsets[], String types[], int posIncrements[], int posLengths[], boolean graphOffsetsAreCorrect, byte[][] payloads) throws IOException {
-    assertTokenStreamContents(a.tokenStream("dummy", input), output, startOffsets, endOffsets, types, posIncrements, posLengths, input.length(), null, null, graphOffsetsAreCorrect, payloads);
+    assertTokenStreamContents(a.tokenStream("dummy", input), output, startOffsets, endOffsets, types, posIncrements, posLengths, input.length(), null, null, graphOffsetsAreCorrect, payloads, null);
```
sarowe, line 409: Since you put back the original method, I think you can also revert this change?

sarowe (changes requested):

lucene/analysis/common/src/java/org/apache/lucene/analysis/miscellaneous/TypeAsSynonymFilterFactory.java
gus-asf: Ah cool.

lucene/test-framework/src/java/org/apache/lucene/analysis/BaseTokenStreamTestCase.java
gus-asf: ok

gus-asf:
In re-reading things, I think it just dawned on me that I'm not quite maintaining back compatibility after all. Previously no flags at all were transferred, yet my ~0 default now copies all flags. ( x & ~0 = x) but the prior functionality was no flags copied which would be ( x & 0 = 0) (plus we could test for 0 and just not set flags at all)

gus-asf:
Oh darn, out thought myself looking at the diff where I added a setFlags.... forgot that the stored state is copying flags already never-mind.

gus-asf:
@sarowe I think I've addressed your concerns LMK if you have further thoughts

sarowe:
> @sarowe I think I've addressed your concerns LMK if you have further thoughts

Yes you have, with one exception: I left a comment that AFAICT you haven't addressed:

> .../TypeAsSynonymFilter.java
> Consider adding the new features to the class comment

gus-asf:
Oh I added it to the Factory class which already has a solr xml example (https://github.com/apache/lucene-solr/pull/1965/files#diff-6efa63ada0f2520335d809359639f4c280a4ffbed3cac47927f2e47009f5f7f7R41)

sarowe:
> Oh I added it to the Factory class which already has a solr xml example (https://github.com/apache/lucene-solr/pull/1965/files#diff-6efa63ada0f2520335d809359639f4c280a4ffbed3cac47927f2e47009f5f7f7R41)

Yeah I saw that, but the Filter is an entry point independent of the Factory, and its class comment contains a brief description of the filter: https://github.com/apache/lucene-solr/blob/master/lucene/analysis/common/src/java/org/apache/lucene/analysis/miscellaneous/TypeAsSynonymFilter.java#L29-L32
.

I was thinking another phrase or two could be added mentioning the ignore list and the flag masking capabilities, but if you disagree, that's fine.

gus-asf:
I don't disagree exactly, just seemed like that was where the doc was. Separately it might be good to have a policy that the doc is either on the factory or the filter and @ see to the other to avoid duplication. Happy to add a tidbit on the filter for now.

sarowe (approved):

PR #1966, opened 2020-10-08
LUCENE-9574 Add DropIfFlaggedFilterFactory

gus-asf:

### DIFF
--- a/lucene/analysis/common/src/java/org/apache/lucene/analysis/miscellaneous/DropIfFlaggedFilter.java
+++ b/lucene/analysis/common/src/java/org/apache/lucene/analysis/miscellaneous/DropIfFlaggedFilter.java
@@ -1,0 +1,90 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.lucene.analysis.miscellaneous;
+
+import java.io.IOException;
+import java.util.UUID;
+
+import org.apache.lucene.analysis.TokenFilter;
+import org.apache.lucene.analysis.TokenStream;
+import org.apache.lucene.analysis.tokenattributes.CharTermAttribute;
+import org.apache.lucene.analysis.tokenattributes.FlagsAttribute;
+
+/**
+ * Allows Tokens with a given combination of flags to be dropped.
+ *
+ * @see DropIfFlaggedFilterFactory
+ */
+public class DropIfFlaggedFilter extends TokenFilter {
+
+  private int dropFlags;
+
+  private CharTermAttribute attribute = getAttribute(CharTermAttribute.class);
+  private boolean firstToken = true;
+
+  /**
+   * Construct a token stream filtering the given input.
+   *
+   * @param input the source stream
+   * @param dropFlags a combination of flags that indicates that the token should be dropped.
+   */
+  @SuppressWarnings("WeakerAccess")
+  protected DropIfFlaggedFilter(TokenStream input, int dropFlags) {
+    super(input);
+    this.dropFlags = dropFlags;
+  }
+
+  @Override
+  public final boolean incrementToken() throws IOException {
+    boolean result;
+    boolean dropToken;
+    do {
+      result = input.incrementToken();
+      dropToken = (getAttribute(FlagsAttribute.class).getFlags() & dropFlags) == dropFlags;
+      if (firstToken && dropToken) {
+        // we've attempted to drop the very first token. If the token had a synonym, this can cause:
+        // java.lang.IllegalArgumentException: first position increment must be > 0 (got 0) for field '_text_'
+        // Since there is no way to look ahead or behind in the stream, all we can do is convert to something
+        // that will never match any other token. Thus we make it "un-findable" and therefore effectively
+        // deleted from a user's perspective (ignoring an extremely low probability of duplication by chance)
+        attribute.setEmpty();
+        attribute.append(getRandomToken());
+        dropToken = false;
+      }
+      firstToken = false;
+    } while (result && dropToken);
+    return result;
+  }
+
+  String getRandomToken() {
+    // The following tactic is a hazard for threaded environments because of locking in SecureRandom, but
+    // this case (where the very first token is dropped) should hopefully be an uncommon case for most users
+    // and still result in low thread contention. An indexing or query situation with frequent drops of the first
+    // token could suffer. Open to suggestions for a better method, possibly something involving ThreadLocalRandom
+    // but some serious thought needs to go into how possible duplication is with such a change. Also of concern are
+    // the cases where java uses a RNG that can run out of entropy and block... Erring on the side of correct
+    // functionality over performance for now.
+    return String.valueOf(UUID.randomUUID());
+  }
+
+  @Override
+  public void reset() throws IOException {
+    firstToken = true;
+    super.reset();
+  }
+}
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -15,7 +15,7 @@ API Changes
 * LUCENE-8474: RAMDirectory and associated deprecated classes have been
   removed. (Dawid Weiss)
 
-* LUCENE-3041: The deprecated Weight#extractTerms() method has been 
+* LUCENE-3041: The deprecated Weight#extractTerms() method has been
   removed (Alan Woodward, Simon Willnauer, David Smiley, Luca Cavanna)
 
 * LUCENE-8805: StoredFieldVisitor#stringField now takes a String rather than a
@@ -92,7 +92,7 @@ Improvements
 
 * LUCENE-8937: Avoid agressive stemming on numbers in the FrenchMinimalStemmer.
   (Adrien Gallou via Tomoko Uchida)
-  
+
 * LUCENE-8984: MoreLikeThis MLT is biased for uncommon fields (Andy Hind via Anshum Gupta)
 
 * LUCENE-8596: Kuromoji user dictionary now accepts entries containing hash mark (#) that were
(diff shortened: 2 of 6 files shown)


rmuir:
please use https://github.com/apache/lucene-solr/blob/master/lucene/core/src/java/org/apache/lucene/analysis/FilteringTokenFilter.java as a subclass and just implement accept() logic to determine whether a token should survive. It handles all the hairy parts, no securerandom needed

uschindler:
Yes, @rmuir is right. The TokenFilter should only subclass FilteringTokenFilter and implement accept(). By that all logic is obsolete and it get's a one-liner.

I am not sure what SecureRandom or UUID has to do here. We should maybe allow to pass fillter token as parameter. If not given maybe default to empty token?

uschindler:
The fillter token is obsolete. If you implement FilteringTokenFilter's accept don't try to inject tokens like this. Tests will pass.

lucene/analysis/common/src/java/org/apache/lucene/analysis/miscellaneous/DropIfFlaggedFilter.java
```diff
@@ -0,0 +1,90 @@
...
+    this.dropFlags = dropFlags;
+  }
+
+  @Override
+  public final boolean incrementToken() throws IOException {
+    boolean result;
+    boolean dropToken;
+    do {
+      result = input.incrementToken();
+      dropToken = (getAttribute(FlagsAttribute.class).getFlags() & dropFlags) == dropFlags;
```
uschindler, line 58: you have to move this to final fiel and use FlagsAttribute flagsAttr = addAttribute(FlagsAttribute.class)

uschindler:
The problem why you hit the exception is wrong implementation. Don't look at tokens before calling incrementToken().
In addition, you must call addAttribute() before the loop, so the FlagsAttribute instance is created before. Please drop the whole PR. It's plain wrong.

Most of tests are not needed if you implement FilteringTokenFilter.

uschindler (changes requested):
Please reimplement the whole filter according to TokenFilter standards. Please read documentation. This is plain wrong.

uschindler:
FilteringTokenFilter also corrects positions. No filler tokens needed.

uschindler:
This would be your filter, plain simple, correct positions, no fillter tokens needed. All bug-specific tests are obsolete, as this type of Filter is tested to hell (we have many token-dropping filters):

```java
public final class DropIfFlaggedFilter extends FilteringTokenFilter {
  private final FlagsAttribute flagsAtt = addAttribute(FlagsAttribute.class);

  private final int dropFlags;

  public DropIfFlaggedFilter (TokenStream in, int dropFlags) {
    super(in);
    this.dropFlags = dropFlags;
  }

  @Override
  public boolean accept() {
    return (flagsAtt.getFlags() & dropFlags) != dropFlags; // TODO maybe == 0 if all or any flag is enough
  }
}
```

lucene/analysis/common/src/java/org/apache/lucene/analysis/miscellaneous/DropIfFlaggedFilter.java
```diff
@@ -0,0 +1,90 @@
...
+import org.apache.lucene.analysis.tokenattributes.FlagsAttribute;
+
+/**
+ * Allows Tokens with a given combination of flags to be dropped.
+ *
+ * @see DropIfFlaggedFilterFactory
+ */
+public class DropIfFlaggedFilter extends TokenFilter {
+
+  private int dropFlags;
```
uschindler, line 35: must be final

gus-asf:
HI @uschindler , your comments sound like you think I will be unwilling to take feedback. I'm not sure why you feel this way and seem to need to say the same thing repeatedly. This is broken out from the original contribution ticket for the very purpose of getting feedback. That said I do have a concern with the solution that FilteringTokenFilter provides... Consider this case:

Text: "January 401(k) contribution"
Whitespace tok: "January"(pi:1) "401l"(pi1) "contribution"(pi1)
PatternTypingFilter: "January"(pi:1), "401(k)"(pi1:flag:2, type:legal2_401_k), "contribution"(pi:1)
TokenAnalyzerFilter:"january"(pi:1),"401"(pi1:flag:2, type:legal2_401_k),"k",  "contribution"(pi:1)

uschindler:
Read the code if FilteringTokenFilter. It will fix the positions. What you describes is not an issue, as this also affects StopFilter or others.

If the first token is removed, FilteringTokenFilter will fix the position increment. Inserting a bullshit token is not needed.

There are tons of tests for this.

uschindler:
Your filter does not handle positions at all, so it will also corrupt the graph of later tokens are removed.
It's a state machine and you have to understand it. Your problem was thought about years ago, there's no need for discussion.

Sorry for being harsh, but I fortunately I have to say: your work does not follow any documentation about how a tokenfilters should behave. If you read javadocs it's all explained. The whole thing is plain wrong an using that at the Library of Congress would feel me really bad...

uschindler:
Hi maybe let me explain it:
With the filter how you implemented it originally, you have the problem of the first token (that has a synonym) leading to a 0 position increment on the second token. If all token filters dropping tokens would be implemented like yours (just dropping away tkens), also StopFilter removing a stop word with synonym would lead to the fact. We also have many other token filters like that.

So my suggestion here is the following: Please use the abstract implementation for those dropping filters, called `FilteringTokenFilter`. Although you think you need to look forward/backwards (which is indeed not possible), the correct way to handle this is to change the PositionIncrementAttribute on the token following the dropped tokens. If you remove a token with no increment (the second synonym), nothing needs to be done (increment is 0). If you remove the first token (which has increment 1 or maybe 2), the increment needs to be moved to the following token. So if the very first token of the very first synonmy needs to be removed, the position increment (1) needs to be added to the second token. If the second token is a synonym, then it gets "upgraded" to first token (0 + 1 = 1). If the second token is a standard token, then it get increment (1 + 1 = 2). So the tokenstream knows that theres a gap at the first position.

FilteringTokenFilter handles this by taking care of all positions and fixing those of following tokens. Read the source code, it just sums up increments of dropped tokens and adds them to the next token that is kept. This is thouroughly testes.

The proposed TokenFilter here is nothing special, it just behaves like any other StopFilter-like filter. The bug you mentuon was fixed in StopFilter about 10 years ago (this was one of my first commits with @rmuir back at that time).

So please redo this PR:
- Copypaste the TokenFilter as posted above (very easy)
- Fix your tests to not expect a "dummy token".

If you have a problem with your other filter, fix that one.

I read your comment above, the discussion is still not correct. The tokenfilter graph must be correct. SpanQueries work correctly with FilteringTokenFilter.

If you want a different behaviour, open a separate issue, as a change around that needs to be done in FilteringTokenFilter, not a subclass or a reimplemented filter.

Sorry for being aggressive, but - sorry - the code you posted had really bad problems and was not accoring to any coding standards.

gus-asf:
In retrospect/review I can see that my solution does have some additional problems that probably just haven't been showing up yet because the cases in question aren't that common in the actual usage (but probably will crop up eventually). More importantly even if it did have a benefit in a particular case, a component in a library it shouldn't buck the behavioral trend like that. I did arrive there for a reason however.

I just hope you can understand that if one reads the documentation, does some work with it and then forgets something after not using it for a year or so, it can be hard to know that you've have forgotten something, especially when what you do remember _seems_ to be working. Would be nice if lucene had docs that included a table of contents or navigation by topic so that one didn't have to wade through a single (or a small number of) giant javadoc pages with many sections unrelated to whatever one is actually trying to understand at the moment.

Unfortunately this will have knock on effects on some other parts of the SIP, mostly because I need to solve the "extra" synonym problem now, because instead of sometimes loosing a single position, I'll now gain N positions where N is the number of times a synonym generating token got split before synonyms were generated, magnifying the problem and pushing it in a direction much more noticeable to users. Such is life and it's probably better in the long run.

lucene/analysis/common/src/java/org/apache/lucene/analysis/miscellaneous/DropIfFlaggedFilter.java
gus-asf: I guess this has been working because org.apache.lucene.analysis.Token happens to have this attribute already, but of course that's not guaranteed to be the impl one gets.

The final field... I've seen this done in a number of places including doc examples but I don't recall (and haven't been able to locate documentation as to) why it's important. Is there anything to it beyond saving an Class.cast() and a Map.get() call? It is a very hot area of the code so maybe something that small does matter here... Also I wonder if anything like the following has been considered so that users don't have to explicitly add attributes:

```
  public final <T extends Attribute> T getAttributeSafe(Class<T> attClass) {
    return attClass.cast(attributes.computeIfAbsent(attClass, this::addAttribute));
  }
```
That doesn't quite work with the current generics and obviously not for this ticket but is that an already discarded idea?

rmuir:
> Would be nice if lucene had docs that included a table of contents or navigation by topic so that one didn't have to wade through a single (or a small number of) giant javadoc pages with many sections unrelated to whatever one is actually trying to understand at the moment.

As far as the docs go, I do think for this stuff, it is pretty prominent: if you go to lucene.apache.org and click "release docs" it takes you here: https://lucene.apache.org/core/8_6_3/index.html

Right at the top under "Getting Started" are two decent links that help with writing analysis stuff:
1. Analysis Overview: https://lucene.apache.org/core/8_6_3/core/org/apache/lucene/analysis/package-summary.html#package.description
2.TokenStream Consumer Workflow: https://lucene.apache.org/core/8_6_3/core/org/apache/lucene/analysis/TokenStream.html

msokolov:
> > Would be nice if lucene had docs that included a table of contents or navigation by topic so that one didn't have to wade through a single (or a small number of) giant javadoc pages with many sections unrelated to whatever one is actually trying to understand at the moment.
> 
> As far as the docs go, I do think for this stuff, it is pretty prominent: if you go to lucene.apache.org and click "release docs" it takes you here: https://lucene.apache.org/core/8_6_3/index.html
> 
> Right at the top under "Getting Started" are two decent links that help with writing analysis stuff:
> 
>     1. Analysis Overview: https://lucene.apache.org/core/8_6_3/core/org/apache/lucene/analysis/package-summary.html#package.description
>       As far as the docs go, I do think for this stuff, it is pretty prominent: if you go to lucene.apache.org and click "release docs" it takes you here

Maybe I'm blind, but I scanned that page and do not see the "release docs" link - tried Ctrl-F too

dweiss:
>
> Maybe I'm blind, but I scanned that page and do not see the "release docs" link - tried Ctrl-F too
>
>
The link is on Lucene's page, here:
https://lucene.apache.org/core/

uschindler:
Hi,
I would really recommend on reading the documentation as provided with links by Robert. The whole issue discussed here is lengthly explained there. It also states, that when you remove tokens, the positionIncrement must be adapted. And for convenience and bugfree tokenfilters use FilteringTokenFilter for that.

As said before, if you need so additional convenience for your specific use case (like adding marker tokens) this should NOT be done in this filter. The DropIfFlaggedFilter is a typical example on a filtering filter. The magic have to be according to the standard also mentioned in the docs.

As said before, if you need to adjust positions (e.g. remove gaps), this should be a separate filter.

This pull request should just add the filter cas copypasted in my comment above and the correct amount of tests added. But as it extends FilteringTokenFilter no position tests are needed.

I wish you a nice weekend. Maybe we have a better start with the next tokenizer components you intend to submit for the SIP / QueryParser. Sorry for the hassle!

msokolov:
> The link is on Lucene's page, here: https://lucene.apache.org/core/

Thanks, @dweiss . The Documentation link is nice and prominent there

PR #1967, opened 2020-10-08
LUCENE-9577: Move Lucene/Solr Documentation assembly to subproject

uschindler:
This PR moves the changes.txt formatting, site Javadocs, markdown formatting to subprojects (`:lucene:documentation`, `:solr:documentation`.

This is still WIP, not everything works.


uschindler:
This also removed some unnecessary assets from Solr's documentation. Those are leftovers from time before the refguide was included.

dweiss (approved):
Looks good to me, thanks for taking care of it, Uwe.

uschindler:
Thanks Dawid. There's still work to do. The Lucene packaging brings a totally strange dependency error.
Solr works. The only difference is configuration name. But it was too late yesterday.
I will also cleanup some parts where I repeatedly used 'project.parent.something' to get out of documentation subproject to Lucene or Solr main.
But it gets closed. Once I am done it would be good to have another look.

dweiss:
I only looked at the diff. If you need my help, ping me. I'm working but I can take a look later to see what that error is.

uschindler:
I fixed the problem with the dependency. Was some explicit exclusion of the "empty" subprojects. Maybe this should be changed to only collect project that actually have the correct configuration.

uschindler:
Hi @dweiss, 
I think all is working now. Can you have a quick look.

This should also make the split between lucene and solr easier, as the changes2html per script and resources was moved to the `gradle/documentation` subdir.

dweiss:
LGTM, Uwe.

uschindler:
I also fixed the links to the new output folder in the ref guide

PR #1968, opened 2020-10-09
LUCENE-9562: Unify 'analysis' package with produced artifact names (-analyzers- to -analysis-)

dweiss:


msokolov:
+1 - this would be for 9.0 only, I assume?

dweiss:
Correct, no backport planned to 8x.

PR #1971, opened 2020-10-09
SOLR-14900: 8.x changes to update Jekyll

ctargett:
PR #1923 will upgrade Jekyll to 4.x for the master branch. I propose to do the same thing with our Ant build in branch_8x. This also fixes the code syntax styling bug fixed in the other PR and updates the README.adoc to reflect the new version requirements.

I don't intend to merge this until _after_ 8.7 is released because it will require anyone who wants to build the Ref Guide in 8.x to upgrade Ruby to at least 2.5 (if it's not already) and upgrade Jekyll. If someone is trying to do something quick I don't want to slow them down.


ctargett:
We can't use Jekyll 4 due to problems with its dependencies on Windows machines (see also https://issues.apache.org/jira/browse/SOLR-15520). Closing this.

PR #1972, opened 2020-10-11
SOLR-14915: Prometheus-exporter does not depend on Solr-core any longer

dsmiley:
And enhanced gradle config to operate as an application
https://issues.apache.org/jira/browse/SOLR-14915

It's cool that I can now run this thing via `gw run` :-)
The Java-Application plugin for Gradle has other nice distribution facilities that render our attempts to hand-edit startup scripts obsolete.  However, using it would mean a much larger Solr distribution as it'd have duplicated libs that Solr has.  Any way, it's nice to leave here for people who might want to activate it manually.

As-is there is something not quite right.  I modified the solr-exporter bash script to not add all of the Solr server libs to the classpath.  However, the assembly task only populates a small number of dependencies into the "lib" here.  In particular I don't see transitive dependencies such as Jackson (used by jackson-jq).    @dweiss do you have suggestions here?  Alternatively, I could remove my changes to solr-exporter scripts and accept that they are going to add a ton more dependencies than actually needed.  That's okay.

`lucene-solr/solr/packaging/build/solr-9.0.0-SNAPSHOT/contrib/prometheus-exporter`
```
├── bin
│   ├── main
│   ├── solr-exporter
│   ├── solr-exporter.cmd
│   └── test
├── conf
│   ├── grafana-solr-dashboard.json
│   └── solr-exporter-config.xml
└── lib
    ├── argparse4j-0.8.1.jar
    ├── disruptor-3.4.2.jar
    ├── jackson-jq-0.0.8.jar
    ├── simpleclient-0.2.0.jar
    ├── simpleclient_common-0.2.0.jar
    └── simpleclient_httpserver-0.2.0.jar
```


solr/contrib/prometheus-exporter/src/java/org/apache/solr/prometheus/exporter/MetricsConfiguration.java
```diff
@@ -66,13 +76,27 @@ public PrometheusExporterSettings getSettings() {
...
+    // See solr-core XmlConfigFile
+    final DocumentBuilderFactory dbf = DocumentBuilderFactory.newInstance();
+    try {
+      dbf.setXIncludeAware(true);
+      dbf.setNamespaceAware(true);
+    } catch (UnsupportedOperationException e) {
+      log.warn("{} XML parser doesn't support XInclude option", file);
+    }
+
+    final Document document = dbf.newDocumentBuilder().parse(file.toFile());
```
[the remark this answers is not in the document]
dsmiley: Acknowledged, and doesn't matter when the XML file is a local configuration file as opposed to external input.

dweiss:
Well, yes. I know it's tempting to use gradle's "ready-to-use" plugins but the situation is different when you have a stand-alone project and a multi-module aggregate. Let me explain.

gradle plugins add tons of stuff (configurations, dependencies, tasks, conventions) and hide what's going on behind the scenes. This has pros and cons. While you can use the automatically-added 'run' task, the subproject is also populated with stuff that will increase the overall build time (like the stand-alone distributions on assemble) and may conflict with other parts of the multi-module build.

For this reason I typically prefer to be explicit about packaging and distribution aspects, even if some things need to be done manually. My preference would be to just keep using java-library and add corresponding entries to the JAR's manifest; the "run" task is still achievable with plain java-library - Lucene's Luke configuration has an example of how this can be done.  Please do as you please though - I don't even know what this particular contrib is :)

As for your question why not all JARs are copied - this is a larger question that applies to the current Solr packaging. I tried to emulate the way ant worked (in packaging.gradle) - this means copying the module's "unique" set of JARs to avoid duplication. This isn't easy and makes configuration management fairly hairy. I think stand-alone tools like prometheus-exporter or luke should have full executable distribution in the binary release - even if this means duplicating some JARs (including Lucene JARs). This makes it easier to reason about what they need, makes the configuration simpler... at the cost of increased distribution size.

The alternative is to hand-edit the build and include what's needed or hand-edit the scripts to point at the right JARs.

I'm really not sure which way is right (I believe the full set of artifacts required to run a stand-alone tool is better but it's my personal opinion).

dsmiley:
Based on your advise, I put back the "java-library" plugin in place of the application one, but added a "run" task.

The prometheus-exporter is fundamentally dissimilar to the other contribs.  The others have Solr plugins -- they run *inside* Solr, and thus don't need "lib" directories containing any of the libs that are in Solr.  But I don't think the prometheus-exporter should have such a limited list.  I see `gradle/solr/packaging.gradle` and it treats all contribs as the same.  So I added more assemblePackage code here that ends up adding other libs like jackson because it does not exclude all so-called "Solr platform libs".   Also, I ensured all server/lib/ext JARs (which are all logging) don't get copied to any contrib's libs.

solr/contrib/prometheus-exporter/src/java/org/apache/solr/prometheus/exporter/SolrExporter.java
```diff
@@ -68,7 +66,7 @@
   private static final String[] ARG_CONFIG_FLAGS = {"-f", "--config-file"};
   private static final String ARG_CONFIG_METAVAR = "CONFIG";
   private static final String ARG_CONFIG_DEST = "configFile";
-  private static final String ARG_CONFIG_DEFAULT = "solr-exporter-config.xml";
+  private static final String ARG_CONFIG_DEFAULT = "conf/solr-exporter-config.xml";
```
dsmiley, line 69: @HoustonPutman you removed the conf/ and I'm putting it back.  I'm not sure why it worked with it gone.  It may be related to other changes in this PR that it's needed again.  I tested that this works here both with `gw run` and via executing normally from the built assembly.  I didn't test Docker yet.

dsmiley (commented):
FYI:
````
├── bin
│   ├── solr-exporter
│   └── solr-exporter.cmd
├── conf
│   ├── grafana-solr-dashboard.json
│   ├── log4j2.xml
│   └── solr-exporter-config.xml
└── lib
    ├── argparse4j-0.8.1.jar
    ├── caffeine-2.8.4.jar
    ├── jackson-annotations-2.10.1.jar
    ├── jackson-core-2.10.1.jar
    ├── jackson-databind-2.10.1.jar
    ├── jackson-jq-0.0.8.jar
    ├── simpleclient-0.2.0.jar
    ├── simpleclient_common-0.2.0.jar
    └── simpleclient_httpserver-0.2.0.jar

3 directories, 14 files
````

**Comment by dsmiley** [comment]
This has been ready I think; I'll merge Monday.

File: solr/contrib/prometheus-exporter/src/java/org/apache/solr/prometheus/exporter/SolrExporter.java (hunk 2 above)
Comment (HoustonPutman) [reply solr/contrib/prometheus-exporter/src/java/org/apache/solr/prometheus/exporter/SolrExporter.java:69]: So the reason I removed the `conf/` is because I was reading the file off of the classpath. That way you can run `solr-exporter` from anywhere, not just `contrib/prometheus-exporter`.

The reason it doesn't work anymore is that we aren't using the SolrResourceLoader in this PR. I have a fix made that should allow for both the default config to be read off of the classpath or other configs specified via a file location. I can push it to this branch if you don't mind.

Comment (dsmiley) [reply solr/contrib/prometheus-exporter/src/java/org/apache/solr/prometheus/exporter/SolrExporter.java:69]: Sure; push if you want.  I don't think it's important to read off the classpath so long as it reads it from somewhere... but if it works, that's nice.

**Comment by HoustonPutman** [comment]
I didn't know there was a stigma around `File`, but made the change to use `Path` and `Files` instead.

File: solr/contrib/prometheus-exporter/src/java/org/apache/solr/prometheus/exporter/MetricsConfiguration.java [hunk 3]
```diff
@@ -66,13 +81,36 @@ public PrometheusExporterSettings getSettings() {
...
+    final DocumentBuilderFactory dbf = DocumentBuilderFactory.newInstance();
+    try {
+      dbf.setXIncludeAware(true);
+      dbf.setNamespaceAware(true);
+    } catch (UnsupportedOperationException e) {
+      log.warn("{} XML parser doesn't support XInclude option", path);
+    }
+
+    Document document;
+    File file = new File(path);
```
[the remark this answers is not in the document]
dsmiley: Not a problem as the config can only be specified by an admin on the local machine.

solr/contrib/prometheus-exporter/src/java/org/apache/solr/prometheus/exporter/MetricsConfiguration.java
```diff
@@ -77,22 +79,22 @@ public PrometheusExporterSettings getSettings() {
...
+      log.warn("{} XML parser doesn't support XInclude option", resource);
     }
 
     Document document;
-    File file = new File(path);
-    if (file.isFile()) {
-      document = dbf.newDocumentBuilder().parse(file);
+    Path path = Path.of(resource);
+    if (Files.exists(path)) {
+      document = dbf.newDocumentBuilder().parse(path.toAbsolutePath().toString());
```
dsmiley, line 95: This `parse(String)` is documented to take a URI.  I don't think a file path is a URI?  If you look at the implementation of the variant that takes a File, it converts it to a URI then and then an ASCII string.  This could be done via `path.toUri().toASCIIString()`.   FWIW I'd also be okay with just doing path.toFile() here.

HoustonPutman: Good catch. I tested it in multiple configurations, but there's probably some edge case it doesn't work for. I'll use `path.toUri().toASCIIString()` since the `parse(file)` method just gets the URI of the file and passes that to the `parse(uri)` implementation anyways. Might as well skip the middleman-method.

PR #1976, opened 2020-10-12
LUCENE-9578: TermRangeQuery empty string lower bound edge case

cbuescher:
# Description

Currently a TermRangeQuery with the empty String ("") as lower bound and
includeLower=false leads internally constructs an Automaton that doesn't match
anything. This is unexpected expecially for open upper bounds where any string
should be considered to be "higher" than the empty string.

# Solution

This PR changes "Automata#makeBinaryInterval" so that for an empty string lower
bound and an open upper bound, any String should match the query regardless or
the includeLower flag.

# Tests

Added two new tests to `TestAutomaton`.

# Checklist

- [x] I have run `./gradlew check`

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/util/automaton/Automata.java
+++ b/lucene/core/src/java/org/apache/lucene/util/automaton/Automata.java
@@ -85,7 +85,15 @@ final public class Automata {
     a.finishState();
     return a;
   }
-  
+
+  /**
+   * Returns a new (deterministic) automaton that accepts all binary terms except
+   * the empty string.
+   */
+  public static Automaton makeAnyBinaryExceptEmpty() {
+    return makeCharRange(0, 255);
+  }
+
   /**
    * Returns a new (deterministic) automaton that accepts any single codepoint.
    */
@@ -255,7 +263,11 @@ final public class Automata {
     } else {
       cmp = -1;
       if (min.length == 0) {
-        return makeAnyBinary();
+        if (minInclusive) {
+          return makeAnyBinary();
+        } else {
+          return makeAnyBinaryExceptEmpty();
+        }
       }
     }
 
--- a/lucene/core/src/java/org/apache/lucene/util/automaton/Automata.java
+++ b/lucene/core/src/java/org/apache/lucene/util/automaton/Automata.java
@@ -90,7 +90,7 @@ final public class Automata {
    * Returns a new (deterministic) automaton that accepts all binary terms except
    * the empty string.
    */
-  public static Automaton makeAnyBinaryExceptEmpty() {
+  public static Automaton makeNonEmptyBinary() {
     Automaton a = new Automaton();
     int s1 = a.createState();
     int s2 = a.createState();
@@ -273,7 +273,7 @@ final public class Automata {
         if (minInclusive) {
           return makeAnyBinary();
         } else {
-          return makeAnyBinaryExceptEmpty();
+          return makeNonEmptyBinary();
         }
       }
     }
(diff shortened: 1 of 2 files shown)


lucene/core/src/java/org/apache/lucene/util/automaton/Automata.java
```diff
@@ -254,7 +254,7 @@ public static Automaton makeBinaryInterval(BytesRef min, boolean minInclusive, B
       cmp = min.compareTo(max);
     } else {
       cmp = -1;
-      if (min.length == 0 && minInclusive) {
+      if (min.length == 0) {
```
jpountz, line 257: this looks wrong as we should still make sure that the empty string is rejected if min=="" and minInclusive==false?

cbuescher:
@jpountz thanks for the review, I added a commit that rejects the empty string when "includeMin == false" and added tests for that edge case as well.

lucene/core/src/java/org/apache/lucene/util/automaton/Automata.java
```diff
@@ -85,7 +85,15 @@ public static Automaton makeAnyBinary() {
...
     return a;
   }
-  
+
+  /**
+   * Returns a new (deterministic) automaton that accepts all binary terms except
+   * the empty string.
+   */
+  public static Automaton makeAnyBinaryExceptEmpty() {
+    return makeCharRange(0, 255);
```
jpountz, line 94: I don't think this is right since this only accepts strings of length 1.

cbuescher: Yes, I was missing that this doesn't include a self-referencing accept state. Fixed that and added more tests to check this is working now.

cbuescher:
@jpountz thanks for the review, I added another commit that should fix the problem you detected.

lucene/core/src/java/org/apache/lucene/util/automaton/Automata.java
```diff
@@ -85,7 +85,22 @@ public static Automaton makeAnyBinary() {
...
+
+  /**
+   * Returns a new (deterministic) automaton that accepts all binary terms except
+   * the empty string.
+   */
+  public static Automaton makeAnyBinaryExceptEmpty() {
+    Automaton a = new Automaton();
+    int s1 = a.createState();
+    int s2 = a.createState();
+    a.setAccept(s2, true);
```
jpountz, line 97: Do we need to call setAccept on s1 too? Or is it done implicitly elsewhere (I'm not completely familiar with the Automaton code)

lucene/core/src/java/org/apache/lucene/util/automaton/Automata.java
```diff
@@ -85,7 +85,22 @@ public static Automaton makeAnyBinary() {
     a.finishState();
     return a;
   }
-  
+
+  /**
+   * Returns a new (deterministic) automaton that accepts all binary terms except
+   * the empty string.
+   */
+  public static Automaton makeAnyBinaryExceptEmpty() {
```
jpountz, line 93: Maybe call `makeNonEmptyBinary`?

lucene/core/src/java/org/apache/lucene/util/automaton/Automata.java
jpountz: Oops ignore this comment I was totally confused. :)

jpountz:
I had a deeper look to check whether this could be an indicator of other bugs lurking, but it doesn't seem so and the fix looks good to me. For reference here's what the automaton looks like today with an empty string as a min bound and includeMin==false before the above fix.
empty

cbuescher:
@jpountz thanks, I renamed the method. btw. would this change also be backported to the latest 8.7 branch so we can use it in ES to fix https://github.com/elastic/elasticsearch/issues/63386?

jpountz:
The 8.7 branch has not been cut yet so this will be in 8.7. This change looked safe to me so I felt free to merge in spite of the branch being imminently cut.

PR #1977, opened 2020-10-12
SOLR-14907: Support single file upload/overwrite in configSet API

HoustonPutman:
https://issues.apache.org/jira/browse/SOLR-14907

### DIFF
--- a/solr/core/src/java/org/apache/solr/handler/admin/ConfigSetsHandler.java
+++ b/solr/core/src/java/org/apache/solr/handler/admin/ConfigSetsHandler.java
@@ -170,10 +170,14 @@ public class ConfigSetsHandler extends RequestHandlerBase implements PermissionN
 
     boolean overwritesExisting = zkClient.exists(configPathInZk, true);
 
-    if (overwritesExisting && !req.getParams().getBool(ConfigSetParams.OVERWRITE, false)) {
-      throw new SolrException(ErrorCode.BAD_REQUEST,
-          "The configuration " + configSetName + " already exists in zookeeper");
-    }
+    // Get upload parameters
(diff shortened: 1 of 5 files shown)


solr/core/src/java/org/apache/solr/handler/admin/ConfigSetsHandler.java
```diff
@@ -170,10 +170,14 @@ private void handleConfigUploadRequest(SolrQueryRequest req, SolrQueryResponse r
...
 
-    if (overwritesExisting && !req.getParams().getBool(ConfigSetParams.OVERWRITE, false)) {
-      throw new SolrException(ErrorCode.BAD_REQUEST,
-          "The configuration " + configSetName + " already exists in zookeeper");
-    }
+    // Get upload parameters
+    String singleFilePath = req.getParams().get(ConfigSetParams.FILE_PATH, "");
+    boolean allowOverwrite = req.getParams().getBool(ConfigSetParams.OVERWRITE, false);
+    // Cleanup is not allowed while using singleFilePath upload
+    boolean cleanup = singleFilePath.isEmpty() && req.getParams().getBool(ConfigSetParams.CLEANUP, false);
```
tflobbe, line 177: should we error instead of silently ignoring the `cleanup` param? it defaults to `false`, so someone must have explicitly set it to `true`

tflobbe (approved):
LGTM

solr/core/src/java/org/apache/solr/handler/admin/ConfigSetsHandler.java
HoustonPutman: Added error handling for that and cases where a bad filePath is given.

PR #1978, opened 2020-10-13
LUCENE-9524: Fix NPE in SpanWeight#explain when no scoring is require…

zacharymorn:
# Description
`SpanWeight#explain` uses `Similarity.SimScorer` to generate explanations, and may fail with NullPointerException when scoring is not needed and thus `Similarity.SimScorer` is set to null, such as matching for must not occur span query.

# Solution
The solution is to check for null `Similarity.SimScorer` in `SpanWeight#explain`, and provides alternative explanation.

# Tests
Added 1 integration test to `TestMemoryIndex` that was broken before the fix.

# Checklist

### DIFF
--- a/lucene/core/build.gradle
+++ b/lucene/core/build.gradle
@@ -22,5 +22,4 @@ description = 'Lucene core library'
 dependencies {
   testImplementation project(':lucene:codecs')
   testImplementation project(':lucene:test-framework')
-  testImplementation project(':lucene:queryparser')
 }
--- a/lucene/core/build.gradle
+++ b/lucene/core/build.gradle
@@ -22,4 +22,5 @@ description = 'Lucene core library'
 dependencies {
   testImplementation project(':lucene:codecs')
   testImplementation project(':lucene:test-framework')
+  testImplementation project(':lucene:queryparser')
 }
--- a/lucene/memory/src/test/org/apache/lucene/index/memory/TestMemoryIndex.java
+++ b/lucene/memory/src/test/org/apache/lucene/index/memory/TestMemoryIndex.java
@@ -57,18 +57,14 @@ import org.apache.lucene.index.SortedNumericDocValues;
 import org.apache.lucene.index.SortedSetDocValues;
 import org.apache.lucene.index.Term;
 import org.apache.lucene.index.TermsEnum;
-import org.apache.lucene.queryparser.classic.ParseException;
-import org.apache.lucene.queryparser.complexPhrase.ComplexPhraseQueryParser;
 import org.apache.lucene.search.CollectionStatistics;
 import org.apache.lucene.search.DocIdSetIterator;
 import org.apache.lucene.search.IndexSearcher;
 import org.apache.lucene.search.MatchAllDocsQuery;
 import org.apache.lucene.search.PhraseQuery;
 import org.apache.lucene.search.Query;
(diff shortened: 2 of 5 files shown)


lucene/memory/src/test/org/apache/lucene/index/memory/TestMemoryIndex.java
```diff
@@ -618,4 +622,21 @@ public void testToStringDebug() {
         "fields=2, terms=2, positions=3", mi.toStringDebug());
   }
 
+  public void testExplain() throws ParseException, IOException {
```
jpountz, line 625: I don't think this is specific to MemoryIndex, so can you move this test to TestSpansExplanations?

zacharymorn: Done

lucene/memory/src/test/org/apache/lucene/index/memory/TestMemoryIndex.java
```diff
@@ -618,4 +622,21 @@ public void testToStringDebug() {
...
+    Query query = new ComplexPhraseQueryParser("content", analyzer).parse(queryString);
+
+    MemoryIndex memoryIndex = new MemoryIndex(true);
+    memoryIndex.addField("content", text, analyzer);
+
+    IndexSearcher searcher = memoryIndex.createSearcher();
+
+    TopDocs topDocs = searcher.search(query, 1);
+    ScoreDoc match = topDocs.scoreDocs[0];
+    searcher.explain(query, match.doc);
```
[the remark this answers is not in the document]
zacharymorn: I've tried this and it seems to be a bit more complicated to do. The problem here is that under the current test setup, the new explanation string code will only be exercised when there's a match between the (none-scoring) prohibited clause and the text.  So the matched topDoc that exercised that code path will actually not include that particular explanation as it must have been matching alternative clause instead.

Therefore, I'm considering two alternatives here. One would be to test `SpanWeight#explain` directly with null `SimScorer`; the other would be to setup another test to match span query without scoring, but I tried query with Occur.FILTER and MUST and see there's still a 1.0 boost associated and valid `SimScorer`. So I'm not sure if the latter approach is something actually achievable / meaningful to implement? 

Screen Shot 2020-10-14 at 7 26 18 PM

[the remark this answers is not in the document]
zacharymorn: I've found a work-around in test and able to test with tighter condition, and assert for explanation message. Could you please let me know if this looks good to you?

lucene/core/build.gradle
```diff
@@ -22,4 +22,5 @@ description = 'Lucene core library'
 dependencies {
   testImplementation project(':lucene:codecs')
   testImplementation project(':lucene:test-framework')
+  testImplementation project(':lucene:queryparser')
```
jpountz, line 25: we should be able to remove this now?

jpountz (approved):
Thanks, the new test looks good to me! I left a minor comment about a dependency that looks no longer needed.

lucene/core/build.gradle
zacharymorn: Ops yes removed.

jpountz:
Thanks @zacharymorn !

PR #1979, opened 2020-10-14
LUCENE-9574 Add DropIfFlaggedFilterFactory

gus-asf:


gus-asf:
Oh lovely, git hub decided to convert my ( originally all lower case) @ since to a user's name...

lucene/analysis/common/src/java/org/apache/lucene/analysis/miscellaneous/DropIfFlaggedFilter.java
```diff
@@ -0,0 +1,52 @@
...
+  private final int dropFlags;
+
+  /**
+   * Construct a token stream filtering the given input.
+   *
+   * @param input the source stream
+   * @param dropFlags a combination of flags that indicates that the token should be dropped.
+   */
+  @SuppressWarnings("WeakerAccess")
+  protected DropIfFlaggedFilter(TokenStream input, int dropFlags) {
```
uschindler, line 42: Can you make the class final and make the constructor public? Why is this protected?

lucene/analysis/common/src/java/org/apache/lucene/analysis/miscellaneous/DropIfFlaggedFilter.java
```diff
@@ -0,0 +1,52 @@
...
+import org.apache.lucene.analysis.tokenattributes.FlagsAttribute;
+
+/**
+ * Allows Tokens with a given combination of flags to be dropped. If all flags specified are present
+ * the token is dropped, otherwise it is retained.
+ *
+ * @see DropIfFlaggedFilterFactory
+ * @since 8.8.0
+ */
+public class DropIfFlaggedFilter extends FilteringTokenFilter {
```
uschindler, line 31: You should make the class final, too. That's not a requirement, but would be in line with other filters. The reason is: We don't want subclasses, as we prefer delegation over inheritance.

uschindler (changes requested):
The TokenFilter looks fine to me, I only ask to make its constructor public and remove the weaker access annotation.

I am not so happy about the tests using the "obsolete" Token class. I'd use a MockTokenizer with a string instead.

lucene/CHANGES.txt
```diff
@@ -19,7 +19,7 @@ API Changes
 * LUCENE-8474: RAMDirectory and associated deprecated classes have been
   removed. (Dawid Weiss)
 
-* LUCENE-3041: The deprecated Weight#extractTerms() method has been
```
uschindler, line 22: Hi could you please undo all the formaating changes in the CHANGES.txt? This makes merging of other pull requests hard.
Maybe your editor reformatted the whole file. I'd just copy the new changes entry into the file.

gus-asf: Yeah apparently the file has a bazillion lines that end in spaces. I typically want this on in the editor to avoid silly extra white-space in code, I'll have to figure out how to turn it off temporarily. Might not be bad to have a commit at some point that fixes this up, but we don't have to do it here.

lucene/analysis/common/src/java/org/apache/lucene/analysis/miscellaneous/DropIfFlaggedFilter.java
gus-asf: That said if it's a standard for the project I'll do it. I have my disagreements but we can hash that out elsewhere.

lucene/analysis/common/src/java/org/apache/lucene/analysis/miscellaneous/DropIfFlaggedFilter.java
gus-asf: Not sure why that wound up protected.  Thx for noticing this.

gus-asf:
> I am not so happy about the tests using the "obsolete" Token class. I'd use a MockTokenizer with a string instead.

I don't see how this would allow me to set flags on the individual tokens? Do you have an example of that?

lucene/analysis/common/src/java/org/apache/lucene/analysis/miscellaneous/DropIfFlaggedFilter.java
uschindler: Thanks. All fine.

uschindler (approved):
All fine, thanks!

PR #1987, opened 2020-10-15
SOLR-14937: Correct client.queryDefaults().set(...) calls in some JSON facet tests.

cpoerschke:
https://issues.apache.org/jira/browse/SOLR-14937


munendrasn (approved):
+1

PR #1992, opened 2020-10-16
SOLR-14939: JSON range faceting to support cache=false parameter

cpoerschke:
https://issues.apache.org/jira/browse/SOLR-14939


solr/core/src/java/org/apache/solr/search/facet/FacetRangeProcessor.java
```diff
@@ -531,7 +533,20 @@ private  SimpleOrderedMap getRangeCountsIndexed() throws IOException {
...
   private DocSet[] intersections;
   private void rangeStats(Range range, int slot) throws IOException {
-    Query rangeQ = sf.getType().getRangeQuery(null, sf, range.low == null ? null : calc.formatValue(range.low), range.high==null ? null : calc.formatValue(range.high), range.includeLower, range.includeUpper);
+    final Query rangeQ;
+    {
+      final Query rangeQuery = sf.getType().getRangeQuery(null, sf, range.low == null ? null : calc.formatValue(range.low), range.high==null ? null : calc.formatValue(range.high), range.includeLower, range.includeUpper);
+      if (fcontext.cache) {
+        rangeQ = rangeQuery;
+      } else if (rangeQuery instanceof ExtendedQuery) {
+        ((ExtendedQuery) rangeQuery).setCache(fcontext.cache);
```
madrob, line 542: Here (and in the else) I think I would explicitly do `setCache(false)` as it feels more readable to me, but I don't have strong opinions on that.

solr/core/src/test/org/apache/solr/search/facet/TestJsonRangeFacets.java
```diff
@@ -41,6 +42,7 @@ public static void beforeTests() throws Exception {
     if (Boolean.getBoolean(NUMERIC_POINTS_SYSPROP)) System.setProperty(NUMERIC_DOCVALUES_SYSPROP,"true");
 
     initCore("solrconfig-tlog.xml","schema_latest.xml");
+    cache = random().nextBoolean();
```
madrob, line 45: Might as well store the string directly?

madrob (approved):

madrob (approved):

PR #1993, opened 2020-10-16
.gitignore clean up

dsmiley:


.gitignore
```diff
@@ -1,42 +1,30 @@
...
-__pycache__
-/dev-tools/scripts/scripts.iml
-.DS_Store
+### At the root
+
+/.gradle/
+/.idea/
+# Ignore the generated local settings file.
+/gradle.properties
+/*~ # TODO?
```
msokolov, line 7: This should be plain `*~` not rooted at `/` - it's for emacs backup files

.gitignore
```diff
@@ -1,42 +1,33 @@
...
-.idea/
+build/
 
-# Ignore the generated local settings file.
-gradle.properties
+# macOS
+.DS_Store
+
+# Eclipse Gradle oddity
+**/bin/default/
```
madrob, line 23: Move this to the other Eclipse section?

dsmiley: What you refer to as "the other Eclipse section" is in a higher level "at the root", and this rule here is not at the root, it's at any directory.  That said, I suppose I should toss out that "at the root" label as it's better to put all these eclipse rules together.

madrob: I don't use Eclipse so this doesn't really affect me much, but I think it's easier to have contiguous sections I can ignore instead of it being piecemeal throughout the file.

dsmiley:
@msokolov you added `.#*` -- what comment should I use in this file to explain what this is?

msokolov:
> @msokolov you added .#* -- what comment should I use in this file to explain what this is?
It's more cruft emacs sometimes leaves behind _ I think in this case it's an autosave backup file left if you exited while editing. On second thought, we probably don't need to list this here - it shouldn't arise as a normal thing

.gitignore
```diff
@@ -1,43 +1,27 @@
...
+/gradle.properties
 .gradle/
-.idea/
+build/
 
-# Ignore the generated local settings file.
-gradle.properties
+# IntelliJ IDEA
+/.idea/
+#    IntelliJ creates this folder, ignore.
```
dweiss, line 9: This is caused by the fact that it's a composite build and we don't configure it with such extra care as we do with other projects (we don't override intellij's defaults). I don't think this matters.

dweiss (approved):

PR #1995, opened 2020-10-16
LUCENE-9575 Add PatternTypingFilter to annotate tokens with flags and types

gus-asf:

### DIFF
--- a/lucene/analysis/common/src/java/org/apache/lucene/analysis/pattern/PatternTypingFilter.java
+++ b/lucene/analysis/common/src/java/org/apache/lucene/analysis/pattern/PatternTypingFilter.java
@@ -24,8 +24,6 @@ import org.apache.lucene.analysis.tokenattributes.FlagsAttribute;
 import org.apache.lucene.analysis.tokenattributes.TypeAttribute;
 
 import java.io.IOException;
-import java.util.LinkedHashMap;
-import java.util.Map;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
@@ -41,12 +39,12 @@ import java.util.regex.Pattern;
  */
 public class PatternTypingFilter extends TokenFilter {
 
-  private final Map<Pattern, Map.Entry<String, Integer>> replacementAndFlagByPattern;
+  private final PatternTypingRule[] replacementAndFlagByPattern;
   private final CharTermAttribute termAtt = addAttribute(CharTermAttribute.class);
   private final FlagsAttribute flagAtt = addAttribute(FlagsAttribute.class);
   private final TypeAttribute typeAtt = addAttribute(TypeAttribute.class);
 
-  public PatternTypingFilter(TokenStream input, LinkedHashMap<Pattern, Map.Entry<String, Integer>> replacementAndFlagByPattern) {
+  public PatternTypingFilter(TokenStream input,  PatternTypingRule... replacementAndFlagByPattern) {
     super(input);
     this.replacementAndFlagByPattern = replacementAndFlagByPattern;
   }
@@ -54,18 +52,44 @@ public class PatternTypingFilter extends TokenFilter {
   @Override
   public final boolean incrementToken() throws IOException {
     if (input.incrementToken()) {
-      for (Map.Entry<Pattern, Map.Entry<String, Integer>> patRep : replacementAndFlagByPattern.entrySet()) {
-        Matcher matcher = patRep.getKey().matcher(termAtt);
+      for (PatternTypingRule rule : replacementAndFlagByPattern) {
+        Matcher matcher = rule.getPattern().matcher(termAtt);
         if (matcher.find()) {
-          Map.Entry<String, Integer> replAndFlags = patRep.getValue();
           // allow 2nd reset() and find() that occurs inside replaceFirst to avoid excess string creation
-          typeAtt.setType(matcher.replaceFirst(replAndFlags.getKey()));
-          flagAtt.setFlags(replAndFlags.getValue());
+          typeAtt.setType(matcher.replaceFirst(rule.getTypeTemplate()));
+          flagAtt.setFlags(rule.getFlags());
           return true;
         }
       }
       return true;
     }
     return false;
+  }
+
+  /**
+   * Value holding class for pattern typing rules.
+   */
+  public static class PatternTypingRule {
+    private final Pattern pattern;
+    private final int flags;
+    private final String typeTemplate;
+
+    public PatternTypingRule(Pattern pattern, int flags, String typeTemplate) {
+      this.pattern = pattern;
+      this.flags = flags;
+      this.typeTemplate = typeTemplate;
+    }
+
+    public Pattern getPattern() {
+      return pattern;
+    }
+
+    public int getFlags() {
+      return flags;
+    }
+
+    public String getTypeTemplate() {
+      return typeTemplate;
+    }
   }
 }
--- a/lucene/analysis/common/src/java/org/apache/lucene/analysis/pattern/PatternTypingFilter.java
+++ b/lucene/analysis/common/src/java/org/apache/lucene/analysis/pattern/PatternTypingFilter.java
@@ -1,0 +1,78 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.lucene.analysis.pattern;
+
+import java.io.IOException;
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import org.apache.lucene.analysis.TokenFilter;
+import org.apache.lucene.analysis.TokenStream;
+import org.apache.lucene.analysis.tokenattributes.CharTermAttribute;
+import org.apache.lucene.analysis.tokenattributes.FlagsAttribute;
+import org.apache.lucene.analysis.tokenattributes.TypeAttribute;
+
+/**
+ * Set a type attribute to a parameterized value when tokens are matched by any of a several regex patterns. The
+ * value set in the type attribute is parameterized with the match groups of the regex used for matching.
+ * In combination with TypeAsSynonymFilter and DropIfFlagged filter this can supply complex synonym patterns
+ * that are protected from subsequent analysis, and optionally drop the original term based on the flag
+ * set in this filter. See {@link PatternTypingFilterFactory} for full documentation.
+ *
+ * @since 8.8.0
+ * @see PatternTypingFilterFactory
+ */
+public class PatternTypingFilter extends TokenFilter {
+
+  private final Map<Pattern, String> patterns;
+  private final Map<Pattern, Integer> flags;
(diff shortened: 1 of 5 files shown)


lucene/analysis/common/src/java/org/apache/lucene/analysis/pattern/PatternTypingFilter.java
```diff
@@ -0,0 +1,78 @@
...
+  public PatternTypingFilter(TokenStream input, LinkedHashMap<Pattern,String> patterns, Map<Pattern,Integer> flags) {
+    super(input);
+    this.patterns = patterns;
+    this.flags = flags;
+  }
+
+  @Override
+  public final boolean incrementToken() throws IOException {
+    if (input.incrementToken()) {
+      if (hasAttribute(CharTermAttribute.class)) {
```
uschindler, line 59: hasAttribute is obsolete as it always returns true. Explanation: All tokenizers and filters in a chain share the same attribute instance. As you added it in line 46, it will alwys be there. So remove the whole if statemen.

lucene/analysis/common/src/java/org/apache/lucene/analysis/pattern/PatternTypingFilter.java
```diff
@@ -0,0 +1,78 @@
...
+        String termText = termAtt.toString();
```
uschindler, line 60: You dont need to convert the termattribute to a termText String. You can dircetly pass termAtt to matcher, as termAtt implements CharSequence.

lucene/analysis/common/src/java/org/apache/lucene/analysis/pattern/PatternTypingFilter.java
```diff
@@ -0,0 +1,78 @@
...
+        for (Map.Entry<Pattern, String> patRep : patterns.entrySet()) {
+          Pattern pattern = patRep.getKey();
+          Matcher matcher = pattern.matcher(termText);
```
uschindler, line 63: pass `termAtt` instead of `termText` here.

lucene/analysis/common/src/java/org/apache/lucene/analysis/pattern/PatternTypingFilter.java
```diff
@@ -0,0 +1,78 @@
...
+          String replaced = matcher.replaceFirst(patRep.getValue());
+          // N.B. Does not support producing a synonym identical to the original term.
+          // Avoids having to match() then replace() which performs a second find().
+          if (!replaced.equals(termText)) {
```
uschindler, line 67: here the question is how to do the comparison without converting to a String. My idea:
Use <https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/regex/Matcher.html#replaceFirst(java.util.function.Function)> to do the replacement. By that you get start() and end() of match and you can also check that the whole region of the term matched. IMHO, the filter should make sure that the regex always matches all the input? Or is there another intention?

uschindler (changes requested):

lucene/analysis/common/src/java/org/apache/lucene/analysis/pattern/PatternTypingFilter.java
```diff
@@ -0,0 +1,78 @@
...
+ * In combination with TypeAsSynonymFilter and DropIfFlagged filter this can supply complex synonym patterns
+ * that are protected from subsequent analysis, and optionally drop the original term based on the flag
+ * set in this filter. See {@link PatternTypingFilterFactory} for full documentation.
+ *
+ * @since 8.8.0
+ * @see PatternTypingFilterFactory
+ */
+public class PatternTypingFilter extends TokenFilter {
+
+  private final Map<Pattern, String> patterns;
```
uschindler, line 44: would't it be better to only have one hashmap, returning tuples of replacement and flags? You may use <https://docs.oracle.com/javase/9/docs/api/java/util/Map.html#entry-K-V-> (e.g. `Map.entry(replacement, flag)` as a workaround to hold two values as a pair.

lucene/analysis/common/src/java/org/apache/lucene/analysis/pattern/PatternTypingFilterFactory.java
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  @Override
+  public void inform(ResourceLoader loader) throws IOException {
+    List<String> lines = getLines(loader, patternFile);
+    // format: # regex ::: typename[_$1[_$2 ...]]    (technically _$1 does not need the '_' but it usually makes sense)
+    // eg: 2 (\d+\(?([a-z])\)?\(?(\d+)\)? ::: legal3_$1_$2_3
+    // which yields legal3_501_c_3 for 501(c)(3) or 501c3 and sets the second lowest bit in flags
+    for (String line : lines) {
+      if (line.trim().startsWith("#")) {
```
danmuzi, line 100: This part seems to be checked in WordlistLoader.getLines!
https://github.com/apache/lucene-solr/blob/f7be9e83e43a549aa92ca6dc26972325f418100a/lucene/core/src/java/org/apache/lucene/analysis/AbstractAnalysisFactory.java#L280-L282

https://github.com/apache/lucene-solr/blob/f7be9e83e43a549aa92ca6dc26972325f418100a/lucene/core/src/java/org/apache/lucene/analysis/WordlistLoader.java#L217-L227

lucene/analysis/common/src/java/org/apache/lucene/analysis/pattern/PatternTypingFilterFactory.java
```diff
@@ -0,0 +1,120 @@
...
+        continue; // allow for comments
+      }
+      int firstSpace = line.indexOf(" "); // no leading spaces allowed
+      int flagsVal = Integer.parseInt(line.substring(0,firstSpace));
```
danmuzi, line 104: Space character is missing after the comma in the substring method parameter!

lucene/analysis/common/src/java/org/apache/lucene/analysis/pattern/PatternTypingFilterFactory.java
```diff
@@ -0,0 +1,120 @@
...
+      line = line.substring(firstSpace + 1);
+      String[] split = line.split(" ::: "); // arbitrary, unlikely to occur in a useful regex easy to read
+      if (split.length != 2) {
+        throw new RuntimeException("The PatternRecognizerFilter: Always two there are, no more, no less, a pattern and a replacement (separated by ' ::: ' )");
```
danmuzi, line 108: Perhaps PatternTypingFilter is right instead of PatternRecognizerFilter.

danmuzi (commented):
Thanks, Gus.
I've left some minor comments!

lucene/analysis/common/src/java/org/apache/lucene/analysis/pattern/PatternTypingFilter.java
gus-asf: I think the problem with this is in the event of a non-match start() and end() and any other MatchResult method will throw exceptions... that sounds very expensive since matches would be rare. As for matching the whole input, I intend to leave that up to the writer of the regex. We're not replacing the text, just adding a type/flags. No reason to limit the user's options there I think.

lucene/analysis/common/src/java/org/apache/lucene/analysis/pattern/PatternTypingFilterFactory.java
```diff
@@ -0,0 +1,120 @@
...
+   */
+  public static final String NAME = "patternTyping";
+
+  private final String patternFile;
+  // todo, perhaps this could be turned into an FST (wanted: regex FST with capturing groups in java)
+  private final LinkedHashMap<Pattern, String> patterns = new LinkedHashMap<>();
+  private final Map<Pattern, Integer> flags = new HashMap<>();
+
+  /**
+   * Creates a new KeepWordFilterFactory
```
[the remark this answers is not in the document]
gus-asf: thx

lucene/analysis/common/src/java/org/apache/lucene/analysis/pattern/PatternTypingFilter.java
gus-asf: I guess maybe, on the idea that matches will be rare in most cases, it might be better to just let the find() happen a second time when a match does occur to avoid the string creation?
```
        if (matcher.matches()) {
          String replaced = matcher.replaceFirst(pattern);
          typeAtt.setType(replaced);
... etc
        }
```

lucene/analysis/common/src/java/org/apache/lucene/analysis/pattern/PatternTypingFilterFactory.java
```diff
@@ -0,0 +1,115 @@
...
+  /**
+   * SPI name
+   */
+  public static final String NAME = "patternTyping";
+
+  private final String patternFile;
+  private final LinkedHashMap<Pattern, Map.Entry<String, Integer>> replacementAndFlagByPattern = new LinkedHashMap<>();
+
+  /**
+   * Creates a new PatternTypingFilter
```
danmuzi, line 75: Thank you for checking my comments.
What do you think about PatternTypingFilterFactory?
It's changed from KeepWordFilterFactory to PatternTypingFilter.

gus-asf: oh darn thx :)

lucene/analysis/common/src/java/org/apache/lucene/analysis/pattern/PatternTypingFilter.java
gus-asf: @uschindler any further thoughts? If you agree with the above, I think all would be resolved.

uschindler: I think that's fine.

lucene/analysis/common/src/java/org/apache/lucene/analysis/pattern/PatternTypingFilter.java
```diff
@@ -0,0 +1,71 @@
...
+ * @since 8.8.0
+ */
+public class PatternTypingFilter extends TokenFilter {
+
+  private final Map<Pattern, Map.Entry<String, Integer>> replacementAndFlagByPattern;
+  private final CharTermAttribute termAtt = addAttribute(CharTermAttribute.class);
+  private final FlagsAttribute flagAtt = addAttribute(FlagsAttribute.class);
+  private final TypeAttribute typeAtt = addAttribute(TypeAttribute.class);
+
+  public PatternTypingFilter(TokenStream input, LinkedHashMap<Pattern, Map.Entry<String, Integer>> replacementAndFlagByPattern) {
```
uschindler, line 49: Explicitely saying LinkedHashMap sounds strange. I know the list must be sorted, so it's more a list.

I liked it use a <Pattern,String,Integer> record. This would be a classical exple of the new Java 6 record types!

This is a public API, any client code may call this - also non-Solr users. So maybe the constructor argument should be `a List<PatternTypingRule>` (a new class, which may actually be a Record in Java 16/17).

uschindler (changes requested):

lucene/analysis/common/src/java/org/apache/lucene/analysis/pattern/PatternTypingFilter.java
uschindler: FYI, I was not aware that the Map of Patterns and Elements are exposed as public API.

uschindler: If we have a class `PatternTypingRule` we can make constructors to create those: `new PatternTypeingRule(Pattern pattern, String replacement, int flag)` but also `new PatternTypingRule(String pattern, String replacement, int flag)`.

I would strogly prefer tono misuse maps

gus-asf: Updated. I agree that this is cleaner. I've started getting some time pressure on this due to end of month approaching so hopefully if you like what I've done we can get this committed. I think we've made good improvements and further refinements have time to be contemplated before 8.8. Getting it in will allow an updated patch for the primary AQP feature.

fsparv:
@uschindler can we commit this now?

gus-asf:
> @uschindler can we commit this now?

(that's me just wrong browser, so wrong account :) )

PR #1996, opened 2020-10-16
SOLR-14907: Adding V2 API for ConfigSet Upload.

HoustonPutman:
https://issues.apache.org/jira/browse/SOLR-14907

Adding v2 APIs.

### DIFF
--- a/solr/core/src/java/org/apache/solr/handler/ClusterAPI.java
+++ b/solr/core/src/java/org/apache/solr/handler/ClusterAPI.java
@@ -47,6 +47,7 @@ import org.apache.solr.response.SolrQueryResponse;
 import static org.apache.solr.client.solrj.SolrRequest.METHOD.DELETE;
 import static org.apache.solr.client.solrj.SolrRequest.METHOD.GET;
 import static org.apache.solr.client.solrj.SolrRequest.METHOD.POST;
+import static org.apache.solr.client.solrj.SolrRequest.METHOD.PUT;
 import static org.apache.solr.cloud.api.collections.OverseerCollectionMessageHandler.REQUESTID;
 import static org.apache.solr.common.params.CollectionParams.CollectionAction.ADDROLE;
 import static org.apache.solr.common.params.CollectionParams.CollectionAction.CLUSTERPROP;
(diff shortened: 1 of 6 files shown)


solr/core/src/java/org/apache/solr/handler/ClusterAPI.java
```diff
@@ -133,6 +134,62 @@ public void create(PayloadObj<CreateConfigInfo> obj) throws Exception {
...
+  public void uploadConfigSetWithOverwrite(SolrQueryRequest req, SolrQueryResponse rsp) throws Exception {
+    req = wrapParams(req,
+            "action", ConfigSetParams.ConfigSetAction.UPLOAD.toString(),
+            CommonParams.NAME, req.getPathTemplateValues().get("name"),
+            ConfigSetParams.OVERWRITE, true,
+            ConfigSetParams.CLEANUP, req.getParams().getBool(ConfigSetParams.CLEANUP, false));
+    configSetsHandler.handleRequestBody(req, rsp);
+  }
+
+  @EndPoint(method = POST,
```
noblepaul, line 161: Let's use a proper REST mechanism here

`path ="/cluster/configs/{configname}/*`

solr/core/src/java/org/apache/solr/handler/ClusterAPI.java
```diff
@@ -133,6 +134,62 @@ public void create(PayloadObj<CreateConfigInfo> obj) throws Exception {
...
+      throw new SolrException(SolrException.ErrorCode.BAD_REQUEST, "In order to insert a file into a configSet, a valid 'path' parameter must be provided.");
+    }
+    req = wrapParams(req,
+            "action", ConfigSetParams.ConfigSetAction.UPLOAD.toString(),
+            CommonParams.NAME, req.getPathTemplateValues().get("name"));
+    configSetsHandler.handleRequestBody(req, rsp);
+  }
+
+  @EndPoint(method = PUT,
+      path =   "/cluster/configs/{name}/file",
```
noblepaul, line 177: same as above

noblepaul (changes requested):

solr/core/src/java/org/apache/solr/handler/ClusterAPI.java
HoustonPutman: Would you be ok with `/cluster/configs/{configname}/file/*`? That's very restful, since it would be updating the `*` `file` under the `{configname}` `config`

HoustonPutman: I've updated the PR to reflect this new API.

noblepaul: I'm Ok with both

`/cluster/configs/{configname}/files/*`

and

`/cluster/configs/{configname}/*`


@janhoy @dsmiley @sigram  WDYT?

dsmiley: I don't see the point of the "/files/" component of the path.  The files within the configSet are directly in it, so I think it should be that way.  I think of this very much like a file in a folder, where the folder is the configSet.

Just checking that we support nested structure within the configSet?

HoustonPutman: We do support the nested structure.

The `/files/` part defines what the * after it is, and allows us to extend in the future. For example possible adding a `/cluster/configs/{configname}/properties` PUT action to update the properties for that config. It fits pretty well in the [guidelines](https://restfulapi.net/resource-naming/).

It's not that important to me though if y'all agree with taking it out.

noblepaul: >Just checking that we support nested structure within the configSet?

We could. But, I'm OK to add support for that later

dsmiley: To support properties manipulation in the future, I think that could be done via a POST to the particular path to the properties file in the configSet (I forget if it exists already or maybe should be exactly params.json) with commands in the JSON to delete/set/update etc.

HoustonPutman: Ok sounds good, that would certainly work. I'll take out the `/files/` part of the path, and update the PR and Jira.

HoustonPutman:
Are there any concerns now that the API has been sorted?

noblepaul:
LGTM

noblepaul (approved):

solr/core/src/java/org/apache/solr/handler/ClusterAPI.java
murblanc: I'm not a REST specialist but I think I would prefer to leave the `files` component of the path in. Wouldn't this allow changing other things in the future and makes the API more self describing?

Would it really be easy to add it later?
If we wanted to add properties for example, how to tell if `/cluster/configs/{configname}/properties/whatever` is a property `whatever` or if `properties/whatever` is actually a file?

murblanc: Also "file" is a pretty generic abstraction that can represent many things (a schema, properties, synonyms etc). We might eventually want to address such ressources by what they actually are (schema, properties, synonyms) rather than by the way in which they are stored (files).

HoustonPutman: I agree @murblanc, it lets us future-proof the API even if we don't have current plans on expanding beyond files.

murblanc: @dsmiley, can users of Solr pick random names for some ConfigSet files? I believe code will be easier to understand/maintain if we don't have to parse file names to detect if they are files or some other form of separator (and then the API would be inconsistent: files will not have its `/files` segment in the part whereas properties will be defined after `/properties`).

Examples in [https://restfulapi.net/resource-naming/](https://restfulapi.net/resource-naming/) would be a vote for keeping `/files` in the path IMO.

anshumg (approved):
I prefer @dsmiley 's suggestion here around removing the `files` bit from the path. The API in my opinion only provides convenience of uploading configsets/files and doesn't provide any abstractions around it.

PR #1998, opened 2020-10-17
LUCENE-9455: ExitableTermsEnum should sample timeout and interruption check before calling next()

zacharymorn:
# Description

ExitableTermsEnum should sample timeout and interruption check before calling next()

# Solution

Add the same sampling logic as in https://github.com/elastic/elasticsearch/blob/4af4eb99e18fdaadac879b1223e986227dd2ee71/server/src/main/java/org/elasticsearch/search/internal/ExitableDirectoryReader.java#L164-L168

# Tests
Updated unit test to test for sampling skipping check

# Checklist

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/ExitableDirectoryReader.java
+++ b/lucene/core/src/java/org/apache/lucene/index/ExitableDirectoryReader.java
@@ -496,7 +496,8 @@ public class ExitableDirectoryReader extends FilterDirectoryReader {
    * exitable enumeration of terms.
    */
   public static class ExitableTermsEnum extends FilterTermsEnum {
-    private static final int MAX_CALLS_BEFORE_QUERY_TIMEOUT_CHECK = (1 << 4) - 1; // 15
+    // Create bit mask in the form of 0000 1111 for efficient checking
+    private static final int NUM_CALLS_PER_TIMEOUT_CHECK = (1 << 4) - 1; // 15
     private int calls;
     private QueryTimeout queryTimeout;
     
@@ -504,15 +505,15 @@ public class ExitableDirectoryReader extends FilterDirectoryReader {
     public ExitableTermsEnum(TermsEnum termsEnum, QueryTimeout queryTimeout) {
       super(termsEnum);
       this.queryTimeout = queryTimeout;
-      checkAndThrowWithSampling();
+      checkTimeoutWithSampling();
     }
 
     /**
      * Throws {@link ExitingReaderException} if {@link QueryTimeout#shouldExit()} returns true,
      * or if {@link Thread#interrupted()} returns true.
      */
-    private void checkAndThrowWithSampling() {
-      if ((calls++ & MAX_CALLS_BEFORE_QUERY_TIMEOUT_CHECK) == 0) {
+    private void checkTimeoutWithSampling() {
+      if ((calls++ & NUM_CALLS_PER_TIMEOUT_CHECK) == 0) {
         if (queryTimeout.shouldExit()) {
           throw new ExitingReaderException("The request took too long to iterate over terms. Timeout: "
               + queryTimeout.toString()
@@ -527,7 +528,7 @@ public class ExitableDirectoryReader extends FilterDirectoryReader {
     @Override
     public BytesRef next() throws IOException {
       // Before every iteration, check if the iteration should exit
-      checkAndThrowWithSampling();
+      checkTimeoutWithSampling();
       return in.next();
     }
   }
--- a/lucene/core/src/java/org/apache/lucene/index/ExitableDirectoryReader.java
+++ b/lucene/core/src/java/org/apache/lucene/index/ExitableDirectoryReader.java
@@ -496,35 +496,38 @@ public class ExitableDirectoryReader extends FilterDirectoryReader {
    * exitable enumeration of terms.
    */
   public static class ExitableTermsEnum extends FilterTermsEnum {
-
+    private static final int MAX_CALLS_BEFORE_QUERY_TIMEOUT_CHECK = (1 << 4) - 1; // 15
+    private int calls;
     private QueryTimeout queryTimeout;
     
     /** Constructor **/
     public ExitableTermsEnum(TermsEnum termsEnum, QueryTimeout queryTimeout) {
       super(termsEnum);
       this.queryTimeout = queryTimeout;
-      checkAndThrow();
+      checkAndThrowWithSampling();
     }
 
     /**
      * Throws {@link ExitingReaderException} if {@link QueryTimeout#shouldExit()} returns true,
      * or if {@link Thread#interrupted()} returns true.
      */
-    private void checkAndThrow() {
-      if (queryTimeout.shouldExit()) {
-        throw new ExitingReaderException("The request took too long to iterate over terms. Timeout: " 
-            + queryTimeout.toString()
-            + ", TermsEnum=" + in
-        );
-      } else if (Thread.interrupted()) {
-        throw new ExitingReaderException("Interrupted while iterating over terms. TermsEnum=" + in);
+    private void checkAndThrowWithSampling() {
+      if ((calls++ & MAX_CALLS_BEFORE_QUERY_TIMEOUT_CHECK) == 0) {
+        if (queryTimeout.shouldExit()) {
+          throw new ExitingReaderException("The request took too long to iterate over terms. Timeout: "
+              + queryTimeout.toString()
+              + ", TermsEnum=" + in
+          );
+        } else if (Thread.interrupted()) {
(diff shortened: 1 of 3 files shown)


lucene/core/src/test/org/apache/lucene/index/TestExitableDirectoryReader.java
```diff
@@ -163,6 +163,17 @@ public void testExitableFilterTermsIndexReader() throws Exception {
...
+    // Set a fairly high timeout value (infinite) and expect the query to complete in that time frame.
+    // Not checking the validity of the result, but checking the sampling kicks in to reduce the number of timeout check
+    CountingQueryTimeout queryTimeout = new CountingQueryTimeout();
+    directoryReader = DirectoryReader.open(directory);
+    exitableDirectoryReader = new ExitableDirectoryReader(directoryReader, queryTimeout);
+    reader = new TestReader(getOnlyLeafReader(exitableDirectoryReader));
+    searcher = new IndexSearcher(reader);
+    searcher.search(query, 10);
+    reader.close();
+    assertEquals(3, queryTimeout.getShouldExitCallCount());
```
bruno-roustant, line 175: Here we compare to the expected call count 3. Because 3 TermsEnum are created: it is a PrefixQuery and there is one TermsEnum created for AutomatonQuery.intersect() (the next call timeout check is skipped once), then 2 TermsEnum created for the 2 matching terms "one" and "ones").

Would it be clearer to have a separate test method?
We could index more docs, for example 50 with a prefix-generated term (e.g. "term"+increment).
There would be a PrefixQuery for "term", the same test code, and we would test the TIMEOUT_CHECK_SAMPLING:
- 1 TermsEnum for Automaton.intersect(), and the next calls would be sampled (50/TIMEOUT_CHECK_SAMPLING=3) => call count +4
- 1 TermsEnum for each enumerated term => call count +50
We could verify call count = 54 (instead of 101 without sampling)

bruno-roustant (changes requested):

bruno-roustant:
Overall I wonder if we can do better with the sampling. The goal is to avoid doing numerous repetitive calls to QueryTimeout.shouldExit(). This is essentially the case for multi-terms queries. But actually for multi-terms queries, a new TermsEnum is created for each matching term (to get doc ids). So we end up only sampling half of the calls to QueryTimeout.shouldExit() since the other half is done by the ExitableTermsEnum constructor which is not sampled.
It would be better to also sample the ExitableTermsEnum constructor, but I don't know yet how to do that.

bruno-roustant:
I have an idea to sample the timeout check in ExitableTermsEnum constructor. Obviously not based on the counter which is always 0 at that point, but rather based on pseudo randomness.
We could check if ((System.identityHashCode(this) & TIMEOUT_CHECK_SAMPLING) == 0) before calling checkTimeoutWithSampling().

lucene/core/src/java/org/apache/lucene/index/ExitableDirectoryReader.java
```diff
@@ -496,35 +496,38 @@ public TermsEnum iterator() throws IOException {
    * exitable enumeration of terms.
    */
   public static class ExitableTermsEnum extends FilterTermsEnum {
-
+    private static final int MAX_CALLS_BEFORE_QUERY_TIMEOUT_CHECK = (1 << 4) - 1; // 15
```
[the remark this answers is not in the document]
zacharymorn: Renamed to `NUM_CALLS_PER_TIMEOUT_CHECK`

For 
> this value must be a power of two

Which value are you referring to, as "15" is not power of two?  I've added this comment to clarify the purpose of computation:
```
// Create bit mask in the form of 0000 1111 for efficient checking 
``` 
What do you think?

lucene/core/src/java/org/apache/lucene/index/ExitableDirectoryReader.java
```diff
@@ -496,35 +496,38 @@ public TermsEnum iterator() throws IOException {
...
      */
-    private void checkAndThrow() {
-      if (queryTimeout.shouldExit()) {
-        throw new ExitingReaderException("The request took too long to iterate over terms. Timeout: " 
-            + queryTimeout.toString()
-            + ", TermsEnum=" + in
-        );
-      } else if (Thread.interrupted()) {
-        throw new ExitingReaderException("Interrupted while iterating over terms. TermsEnum=" + in);
+    private void checkAndThrowWithSampling() {
```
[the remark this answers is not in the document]
zacharymorn: Done.

lucene/core/src/test/org/apache/lucene/index/TestExitableDirectoryReader.java
zacharymorn: > Because 3 TermsEnum are created: it is a PrefixQuery and there is one TermsEnum created for AutomatonQuery.intersect() (the next call timeout check is skipped once), then 2 TermsEnum created for the 2 matching terms "one" and "ones").

Yes this is exactly what happened. Without the check the call count will be 4. 

------
I've tried to come up with a new test like what you suggested:
```
  /**
   * Tests time out check sampling of TermsEnum iterations
   * @throws Exception on error
   */
  public void testExitableTermsEnumSampleTimeoutCheck() throws Exception {
    try(Directory directory = newDirectory()) {
      try(IndexWriter writer = new IndexWriter(directory, newIndexWriterConfig(new MockAnalyzer(random())))) {
        for(int i = 0; i < 50; i++) {
          Document d1 = new Document();
          d1.add(newTextField("default", "term" + i, Field.Store.YES));
          writer.addDocument(d1);
        }

        writer.forceMerge(1);
        writer.commit();

        DirectoryReader directoryReader;
        DirectoryReader exitableDirectoryReader;
        IndexReader reader;
        IndexSearcher searcher;

        Query query = new PrefixQuery(new Term("default", "term"));

        // Set a fairly high timeout value (infinite) and expect the query to complete in that time frame.
        // Not checking the validity of the result, but checking the sampling kicks in to reduce the number of timeout check
        CountingQueryTimeout queryTimeout = new CountingQueryTimeout();
        directoryReader = DirectoryReader.open(directory);
        exitableDirectoryReader = new ExitableDirectoryReader(directoryReader, queryTimeout);
        reader = new TestReader(getOnlyLeafReader(exitableDirectoryReader));
        searcher = new IndexSearcher(reader);
        searcher.search(query, 300);
        reader.close();
        assertEquals(54, queryTimeout.getShouldExitCallCount());
      }
    }
  }
``` 
This test actually failed as the actual call count is 5 instead of 54. After some debugging it turns out that the threshold setting here https://github.com/apache/lucene-solr/blob/67ecd8ff9ad3640016424ded86bfaaadd815b96d/lucene/core/src/java/org/apache/lucene/search/MultiTermQueryConstantScoreWrapper.java#L118 also help reduce the query timeout check call count. If I were to change the threshold to 100, then the test will pass.

lucene/core/src/java/org/apache/lucene/index/ExitableDirectoryReader.java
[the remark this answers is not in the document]
bruno-roustant: Sounds good to me. Indeed I wrote too fast, I meant power of two mask, which is (power-of-two - 1).

lucene/core/src/test/org/apache/lucene/index/TestExitableDirectoryReader.java
bruno-roustant: This test is good. Can you actually add it to the PR? It should replace the lines added to testExitableFilterTermsIndexReader().

Your investigation reveals something I overlooked - the MultiTerm query is not rewritten to TermQueries above 16 matching terms.
- Just replace assertEquals(54, ...) by assertEquals(5, ...).
- Please add a comment above the assertEquals(5, ...) to explain it depends on ExitableDirectoryReader.ExitableTermsEnum.NUM_CALLS_PER_TIMEOUT_CHECK and MultiTermQueryConstantScoreWrapper.BOOLEAN_REWRITE_TERM_COUNT_THRESHOLD (so if someone changes one of these constants - unlikely -, she will know why this test fail and that it needs to be adjusted).
- I remove my remark about sampling ExitableTermsEnum constructor because we don't need it.

bruno-roustant:
So thanks to MultiTermQueryConstantScoreWrapper.BOOLEAN_REWRITE_TERM_COUNT_THRESHOLD we actually don't need to sample ExitableTermsEnum constructor.

lucene/core/src/test/org/apache/lucene/index/TestExitableDirectoryReader.java
zacharymorn: Sounds good. I've added another commit to separate out the test, and adjusted assertion and added comments.

bruno-roustant (approved):

PR #1999, opened 2020-10-17
SOLR-14943 Rework Monitoring Solr with Prometheus and Grafana ref guide to be clearer

epugh:
# Description

Update the ref guide based on going through the documentation and setting it up for the first time myself.

# Solution

Some fixes to example code, some transitions for folks like myself who don't know about these tools already!


epugh:
I'd love some clarity on if this contrib is really experimental or not.  If not, I'd love to remove that warning!   Everything in Solr is subject to change over time ;-)

janhoy (approved):
Thanks for cleaning up and making it more beginner friendly.

We could perhaps also supply a docker-compose file which sets up everything for easy testing? But that’s for another jira.

PR #2000, opened 2020-10-17
LUCENE-9318: Clean up package name conflicts for backward-codecs

mocobeta:
This resolves split packages between core and backward-codecs.

- `o.a.l.codecs` in backward-codecs is renamed to `o.a.l.backward_codecs`.
- `o.a.l.index` in backward-codecs (only exists in src/test) is renamed to `o.a.l.backward_index`.
- packages in lucene-core are unchanged.

Passes test and precommit.


lucene/core/src/java/org/apache/lucene/index/IndexUpgrader.java
```diff
@@ -77,9 +77,10 @@ private static void printUsage() {
   public static void main(String[] args) throws IOException {
     parseArgs(args).upgrade();
   }
-  
+
+  /** Parse arguments. */
   @SuppressForbidden(reason = "System.out required: command line tool")
-  static IndexUpgrader parseArgs(String[] args) throws IOException {
+  public static IndexUpgrader parseArgs(String[] args) throws IOException {
```
mocobeta, line 83: This have to be public because the renamed o.a.l.backward_index.TestBackwardsCompatibility refers this.

msokolov (approved):
Hi Tomoko - thanks for taking care of this. The approach looks sound to me. I did not review every file! But I saw where you renamed the classes in the META-INF/...codecs, so that should work (and I assume that tests bore it out) since we store the service name in the index, not the classname. 

It's nice that you were able to fix the conflicts among javadocs. Should there be a followup issue about the codec javadocs I wonder? Today we have a weird situation where the latest javadoc keeps moving from package to package. I wonder if we could move it to the codecs/ package. Also, I don't really understand why we have these package.html files rather than package-info.java - is it because of these cross-module package conflicts? Anyway this commit should stay clean refactor/renaming, 

So +1

rmuir:
{quote}
Also, I don't really understand why we have these package.html files rather than package-info.java - is it because of these cross-module package conflicts?
{quote}

That's correct. We can clean these up later after the package conflicts are fixed.

lucene/core/src/java/org/apache/lucene/index/IndexUpgrader.java
rmuir: Should the test maybe just call `IndexUpgader.main(args)` instead of `upgrader = parseArgs(args) ... upgrader.upgrade()?` @uschindler ?

uschindler: I don't know the code. But maybe indexupgrader can just provide an API to call it without first creating some arguments. This would also make it easier to use for 3rd party code.

mocobeta: It should be possible to just call `main()` instead of calling parseArgs() and upgrade() separately. Can we omit the assertion error regarding to `parseArgs()`?

```
        /*
        IndexUpgrader upgrader = null;
        try {
          upgrader = IndexUpgrader.parseArgs(args.toArray(new String[0]));
        } catch (Exception e) {
          throw new AssertionError("unable to parse args: " + args, e);
        }
        upgrader.upgrade();
         */
        // this could throw IOException. 
        // we cannot identify whether the error occurs in parseArgs() or upgrade()
        IndexUpgrader.main(args.toArray(new String[0]));
```

uschindler: I think so.

mocobeta: fixed the test and reverted the change for IndexUpgrader. https://github.com/apache/lucene-solr/pull/2000/commits/2b9c1bfbd7ca589a297813938dbbe96472e27ad0

mocobeta:
Thanks guys for review.
I merged the latest master and resolved conflicts (had a bit trouble though...)

I have checked the changes again whether there is no unexpected modifications. I'll merge it soon after the checks by Github action are done. If any follow-ups are needed, I'll catch them later.

msokolov:
> I merged the latest master and resolved conflicts (had a bit trouble though...)
Woops! sorry if the new codec and related changes got in the way ;) Looks as if you overcame it

PR #2002, opened 2020-10-18
SOLR Mark only @deprecated tag, not @Deprecated annotation

dsmiley:
For DIH, HDFS, Velocity.
These [will] *move* outside the Solr project, and are not disappearing from a user's standpoint.  Use of Deprecated annotation on a plugin will cause a warning on startup, which will needlessly alarm users for these plugins.

See https://issues.apache.org/jira/browse/SOLR-12987?page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel&focusedCommentId=17203524#comment-17203524
And raised recently on the dev list.


epugh:
Would it be possible to have those depredations have a link to the NEW home of these projects?   So the person who sees the tag knows where to go to?

dsmiley:
@epugh I modified to include a link to https://cwiki.apache.org/confluence/display/SOLR/Deprecations which I think is easier to maintain and sufficient

dsmiley:
@chatman are you okay with what I'm doing here?

PR #2004, opened 2020-10-19
SOLR-14942: Reduce leader election time on node shutdown

shalinmangar:
This PR is based on the work of @CaoManhDat 

# Description

The shutdown process waits for all replicas/cores to be closed before removing the election node of the leader. This can take some time due to index flush or merge activities on the leader cores and delays new leaders from being elected. This leads to long gap of time during which indexing requests fail.

# Solution

This PR introduces a Phaser that registers and deregisters all update requests. The shutdown process pauses indexing requests, waits for all in-flight indexing requests to complete, removes election nodes (thus triggering leader election) and then closes all replicas.

# Tests

Since this PR makes changes to Solr's shutdown process, it is difficult to write a unit/integration tests inside Solr itself. This was tested by creating a test harness outside of Solr. The steps were:
- Create a Collection with 1 shard, 3 replicas
- Do heavy indexing when all replicas are ACTIVE
- Shut down the leader node
- Measure the leaderless time (time during which there is no live/active leader)

Before the patch | After the patch
-- | --
Shutting down time: 100ms Leader election: 46s | Shutting down time: 12s Leader election: 3s
Shutting down time: 37s Leader election: 3s | Shutting down time: 31s Leader election: 3s
Shutting down time: 200ms Leader election: 17s | Shutting down time: 100ms Leader election: 3s
Shutting down time: 200ms Leader election: 27s | Shutting down time: 4s Leader election: 3s

Since the patch introduces waiting for all updates to finish before giving up the leadership so the shut down time gets increased on average, but also it makes leader election more stable and improves consistency.

# Checklist

### DIFF
--- a/solr/core/src/java/org/apache/solr/cloud/ZkController.java
+++ b/solr/core/src/java/org/apache/solr/cloud/ZkController.java
@@ -1838,6 +1838,22 @@ public class ZkController implements Closeable {
     }
   }
 
+  public void tryCancelAllElections() {
+    if (zkClient.isClosed()) {
+      return;
+    }
+    electionContexts.values().parallelStream().forEach(context -> {
+      try {
+        context.cancelElection();
+        context.close();
+      } catch (InterruptedException e) {
+        Thread.currentThread().interrupt();
+      } catch (KeeperException e) {
+        log.error("Error on cancelling elections of {}", context.leaderPath, e);
+      }
+    });
+  }
+
   private ZkCoreNodeProps waitForLeaderToSeeDownState(
       CoreDescriptor descriptor, final String coreZkNodeName) throws SessionExpiredException {
     // try not to wait too long here - if we are waiting too long, we should probably
--- a/solr/core/src/java/org/apache/solr/cloud/ZkController.java
+++ b/solr/core/src/java/org/apache/solr/cloud/ZkController.java
@@ -653,7 +653,12 @@ public class ZkController implements Closeable {
     customThreadPool.submit(() -> Collections.singleton(overseer).parallelStream().forEach(IOUtils::closeQuietly));
 
     try {
-      customThreadPool.submit(() -> electionContexts.values().parallelStream().forEach(IOUtils::closeQuietly));
+      customThreadPool.submit(() -> {
+        Collection<ElectionContext> values = electionContexts.values();
+        synchronized (electionContexts) {
+          values.forEach(IOUtils::closeQuietly);
+        }
+      });
 
     } finally {
 
@@ -1838,20 +1843,26 @@ public class ZkController implements Closeable {
     }
   }
 
+  /**
+   * Attempts to cancel all leader elections. This method should be called on node shutdown.
+   */
   public void tryCancelAllElections() {
     if (zkClient.isClosed()) {
       return;
     }
-    electionContexts.values().parallelStream().forEach(context -> {
-      try {
-        context.cancelElection();
-        context.close();
-      } catch (InterruptedException e) {
-        Thread.currentThread().interrupt();
-      } catch (KeeperException e) {
-        log.error("Error on cancelling elections of {}", context.leaderPath, e);
-      }
-    });
+    Collection<ElectionContext> values = electionContexts.values();
+    synchronized (electionContexts) {
+      values.forEach(context -> {
+        try {
+          context.cancelElection();
+          context.close();
+        } catch (InterruptedException e) {
+          Thread.currentThread().interrupt();
+        } catch (KeeperException e) {
+          log.warn("Error on cancelling elections of {}", context.leaderPath, e);
+        }
+      });
+    }
   }
 
   private ZkCoreNodeProps waitForLeaderToSeeDownState(
--- a/solr/core/src/java/org/apache/solr/cloud/ZkController.java
+++ b/solr/core/src/java/org/apache/solr/cloud/ZkController.java
@@ -648,9 +648,9 @@ public class ZkController implements Closeable {
 
     ExecutorService customThreadPool = ExecutorUtil.newMDCAwareCachedThreadPool(new SolrNamedThreadFactory("closeThreadPool"));
 
-    customThreadPool.submit(() -> Collections.singleton(overseerElector.getContext()).parallelStream().forEach(IOUtils::closeQuietly));
+    customThreadPool.submit(() -> IOUtils.closeQuietly(overseerElector.getContext()));
 
-    customThreadPool.submit(() -> Collections.singleton(overseer).parallelStream().forEach(IOUtils::closeQuietly));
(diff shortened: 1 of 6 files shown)


CaoManhDat:
+1 LGTM!

CaoManhDat (approved):

solr/core/src/java/org/apache/solr/cloud/ZkController.java
```diff
@@ -1838,6 +1838,22 @@ private void checkStateInZk(CoreDescriptor cd) throws InterruptedException, NotI
     }
   }
 
+  public void tryCancelAllElections() {
+    if (zkClient.isClosed()) {
+      return;
+    }
+    electionContexts.values().parallelStream().forEach(context -> {
```
madrob, line 1845: How many contexts do we expect here. Does it actually make sense to do a parallel stream?

solr/core/src/java/org/apache/solr/update/SolrCoreState.java
```diff
@@ -86,7 +103,45 @@ public boolean decrefSolrCoreState(IndexWriterCloser closer) {
...
+    if (pauseUpdateRequests.compareAndSet(false, true)) {
+      inflightUpdatesCounter.register();
+      inflightUpdatesCounter.arriveAndAwaitAdvance();
+    }
+  }
+
+  /**
+   * Unpauses update requests to this core
+   */
+  public void unpauseUpdates() {
```
madrob, line 121: This is never used?

solr/core/src/java/org/apache/solr/update/SolrCoreState.java
```diff
@@ -86,7 +103,45 @@ public boolean decrefSolrCoreState(IndexWriterCloser closer) {
...
+  /**
+   * Registers in-flight update requests to this core.
+   *
+   * @return true if request was registered, false if update requests are paused
+   */
+  public boolean registerInFlightUpdate() {
+    if (pauseUpdateRequests.get()) {
+      return false;
+    }
+    inflightUpdatesCounter.register();
```
madrob, line 134: Do we need to check for termination?

solr/core/src/java/org/apache/solr/update/SolrCoreState.java
```diff
@@ -86,7 +103,45 @@ public boolean decrefSolrCoreState(IndexWriterCloser closer) {
...
-  
+
+  /**
+   * Pauses all update requests to this core and waits (indefinitely) for all in-flight
+   * update requests to finish
+   */
+  public void pauseUpdatesAndAwaitInflightRequests() {
+    if (pauseUpdateRequests.compareAndSet(false, true)) {
+      inflightUpdatesCounter.register();
+      inflightUpdatesCounter.arriveAndAwaitAdvance();
```
madrob, line 114: This is in the shutdown path, should we add a timeout? What happens if a disk is slow on write and something else times out underneath us?

solr/core/src/java/org/apache/solr/cloud/ZkController.java
shalinmangar: No, it does not. Parallelstream was being used for close in ZkController.close() as well so this was basically copied code. However I checked and it is a lightweight operation. It basically removes a zk node and sets a volatile member. I have replaced this call with a serial forEach in both places.

@hossman pointed out that this iteration wasn't safe so I have put the forEach inside a synchronized block

solr/core/src/java/org/apache/solr/update/SolrCoreState.java
shalinmangar: Removed

solr/core/src/java/org/apache/solr/update/SolrCoreState.java
shalinmangar: According to the javadocs, register() is a no-op if phaser has terminated so we don't need to worry about it.

solr/core/src/java/org/apache/solr/update/SolrCoreState.java
shalinmangar: Good point. I have added a small timeout (2500 ms) with the same value as the one used inside RecoveryStrategy for leader to finish in-flight requests. I think this default should be enough but it can be changed using a sysprop if needed.

solr/core/src/java/org/apache/solr/update/SolrCoreState.java
madrob: Should we return an error in this case? Does termination imply that the core is closed?

solr/core/src/java/org/apache/solr/core/CoreContainer.java
```diff
@@ -38,10 +38,7 @@
 import java.util.Properties;
 import java.util.Set;
 import java.util.UUID;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ExecutionException;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Future;
+import java.util.concurrent.*;
```
madrob, line 41: Please check your IDE settings to not collapse imports like this.

solr/core/src/java/org/apache/solr/update/SolrCoreState.java
madrob: Maybe `return inflightUpdatesCounter.register() > 0`? Or `>=`, I'm not sure.

solr/core/src/java/org/apache/solr/cloud/ZkController.java
```diff
@@ -653,7 +653,12 @@ public void close() {
     customThreadPool.submit(() -> Collections.singleton(overseer).parallelStream().forEach(IOUtils::closeQuietly));
```
madrob, line 653: While we're here, this line (and a few others) should be `customThreadPool.submit(() -> IOUtils.closeQuietly(overseer);` I have no idea why we're creating a collection and a stream for a single object.

shalinmangar: I have cleaned it up

solr/core/src/java/org/apache/solr/update/SolrCoreState.java
shalinmangar: Thanks for bringing my attention to termination. The usage of the phaser was wrong before because it would get terminated the moment there were no in-flight requests. We actually never need to terminate the phaser at all. Now, we override onAdvance() and disable termination such that the phase keeps advancing and wraps around after reaching 65536.

shalinmangar:
@madrob would you like to make a final pass at this PR? I wish to merge to master and back port to 8x today.

madrob (approved):

madrob:
Please add a CHANGES entry, and credit @CaoManhDat as well, if this work was based on initial work done by him.

PR #2005, opened 2020-10-19
Add example for ConfigSet create with properties map.

HoustonPutman:
The previous example in the documentation only used the less-user-friendly `configSetProp.<name>` option.


noblepaul (approved):
LGTM

PR #2010, opened 2020-10-20
SOLR-12182: Don't persist base_url in ZK as the scheme is variable, compute from node_name instead

thelabdude:
# Description

See JIRA for description of the issue: https://issues.apache.org/jira/browse/SOLR-12182

# Solution

This PR computes the `base_url` for a Replica using the stored `node_name` and a global `urlScheme` rather than storing the `base_url` in `state.json`. This avoids storing an incorrect URL scheme for replicas in persistent storage. The `base_url` is computed when read back from ZK and dropped when marshaling the Replica state to JSON. This also means we don't need a migration tool as stored state is "healed" on-the-fly when read back from ZK.

The unfortunate aspect of this PR is we need to keep the URL scheme for the cluster in a global variable (so that it is available when reading from ZK). The global `urlScheme` still comes from the cluster property but is then stored in a global singleton, see: `org.apache.solr.common.cloud.UrlScheme`. Alternatively, we could just keep the `urlScheme` in a static in ZkStateReader, I felt the global singleton `UrlScheme.INSTANCE` made it clearer that this was a global thing but it also made more sense with my first implementation that tried to make rolling restart upgrades to TLS less chaotic. It's a trivial change to move all this over to ZkStateReader and remove UrlScheme.

I initially tried setting a `ThreadLocal` that gives access to the `urlScheme` whenever we need to read these props from ZK. However, that ended up being problematic because we tend to read ZkNodeProps from ZK in many places. In reality, the `urlScheme` really is an immutable global variable that should be set once during initialization by reading from the cluster property stored in ZK. So I felt trying to treat this global as something that was highly dynamic made the code overly cumbersome. Put simply, we shouldn't support `urlScheme` changing in a live node after initialization, it's bad for business.

I also tried to get rid of the `urlScheme` cluster property (re: https://issues.apache.org/jira/browse/SOLR-10202) but I'm not sure how SolrCloud client applications can resolve the correct `urlScheme` for the cluster without this property? On the server-side, sure we can just get the `urlScheme` from a Java System Property, but that won't be set for remote client applications that initialize via a connection to ZooKeeper. So I'm keeping the cluster property `urlScheme` for now.

We also need to consider how to enable TLS on an existing cluster (with active collections) using a rolling restart process. The current `org.apache.solr.cloud.SSLMigrationTest` just stopped all test nodes at once and then brought them back with TLS enabled.

Based on feedback, I've since removed the option to pull the active urlScheme from live nodes as we're not able to ensure zero-downtime when moving from `http` -> `https` for clusters with existing collections and live traffic. Put simply, the feature was a bit trappy in that it tried to reduce chaos when doing a rolling restart to enable TLS, but it made no guarantees. Thus, users just need to be sure to enable TLS before building production clusters!

Lastly, I've tried to clean-up some of the places that access the baseUrl on replicas to be more consistent, so you'll see some of that in this PR as well.

# Tests

Many existing tests cover regression caused by these code changes. Added simple unit test for UrlScheme.

# Checklist


solr/core/src/java/org/apache/solr/core/ZkContainer.java
```diff
@@ -31,6 +31,7 @@
 import java.util.function.Supplier;
 
 import org.apache.commons.lang3.StringUtils;
+import org.apache.solr.common.cloud.GlobalStateVars;
```
madrob, line 34: nit: alphabetize imports

solr/core/src/test/org/apache/solr/cloud/SSLMigrationTest.java
```diff
@@ -63,7 +61,7 @@ public void test() throws Exception {
     testMigrateSSL(new SSLTestConfig(false, false));
   }
   
-  public void testMigrateSSL(SSLTestConfig sslConfig) throws Exception {
+  private void testMigrateSSL(SSLTestConfig sslConfig) throws Exception {
```
madrob, line 64: ?

solr/solrj/src/java/org/apache/solr/common/cloud/GlobalStateVars.java
```diff
@@ -0,0 +1,79 @@
...
+ * really should be treated like a static global that is set at initialization and not altered after.
+ */
+public class GlobalStateVars {
+
+  public static final String HTTP = "http";
+  public static final String HTTPS = "https";
+  public static final String SCHEME_VAR = "${scheme}://";
+  public static final String HTTPS_PORT_PROP = "solr.jetty.https.port";
+
+  private static final GlobalStateVars _singleton = new GlobalStateVars();
```
madrob, line 32: I prefer the enum singleton pattern

solr/solrj/src/java/org/apache/solr/common/cloud/GlobalStateVars.java
```diff
@@ -0,0 +1,79 @@
...
+      throw new IllegalArgumentException("Invalid urlScheme: "+urlScheme);
+    }
+  }
+
+  /**
+   * Given a URL with a replaceable parameter for the scheme, return a new URL with the correct scheme applied.
+   * @param url A URL to change the scheme (http|https)
+   * @return A new URL with the correct scheme or null if the supplied url remains unchanged.
+   */
+  public String applyUrlSchemeIfChanged(final String url) {
```
madrob, line 62: null on unchanged seems like a trap to me

solr/solrj/src/java/org/apache/solr/common/cloud/ZkCoreNodeProps.java
```diff
@@ -20,6 +20,9 @@
   private ZkNodeProps nodeProps;
   
   public ZkCoreNodeProps(ZkNodeProps nodeProps) {
+    if (nodeProps == null) {
```
madrob, line 23: Use Objects.requireNonNull

solr/solrj/src/java/org/apache/solr/common/cloud/ZkNodeProps.java
```diff
@@ -105,6 +111,17 @@ public static ZkNodeProps load(byte[] bytes) {
...
   @Override
   public void write(JSONWriter jsonWriter) {
+    final String baseUrl = (String)propMap.get(ZkStateReader.BASE_URL_PROP);
+    if (baseUrl != null && !baseUrl.startsWith(SCHEME_VAR)) {
+      final int at = baseUrl.indexOf("://");
+      if (at != -1) {
+        final String updatedUrl = SCHEME_VAR + baseUrl.substring(at+3);
+        Map<String,Object> modMap = new LinkedHashMap<>(propMap);
+        modMap.put(ZkStateReader.BASE_URL_PROP, updatedUrl);
+        jsonWriter.write(modMap);
```
madrob, line 121: I think it's clearer here if we only have a single return from the method, so we should fall through or have an else.

solr/core/src/java/org/apache/solr/core/ZkContainer.java
```diff
@@ -263,4 +275,27 @@ public void close() {
...
+    final String urlScheme = zkCtrl.getZkStateReader().getClusterProperty(ZkStateReader.URL_SCHEME, null);
+    if (StringUtils.isNotEmpty(urlScheme)) {
+      // track the urlScheme in a global so we can use it during ZK read / write operations for cluster state objects
+      GlobalStateVars.singleton().setUrlScheme(urlScheme);
+    } else {
+      final String urlSchemeFromSysProp = System.getProperty(URL_SCHEME, HTTP);
+      if (HTTPS.equals(urlSchemeFromSysProp)) {
+        // it's OK to set the cluster prop to https b/c we're making the cluster more secure, not less
+        // wouldn't do this for http though
+        log.warn("Cluster property 'urlScheme' not set but system property is set to 'https'. Updating the cluster property to match.");
```
madrob, line 290: If we have an initially HTTP cluster, and then one node comes up with HTTPS, then that will bring down the rest of the cluster, right? I don't think we can do this rolling, however.

solr/solrj/src/java/org/apache/solr/common/cloud/GlobalStateVars.java
```diff
@@ -0,0 +1,79 @@
...
+ */
+package org.apache.solr.common.cloud;
+
+import static org.apache.solr.common.cloud.ZkStateReader.URL_SCHEME;
+
+/**
+ * Singleton access to global vars in persisted state, such as the urlScheme, which although is stored in ZK as a cluster property
+ * really should be treated like a static global that is set at initialization and not altered after.
+ */
+public class GlobalStateVars {
```
madrob, line 25: I don't think we want to encourage further use of this pattern. Let's call it something more specific like `UrlSchemeState`

madrob (commented):
Please update the ref guide and add a CHANGES entry.

solr/solrj/src/java/org/apache/solr/common/cloud/GlobalStateVars.java
thelabdude: yeah, it's ugly for sure, but wanted to capture the fact that the url didn't change ... how about Optional instead?

solr/core/src/test/org/apache/solr/cloud/SSLMigrationTest.java
thelabdude: it's an internal test method and doesn't need to be public but I'll rename it to make it clear

solr/solrj/src/java/org/apache/solr/common/cloud/GlobalStateVars.java
thelabdude: ah yes, good call, thanks

solr/core/src/java/org/apache/solr/core/ZkContainer.java
thelabdude: I'll just remove this, not strictly necessary (was trying to save the user some work but could have side-effects)

solr/solrj/src/java/org/apache/solr/common/cloud/GlobalStateVars.java
thelabdude: I agree, thanks for confirming ;-)

thelabdude:
Thanks for the quick turn-around with a review @madrob  ... will update CHANGES.txt and ref guide next

murblanc:
Why store `${scheme}://` in Zookeeper? We could store the rest of the URL (`host:port/path`) given we'll rebuild the URL from configuration available elsewhere.
For existing URLs with a scheme we could just remove/ignore the scheme and use the configured one instead.

thelabdude:
@murblanc right, using `${scheme}://` doesn't matter at all except for making it clear that gets replaced at runtime, at least to me, it makes it clear that is a variable that gets replaced. If the consensus is we don't want it there, that's fine too, as it really doesn't matter for the impl.

thelabdude:
Switched this to draft as we need a better solution for migrating from using `http` -> `https` using a rolling restart where some nodes will still be using `http` and others `https` (temporarily). Currently, switching the urlScheme over to https globally before all nodes have restarted could lead to unnecessary thrashing and cause downtime as leader elections would be effected while the rolling restart is progressing.

thelabdude:
I think this is ready for review again ... still have to create CHANGES.txt entry and update the ref guide but let's get through another round of reviews, esp. around the approach I'm taking here. Once we're in agreement on the overall design, I'll update the docs.

anshumg (approved):
This looks great!

I plan to manually test this out with some custom setups soon and will update with my findings.

I'm not too sure about the option for rolling restart during migration of urlScheme, but this seems like a reasonable middle path.

anshumg (commented):
Accidentally pressed the 're request review' button.

anshumg:
We should restrict this PR to just handle the removal of persisted base_url in ZK.

For users who intend to enable/disable TLS for their cluster, they should plan for downtime and as guaranteeing a safe rolling restart during such a transition is not possible right now. 

I think we're better off not trying for a best effort during a rolling restart here and removing the live_nodes part of the code change to keep things simple.

solr/core/src/java/org/apache/solr/cloud/ZkController.java
```diff
@@ -1401,8 +1420,7 @@ public ZkCoreNodeProps getLeaderProps(final String collection,
         byte[] data = zkClient.getData(
             ZkStateReader.getShardLeadersPath(collection, slice), null, null,
             true);
-        ZkCoreNodeProps leaderProps = new ZkCoreNodeProps(
-            ZkNodeProps.load(data));
+        ZkCoreNodeProps leaderProps = new ZkCoreNodeProps(ZkNodeProps.load(data));
```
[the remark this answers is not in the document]
thelabdude: that'll teach me to fix weird whitespacing!

solr/core/src/java/org/apache/solr/cloud/ZkController.java
```diff
@@ -2131,12 +2128,7 @@ public LeaderElector getOverseerElector() {
   static String generateNodeName(final String hostName,
                                  final String hostPort,
                                  final String hostContext) {
-    try {
-      return hostName + ':' + hostPort + '_' +
-          URLEncoder.encode(trimLeadingAndTrailingSlashes(hostContext), "UTF-8");
-    } catch (UnsupportedEncodingException e) {
-      throw new Error("JVM Does not seem to support UTF-8", e);
-    }
+    return hostName + ':' + hostPort + '_' + URLEncoder.encode(trimLeadingAndTrailingSlashes(hostContext), StandardCharsets.UTF_8);
```
anshumg, line 2131: Java8 doesn't have `URLEncoder.encode(String, Charset)`. Also, it'd through the exception on 8x.
I like it this way, just saying that you might have to remember that when merging.

solr/core/src/test/org/apache/solr/cloud/LeaderElectionTest.java
```diff
@@ -414,6 +425,11 @@ private void waitForLeader(List<ClientThread> threads, int seq)
 
   private int getLeaderThread() throws KeeperException, InterruptedException {
     String leaderUrl = getLeaderUrl("collection1", "shard1");
+    // strip off the scheme
+    final int at = leaderUrl.indexOf("://");
+    if (at != -1) {
+      leaderUrl = leaderUrl.substring(at+3);
```
anshumg, line 431: Can you please add a space around `+`.

anshumg (approved):
Just a couple of minor things but looks good to me!

noblepaul:
Before I go further, can I know

* Does this work for a rolling restart?
* How does one migrate to this?

thelabdude:
Hi @noblepaul thanks for taking a look ... so we decided to not try to handle the upgrade to TLS using a rolling restart (as described in my comment ^) in this PR ... also see Anshum's comments. I initially had a way to get the "current" urlScheme for each live node as with a rolling restart, you'll have a mix of nodes with TLS enabled and some not yet, but we felt that could be a little trappy b/c it really doesn't address the client applications. So our advice now is just suck it up and take the down time to enable TLS. Basically, we don't want to promise the community a zero-downtime upgrade to enable TLS, because it is a hard thing to promise. The live nodes approach (see commit history in this PR) works on the server side, but doesn't address client applications. Probably other weird issues too ... FWIW ~ that's the current experience as well, so we're not any worse off. I find it highly unlikely that users will enable TLS after building up a large production cluster anyway, that seems like it wouldn't happen in practice.

Regarding migrating to this: I don't think these changes would require any migration process. Currently, `node_name` is stored in the state in ZK (as you know), so the stored `base_url` will just be ignored and re-created when reading from ZK using the `node_name`.

noblepaul:
> we don't want to promise the community a zero-downtime upgrade to enable TLS, because it is a hard thing to promise

I don't have a problem if there is a downtime when you upgrade from one schem to another. I only worry if the scheme remains same and a rolling restart is not posisble.

noblepaul:
LGTM
👍

thelabdude:
thank you for the review @noblepaul and @anshumg ... going to merge to master now.

broberson721:
UrlScheme.INSTANCE became a singleton with this PR. That is causing failures for use cases where a single client connects to some clusters with SSL enabled and some without SSL enabled.. Are there any plans to fix this in a future release?

thelabdude:
yeah we need to fix that ... the singleton works great for the server-side but doesn't work for multiple clusters with different SSL config as you describe ... I'll come up with something for 8.11.2 and 9. In the meantime, I guess you can use a pre-8.8 SolrJ?

broberson721:
Appreciate that.. Any sense how off 8.11.2/this fix would be? A couple months? Or much longer?

thelabdude:
Can't really speak to a release date for 8.11.2 but bug fix releases can happen pretty quickly. On the other hand, it's better to let 8.11.1 live in the wild for a little bit before rushing out another bug fix release. If there's not traction on that from others in the community in a few weeks, I'll get the release going myself.

I've already been looking into a related issue, so will address this here as well: https://issues.apache.org/jira/browse/SOLR-15587

broberson721:
Understood. And thanks. We're hoping a fix gets out there before we have to patch due to security issues. I think the big security issue with 8.7 (CVE-2021-29943) doesn't affect us. But it's usually just a matter of time.

thelabdude:
@broberson721 ~ I'm doing the work for this issue here: https://github.com/apache/solr/pull/460 ... if you want to follow along. Thanks again for reporting, obviously I hadn't considered the use-case of multiple clients in the same JVM talking to different Solr clusters ;-)

PR #2017, opened 2020-10-21
SOLR-14957: Add Prometheus Exporter to docker PATH. Fix classpath issues.

HoustonPutman:
https://issues.apache.org/jira/browse/SOLR-14957

While testing this out, I was unable to make the `solr-exporter` run without fixing classpath issues. I think this pr, https://github.com/apache/lucene-solr/pull/1972, was trying to fix a similar thing but with a bigger scope. I ultimately support that goal of that PR, but thought I might as well include my fixes in this PR as they are small and can be easily overridden if necessary.

I made the necessary changes to get the exporter to run with default settings:

- Adding prometheus-exporter/conf to the classpath and changing the default config file to be the location of the file on the classpath
- Adding the `solr/server/lib/ext` jars to the classpath in order to get logging to work.


dsmiley:
+1 LGTM; I see you are putting SLF4J JARs on the classpath, so I assume that was the issue you encountered?  What's in conf?  Oh the configuration... yes I ran into this issue and I solved it by setting the CWD from Docker to prometheus-exporter's dir.  Please be sure to Q/A both manual execution (i.e. as if from a distribution) and via Docker.

I thought you might have also added the env var overrides while you were at it but maybe you want to do that on another issue?  You could do it here; IMO it fits with the scope of making prometheus-exporter easier to use via Docker than is today (possible but some quirks).

dsmiley (approved):

HoustonPutman:
Yeah it was the SLF4J JARs that were the issue, as well as the configuration file for the metrics. I will test manually as well before committing.

I was planning on putting the env var overrides in a separate PR, because that can be backported to 8x, while these changes are really only for 9.0. I guess the configuration classpath issue can also be backported to 8.x, but that's the only change here that's really applicable.

HoustonPutman:
Confirmed, can run the prometheus exporter via the `packaging/build/dev` folder with the default options as well as specifying a concrete path for the metrics config.

PR #2018, opened 2020-10-21
LUCENE-9582: rename VectorValues.ScoreFunction to SearchStrategy

msokolov:
This is - renaming as the title says, plus a light refactoring, and adding `o.a.l.util.VectorUtil` to house the score function implementations.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/util/VectorUtil.java
+++ b/lucene/core/src/java/org/apache/lucene/util/VectorUtil.java
@@ -60,7 +60,7 @@ public final class VectorUtil {
     return res;
   }
 
-  public static float squareSum(float[] v1, float[] v2) {
+  public static float squareDistance(float[] v1, float[] v2) {
     assert v1.length == v2.length;
     float squareSum = 0.0f;
     int dim = v1.length;
--- a/lucene/core/src/java/org/apache/lucene/util/VectorUtil.java
+++ b/lucene/core/src/java/org/apache/lucene/util/VectorUtil.java
@@ -1,0 +1,74 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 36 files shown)


lucene/core/src/java/org/apache/lucene/util/VectorUtil.java
```diff
@@ -0,0 +1,74 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.lucene.util;
+ 
+/**
+ * Utilities for computations with numeric arrays
+ */
+public final class VectorUtil {
```
jtibshirani, line 23: Should we add a light unit test for these utility methods?

jtibshirani (commented):
The `SearchStrategy` concept and naming makes sense to me.

lucene/core/src/java/org/apache/lucene/util/VectorUtil.java
msokolov: oh, good call. I will add

lucene/queries/src/java/org/apache/lucene/queries/intervals/IntervalQuery.java
```diff
@@ -99,7 +99,7 @@ public IntervalQuery(String field, IntervalsSource intervalsSource, float pivot,
   private IntervalQuery(String field, IntervalsSource intervalsSource, IntervalScoreFunction scoreFunction) {
     Objects.requireNonNull(field, "null field aren't accepted");
     Objects.requireNonNull(intervalsSource, "null intervalsSource aren't accepted");
-    Objects.requireNonNull(scoreFunction, "null scoreFunction aren't accepted");
+    Objects.requireNonNull(scoreFunction, "null searchStrategy aren't accepted");
```
jtibshirani, line 102: Just noticed this rename seems accidental.

lucene/core/src/test/org/apache/lucene/util/TestVectorUtil.java
```diff
@@ -0,0 +1,83 @@
...
+    float[] v = randomVector();
+    assertEquals(0, VectorUtil.squareSum(v, v), 1e-5);
+  }
+
+  public void testBasicSquareSum() {
+    assertEquals(12, VectorUtil.squareSum(new float[]{1, 2, 3}, new float[]{-1, 0, 5}), 0);
+  }
+
+  public void testRandomSquareSum() {
+    // the MSE of a vector with its inverse is equal to four times the sum of squares of its components
```
jtibshirani, line 53: Small comment: I don't think you mean MSE here since it's not a mean, it could be 'squared distance' ?

jtibshirani (approved):

msokolov:
thanks for the review! I'll fix when merging

PR #2020, opened 2020-10-22
SOLR-14949: Ability to customize Solr Docker build

HoustonPutman:
https://issues.apache.org/jira/browse/SOLR-14949

Custom inputs:
- Docker Image repo
- Docker Image tag
- Docker Image name
- Base Docker Image

Also added a gradlew helpDocker page.


help/docker.txt
```diff
@@ -0,0 +1,53 @@
+Docker Images for Solr
+======================
+
+Solr docker images are built using Palantir's Docker Gradle plugin, https://github.com/palantir/gradle-docker.
+
+Common Inputs
+-------------
+
+The docker image and it's tag can be customized via the following options, all accepted via both Environment Variables and Gradle Properties.
```
madrob, line 9: s/it's/its

.github/workflows/docker-test.yml
```diff
@@ -17,6 +17,10 @@ jobs:
 
     runs-on: ubuntu-latest
 
+    env:
+      DOCKER_SOLR_IMAGE_REPO: github-pr/solr
```
madrob, line 21: Does this mean we will publish all of our PR docker images somewhere?

help/docker.txt
```diff
@@ -0,0 +1,53 @@
...
+   Gradle Property: -Pdocker.solr.imageName
+
+Building
+--------
+
+In order to build the Solr Docker image, run:
+
+gradlew docker
+
+The docker build task (`gradlew docker`) accepts the following inputs, in addition to the common inputs listed above:
```
madrob, line 33: This is displayed as plain text, not markdown.

solr/docker/tests/cases/version/test.sh
```diff
@@ -1,45 +0,0 @@
-#!/bin/bash
```
madrob, line 1: Deleted because we can set the tag and version to be anything now?

HoustonPutman: Yeah, this is something that might be useful when we start creating release artifacts, but the test we want will probably look quite different. This is pretty shallow and doesn't actually provide a whole lot.

.github/workflows/docker-test.yml
HoustonPutman: nah this doesn't actually push. It's just testing that the custom repo/tag works.

I'm pretty sure it would fail if it tried, because github-pr isn't a docker hub repo.

help/docker.txt
```diff
@@ -0,0 +1,53 @@
...
+
+Common Inputs
+-------------
+
+The docker image and it's tag can be customized via the following options, all accepted via both Environment Variables and Gradle Properties.
+
+Docker Image Repository:
+   Default: "apache/solr"
+   EnvVar: DOCKER_SOLR_IMAGE_REPO
+   Gradle Property: -Pdocker.solr.imageRepo
```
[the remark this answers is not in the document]
HoustonPutman: I could go either way. 

Say we add a prometheus exporter. Which would we prefer?
- `docker.solr.imageRepo` and `docker.prometheusExporter.imageRepo`
- `solr.docker.imageRepo` and `prometheusExporter.docker.imageRepo`

HoustonPutman:
@dweiss Is the way I am declaring and obtaining variables the correct way of doing it?

help/docker.txt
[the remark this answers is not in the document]
HoustonPutman: Nevermind, I agree with you. Made the changes.

dweiss:
Hmm... sorry, missed your mention/ request somehow. Yeah - these functions are intended to read variables from multiple locations so it looks ok to me. I didn't test it (or use docker much for that matter).

HoustonPutman:
Thanks for the sanity check!

I've tested pretty thoroughly, and the PR test does some checks on its own too.

dsmiley:
FWIW I don't think /help/docker.txt is a good place for this because the Docker module has it's very own folder, and so shouldn't it go there?  I've been poking around the docker support today in the docker folder and had no clue this help file existed.

HoustonPutman:
That's fair, we can certainly move it. I included it there because that seemed like the place to put help files for gradle usage. All of the files used in the `help.gradle` are in that directory, but I guess there's no actual requirement for that.

PR #2022, opened 2020-10-23
LUCENE-9004: KNN vector search using NSW graphs

msokolov:
Phew this has been a long time coming, but I think it is in good shape now. We started with a scratchy prototype about a year ago, then @mocobeta got it on a better footing by adding a new codec and also implemented the full hierarchical algorithm, making the graph search faithful to the published literature. Then we took a step back to add underlying vector format as a separate patch, now landed. This patch builds on the new vector format, providing KNN search with NSW graphs. It's the simplest implementation I could tease out (single layer graph, simple neighbor selection, no max fanout control), but I think it will be a good foundation. I've done some pretty extensive performance testing and hyperparameter exploration using the (included) KnnGraphTester with some proprietary data, and get good results. I will follow up later with specifics, but single-threaded latencies in a few ms on my i7 laptop over a 1M x 256-dim dataset seems pretty good. Followups will include repeatable benchmarks on public datasets.


lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraph.java
```diff
@@ -0,0 +1,235 @@
...
+ *   Thus <code>fanout</code> is like a combination of <code>ef</code> (search beam width) from the 2016 paper and <code>m</code> from the 2014 paper.
+ *   </li>
+ * </ul>
+ *
+ * <p>Note: The graph may be searched by multiple threads concurrently, but updates are not thread-safe. Also note: there is no notion of
+ * deletions. Document searching built on top of this must do its own deletion-filtering.</p>
+ */
+public final class HnswGraph {
+
+  // each entry lists the neighbors of a node, in node order
```
mikemccand, line 66: A node corresponds to an indexed vector, right?  And it is represented using its compact vector ordinal?

lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraph.java
```diff
@@ -0,0 +1,235 @@
...
+      default:
+        throw new IllegalStateException("invalid search strategy: " + searchStrategy);
+    }
+  }
+
+  /**
+   * Return whether the given strategy returns lower values for nearer vectors
+   * @param searchStrategy the strategy
+   */
+  public static boolean isReversed(VectorValues.SearchStrategy searchStrategy) {
```
mikemccand, line 196: Maybe we could add this as a method on the `enum` constants instead?  So then caller could just call `SearchStrategy.isReversed`?

lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraph.java
```diff
@@ -0,0 +1,235 @@
...
+     */
+  }
+
+  /**
+   * Calculates a similarity score between the two vectors with a specified function.
+   * @param v1 a vector
+   * @param v2 another vector, of the same dimension
+   * @return the value of the strategy's score function applied to the two vectors
+   */
+  public static float compare(float[] v1, float[] v2, VectorValues.SearchStrategy searchStrategy) {
```
mikemccand, line 181: Maybe this could also be a method on `SearchStrategy` enum?

lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraphBuilder.java
```diff
@@ -0,0 +1,186 @@
...
+/**
+ * Builder for HNSW graph. See {@link HnswGraph} for a gloss on the algorithm and the meaning of the hyperparameters.
+ */
+public final class HnswGraphBuilder {
+
+  // default random seed for level generation
+  private static final long DEFAULT_RAND_SEED = System.currentTimeMillis();
+
+  // expose for testing.
+  public static long randSeed = DEFAULT_RAND_SEED;
```
mikemccand, line 39: Does our `test-framework` `-Dtests.seed=XXX` fix the seed here too?  So that (hopefully!) test failures are reproducible.

lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraphBuilder.java
```diff
@@ -0,0 +1,186 @@
...
+
+  // These "default" hyperparameter settings are exposed (and nonfinal) to enable performance testing
+  // since the indexing API doesn't provide any control over them.
```
mikemccand, line 42: Oh I see -- users cannot (easily) change these defaults while indexing?  I would think we could provide these arguments to the `Codec` at write time and route them down to this `HNSW` building process...

lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraph.java
```diff
@@ -0,0 +1,235 @@
...
+    } else {
+      assert node2 == graph.size();
+      arcs2 = new ArrayList<>();
+      graph.add(arcs2);
+    }
+    arcs2.add(new Neighbor(node1, score));
+
+    // ensure #arcs <= maxConnections
+    /*
+    if (maxConnections > 0) {
```
mikemccand, line 169: Hmm did we decide not to impose the `maxConnections`?

lucene/core/src/java/org/apache/lucene/codecs/lucene90/Lucene90VectorFormat.java
```diff
@@ -54,4 +56,10 @@ public VectorReader fieldsReader(SegmentReadState state) throws IOException {
     return new Lucene90VectorReader(state);
   }
 
+  static boolean isHnswStrategy(VectorValues.SearchStrategy searchStrategy) {
```
mikemccand, line 59: Maybe also make this a method on `SearchStrategy` enum?

lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraph.java
```diff
@@ -0,0 +1,235 @@
...
+ *
+ * <p>Note: The graph may be searched by multiple threads concurrently, but updates are not thread-safe. Also note: there is no notion of
+ * deletions. Document searching built on top of this must do its own deletion-filtering.</p>
+ */
+public final class HnswGraph {
+
+  // each entry lists the neighbors of a node, in node order
+  private final List<List<Neighbor>> graph;
+
+  HnswGraph() {
```
mikemccand, line 69: Is this class only used at segment-write time?  Or, is this also used to represent the `HNSWGraph` at search time?

OK I think we do not use this at search time -- instead we `.seek` to a node, and then iterate through its `int` neighbors.  I guess we do not store the scores for each arc in the index.

From your benchmarks, do you have a sense of how much heap is required (as a function of number of vectors) by this class to construct the ANN neighbors?

After the one vector field for one segment is written, this class is GC'able?  So if I have three vector fields, the heap will only have one `HnswGraph` living at a time?

lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraph.java
```diff
@@ -0,0 +1,235 @@
...
+    return new HnswGraphValues();
+  }
+
+  /**
+   * Present this graph as KnnGraphValues, used for searching while inserting new nodes.
+   */
+  private class HnswGraphValues extends KnnGraphValues {
+
+    private int arcUpTo;
+    private int[] arcs;
```
mikemccand, line 217: An arc is the node id (compact ordinal) for the destination of the arc, right?  I.e. the neighbor nodes.

Maybe rename `arcs` to `neighborNodes`?

lucene/core/src/java/org/apache/lucene/util/hnsw/BoundsChecker.java
```diff
@@ -0,0 +1,74 @@
...
+        if (reversed) {
+            return new Min();
+        } else {
+            return new Max();
+        }
+    }
+
+    static class Max extends BoundsChecker {
+        Max() {
+            bound = -Float.MAX_VALUE;
```
mikemccand, line 44: At first I thought we should use `Float.MIN_VALUE` here, but that is not right, since it is the smallest positive float.

But then I worried whether negating `Float.MAX_VALUE` produces a valid float (e.g. you cannot safely negate `Integer.MIN_VALUE`).

Maybe just use `Float.NEGATIVE_INFINITY` here and `Float.POSITIVE_INFINITY` for `Min`?

lucene/core/src/java/org/apache/lucene/codecs/lucene90/Lucene90VectorFormat.java
```diff
@@ -54,4 +56,10 @@ public VectorReader fieldsReader(SegmentReadState state) throws IOException {
...
+    return searchStrategy == VectorValues.SearchStrategy.DOT_PRODUCT_HNSW ||
+        searchStrategy == VectorValues.SearchStrategy.EUCLIDEAN_HNSW;
+  }
+
+
```
mikemccand, line 64: Remove extra blank line?

lucene/core/src/java/org/apache/lucene/codecs/lucene90/Lucene90VectorFormat.java
```diff
@@ -33,15 +34,16 @@
...
   static final String META_EXTENSION = "vem";
   static final String VECTOR_DATA_EXTENSION = "vec";
+  static final String VECTOR_INDEX_EXTENSION = "vex";
 
   static final int VERSION_START = 0;
   static final int VERSION_CURRENT = VERSION_START;
 
   /** Sole constructor */
-  public Lucene90VectorFormat() {
+  Lucene90VectorFormat() {
```
mikemccand, line 46: Hmm, why no longer `public`?  Users making their own Codec might want to use this constructor I think?

lucene/core/src/java/org/apache/lucene/codecs/lucene90/Lucene90VectorFormat.java
```diff
@@ -24,6 +24,7 @@
 import org.apache.lucene.codecs.VectorWriter;
 import org.apache.lucene.index.SegmentReadState;
 import org.apache.lucene.index.SegmentWriteState;
+import org.apache.lucene.index.VectorValues;
 
 /**
  * Lucene 9.0 vector format, which encodes dense numeric vector values.
```
mikemccand, line 30: Hmm, remove the `TODO` in here?

Add `@lucene.experimental`?

lucene/core/src/java/org/apache/lucene/index/KnnGraphValues.java
```diff
@@ -0,0 +1,60 @@
...
+ */
+
+package org.apache.lucene.index;
+
+import java.io.IOException;
+
+import static org.apache.lucene.search.DocIdSetIterator.NO_MORE_DOCS;
+
+/**
+ * Access to per-document friends lists in a (hierarchical) knn search graph.
```
mikemccand, line 25: s/`friends`/`neighbors`?

lucene/core/src/java/org/apache/lucene/index/KnnGraphValues.java
```diff
@@ -0,0 +1,60 @@
...
+   * @param target must be a valid node in the graph, ie. &ge; 0 and &lt; {@link VectorValues#size()}.
+   */
+  public abstract void seek(int target) throws IOException;
+
+  /**
+   * Iterates over the neighbor list. It is illegal to call this method after it returns
+   * NO_MORE_DOCS without calling {@link #seek(int)}, which resets the iterator.
+   * @return a node ordinal in the graph, or NO_MORE_DOCS if the iteration is complete.
+   */
+  public abstract int nextArc() throws IOException ;
```
mikemccand, line 44: Remove extra space before `;`?

lucene/core/src/java/org/apache/lucene/index/KnnGraphValues.java
```diff
@@ -0,0 +1,60 @@
...
+
+import java.io.IOException;
+
+import static org.apache.lucene.search.DocIdSetIterator.NO_MORE_DOCS;
+
+/**
+ * Access to per-document friends lists in a (hierarchical) knn search graph.
+ * TODO: replace with SortedNumericDocValues??
+ */
+public abstract class KnnGraphValues {
```
mikemccand, line 28: Add `@lucene.experimental`?

lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraph.java
```diff
@@ -0,0 +1,235 @@
...
+  // each entry lists the neighbors of a node, in node order
+  private final List<List<Neighbor>> graph;
+
+  HnswGraph() {
+    graph = new ArrayList<>();
+    graph.add(new ArrayList<>());
+  }
+
+  /**
+   * Searches for the nearest neighbors of a query vector.
```
mikemccand, line 75: This is expected to normally be used just during indexing?  Maybe explain that in the javadoc?  I.e. at search time, users should use the other `search` method that pulls neighbors from the index and recurses.

mikemccand (commented):
This is impressive @msokolov!

I left a bunch of small comments, mostly asking naive questions ;)

Are we planning on adding a `VectorsQuery` in a follow-on issue/PR, to make it simpler for users to combine ANN search with other clauses in a `BooleanQuery`, to ensure deletions are filtered out, etc.?

lucene/core/src/java/org/apache/lucene/codecs/lucene90/Lucene90VectorReader.java
```diff
@@ -102,23 +122,28 @@ private void readFields(ChecksumIndexInput meta, FieldInfos infos) throws IOExce
...
+      throw new CorruptIndexException("Invalid search strategy id: " + searchStrategyId, input);
+    }
+    return VectorValues.SearchStrategy.values()[searchStrategyId];
+  }
+
+  private FieldEntry readField(DataInput input) throws IOException {
+    VectorValues.SearchStrategy searchStrategy = readSearchStrategy(input);
+    switch(searchStrategy) {
+      case NONE:
+        return new FieldEntry(input, searchStrategy);
```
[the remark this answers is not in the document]
msokolov: right; they may choose an indexing approach based on a strategy (the names indicate algorithm/metric, but could be anything), or NONE. That can be useful for calculating vector score only for the purpose of ranking hits matched by other means.

lucene/core/src/java/org/apache/lucene/codecs/lucene90/Lucene90VectorReader.java
```diff
@@ -165,42 +191,88 @@ public VectorValues getVectorValues(String field) throws IOException {
     return new OffHeapVectorValues(fieldEntry, bytesSlice);
   }
 
+  // exposed for testing
```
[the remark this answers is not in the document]
msokolov: Fair; mostly I think the idea was to minimize the new API surface area that needs to be documented and maintained, trying to stick to VectorValues as the main public API with the KnnGraph as an internal indexing data structure

lucene/core/src/java/org/apache/lucene/index/VectorValues.java
```diff
@@ -74,6 +74,18 @@ public BytesRef binaryValue() throws IOException {
...
   }
 
+  /**
+   * Return the k nearest neighbor documents as determined by comparison of their vector values
+   * for this field, to the given vector, by the field's search strategy. If the search strategy is
+   * reversed, lower values indicate nearer vectors, otherwise higher scores indicate nearer
+   * vectors. Unlike relevance scores, vector scores may be negative.
+   * @param target the vector-valued query
+   * @param k      the number of docs to return
+   * @param fanout control the accuracy/speed tradeoff - larger values give better recall at higher cost
```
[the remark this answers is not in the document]
msokolov: Yeah, I think there needs to be a follow-on exposing the index-time controls, which indeed are much more potent than this search-time fanout, which has only a small impact on recall and latency. In this patch they are globals in HnswGraphBuilder, but there is no API for setting them. I am thinking the index-time hyperparameters would be specified in IndexWriterConfig?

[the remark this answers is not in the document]
mikemccand: > Yeah, I think there needs to be a follow-on exposing the index-time controls, which indeed are much more potent than this search-time fanout, which has only a small impact on recall and latency. In this patch they are globals in HnswGraphBuilder, but there is no API for setting them.

OK, makes sense.

> I am thinking the index-time hyperparameters would be specified in IndexWriterConfig?

Hmm, maybe these could be codec level controls?  Or maybe `FieldInfo`?  They would be per-vector-field configuration right?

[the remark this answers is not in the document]
msokolov: Yeah that's a good point. While experimenting with GloVe I'm learning that different settings are appropriate for different vectors, so field-level control might be needed. I'm not sure how codec-level controls are exposed. Don't Codecs get created automatically using no-args constructors and service autodiscovery? Did you mean something like perFieldVectorFormat? Except I doubt we need a new format; it's more about some metadata values that we would store in the field, so I think yeah it would go in FieldInfo. But I'm reluctant to expose hnsw-specific hyperparameters in `VectorField`, which we want to support other algorithms as well. Maybe this is a good use case for `IndexableField.getAttributes()`?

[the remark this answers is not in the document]
msokolov: Yeah this probably needs to be field level. Different strokes for different collections of vectors. I'm not sure how to expose since the parameters will be different for different ANN implementations. Might be a good use case for generic IndexedField.attributes?

lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraph.java
msokolov: Yes, although this class is agnostic as to the interpretation of these nodes. They could just as well be docIds, or social security numbers. But yeah, the only usage is as you describe, so I'll add a note to the javadoc.

lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraph.java
msokolov: yeah that does seem cleaner. It was what I had in an earlier revision, but got thrown out with the bathwater of the vector scoring implementations.

lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraph.java
msokolov: No this method is intended for use both when indexing and when searching. I guess this method could go in some other class (KnnGraphValues??) since it doesn't rely on Hnsw in any direct way (it only needs an abstract source of vectors and graph values). But it does use Neighbor/Neighbors and is heavily used *by* HnswGraphBuilder, so it seems natural to have it here.

lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraphBuilder.java
msokolov: yes!

lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraph.java
msokolov: Coming in the next revision

lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraph.java
msokolov: Right, we do not instantiate this class at search time, although we do use the static `search` method, which only relies on the abstract vector and graph values. Heap usage is quite high! It will scale with graph fanout * number of vectors. For each node, we store a PriorityQueue of Neighbor objects; each Neighbor has an int node id and a float score. I think for 1M vectors, with fanout of 64, we'll see around 1.5G heap usage, with lots of tiny objects on the heap during flush/merge. This can be dramatically reduced by eliminating the Neighbor objects in favor of parallel arrays, and the priority queues can then store int ordinals rather than object pointers, but I think that can be a fast follow?

lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraph.java
msokolov: sure, neighborNodes it is

lucene/core/src/java/org/apache/lucene/util/hnsw/BoundsChecker.java
msokolov: Well, `-Float.MAX_VALUE` is reliable since the sign bit is separate from all the rest of the bits, but `Float.NEGATIVE_INFINITY` is less than that, so if we use it then any finite float will be a valid score, so +1 for that, and for +infinity as well

lucene/core/src/java/org/apache/lucene/index/VectorValues.java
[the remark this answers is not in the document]
mikemccand: > Don't Codecs get created automatically using no-args constructors and service autodiscovery? 

They do at read (search) time!  But at write time, you can pass parameters that alter how the Codec does its work, as long as the resulting index is then readable at search time with no-args constructors.

I vaguely remember talking about having ways for Codec at read-time to also take options, but I'm not sure that was ever fully designed / pushed ... @s1monw may remember?

> But I'm reluctant to expose hnsw-specific hyperparameters in VectorField, which we want to support other algorithms as well.
> Might be a good use case for generic IndexedField.attributes?

Yeah, maybe?  I agree it is not obvious where the API should live and how it then finds its way into the ANN data structure construction when writing each segment.

lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraph.java
mikemccand: Yeah, +1 for fast follow!

lucene/core/src/java/org/apache/lucene/codecs/lucene90/Lucene90VectorReader.java
```diff
@@ -22,39 +22,60 @@
...
+    long[] checksumRef = new long[1];
+    vectorData = openDataInput(state, versionMeta, Lucene90VectorFormat.VECTOR_DATA_EXTENSION, Lucene90VectorFormat.VECTOR_DATA_CODEC_NAME, checksumRef);
+    vectorIndex = openDataInput(state, versionMeta, Lucene90VectorFormat.VECTOR_INDEX_EXTENSION, Lucene90VectorFormat.VECTOR_INDEX_CODEC_NAME, checksumRef);
+    checksumSeed = checksumRef[0];
+  }
+
+  private int readMetadata(SegmentReadState state, String fileExtension) throws IOException {
+    String metaFileName = IndexFileNames.segmentFileName(state.segmentInfo.name, state.segmentSuffix, fileExtension);
     int versionMeta = -1;
+    long checksum = -1;
```
mikemccand, line 78: Hmm is this unused?

lucene/core/src/java/org/apache/lucene/codecs/lucene90/Lucene90VectorReader.java
```diff
@@ -277,24 +351,46 @@ public long cost() {
...
     @Override
-    public RandomAccess randomAccess() {
+    public RandomAccessVectorValues randomAccess() {
       return new OffHeapRandomAccess(dataIn.clone());
     }
 
+    @Override
+    public TopDocs search(float[] vector, int topK, int fanout) throws IOException {
+      // use a seed that is fixed for the index so we get reproducible results for the same query
+      final Random random = new Random(checksumSeed);
```
mikemccand, line 361: Clever seed!

mikemccand (approved):
Thanks @msokolov, this is an awesome progress-not-perfection start to adding ANN to Lucene!

lucene/core/src/java/org/apache/lucene/index/VectorValues.java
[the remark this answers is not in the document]
s1monw: @mikemccand it was pushed but removed again in https://issues.apache.org/jira/browse/LUCENE-9257 a little while ago. I get why it's removed but it seems useful maybe we add it back?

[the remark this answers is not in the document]
mikemccand: Ahh OK thanks for the context @s1monw.

uschindler:
Hi,
this broke javadocs:
> /home/jenkins/workspace/Lucene-Solr-master-Linux/lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraph.java:43: error: heading used out of sequence: <H3>, compared to implicit preceding heading: <H1>
 * <h3>Hyperparameters</h3>
   ^
1 error

This line is wrong: https://github.com/apache/lucene-solr/blob/b36b4af22bb76dc42b466b818b417bcbc0deb006/lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraph.java#L43
(should be `<h2>`)

PR #2028, opened 2020-10-25
LUCENE-9322: Make sure to account for vectors in SortingCodecReader.

jtibshirani:
This ensures that vector values are sorted when sorting an index after-the-fact
using `SortingCodecReader`.


msokolov:
Oh! Good catch, I forgot about this one. I guess that FilteringCodecReader is a bit trappy, but probably OK given how infrequently we add new formats.

msokolov (approved):

PR #2030, opened 2020-10-26
Fix CHANGES.txt for 8.7

atris:
Fixing merge issues with 8.7 change notes

### DIFF
--- a/solr/CHANGES.txt
+++ b/solr/CHANGES.txt
@@ -10,108 +10,6 @@ Consult the lucene/CHANGES.txt file for additional, low level, changes in this r
 
 New Features
 ---------------------
-* SOLR-14151 Make schema components load from packages (noble)
-
-* SOLR-14588: Introduce Circuit Breaker Infrastructure and a JVM heap usage memory tracking circuit breaker implementation (Atri Sharma)
-
-* SOLR-14615: CPU Utilization Based Circuit Breaker (Atri Sharma)
-
-Improvements
-----------------------
-* LUCENE-8984: MoreLikeThis MLT is biased for uncommon fields (Andy Hind via Anshum Gupta)
-
-* SOLR-14223: PKI Auth can bootstrap from existing key files instead of creating new keys on startup (Mike Drob)
-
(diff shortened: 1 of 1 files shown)


munendrasn:
Some change log entries which are meant only for master (due to backward compatibility break) has been backported to branch_8x in this commit https://github.com/apache/lucene-solr/commit/c841f63e4ba8615d648f00741d4ee3628cd60954
I think we would need to fix that too.

solr/CHANGES.txt
```diff
@@ -177,6 +77,15 @@ Improvements
...
 
+* SOLR-14223: PKI Auth can bootstrap from existing key files instead of creating new keys on startup (Mike Drob)
+
+* SOLR-11725: Use corrected sample formula for computing stdDev and variance in JSON aggregations
+  (hossman, Munendra S N, yonik)
+
+* SOLR-14387: SolrClient.getById() will escape comma separater within ids (Markus Schuch via Mike Drob)
+
+* SOLR-10814: Add short-name feature to RuleBasedAuthz plugin (Mike Drob, Hrishikesh Gadre)
+
```
munendrasn, line 88: AFAIK, this shouldn't be in either 8x or 8_7 branches. These are master only changes

munendrasn: I have created PR #2033 which fixes changelog for branch_8x. If it looks good then, we can port it to branc_8_7

munendrasn: Also, there are entries like below
```
* SOLR-14859: DateRangeField now throws errors when invalid field/fieldType options specified; no longer silently accepts incompatible option values
  (Jason Gerlowski, Chris Hostetter, Munendra S N)
```
This was released in 8.6.3, not sure if it should be again included in 8.7 release

jpountz: Based on our release process, changes entries for released versions should not appear in the changes of unreleased versions, so we should drop this one indeed. See https://cwiki.apache.org/confluence/display/LUCENE/ReleaseTodo#ReleaseTodo-SynchronizeCHANGES.txt.

PR #2037, opened 2020-10-26
LUCENE-9583: extract separate RandomAccessVectorValues interface

msokolov:
This moves the random access vector interface out of VectorValues with the idea of making it less prominent. All implementations still support it, and it is required by the codec, but the public-facing VectorValues api becomes cleaner.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/VectorValues.java
+++ b/lucene/core/src/java/org/apache/lucene/index/VectorValues.java
@@ -75,68 +75,23 @@ public abstract class VectorValues extends DocIdSetIterator {
   }
 
   /**
-   * Return a random access interface over this iterator's vectors. Calling the RandomAccess methods will
-   * have no effect on the progress of the iteration or the values returned by this iterator. Successive calls
-   * will retrieve independent copies that do not overwrite each others' returned values.
+   * Return the k nearest neighbor documents as determined by comparison of their vector values
+   * for this field, to the given vector, by the field's search strategy. If the search strategy is
+   * reversed, lower values indicate nearer vectors, otherwise higher scores indicate nearer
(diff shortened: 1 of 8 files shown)


lucene/core/src/java/org/apache/lucene/index/VectorValues.java
```diff
@@ -75,68 +75,23 @@ public BytesRef binaryValue() throws IOException {
...
-     * reversed, lower values indicate nearer vectors, otherwise higher scores indicate nearer
-     * vectors. Unlike relevance scores, vector scores may be negative.
-     * @param target the vector-valued query
-     * @param k      the number of docs to return
-     * @param fanout control the accuracy/speed tradeoff - larger values give better recall at higher cost
-     * @return the k nearest neighbor documents, along with their (searchStrategy-specific) scores.
-     */
-    TopDocs search(float[] target, int k, int fanout) throws IOException;
-  }
+  public abstract TopDocs search(float[] target, int k, int fanout) throws IOException;
```
mikemccand, line 87: Is it intentional that we are leaving `search` on `VectorValues`, but moving `searchStrategy` to `RandomAccessVectorValues`?

mikemccand (approved):

lucene/core/src/java/org/apache/lucene/index/VectorValues.java
msokolov: Right, it's not moved actually just *still* in both places. The RA one is basically always a delegate for the other one. There is some ongoing discussion about this whole topic on the JIRA, so perhaps we'll ultimately end up with something different, but for the nonce I think we can commit this in order to make progress while we iterate.

PR #2038, opened 2020-10-26
SOLR-14955: Add env var options to Prometheus Export scripts.

HoustonPutman:
https://issues.apache.org/jira/browse/SOLR-14955

### DIFF
--- a/solr/contrib/prometheus-exporter/bin/solr-exporter
+++ b/solr/contrib/prometheus-exporter/bin/solr-exporter
@@ -1,5 +1,4 @@
-#!/bin/sh
-
+#!/usr/bin/env bash
 #
 # Licensed to the Apache Software Foundation (ASF) under one or more
 # contributor license agreements.  See the NOTICE file distributed with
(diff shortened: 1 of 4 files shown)


solr/solr-ref-guide/src/monitoring-solr-with-prometheus-and-grafana.adoc
```diff
@@ -81,34 +81,44 @@ $ ./bin/solr-exporter -p 9854 -z localhost:2181/solr -f ./conf/solr-exporter-con
...
+All parameters can be provided via an environment variable, instead of through the command line.
 
 `h`, `--help`::
 Displays command line help and usage.
 
-`-p`, `--port`::
-The port where Prometheus will listen for new data. This port will be used to configure Prometheus. It can be any port not already in use on your server. The default is `9983`.
+`-p`, `--port`, `$PORT`::
+The port where Prometheus will listen for new data. This port will be used to configure Prometheus.
+It can be any port not already in use on your server. The default is `9983`.
```
janhoy, line 92: I find it puzzling that default port is the same as the default port for embedded Zookeeper, i.e. if you start `bin/solr -c` and then start exporter on same host, you'll have a port collision.

solr/contrib/prometheus-exporter/bin/solr-exporter
```diff
@@ -135,15 +134,42 @@ if $cygwin; then
...
 fi
 
+# Convert Environment Variables to Command Line Options
+EXPORTER_ARGS=()
+
+if [[ -n "$CONFIG_FILE" ]]; then
+  EXPORTER_ARGS+=(-f "$CONFIG_FILE")
+fi
+
+if [[ -n "$PORT" ]]; then
```
janhoy, line 144: I like the simplicity of `PORT`, but is there a risk that such generic naming could lead to conflict, i.e. that a user hast this env set for some other app, and then the exporter will use it?

janhoy (approved):
LGTM, just a few comments. Have not tested it live.

solr/solr-ref-guide/src/monitoring-solr-with-prometheus-and-grafana.adoc
HoustonPutman: That is very fair, and something we could certainly change in a major version, such as 9.0.

It should probably be done separately, since these changes are not backwards-incompatible.

solr/contrib/prometheus-exporter/bin/solr-exporter
HoustonPutman: Yeah, that could certainly happen. `EXPORTER_PORT` would work perfectly fine. I went the direction decided in https://github.com/apache/lucene-solr/pull/887 on variable naming. I don't have a strong opinion either way.

janhoy: For a docker env it's not a problem, and for most servers you'll have a separate service user with clean shell. Developers could unintentionally have the PORT var set on their laptop, but I don't think we should make design decisions from that.

On a side note, I tested what would happen if you have both env `$PORT=1111` and arg `-p 2222`, and the last arg will win, thus you can override any env with explicit command arg, which I guess is a good thing.

solr/solr-ref-guide/src/monitoring-solr-with-prometheus-and-grafana.adoc
janhoy: https://issues.apache.org/jira/browse/SOLR-14972

solr/contrib/prometheus-exporter/bin/solr-exporter
HoustonPutman: That is good to know. In that case we should probably stick with `$PORT`, as there's an easy fix in the case where someone is affected by it.

PR #2040, opened 2020-10-27
SOLR-14965 add overseer queue size metrics

saatchibhalla:
# Description

The Overseer work queues stored in ZK and abstracted by the Overseer can give us a good indication of the health of the cluster - if messages are taking too long to dequeue or the queue is growing too large, we know that the Overseer is overloaded and we are going to overwhelm the cluster.

This work adds metrics to track the size of the Overseer queues (ClusterStateUpdate queue and Collections Work Queue).  

# Solution

Registered these two size metrics in a shared metrics registry for the Overseer. The Overseer shared metrics registry is only initialized upon reference so this shouldn't have any impact on metrics when run in non-cloud mode. Also updated the local solr-exporter-config.xml so that these metrics are exported and can be viewed locally in Grafana. 

# Tests

Local testing was accomplished by sending CreateCollection load and monitoring a local grafana dashboard which accessed the metrics through prometheus-exporter. As the size of these queues change extremely quickly, it was difficult to see changes in these metrics quickly enough--and if the requests were sent too quickly we'd see local solr hosts OOMing. However, I was able to verify that the metrics are updating as the size of the queues change, and in a real environment with more sustained load we should see these metrics more accurately represent the state of the Overseer.

Testing confirmed that we don't see the Overseer metrics registry in non-cloud mode.

**/admin/metrics response containing added metrics**
Screen Shot 2020-10-27 at 5 56 48 PM

**local testing sending CreateCollection request load and seeing metrics update in real time**
image

# Checklist


solr/core/src/java/org/apache/solr/cloud/Overseer.java
```diff
@@ -173,6 +175,9 @@
 
     private final Stats zkStats;
 
+    private final SolrMetricManager metricManager = new SolrMetricManager();
```
sigram, line 178: This is wrong - you should never create instances of SolrMetricManager in your components. Always use the single instance constructed in `CoreContainer.getMetricManager()`.

solr/core/src/java/org/apache/solr/cloud/Overseer.java
```diff
@@ -185,6 +190,9 @@ public ClusterStateUpdater(final ZkStateReader reader, final String myId, Stats
       this.completedMap = getCompletedMap(zkClient);
       this.myId = myId;
       this.reader = reader;
+
+      registryName = SolrMetricManager.getRegistryName(SolrInfoBean.Group.overseer);
+      metricManager.registerGauge(null, registryName, () -> stateUpdateQueue.getZkStats().getQueueLength(), "overseer", true, "stateUpdateQueueSize", "queue");
```
sigram, line 195: You should create a `SolrMetricsContext` in Overseer, and create a child context here, and then pass it as the first argument to `registerGauge`. This helps to properly unregister old gauges from the registry (because a new instance of `ClusterStateUpdater` is created each time this Overseer instance becomes the leader, and then it registers a new instance of gauge with the same name). Using a static string "overseer" defeats this mechanism, because it falsely indicates that the life-cycle of `ClusterStateUpdater` is the same as its parent object Overseer.

Then in the `ClusterStateUpdater.close()` method you should unregister this gauge using the value of `solrMetricsContext.getTag()`. This prevents obscure reference leaks in the metrics registry.

solr/core/src/java/org/apache/solr/cloud/Overseer.java
```diff
@@ -544,6 +552,7 @@ private LeaderStatus amILeader() {
     @Override
       public void close() {
         this.isClosed = true;
+        metricManager.unregisterGauges(registryName, "overseer");
```
sigram, line 555: Please use the context tag value here.

solr/core/src/java/org/apache/solr/cloud/OverseerTaskProcessor.java
```diff
@@ -94,6 +96,8 @@
   private boolean isClosed;
 
   private volatile Stats stats;
+  private final SolrMetricManager metricManager = new SolrMetricManager();
```
sigram, line 99: See above - this should never be done.

solr/core/src/java/org/apache/solr/cloud/OverseerTaskProcessor.java
```diff
@@ -154,6 +158,9 @@ public OverseerTaskProcessor(ZkStateReader zkStateReader, String myId,
     this.runningTasks = ConcurrentHashMap.newKeySet();
     this.completedTasks = new ConcurrentHashMap<>();
     thisNode = Utils.getMDCNode();
+
+    registryName = SolrMetricManager.getRegistryName(SolrInfoBean.Group.overseer);
+    metricManager.registerGauge(null, registryName, () -> workQueue.getZkStats().getQueueLength(), "overseer", true, "collectionWorkQueueSize", "queue");
```
sigram, line 163: See above comments - you should create a child context here (from the one created in Overseer) and use it in this call, and then unregister this gauge using this context's tag value.

solr/core/src/java/org/apache/solr/cloud/OverseerTaskProcessor.java
```diff
@@ -386,6 +393,7 @@ private void cleanUpWorkQueue() throws KeeperException, InterruptedException {
 
   public void close() {
     isClosed = true;
+    metricManager.unregisterGauges(registryName, "overseer");
```
sigram, line 396: See above - use the tag value from the context.

solr/core/src/java/org/apache/solr/metrics/SolrMetricManager.java
```diff
@@ -423,13 +429,13 @@ public boolean hasRegistry(String name) {
   /**
    * Check for predefined shared registry names. This compares the input name
    * with normalized names of predefined shared registries -
-   * {@link #JVM_REGISTRY} and {@link #JETTY_REGISTRY}.
+   * {@link #JVM_REGISTRY}, {@link #JETTY_REGISTRY}, and {@link #OVERSEER_REGISTRY}
```
sigram, line 432: This is wrong - Overseer registry CANNOT be a shared registry because it's possible to have multiple Overseers in a single JVM, especially in unit tests. This should be a regular per-CoreContainer registry, which is created within the scope of the instance available from CC.getMetricManager().

solr/core/src/java/org/apache/solr/metrics/SolrMetricManager.java
```diff
@@ -423,13 +429,13 @@ public boolean hasRegistry(String name) {
...
    *
    * @param registry already normalized name
    * @return true if the name matches one of shared registries
    */
   private static boolean isSharedRegistry(String registry) {
-    return JETTY_REGISTRY.equals(registry) || JVM_REGISTRY.equals(registry);
+    return JETTY_REGISTRY.equals(registry) || JVM_REGISTRY.equals(registry) || OVERSEER_REGISTRY.equals(registry);
```
sigram, line 438: See above - this can't be a shared registry.

sigram (changes requested):

solr/core/src/java/org/apache/solr/cloud/Overseer.java
saatchibhalla: Thanks for the feedback, will update accordingly.

megancarey: @sigram Thanks for the feedback! Is there documentation around adding metrics that covers things like this?

saatchibhalla: @sigram Updated the PR with those proposed changes. Whenever you're free to take another look I'd appreciate it!

solr/core/src/java/org/apache/solr/cloud/Overseer.java
```diff
@@ -636,7 +650,7 @@ public synchronized void start(String id) {
     ThreadGroup ccTg = new ThreadGroup("Overseer collection creation process.");
 
     OverseerNodePrioritizer overseerPrioritizer = new OverseerNodePrioritizer(reader, getStateUpdateQueue(), adminPath, shardHandler.getShardHandlerFactory());
-    overseerCollectionConfigSetProcessor = new OverseerCollectionConfigSetProcessor(reader, id, shardHandler, adminPath, stats, Overseer.this, overseerPrioritizer);
+    overseerCollectionConfigSetProcessor = new OverseerCollectionConfigSetProcessor(reader, id, shardHandler, adminPath, stats, Overseer.this, overseerPrioritizer, solrMetricsContext);
```
megancarey, line 653: Are there any tests that instantiate OverseerCollectionConfigSetProcessor? (Just want to make sure this is the only time this method is used, since we changed the signature)

solr/core/src/java/org/apache/solr/cloud/OverseerTaskProcessor.java
```diff
@@ -134,13 +136,14 @@ public String toString() {
...
-                                        DistributedMap completedMap,
-                                        DistributedMap failureMap) {
+                               Stats stats,
+                               OverseerMessageHandlerSelector selector,
+                               OverseerNodePrioritizer prioritizer,
+                               OverseerTaskQueue workQueue,
+                               DistributedMap runningMap,
+                               DistributedMap completedMap,
+                               DistributedMap failureMap,
+                               SolrMetricsContext solrMetricsContext) {
```
megancarey, line 146: Same question here - would be good to double-check that this is the only place we instantiate OverseerTaskProcessor

solr/solr-ref-guide/src/metrics-reporting.adoc
```diff
@@ -53,6 +53,12 @@ This registry is returned at `solr.jvm` and includes the following information.
 * System properties such as Java information, various installation directory paths, ports, and similar information. You can control what appears here by modifying `solr.xml`.
 // TODO for 7.0 fix this
 
+=== Overseer Registry
```
megancarey, line 56: We can update this doc comment now, since we are using the CoreContainer registry

saatchibhalla: chose to leave it this way - although we're using MetricManager from CoreContainer i'm still registering these metrics under "solr.overseer."

solr/core/src/java/org/apache/solr/cloud/Overseer.java
saatchibhalla: Updated the one test that calls this thanks for the catch!

megancarey (approved):
Looks good to me!

saatchibhalla:
@sigram Would you mind taking another look at this PR whenever you have a chance?

saatchibhalla:
> Saatchi; could you please comment here what this change in the config is about? Did you blend two metrics into one, and if so why? Maybe could you share a snippet of the prometheus exporter's output so we know clearly what the output of this "looks like"

Hey David, sorry for the delay in getting back to you. I looked at this again, and decided that the two overseer queue metrics should be exported as separate metrics (rather than grouping them together under a single exported metric). Each of these queues represent separate responsibilities of the Overseer, and this allows for more flexibility in the future to add more metrics for each queue and group those together per queue. 

Based on the current status of solr-exporter.xml this is how the exported metrics will look:
{
  "name": "solr_metrics_overseer_stateUpdateQueueSize",
  "type": "GAUGE",
  "help": "See following URL: https://lucene.apache.org/solr/guide/metrics-reporting.html",
  "label_names": [],
  "label_values": [],
  "value": 0
}
{
  "name": "solr_metrics_overseer_collectionWorkQueueSize",
  "type": "GAUGE",
  "help": "See following URL: https://lucene.apache.org/solr/guide/metrics-reporting.html",
  "label_names": [],
  "label_values": [],
  "value": 0
}

solr/core/src/java/org/apache/solr/cloud/Overseer.java
sigram: @megancarey unfortunately, there's not much documentation apart from the javadocs and user-level info in the RefGuide. We should create a doc in solr/dev-docs/metrics.

dsmiley:
Looks good now Saatchi!  I'll merge tomorrow with the following CHANGES.txt note under "Improvements":

    SOLR-14965: metrics: Adds two metrics to the SolrCloud Overseer: solr_metrics_overseer_stateUpdateQueueSize
    and solr_metrics_overseer_collectionWorkQueueSize with corresponding entries in the the Prometheus exporter's
    default/stock configuration.  (Saatchi Bhalla, Megan Carey, Andrzej Białecki, David Smiley)

@sigram if you prefer to or if you have lingering concerns, let me know.

saatchibhalla:
Thanks @dsmiley! Looks good to me.

PR #2051, opened 2020-10-30
LUCENE-9594 Add linear function for FeatureField

mayya-sharipova:
This adds a linear function and newLinearQuery for FeatureField

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/document/FeatureField.java
+++ b/lucene/core/src/java/org/apache/lucene/document/FeatureField.java
@@ -66,8 +66,11 @@ import org.apache.lucene.search.similarities.Similarity.SimScorer;
  * 2<sup>-8</sup> = 0.00390625.
  * <p>
  * Given a scoring factor {@code S > 0} and its weight {@code w > 0}, there
- * are three ways that S can be turned into a score:
+ * are four ways that S can be turned into a score:
  * <ul>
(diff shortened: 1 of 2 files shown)


lucene/core/src/java/org/apache/lucene/document/FeatureField.java
```diff
@@ -66,8 +66,11 @@
  * 2<sup>-8</sup> = 0.00390625.
  * <p>
  * Given a scoring factor {@code S > 0} and its weight {@code w > 0}, there
- * are three ways that S can be turned into a score:
+ * are four ways that S can be turned into a score:
  * <ul>
+ *   <li>{@link #newLinearQuery w * S}. This is the simplest function
+ *       where no transformation is applied on the feature value, and
+ *       the feature value itself multiplied by weight defines the score.
```
jpountz, line 73: "the simplest" might suggest that this would be a good one to start with, when in fact I think this is the most expert function in my opinion given that it expects the feature to already be encoded in the index in a way that makes sense for scoring. Maybe javadocs should better convey that this is expert functionality?

jpountz (commented):
This makes sense to me, but I think the javadocs should be clearer that this is expert functionality as it makes some strong assumptions on how features are encoded in the index?

mayya-sharipova:
@jpountz Thank you for the feedback, it makes sense. Addressed in 2dfa5511

jpountz (approved):

PR #2055, opened 2020-11-02
SOLR-14978 OOM Killer in Foreground

madrob:
https://issues.apache.org/jira/browse/SOLR-14978


solr/bin/solr
```diff
@@ -2172,6 +2172,12 @@ function start_solr() {
     SOLR_OPTS+=($AUTHC_OPTS)
   fi
 
+  # If a heap dump directory is specified, enable it in SOLR_OPTS
+  if [[ -n "$SOLR_HEAP_DUMP_DIR" ]]; then
```
HoustonPutman, line 2176: I kind of like the idea of having a boolean flag to turn on heap dumps, and have a default location in the logs directory. This logic would be fine, but having an additional check above that checks that `SOLR_HEAP_DUMP_DIR` is empty and `SOLR_HEAP_DUMP` is true, then sets `SOLR_HEAP_DUMP_DIR="${SOLR_LOG_DIR}/dumps"`.

Just a thought, for ease of use.

solr/docker/include/scripts/solr-fg
```diff
@@ -15,21 +15,19 @@ if [[ -z "${OOM:-}" ]]; then
   OOM='none'
 fi
 case "$OOM" in
```
HoustonPutman, line 17: Should this also be moved to bin/solr?

solr/bin/solr
```diff
@@ -2172,6 +2172,12 @@ function start_solr() {
     SOLR_OPTS+=($AUTHC_OPTS)
   fi
```
HoustonPutman, line 2174:
```suggestion

# If a heap dump directory is specified, enable it in SOLR_OPTS
if [[ -z "$SOLR_HEAP_DUMP_DIR" ]] && [[ "$SOLR_HEAP_DUMP" == "true" ]]; then
  SOLR_HEAP_DUMP_DIR="${SOLR_LOGS_DIR}/dumps"
fi

```

I like the idea of having a boolean flag that lets you say that you want a heap dump at the default location of heap dumps. Not a necessity, just a thought.

solr/docker/include/scripts/solr-fg
madrob: I don't think so. It might make sense to remove it entirely, but in bin/solr these options can be specified directly instead.

HoustonPutman: I wonder what effect these even have if we are going to run the oom_script by default, which will kill the solr process

madrob: After some experimentation, I don't think these do anything.

solr/bin/solr
```diff
@@ -2172,6 +2172,16 @@ function start_solr() {
...
 
+  # If a heap dump directory is specified, enable it in SOLR_OPTS
+  if [[ -z "$SOLR_HEAP_DUMP_DIR" ]] && [[ "$SOLR_HEAP_DUMP" == "true" ]]; then
+    SOLR_HEAP_DUMP_DIR="${SOLR_LOGS_DIR}/dumps"
+  fi
+  if [[ -n "$SOLR_HEAP_DUMP_DIR" ]]; then
+    SOLR_OPTS+=("-XX:+HeapDumpOnOutOfMemoryError")
+    SOLR_OPTS+=("-XX:HeapDumpPath=$SOLR_HEAP_DUMP_DIR/solr-$(date +%s)-pid$$.hprof")
+    mkdir -p "$SOLR_HEAP_DUMP_DIR" 2>/dev/null
```
HoustonPutman, line 2182: Why are you piping the errors here? `mkdir -p` should be safe right?

HoustonPutman (approved):
Tested in multiple configurations, works very well!
+1 from me

One question that isn't super important, I was just curious.

solr/bin/solr
madrob: I was doing it to match the call at https://github.com/apache/lucene-solr/blob/8f75d6ff4f0d981136f64a28c173b5ac4a7263ec/solr/bin/solr#L2243 but that also had additional checks afterwards. I extracted the commonality here and call that function instead.

mkdir can fail if you don't have permissions to that directory path, which would also indicate that the process won't have permissions later to write the log files.

PR #2056, opened 2020-11-02
SOLR-14971: Handle atomic-removes on uncommitted docs

gerlowskija:
# Description

Atomic Update 'remove' operations currently fail on several numeric fieldtypes when the "existing" doc hasn't been committed yet and is fetched from the UpdateLog.

# Solution

Modify logic for "remove" operation in AtomicUpdateDocumentMerger to check for comparable but different types (e.g. Long vs Integer).

# Tests

Manual testing for XML/JSON/javabin requests.  Automated tests for Javabin and XML formats. 

# Checklist


munendrasn:
> As PR is open, sharing the comment here instead of JIRA

I was able to reproduce the issue with the script shared in JIRA. Regarding the approach, I agree SolrInputDocument returning field values in the actual type even for uncommitted docs is a better solution (approach 1) but currently not sure of the impact

I was able to reproduce the same issue with `add-distinct` atomic operation too. `add`, `set` doesn't have this problem. I think `removeregex` might have the same issue but I'm assuming this operation would be used with string/text fields, so probably not an issue? 🤔 

Coming to the PR, 
* This doesn't handle the case when the original values are passed as string. Suppose `values: ["2", "3"]`, then uncommitted doc would contain values as a string (This can happen if the users are using XML format (older version of solrJ sends update docs in XML format, not sure of the latest). In such cases, it will still not work
* would the PR support date fields type too?

I think to handle these cases, we need to convert the existing values to native types using
Something like this but not sure of performance overhead in case of large fields
```java
Collection<Object> original = existingField.getValues();
original.stream().map(val -> sf.getType().toNativeType(object)).collect(Collectors.toList())
```

gerlowskija:
I thought of doing a `toNativeType` conversion on the "original" Collection, but didn't like how it changed SolrInputDocument values for field-vals supposedly unaffected by the operation (e.g. an atomic remove of 1 of 5 values would change all 5 to native-types).  I was leery of what problems that might cause downstream.

But if there are input edge cases I'm missing in the current iteration, maybe it's worth revisiting that approach.  I'll take a look at that and some of the other things you mentioned (esp. add-distinct reproduction).

gerlowskija:
Hey @munendrasn new code should address your concerns:

- handling for the string values that come through when requests are made using XML.
- handling for the "add-distinct" class of operations.
- additional tests to cover the two cases above.

I ended up _not_ going with the approach you suggested above of using `toNativeType()` to convert the whole list of "original" values, primarily for the performance concern I mentioned above.  I'm not in love with the code I've got now though - the type checks are pretty ugly.  So maybe I'll flip on this after a bit more consideration.

gerlowskija:
Deciding to keep the implementation as-is for now.  We can revisit if other type-related bugs pop up, or anyone has insight into what the downstream effects of the bulk `toNativeType` call might be on uncommitted docs.

One last note, in writing the tests for this PR I noticed that we have an abundance of test files for atomic updates, with a lot of overlap among them.  I didn't want to dive into sorting those out here, but I'm going to file a follow up ticket for cutting out some of the overlap among these.

PR #2059, opened 2020-11-03
LUCENE-9595: Make Component2D#withinPoint implementations consistent with ShapeQuery logic

iverase:
This commit changes the logic of Component2D#withinPoint to return WithinRelation.NOTWITHIN when the point is inside the component.


iverase:
@nknize Do you have an opinion about this change?

PR #2062, opened 2020-11-04
LUCENE-9589 Swedish minimal stemmer

janhoy:
See https://issues.apache.org/jira/browse/LUCENE-9589

This impl is based on [SwedishLightStemmer](https://github.com/apache/lucene-solr/blob/master/lucene/analysis/common/src/java/org/apache/lucene/analysis/sv/SwedishLightStemmer.java), concentrating on the plural endings only, inspired by [NorwegianMinimalStemmer](https://github.com/apache/lucene-solr/blob/master/lucene/analysis/common/src/java/org/apache/lucene/analysis/no/NorwegianMinimalStemmer.java).

Some of the examples tested in `minimal.txt` are fetched from https://en.wikipedia.org/wiki/Swedish_grammar, not any scientific rule book of any kind.

### DIFF
--- a/lucene/analysis/common/src/java/org/apache/lucene/analysis/sv/SwedishMinimalStemFilterFactory.java
+++ b/lucene/analysis/common/src/java/org/apache/lucene/analysis/sv/SwedishMinimalStemFilterFactory.java
@@ -1,0 +1,60 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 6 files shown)


rmuir (approved):
this looks great, a toned-down version of the light stemmer.

lucene/analysis/common/src/java/org/apache/lucene/analysis/sv/SwedishMinimalStemFilterFactory.java
```diff
@@ -0,0 +1,60 @@
...
+
+
+import org.apache.lucene.analysis.TokenFilterFactory;
+import org.apache.lucene.analysis.TokenStream;
+
+import java.util.Map;
+
+/** 
+ * Factory for {@link SwedishMinimalStemFilter}.
+ * <pre class="prettyprint">
```
dweiss, line 27: Would code tag read better here?
https://reflectoring.io/howto-format-code-snippets-in-javadoc/#pre--code

janhoy: Peraps, but all the other zillion factories use pre, and btw we're going to get rid fo solr xml syntax anyway and replace it with some other way of documenting options. Let's defer all of that to LUCENE-7964.

PS. I found a typo in the field-type name in line 28, which I changed from `name="text_svlgtstem"` to `name="text_svminstem"`. I'll also add `@since` tags to the other classes.

dweiss: You have to use both, actually. The code tag only helps you to dodge HTML entity escaping (<>) and makes the code easier on the eyes, that's all.

janhoy:
Closing in favor of PR in new repo

PR #2063, opened 2020-11-04
LUCENE-9599 Make comparator aware of index sorting

mayya-sharipova:
Currently, if search sort is equal to index sort,  we have an early
termination in TopFieldCollector. As we work to enhance comparators
to provide skipping functionality (PR #1351), we would like to
move this termination functionality on index sort from
TopFieldCollector to comparators.

This patch does the following:
- Add method usesIndexSort to LeafFieldComparator
- Make numeric comparators aware of index sort and early terminate on
  collecting all competitive hits
- Move TermValComparator and TermOrdValComparator from FieldComparator
  to comparator package, for all comparators to be in the same package
- Enhance TermOrdValComparator to provide skipping functionality when
  index is sorted

One item left for TODO for a following PR is to remove the logic of
early termination from TopFieldCollector. We can do that once
we ensure that all BulkScorers are using iterators from collectors
that can skip non-competitive docs.

Relates to #1351


lucene/core/src/java/org/apache/lucene/search/comparators/TermOrdValComparator.java
```diff
@@ -0,0 +1,324 @@
...
+ * {@link org.apache.lucene.index.LeafReader#getSortedDocValues(String)}),
+ * and does most comparisons using the ordinals.
+ * For medium to large results, this comparator will be much faster
+ * than {@link org.apache.lucene.search.comparators.TermValComparator}.
+ * For very small result sets it may be slower.
+ *
+ * The comparator provides an iterator that can efficiently skip
+ * documents when search sort is done according to the index sort.
+ */
+public class TermOrdValComparator extends FieldComparator<BytesRef> {
```
msokolov, line 47: Is this just moved to top-level class, or were there changes? It's hard to tell in the PR

msokolov: Oh I see reverse was added..

lucene/core/src/java/org/apache/lucene/search/comparators/TermOrdValComparator.java
```diff
@@ -0,0 +1,324 @@
...
+      } else if (bottomOrd >= docOrd) {
+        // the equals case always means bottom is > doc
+        // (because we set bottomOrd to the lower bound in setBottom):
+        result = 1;
+      } else {
+        result = -1;
+      }
+      // for the index sort case, if we encounter a first doc that is non-competitive,
+      // and the hits threshold is reached, we can update the iterator to skip the rest of docs
+      if (indexSort && (reverse ? result >= 0 : result <= 0)) {
```
msokolov, line 224: oh it looks like there was some new stuff here related to indexSort. I wonder if it would be possible to separate the change moving these classes to top level for the sake of review? If it's a pain, perhaps you could just comment on what the changes are here?

lucene/core/src/test/org/apache/lucene/search/TestFieldSortOptimizationSkipping.java
```diff
@@ -485,4 +491,186 @@ public void testDocSort() throws IOException {
     dir.close();
   }
 
+  public void testNumericSortOptimizationIndexSort() throws IOException {
```
msokolov, line 494: Do we have a test for the case where the index sort has multiple Sorts in it and the query sort matches its first n components?

msokolov (commented):
Thanks for tackling; it should be great to unify this with our approach to WAND!

Initially I had some concerns about how we would handle the multi-threaded case (I have an optimization that shares some information across threads coordinated by the Collector that never got pushed), but I think we could implement it in this framework too.

lucene/core/src/java/org/apache/lucene/search/comparators/TermOrdValComparator.java
mayya-sharipova: @msokolov Thank you for the review and feedback. Sorry, indeed it difficult to see the changes this way. In commits f14f11371c8d and 98a02981e80e, I did as you suggested, I kept `TermOrdValComparator` and `TermValComparator` in FieldComparator, but still there are substantial changes,  so I will summarize them here:

-  Before `TermOrdValComparator` and `TermValComparator` were extending FieldComparator<BytesRef>  and also implementing LeafFieldComparator.  Now each of these comparators have their corresponding inner classes that implement `LeafFieldComparator` and all methods related to leaf comparator were moved there.
- `TermOrdValComparator` was enhanced to provide skipping functionality when index is sorted

lucene/core/src/test/org/apache/lucene/search/TestFieldSortOptimizationSkipping.java
mayya-sharipova: @msokolov Thanks for the feedback. Addressed in d2909aa

lucene/core/src/test/org/apache/lucene/search/TestFieldSortOptimizationSkipping.java
```diff
@@ -485,4 +491,249 @@ public void testDocSort() throws IOException {
...
+      if (sortedValues[randomIdxs[i]] > 0) {
+        doc.add(new NumericDocValuesField("field1", sortedValues[randomIdxs[i]]));
+      }
+      writer.addDocument(doc);
+      if (i == 30) {
+        writer.flush();
+      }
+    }
+    final IndexReader reader = DirectoryReader.open(writer);
+    writer.close();
```
msokolov, line 518: If we use try-with-resources, we won't need the close() and we will safely close in case of unexpected failures

lucene/core/src/test/org/apache/lucene/search/TestFieldSortOptimizationSkipping.java
```diff
@@ -485,4 +491,249 @@ public void testDocSort() throws IOException {
...
+
+    IndexSearcher searcher = newSearcher(reader);
+    final int numHits = randomIntBetween(1, numDocs - 10);
+    final int totalHitsThreshold = randomIntBetween(1, numDocs - 10);
+    {
```
msokolov, line 523: why the extra block?

lucene/core/src/test/org/apache/lucene/search/TestFieldSortOptimizationSkipping.java
```diff
@@ -485,4 +491,249 @@ public void testDocSort() throws IOException {
...
+      // test that optimization is run when search sort is equal to the index sort
+      TopFieldCollector collector = TopFieldCollector.create(indexSort, numHits, null, totalHitsThreshold);
+      searcher.search(new MatchAllDocsQuery(), collector);
+      TopDocs topDocs = collector.topDocs();
+      assertTrue(collector.isEarlyTerminated());
```
msokolov, line 528: does this test distinguish between the kind of early termination we had before, and what this provides?

lucene/core/src/test/org/apache/lucene/search/TestFieldSortOptimizationSkipping.java
msokolov: Oh I see, you can scope the variables - nice! Although in this case it seems unneccessary

msokolov (commented):
Phew this is a big change- I don't pretend to understand all of it, but I read through and added some minor comments and questions

lucene/core/src/java/org/apache/lucene/search/TopFieldCollector.java
```diff
@@ -100,6 +100,7 @@ boolean thresholdCheck(int doc) throws IOException {
         // since docs are visited in doc Id order, if compare is 0, it means
         // this document is largest than anything else in the queue, and
         // therefore not competitive.
+        // TODO: remove early termination in TopFieldCollector, as this should be managed by comparators
```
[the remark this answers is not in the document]
mayya-sharipova: Indeed, this is a no-op operation, and will be removed in a following PR.
I left it for now, as we also need to ensure that all necessary BulkScorers use collectors/comparators iterators.

lucene/core/src/java/org/apache/lucene/search/TopFieldCollector.java
```diff
@@ -139,6 +140,7 @@ void collectAnyHit(int doc, int hitsCollected) throws IOException {
 
     @Override
     public void setScorer(Scorable scorer) throws IOException {
+      if (canEarlyTerminate) comparator.usesIndexSort();
```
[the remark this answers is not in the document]
mayya-sharipova: Addressed in c52d9a7

lucene/core/src/java/org/apache/lucene/search/comparators/NumericComparator.java
```diff
@@ -218,7 +237,7 @@ public void visit(int docID, byte[] packedValue) {
 
         @Override
         public DocIdSetIterator competitiveIterator() {
-            if (enableSkipping == false) return null;
+            if (enableSkipping == false && indexSort == false) return null;
```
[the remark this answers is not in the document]
mayya-sharipova: Addressed in c52d9a7.  I also have a plan to a do a  separate PR to just fix formatting.

lucene/core/src/test/org/apache/lucene/search/TestFieldSortOptimizationSkipping.java
mayya-sharipova: Thank you, good suggestion.  Addressed in c52d9a7

lucene/core/src/test/org/apache/lucene/search/TestFieldSortOptimizationSkipping.java
mayya-sharipova: Addressed in c52d9a7

lucene/core/src/java/org/apache/lucene/search/FieldComparator.java
```diff
@@ -387,125 +346,225 @@ protected SortedDocValues getSortedDocValues(LeafReaderContext context, String f
...
+        } else {
+          result = -1;
+        }
+        // for the index sort case, if we encounter a first doc that is non-competitive,
+        // and the hits threshold is reached, we can update the iterator to skip the rest of docs
+        if (indexSort && (reverse ? result >= 0 : result <= 0)) {
+          collectedAllCompetitiveHits = true;
+          if (hitsThresholdReached && iteratorUpdated == false) {
+            competitiveIterator = DocIdSetIterator.empty();
+            iteratorUpdated = true;
```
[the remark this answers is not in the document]
mayya-sharipova: `iteratorUpdated` variable removed in c52d9a7.    
`iteratorUpdated`, similarly to early termination in `TopFieldCollector` is not necessary and just extra precaution, for cases if there are some BulkScores that don't use iterators from collectors. 

> Could we instead create a class constant EMPTY iterator

We can't use a static iterator as we advance it once, it because invalid for all other instances.

lucene/core/src/java/org/apache/lucene/search/FieldComparator.java
```diff
@@ -387,125 +346,225 @@ protected SortedDocValues getSortedDocValues(LeafReaderContext context, String f
...
-    @Override
-    public void setScorer(Scorable scorer) {}
   }
   
   /** Sorts by field's natural Term sort order.  All
    *  comparisons are done using BytesRef.compareTo, which is
    *  slow for medium to large result sets but possibly
    *  very fast for very small results sets. */
-  public static class TermValComparator extends FieldComparator<BytesRef> implements LeafFieldComparator {
+  public static class TermValComparator extends FieldComparator<BytesRef> {
```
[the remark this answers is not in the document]
mayya-sharipova: It seems that we [don't support index sort](https://github.com/apache/lucene-solr/blob/master/lucene/core/src/java/org/apache/lucene/search/SortField.java#L524) on `STRING_VAL` type  that will use `TermValComparator` comparator,  thus early termination on index sort is not necessary for this class.

lucene/core/src/test/org/apache/lucene/search/TestFieldSortOptimizationSkipping.java
mayya-sharipova: You are right, they don't distinguish this. The idea that if we remove early termination in TopFieldCollector, these tests should still work.

If all this confusing, I can also upgrade this PR to remove early termination in TopFieldCollector.

mayya-sharipova:
@msokolov Thanks a lot for  another round of review, I tried to address it in c52d9a73

msokolov (approved):
LGTM, thanks!

jimczi (changes requested):
I wonder if we should keep the logic for early termination based on index sorting in the top field collector. We don't really need to move this code, we could just disable the numeric sort optimization when the query sort is congruent with the index sort ? That feels simpler to keep these two optimizations separated since the index sort can be applied on multiple fields (and not only the primary sort). 
Does that makes sense @mayya-sharipova ?

mayya-sharipova:
@msokolov @jimczi Thank you for the review.

As per @jimczi's latest comment, since it makes more sense to disable sort optimization on index sort in comparators rather then move this logic from `TopFieldCollector` to comparators, I am closing this PR,  and I have opened a new much smaller  [PR](https://github.com/apache/lucene-solr/pull/2075)  that will implement this plan. 

Sorry for the trouble.

PR #2064, opened 2020-11-05
LUCENE-9600: Clean up package name conflicts between misc and core modules

mocobeta:
This refactors misc module to resolve package name conflicts between misc and lucene-core.

- moves `o.a.l.document`, `o.a.l.index`, `o.a.l.search`, `o.a.l.store`, `o.a.l.util`, and their sub-packages under `o.a.l.misc`.
- made some refactorings to remove unnecessary accesses to package-private methods in `lucene-core`.
- made methods in o.a.l.index.DocHelper and o.a.l.util.FSTTester in test-framework publicly accessible for unit testing.
- lucene-core isn't touched.


lucene/misc/src/java/org/apache/lucene/misc/index/IndexSplitter.java
```diff
@@ -123,8 +118,8 @@ private SegmentCommitInfo getInfo(String name) {
 
   public void remove(String[] segs) throws IOException {
     for (String n : segs) {
-      int idx = getIdx(n);
-      infos.remove(idx);
+      SegmentCommitInfo info = getInfo(n);
+      infos.remove(info);
```
mocobeta, line 122: SegmentInfos.remove(int) isn't accessible from external packages. This method call can be switched to SegmentInfo.remove(SegmentCommitInfo); that is already opened to public.

lucene/misc/src/java/org/apache/lucene/misc/index/MultiPassIndexSplitter.java
```diff
@@ -101,7 +111,7 @@ public void split(IndexReader in, Directory[] outputs, boolean seq) throws IOExc
           .setOpenMode(OpenMode.CREATE));
       System.err.println("Writing part " + (i + 1) + " ...");
       // pass the subreaders directly, as our wrapper's numDocs/hasDeletetions are not up-to-date
-      final List<? extends FakeDeleteLeafIndexReader> sr = input.getSequentialSubReaders();
+      final List<? extends FakeDeleteLeafIndexReader> sr = input.getSequentialSubReadersWrapper();
```
mocobeta, line 114: BaseCompositeReader.getSequentialSubReaders() is a protected method so it cannot be directly called from here. A wrapper method is needed to move this class to `o.a.l.misc.index` from `o.a.l.index`.

lucene/misc/src/test/org/apache/lucene/misc/index/TestIndexSplitter.java
```diff
@@ -60,9 +67,9 @@ public void test() throws Exception {
       iw.addDocument(doc);
     }
     iw.commit();
-    DirectoryReader iwReader = iw.getReader();
-    assertEquals(3, iwReader.leaves().size());
-    iwReader.close();
+    //DirectoryReader iwReader = iw.getReader();
+    //assertEquals(3, iwReader.leaves().size());
+    //iwReader.close();
```
mocobeta, line 72: IndexWriter.getReader() is package-private so it cannot be accessed from here. It would be better that we omit this assertion (it seems these lines are not the main part of the test anyway), instead of opening IW.getReader() up to public?

lucene/test-framework/src/java/org/apache/lucene/util/fst/FSTTester.java
```diff
@@ -95,7 +95,7 @@ private static BytesRef toBytesRef(IntsRef ir) {
     return br;
   }
 
-  static String getRandomString(Random random) {
+  public static String getRandomString(Random random) {
```
mocobeta, line 98: Looks like this should be publicly accessible for tests in any modules?

lucene/test-framework/src/java/org/apache/lucene/util/fst/FSTTester.java
```diff
@@ -121,7 +121,7 @@ static String simpleRandomString(Random r) {
     return new String(buffer, 0, end);
   }
 
-  static IntsRef toIntsRef(String s, int inputMode) {
+  public static IntsRef toIntsRef(String s, int inputMode) {
```
mocobeta, line 124: Looks like this should be publicly accessible for tests in any modules?

lucene/test-framework/src/java/org/apache/lucene/index/DocHelper.java
```diff
@@ -35,7 +35,8 @@
 import org.apache.lucene.search.similarities.Similarity;
 import org.apache.lucene.store.Directory;
 
-class DocHelper {
+/** Helper functions for tests that handles documents */
+public class DocHelper {
```
mocobeta, line 39: Looks like this should be publicly accessible for tests in any modules?

lucene/misc/src/test/org/apache/lucene/misc/index/TestIndexSplitter.java
dweiss: Or maybe switch to DirectoryReader.open(iw...) and keep the assertion? This is public.

lucene/test-framework/src/java/org/apache/lucene/util/fst/FSTTester.java
dweiss: Don't know what it actually does but randomizedtesting has a number of utility methods for random strings:

http://labs.carrotsearch.com/download/randomizedtesting/2.0.0/api/randomizedtesting-runner/com/carrotsearch/randomizedtesting/generators/RandomStrings.html

lucene/misc/src/test/org/apache/lucene/misc/index/TestIndexSplitter.java
mocobeta: Thanks! DirectoryReader.open(iw) and IW.getReader() are actually equivalent. https://github.com/apache/lucene-solr/pull/2064/commits/4858a5d8347c72364c4a6572345a3fc52f43cc77

lucene/test-framework/src/java/org/apache/lucene/util/fst/FSTTester.java
mocobeta: Yes, they look similar except for that Lucene's TestUtil selects different codepoint range (`blockStarts` and `blockEnds`) to generate random String instance  each time. I think both one is okay for the test, what do you think?
 
o.a.l.util.TestUtil.randomRealisticUnicodeString()
```
  /** Returns random string of length between min and max codepoints, all codepoints within the same unicode block. */
  public static String randomRealisticUnicodeString(Random r, int minLength, int maxLength) {
    final int end = nextInt(r, minLength, maxLength);
    final int block = r.nextInt(blockStarts.length);
    StringBuilder sb = new StringBuilder();
    for (int i = 0; i < end; i++)
      sb.appendCodePoint(nextInt(r, blockStarts[block], blockEnds[block]));
    return sb.toString();
  }
```

com.carrotsearch.randomizedtesting.generators.UnicodeGenerator.ofCodePointsLength()
```
  public String ofCodePointsLength(Random r, int minCodePoints, int maxCodePoints) {
    int length = RandomNumbers.randomIntBetween(r, minCodePoints, maxCodePoints);
    int [] chars = new int [length];
    for (int i = 0; i < chars.length; i++) {
      int v = RandomNumbers.randomIntBetween(r, 0, CODEPOINT_RANGE);
      if (v >= Character.MIN_SURROGATE)
        v += SURROGATE_RANGE;
      chars[i] = v;
    }
    return new String(chars, 0, chars.length);
  }
```

dweiss: Not sure whether the difference matters. It shouldn't at first glance.

mocobeta: If there's no good reason to change, I'd like to keep the original test code here and make the test utility method publicly accessible (the change shouldn't hurt anything, to me) :)

mocobeta:
The changes look okay to me; the refactoring on misc module shouldn't change its behaviour while some classes are moved under o.a.l.misc package.
I'm going to merge it into the master after waiting for a couple of days, if there is no any comment/objection about the changes.

jpountz (approved):
LGTM.
We should think about moving InetAddressPoint to core.

lucene/test-framework/src/java/org/apache/lucene/util/fst/FSTTester.java
dsmiley: If you're going to make this method public... I think even a tiny bit of documentation would be helpful.  Like maybe recommend callers consider the methods Dawid mentioned in RandomizedTesting _instead_.  After all, we're only making this public for some test in misc to call it.  We aren't signaling to all of Lucene, Solr, and any of the plugin writers, that we've got some awesome random string generating that everyone should use :-)

FWIW I prefer that we maintain less code and use whatever's in RandomizedTesting when possible, but I won't stand in the way of your judgement here.

dsmiley (approved):

lucene/test-framework/src/java/org/apache/lucene/util/fst/FSTTester.java
mocobeta: I'd use RandomStrings instead of making the FSTTester method public and adding extra documentation for it (I prefer only one recommended way to implement one thing), if we can safely switch them. 

I checked the `FSTTester.getRandomString()` method if we can safely replace it with `RandomStrings.randomUnicodeOfCodepointLength()`. It seems there is a small, but intentional difference (it generates strings that consists of characters within the range of codepoint 97('a') to 102('f'), with a 50% possibility)... not fully sure the details, but it somewhat serves for testing FST?

https://github.com/apache/lucene-solr/blob/master/lucene/test-framework/src/java/org/apache/lucene/util/fst/FSTTester.java#L98-L122

mocobeta: RandomStrings might be okay here but I'm not so confident about switching to it because of the difference... i aded a note to the method's javadoc. https://github.com/apache/lucene-solr/pull/2064/commits/a0d97b595f8d01f81db6d1f068aa8ae11d47887e

PR #2065, opened 2020-11-06
SOLR-14977 :  ContainerPlugins should be configurable

noblepaul:


solr/core/src/java/org/apache/solr/api/ContainerPluginsRegistry.java
```diff
@@ -372,9 +396,27 @@ public void init() throws Exception {
         holders.add(new ApiHolder((AnnotatedApi) api));
       }
     }
+
+    @SuppressWarnings("rawtypes")
+    private Class getConfigObj(Object o) {
+      Type[] interfaces = o.getClass().getGenericInterfaces();
```
madrob, line 402: This is super fragile and fails if we have `Widget implements ConfigurablePlugin<WidgetInfo>` and then `NewerWidget extends Widget`

solr/core/src/java/org/apache/solr/api/ContainerPluginsRegistry.java
```diff
@@ -372,9 +396,27 @@ public void init() throws Exception {
         holders.add(new ApiHolder((AnnotatedApi) api));
       }
     }
+
+    @SuppressWarnings("rawtypes")
```
madrob, line 400: At the very least, don't suppress raw types here.

solr/core/src/java/org/apache/solr/api/ConfigurablePlugin.java
```diff
@@ -0,0 +1,31 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.solr.api;
+
+/**Implement this interface if your plugin needs to accept some configuration
+ * 
+ * @param <T> the configuration Object type
+ */
+public interface ConfigurablePlugin<T> {
```
madrob, line 24: Does T need to extend `PluginMeta` as you do in the unit test or no?

solr/core/src/java/org/apache/solr/api/ContainerPluginsRegistry.java
```diff
@@ -69,7 +72,8 @@
 public class ContainerPluginsRegistry implements ClusterPropertiesListener, MapWriter, Closeable {
   private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
 
-  private final ObjectMapper mapper = SolrJacksonAnnotationInspector.createObjectMapper();
+  private static final ObjectMapper mapper = SolrJacksonAnnotationInspector.createObjectMapper()
+      .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
```
madrob, line 76: Is this a security risk? Why do we need this?

solr/core/src/java/org/apache/solr/api/ConfigurablePlugin.java
noblepaul: No,it doe snot have to

solr/core/src/java/org/apache/solr/api/ContainerPluginsRegistry.java
noblepaul: This is not a security risk. In general it's better to ignore extra attributes for backcompat. Imagine adding an extra attribute to your plugin in a newer version and old Solr nodes failing because of that

solr/core/src/java/org/apache/solr/api/ContainerPluginsRegistry.java
noblepaul: Yeah, this can be improved

noblepaul: fixed

solr/core/src/java/org/apache/solr/api/ContainerPluginsRegistry.java
noblepaul: it's not required

solr/core/src/java/org/apache/solr/api/ConfigurablePlugin.java
```diff
@@ -0,0 +1,31 @@
...
+
+  /**This is invoked soon after the Object is initialized
+   * 
+   * @param cfg value deserialized from JSON
+   */
+  void initConfig(T cfg);
```
sigram, line 30: Maybe `configure(T cfg)`? the current name looks awkward.

solr/core/src/java/org/apache/solr/api/AnnotatedApi.java
```diff
@@ -222,7 +223,8 @@ public void call(SolrQueryRequest req, SolrQueryResponse rsp) {
     final String command;
     final MethodHandle method;
     final Object obj;
-    ObjectMapper mapper = SolrJacksonAnnotationInspector.createObjectMapper();
+    ObjectMapper mapper = SolrJacksonAnnotationInspector.createObjectMapper()
```
sigram, line 226: `ObjectMapper` is a relatively heavy object, we should not create new instances in every class that needs it - maybe put a common instance somewhere in Utils?

solr/core/src/java/org/apache/solr/api/ContainerPluginsRegistry.java
```diff
@@ -114,6 +118,16 @@ public synchronized ApiInfo getPlugin(String name) {
     return currentPlugins.get(name);
   }
 
+  static class PluginMetaHolder {
+    private final Map<String, Object> original;
+    private final PluginMeta meta;
```
sigram, line 123: Does `PluginMeta` still need a separate `pathPrefix`? I think this could become a config property.

solr/core/src/java/org/apache/solr/api/ContainerPluginsRegistry.java
```diff
@@ -349,7 +365,7 @@ public ApiInfo(PluginMeta info, List<String> errs) {
       }
     }
 
-    @SuppressWarnings({"rawtypes"})
+    @SuppressWarnings({"rawtypes","unchecked"})
```
sigram, line 368: Whitespace.

solr/core/src/java/org/apache/solr/api/ContainerPluginsRegistry.java
```diff
@@ -360,6 +376,14 @@ public void init() throws Exception {
       } else {
         throw new RuntimeException("Must have a no-arg constructor or CoreContainer constructor ");
       }
+      if (instance instanceof ConfigurablePlugin) {
+        Class c = getConfigClass((ConfigurablePlugin<?>) instance);
+        if(c != null) {
```
sigram, line 381: Whitespace.

solr/core/src/test/org/apache/solr/handler/TestContainerPlugin.java
```diff
@@ -312,6 +340,45 @@ public void testApiFromPackage() throws Exception {
...
+    @EndPoint(method = GET,
+        path = "/hello/plugin",
+        permission = PermissionNameProvider.Name.READ_PERM)
+    public void m2(SolrQueryRequest req, SolrQueryResponse rsp) {
+      rsp.add("config", cfg);
+    }
+
+  }
+
+  public static class CConfig extends PluginMeta {
```
sigram, line 369: This example may be confusing because in general case configuration classes don't have to subclass `PluginMeta`. I propose removing this subclassing here to make it clear that's the case.

sigram (changes requested):

solr/core/src/java/org/apache/solr/api/ContainerPluginsRegistry.java
noblepaul: It's a standard property , that is a part of the plugin definition

solr/core/src/test/org/apache/solr/handler/TestContainerPlugin.java
```diff
@@ -366,7 +381,7 @@ public void m2(SolrQueryRequest req, SolrQueryResponse rsp) {
 
   }
 
-  public static class CConfig extends PluginMeta {
+  public static class CConfig implements ReflectMapWriter {
```
sigram, line 384: I don't understand this change. Do all config beans have to implement `ReflectMapWriter`?

If that's the case, and all config beans have to implement ReflectMapWriter then that's what the generic type in ConfigurablePlugin<T> should use. If not then this test is misleading again because it suggests you have to implement additional interfaces to have a usable config bean.

sigram (changes requested):

solr/core/src/java/org/apache/solr/api/ContainerPluginsRegistry.java
sigram: It made sense when all plugins were request handlers, which is no longer true. In case of plugins that are not handlers a "standard" `pathPrefix` property doesn't make sense.

solr/core/src/java/org/apache/solr/api/AnnotatedApi.java
```diff
@@ -62,6 +63,8 @@
 
 public class AnnotatedApi extends Api implements PermissionNameProvider , Closeable {
   private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+  private static final ObjectMapper mapper = SolrJacksonAnnotationInspector.createObjectMapper()
```
sigram, line 66: Oh, I didn't notice before that it was created for each request. That was horrible!

Still, we already have one instance here and another (configured the same) in `ContainerPluginsRegistry`. We should probably expose this instance here and reuse it in `ContainerPluginsRegistry`.

solr/core/src/java/org/apache/solr/api/ConfigurablePlugin.java
sigram, line 24: Should this be `<T extends ReflectMapWriter>` ?

solr/core/src/java/org/apache/solr/api/AnnotatedApi.java
noblepaul: It was not created per request. it was created per API

solr/core/src/java/org/apache/solr/api/ConfigurablePlugin.java
noblepaul: may be `<T extends MapWriter>`

solr/core/src/test/org/apache/solr/handler/TestContainerPlugin.java
noblepaul: All beans must implement `MapWriter`

solr/core/src/java/org/apache/solr/api/ContainerPluginsRegistry.java
noblepaul: `PluginMeta` is the information used by the framework. It's optional, so your plugin can choose to ignore it

solr/core/src/test/org/apache/solr/handler/TestContainerPlugin.java
sigram: Then we have to explicitly say so in `ConfigurablePlugin<T extends MapWriter>`.

solr/core/src/java/org/apache/solr/api/ContainerPluginsRegistry.java
sigram: I know that I can ignore it - my point was that this property is a relic of the time when we allowed only Api handlers as plugins. Now when non-Api plugins are first-class citizens half of the time this property doesn't make sense because it's specific only to Api plugins - so it should not be exposed as a standard property for all plugins.

noblepaul: It can be a specific sub-class of `PluginMeta` say `EndPointPluginMeta` .

However, that change does not belong here

sigram: Let's make it a sub-class.

(I would argue that it belongs in this PR because it's a part of the configuration mechanism for the plugins that this PR defines. Since we are adding a flexible config bean it doesn't make sense to still keep the old primitive field. And you can always ADD fields in subclasses but removing them is much harder... ;) )

PR #2066, opened 2020-11-06
SOLR-14975: Optimize CoreContainer.getAllCoreNames and getLoadedCoreNames.

bruno-roustant:


dsmiley (changes requested):
I observe SolrCores.getCoreDescriptors could be optimized as well but I suppose it's fine for now.

solr/core/src/java/org/apache/solr/core/CoreContainer.java
```diff
@@ -1568,33 +1568,49 @@ private void resetIndexDirectory(CoreDescriptor dcore, ConfigSet coreConfig) {
   }
 
   /**
-   * @return a Collection of registered SolrCores
+   * Gets the permanent (non-transient) cores that are currently loaded.
+   *
+   * @return An unsorted collection.
    */
   public Collection<SolrCore> getCores() {
```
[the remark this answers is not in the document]
bruno-roustant: Very good point!

solr/core/src/java/org/apache/solr/core/SolrCores.java
```diff
@@ -198,35 +188,91 @@ protected SolrCore putCore(CoreDescriptor cd, SolrCore core) {
...
+      Set<String> transientCoreNames = transientCoreCache == null ? Collections.emptySet() : transientCoreCache.getAllCoreNames();
+      return distinctSetsUnion(residentDescriptors.keySet(), transientCoreNames);
+    }
+  }
+
+  /**
+   * Makes the union of two distinct sets.
+   */
+  private static <T> Collection<T> distinctSetsUnion(Set<T> set1, Set<T> set2) {
+    assert areSetsDistinct(set1, set2);
```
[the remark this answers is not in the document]
bruno-roustant: Yes it must be a snapshot (like in getCores()). I'll add the comment.

solr/core/src/java/org/apache/solr/core/SolrCores.java
```diff
@@ -198,35 +188,91 @@ protected SolrCore putCore(CoreDescriptor cd, SolrCore core) {
...
-  Set<String> getLoadedCoreNames() {
-    Set<String> set;
-
+  Collection<String> getLoadedCoreNames() {
     synchronized (modifyLock) {
-      set = new TreeSet<>(cores.keySet());
-      if (getTransientCacheHandler() != null) {
-        set.addAll(getTransientCacheHandler().getLoadedCoreNames());
-      }
+      TransientSolrCoreCache transientCoreCache = getTransientCacheHandler();
```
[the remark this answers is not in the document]
bruno-roustant: There is a comment in TransientSolrCoreCacheFactory.newInstance() line 60 that says a bad config should not prevent the CoreContainer from loading. It logs an continues.
Then SolrCores.getTransientCacheHandler() checks transientCoreCache null. If so it logs an error and returns null.

Before this first PR I tried to replace the null by a NOOP TransientSolrCoreCache singleton. But I realize now that I could use an instance with the parent container. Let's do it, it'll save a lot of null checks.

solr/core/src/java/org/apache/solr/cloud/ReplicateFromLeader.java
```diff
@@ -69,7 +71,10 @@ public void startReplication(boolean switchTransactionLog) {
         if (cc.isShutDown()) {
           return;
         } else {
-          throw new SolrException(SolrException.ErrorCode.SERVER_ERROR, "SolrCore not found:" + coreName + " in " + cc.getLoadedCoreNames());
+          Collection<String> loadedCoreNames = cc.getLoadedCoreNames();
+          throw new SolrException(SolrException.ErrorCode.SERVER_ERROR, "SolrCore not found:" + coreName + " in "
```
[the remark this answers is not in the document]
bruno-roustant: Ok. I'll share in a new method CloudUtil.getLoadedCoreNamesAsString().

bruno-roustant:
I'll continue soon with two other improvements:
- Optimize SolrCores.getCoreDescriptors().
- NOOP TransientSolrCoreCache.

solr/core/src/java/org/apache/solr/core/SolrCores.java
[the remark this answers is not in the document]
dsmiley: @ErickErickson IMO a bad config should stop Solr in its tracks.  Why allow this?  I'd rather disallow it and thus we'd *always* have either the default impl or some custom impl, but _something_.

bruno-roustant:
I'm looking at optimizing SolrCores.getCoreDescriptors() but actually it would require to add a new TransientSolrCoreCache.getTransientDescriptors() method. This means the public interface TransientSolrCoreCache would not be compatible. I suppose we shouldn't do that in this PR but rather in another for 9.x only? @dsmiley

ErickErickson:
David:

I’ve no objection to removing all the if (transientHandler==null) tests if we fail on initialization.

Bruno:

Yeah, I think 9.0 only is the way to go.

bruno-roustant:
Finally I optimized SolrCore.getCoreDescriptors(). The version in this commit is the one for the 8.x branch - TransientSolrCoreCache.getTransientDescriptors() stays compatible with a default (sub-optimal but still better) implementation overridden by the optimal implementation in TransientSolrCoreCacheDefault.
When I merge this PR, I'll simply remove the default impl on master.

bruno-roustant:
So @dsmiley @ErickErickson do we all agree that we throw an exception in in TransientSolrCoreCacheFactory.newInstance() line 60 and we never return null?

solr/core/src/java/org/apache/solr/core/SolrCores.java
```diff
@@ -490,21 +492,19 @@ public CoreDescriptor getCoreDescriptor(String coreName) {
   }
 
   /**
-   * Get the CoreDescriptors for every SolrCore managed here
-   * @return a List of CoreDescriptors
+   * Get the CoreDescriptors for every {@link SolrCore} managed here (permanent and transient, loaded and unloaded).
+   *
+   * @return An unordered list copy. This list can be modified by the caller (e.g. sorted).
```
dsmiley, line 497: Collectors.toList says that there is no guarantees on the mutability.  So remove the docs about the mutability guarantee of what we return because we can't guarantee that.
Or, simply use ArrayList and call addAll.  I prefer this; use of Streams here doesn't appear to add value.

dsmiley:
> So @dsmiley @ErickErickson do we all agree that we throw an exception in in TransientSolrCoreCacheFactory.newInstance() line 60 and we never return null?

+1 definitely

solr/core/src/java/org/apache/solr/core/SolrCores.java
bruno-roustant: I'll use ArrayList.

bruno-roustant:
Ok, now I think I integrated all comments.

ErickErickson:
+1

solr/core/src/java/org/apache/solr/core/SolrCores.java
```diff
@@ -121,8 +101,8 @@ protected void close() {
         // make a copy of the cores then clear the map so the core isn't handed out to a request again
         coreList.addAll(cores.values());
         cores.clear();
-        if (transientSolrCoreCache != null) {
-          coreList.addAll(transientSolrCoreCache.prepareForShutdown());
+        if (transientSolrCoreCacheFactory != null) {
```
dsmiley, line 104: Do we still need a null check here?  And why the factory vs the cache itself?

dsmiley (approved):

bruno-roustant:
@dsmiley I pushed a new commit to have a NO OP TransientSolrCacheFactory before SolrCores.load() is called.
This is to have your opinion, whether to keep it or remove it.
If SolrCores is used before calling SolrCores.load() it results in a SolrException in getTransientCacheHandler(). This may happen for current users that previously didn't have to call load() at the very beginning after SolrCores creation. Do you think it could be a behavior backward incompatibility?

solr/core/src/java/org/apache/solr/core/SolrCores.java
```diff
@@ -62,55 +51,44 @@
...
     Collection<SolrCore> coreList = new ArrayList<>();
 
-    
-    TransientSolrCoreCache transientSolrCoreCache = getTransientCacheHandler();
-    // Release observer
-    if (transientSolrCoreCache != null) {
-      transientSolrCoreCache.close();
-    }
+    // Release transient core cache.
+    getTransientCacheHandler().close();
```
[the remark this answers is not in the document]
dsmiley: @bruno-roustant the muse bot makes a good point; there should be a synchronized(modifyLock) around grabbing getTransientCacheHandler and calling close on it.

solr/core/src/java/org/apache/solr/core/SolrCores.java
```diff
@@ -51,7 +51,7 @@
   // to essentially queue them up to be handled via pendingCoreOps.
   private static final List<SolrCore> pendingCloses = new ArrayList<>();
 
-  private TransientSolrCoreCacheFactory transientSolrCoreCacheFactory;
+  private TransientSolrCoreCacheFactory transientSolrCoreCacheFactory = TransientSolrCoreCacheFactory.NO_OP;
```
dsmiley, line 54: Under what circumstance do we need this no-op impl to prevent an NPE?

bruno-roustant: If a SolrCores method accessing getTransientCacheHandler() is called before calling load().
For example an asynchronous periodic thread that would be started with CoreContainer and that would call getLoadedCoreNames(). This worked before even if SolrCores.load() was called after. (first calls would return without counting transient cores, and subsequent calls after load() would count transient cores)

solr/core/src/java/org/apache/solr/core/SolrCores.java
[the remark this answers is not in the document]
bruno-roustant: +1 thanks muse!

solr/core/src/java/org/apache/solr/core/SolrCores.java
dsmiley: then SolrCores.load could be called much sooner, basically right after the resourceLoader is ready.  

Also, maybe this other thread ought to wait to start till some later time.  Perhaps ideally there would be an event publishing mechanism, which doesn't exist currently, I know.  Or alternatively just have some CountDownLatch signal, like signaling when Solr will begin loading cores.

solr/core/src/java/org/apache/solr/core/SolrCores.java
```diff
@@ -536,7 +538,9 @@ public void queueCoreToClose(SolrCore coreToClose) {
    * @return the cache holding the transient cores; never null.
    */
   public TransientSolrCoreCache getTransientCacheHandler() {
```
dsmiley, line 540: I think it should be a prerequisite that the caller acquire the lock a-priori.  For example, close() seems to need to keep holding this lock to call close() on it.

bruno-roustant: I don't think so. Let's say transientSolrCoreCacheFactory = A. If there is a race and load() is called between getTransientCacheHandler() and TransientSolrCoreCache.close(). getTransientCacheHandler() returns A, then load() sets B, and then A.close() is called. It is the same result as the sequence { getTransientCacheHandler().close() } atomically on A, and then load() B.

But actually it doesn't matter. So I can add a synchronized (modifyLock) around if you prefer.
I'll keep anyway the synchronized inside getTransientCacheHandler() because it is public.

dsmiley: Okay; I'm fine with the lock to be safe.

dsmiley (approved):
Great!

PR #2067, opened 2020-11-06
SOLR-14987: Reuse HttpSolrClient per node vs. one per Solr core when using CloudSolrStream

thelabdude:
# Description

For collections with many shards (or aliases with many collections and some shards), `CloudSolrStream` will end up creating a new `HttpSolrClient` for every `SolrStream` it opens because the cache key is the full core URL, such as: `http://127.0.0.1:63460/solr/collection4_shard4_replica_n6/`

In addition, `CloudSolrStream#getSlices` was calling `clusterState.getCollectionsMap()` which pre-emptively loads all `LazyCollectionRef` from ZK unnecessarily. This could cause an issue with clusters with many collections and slow down the streaming expression execution.

# Solution

In this PR, I've introduced a new ctor in `SolrStream` to pass the Replica's `baseUrl` and `core` as separate parameters. This leads to reusing the same `HttpSolrClient` for the same node because the cache key is now `http://127.0.0.1:63460/solr/`. I chose this new ctor approach because `CloudSolrStream` is not the only consumer of `SolrStream` and it knows how the list of URLs where constructed from cluster state, so it can safely make the decision about passing the core and reusing clients.

When the request is sent to the remote core, we need to add the core name to the path. This happens in `SolrStream#constructParser`. This method was public and takes a SolrClient (even though SolrStream already has an HttpSolrClient created in the `open` method); I've changed the signature to be private and use the client opened in the `open` method.

# Tests

Added a new test `testCloudStreamClientCache` in `StreamingTest` to verify the SolrStreams created by the CloudStream have the correct baseUrl (without the core).

# Checklist


thelabdude:
I don't think the current solution solves the problem of clusters many nodes, you'll still get a bunch of client objects, still investigating what can be done about that

solr/solrj/src/java/org/apache/solr/client/solrj/io/stream/CloudSolrStream.java
```diff
@@ -334,11 +334,6 @@ private StreamComparator parseComp(String sort, String fl) throws IOException {
   public static Slice[] getSlices(String collectionName, ZkStateReader zkStateReader, boolean checkAlias) throws IOException {
     ClusterState clusterState = zkStateReader.getClusterState();
 
-    Map<String, DocCollection> collectionsMap = clusterState.getCollectionsMap();
```
madrob, line 337: related: can we update the javadoc on clusterState.getCollectionsMap to be more explicit that it _will_ make a call to zk, instead of the current _may_

solr/solrj/src/java/org/apache/solr/client/solrj/io/stream/CloudSolrStream.java
```diff
@@ -334,11 +334,6 @@ private StreamComparator parseComp(String sort, String fl) throws IOException {
...
-
-    //TODO we should probably split collection by comma to query more than one
-    //  which is something already supported in other parts of Solr
-
     // check for alias or collection
```
madrob, line 337: Should we cache the value of `zkStateReader.getAliases` below to avoid volatile reads?

solr/solrj/src/java/org/apache/solr/client/solrj/io/stream/SolrStream.java
```diff
@@ -126,6 +135,17 @@ public void open() throws IOException {
     }
   }
 
+  private String getNodeUrl() {
```
madrob, line 138: Can we precomute this in the constructor?

solr/solrj/src/java/org/apache/solr/client/solrj/io/stream/CloudSolrStream.java
thelabdude: good idea

thelabdude:
Hi @joel-bernstein wasn't able to assign you as a reviewer on this, but would love for you to take a look when convenient.

solr/solrj/src/java/org/apache/solr/client/solrj/io/stream/CloudSolrStream.java
```diff
@@ -334,88 +330,76 @@ private StreamComparator parseComp(String sort, String fl) throws IOException {
...
+        .map(c -> clusterState.getCollectionOrNull(c, true))
         .filter(Objects::nonNull)
         .flatMap(docCol -> Arrays.stream(docCol.getActiveSlicesArr()))
         .collect(Collectors.toList());
     if (!slices.isEmpty()) {
-      return slices.toArray(new Slice[slices.size()]);
-    }
-
-    // Check collection case insensitive
-    for(Entry<String, DocCollection> entry : collectionsMap.entrySet()) {
```
thelabdude, line 364: I removed this b/c I don't think we should try to accommodate improperly cased collection names. No tests broke, but let me know if we need this for some reason I don't understand

solr/solrj/src/java/org/apache/solr/client/solrj/io/stream/CloudSolrStream.java
thelabdude: Moved the call to getAliases out of the for loop

solr/solrj/src/java/org/apache/solr/client/solrj/io/stream/CloudSolrStream.java
```diff
@@ -334,88 +330,76 @@ private StreamComparator parseComp(String sort, String fl) throws IOException {
...
     try {
+      final Stream<SolrStream> streamOfSolrStream;
+      if (streamContext != null && streamContext.get("shards") != null) {
+        // stream of shard url with core
+        streamOfSolrStream = getShards(this.zkHost, this.collection, this.streamContext, mParams).stream()
+            .map(s -> new SolrStream(s, mParams));
+      } else {
+        // stream of replicas to reuse the same SolrHttpClient per baseUrl
+        // avoids re-parsing data we already have in the replicas
+        streamOfSolrStream = getReplicas(this.zkHost, this.collection, this.streamContext, mParams).stream()
```
thelabdude, line 371: Here we're keeping the Replica so we have direct access to its baseUrl and core name instead of parsing those out of the shardUrl

solr/solrj/src/java/org/apache/solr/client/solrj/io/stream/SolrStream.java
```diff
@@ -268,8 +275,7 @@ private Map mapFields(Map fields, Map<String,String> mappings) {
     return fields;
   }
 
-  // temporary...
-  public TupleStreamParser constructParser(SolrClient server, SolrParams requestParams) throws IOException, SolrServerException {
+  private TupleStreamParser constructParser(SolrParams requestParams) throws IOException, SolrServerException {
```
thelabdude, line 278: Didn't seem like this method needed to be public and we already get a SolrClient in the open method, so no need to pass it. However, this breaks a public method signature, so is only for Solr 9.x and shouldn't be back-ported to 8.x

thelabdude:
Plan to merge this into master later today.

PR #2068, opened 2020-11-07
LUCENE-8982: Separate out native code to another module to allow cpp build with gradle

zacharymorn:
# Description
When working on https://issues.apache.org/jira/browse/LUCENE-8982, we realized that there may no longer be a way to build native cpp code due to migration away from ant. This patch separate out native code to another module to allow cpp build with gradle.

# Solution
Create a new lucene module and new parameter `-Pbuild.native=true` for cpp code compilation.

# Tests
Pending

# Checklist


lucene/native/build.gradle
```diff
@@ -0,0 +1,42 @@
...
+ * For details, please see https://github.com/gradle/gradle-native/issues/352#issuecomment-461724948
+ */
+apply plugin: 'cpp-library'
+
+description = 'Module for native code'
+
+library {
+    baseName = 'NativePosixUtil'
+
+    privateHeaders.from file(System.getProperty('java.home') + '/include')
```
dweiss, line 29: this should be resolved relative to the JVM we compile against, not gradle's JVM. See alternative-jdk-support.gradle and use root project's property:
```
// Set up root project's property.
rootProject.ext.runtimeJava = altJvm 
```

lucene/native/build.gradle
```diff
@@ -0,0 +1,42 @@
...
+    privateHeaders.from file(System.getProperty('java.home') + '/include/darwin')
```
dweiss, line 30: Maybe we should check for existence of these folders and add them only if they exist? Seems strange to add includes for all of them.

dweiss (commented):
I added some general comments. The code, as it is now, doesn't compile for me on Windows (fails to find includes as they're under win32). Even if I add the includes, it fails on compiling the source code. I think the whole native compilation thing should be ignored on Windows as it's not going to work now.

lucene/native/build.gradle
```diff
@@ -0,0 +1,42 @@
...
+    privateHeaders.from file(System.getProperty('java.home') + '/../include/solaris')
```
dweiss, line 31: Are these really a folder up from java.home? On my linux (java 11-15) they seem to be right under JDK's home (although I install Java manually, not via distribution-specific package manager).

zacharymorn: These paths were ported based on the original ant target definition below :

https://github.com/apache/lucene-solr/blob/005fec5a58e1993a1ce7ae235060ad3bf2adb9c7/lucene/contrib/misc/build.xml#L44-L45

My guess is these paths might now be deprecated in newer JDK installations. 

Do you think this native build feature should be back ported to prior versions? If so, I think it might be safer to have those paths as well.

lucene/native/build.gradle
zacharymorn: In a newer commit I check for platform first before including them. Will that work as well?

lucene/native/build.gradle
zacharymorn: Done.

zacharymorn:
> I added some general comments. The code, as it is now, doesn't compile for me on Windows (fails to find includes as they're under win32). Even if I add the includes, it fails on compiling the source code. I think the whole native compilation thing should be ignored on Windows as it's not going to work now.

Thanks for the comments! I think the reason of failure even with /win32 inclusion is that the native module currently only includes `NativePoxisUtil.cpp`, which seems to be for unix only. The windows equivalent`WindowsDirectory.cpp` still sits in `misc` module and hasn't been moved over yet.

From where we are now with the native module, in order to support Windows, we probably just need to add some logic in `build.gradle` to pick the right source and include folders, as well as passing in custom compiler flags. But yes I also feel that this can be tackled in a later stage after the changes for unix are in good shape.

lucene/native/build.gradle
dweiss: These paths are wrong, I think. I don't know where they came from but they should correspond to the layout of newer JDK distributions (openjdk). 

I'm not sure Solaris is still needed - I can't test it on Solaris.

dweiss:
bq. The windows equivalentWindowsDirectory.cpp still sits in misc module and hasn't been moved over yet.

Ah... I didn't realize this is the case then - sorry. Give me a day or two, I'll try to make those files compile under Windows and maybe we can do a clean patch that moves everything.

dweiss:
Zach I've cleaned up the native build a bit - moved it under lucene/misc, added Windows build (it does build the native library for me). I didn't check whether it works on a Mac but I suspect it should.

I also left the native project included by default in settings (removed the "optional" flag). Gradle's cpp plugin ignores the project on platforms not explicitly mentioned in the targetMachines - I am curious whether we'll blow up something or if it's just going to work. 

While I don't particularly like having native code in Lucene, I think it's better than it used to be (mixed cpp code with java code, etc.).

I allowed myself to commit directly to your fork, hope you don't mind (please test it out!).

lucene/misc/native/build.gradle
```diff
@@ -0,0 +1,69 @@
...
+ * For details, please see https://github.com/gradle/gradle-native/issues/352#issuecomment-461724948
+ */
+import org.apache.tools.ant.taskdefs.condition.Os
+
+description = 'Module for native code'
+
+apply plugin: 'cpp-library'
+
+library {
+  baseName = 'NativePosixUtil'
```
dweiss, line 29: I wonder if we should rename the resulting native library something more specific...

dweiss:
Byt. if we do have to add that explicit 'build.native' option then I'd implement it as a task (graph) exclusion rather than project exclusion. Windows users in particular may complain as the plugin requires visual studio...

https://docs.gradle.org/current/userguide/building_cpp_projects.html#sec:cpp_supported_tool_chain

So something like this on all native project's tasks (conditionally):
https://discuss.gradle.org/t/removing-tasks-from-taskgraph-remove-a-task-dependency/394
or filter them out entirely:

lucene/misc/native/build.gradle
zacharymorn: Given this now also includes the Windows one, and the cpp code focus on doing file IO, I'm guessing something like `NativeIOUtil` might work ?

zacharymorn:
> Zach I've cleaned up the native build a bit - moved it under lucene/misc, added Windows build (it does build the native library for me). I didn't check whether it works on a Mac but I suspect it should.
> 
> I also left the native project included by default in settings (removed the "optional" flag). Gradle's cpp plugin ignores the project on platforms not explicitly mentioned in the targetMachines - I am curious whether we'll blow up something or if it's just going to work.
> 
> While I don't particularly like having native code in Lucene, I think it's better than it used to be (mixed cpp code with java code, etc.).
> 
> I allowed myself to commit directly to your fork, hope you don't mind (please test it out!).

Thanks Dawid for the changes! I tested it out in my mac and it built fine as well. 

I originally separated out into an independent native module thinking that it could host future native code as well, but I guess that's probably just pre-mature optimization as it hasn't been the case for the last few years.

> I am curious whether we’ll blow up something or if it’s just going to work.

Just curious, are there any cross-platform tests as well in the pipeline that can confirm this? How do we verify this other than running local builds?

lucene/misc/native/build.gradle
```diff
@@ -0,0 +1,69 @@
...
+      file("${javaHome}/include/win32"),
+      file("${javaHome}/include/darwin"),
+      file("${javaHome}/include/linux"),
+      file("${javaHome}/include/solaris")]) {
+    if (path.exists()) {
+      systemIncludes.from path
+    }
+  }
+
+  compilerArgs.add '-fPIC'
```
zacharymorn, line 64: Just curious, shall we also modify the compiler args when it’s on Windows, to be the same with what's used before?

https://github.com/apache/lucene-solr/blob/ec9a659845973a0dd0ee7c04e0075db818ed118d/lucene/misc/src/java/org/apache/lucene/store/WindowsDirectory.java#L31-L35 

A quick search shows that some of these flags might be specific to MinGW compiler though, so I'm not sure if these flags are still relevant.

lucene/misc/native/build.gradle
dweiss: Sure. Maybe LuceneNativeIO though so that it's clear what dll it is?

lucene/misc/native/build.gradle
dweiss: I wouldn't add anything if it works (and it currently does for me). There are compiler toolchain-specific ways of adding options but like I said - if it's not required, I wouldn't worry about it.

lucene/misc/src/java/org/apache/lucene/store/NativeUnixDirectory.java
```diff
@@ -47,10 +47,10 @@
  *
  * <p>To use this you must compile
  * NativePosixUtil.cpp (exposes Linux-specific APIs through
- * JNI) for your platform, by running <code>ant
- * build-native-unix</code>, and then putting the resulting
- * <code>libNativePosixUtil.so</code> (from
- * <code>lucene/build/native</code>) onto your dynamic
+ * JNI) for your platform, by running <code>
+ * ./gradlew build -Pbuild.native=true</code>, and then putting the resulting
```
dweiss, line 51: I removed the flag and forgot to update the docs here.

lucene/packaging/build.gradle
```diff
@@ -32,7 +32,9 @@ def includeInBinaries = project(":lucene").subprojects.findAll {subproject ->
         ":lucene:packaging",
         ":lucene:documentation",
         // Exclude parent container project of analysis modules (no artifacts).
-        ":lucene:analysis"
+        ":lucene:analysis",
+        // Exclude native module, which requires manual copying and enabling
+        ":lucene:native"
```
dweiss, line 37: This reference needs an updated path - it'll probably fail when building the whole project.

lucene/misc/native/build.gradle
zacharymorn: Ok cool. I've updated the comment here to reflect the new way of building the code.

lucene/misc/src/java/org/apache/lucene/store/NativeUnixDirectory.java
zacharymorn: Updated.

lucene/packaging/build.gradle
zacharymorn: It did fail, I've updated the reference here and confirmed it's able to build from directory root.

lucene/misc/native/build.gradle
zacharymorn: Done.

.gitignore
```diff
@@ -8,6 +8,10 @@ build/
 /.idea/
 #    IntelliJ creates this folder, ignore.
 /dev-tools/missing-doclet/out/
+*.iml
```
dweiss, line 11: Hmm... how come you have intellij's old-style files and not the folder-based layout? :)

dweiss (commented):
All of it looks good, Zach. Sorry about my lack of consistency here but on as second thought I think we should add a safety switch of perhaps force-disabling the native build - just in case we commit it in and have to do it for some reason. 

I'll work on this later today and commit it in. 

I also feel tempted to add tests that utilize this native code. This can come as a separate issue though.

dweiss:
Added the kill switch, Zach. I think you need to merge with master and then update the overview file that Tomoko moved to package-info.java (grep for "build-native-unix" and you'll see old instructions currently in overview.html).

.gitignore
zacharymorn: Hmm I seems to recall these files were generated after running `./gradlew idea`, but I removed them in the latest commit anyway.

lucene/misc/native/src/main/posix/NativePosixUtil.cpp
```diff
@@ -38,12 +38,12 @@
...
 #ifdef LINUX
 /*
- * Class:     org_apache_lucene_store_NativePosixUtil
+ * Class:     org_apache_lucene_misc_store_NativePosixUtil
  * Method:    posix_fadvise
  * Signature: (Ljava/io/FileDescriptor;JJI)V
  */
 extern "C"
-JNIEXPORT jint JNICALL Java_org_apache_lucene_store_NativePosixUtil_posix_1fadvise(JNIEnv *env, jclass _ignore, jobject fileDescriptor, jlong offset, jlong len, jint advice)
+JNIEXPORT jint JNICALL Java_org_apache_lucene_misc_store_NativePosixUtil_posix_1fadvise(JNIEnv *env, jclass _ignore, jobject fileDescriptor, jlong offset, jlong len, jint advice)
```
zacharymorn, line 46: These paths need to be adjusted as the native code has been moved in d1110394e9c

zacharymorn:
> Added the kill switch, Zach. I think you need to merge with master and then update the overview file that Tomoko moved to package-info.java (grep for "build-native-unix" and you'll see old instructions currently in overview.html).

Updated.

zacharymorn:
> All of it looks good, Zach. Sorry about my lack of consistency here but on as second thought I think we should add a safety switch of perhaps force-disabling the native build - just in case we commit it in and have to do it for some reason.
> 
> I'll work on this later today and commit it in.
> 
> I also feel tempted to add tests that utilize this native code. This can come as a separate issue though.

I just pushed up a commit that adds tests to exercise the native library loading part. Seems to be working fine on my Mac. Could you please give it a try to see it works on Windows as well? 

On the other hand, I think these tests will break if run from IDEs. Do we need to support that in this PR?

.gitignore
dweiss: Right... gradlew idea is a plugin - it does generate those files. IntelliJ has native gradle support nowadays. See help/IDEs.txt; perhaps it should be clarified there.

lucene/misc/native/src/main/posix/NativePosixUtil.cpp
dweiss: Yeah. We really should try to add a test that tries to run with these libs. I don't know how to handle this yet - can be a follow-up issue.

dweiss:
> On the other hand, I think these tests will break if run from IDEs. Do we need to support that in this PR?

Oh, thanks! I'll take a look after I come back from work. I think the PR should be clean in that it doesn't break other people's workflow, so yes - if you added tests they should run or be quietly ignored if they're not supported. I'll take a look.

zacharymorn:
> > On the other hand, I think these tests will break if run from IDEs. Do we need to support that in this PR?
> 
> Oh, thanks! I'll take a look after I come back from work. I think the PR should be clean in that it doesn't break other people's workflow, so yes - if you added tests they should run or be quietly ignored if they're not supported. I'll take a look.

Agreed. I just looked around as well, and it seems @Nightly annotation might be a possible work around here to disable running these tests from IDE. Granted it's a bit of a hack as it also disabled running those tests from developer's local gradle build by default, but at least these tests will get run in nightly build and bugs can be caught (albeit a bit late).

Another solution is perhaps to create a new annotation to disable tests running from IDE, if such a thing can be detected ?

dweiss:
Thanks Zach. I looked at what you did with tests and I think this can be done in a cleaner way that always works. I'll show you how, give me some time.

Also, update your IDE"s formatter settings to the convention used throughout the code (you used 4 spaces indentation). May be helpful in other patches too.

I'll get back to you.

zacharymorn:
> Thanks Zach. I looked at what you did with tests and I think this can be done in a cleaner way that always works. I'll show you how, give me some time.

Sounds good! Look forward to it.

> Also, update your IDE"s formatter settings to the convention used throughout the code (you used 4 spaces indentation). May be helpful in other patches too.

Ops sorry. Just updated it to use 2 spaces indentation.

dweiss:
No worries. I've pushed the code the way I think it should work - please take a look, let me know if you don't understand something. I tested on Windows and Linux, runs fine. The 'tests.native' flag is set automatically depending on 'build.native' but is there separately just in case somebody wished to manually enable those tests from IDE level.

zacharymorn:
> No worries. I've pushed the code the way I think it should work - please take a look, let me know if you don't understand something. I tested on Windows and Linux, runs fine. The 'tests.native' flag is set automatically depending on 'build.native' but is there separately just in case somebody wished to manually enable those tests from IDE level.

Wow these look pretty advanced! Pretty sure I can't come up with them myself (and I do have a question that I can't seems to find the answer to readily online). I also like that these tests can be run from dev's local environment as well (and they run fine on my mac) compared to the @Nightly annotation approach. Thanks Dawid!

gradle/native/disable-native.gradle
```diff
@@ -17,20 +17,65 @@
...
+    ":lucene:misc:native"
+]}
+
+def javaProjectsWithNativeDeps = allprojects.findAll {it.path in [
+    ":lucene:misc"
+]}
+
+// Set up defaults for projects with native dependencies.
+configure(javaProjectsWithNativeDeps, {
+  configurations {
```
zacharymorn, line 36: This configuration block seems very auto-magical to me in that it somehow gets linked to `:lucene:misc:native/build` folder (no reference to `nativeProjects` here),  and the `copyNativeDeps` task below copies only the needed library artifact without all the nested folder structure (nor does it seems to copy any other random file I created in `:lucene:misc:native/build` folder to test thing out). Is this some convention trigged by the two attribute settings below?

dweiss: Right. This is gradle deep-waters but not too complicated once you get the hang of it. I am personally font of simple and straightforward code, coming from assembly, but these days it's hard to avoid higher-level constructs. Sorry about this!

So, the code you asked about essentially applies ('configure') a code block (secont argument to configure - the closure) to all objects that are in the collection that is the first argument to configure. When you look at that patch we create two such collections - nativeProjects and javaProjectsWithNativeDeps, each one with Project instances that match a corresponding path. This is very much as if you copy-pasted the same code in each gradle.build of each corresponding project - avoids repetition, centralises common configuration.

The inside of this closure block is where things get more interesting and more gradle-sque. The "configurations" block is a project property that declares the project's configurations to which artifacts (or dependencies) can be added. You should understand how this works because this is the core of how it works. See here, for example (gradle documentation is very good, if sometimes scattered across multiple documents):

https://docs.gradle.org/current/userguide/dependency_management.html#declaring-dependencies

So, for any project in javaProjectsWithNativeDeps we declare a configuration called "nativeDeps". To this configuration we will attach a dependency to a "native" project - our cpp-library modules. In case of :lucene:misc we attach it inside build.gradle:
```
nativeDeps project(":lucene:misc:native")
```

Then the code declares two project "extra (ext) properties - testOptions and nativeDepsDir. The testOptions is a an array we use to collect various test attributes from bits and pieces of configuration scattered around other files. To make the evaluation work in the right order (ensure the array is there) we put everything in a deferred-execution block (this is why plugins.withType(JavaPlugin) stuff is needed...).

Finally, we apply a code closure configuring any of the project's tasks that have the type "Test", adding java.library.path system property pointing at the nativeDepsDir property location and a dependency on a task that will copy all dependencies (artifacts produced by dependencies attached to the nativeDeps configuration) to that folder. These two steps are only needed if we indeed plan to run with native libraries so they're wrapped in an 'if'.

> and the copyNativeDeps task below copies only the needed library artifact without all the nested folder structure

It actually synchronizes the state of the target folder with source files. Note that the "source" files are pointed to by the configuration (artifacts produced by this configuration dependencies). The cpp-library produces *multiple* configuration variants (debug, release - with possibly multiple artifacts). This is where it gets tricky and this is why we declared the nativeDeps configuration with some "attributes" that allow gradle to pick just one of the produced configurations out of multiple choices. In most Java projects you'd never have to use this but you combined cpp and java. Try removing those attributes from configuration definition and see what happens. Then see this:

https://docs.gradle.org/current/userguide/variant_model.html 

Sorry for throwing so much on you but you asked! :)

zacharymorn: Wow thanks Dawid so much for the detailed and thorough explanations and pointers! I did come across some of these documentations individually before, but it's just that the combination of configuration + variant + sync + cpp confuses me quite a bit. But it's pretty clear now. Thanks again for your patience and guidance here!

I do have a final question though. It seems at this point we should also be able to easily support packaging the compiled native code into misc jar package (and change to use the optimized variant of the build), to remove the final manual step to copy around the native library artifact for production usage of `WindowsDirectory` or `NativeUnixDirectory`? Do you think this is a good direction to be taken ?

dweiss: This is something I considered but eventually left out. Technically, it's trivial to do but my concern is that then the content of the "lucene distribution" depends on the machine you build it on - native libraries will vary and you can't (?) easily compile all libraries for the most common platforms to include all of them in the distribution... I don't know how to solve it. 

Unless those native extensions offer super performance boosts I'd rather treat them as "sandbox", expert-level stuff and not include them in the distribution by default? You can bring this issue up on the dev list, perhaps other people have an opinion here - I have never used those extensions myself.

zacharymorn: > Unless those native extensions offer super performance boosts I'd rather treat them as "sandbox", expert-level stuff and not include them in the distribution by default?

Makes sense. I think as part of https://issues.apache.org/jira/browse/LUCENE-8982 some performance measurement utility may come out, so by then this can be more objectively measured to help further discussions.

I guess this PR is ready for merge then? Please let me know if there's any other improvement you would like me to make. Thanks again for your help on this!

dweiss: Yep, I think it's ready! Please monitor jenkins so that we know should something go wrong - if it does we can just turn native builds off (by default).

zacharymorn: Sure will do. Just to confirm these are all the jobs right https://ci-builds.apache.org/job/Lucene/, and I guess job `Lucene-Artifacts-master` requires the most attention / monitoring?

dweiss: I think the build mailing list is the easiest to spot problems - builds@lucene.apache.org. There will be some builds there from branches (these can be ignored) but otherwise these are worth taking an occasional look at.

dweiss:
I've added changes entry and committed it in, thanks @zacharymorn !

PR #2069, opened 2020-11-09
LUCENE-9378: Make it possible to configure how to trade speed for compression on doc values.

jpountz:
This adds a switch to `Lucene80DocValuesFormat` which allows to
configure whether to prioritize retrieval speed over compression ratio
or the other way around. When prioritizing retrieval speed, binary doc
values are written using the exact same format as before more aggressive
compression got introduced.

### DIFF
--- a/lucene/backward-codecs/src/java/org/apache/lucene/backward_codecs/lucene87/Lucene87Codec.java
+++ b/lucene/backward-codecs/src/java/org/apache/lucene/backward_codecs/lucene87/Lucene87Codec.java
@@ -113,7 +113,7 @@ public class Lucene87Codec extends Codec {
    */
   public Lucene87Codec(Mode mode) {
     super("Lucene87");
-    this.storedFieldsFormat = new Lucene87StoredFieldsFormat(Objects.requireNonNull(mode.storedMode));
+    this.storedFieldsFormat = new Lucene87StoredFieldsFormat(Objects.requireNonNull(mode).storedMode);
     this.defaultFormat = new Lucene84PostingsFormat();
     this.defaultDVFormat = new Lucene80DocValuesFormat(mode.dvMode);
   }
--- a/lucene/backward-codecs/src/java/org/apache/lucene/backward_codecs/lucene87/Lucene87Codec.java
+++ b/lucene/backward-codecs/src/java/org/apache/lucene/backward_codecs/lucene87/Lucene87Codec.java
@@ -36,6 +36,7 @@ import org.apache.lucene.codecs.lucene50.Lucene50CompoundFormat;
 import org.apache.lucene.codecs.lucene50.Lucene50LiveDocsFormat;
 import org.apache.lucene.codecs.lucene50.Lucene50TermVectorsFormat;
 import org.apache.lucene.codecs.lucene60.Lucene60FieldInfosFormat;
+import org.apache.lucene.codecs.lucene80.Lucene80DocValuesFormat;
 import org.apache.lucene.codecs.lucene80.Lucene80NormsFormat;
 import org.apache.lucene.codecs.lucene84.Lucene84PostingsFormat;
 import org.apache.lucene.codecs.lucene86.Lucene86PointsFormat;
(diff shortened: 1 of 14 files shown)


lucene/backward-codecs/src/java/org/apache/lucene/backward_codecs/lucene87/Lucene87Codec.java
```diff
@@ -56,6 +57,23 @@
...
+  public static enum Mode {
+    /** Trade compression ratio for retrieval speed. */
+    BEST_SPEED(Lucene87StoredFieldsFormat.Mode.BEST_SPEED, Lucene80DocValuesFormat.Mode.BEST_SPEED),
+    /** Trade retrieval speed for compression ratio. */
+    BEST_COMPRESSION(Lucene87StoredFieldsFormat.Mode.BEST_COMPRESSION, Lucene80DocValuesFormat.Mode.BEST_COMPRESSION);
+
+    private final Lucene87StoredFieldsFormat.Mode storedMode;
+    private final Lucene80DocValuesFormat.Mode dvMode;
+
+    private Mode(Lucene87StoredFieldsFormat.Mode storedMode, Lucene80DocValuesFormat.Mode dvMode) {
```
mikemccand, line 71: Nice!  So we roll up the tradeoffs to Codec level which will then tell each format how to tradeoff.

lucene/core/src/test/org/apache/lucene/codecs/lucene80/BaseLucene80DocValuesFormatTestCase.java
```diff
@@ -286,7 +278,7 @@ private void doTestTermsEnumRandom(int numDocs, Supplier<String> valuesProducer)
     conf.setMergeScheduler(new SerialMergeScheduler());
     // set to duel against a codec which has ordinals:
     final PostingsFormat pf = TestUtil.getPostingsFormatWithOrds(random());
-    final DocValuesFormat dv = new Lucene80DocValuesFormat();
+    final DocValuesFormat dv = getCodec().docValuesFormat();
```
mikemccand, line 281: Will this randomize between the different `Mode` tradeoffs?

lucene/core/src/test/org/apache/lucene/codecs/lucene80/TestBestSpeedLucene80DocValuesFormat.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+package org.apache.lucene.codecs.lucene80;
+
+import org.apache.lucene.codecs.Codec;
+import org.apache.lucene.util.TestUtil;
+
+/**
+ * Tests Lucene80DocValuesFormat
+ */
+public class TestBestSpeedLucene80DocValuesFormat extends BaseLucene80DocValuesFormatTestCase {
```
mikemccand, line 25: Do we also have a dedicated `TestBestCompressedLucene80DocValuesFormat`?

mikemccand (commented):
Thank you @jpountz!  I left a few small questions, but this looks great.

At write time, Lucene user can choose whether they want smaller index (compressed) or faster search (`BEST_SPEED`), and the resulting index has Lucene's normal full back compat guarantee.

Let's make sure Lucene's `test-framework` is fully randomize these write-time `Codec` choices so we get good test coverage of all options.

lucene/core/src/test/org/apache/lucene/codecs/lucene80/TestBestSpeedLucene80DocValuesFormat.java
mikemccand: Oh nevermind I see you opened followon issue for this: https://issues.apache.org/jira/browse/LUCENE-9602

jpountz: You should see a `TestBestCompressedLucene80DocValuesFormat` file as well in this PR. I opened LUCENE-9602 specifically for backward compatibility and make sure we check in indices created by BEST_COMPRESSION in our source tree after every release to make sure we have good bw compatibility coverage.

lucene/core/src/test/org/apache/lucene/codecs/lucene80/BaseLucene80DocValuesFormatTestCase.java
jpountz: It's not randomizing, we are testing both modes explicitly via TestBestSpeedLucene80DocValuesFormat on one hand and TestBestCompressionLucene80DocValuesFormat on the other hand.

lucene/backward-codecs/src/java/org/apache/lucene/backward_codecs/lucene87/Lucene87Codec.java
jpountz: Right. It's still possible to made different choices for stored fields and doc values given that we allow configuration of doc values on a per-field basis, but this should at least keep simple use simple with one switch that configures stored fields and doc values at the same time.

mikemccand: Great!  Simple for common use cases ("I want best compression" or "I want fastest search"), and complex for complex use cases (I want separate control for each part of the index).

mikemccand (approved):
Thanks @jpountz!

mikemccand (approved):
Thanks @jpountz!

mikemccand (approved):

PR #2070, opened 2020-11-09
LUCENE-9536: Correct the OrdinalMap optimization.

jtibshirani:
Previously we only checked that the first segment's ordinal deltas were all
zero. This didn't account for some rare cases where some of the segment's
ordinals are filtered out, so the ordinals aren't contiguous. In these cases we
fill in dummy values for the missing ordinal deltas. So a segment's ordinals
can appear to match the global ordinals perfectly, but not actually contain all
the terms.

Such a case can arise when using a FilteredTermsEnum, for example when merging
a segment with deletions.


jtibshirani:
This should fix the failures we're seeing like `TestLucene70DocValuesFormat#testSparseSortedVariableLengthVsStoredFields` and `TestSimpleTextDocValuesFormat#testSparseSortedFixedLengthVsStoredFields`. Note to self: run the whole test suite a bunch of times when changing subtle logic!!

jpountz (approved):

PR #2071, opened 2020-11-10
LUCENE-9322: Some fixes to SimpleTextVectorFormat.

jtibshirani:
* Make sure the files are unique by renaming the term vectors extension to `tvc`.
* Fix a bug where reading a vector would drop the leading digit of the first element.

### DIFF
--- a/lucene/codecs/src/java/org/apache/lucene/codecs/simpletext/SimpleTextVectorReader.java
+++ b/lucene/codecs/src/java/org/apache/lucene/codecs/simpletext/SimpleTextVectorReader.java
@@ -245,8 +245,8 @@ public class SimpleTextVectorReader extends VectorReader {
 
     private void readVector(float[] value) throws IOException {
       SimpleTextUtil.readLine(in, scratch);
-      // skip leading " [" and strip trailing "]"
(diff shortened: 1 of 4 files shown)


jtibshirani:
I found these issues while fixing the following failing test:

```
./gradlew test --tests TestSortingCodecReader.testSortOnAddIndicesRandom -Dtests.seed=B38EBA45728D5FB1
```

lucene/codecs/src/java/org/apache/lucene/codecs/simpletext/SimpleTextVectorReader.java
```diff
@@ -245,8 +245,8 @@ private void readAllVectors() throws IOException {
 
     private void readVector(float[] value) throws IOException {
       SimpleTextUtil.readLine(in, scratch);
-      // skip leading " [" and strip trailing "]"
-      String s = new BytesRef(scratch.bytes(), 2, scratch.length() - 3).utf8ToString();
+      // skip leading "[" and strip trailing "]"
+      String s = new BytesRef(scratch.bytes(), 1, scratch.length() - 2).utf8ToString();
```
msokolov, line 249: Wow, how did this ever work; we must never have tested it. grr. Thank you for cleaning up!

msokolov (approved):

PR #2075, opened 2020-11-10
LUCENE-9599 Disable sort optim on index sort

mayya-sharipova:
Disable sort optimization in comparators on index sort.

Currently, if search sort is equal or a part of the index sort, we have
an early termination in TopFieldCollector.
But comparators are not aware of the index sort, and may run
sort optimization even if the search sort is congruent with
the index sort.

This patch:
- make leaf comparators aware that search sort is congruent
with the index sort.
- disables sort optimization in comparators in this case.
- removes a private  MultiComparatorLeafCollector class as the only
class that extended that class was TopFieldLeafCollector that
now incorporates the logic of the deleted class.

Relates to #1351

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/TopFieldCollector.java
+++ b/lucene/core/src/java/org/apache/lucene/search/TopFieldCollector.java
@@ -353,7 +353,7 @@ public abstract class TopFieldCollector extends TopDocsCollector<Entry> {
           && queueFull
           && hitsThresholdChecker.isThresholdReached()) {
       assert bottom != null;
-      float minScore = ((FieldComparator.RelevanceComparator) firstComparator).value(bottom.slot);
+      float minScore = (float) firstComparator.value(bottom.slot);
       if (minScore > minCompetitiveScore) {
         scorer.setMinCompetitiveScore(minScore);
         minCompetitiveScore = minScore;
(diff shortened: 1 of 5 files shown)


mayya-sharipova:
I've caught with @jimczi offline, and we have decided that  as all index segments are sorted in the same way, `TopFieldCollector` should inform `FieldComparator`  about index sort only once and not for every leaf.
This change is reflected in 5aa08f51b.

For 8.x branch `FieldComparator::usesIndexSort` method will be combined into [FieldComparator::setCanUsePoints](https://github.com/apache/lucene-solr/blob/branch_8x/lucene/core/src/java/org/apache/lucene/search/FieldComparator.java#L152)

@jimczi This PR is ready for another round of review, please review when you have time

mayya-sharipova:
I've chatted with @jimczi offline, and we have decided to rename `usesIndexSort()` to a more general `disableSkipping` for 
future use cases where `TopFieldCollector` may want to disable skipping functionality of comparators. 

Commit ae710614d1 addresses this change.  @jimczi can you please continue the review when you have time.

lucene/core/src/java/org/apache/lucene/search/TopFieldCollector.java
```diff
@@ -360,8 +352,8 @@ protected void updateMinCompetitiveScore(Scorable scorer) throws IOException {
     if (canSetMinScore
           && queueFull
           && hitsThresholdChecker.isThresholdReached()) {
-      assert bottom != null && relevanceComparator != null;
-      float minScore = relevanceComparator.value(bottom.slot);
+      assert bottom != null;
+      float minScore = ((FieldComparator.RelevanceComparator) firstComparator).value(bottom.slot);
```
jimczi, line 356: Can we avoid the cast here

jimczi (approved):
LGTM

PR #2080, opened 2020-11-13
LUCENE-8947: Skip field length accumulation when norms are disabled

dxl360:
# Description

Lucene accumulates the total length when indexing the field. But when we use custom term frequencies to hold arbitrary scoring signals, Lucene will run into integer overflow error during accumulation if the scoring signals and the number of tokens are too large. This PR aims to fix this issue https://issues.apache.org/jira/browse/LUCENE-8947

# Solution

Skip the field length accumulation when norms is disabled.

# Tests

The test tries to index a field with extremely large custom term frequency 
- Successfully index the field that omits norms
- Expect to trigger the integer overflow error when indexing the same field with norms disabled

# Checklist

### DIFF
--- a/lucene/core/src/test/org/apache/lucene/index/TestCustomTermFreq.java
+++ b/lucene/core/src/test/org/apache/lucene/index/TestCustomTermFreq.java
@@ -458,4 +458,51 @@ public class TestCustomTermFreq extends LuceneTestCase {
 
     IOUtils.close(w, dir);
   }
+
+  // LUCENE-8947: Indexing fails with "too many tokens for field" when using large enough custom term frequencies
+  //              to overflow int on accumulation
+  public void testFieldLengthAccumulation() throws Exception {
+    Directory dir = newDirectory();
+    IndexWriter w = new IndexWriter(dir, new IndexWriterConfig(new MockAnalyzer(random())));
+
+    // when norms is disabled, skip field length accumulation
+    Document doc = new Document();
+    FieldType fieldType = new FieldType(TextField.TYPE_NOT_STORED);
+    fieldType.setIndexOptions(IndexOptions.DOCS_AND_FREQS);
+    fieldType.setOmitNorms(true);
+    Field field = new Field("field",
+            new CannedTermFreqs(new String[] {"foo", "bar"},
+                    new int[] {1073741824, 1073741824}),
+            fieldType);
+    doc.add(field);
+    w.addDocument(doc);
+
+    IndexReader r = DirectoryReader.open(w);
+    PostingsEnum postings = MultiTerms.getTermPostingsEnum(r, "field", new BytesRef("bar"), (int) PostingsEnum.FREQS);
+    assertNotNull(postings);
+    assertEquals(0, postings.nextDoc());
+    assertEquals(1073741824, postings.freq());
+    assertEquals(NO_MORE_DOCS, postings.nextDoc());
+
+    postings = MultiTerms.getTermPostingsEnum(r, "field", new BytesRef("foo"), (int) PostingsEnum.FREQS);
+    assertNotNull(postings);
+    assertEquals(0, postings.nextDoc());
+    assertEquals(1073741824, postings.freq());
+    assertEquals(NO_MORE_DOCS, postings.nextDoc());
+
+    // when norms is enabled, field length accumulation will throw exception due to integer overflow
+    doc.clear();
+    fieldType.setOmitNorms(false);
+    field = new Field("field",
+            new CannedTermFreqs(new String[] {"foo", "bar"},
+                    new int[] {1073741824, 1073741824}),
+            fieldType);
+    doc.add(field);
+    IllegalArgumentException expected = expectThrows(IllegalArgumentException.class, () -> {
+      w.addDocument(doc);
+    });
+    assertTrue(expected.getMessage().contains("too many tokens for field \"field\""));
+
+    IOUtils.close(r, w, dir);
+  }
 }
--- a/lucene/core/src/java/org/apache/lucene/index/IndexingChain.java
+++ b/lucene/core/src/java/org/apache/lucene/index/IndexingChain.java
@@ -1027,7 +1027,9 @@ final class IndexingChain implements Accountable {
           invertState.lastStartOffset = startOffset;
 
           try {
-            invertState.length = Math.addExact(invertState.length, invertState.termFreqAttribute.getTermFrequency());
+            if (fieldInfo.omitsNorms() == false) {
+              invertState.length = Math.addExact(invertState.length, invertState.termFreqAttribute.getTermFrequency());
+            }
           } catch (ArithmeticException ae) {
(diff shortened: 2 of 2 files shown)


lucene/core/src/test/org/apache/lucene/index/TestCustomTermFreq.java
```diff
@@ -458,4 +458,50 @@ public void testFieldInvertState() throws Exception {
 
     IOUtils.close(w, dir);
   }
+
+  // LUCENE-8947: Indexing fails with "too many tokens for field" when using custom term frequencies
```
mikemccand, line 462: Maybe add `when using large enough custom term frequencies to overflow int on accumulation`?

mikemccand (commented):
Looks great @dxl360!  I left one small comment.

lucene/core/src/test/org/apache/lucene/index/TestCustomTermFreq.java
dxl360: Yeah that is a more precise description. I just added it to the comment.

rmuir:
I'm concerned about this change: other things will overflow if you have too many term frequencies in a field. Currently frequency is bounded by 2^32-1 within a doc, and you can only have 2^32-1 documents in the index, so stats like `totalTermFreq` and `sumTotalTermFreq` can't overflow. But with this change it would be easy to do this and break scoring, fail checkindex, etc.

dxl360:
> I'm concerned about this change: other things will overflow if you have too many term frequencies in a field. Currently frequency is bounded by 2^32-1 within a doc, and you can only have 2^32-1 documents in the index, so stats like `totalTermFreq` and `sumTotalTermFreq` can't overflow. But with this change it would be easy to do this and break scoring, fail checkindex, etc.

dxl360:
Original implementation accumulates `int invertState.length` (number of tokens) by term frequency and will overflow if the term frequency is too large. Can we increment `length` by 1 for each token when we use custom term frequencies to hold arbitrary scoring signals (norms is disabled)? In this way, the number of tokens is bounded by 2,147,483,647 and `long totalTermFreq/sumTotalTermFreq` won't overflow.

rmuir:
`totalTermFreq`/`sumTotalTermFreq` are about term frequencies, nothing to do with the norms... but this norms check is the only thing guarding against overflow. we can't just disable the check like this without causing additional problems.

dxl360:
Had offline discussion with @mikemccand. Maybe we can change the type of `invertState.length` from `int` to `long` and keep the current check on field length/termFreq accumulation but safely cast the length back to `int` when calculating the norms. `totalTermFreq` and `sumTotalTermFreq` are both `long` and is not expected to be broken by `invertState.length`.

mikemccand:
I now understand @rmuir 's concern: because today we force sum of term freq within a single document to fit in `int` (during this `invertState.length` accumulation for norms), and because Lucene allows at most `Integer.MAX_VALUE` documents, we know that `totalTermFreq` (per term sum of term freq across all docs) cannot possibly overflow `long`.

Hmm, but I think `sumTotalTermFreq`, which is per field sum of all `totalTermFreq` across all terms in that field, could overflow `long` even today, in and adversarial case.  And it would not be detected by Lucene...

I think it is weird to rely on norms (doc length) accumulation to prevent overflow of `totalTermFreq` and `sumTotalTermFreq`, especially if norms are in fact disabled for that field?

How about decoupling these two problems?  First, let's fix the aggregation of `totalTermFreq` and `sumTotalTermFreq` to explicitly catch any overflow instead of just doing the dangerous `+=` today: https://github.com/apache/lucene-solr/blob/master/lucene/core/src/java/org/apache/lucene/codecs/PushPostingsWriterBase.java#L142 and https://github.com/apache/lucene-solr/blob/master/lucene/core/src/java/org/apache/lucene/codecs/blocktree/BlockTreeTermsWriter.java#L915?  I.e. switch these accumluations to `Math.addExact`.  This will explicitly catch `long` overflow for either of these stats.

Second, let's not accumulate `invertState.length` if norms are disabled (@dxl360 's current PR).

I think this would solve both issues well -- prevent `long` overflow, and allow large custom term frequencies in one document when norms are disabled.

rmuir:
>> Hmm, but I think sumTotalTermFreq, which is per field sum of all totalTermFreq across all terms in that field, could overflow long even today, in and adversarial case. And it would not be detected by Lucene...

I don't think so. I like to think of this as "number of tokens" in the corpus. Because each doc is limited to Integer.MAX_VALUE and there can only be Integer.MAX_VALUE docs, sumTotalTermFreq can't overflow. and totalTermFreq is <= sumTotalTermFreq (it would be equal, in a degraded case where all your documents only have a single word repeated many times).

rmuir:
> How about decoupling these two problems? First, let's fix the aggregation of totalTermFreq and sumTotalTermFreq to explicitly catch any overflow instead of just doing the dangerous += today: https://github.com/apache/lucene-solr/blob/master/lucene/core/src/java/org/apache/lucene/codecs/PushPostingsWriterBase.java#L142 and https://github.com/apache/lucene-solr/blob/master/lucene/core/src/java/org/apache/lucene/codecs/blocktree/BlockTreeTermsWriter.java#L915? I.e. switch these accumluations to Math.addExact. This will explicitly catch long overflow for either of these stats.

I don't think this is correct. You wouldn't trip this until after merge, far after you've already overflowed the values and caused broken search results (assuming you have more than one segment).

mikemccand:
> > > Hmm, but I think sumTotalTermFreq, which is per field sum of all totalTermFreq across all terms in that field, could overflow long even today, in and adversarial case. And it would not be detected by Lucene...
> 
> I don't think so. I like to think of this as "number of tokens" in the corpus. Because each doc is limited to Integer.MAX_VALUE and there can only be Integer.MAX_VALUE docs, sumTotalTermFreq can't overflow. and totalTermFreq is <= sumTotalTermFreq (it would be equal, in a degraded case where all your documents only have a single word repeated many times).

Ahh you're right ... no more than `Integer.MAX_VALUE` tokens in one document, OK.

> > How about decoupling these two problems? First, let's fix the aggregation of totalTermFreq and sumTotalTermFreq to explicitly catch any overflow instead of just doing the dangerous += today: https://github.com/apache/lucene-solr/blob/master/lucene/core/src/java/org/apache/lucene/codecs/PushPostingsWriterBase.java#L142 and https://github.com/apache/lucene-solr/blob/master/lucene/core/src/java/org/apache/lucene/codecs/blocktree/BlockTreeTermsWriter.java#L915? I.e. switch these accumluations to Math.addExact. This will explicitly catch long overflow for either of these stats.
> 
> I don't think this is correct. You wouldn't trip this until after merge, far after you've already overflowed the values and caused broken search results (assuming you have more than one segment).

Hrmph, also correct, boo.

Alright I guess there is nothing we can fix here ... applications simply must not create > `Integer.MAX_VALUE` term frequencies in one doc/field.

PR #2081, opened 2020-11-14
LUCENE-9413: Add CJKWidthCharFilter and its factory.

mocobeta:
This adds a char filter (and its factory) which is the exact counterpart of o.a.l.a.cjk.CJKWidthFilter. The char filter would be useful especially for dictionary-based CJK analyzers; e.g. kuromoji.

https://issues.apache.org/jira/browse/LUCENE-9413


rmuir (approved):
Looks good!

mocobeta:
Merged. Thank you @rmuir for review!

PR #2083, opened 2020-11-16
SOLR-15001 Docker: require init_var_solr.sh

dsmiley:
https://issues.apache.org/jira/browse/SOLR-15001

There are two distinct commits here.  The second is just to the build.gradle.  I struggled with the Docker gradle build because the inputs/outputs were not configured correctly which confused Gradle's cool incremental build.  I also disagree with defining a verbose class file inside a build file when it won't be re-used by other modules -- just do ad-hoc instead.  Build automation is best suited to scripting languages IMO.


HoustonPutman:
I agree with not liking the cumbersome Test class, the reason I added it was to allow for easy inclusion/exclusion rules for test cases. I think this PR loses that functionality. You could add it back in with 

```groovy
testsInclude = propertyOrEnvOrDefault("solr.docker.tests.include", "SOLR_DOCKER_TESTS_INCLUDE", "")
testsExclude = propertyOrEnvOrDefault("solr.docker.tests.exclude", "SOLR_DOCKER_TESTS_EXCLUDE", "")
```

You would just need to edit `help/docker.txt` to make sure it's up to date with how to use the test inputs.

dsmiley:
Ah; thanks for pointing out the point of that configuration; now I see.  I'll remove these changes from this PR and have a separate PR dedicated to overhauling/simplifying the Docker build.

solr/docker/include/scripts/init-var-solr
```diff
@@ -31,31 +31,31 @@ function check_dir_writability {
 }
 
 if [ ! -d "$DIR/data" ]; then
-    echo "Creating $DIR/data"
+    #echo "Creating $DIR/data"
     check_dir_writability "$DIR"
     mkdir "$DIR/data"
     chmod 0770 "$DIR/data"
```
madrob, line 37: Combine these as above?

madrob (changes requested):
This change breaks

```
$ docker run -it apache/solr:9.0.0-SNAPSHOT solr -f
The currently defined JAVA_HOME (/usr/local/openjdk-11) refers to a location
where java was found but jstack was not found. Continuing.

Solr home directory /var/solr/data not found!```

dsmiley:
Mike: Wow; it's embarrassing my PR broke that case.  Thanks for your review!  
I now made init_var_solr occur in the entrypoint, and thus it's guaranteed to run.  Note that users can opt-out by setting NO_INIT_VAR_SOLR=yes if for some reason it's necessary.

madrob (approved):

PR #2084, opened 2020-11-17
LUCENE-9592: Loosen equality checks in TestVectorUtil.

jtibshirani:
TestVectorUtil occasionally fails because of floating point errors. This
change slightly increases the epsilon in equality checks -- testing shows that
this will greatly decrease the chance of failure.


msokolov (approved):
Thanks! Yes, I think your analysis is good - we see people experimenting with quantizing down to a byte-per-sample while maintaining good near-neighbor matching results. https://ai.googleblog.com/2020/07/announcing-scann-efficient-vector.html, for example, showed some promising results in this vein. Are you able to push the commit yet?

jtibshirani:
> Are you able to push the commit yet?

Not yet, I'm still waiting on account credentials.

PR #2085, opened 2020-11-18
LUCENE-9508: Fix DocumentsWriter to block threads until unstalled

s1monw:
DWStallControl expects the caller to loop on top of the wait call to make
progress with flushing if the DW is stalled. This logic wasn't applied such that
DW only stalled for one second and then released the indexing thread. This can cause
OOM if for instance during a full flush one DWPT gets stuck and other threads keep on
indexing.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/DocumentsWriter.java
+++ b/lucene/core/src/java/org/apache/lucene/index/DocumentsWriter.java
@@ -371,19 +371,15 @@ final class DocumentsWriter implements Closeable, Accountable {
   private boolean preUpdate() throws IOException {
     ensureOpen();
     boolean hasEvents = false;
-
-    if (flushControl.anyStalledThreads() || (flushControl.numQueuedFlushes() > 0 && config.checkPendingFlushOnUpdate)) {
+    while (flushControl.anyStalledThreads() || (flushControl.numQueuedFlushes() > 0 && config.checkPendingFlushOnUpdate)) {
       // Help out flushing any queued DWPTs so we can un-stall:
-      do {
-        // Try pick up pending threads here if possible
-        DocumentsWriterPerThread flushingDWPT;
-        while ((flushingDWPT = flushControl.nextPendingFlush()) != null) {
-          // Don't push the delete here since the update could fail!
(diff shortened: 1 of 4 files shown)


s1monw:
@mikemccand can you take a look pls

lucene/core/src/test/org/apache/lucene/index/TestIndexWriter.java
```diff
@@ -4258,4 +4258,47 @@ public void testPendingNumDocs() throws Exception {
       }
     }
   }
+
+  public void testIndexWriterBlocksOnStall() throws IOException, InterruptedException {
+    try (Directory dir = newDirectory()) {
+      try (IndexWriter writer = new IndexWriter(dir, newIndexWriterConfig())) {
```
dweiss, line 4264: pull it up into the same try (semicolon-separated)?

lucene/core/src/java/org/apache/lucene/index/DocumentsWriter.java
```diff
@@ -371,19 +371,15 @@ public void close() throws IOException {
   private boolean preUpdate() throws IOException {
     ensureOpen();
     boolean hasEvents = false;
-
-    if (flushControl.anyStalledThreads() || (flushControl.numQueuedFlushes() > 0 && config.checkPendingFlushOnUpdate)) {
```
dweiss, line 375: oh, wonderful.

dweiss (approved):
LGTM.

lucene/core/src/test/org/apache/lucene/index/TestIndexWriter.java
s1monw: it's more clear that way what is closed first which is important here. That's why I do it this way.

lucene/core/src/java/org/apache/lucene/index/DocumentsWriter.java
s1monw: yeah, details...

dnhatn (approved):
Great catch! LGTM.

PR #2087, opened 2020-11-18
LUCENE-9613: Split ordinals into blocks when it saves 10+% space.

jpountz:
This changes the way that doc values write ordinals so that Lucene splits into
blocks if it would save more than 10% space, like we do for numeric fields. By
the way, we are now using the same code path to write ordinals and numbers.


jpountz:
Superseded by https://github.com/apache/lucene/pull/186

PR #2088, opened 2020-11-18
LUCENE-9617: Reset lowestUnassignedFieldNumber in FieldNumbers.clear()

msfroh:
FieldNumbers.clear() is called from IndexWriter.deleteAll(), which is
supposed to completely reset the state of the index. This includes
clearing all known fields.

Prior to this change, it would allocate progressively higher field
numbers, which results in larger and  larger arrays for
FieldInfos.byNumber, effectively "leaking" field numbers every time
deleteAll() is called.

# Description

If you run a loop that repeatedly adds documents to an IndexWriter and calls deleteAll, new fields will be given progressively higher index numbers. Since these index numbers are reflected in the size of the FieldInfos.byNumber array, this is effectively a memory leak.

# Solution

Reset lowestUnassignedFieldNumber to -1 when FieldNumbers.clear() is called (resetting the state to that of a newly-created FieldNumbers instance).

# Tests

Added a unit test to confirm that after calling clear(), the next field is given number 0.

# Checklist

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -184,8 +184,8 @@ Bug fixes
 * LUCENE-9365: FuzzyQuery was missing matches when prefix length was equal to the term length
   (Mark Harwood, Mike Drob)
 
-* LUCENE-9617: Reset lowestUnassignedFieldNumber on FieldNumbers.clear(), to avoid leaking
-  field numbers on IndexWriter.deleteAll(). (Michael Froh)
+* LUCENE-9617: Fix per-field memory leak in IndexWriter.deleteAll(). Reset next available internal
+  field number to 0 on FieldInfos.clear(), to avoid wasting FieldInfo references. (Michael Froh)
 
 Other
 
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -184,6 +184,9 @@ Bug fixes
 * LUCENE-9365: FuzzyQuery was missing matches when prefix length was equal to the term length
   (Mark Harwood, Mike Drob)
 
+* LUCENE-9617: Reset lowestUnassignedFieldNumber on FieldNumbers.clear(), to avoid leaking
+  field numbers on IndexWriter.deleteAll(). (Michael Froh)
+
 Other
 
 * LUCENE-9312: Allow gradle builds against arbitrary JVMs. (Tomoko Uchida, Dawid Weiss)
--- a/lucene/core/src/test/org/apache/lucene/index/TestFieldInfos.java
+++ b/lucene/core/src/test/org/apache/lucene/index/TestFieldInfos.java
@@ -187,4 +187,23 @@ public class TestFieldInfos extends LuceneTestCase {
     writer.close();
     dir.close();
   }
+
+  public void testFieldNumbersAutoIncrement() {
+    FieldInfos.FieldNumbers fieldNumbers = new FieldInfos.FieldNumbers("softDeletes");
+    for (int i = 0; i < 10; i++) {
+      fieldNumbers.addOrGet("field" + i, -1, IndexOptions.NONE, DocValuesType.NONE,
+          0, 0, 0, 0,
+          VectorValues.SearchStrategy.NONE, false);
+    }
+    int idx = fieldNumbers.addOrGet("EleventhField", -1, IndexOptions.NONE, DocValuesType.NONE,
+        0, 0, 0, 0,
+        VectorValues.SearchStrategy.NONE, false);
+    assertEquals("Field numbers 0 through 9 were allocated", 10, idx);
+
+    fieldNumbers.clear();
+    idx = fieldNumbers.addOrGet("PostClearField", -1, IndexOptions.NONE, DocValuesType.NONE,
+        0, 0, 0, 0,
+        VectorValues.SearchStrategy.NONE, false);
+    assertEquals("Field numbers should reset after clear()", 0, idx);
+  }
 }
--- a/lucene/core/src/java/org/apache/lucene/index/FieldInfos.java
+++ b/lucene/core/src/java/org/apache/lucene/index/FieldInfos.java
@@ -481,6 +481,7 @@ public class FieldInfos implements Iterable<FieldInfo> {
       indexOptions.clear();
       docValuesType.clear();
       dimensions.clear();
+      lowestUnassignedFieldNumber = -1;
     }
 
     synchronized void setIndexOptions(int number, String name, IndexOptions indexOptions) {
--- a/lucene/core/src/test/org/apache/lucene/index/TestIndexWriterDelete.java
+++ b/lucene/core/src/test/org/apache/lucene/index/TestIndexWriterDelete.java
@@ -20,6 +20,7 @@ package org.apache.lucene.index;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.PrintStream;
+import java.io.UncheckedIOException;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
@@ -28,6 +29,7 @@ import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLong;
 
 import org.apache.lucene.analysis.Analyzer;
 import org.apache.lucene.analysis.MockAnalyzer;
@@ -36,6 +38,7 @@ import org.apache.lucene.document.Document;
 import org.apache.lucene.document.Field;
 import org.apache.lucene.document.FieldType;
 import org.apache.lucene.document.NumericDocValuesField;
+import org.apache.lucene.document.StoredField;
 import org.apache.lucene.document.StringField;
 import org.apache.lucene.search.IndexSearcher;
 import org.apache.lucene.search.ScoreDoc;
@@ -355,17 +358,17 @@ public class TestIndexWriterDelete extends LuceneTestCase {
     if (VERBOSE) {
       System.out.println("\nTEST: now final deleteAll");
     }
-    
+
     modifier.deleteAll();
     for (Thread thread : threads) {
       thread.join();
     }
-    
+
     if (VERBOSE) {
       System.out.println("\nTEST: now close");
     }
     modifier.close();
-    
+
     DirectoryReader reader = DirectoryReader.open(dir);
     if (VERBOSE) {
       System.out.println("\nTEST: got reader=" + reader);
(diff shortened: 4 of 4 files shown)


rmuir:
I'm suspicious that this is safe to do. What if another thread is calling addDocument at the same time?

msfroh:
> I'm suspicious that this is safe to do. What if another thread is calling addDocument at the same time?

As long as `FieldNumbers.clear()` is only called from `IndexWriter.deleteAll()`, my understanding is that the safety is provided by the `try (Closeable finalizer = docWriter.lockAndAbortAll()) {` block, which (I think) guarantees that any concurrent indexing is blocked until the lock is released.

lucene/CHANGES.txt
```diff
@@ -184,6 +184,9 @@ Bug fixes
 * LUCENE-9365: FuzzyQuery was missing matches when prefix length was equal to the term length
   (Mark Harwood, Mike Drob)
 
+* LUCENE-9617: Reset lowestUnassignedFieldNumber on FieldNumbers.clear(), to avoid leaking
```
dsmiley, line 187: Can you rewrite in terms of what a user might understand?  e.g.
`IndexWriter.deleteAll now resets internal field numbers; prevents ever-increasing numbers in unusual use-cases`
The latter part of what you wrote isn't bad but the first part is technical mumbo-jumbo that only Lucene deep divers would even recognize.

lucene/core/src/test/org/apache/lucene/index/TestFieldInfos.java
```diff
@@ -187,4 +187,23 @@ public void testMergedFieldInfos_singleLeaf() throws IOException {
...
+      fieldNumbers.addOrGet("field" + i, -1, IndexOptions.NONE, DocValuesType.NONE,
+          0, 0, 0, 0,
+          VectorValues.SearchStrategy.NONE, false);
+    }
+    int idx = fieldNumbers.addOrGet("EleventhField", -1, IndexOptions.NONE, DocValuesType.NONE,
+        0, 0, 0, 0,
+        VectorValues.SearchStrategy.NONE, false);
+    assertEquals("Field numbers 0 through 9 were allocated", 10, idx);
+
+    fieldNumbers.clear();
```
dsmiley, line 203: My only problem with unit tests like this is that it doesn't test what we _really_ want to know -- that when IW.deleteAll() is called (a user level thing, fieldNumbers.clear() is not), that the field numbers get reset.

dsmiley (changes requested):

rmuir:
> As long as FieldNumbers.clear() is only called from IndexWriter.deleteAll(), my understanding is that the safety is provided by the try (Closeable finalizer = docWriter.lockAndAbortAll()) { block, which (I think) guarantees that any concurrent indexing is blocked until the lock is released.

Thanks, maybe there is a way to improve the testing of `deleteAll` to better enforce this? Lucene is not testing this method much, but I know some users (e.g. Solr) are using it often. My concern is some race condition that ultimately creates segments with unaligned field numbers. This would be a disaster and definitely result in corruption (think, stored fields merging etc which copies binary/compressed data directly).

lucene/core/src/test/org/apache/lucene/index/TestFieldInfos.java
mikemccand: Uhm, but this is the purpose of a unit-test?

It should test the lowest "unit" of the software, in this case `FieldInfos` and its `clear` method.

The fact that `IW.deleteAll()` is the "user space manifestation" of this particular low level problem is actually unimportant, from the standpoint of this unit test.  The unit test should only test the smallest / shrink-wrapped problem, which is indeed that `FieldInfos.clear()` failed to do its job correctly.

We could consider adding a unit test that "IW.deleteAll does not leak memory", but I don't think that is required for committing this fix to an obvious bug  ...

mikemccand:
> > As long as FieldNumbers.clear() is only called from IndexWriter.deleteAll(), my understanding is that the safety is provided by the try (Closeable finalizer = docWriter.lockAndAbortAll()) { block, which (I think) guarantees that any concurrent indexing is blocked until the lock is released.
> 
> Thanks, maybe there is a way to improve the testing of `deleteAll` to better enforce this? Lucene is not testing this method much, but I know some users (e.g. Solr) are using it often. My concern is some race condition that ultimately creates segments with unaligned field numbers. This would be a disaster and definitely result in corruption (think, stored fields merging etc which copies binary/compressed data directly).

Maybe a new unit test that randomly calls `deleteAll()` and other IW APIs across threads?

Hmm, actually, it looks like `TestIndexWriter.testThreadInterruptDeadlock` is doing just that already.

`TestIndexWriterDelete` tests specific interesting cases.  `TestIndexWriterMergePolicy` seems to test the interaction of `deleteAll` and merge-on-commit.

lucene/CHANGES.txt
mikemccand: How about something like this?:

```
Fix FieldInfos.clear() to reset its internal next field number state back to 0, which was otherwise causing a slow memory leak for applications relying on IndexWriter.deleteAll.
```

mikemccand:
Maybe we could add a simple unit test here that keeps re-indexing the same small set of docs with many unique fields, then calls `IW.deleteAll`, many times?  Such that the test would OOME on trunk today, and then pass with the fix?  Would that test take too long to OOME (how slow is the leak?)?  If so we could mark it as `@Slow`.

lucene/CHANGES.txt
msfroh: Thinking backwards from the user (who probably cares about the slow memory leak more than anything), how about:

```
Fix per-field memory leak in IndexWriter.deleteAll(). Reset next available internal field number to 0 on FieldInfos.clear(), to avoid wasting FieldInfo references.
```

It gets the user-visible impact out front, explains the fix, and what was leaking, all in about 20 words.

msfroh:
> Maybe we could add a simple unit test here that keeps re-indexing the same small set of docs with many unique fields, then calls IW.deleteAll, many times? Such that the test would OOME on trunk today, and then pass with the fix? Would that test take too long to OOME (how slow is the leak?)? If so we could mark it as @Slow.

I'll give that a shot. With trunk, my code ran for about 12 hours and was still limping along (at a rate of ~5 documents/second, versus ~250 documents/second in the first minute). But I was doing some heavy analysis and flushing to disk. Adding a doc with a bunch of doc value fields, calling deleteAll(), and repeating, I think the problem should reproduce pretty quickly. I think I still need to trigger a `FieldInfos.finish()` to force it to allocate the large array. Alternatively, we would get an overflow on `lowestUnassignedFieldNumber` (though it looks like FieldNumbers will happily just give out negative field numbers in that case).

Anyway, I'll try to come up with a test that breaks quickly.

mikemccand:
OK thanks @msfroh.  This indeed sounds like a very inefficient way to test for this particular bug!  But maybe you could mark it `@Monster`, like e.g. `Test2BBinaryDocValues` and such?  These tests get run only during release (if even, hmm).

I think it is out of scope for you to prove, here, that usage of this `FieldInfos.clear()` API is 100% safe -- we cannot boil the ocean with otherwise well contained bug fixes like this one.  Progress not perfection :)

Maybe we can at least add a warning to javadocs for `IndexWriter.deleteAll` that any concurrently running indexing operations (`addDocument`, etc.) may or may not be deleted, i.e. the behavior of such concurrent operations are not defined / predictable.

lucene/core/src/test/org/apache/lucene/index/TestFieldInfos.java
dsmiley: I agree that nothing more than what was done is _required_.  My feedback reflects my low opinion of the value of some unit tests like this one.  I've seen low level unit tests before that seem nice at a glance yet the component still doesn't work as intended due to other reasons.  From the conversation over the course of today I see here... it seems my point has some credence ;-)

msfroh:
I managed to trigger an OutOfMemoryError, at least within my IDE by burning through 50 million fields, then trying to add one more document and flushing.

Curiously, the test wasn't failing on the OOME itself, but rather we fail an assertion when trying to rollback on tragedy (where the tragedy was the OOME). Here's that stack trace:

```
java.lang.AssertionError: pendingNumDocs 1 != 0 totalMaxDoc
	at __randomizedtesting.SeedInfo.seed([ABBF17C4E0FCDEE5:DDC8E99910AFC8FF]:0)
	at org.apache.lucene.index.IndexWriter.rollbackInternal(IndexWriter.java:2398)
	at org.apache.lucene.index.IndexWriter.maybeCloseOnTragicEvent(IndexWriter.java:5196)
	at org.apache.lucene.index.IndexWriter.tragicEvent(IndexWriter.java:5186)
	at org.apache.lucene.index.IndexWriter.doFlush(IndexWriter.java:3932)
	at org.apache.lucene.index.IndexWriter.flush(IndexWriter.java:3874)
	at org.apache.lucene.index.IndexWriter.flush(IndexWriter.java:3853)
	at org.apache.lucene.index.TestIndexWriterDelete.testDeleteAllRepeated(TestIndexWriterDelete.java:496)
```

mikemccand:
> Curiously, the test wasn't failing on the OOME itself, but rather we fail an assertion when trying to rollback on tragedy (where the tragedy was the OOME). Here's that stack trace:

Well, that is spooky -- this looks like a pre-existing bug in `IndexWriter`'s `abort`!  But I don't think we must fix that before committing this fix?  Can you open a new issue for it?

msfroh:
> Well, that is spooky -- this looks like a pre-existing bug in `IndexWriter`'s `abort`! But I don't think we must fix that before committing this fix? Can you open a new issue for it?

https://issues.apache.org/jira/browse/LUCENE-9621

mikemccand:
Great, thank you for adding the new "user space impact" test case @msfroh (in addition to the low-level unit test)!  @dsmiley does that address your concern about the lower level unit test?

Net/net I think this is ready -- I think the existing tests we have that concurrently {{deleteAll}} help address @rmuir's concern too.  I will beast this change for a while to see if there are any interesting failures ...

dsmiley (approved):

s1monw:
> Well, that is spooky -- this looks like a pre-existing bug in IndexWriter's abort! But I don't think we must fix that before committing this fix? Can you open a new issue for it?

I am not sure we can guarantee anything on an OOM? 

> I'm suspicious that this is safe to do. What if another thread is calling addDocument at the same time?

The explanation from @msfroh is correct here we basically stop the world here since we hold the IW lock, lock all DWPT, prevent the creation of new DWPT and abort all merges. That's basically all we can do here. I made the ask for removing this method a couple of times since it creates a ton of friction and is hard to test. IMO we should rather ask the user to close the IW, open a new one that re-creates the index altogether.

mikemccand:
> I am not sure we can guarantee anything on an OOM?

Yeah I suppose that is essentially nearly hopeless, sigh -- this is why `IndexWriter` aborts on OOME.

> The explanation from @msfroh is correct here we basically stop the world here since we hold the IW lock, lock all DWPT, prevent the creation of new DWPT and abort all merges. That's basically all we can do here.

+1

> I made the ask for removing this method a couple of times since it creates a ton of friction and is hard to test. IMO we should rather ask the user to close the IW, open a new one that re-creates the index altogether.

I think this method serves an important purpose/use-case: it lets you efficiently clear the whole index, transactionally, so that e.g. searchers can see the old index, and then the new index, without seeing the fully deleted index at all.  I think we should fix its bugs instead!

s1monw:
> I think this method serves an important purpose/use-case: it lets you efficiently clear the whole index, transactionally, so that e.g. searchers can see the old index, and then the new index, without seeing the fully deleted index at all. I think we should fix its bugs instead!

I know, that's what you said about the last 2 subtile bugs.

mikemccand:
> > I think this method serves an important purpose/use-case: it lets you efficiently clear the whole index, transactionally, so that e.g. searchers can see the old index, and then the new index, without seeing the fully deleted index at all. I think we should fix its bugs instead!
> 
> I know, that's what you said about the last 2 subtile bugs.

Fair enough ;)

mikemccand:
I beasted this PR for 36 iterations and uncovered only an [unrelated issue](https://issues.apache.org/jira/browse/LUCENE-9638) -- I think this is ready!  I'll push soon.  Thanks @msfroh!

PR #2089, opened 2020-11-18
SOLR-14999: Option to set the advertised port for Solr.

HoustonPutman:
https://issues.apache.org/jira/browse/SOLR-14999

I am open to suggestions on the naming of this new option. Currently I have `-Dsolr.port.advertise` and `SOLR_PORT_ADVERTISE`, because I think those are more descriptive names than `hostPort`.

This is backwards compatible with the current `solr.xml`s, because the default value (`jetty.port`) is now checked in the Java code, rather than the xml.


solr/solr-ref-guide/src/format-of-solr-xml.adoc
```diff
@@ -134,9 +134,10 @@ The hostname Solr uses to access cores.
 The url context path.
 
 `hostPort`::
-The port Solr uses to access cores.
+The port Solr uses to access cores, and advertise Solr node locations through liveNodes.
 +
-In the default `solr.xml` file, this is set to `${jetty.port:8983}`, which will use the Solr port defined in Jetty, and otherwise fall back to 8983.
+In the default `solr.xml` file, this is set to `${solr.port.advertise:0}`.
```
madrob, line 139: I'd like a sentence here giving an example why the user might care about it, i.e. solr running behind a proxy.

PR #2092, opened 2020-11-20
SOLR-15009 Propogate IOException from DF.exists

madrob:
https://issues.apache.org/jira/browse/SOLR-15009

### DIFF
--- a/solr/core/src/java/org/apache/solr/core/CachingDirectoryFactory.java
+++ b/solr/core/src/java/org/apache/solr/core/CachingDirectoryFactory.java
@@ -16,9 +16,11 @@
  */
 package org.apache.solr.core;
 
-import java.io.File;
 import java.io.IOException;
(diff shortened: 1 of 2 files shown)


madrob:
@dsmiley  you might find this interesting based on your other comment about Path v File

dsmiley (approved):
LGTM

madrob:
@TomMD musebot seems to be stuck - any ideas?

TomMD:
@madrob I see a backend failure which lacks sufficient explanation.  We have opened a ticket to investigate further.  In the mean time I have re-started analysis of this PR.  I can babysit this - if the job does not finish within the next hour then we'll have more to investigate.  Thank you for asking about this, it's certainly not a situation that should come up.

madrob:
Committed in cb5ba42bd7d9777eaf76705f17a3407fd2897b10

PR #2093, opened 2020-11-23
LUCENE-9606: Wrap boolean queries generated by shape fields with a Constant score query

iverase:
When querying a shape field with a Geometry collection and a CONTAINS spatial relationship, the query is rewritten as a boolean query. We should wrap the resulting query with a ConstantScoreQuery.


dsmiley (approved):

iverase:
Thanks @dsmiley!

PR #2101, opened 2020-11-27
SOLR-15016 Replica placement plugins should use container plugins API / configs

sigram:
See Jira for more details.


sigram:
This branch includes changes in branch `jira/solr-15004`.

solr/core/src/java/org/apache/solr/core/CoreContainer.java
```diff
@@ -896,6 +900,9 @@ public void load() {
       containerHandlers.getApiBag().registerObject(containerPluginsApi.readAPI);
       containerHandlers.getApiBag().registerObject(containerPluginsApi.editAPI);
 
+      // get the placement plugin
```
murblanc, line 902: I'd rather comment "get the placement plugin **factory**"
And possibly specify what's the plugin factory lifecycle wrt configuration? Now that everything is a bit more implicit than it previously was, I don't get (yet...) when exactly the configuration is passed. I believe comments related to this would be useful.

solr/core/src/java/org/apache/solr/cluster/placement/PlacementPluginFactory.java
```diff
@@ -18,14 +18,22 @@
...
 
 /**
- * Factory implemented by client code and configured in {@code solr.xml} allowing the creation of instances of
+ * Factory implemented by client code and configured in container plugins allowing the creation of instances of
  * {@link PlacementPlugin} to be used for replica placement computation.
+ * <p>Note: configurable factory implementations should also implement
+ * {@link org.apache.solr.api.ConfigurablePlugin} with the appropriate configuration
+ * bean type.</p>
  */
 public interface PlacementPluginFactory {
```
murblanc, line 27: Shouldn't this interface extend `ConfigurablePlugin<T extends MapWriter>` so that concrete config classes such as `AffinityPlacementFactory` do not have to implement multiple interfaces?

murblanc: That would force every placement plugin to be configurable though I guess...

solr/core/src/java/org/apache/solr/core/CoreContainer.java
```diff
@@ -257,6 +260,7 @@ public CoreLoadFailure(CoreDescriptor cd, Exception loadFailure) {
 
   // initially these are the same to collect the plugin-based listeners during init
   private ClusterEventProducer clusterEventProducer;
+  private PlacementPluginFactory placementPluginFactory;
```
murblanc, line 262: I believe we have a synchronization issue here on access to that variable. It is not `final` nor `volatile`, access is not synchronized but it is accessed from multiple threads (the command execution Overseer threads calling `getPlacementPluginFactory()`).

sigram: Since we always use `DelegatingPlacementPluginFactory` I'm going to create a final instance of this wrapper - and then it will be initialized once the plugins registry is ready.

solr/core/src/java/org/apache/solr/cloud/api/collections/AddReplicaCmd.java
```diff
@@ -144,7 +140,8 @@ public void call(ClusterState state, ZkNodeProps message, @SuppressWarnings({"ra
       }
     }
 
-    List<CreateReplica> createReplicas = buildReplicaPositions(ocmh.cloudManager, clusterState, collectionName, message, replicaTypesVsCount)
+    List<CreateReplica> createReplicas = buildReplicaPositions(ocmh.cloudManager, clusterState, collectionName, message, replicaTypesVsCount,
+        ocmh.overseer.getCoreContainer().getPlacementPluginFactory().createPluginInstance())
```
murblanc, line 144: Unclear to me: what happens here when placement plugins are not configured and we use for example the legacy assign strategy?

murblanc: Ok, I see that's how we tell which one is configured by this value being `null`... Can't say I really like it, but at least it should be commented (here and in `PlacementPluginFactory` as well?).

sigram: The plugin is null and `Assign.createAssignStrategy` provides a `LegacyAssignStrategy`.

solr/core/src/java/org/apache/solr/api/ContainerPluginsRegistry.java
```diff
@@ -239,8 +238,7 @@ private static String getActualPath(ApiInfo apiInfo, String path) {
 
   @SuppressWarnings({"rawtypes", "unchecked"})
   private static  Map<String, String> getTemplateVars(PluginMeta pluginMeta) {
-    Map result = makeMap("plugin-name", pluginMeta.name, "path-prefix", pluginMeta.pathPrefix);
-    return result;
+    return (Map) makeMap("plugin-name", pluginMeta.name, "path-prefix", pluginMeta.pathPrefix);
```
murblanc, line 241: Should replace the call to `makeMap` by `Map.of(...)`. This allows removing the cast and removing the `@SuppressWarnings`.

solr/core/src/java/org/apache/solr/api/ContainerPluginsRegistry.java
```diff
@@ -422,7 +421,10 @@ public static Class getConfigClass(ConfigurablePlugin<?> o) {
       for (Type type : interfaces) {
         if (type instanceof ParameterizedType) {
           ParameterizedType parameterizedType = (ParameterizedType) type;
-          if (parameterizedType.getRawType() == ConfigurablePlugin.class) {
+          Type rawType = parameterizedType.getRawType();
+          if (rawType == ConfigurablePlugin.class ||
```
murblanc, line 425: Is it worth keeping `rawType == ConfigurablePlugin.class` in the if condition? It's covered by the second operand of the OR.

solr/core/src/java/org/apache/solr/cluster/placement/PlacementPluginFactory.java
```diff
@@ -17,15 +17,37 @@
...
+   */
+  String PLUGIN_NAME = ".placement-plugin";
+
+  /**
+   * Returns an instance of the plugin that will be repeatedly (and concurrently) called to compute placement. Multiple
    * instances of a plugin can be used in parallel (for example if configuration has to change, but plugin instances with
    * the previous configuration are still being used).
    */
-  PlacementPlugin createPluginInstance(PlacementPluginConfig config);
+  PlacementPlugin createPluginInstance();
```
murblanc, line 40: See previous comment in [AddReplicaCmd](https://github.com/apache/lucene-solr/pull/2101/files#diff-9190d76945a6b15eee98e2a3725de99d342c635b87cec2992cc2a443c247e3e7R144).
A comment should be added here saying that when this returns null it means legacy assign strategy.

solr/core/src/java/org/apache/solr/api/ContainerPluginsRegistry.java
sigram: Right, I wrestled with this expressions to make it work and forgot to simplify.

solr/core/src/java/org/apache/solr/cluster/events/impl/CollectionsRepairEventListener.java
```diff
@@ -76,10 +78,12 @@
   private int waitForSecond = DEFAULT_WAIT_FOR_SEC;
 
   private ScheduledThreadPoolExecutor waitForExecutor;
+  private PlacementPluginFactory<? extends MapWriter> placementPluginFactory;
```
murblanc, line 81: Should this rather be `<? extends PlacementPluginConfig>`?

sigram: Yes.

solr/core/src/java/org/apache/solr/core/CoreContainer.java
```diff
@@ -2180,6 +2188,10 @@ public ClusterEventProducer getClusterEventProducer() {
     return clusterEventProducer;
   }
 
+  public PlacementPluginFactory<? extends MapWriter> getPlacementPluginFactory() {
```
murblanc, line 2191: Shouldn't this be `<? extends PlacementPluginConfig>`?

solr/core/src/test/org/apache/solr/cluster/placement/impl/PlacementPluginIntegrationTest.java
```diff
@@ -109,4 +132,108 @@ public void testMinimizeCores() throws Exception {
     assertEquals("min cores too low", 1, minCores);
   }
 
+  @Test
+  @SuppressWarnings("unchecked")
+  public void testDynamicReconfiguration() throws Exception {
+    PlacementPluginFactory<? extends MapWriter> pluginFactory = cc.getPlacementPluginFactory();
```
murblanc, line 138: `<? extends PlacementPluginConfig>`?

solr/core/src/test/org/apache/solr/cluster/placement/impl/PlacementPluginIntegrationTest.java
```diff
@@ -109,4 +132,108 @@ public void testMinimizeCores() throws Exception {
...
+        .forceV2(true)
+        .POST()
+        .withPayload(singletonMap("add", plugin))
+        .build();
+    req.process(cluster.getSolrClient());
+
+    version = waitForVersionChange(version, wrapper, 10);
+
+    assertTrue("wrong version " + version, version > 0);
+    PlacementPluginFactory<? extends MapWriter> factory = wrapper.getDelegate();
```
murblanc, line 158: <? extends PlacementPluginConfig>?

sigram: Gah .. well-spotted all these. I'll fix them later today.

solr/core/src/java/org/apache/solr/cluster/events/ClusterEventProducer.java
```diff
@@ -26,7 +26,7 @@
 public interface ClusterEventProducer extends ClusterSingleton, Closeable {
 
   /** Unique name for the registration of a plugin-based implementation. */
-  String PLUGIN_NAME = "cluster-event-producer";
+  String PLUGIN_NAME = ".cluster-event-producer";
```
murblanc, line 29: I don't understand why "." was added, therefore I suggest adding a comment explaining it.

solr/core/src/java/org/apache/solr/cluster/events/impl/ClusterEventProducerFactory.java
```diff
@@ -128,13 +128,17 @@ public void added(ContainerPluginsRegistry.ApiInfo plugin) {
...
         } else if (instance instanceof ClusterEventProducer) {
-          // replace the existing impl
-          if (cc.getClusterEventProducer() instanceof DelegatingClusterEventProducer) {
-            ((DelegatingClusterEventProducer) cc.getClusterEventProducer())
-                .setDelegate((ClusterEventProducer) instance);
+          if (ClusterEventProducer.PLUGIN_NAME.equals(plugin.getInfo().name)) {
+            // replace the existing impl
+            if (cc.getClusterEventProducer() instanceof DelegatingClusterEventProducer) {
+              ((DelegatingClusterEventProducer) cc.getClusterEventProducer())
+                  .setDelegate((ClusterEventProducer) instance);
```
murblanc, line 135: It seems we still have synchronization issues on access to `DelegatingClusterEventProducer.delegate`, right? Might not be on this PR though (and I might have it wrong because I don't fully understand this plugin and events architecture) but in `DelegatingClusterEventProducer` I see non sync access to `delegate` and I wonder...

solr/core/src/java/org/apache/solr/cluster/events/impl/ClusterEventProducerFactory.java
```diff
@@ -149,21 +153,25 @@ public void deleted(ContainerPluginsRegistry.ApiInfo plugin) {
           ClusterEventListener listener = (ClusterEventListener) instance;
           clusterEventProducer.unregisterListener(listener);
         } else if (instance instanceof ClusterEventProducer) {
-          // replace the existing impl with NoOp
-          if (cc.getClusterEventProducer() instanceof DelegatingClusterEventProducer) {
-            ((DelegatingClusterEventProducer) cc.getClusterEventProducer())
-                .setDelegate(new NoOpProducer(cc));
+          if (ClusterEventProducer.PLUGIN_NAME.equals(plugin.getInfo().name)) {
```
murblanc, line 156: There's quite some duplication between this listener method (`deleted`) and method `added` above. Likely could use some factoring out (and maybe slight rework of the logs to differentiate adding and deleting?).

solr/core/src/java/org/apache/solr/cluster/events/ClusterEventProducer.java
sigram: There was a Slack discussion on the naming scheme for pre-defined plugins (on #solr-plugins) and the overall sentiment was to use a dot prefix for such names. It's a convention not a requirement - and it will be documented in SOLR-15022.

solr/core/src/java/org/apache/solr/cluster/events/impl/ClusterEventProducerFactory.java
sigram: I debated whether to do it in this PR or separately... strictly speaking it's not a part of this PR, but I refactored a similar pattern in a listener that belongs to this PR ... so maybe I'll do it here :)

solr/core/src/java/org/apache/solr/cluster/placement/PlacementPluginFactory.java
```diff
@@ -17,15 +17,39 @@
 
 package org.apache.solr.cluster.placement;
 
+import org.apache.solr.api.ConfigurablePlugin;
+
 /**
- * Factory implemented by client code and configured in {@code solr.xml} allowing the creation of instances of
+ * Factory implemented by client code and configured in container plugins allowing the creation of instances of
```
murblanc, line 23: Can we add a reference to how "container plugins" are configured? Not obvious for me and I guess I'm likely not the only one...

murblanc: (referencing the Affinity class and its configuration would likely be good enough here)

solr/core/src/java/org/apache/solr/cluster/placement/impl/PlacementPluginFactoryLoader.java
```diff
@@ -0,0 +1,80 @@
...
+    }
+    ContainerPluginsRegistry.PluginRegistryListener pluginListener = new ContainerPluginsRegistry.PluginRegistryListener() {
+      @Override
+      public void added(ContainerPluginsRegistry.ApiInfo plugin) {
+        if (plugin == null || plugin.getInstance() == null) {
+          return;
+        }
+        Object instance = plugin.getInstance();
+        if (instance instanceof PlacementPluginFactory) {
+          setDelegate(plugin.getInfo(), instance);
```
murblanc, line 49: 1/3: maybe we should cast the second parameter `instance` to `PlacementPluginFactory` here.

solr/core/src/java/org/apache/solr/cluster/placement/impl/PlacementPluginFactoryLoader.java
```diff
@@ -0,0 +1,80 @@
...
+          setDelegate(plugin.getInfo(), null);
+        }
+      }
+
+      @Override
+      public void modified(ContainerPluginsRegistry.ApiInfo old, ContainerPluginsRegistry.ApiInfo replacement) {
+        added(replacement);
+      }
+
+      private void setDelegate(PluginMeta pluginMeta, Object instance) {
```
murblanc, line 69: 2/3: change the signature of this method so that `instance` is a `PlacementPluginFactory`.

solr/core/src/java/org/apache/solr/cluster/placement/impl/PlacementPluginFactoryLoader.java
```diff
@@ -0,0 +1,80 @@
...
+        if (PlacementPluginFactory.PLUGIN_NAME.equals(pluginMeta.name)) {
+          pluginFactory.setDelegate((PlacementPluginFactory<? extends PlacementPluginConfig>) instance);
```
murblanc, line 71: 3/3: remove the cast here, `instance` would be of the expected type.

solr/core/src/java/org/apache/solr/cluster/placement/plugins/MinimizeCoresPlacementFactory.java
```diff
@@ -40,13 +40,23 @@
...
+public class MinimizeCoresPlacementFactory implements PlacementPluginFactory<PlacementPluginFactory.NoConfig> {
 
   @Override
-  public PlacementPlugin createPluginInstance(PlacementPluginConfig config) {
+  public PlacementPlugin createPluginInstance() {
     return new MinimizeCoresPlacementPlugin();
   }
 
+  @Override
+  public void configure(NoConfig cfg) {
```
murblanc, line 51: Maybe `configure()` in `ConfigurablePlugin<T extends MapWriter>` should have a default empty implementation so we don't have to repeat it (here, in `DelegatingPlacementPluginFactory` and `RandomPlacementFactory`).

solr/core/src/java/org/apache/solr/cluster/placement/plugins/MinimizeCoresPlacementFactory.java
```diff
@@ -40,13 +40,23 @@
...
+    // no-op
+  }
+
+  @Override
+  public NoConfig getConfig() {
```
murblanc, line 56: Maybe similarly here, `PlacementPluginFactory` should provide a default implementation returning `NoConfig.INSTANCE` given that out of 4 implementations of the interface, only one returns something else (`AffinityPlacementFactory`).

solr/core/src/java/org/apache/solr/cluster/placement/plugins/RandomPlacementFactory.java
```diff
@@ -0,0 +1,91 @@
...
+ */
+public class RandomPlacementFactory implements PlacementPluginFactory {
+
+  @Override
+  public PlacementPlugin createPluginInstance() {
+    return new RandomPlacementPlugin();
+  }
+
+  public static class RandomPlacementPlugin implements PlacementPlugin {
+    private Random random = new Random();
```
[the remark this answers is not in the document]
murblanc: @TomMD is there a way to annotate the code when we know/decide this is not an issue to silence muse-dev? (@madrob suggested to ask you).

murblanc:
Looks good to me!

solr/core/src/java/org/apache/solr/cluster/placement/PlacementPluginFactory.java
sigram: I'll add a reference to `ContainerPluginsApi#Edit` - JSON formats and the location of the config in `/clusterprops.json` are an implementation detail. Also, container plugin configuration is/will be described in the RefGuide (SOLR-15022).

solr/core/src/java/org/apache/solr/cluster/placement/plugins/MinimizeCoresPlacementFactory.java
sigram: I think it needs to return null here, I don't see any other way to return an instance that is assignable without ClassCastException to any `<T extends PlacementPluginConfig>`.

sigram:
Merged.

solr/core/src/java/org/apache/solr/cluster/placement/plugins/RandomPlacementFactory.java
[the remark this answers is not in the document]
TomMD: @murblanc Sorry for the delay, this was missed.

The above issue is produced by FSB which suggests `@suppressFBWarnings`.  See [here](http://findbugs.sourceforge.net/api/edu/umd/cs/findbugs/annotations/SuppressFBWarnings.html) for the method to suppress in a case by case basis.

If you'd like to remove the warning globally then you can add `ignore = [ "PREDICTABLE_RANDOM" ]` to the `.muse/config.toml` file.

PR #2103, opened 2020-11-27
Reconcile upgrade notes in master

janhoy:
The upgrade notes in ref-guide tends to sometimes be updated in 8.x branches only and not synced to master branch.
This is an attempt, (inspired by and fixes #2102). I diffed the `branch_8x` version with master and saw that the 8x wording in most cases was more up to date. I tried to make sure that the updates are valid for 9.x, and skipped a few refguide links that I I believe may have been since removed, but have not tried to compile the refguide to detect its validity.


ctargett (approved):
This is fine. In a couple cases I think the branch_8x wording is strange when it's on master, but the entire contents of the page will be moved and re-edited when we actually release 9.0 so making sure the general ideas are there is fine.

PR #2104, opened 2020-11-28
SOLR Contribs: Assembly was missing READMEs

dsmiley:
_(9.0/Gradle build)_
I found that `README.md`, `CHANGES.md`, `example/` was not getting in the assembly (the distribution).  I chose to take an excludes-list instead of includes-list approach.  This way, there's a problem in the future, it'll be because we accidentally included stuff (no big deal?) instead of accidentally forgetting stuff (more or a problem).  I tested this manually, and looked at the prometheus one especially.


dweiss:
Sorry, I missed this one. I think the includes approach is actually safer - people use all kinds of IDEs, tools and stuff that produces local files which you simply can't predict. If something is omitted in the distribution should be caught by ITs (which, admittedly we don't have in Solr).

PR #2105, opened 2020-11-28
Remove obsolete dev-tools scripts

dsmiley:
These build files appear to be obsolete.  AFAICT, nothing calls them.  Let's delete them.


dweiss (approved):
I don't think these are used anywhere.

PR #2107, opened 2020-11-30
SOLR-15017: The core's lib/ folder content is not loaded in the classloader anymore when the core's configuration does not define any <lib> element

tmortagne:
Fixes https://issues.apache.org/jira/browse/SOLR-15017

### DIFF
--- a/solr/CHANGES.txt
+++ b/solr/CHANGES.txt
@@ -205,6 +205,9 @@ Bug Fixes
 
 * SOLR-14851: Http2SolrClient doesn't handle keystore type correctly (Andras Salamon via janhoy)
 
+* SOLR-15017: Core lib directories were not being recognized unless the solrconfig included a <lib> directive.
+  (Thomas Mortagne)
+
 Other Changes
 ---------------------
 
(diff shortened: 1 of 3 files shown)


dsmiley:
+1 LGTM.  I assume you will test this manually?  (good enough). LMK when done; I'll merge this.

tmortagne:
I was wondering where would be the best place to add unit tests actually, I could not find a SolrConfigTest but maybe you have a idea of a place dedicated to SolrConfig validation already.

dsmiley:
See `org.apache.solr.core.TestCoreContainer#testSharedLib` which looks interesting.  You might do something similar and add to TestConfig, and ignore the built-in core in TestConfig since that one probably is using <lib>.  You could use `solr/configsets/minimal.conf`

tmortagne:
Unit test added following TestConfig#testLib() but with a different config.

dsmiley:
Excellent.  Please add to solr/CHANGES.txt in the 8.8 bug fix section at the end, the following:
```
* SOLR-15017: Core lib directories were not being recognized unless the solrconfig included a <lib> directive.
  (Thomas Mortagne)
```
Meaningwhile I'm running all tests...

tmortagne:
Fix the conflict with solr/CHANGES.txt.

PR #2108, opened 2020-11-30
LUCENE-9626 represent HNSW graph neighbors using primitive arrays

msokolov:
The subject line is the main thrust, but there are a few subsidiary changes that were needed in order to achieve that (see below), and I made a few incidental improvements to HNSW-related classes.

1. Add a primitive int-valued heap, IntHeap, based on the existing PriorityQueue
2. Convert scores to 16-bit precision in order to pack them into an int along with a neighbor ordinal

The result was about a 16% improvement in indexing times, and fewer, smaller GC pauses noted in profiler.


msokolov:
I found some problems with the first implementation posted here, which used a tricky data structure combining parallel arrays with a separate sortable key held in an (int) priority queue. It had bugs, and I realized was wasteful too. This replaces the int queue with a long queue, where the parallel arrays (float score + int node id) are folded together into a single (sortable by score) long. I measured 20% index time improvement (over the baseline, not the previous impl here).

PR #2109, opened 2020-12-01
LUCENE-9627: Small refactor of codec classes

iverase:
Refactor of codec classes to separate reading header/footer from reading content of the file.

In addition the `Lucene50FieldInfosFormat` class is removed as it is not used in the codebase.


lucene/core/src/java/org/apache/lucene/codecs/lucene50/Lucene50CompoundFormat.java
```diff
@@ -80,52 +89,46 @@ public void write(Directory dir, SegmentInfo si, IOContext context) throws IOExc
...
-        long endOffset = data.getFilePointer();
-        
-        long length = endOffset - startOffset;
-        
-        // write entry for file
-        entries.writeString(IndexFileNames.stripSegmentName(file));
-        entries.writeLong(startOffset);
-        entries.writeLong(length);
-      }
+      writeSegmentInfo(entries, data, dir, si);
```
jpountz, line 93: call it `writeCompoundFile`?

jpountz (approved):

PR #2112, opened 2020-12-01
SOLR-14942: Move request registration to ContentStreamHandlerBase

shalinmangar:
This addresses review feedback [1] from David Smiley on Jira. It moves the request registration to the ContentStreamHandlerBase class instead of doing a hack-ish instanceof check inside HttpSolrCall.

[1] - https://issues.apache.org/jira/browse/SOLR-14942?focusedCommentId=17227639&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17227639


dsmiley (approved):
+1
Thanks for the PR!  I reviewed the diff in IntelliJ (easier to view as it ignores the newline changes) and it looks good.  I wasn't expecting this on ContentStreamHandlerBase but I can see that it's specific to update loading, not general content streams, and so it makes sense this way.  CSHB is maybe not an ideal name but whatever.

CaoManhDat:
LGTM! That is a nice idea to move thing here.

PR #2115, opened 2020-12-02
SOLR-14992 Wait for node down before checking for node up

madrob:
https://issues.apache.org/jira/browse/SOLR-14992

### DIFF
--- a/solr/core/src/test/org/apache/solr/cloud/TestPullReplicaErrorHandling.java
+++ b/solr/core/src/test/org/apache/solr/cloud/TestPullReplicaErrorHandling.java
@@ -111,6 +111,7 @@ public class TestPullReplicaErrorHandling extends SolrCloudTestCase {
     collectionName = suggestedCollectionName();
     expectThrows(SolrException.class, () -> getCollectionState(collectionName));
     cluster.getSolrClient().setDefaultCollection(collectionName);
+    cluster.waitForAllNodes(30);
   }
 
   @Override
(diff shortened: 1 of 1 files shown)


solr/core/src/test/org/apache/solr/cloud/TestPullReplicaErrorHandling.java
```diff
@@ -236,8 +237,9 @@ public void testCloseHooksDeletedOnReconnect() throws Exception {
     JettySolrRunner jetty = getJettyForReplica(s.getReplicas(EnumSet.of(Replica.Type.PULL)).get(0));
     SolrCore core = jetty.getCoreContainer().getCores().iterator().next();
 
-    for (int i = 0; i < 5; i++) {
+    for (int i = 0; i < (TEST_NIGHTLY ? 5 : 2); i++) {
       cluster.expireZkSession(jetty);
+      waitForState("Expecting node to be disconnected", collectionName, activeReplicaCount(1, 0, 0));
```
tflobbe, line 242: Wouldn't it be better to actually check for the node znode `ctime`? Or maybe a bump in `/live_nodes`'s `cversion`?

madrob: There is a window where live node has gone away but state is still active because it hasn't updated yet. if we're just waiting for and watching live nodes, then we can see that go away and complete the test before the cluster has quiesced. this is also how we check in testPullReplicaDisconnectsFromZooKeeper, so for consistency this felt better.

There is still a different race here that the replica could go down and come back up before we start waiting for it to be down the first time (we're expecting the overseer to be slow), which I'm sure @markrmiller would be upset with me over, but we can deal with that when he finishes the rest of his speed up branch.

tflobbe: > There is a window where live node has gone away but state is still active because it hasn't updated yet. 
Have you seen that happening? AFAIK, everywhere that we check if a replica is active we look at the state and the live nodes.
> if we're just waiting for and watching live nodes, then we can see that go away and complete the test before the cluster has quiesced.
We would still have the check in line 243, right? My point was:
1) wait to see a change in live nodes
2) wait for active (line 243 as it is now)

Wouldn't that be safe (assuming no other, unrelated node dies just at this point)?

> There is still a different race here that the replica could go down and come back up before we start waiting for it to be down the first time
Right, that's the one I was concerned about. Difficult to happen, but...

madrob: I think you're technically correct, but I'd rather leave that as a future improvement since we do it in several places.

PR #2118, opened 2020-12-04
SOLR-15031: Prevent null being wrapped in a QueryValueSource

vonbox:
# Description

When parsing a sub query in a function query, FunctionQParser#parseValueSource does not check if the produced query object is null. When it is, it just wraps a null in a QueryValueSource object. This is a cause for NPE's in code consuming that object. Parsed queries can be null, for example when the query string only contains stopwords, so we need handle that condition.

# Solution

Added a null check to FunctionQParser#parseValueSource

# Tests

The problem can be reproduced via:
1. Start solr with the techproducts example collection: solr start -e techproducts
2. Add a stopword to SOLR_DIR/example/techproducts/solr/techproducts/conf/stopwords.txt, for example "at"
3. Execute a function query: http://localhost:8983/solr/techproducts/select?fieldquery={!field%20f=features%20v=%27%22at%22%27}&q={!func}%20if($fieldquery,1,0)

After applying this PR the NPE is gone.

# Checklist

### DIFF
--- a/solr/core/src/java/org/apache/solr/search/FunctionQParser.java
+++ b/solr/core/src/java/org/apache/solr/search/FunctionQParser.java
@@ -362,7 +362,7 @@ public class FunctionQParser extends QParser {
       }
       Query subQuery = subParser.getQuery();
       if (subQuery == null) {
-        valueSource = new DoubleConstValueSource(0.0f);
+        valueSource = new ConstValueSource(0.0f);
       } else if (subQuery instanceof FunctionQuery) {
         valueSource = ((FunctionQuery) subQuery).getValueSource();
       } else {
--- a/solr/core/src/java/org/apache/solr/search/FunctionQParser.java
+++ b/solr/core/src/java/org/apache/solr/search/FunctionQParser.java
@@ -361,7 +361,9 @@ public class FunctionQParser extends QParser {
         ((FunctionQParser)subParser).setParseMultipleSources(true);
       }
       Query subQuery = subParser.getQuery();
-      if (subQuery instanceof FunctionQuery) {
+      if (subQuery == null) {
+        valueSource = new DoubleConstValueSource(0.0f);
+      } else if (subQuery instanceof FunctionQuery) {
         valueSource = ((FunctionQuery) subQuery).getValueSource();
       } else {
         valueSource = new QueryValueSource(subQuery, 0.0f);
--- a/lucene/queries/src/java/org/apache/lucene/queries/function/valuesource/QueryValueSource.java
+++ b/lucene/queries/src/java/org/apache/lucene/queries/function/valuesource/QueryValueSource.java
@@ -42,6 +42,10 @@ public class QueryValueSource extends ValueSource {
   final float defVal;
 
   public QueryValueSource(Query q, float defVal) {
+    super();
+    if (q == null) {
+      throw new IllegalArgumentException("query cannot be null");
+    }
     this.q = q;
     this.defVal = defVal;
   }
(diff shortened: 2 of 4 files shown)


solr/core/src/java/org/apache/solr/search/FunctionQParser.java
```diff
@@ -361,7 +361,9 @@ protected ValueSource parseValueSource(int flags) throws SyntaxError {
         ((FunctionQParser)subParser).setParseMultipleSources(true);
       }
       Query subQuery = subParser.getQuery();
-      if (subQuery instanceof FunctionQuery) {
+      if (subQuery == null) {
+        valueSource = new DoubleConstValueSource(0.0f);
```
madrob, line 365: Why a Double?

solr/core/src/java/org/apache/solr/search/FunctionQParser.java
```diff
@@ -361,7 +361,9 @@ protected ValueSource parseValueSource(int flags) throws SyntaxError {
...
+      } else if (subQuery instanceof FunctionQuery) {
         valueSource = ((FunctionQuery) subQuery).getValueSource();
       } else {
         valueSource = new QueryValueSource(subQuery, 0.0f);
```
madrob, line 369: Should we add a test in QueryValueSource constructor to require non-null?

madrob (changes requested):
Thanks for this fix, can you add your reproduction from the JIRA issue as a unit test?

madrob:
Overall the fix is definitely good, and I think it's correct, just a few minor questions about it for completeness. Thank you for opening the PR!

vonbox:
> Thanks for this fix, can you add your reproduction from the JIRA issue as a unit test?

Thanks for looking into it. I surely agree that every fix should have a unit test. However, though I know how to get around in the Solr code, the test suite is unknown territory. Can you give me a hint where I can add such test (and maybe find some inspiration from existing tests)?

solr/core/src/java/org/apache/solr/search/FunctionQParser.java
vonbox: Having a second look I see that a ConstValueSource is a better alternative, I'll update this.

solr/core/src/java/org/apache/solr/search/FunctionQParser.java
vonbox: I think it would be good to catch problems at the gate, adding this extra check. However, this could result in a risk that these exceptions pop up in places that were (seemed) executing fine. Very small risk, if a QueryValueSource with a null query is created a NPE is inevitable, so I cannot imagine adding such check will cause new problems. Still, I did not dare to add it yet. Can we get a veto of someone on this? (In the meantime I'll try it and run the test suite.)

vonbox: I pushed a commit adding the null check to the QueryValueSource constructor

madrob:
> Can you give me a hint where I can add such test (and maybe find some inspiration from existing tests)?

You can add a test to https://github.com/apache/lucene-solr/blob/master/solr/core/src/test/org/apache/solr/search/function/TestFunctionQuery.java the `text` field in the schema for that test is already configured to use stopwords: `stopworda` and `stopwordb`

vonbox:
Thanks for the hint. I added a test.

madrob:
Thanks for adding the test! it passes for me when run alone, but not as part of the full class, can you verify in your environment as well?

`./gradlew :solr:core:test --tests TestFunctionQuery`

madrob:
Also, would you mind adding an entry to `solr/CHANGES.txt` with how you would like proper credit/attribution? I'll make sure that it gets to the right section, so don't stress about that if you are unsure.

vonbox:
> Thanks for adding the test! it passes for me when run alone, but not as part of the full class, can you verify in your environment as well?
> 
> `./gradlew :solr:core:test --tests TestFunctionQuery`

That's strange, sometimes it succeeds, but often it fails. It looks like documents from previous tests pop up in the result, which makes me wonder what the clearIndex call actually does? I took the easy way out by removing the numFound check. So now it is just making sure no exception is thrown when executing the function, which is a good enough test I think.

PR #2121, opened 2020-12-05
SOLR-10860: Return proper error code for bad input incase of inplace updates

munendrasn:
Additionally,
* Return proper error when inc operation is specified for non-numeric fields

### DIFF
--- a/solr/core/src/java/org/apache/solr/update/processor/AtomicUpdateDocumentMerger.java
+++ b/solr/core/src/java/org/apache/solr/update/processor/AtomicUpdateDocumentMerger.java
@@ -130,7 +130,7 @@ public class AtomicUpdateDocumentMerger {
               break;
             default:
               throw new SolrException(ErrorCode.BAD_REQUEST,
-                  "Unknown operation for the an atomic update, operation ignored: " + key + " for " + getID(toDoc, schema));
+                  "Error:" + getID(toDoc, schema) + " Unknown operation for the an atomic update: " + key);
           }
           // validate that the field being modified is not the id field.
           if (idField.getName().equals(sif.getName())) {
@@ -150,7 +150,7 @@ public class AtomicUpdateDocumentMerger {
   private static String getID(SolrInputDocument doc, IndexSchema schema) {
     String id = "";
     SchemaField sf = schema.getUniqueKeyField();
-    if( sf != null ) {
+    if (sf != null) {
       id = "[doc="+doc.getFieldValue( sf.getName() )+"] ";
     }
     return id;
--- a/solr/core/src/java/org/apache/solr/update/processor/AtomicUpdateDocumentMerger.java
+++ b/solr/core/src/java/org/apache/solr/update/processor/AtomicUpdateDocumentMerger.java
@@ -483,6 +483,11 @@ public class AtomicUpdateDocumentMerger {
   protected void doInc(SolrInputDocument toDoc, SolrInputField sif, Object fieldVal) {
     SolrInputField numericField = toDoc.get(sif.getName());
     SchemaField sf = schema.getField(sif.getName());
+
+    if (sf.getType().getNumberType() == null) {
+      throw new SolrException(ErrorCode.BAD_REQUEST, "'inc' is not supported on non-numeric field " + sf.getName());
+    }
+
     if (numericField != null || sf.getDefaultValue() != null) {
       // TODO: fieldtype needs externalToObject?
       String oldValS = (numericField != null) ?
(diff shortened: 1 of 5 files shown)


mkhludnev (approved):

solr/core/src/java/org/apache/solr/update/processor/AtomicUpdateDocumentMerger.java
```diff
@@ -143,6 +147,15 @@ public SolrInputDocument merge(final SolrInputDocument fromDoc, SolrInputDocumen
     return toDoc;
   }
 
+  private static String getID(SolrInputDocument doc, IndexSchema schema) {
+    String id = "";
```
madrob, line 151: can we default to `(unknown id)` otherwise the error message will look weird I think.

solr/core/src/java/org/apache/solr/update/processor/AtomicUpdateDocumentMerger.java
```diff
@@ -553,7 +574,15 @@ private Object getNativeFieldValue(String fieldName, Object val) {
       return val;
     }
     SchemaField sf = schema.getField(fieldName);
-    return sf.getType().toNativeType(val);
+    try {
+      return sf.getType().toNativeType(val);
+    } catch (SolrException ex) {
+      throw new SolrException(SolrException.ErrorCode.getErrorCode(ex.code()),
+          "Error converting field '" + sf.getName() + "'='" +val+"' to native type, msg=" + ex.getMessage(), ex);
```
madrob, line 581: I don't think we want `msg` copied since it will be in the cause anyway.

solr/core/src/test/org/apache/solr/update/TestInPlaceUpdatesStandalone.java
```diff
@@ -121,6 +123,36 @@ public void deleteAllAndCommit() throws Exception {
...
+    MatcherAssert.assertThat(e.getMessage(), containsString("For input string: \"NOT_NUMBER\""));
+
+    // inc op with null value
+    e = expectThrows(SolrException.class,
+        () -> addAndAssertVersion(version1, "id", "1", "inplace_updatable_float", map("inc", null)));
+    assertEquals(SolrException.ErrorCode.BAD_REQUEST.code, e.code());
+    MatcherAssert.assertThat(e.getMessage(), containsString("Invalid input 'null' for field inplace_updatable_float"));
+
+    e = expectThrows(SolrException.class,
+        () -> addAndAssertVersion(version1, "id", "1", "inplace_updatable_float",
```
madrob, line 150: This surprises me a little bit that we can't increment a float by an integer amount?

munendrasn: We can increment float by an integer. In this particular test input, verifying the case when instead of passing the number, a list of numbers is passed. Previously, Solr used to return 500 with the current changes Bad request would be returned
`"Invalid input '[123]' for field inplace_updatable_float"`

solr/core/src/java/org/apache/solr/update/processor/AtomicUpdateDocumentMerger.java
munendrasn: I'm thinking to rephrase the above error message to something like so that it is better than the previous msg. If the id is not known then, I think maybe it is better not to send anything related id, wdyt?
```
"Error:" + getID(toDoc, schema) + "Unknown operation for the an atomic update : " + key;
```

solr/core/src/java/org/apache/solr/update/processor/AtomicUpdateDocumentMerger.java
munendrasn: cause gets lost in the metadata section of the response so, thought this would give simpler insight into error. Also, trying to follow the same convention as other error messages in DocumentBuilder

solr/core/src/java/org/apache/solr/update/processor/AtomicUpdateDocumentMerger.java
```diff
@@ -143,6 +147,15 @@ public SolrInputDocument merge(final SolrInputDocument fromDoc, SolrInputDocumen
     return toDoc;
   }
 
+  private static String getID(SolrInputDocument doc, IndexSchema schema) {
+    String id = "";
+    SchemaField sf = schema.getUniqueKeyField();
+    if( sf != null ) {
```
jbampton, line 153:
```suggestion
    if ( sf != null ) {
```

madrob:
@munendrasn all of your comments make sense, thanks. let's update the error message as you describe and then this looks good.

madrob (approved):

munendrasn:
@mkhludnev @madrob 
Thanks for the review 🙂

PR #2122, opened 2020-12-05
SOLR-14950: Fix copyfield regeneration with explicit src/dest matching dyn rule

munendrasn:
CopyFields are regenerated in case of replace-field or replace-field-type.
While regenerating, source and destination are checked against fields but source/dest
could match dynamic rule too.
For example,
<copyField source="something_s" dest="spellcheck"/>
<dynamicField name="*_s" type="string"/>
here, something_s is not present in schema but matches the dynamic rule.

To handle the above case, need to check dynamicFieldCache too while regenerating the
copyFields

### DIFF
--- a/solr/core/src/test/org/apache/solr/rest/schema/TestBulkSchemaAPI.java
+++ b/solr/core/src/test/org/apache/solr/rest/schema/TestBulkSchemaAPI.java
@@ -773,6 +773,108 @@ public class TestBulkSchemaAPI extends RestTestBase {
     assertTrue("'bleh_s' copyField rule exists in the schema", l.isEmpty());
   }
 
+  @SuppressWarnings({"rawtypes"})
+  public void testCopyFieldWithReplace() throws Exception {
+    RestTestHarness harness = restTestHarness;
+    String newFieldName = "test_solr_14950";
+
+    // add-field-type
+    String addFieldTypeAnalyzer = "{\n" +
+        "'add-field-type' : {" +
+        "    'name' : 'myNewTextField',\n" +
+        "    'class':'solr.TextField',\n" +
+        "    'analyzer' : {\n" +
+        "        'charFilters' : [{\n" +
+        "                'name':'patternReplace',\n" +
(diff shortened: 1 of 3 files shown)


solr/core/src/test/org/apache/solr/rest/schema/TestBulkSchemaAPI.java
```diff
@@ -773,6 +773,108 @@ public void testCopyFieldRules() throws Exception {
...
+    l = getSourceCopyFields(harness, "bleh_s");
+    assertFalse("'bleh_s' copyField rule doesn't exist", l.isEmpty());
+    assertEquals("bleh_s", ((Map)l.get(0)).get("source"));
+    assertEquals(newFieldName, ((Map)l.get(0)).get("dest"));
+
+    // replace-field-type
+    String replaceFieldTypeAnalyzer = "{\n" +
+        "'replace-field-type' : {" +
+        "    'name' : 'myNewTextField',\n" +
+        "    'class':'solr.TextField',\n" +
```
jbampton, line 845:
```suggestion
        "    'class' : 'solr.TextField',\n" +
```

solr/core/src/test/org/apache/solr/rest/schema/TestBulkSchemaAPI.java
```diff
@@ -773,6 +773,108 @@ public void testCopyFieldRules() throws Exception {
...
+  @SuppressWarnings({"rawtypes"})
+  public void testCopyFieldWithReplace() throws Exception {
+    RestTestHarness harness = restTestHarness;
+    String newFieldName = "test_solr_14950";
+
+    // add-field-type
+    String addFieldTypeAnalyzer = "{\n" +
+        "'add-field-type' : {" +
+        "    'name' : 'myNewTextField',\n" +
+        "    'class':'solr.TextField',\n" +
```
jbampton, line 785:
```suggestion
        "    'class' : 'solr.TextField',\n" +
```

munendrasn:
This is merged

PR #2123, opened 2020-12-07
SOLR-10732: short circuit calls to searcher#numDocs when base is empty

munendrasn:
This avoids constructing redundant queries

### DIFF
--- a/solr/core/src/java/org/apache/solr/request/SimpleFacets.java
+++ b/solr/core/src/java/org/apache/solr/request/SimpleFacets.java
@@ -287,7 +287,7 @@ public class SimpleFacets {
     NamedList<Integer> res = new SimpleOrderedMap<>();
 
     /* Ignore CommonParams.DF - could have init param facet.query assuming
-     * the schema default with query param DF intented to only affect Q.
+     * the schema default with query param DF intended to only affect Q.
      * If user doesn't want schema default for facet.query, they should be
      * explicit.
      */
@@ -306,6 +306,10 @@ public class SimpleFacets {
   }
 
   public void getFacetQueryCount(ParsedParams parsed, NamedList<Integer> res) throws SyntaxError, IOException {
+    if (parsed.docs.size() == 0) {
+      res.add(parsed.key, 0);
+      return;
+    }
     // TODO: slight optimization would prevent double-parsing of any localParams
     // TODO: SOLR-7753
     Query qobj = QParser.getParser(parsed.facetValue, req).getQuery();
@@ -325,6 +329,9 @@ public class SimpleFacets {
    * @see FacetParams#FACET_QUERY
    */
   public int getGroupedFacetQueryCount(Query facetQuery, DocSet docSet) throws IOException {
+    if (docSet.size() == 0) {
+      return 0;
+    }
     // It is okay to retrieve group.field from global because it is never a local param
     String groupField = global.get(GroupParams.GROUP_FIELD);
     if (groupField == null) {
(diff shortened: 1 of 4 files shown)


solr/core/src/java/org/apache/solr/search/facet/FacetProcessor.java
```diff
@@ -419,7 +419,7 @@ void fillBucket(SimpleOrderedMap<Object> bucket, Query q, DocSet result, boolean
       }
       count = result.size();  // don't really need this if we are skipping, but it's free.
     } else {
-      if (q == null) {
+      if (q == null || fcontext.base.size() == 0) {
```
magibney, line 422: The query is already built at this point, so I don't think this particular change actually helps wrt SOLR-10732? (and the `base.size()==0` case is already trivially optimized in `SolrIndexSearcher.numDocs(Query, DocSet)`)

munendrasn: Will remove this before committing

solr/core/src/java/org/apache/solr/request/SimpleFacets.java
```diff
@@ -903,7 +910,7 @@ public void execute(Runnable r) {
 
   private int numDocs(String term, final SchemaField sf, final FieldType ft, final DocSet baseDocset) {
     try {
-      return searcher.numDocs(ft.getFieldQuery(null, sf, term), baseDocset);
+      return baseDocset.size() == 0? 0: searcher.numDocs(ft.getFieldQuery(null, sf, term), baseDocset);
```
magibney, line 913: Line 903 (immediately above) is the only caller of this private method. Would it make sense to do the check in only one of these places? Maybe prefer the higher-level (line 903) check, where the check could also be used to bypass `termCountEntries.sorted(...)`? Could optionally leave `assert baseDocset.size() > 0` here.

solr/core/src/java/org/apache/solr/request/SimpleFacets.java
```diff
@@ -325,6 +329,9 @@ public void getFacetQueryCount(ParsedParams parsed, NamedList<Integer> res) thro
    * @see FacetParams#FACET_QUERY
    */
   public int getGroupedFacetQueryCount(Query facetQuery, DocSet docSet) throws IOException {
+    if (docSet.size() == 0) {
+      return 0;
+    }
```
magibney, line 334: Both callers of this method (`SimpleFacets.getFacetQueryCount(...)` and `RangeFacetProcessor.rangeCount(...)` check for the base `docSet.size()==0` case independently. Could perhaps remove this redundant check (or replace with `assert docSet.size() > 0`)?

dsmiley:
Thanks Munendra, I'm glad you moved this from a patch file to GitHub PR -- way more conducive to getting a (good) review.  I don't bother with patch files any more because I want reviewers, and patch files are a barrier.  I think you already incorporated my suggestion in JIRA; it's good to see Michael's input.

solr/core/src/java/org/apache/solr/search/facet/FacetProcessor.java
munendrasn: This is done https://github.com/apache/lucene-solr/pull/2123/commits/c194e09ca0d2df32acf21875c7625f9e862fdc09

solr/core/src/java/org/apache/solr/request/SimpleFacets.java
munendrasn: sort by count won't be done if baseDocSet size is 0 but I have kept this check in numDocs so that, any future usage can benefit from it

solr/core/src/java/org/apache/solr/request/SimpleFacets.java
munendrasn: Same as above

PR #2129, opened 2020-12-07
Fix format indent from 4 to 2 spaces

mayya-sharipova:
Files in org.apache.lucene.search.comparators package
has a wrong indent of 4 spaces instead of 2.
This patch fixes only the indent from 4 to correct 2 spaces

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/comparators/DocComparator.java
+++ b/lucene/core/src/java/org/apache/lucene/search/comparators/DocComparator.java
@@ -29,157 +29,159 @@ import java.io.IOException;
  * Comparator that sorts by asc _doc
  */
 public class DocComparator extends FieldComparator<Integer> {
-    private final int[] docIDs;
-    private final boolean enableSkipping; // if skipping functionality should be enabled
-    private int bottom;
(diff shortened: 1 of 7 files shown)


msokolov:
Thanks for fixing - I have another one I need to do, too, and I wonder if we shouldn't just push uncontroversial changes like this directly? No need to wait for review I think.

mayya-sharipova:
@msokolov  Thanks for your comment. Indeed having a  `gradlew precommit` to fail on inconsistent code style would be useful.

Thanks for feedback, I will merge the PR.

ErickErickson:
It already fails on _tabs_ rather than spaces, but failing on too many spaces isn’t checked.

That said, rather than a one-off for indentation, I’d rather see the effort go here
rather than a separate precommit check….:

https://issues.apache.org/jira/browse/LUCENE-9564 and SOLR-14920…

BTW, 'gradlew check’ does all the precommit tasks as well as run tests...

mayya-sharipova:
@ErickErickson Thanks for your suggestions, and letting me know about LUCENE-9564, I will keep watching this issue.

PR #2130, opened 2020-12-08
Adding Apache Reporter step in Release Wizard.

HoustonPutman:
The release wizard doesn't currently tell the RM to add the new release to the apache release reporter.

This is a simple change to add that one step at the end of announcing the new version.


anshumg (approved):
LGTM. Looks like this is accessible to all committers and isn't a PMC only link so we should be good to go here.

janhoy (approved):

HoustonPutman:
Thanks for checking on that Anshum

PR #2133, opened 2020-12-09
SOLR-15019: Replica placement API needs a way to fetch existing replica metrics

sigram:
See JIRA for details.


solr/core/src/java/org/apache/solr/cluster/placement/ShardMetrics.java
```diff
@@ -0,0 +1,27 @@
...
+package org.apache.solr.cluster.placement;
+
+import java.util.Optional;
+
+/**
+ *
+ */
+public interface ShardMetrics {
+  Optional<ReplicaMetrics> getLeaderMetrics();
+  Optional<ReplicaMetrics> getReplicaMetrics(String replicaName);
```
sigram, line 26: Perhaps we should add `iterator()` here too, so that the consumers are not required to know the replica name in advance.

solr/core/src/java/org/apache/solr/cluster/placement/impl/CollectionMetricsBuilder.java
```diff
@@ -0,0 +1,117 @@
...
+      replicaMetricsBuilders.put(replicaName, replicaMetricsBuilder);
+      return this;
+    }
+
+    public ShardMetricsBuilder setLeaderMetrics(ReplicaMetricsBuilder replicaMetricsBuilder) {
+      replicaMetricsBuilders.put(LEADER, replicaMetricsBuilder);
+      return this;
+    }
+
+    public static final String LEADER = "__leader__";
```
sigram, line 58: Make this `@VisibleForTesting` or package-private.

solr/core/src/java/org/apache/solr/cluster/placement/impl/CollectionMetricsBuilder.java
```diff
@@ -0,0 +1,117 @@
...
+        public Optional<ReplicaMetrics> getReplicaMetrics(String replicaName) {
+          return Optional.ofNullable(metricsMap.get(replicaName));
+        }
+      };
+    }
+  }
+
+  public static class ReplicaMetricsBuilder {
+    final Map<String, Object> metrics = new HashMap<>();
+    int sizeGB = 0;
```
sigram, line 85: Maybe use `Integer` here? Replicas always have size, even new ones, but if we can return null then we can signal that the size is unknown (eg. couldn't be retrieved from the node).

solr/core/src/java/org/apache/solr/cluster/placement/AttributeFetcher.java
```diff
@@ -78,13 +85,15 @@
   AttributeFetcher fetchFrom(Set<Node> nodes);
 
   /**
-   * Requests a (non node) metric of a given scope and name. To get the value use {@link AttributeValues#getMetric(String, String)}
+   * Requests any metric from any metric registry on each node, using a fully-qualified metric key,
+   * for example <code>solr.jvm:system.properties:user.name</code>.
+   * To get the value use {@link AttributeValues#getNodeMetric(Node, String)}
    */
-  AttributeFetcher requestMetric(String scope, String metricName);
+  AttributeFetcher requestNodeMetric(String metricKey);
```
murblanc, line 92: Move this method closer to the other `requestNodeMetric` above.

solr/core/src/java/org/apache/solr/cluster/placement/AttributeValues.java
```diff
@@ -65,11 +65,16 @@
...
 
   /**
-   * Get a non node related metric of specific scope and name
+   * Get any metric using a fully-qualified metric key.
    */
-  Optional<Double> getMetric(String scope, String metricName);
+  Optional<Object> getNodeMetric(Node node, String metricKey);
+
+  /**
+   *
```
murblanc, line 77: Javadoc :-)

solr/core/src/java/org/apache/solr/cluster/placement/AttributeValues.java
```diff
@@ -65,11 +65,16 @@
...
+   */
+  Optional<CollectionMetrics> getCollectionMetrics(String collectionName);
```
murblanc, line 79: We have a lack of symmetry now between node metrics that return mere `Object`s and collection/shard/replica metrics that are defined types (eventually resolving to a `ReplicaMetrics` that can return an object or specific typed metrics - currently only replica size GB).
Maybe we should call something different the "typed" metric interfaces to distinguish between them and the object being eventually returned?
I suggest that rather than calling the interface `CollectionMetrics`, we call it `CollectionMetricContainer` (and similarly for Shard and Replica). That would help me.

solr/core/src/java/org/apache/solr/cluster/placement/CollectionMetrics.java
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+package org.apache.solr.cluster.placement;
+
+import java.util.Optional;
+
+/**
+ *
+ */
+public interface CollectionMetrics {
```
murblanc, line 25: As suggested elsewhere, I'd rather have this interface named `CollectionMetricsContainer`.

solr/core/src/java/org/apache/solr/cluster/placement/ReplicaMetric.java
```diff
@@ -0,0 +1,113 @@
...
+
+import java.util.Objects;
+import java.util.function.Function;
+
+/**
+ * Replica metric wrapper that defines a short symbolic name of the metric, the corresponding
+ * internal metric name (as reported in <code>solr.core.[collection].[replica]</code> registry)
+ * and the desired format/unit conversion.
+ */
+public class ReplicaMetric<T> {
```
murblanc, line 27: Now I'm confused between `ReplicaMetric` and `ReplicaMetrics`.... So take previous comments with a grain of salt I might have gotten things wrong.

solr/core/src/java/org/apache/solr/cluster/placement/AttributeFetcher.java
```diff
@@ -70,6 +71,12 @@
    */
   AttributeFetcher requestNodeMetric(String metricName, NodeMetricRegistry registry);
 
+  /**
+   * Request collection-level metrics. To get the values use {@link AttributeValues#getCollectionMetrics(String)}.
+   * Note that this request will fetch information from nodes relevant to the collection
+   * replicas and not the ones specified in {@link #fetchFrom(Set)} (though they may overlap).
+   */
+  AttributeFetcher requestCollectionMetrics(SolrCollection solrCollection, Set<ReplicaMetric<?>> metricNames);
```
murblanc, line 79: `metricNames` is not a good name, there are not just names.

solr/core/src/java/org/apache/solr/cluster/placement/ReplicaMetric.java
```diff
@@ -0,0 +1,113 @@
...
+
+  private static final double GB = 1024 * 1024 * 1024;
+  @SuppressWarnings("unchecked")
+  private final Function<Object, T> NO_CONVERTER = v -> {
```
murblanc, line 31: I'd rather have this named `IDENTITY_CONVERTER`

solr/core/src/java/org/apache/solr/cluster/placement/AttributeFetcher.java
sigram: Initially this was a String but then I realized we need more metadata, I'll fix it.

solr/core/src/java/org/apache/solr/cluster/placement/AttributeValues.java
sigram: Hmm... It's clear that this type is not a primitive. Similarly, we're not calling the container for attribute values an `AttributeValuesContainer`. I don't think this change here would actually help in understanding this relationship.

solr/core/src/java/org/apache/solr/cluster/placement/ReplicaMetric.java
sigram: `ReplicaMetric` (singular) is a pseudo-enum that encapsulates both the metric name that we want to use (usually something simple), the internal Solr metrics name (which is usually longer and more complicated) and the optional unit/type conversion for the value retrieved from the metrics API.

solr/core/src/java/org/apache/solr/cluster/placement/CollectionMetrics.java
murblanc: Ok for not renaming, but maybe say something in the Javadoc about this? Are there collection level metrics that are not shard metrics?

solr/core/src/java/org/apache/solr/cluster/placement/ReplicaMetrics.java
```diff
@@ -0,0 +1,29 @@
...
+
+import java.util.Optional;
+
+/**
+ *
+ */
+public interface ReplicaMetrics {
+
+  Double getReplicaSizeGB();
+  Optional<Object> getReplicaMetric(String metricName);
```
murblanc, line 28: Can we change this signature into:
`<T> Optional<T> getReplicaMetric(ReplicaMetric<T> replicaMetric);`
to get rid of the type check and casts of the returned value and the `getName()` on the `ReplicaMetric` parameter?

solr/core/src/java/org/apache/solr/cluster/placement/AttributeValues.java
```diff
@@ -65,11 +65,16 @@
   /**
    * For the given node: metric of specific name and registry
    */
-  Optional<Double> getMetric(Node node, String metricName, AttributeFetcher.NodeMetricRegistry registry);
+  Optional<Object> getMetric(Node node, String metricName, AttributeFetcher.NodeMetricRegistry registry);
```
murblanc, line 68: If `ReplicaMetrics` can be improved as suggested by a [comment](https://github.com/apache/lucene-solr/pull/2133/files#r546733523) there to make better use of `ReplicaMetric<T>` instances and not use generic `Object` values, can we generalize the concept?
Either rename `ReplicaMetric` and have it define metrics for all types of cluster entities (i.e. including nodes) or introduce a similar `NodeMetric<T>` class that would be used here as well as in `AttributeFetcher`?

murblanc: Also, method likely needs to have `Node` in the name given it's a node metric (like method below).

solr/core/src/java/org/apache/solr/cluster/placement/ReplicaMetrics.java
sigram: This will require casts in implementations anyway, but I agree the external API is cleaner this way.

solr/core/src/java/org/apache/solr/cluster/placement/ShardMetrics.java
murblanc: Yes. As well as `ShardMetrics` iterator in `CollectionMetrics`.

solr/core/src/java/org/apache/solr/cluster/placement/impl/CollectionMetricsBuilder.java
```diff
@@ -0,0 +1,116 @@
...
+    public Map<String, ReplicaMetricsBuilder> getReplicaMetricsBuilders() {
+      return replicaMetricsBuilders;
+    }
+
+    public ShardMetricsBuilder setLeaderMetrics(ReplicaMetricsBuilder replicaMetricsBuilder) {
+      replicaMetricsBuilders.put(LEADER, replicaMetricsBuilder);
+      return this;
+    }
+
+    static final String LEADER = "__leader__";
```
murblanc, line 57: I'd rather have a dedicated variable in the `ShardMetricsBuilder` referencing the leader's `ReplicaMetricsBuilder` instance and a dedicated variable in the anonymous subclass of `ShardMetrics` doing the same.
Will be cleaner to read and not less efficient.

solr/core/src/test/org/apache/solr/cluster/placement/Builders.java
```diff
@@ -48,7 +49,11 @@ public static CollectionBuilder newCollectionBuilder(String collectionName) {
     public ClusterBuilder initializeLiveNodes(int countNodes) {
       nodeBuilders = new LinkedList<>();
       for (int n = 0; n < countNodes; n++) {
-        nodeBuilders.add(new NodeBuilder().setNodeName("node_" + n)); // Default name, can be changed
+        NodeBuilder nodeBuilder = new NodeBuilder().setNodeName("node_" + n); // Default name, can be changed
+        nodeBuilder.setTotalDiskGB(10000.0);
+        nodeBuilder.setFreeDiskGB(5000.0);
+        nodeBuilder.setDiskType(AttributeFetcher.DiskHardwareType.SSD);
```
murblanc, line 55: I thought disk type was going away?

solr/core/src/test/org/apache/solr/cluster/placement/impl/PlacementPluginIntegrationTest.java
```diff
@@ -220,6 +231,66 @@ public void testDynamicReconfiguration() throws Exception {
...
+      assertTrue("shard metrics", shardMetricsOpt.isPresent());
+      shard.replicas().forEach(replica -> {
+        Optional<ReplicaMetrics> replicaMetricsOpt = shardMetricsOpt.get().getReplicaMetrics(replica.getReplicaName());
+        assertTrue("replica metrics", replicaMetricsOpt.isPresent());
+        ReplicaMetrics replicaMetrics = replicaMetricsOpt.get();
+        // this should always be present
+        assertNotNull("size", replicaMetrics.getReplicaSizeGB());
+        assertTrue("should be greater than 0 but was " + replicaMetrics.getReplicaSizeGB(),
+            replicaMetrics.getReplicaSizeGB() > 0);
+        assertNotNull("queryRate", replicaMetrics.getReplicaMetric(ReplicaMetric.QUERY_RATE_1MIN.getName()));
```
murblanc, line 288: We need to retrieve the value of a replica metric and verify it (previously it would have been making sure it is cast to the appropriate type, but with the ongoing changes, show how these values are read and make sure things actually works).

murblanc: This will possibly happen earlier in the test code if index size is converted into a normal metric though.

solr/core/src/java/org/apache/solr/cluster/placement/AttributeFetcher.java
```diff
@@ -18,93 +18,53 @@
...
 
-  /**
-   * Requests a (non node) metric of a given scope and name. To get the value use {@link AttributeValues#getMetric(String, String)}
-   */
-  AttributeFetcher requestMetric(String scope, String metricName);
-
   /**
    * Fetches all requested node attributes from all nodes passed to {@link #fetchFrom(Set)} as well as non node attributes
-   * (those requested for example using {@link #requestMetric(String, String)}.
+   * (those requested for example using {@link #requestNodeMetric(NodeMetric)}.
```
murblanc, line 65: Didn't you mean to reference `requestCollectionMetrics` here?

solr/core/src/java/org/apache/solr/cluster/placement/AttributeValues.java
```diff
@@ -65,11 +35,10 @@
   /**
    * For the given node: metric of specific name and registry
```
murblanc, line 36: Javadoc comment needs updating

solr/core/src/java/org/apache/solr/cluster/placement/NodeMetric.java
```diff
@@ -0,0 +1,51 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.solr.cluster.placement;
+
+/**
+ * Node metric identifier, corresponding
+ * to a node-level metric registry and the internal metric name.
+ * Alternatively this identified may use a fully-qualified metric key,
```
murblanc, line 23: typo: identified -> identifier

Also, the second part of the Javadoc comment seems more appropriate for the implementation rather than the interface.
Can we add to the `Registry` enum a `NULL` or `UNSPECIFIED` value so that the implementation registry is never `null`?

solr/core/src/java/org/apache/solr/cluster/placement/MetricAttribute.java
```diff
@@ -0,0 +1,53 @@
...
+package org.apache.solr.cluster.placement;
+
+import java.util.function.Function;
+
+/**
+ * Metric-related attribute of a node or replica. It defines a short symbolic name of the metric, the corresponding
+ * internal metric name and the desired format/unit conversion. Generic type
+ * defines the type of converted values of this attribute.
+ */
+public interface MetricAttribute<T> {
```
murblanc, line 26: I find it a bit weird (naming wise) that interface `MetricAttribute` has subinterfaces `NodeMetric` and `ReplicaMetric`.
Ideally I'd drop the `Attribute` part of the interface name here (and in `MetricAttributeImpl`).

solr/core/src/java/org/apache/solr/cluster/placement/impl/MetricAttributeImpl.java
```diff
@@ -0,0 +1,138 @@
...
+
+import org.apache.solr.cluster.placement.MetricAttribute;
+
+import java.util.Objects;
+import java.util.function.Function;
+
+/**
+ * Base class for {@link MetricAttribute} implementations.
+ */
+public class MetricAttributeImpl<T> implements MetricAttribute<T> {
```
murblanc, line 27: Make class `abstract`.

solr/core/src/java/org/apache/solr/cluster/placement/MetricAttribute.java
```diff
@@ -0,0 +1,53 @@
...
+
+  /**
+   * Return the internal name of a Solr metric associated with this attribute.
+   */
+  String getInternalName();
+
+  /**
+   * Conversion function to convert formats/units of raw values.
+   */
+  Function<Object, T> getConverter();
```
murblanc, line 41: I don't see the value in exposing the converter function in the interface. `convert` method below is sufficient (and its default implementation moved to the abstract implementation `MetricAttributeImpl` (or `MetricImpl` as I'd prefer it to be named)).

solr/core/src/java/org/apache/solr/cluster/placement/ReplicaMetrics.java
```diff
@@ -0,0 +1,28 @@
...
+package org.apache.solr.cluster.placement;
+
+import java.util.Optional;
+
+/**
+ * Strongly-typed replica-level metrics.
+ */
+public interface ReplicaMetrics {
+
+  <T> Optional<T> getReplicaMetric(ReplicaMetric<T> metric);
```
murblanc, line 27: Shouldn't this method be called `getReplicaMetricValue`?

solr/core/src/java/org/apache/solr/cluster/placement/impl/AttributeFetcherImpl.java
```diff
@@ -33,18 +37,18 @@
...
 
-  boolean requestedNodeCoreCount;
-  boolean requestedNodeDiskType;
-  boolean requestedNodeFreeDisk;
-  boolean requestedNodeTotalDisk;
-  boolean requestedNodeHeapUsage;
-  boolean requestedNodeSystemLoadAverage;
-  Set<String> requestedNodeSystemPropertiesSnitchTags = new HashSet<>();
-  Set<String> requestedNodeMetricSnitchTags = new HashSet<>();
+  Set<String> requestedNodeSystemSnitchTags = new HashSet<>();
```
murblanc, line 49: I preferred the previous name `requestedNodeSystemPropertiesSnitchTags` ("system" by itself is too vague).

murblanc: Oh I see it's used for both system properties and environment variables... In which case I'd drop system in the name.

solr/core/src/java/org/apache/solr/cluster/placement/impl/AttributeFetcherImpl.java
```diff
@@ -113,72 +90,46 @@ public AttributeFetcher fetchFrom(Set<Node> nodes) {
...
 
-  @Override
-  public AttributeFetcher requestMetric(String scope, String metricName) {
-    throw new UnsupportedOperationException("Not yet implemented...");
-  }
+  private static final double GB = 1024 * 1024 * 1024;
 
   @Override
   public AttributeValues fetchAttributes() {
     // TODO Code here only supports node related attributes for now
```
murblanc, line 97: remove comment

solr/core/src/java/org/apache/solr/cluster/placement/impl/AttributeFetcherImpl.java
```diff
@@ -113,72 +90,46 @@ public AttributeFetcher fetchFrom(Set<Node> nodes) {
...
+      collection.shards().forEach(shard ->
+          shard.replicas().forEach(replica -> {
+            Set<String> perNodeInternalTags = nodeToReplicaInternalTags
+                .computeIfAbsent(replica.getNode(), n -> new HashSet<>());
+            perNodeInternalTags.addAll(collectionTags);
+          }));
+    });
 
     // Now that we know everything we need to fetch (and where to put it), just do it.
     for (Node node : nodes) {
```
murblanc, line 135: Maybe add a comment here saying fetching could be made concurrent from all nodes at once for perf improvement?

solr/core/src/java/org/apache/solr/cluster/placement/impl/AttributeFetcherImpl.java
```diff
@@ -197,32 +148,78 @@ public AttributeValues fetchAttributes() {
...
 
-    return new AttributeValuesImpl(nodeToCoreCount,
-        nodeToDiskType,
-        nodeToFreeDisk,
-        nodeToTotalDisk,
-        nodeToHeapUsage,
-        nodeToSystemLoadAverage,
-        syspropSnitchToNodeToValue,
-        metricSnitchToNodeToValue);
+    for (Node node : nodeToReplicaInternalTags.keySet()) {
```
murblanc, line 151: I understand here we might be contacting nodes again to fetch replica metrics. I think this is ok but maybe add a comment that there can be an optimization where node metrics and replica metrics are fetched with a single call to each node?
Here we might even contact each node multiple times I believe depending on the collections we need info about and how their replicas are distributed?

solr/core/src/java/org/apache/solr/cluster/placement/impl/AttributeFetcherImpl.java
```diff
@@ -197,32 +148,78 @@ public AttributeValues fetchAttributes() {
...
+                CollectionMetricsBuilder.ReplicaMetricsBuilder replicaMetricsBuilder =
+                    shardMetricsBuilder.getReplicaMetricsBuilders()
+                    .computeIfAbsent(replica.getName(), n -> new CollectionMetricsBuilder.ReplicaMetricsBuilder());
+                replicaMetricsBuilder.setLeader(replica.isLeader());
+                if (replica.isLeader()) {
+                  shardMetricsBuilder.setLeaderMetrics(replicaMetricsBuilder);
+                }
+                Set<ReplicaMetric<?>> requestedMetrics = requestedCollectionNamesMetrics.get(replica.getCollection());
+                if (requestedMetrics == null) {
+                  throw new RuntimeException("impossible error");
```
murblanc, line 173: Likely an error log with the value of `replica.getCollection()` (or add the value to the exception text) could help understand how the impossible happens. Or just remove this check and let the NPE bubble up?

solr/core/src/java/org/apache/solr/cluster/placement/impl/AttributeFetcherImpl.java
```diff
@@ -197,32 +148,78 @@ public AttributeValues fetchAttributes() {
...
+  private static SolrInfoBean.Group getGroupFromMetricRegistry(NodeMetric.Registry registry) {
     switch (registry) {
       case SOLR_JVM:
         return SolrInfoBean.Group.jvm;
       case SOLR_NODE:
         return SolrInfoBean.Group.node;
+      case SOLR_JETTY:
+        return SolrInfoBean.Group.jetty;
       default:
         throw new SolrException(SolrException.ErrorCode.SERVER_ERROR, "Unsupported registry value " + registry);
```
murblanc, line 200: If we do add a `NULL` or `UNDEFINED` `Registry` enum value, it's ok to handle it as `default` here and throw exception (although it getting all the way to here likely denotes a bug in our code).

solr/core/src/java/org/apache/solr/cluster/placement/impl/AttributeFetcherImpl.java
```diff
@@ -197,32 +148,78 @@ public AttributeValues fetchAttributes() {
...
   }
 
-  public static String getMetricSnitchTag(String metricName, NodeMetricRegistry registry) {
-    return SolrClientNodeStateProvider.METRICS_PREFIX + SolrMetricManager.getRegistryName(getGroupFromMetricRegistry(registry), metricName);
+  public static String getMetricSnitchTag(NodeMetric<?> metric) {
+    if (metric.getRegistry() != null) {
+      // regular registry + metricName
+      return SolrClientNodeStateProvider.METRICS_PREFIX +
+          SolrMetricManager.getRegistryName(getGroupFromMetricRegistry(metric.getRegistry())) + ":" + metric.getInternalName();
+    } else if (ImplicitSnitch.tags.contains(metric.getInternalName())) {
```
murblanc, line 209: Can we have the metric instance itself know if it's based on a well known snitch name or a metrics: name? That would made the condition tests  here clearer. Given different constructors are used, that's easy to track.

solr/core/src/java/org/apache/solr/cluster/placement/impl/MetricAttributeImpl.java
```diff
@@ -0,0 +1,138 @@
...
+   */
+  @SuppressWarnings("unchecked")
+  public static final Function<Object, Double> BYTES_TO_GB_CONVERTER = v -> {
+    double sizeInBytes;
+    if (!(v instanceof Number)) {
+      if (v == null) {
+        return null;
+      }
+      try {
+        sizeInBytes = Double.valueOf(String.valueOf(v)).doubleValue();
```
murblanc, line 57: `Double.parseDouble(String.valueOf(v))` does the job.

solr/core/src/java/org/apache/solr/cluster/placement/impl/MetricAttributeImpl.java
```diff
@@ -0,0 +1,138 @@
...
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+    MetricAttribute<?> that = (MetricAttribute<?>) o;
+    return name.equals(that.getName()) && internalName.equals(that.getInternalName()) && converter.equals(that.getConverter());
```
murblanc, line 123: Why the inconsistency between direct field access for `this` instance and getter for `that`?

solr/core/src/java/org/apache/solr/cluster/placement/plugins/AffinityPlacementFactory.java
```diff
@@ -194,7 +195,9 @@ public PlacementPlan computePlacement(Cluster cluster, PlacementRequest request,
 
       // Request all needed attributes
       attributeFetcher.requestNodeSystemProperty(AVAILABILITY_ZONE_SYSPROP).requestNodeSystemProperty(REPLICA_TYPE_SYSPROP);
-      attributeFetcher.requestNodeCoreCount().requestNodeFreeDisk();
+      attributeFetcher
+          .requestNodeMetric(NodeMetricImpl.NUM_CORES)
```
murblanc, line 199: Leaking implementation to the plugin code here.
Something that I think we managed to avoid until now. Shall we continue to try? And expose these metric instances through a factory that will have an interface visible to the client? Or consider it's not worth the trouble?

From `NodeMetricImpl` dependencies go to `ImplicitSnitch` to `ZkStateReader` and to basically the whole codebase.

murblanc: One option (simpler than a factory and less verbose) is to define an enum with metrics in interface `NodeMetric` and have the implementation convert these to the actual instances (in some `if... else` succession ugliness, like was done for example for the `Registry` enum to not leak the actual registry values).

solr/core/src/java/org/apache/solr/cluster/placement/AttributeFetcher.java
sigram: Ah, this is more complicated and it needs to be documented properly - there's a note in `requestCollectionMetrics` about this but I can add it here, too. Collection metrics are fetched from whatever nodes host the replicas, and `fetchFrom` doesn't affect this.

solr/core/src/java/org/apache/solr/cluster/placement/MetricAttribute.java
sigram: There's already a `Metric` class on the classpath, it comes from Dropwizard Metrics... I wanted to avoid the naming conflict. Let me think...

solr/core/src/java/org/apache/solr/cluster/placement/ReplicaMetrics.java
sigram: `AttributeValues.getNodeMetric` follows the same pattern .. maybe both are wrong (or right?).

solr/core/src/java/org/apache/solr/cluster/placement/impl/AttributeFetcherImpl.java
sigram: I disagree, having "system" here makes it clearer what category of metrics we want to keep here. In any case, this is an internal variable name...

solr/core/src/java/org/apache/solr/cluster/placement/impl/AttributeFetcherImpl.java
sigram: Yes, we make two calls per node - one is `getNodeValues` and the other is `getReplicaInfos`.

The number of calls doesn't depend on the number of collections - we already use a set of nodes that covers all replicas for all interesting collections and we only call `getReplicaInfo` once per node (and it returns all replica infos, regardless of collection).

solr/core/src/java/org/apache/solr/cluster/placement/impl/AttributeFetcherImpl.java
sigram: I'll remove this check.

solr/core/src/java/org/apache/solr/cluster/placement/impl/AttributeFetcherImpl.java
sigram: Well-known tags are simple String-s so they are indistinguishable from metric keys (such as `solr.jvm:system.properties:user.name`). We could perhaps move this check to `NodeMetricImpl` and add a pseudo-registry of `WELL_KNOWN_TAG` ... although this sounds maybe too complicated just to prettify these couple lines of code?

solr/core/src/java/org/apache/solr/cluster/placement/impl/MetricAttributeImpl.java
sigram: Ah, this was a problem with class hierarchy - it should cast to `MetricAttributeImpl`, I'll fix this.

solr/core/src/java/org/apache/solr/cluster/placement/impl/CollectionMetricsBuilder.java
```diff
@@ -61,18 +80,19 @@ public ShardMetrics build() {
       replicaMetricsBuilders.forEach((name, replicaBuilder) -> {
         ReplicaMetrics metrics = replicaBuilder.build();
         metricsMap.put(name, metrics);
-        // skip leader from map
         if (replicaBuilder.leader) {
           if (leaderMetricsBuilder == null) {
             leaderMetricsBuilder = replicaBuilder;
           }
-          if (replicaBuilder != leaderMetricsBuilder) {
```
murblanc: Why did you remove that check? It doesn't deserve a new if statement (just `else` to the `leaderMetricsBuilder == null` condition is equivalent) but it's interesting, it means two replicas pretend they are shard leader... Maybe error log?

solr/core/src/java/org/apache/solr/cluster/placement/impl/NodeMetricImpl.java
```diff
@@ -67,7 +67,7 @@ public NodeMetricImpl(String key) {
 
   public NodeMetricImpl(String key, Function<Object, T> converter) {
     super(key, key, converter);
```
murblanc, line 69: Maybe replace implementation by `this(key, Registry.UNSPECIFIED, key, converter);` ?

HoustonPutman:
@sigram is there need to keep this branch around?

PR #2138, opened 2020-12-09
LUCENE-9635: BM25FQuery - Mask encoded norm long value in array lookup

yiluncui:
# Description

Through some experimentation with with the BM25FQuery on long documents, I've discovered that there is a bug that doesn't mask the encoded norm's long value during scoring. For long documents (or long fields) this may cause ArrayIndexOutOfBoundsExceptions.

The line where I suspect the bug is being exposed is here
https://github.com/apache/lucene-solr/blob/master/lucene/sandbox/src/java/org/apache/lucene/sandbox/search/MultiNormsLeafSimScorer.java#L131

Here is a similar use in BM25Similarity with the masking
https://github.com/apache/lucene-solr/blob/c413656b627160d49eb9e9f1f84ec4945db80f0e/lucene/core/src/java/org/apache/lucene/search/similarities/BM25Similarity.java#L233

My experimentation shows that to expose this bug, there must be a match for a token in more than one field (which is what BM25FQuery is for). In addition one of the fields must be >= 32792 tokens long.

I've provided tests in the pull request to demonstrate this.

# Solution

Change the array lookup to norm & 0xff

# Tests

Added tests for single and multiple long documents that exposes this problem.

# Checklist

### DIFF
--- a/lucene/sandbox/src/java/org/apache/lucene/sandbox/search/MultiNormsLeafSimScorer.java
+++ b/lucene/sandbox/src/java/org/apache/lucene/sandbox/search/MultiNormsLeafSimScorer.java
@@ -128,7 +128,7 @@ final class MultiNormsLeafSimScorer {
       for (int i = 0; i < normsArr.length; i++) {
         boolean found = normsArr[i].advanceExact(target);
         assert found;
-        normValue += weightArr[i] * LENGTH_TABLE[(byte) normsArr[i].longValue()];
+        normValue += weightArr[i] * LENGTH_TABLE[Byte.toUnsignedInt((byte) normsArr[i].longValue())];
       }
       current = SmallFloat.intToByte4(Math.round(normValue));
       return true;
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -186,6 +186,9 @@ Bug fixes
 * LUCENE-9365: FuzzyQuery was missing matches when prefix length was equal to the term length
   (Mark Harwood, Mike Drob)
 
+* LUCENE-9635: BM25FQuery - Mask encoded norm long value in array lookup.
+  (Yilun Cui)
+
 Other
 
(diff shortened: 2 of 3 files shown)


jpountz (commented):
Great catch! The fix looks good but I'm unhappy that the text needs to generate such large strings. Could you change the test to e.g. use a similarity that always encodes the length normalization factor as a negative number at index time so that we wouldn't need giant strings to test this?

lucene/sandbox/src/java/org/apache/lucene/sandbox/search/MultiNormsLeafSimScorer.java
```diff
@@ -128,7 +128,7 @@ public boolean advanceExact(int target) throws IOException {
       for (int i = 0; i < normsArr.length; i++) {
         boolean found = normsArr[i].advanceExact(target);
         assert found;
-        normValue += weightArr[i] * LENGTH_TABLE[(byte) normsArr[i].longValue()];
+        normValue += weightArr[i] * LENGTH_TABLE[(byte) normsArr[i].longValue() & 0xff];
```
mikemccand, line 131: Actually, I see that `SimilarityBase.java` uses `Byte.toUnsignedInt` instead -- should we use that here?  It is the same computation?

Or, if you want to stick with `& 0xff`, could you add parens around the `(byte)` cast to make the order of operations clear?

mikemccand (commented):
Thanks @yiluncui -- I left one small comment about how exactly to do the byte to unsigned int conversion.

lucene/sandbox/src/java/org/apache/lucene/sandbox/search/MultiNormsLeafSimScorer.java
yiluncui: I've updated the change using Byte.toUnsignedInt

yiluncui:
Updated the pull request addressing outstanding feedback.

jpountz:
Thank you @yiluncui !

PR #2141, opened 2020-12-11
LUCENE-9346: Support minimumNumberShouldMatch in WANDScorer

zacharymorn:
# Description
Support minimumNumberShouldMatch in WANDScorer

Currently has a few `nocommit` to keep track of questions

# Solution
Similar to `MinShouldMatchSumScorer`, the logic here keeps track of number of matched scorers for each candidate doc, and compares it with `minShouldMatch` to decide if the minimum number of optional clauses have been matched.

# Tests
Passed existing tests (especially those in `TestBooleanMinShouldMatch` and `TestWANDScorer`), and updated some that check for scores.

`./gradlew check` passed with `nocommit` rule commented out for now.

# Checklist


lucene/core/src/java/org/apache/lucene/search/Boolean2ScorerSupplier.java
```diff
@@ -195,10 +201,13 @@ private Scorer opt(Collection<ScorerSupplier> optional, int minShouldMatch,
       for (ScorerSupplier scorer : optional) {
         optionalScorers.add(scorer.get(leadCost));
       }
-      if (minShouldMatch > 1) {
+
+      if (scoreMode == ScoreMode.TOP_SCORES) {
+        return new WANDScorer(weight, optionalScorers, minShouldMatch);
+      } else if (minShouldMatch > 1) {
+        // nocommit minShouldMath > 1 && scoreMode != ScoreMode.TOP_SCORES still requires MinShouldMatchSumScorer.
+        // Do we want to depcate this entirely now ?
```
jbampton, line 209:
```suggestion
        // Do we want to deprecate this entirely now ?
```

jpountz (commented):
Thanks for tackling this! I left a few comments.

lucene/core/src/java/org/apache/lucene/search/Boolean2ScorerSupplier.java
zacharymorn: Fixed. Thanks!

lucene/core/src/java/org/apache/lucene/search/WANDScorer.java
```diff
@@ -130,10 +130,19 @@ private static long scaleMinScore(float minScore, int scalingFactor) {
...
   int upTo; // upper bound for which max scores are valid
 
-  WANDScorer(Weight weight, Collection<Scorer> scorers) throws IOException {
+  int minShouldMatch;
+  int freq;
+
+  WANDScorer(Weight weight, Collection<Scorer> scorers, int minShouldMatch) throws IOException {
     super(weight);
 
+    if (minShouldMatch > scorers.size()) {
```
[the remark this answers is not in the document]
zacharymorn: Makes sense. I've updated the condition and the exception message.

lucene/core/src/java/org/apache/lucene/search/WANDScorer.java
```diff
@@ -130,10 +130,19 @@ private static long scaleMinScore(float minScore, int scalingFactor) {
...
+      throw new IllegalArgumentException("minShouldMatch should be <= the number of scorers");
+    }
+
     this.minCompetitiveScore = 0;
+    this.minShouldMatch = minShouldMatch > 0 ? minShouldMatch : 0;
```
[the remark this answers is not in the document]
zacharymorn: I put it there originally to guard against any accidental invalid negative minShouldMatch being passed in, but yeah it should probably be handled as assertion. I've updated it.

lucene/core/src/java/org/apache/lucene/search/WANDScorer.java
```diff
@@ -271,6 +280,19 @@ public boolean matches() throws IOException {
             return false;
           }
         }
+
+        // minCompetitiveScore satisfied, now checks if the doc has enough required number of matches
+        // Combining this loop with the above makes for complicated conditional checks, so keeping them separate for readability
```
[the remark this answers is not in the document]
zacharymorn: Makes sense. I gave this a second try and actually able to arrive at a clean looking solution. Please let me know if this version looks good.

lucene/core/src/java/org/apache/lucene/search/Boolean2ScorerSupplier.java
```diff
@@ -230,10 +232,13 @@ private Scorer opt(
       for (ScorerSupplier scorer : optional) {
         optionalScorers.add(scorer.get(leadCost));
       }
-      if (minShouldMatch > 1) {
+
+      if (scoreMode == ScoreMode.TOP_SCORES) {
+        return new WANDScorer(weight, optionalScorers, minShouldMatch);
+      } else if (minShouldMatch > 1) {
+        // nocommit minShouldMath > 1 && scoreMode != ScoreMode.TOP_SCORES still requires MinShouldMatchSumScorer.
+        // Do we want to deprecate this entirely now ?
```
jpountz, line 240: Maybe not now in order to keep this PR contained, but it would be nice if we could handle this case with WANDScorer too indeed.

lucene/core/src/java/org/apache/lucene/search/WANDScorer.java
```diff
@@ -130,10 +130,21 @@ private static long scaleMinScore(float minScore, int scalingFactor) {
 
   int upTo; // upper bound for which max scores are valid
 
-  WANDScorer(Weight weight, Collection<Scorer> scorers) throws IOException {
+  int minShouldMatch;
```
jpountz, line 133: Let's make it final?

lucene/core/src/java/org/apache/lucene/search/WANDScorer.java
```diff
@@ -130,10 +130,21 @@ private static long scaleMinScore(float minScore, int scalingFactor) {
...
+  WANDScorer(Weight weight, Collection<Scorer> scorers, int minShouldMatch) throws IOException {
     super(weight);
 
+    if (minShouldMatch >= scorers.size()) {
+      throw new IllegalArgumentException("minShouldMatch should be < the number of scorers");
+    }
+
     this.minCompetitiveScore = 0;
+
+    assert minShouldMatch >=0 : "minShouldMatch should not be negative, but got " + minShouldMatch;
```
jpountz, line 145:
```suggestion
    assert minShouldMatch >= 0 : "minShouldMatch should not be negative, but got " + minShouldMatch;
```

lucene/core/src/java/org/apache/lucene/search/WANDScorer.java
```diff
@@ -447,6 +461,23 @@ private int doNextCompetitiveCandidate() throws IOException {
       }
     }
 
+      // nocommit Do we still need the following given TwoPhaseIterator.matches already performs the check
+//     minCompetitiveScore satisfied, now checks if the doc has enough required number of matches
```
jpountz, line 465: Indeed I don't think we need it, but I like doing it so that the approximation doesn't return documents knowing that no match is possible.

lucene/core/src/java/org/apache/lucene/search/WANDScorer.java
```diff
@@ -447,6 +461,23 @@ private int doNextCompetitiveCandidate() throws IOException {
...
+//    while (freq < minShouldMatch) {
+//      if (freq + tailSize >= minShouldMatch) {
+//        // a match on doc is still possible, try to
+//        // advance scorers from the tail
+//        advanceTail();
```
jpountz, line 470: I don't think we should advanceTail here, which may be costly. We should just make sure that `freq + tailSize >= minShouldMatch` and otherwise call your `else` block.

jpountz (commented):
I left some comments but it looks great in general. Could you add more tests as well, e.g. a variant of TestWANDScorer#testBasics that configures a minShouldMatch of 2?

lucene/core/src/java/org/apache/lucene/search/WANDScorer.java
zacharymorn: Done.

lucene/core/src/java/org/apache/lucene/search/Boolean2ScorerSupplier.java
zacharymorn: Sounds good. I've removed the no commit comment for now, but let me give it a second try in the next few days.

lucene/core/src/java/org/apache/lucene/search/Boolean2ScorerSupplier.java
```diff
@@ -74,11 +74,13 @@ private long computeCost() {
...
-              optionalScorers.stream().mapToLong(ScorerSupplier::cost),
-              optionalScorers.size(),
-              minShouldMatch);
+      // nocommit The cost calculation here copies that in WANDScorer's constructor, and may need to be adjusted?
+      final long shouldCost = scoreMode == ScoreMode.TOP_SCORES ?
+                              optionalScorers.stream().mapToLong(ScorerSupplier::cost).sum() :
+                              MinShouldMatchSumScorer.cost(
+                                      optionalScorers.stream().mapToLong(ScorerSupplier::cost),
+                                      optionalScorers.size(),
+                                      minShouldMatch);
```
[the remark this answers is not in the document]
zacharymorn: I see. Previously I wasn't entirely sure about the handling here. I've reverted the related changes and updated `WANDScorer` to share the same cost function with that of `minShouldMatchSumScorer`.

lucene/core/src/java/org/apache/lucene/search/WANDScorer.java
zacharymorn: Makes sense.

lucene/core/src/java/org/apache/lucene/search/WANDScorer.java
zacharymorn: Makes sense. I've updated the code to include this check in the while loop condition.

zacharymorn:
> I left some comments but it looks great in general. Could you add more tests as well, e.g. a variant of TestWANDScorer#testBasics that configures a minShouldMatch of 2?

Thanks for the review and comments! I've added a variant of `TestWANDScorer#testBasics` and broke it apart into 3 separate test methods. I also ran `gradlew :lucene:core:spotlessApply` to do some auto formatting in order to keep the build / validation happy.

jpountz (approved):
Thanks this looks good to me. I'll merge in the next few days.

lucene/core/src/java/org/apache/lucene/search/Boolean2ScorerSupplier.java
zacharymorn: I've taken a second look at this, and had the following thoughts / questions:

Previously, in Boolean2ScorerSupplier#opt, the scorer picking logic is

```
if (minShouldMatch > 1) {
  return MinShouldMatchSumScorer...
} else if (scoreMode == ScoreMode.TOP_SCORES) {
  return WANDScorer...
} else {
  return DisjunctionSumScorer...
}
```

With the current changes in this PR, we now have

```
if (scoreMode == ScoreMode.TOP_SCORES) {
  return WANDScorer...
} else if (minShouldMatch > 1) {
  return MinShouldMatchSumScorer...
} else {
  return DisjunctionSumScorer...
}
```

I noticed that with trivial changes, WANDScorer seems to be usable for all ScoreMode when `scoreMode.needsScore == true` as well (passed all existing tests with trivial updates). So we can arrive at something like this

```
if (scoreMode.needsScore()) {
  return WANDScorer...
} else if (minShouldMatch > 1) {
  return MinShouldMatchSumScorer...
} else {
  return DisjunctionSumScorer...
}
```

Is there a reason `WANDScorer` is only used for TOP_DOCS before? I did read up the documentation and discussion in https://issues.apache.org/jira/browse/LUCENE-4100, but it's a bit long so I'm not sure I'm following all the details there.

Following the above, since `WANDScorer` can already handle the case where `minShouldMatch == 0`, then I can see MinShouldMatchSumScorer be merged into WANDScorer mostly to handle non-scoring mode, arriving at 

```
if (?) {
  return WANDScorer...
} else {
  return DisjunctionSumScorer...
}
```

At this point, it actually seems like `WANDScorer` can handle all the cases already, so `DisjunctionSumScorer` can be deprecated as well to just return `WANDScorer` here? 

Is the above the right reasoning of how things would go for the deprecation?

lucene/core/src/java/org/apache/lucene/search/ScorerUtil.java
```diff
@@ -0,0 +1,49 @@
...
+ * limitations under the License.
+ */
+package org.apache.lucene.search;
+
+import java.util.stream.LongStream;
+import java.util.stream.StreamSupport;
+import org.apache.lucene.util.PriorityQueue;
+
+/** Util class for Scorer related methods */
+public class ScorerUtil {
```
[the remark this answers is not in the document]
zacharymorn: Done.

zacharymorn:
> Thanks this looks good to me. I'll merge in the next few days.

Thanks Adrien for the review and approval! I also took a look and thought about the possible steps involved in the complete replacement of MinShouldMatchSumScorer with WANDScorer, and left some thoughts / questions in the conversation above. But yeah in general I also feel that will significantly increase the scope of this ticket / PR. 

Please let me know what you think.

lucene/core/src/java/org/apache/lucene/search/Boolean2ScorerSupplier.java
jpountz: > Is there a reason WANDScorer is only used for TOP_DOCS before?

Mostly because it's a bit slower since it needs to track two priority queues and one linked list, when the simple DisjunctionScorer only needs to track a single priority queue.

> since WANDScorer can already handle the case where minShouldMatch == 0, then I can see MinShouldMatchSumScorer be merged into WANDScorer mostly to handle non-scoring mode

+1 to give it a try in a follow-up PR!

jpountz:
@zacharymorn I merged your PR because it was good progress already, but I'm also +1 on your idea of replacing MinShouldMatchSumScorer with WANDScorer since they share very similar logic. Let's give it a try in a follow-up PR?

zacharymorn:
> @zacharymorn I merged your PR because it was good progress already, but I'm also +1 on your idea of replacing MinShouldMatchSumScorer with WANDScorer since they share very similar logic. Let's give it a try in a follow-up PR?

hi @jpountz, sorry for the delay and thanks for merging this PR! I've created a new jira issue https://issues.apache.org/jira/browse/LUCENE-9668 to follow up on this, and will work on it and create a new PR in the next few days.

PR #2152, opened 2020-12-17
SOLR-14034: remove deprecated min_rf references

trdillon:
# Description

Remove min_rf and MIN_REPFACT references in code, tests and documentation.

# Solution

Incrementally remove min_rf and MIN_REPFACT references.

# Tests

./gradlew check

# Checklist

### DIFF
--- a/solr/core/src/test/org/apache/solr/cloud/HttpPartitionTest.java
+++ b/solr/core/src/test/org/apache/solr/cloud/HttpPartitionTest.java
@@ -162,7 +162,7 @@ public class HttpPartitionTest extends AbstractFullDistribZkTestBase {
       createCollection(testCollectionName, "conf1", 1, 2);
       cloudClient.setDefaultCollection(testCollectionName);
 
-      sendDoc(1, 2);
+      sendDoc(1);
 
       JettySolrRunner leaderJetty = getJettyOnPort(getReplicaPort(getShardLeader(testCollectionName, "shard1", 1000)));
       List<Replica> notLeaders =
@@ -176,7 +176,7 @@ public class HttpPartitionTest extends AbstractFullDistribZkTestBase {
       leaderProxy.close();
 
       // indexing during a partition
-      int achievedRf = sendDoc(2, 1, leaderJetty);
+      int achievedRf = sendDoc(2, leaderJetty);
       assertEquals("Unexpected achieved replication factor", 1, achievedRf);
       try (ZkShardTerms zkShardTerms = new ZkShardTerms(testCollectionName, "shard1", cloudClient.getZkStateReader().getZkClient())) {
         assertFalse(zkShardTerms.canBecomeLeader(notLeaders.get(0).getName()));
@@ -228,7 +228,7 @@ public class HttpPartitionTest extends AbstractFullDistribZkTestBase {
     leaderProxy.close();
 
     // indexing during a partition
-    sendDoc(2, null, leaderJetty);
+    sendDoc(2, leaderJetty);
     // replica should publish itself as DOWN if the network is not healed after some amount time
     waitForState(testCollectionName, notLeader.getName(), DOWN, 10000);
     
@@ -282,7 +282,7 @@ public class HttpPartitionTest extends AbstractFullDistribZkTestBase {
         }
       }
       // always send doc directly to leader without going through proxy
-      sendDoc(d + 4, null, leaderJetty); // 4 is offset as we've already indexed 1-3
+      sendDoc(d + 4, leaderJetty); // 4 is offset as we've already indexed 1-3
     }
     
     // restore connectivity if lost
@@ -352,7 +352,7 @@ public class HttpPartitionTest extends AbstractFullDistribZkTestBase {
     leaderProxy.close();
     
     // indexing during a partition
-    sendDoc(2, null, leaderJetty);
+    sendDoc(2, leaderJetty);
     
     Thread.sleep(sleepMsBeforeHealPartition);
     proxy0.reopen();
@@ -360,7 +360,7 @@ public class HttpPartitionTest extends AbstractFullDistribZkTestBase {
     SocketProxy proxy1 = getProxyForReplica(notLeaders.get(1));
     proxy1.close();
     
-    sendDoc(3, null, leaderJetty);
+    sendDoc(3, leaderJetty);
     
     Thread.sleep(sleepMsBeforeHealPartition);
     proxy1.reopen();
@@ -527,22 +527,18 @@ public class HttpPartitionTest extends AbstractFullDistribZkTestBase {
     return getHttpSolrClient(url);
   }
 
-  protected int sendDoc(int docId) throws Exception {
-    return sendDoc(docId, null);
-  }
-
   // Send doc directly to a server (without going through proxy)
-  protected int sendDoc(int docId, Integer minRf, JettySolrRunner leaderJetty) throws IOException, SolrServerException {
+  protected int sendDoc(int docId, JettySolrRunner leaderJetty) throws IOException, SolrServerException {
     try (HttpSolrClient solrClient = new HttpSolrClient.Builder(leaderJetty.getBaseUrl().toString()).build()) {
-      return sendDoc(docId, minRf, solrClient, cloudClient.getDefaultCollection());
+      return sendDoc(docId, solrClient, cloudClient.getDefaultCollection());
     }
   }
 
-  protected int sendDoc(int docId, Integer minRf) throws Exception {
-    return sendDoc(docId, minRf, cloudClient, cloudClient.getDefaultCollection());
+  protected int sendDoc(int docId) throws Exception {
+    return sendDoc(docId, cloudClient, cloudClient.getDefaultCollection());
   }
 
-  protected int sendDoc(int docId, Integer minRf, SolrClient solrClient, String collection) throws IOException, SolrServerException {
+  protected int sendDoc(int docId, SolrClient solrClient, String collection) throws IOException, SolrServerException {
     SolrInputDocument doc = new SolrInputDocument();
     doc.addField(id, String.valueOf(docId));
     doc.addField("a_t", "hello" + docId);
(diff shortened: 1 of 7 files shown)


madrob (commented):
LGTM, will wait for another set of eyes before committing. @cpoerschke?

@trdillon can you add an entry to `solr/CHANGES` with your preferred attribution?

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseCloudSolrClient.java
```diff
@@ -716,7 +716,6 @@ protected RouteException getRouteException(SolrException.ErrorCode serverError,
         if (rf == null || routeRf < rf)
           rf = routeRf;
       }
-      minRf = (Integer)header.get(UpdateRequest.MIN_REPFACT);
```
cpoerschke, line 719: Suggest to also remove the `Integer minRf = null;` at line 696 above.

solr/core/src/test/org/apache/solr/cloud/HttpPartitionTest.java
```diff
@@ -548,9 +548,6 @@ protected int sendDoc(int docId, Integer minRf, SolrClient solrClient, String co
     doc.addField("a_t", "hello" + docId);
 
     UpdateRequest up = new UpdateRequest();
-    if (minRf != null) {
-      up.setParam(UpdateRequest.MIN_REPFACT, String.valueOf(minRf));
-    }
```
cpoerschke, line 553: Hi @trdillon - thanks for opening this pull request!

Okay, I think I see it now, so your https://issues.apache.org/jira/browse/SOLR-14034?focusedCommentId=17223427&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17223427 question is about this sort of test change here i.e. whether or not `minRf` should remain an unused `sendDoc` argument or whether or not it should be removed. I'm thinking here in `HttpPartitionTest` removal makes sense (haven't yet looked at `ReplicationFactorTest`). What do you think?

solr/core/src/test/org/apache/solr/cloud/ReplicationFactorTest.java
```diff
@@ -461,38 +447,24 @@ protected void doDelete(UpdateRequest req, String msg, int expectedRf, int retri
...
-    boolean minRfExplicit = maybeAddMinRfExplicitly(minRf, up);
     SolrInputDocument doc = new SolrInputDocument();
     doc.addField(id, String.valueOf(docId));
     doc.addField("a_t", "hello" + docId);
     up.add(doc);
-    return runAndGetAchievedRf(up, minRfExplicit, minRf);
+    return runAndGetAchievedRf(up, minRf);
   }
   
-  private int runAndGetAchievedRf(UpdateRequest up, boolean minRfExplicit, int minRf) throws SolrServerException, IOException {
```
cpoerschke, line 472: How about also removing the `minRf` argument here since it's now no longer used?

solr/core/src/test/org/apache/solr/cloud/ReplicationFactorTest.java
```diff
@@ -461,38 +447,24 @@ protected void doDelete(UpdateRequest req, String msg, int expectedRf, int retri
...
-    }
     return cloudClient.getMinAchievedReplicationFactor(cloudClient.getDefaultCollection(), response);
   }
 
   private void assertMinRfInResponse(int minRf, NamedList<Object> response) {
-    Object minRfFromResponse = response.findRecursive("responseHeader", UpdateRequest.MIN_REPFACT);
+    Object minRfFromResponse = response.findRecursive("responseHeader");
     assertNotNull("Expected min_rf header in the response", minRfFromResponse);
     assertEquals("Unexpected min_rf in response", ((Integer)minRfFromResponse).intValue(), minRf);
   }
```
cpoerschke, line 466: Looks like `assertMinRfInResponse` is now also unused then, if so suggest to remove it also.

cpoerschke:
At https://github.com/apache/lucene-solr/blob/releases/lucene-solr/8.7.0/solr/core/src/java/org/apache/solr/update/processor/DistributedUpdateProcessor.java#L754-L756 there's also a `Implementing min_rf here was a bit tricky. ...` comment reference to `min_rf` -- any thoughts on either removing it or rewording it somehow?

solr/core/src/test/org/apache/solr/cloud/HttpPartitionTest.java
trdillon: @cpoerschke Thanks for all your help with this one.

`sendDoc` seems OK to remove from `HttpPartitionTest` as it's only passed `minRf` as an argument, but in `ReplicationFactorTest` it is passed `expectedRf` as an argument in `addDocs`:
https://github.com/apache/lucene-solr/blob/98f12f4aeb9f6ec9f5c4de53f9faddc41043df59/solr/core/src/test/org/apache/solr/cloud/ReplicationFactorTest.java#L417

`sendDocsWithRetry`  is implemented here and used in a few different test classes (`DistributedVersionInfoTest`, `LeaderFailoverAfterPartitionTest` and `ForceLeaderTest`) with `minRf` as an argument:
https://github.com/apache/lucene-solr/blob/98f12f4aeb9f6ec9f5c4de53f9faddc41043df59/solr/test-framework/src/java/org/apache/solr/cloud/AbstractFullDistribZkTestBase.java#L941

But it's also implemented using `expectedRfDBQ` and `expectedRf` in `ReplicationFactorTest`:

- https://github.com/apache/lucene-solr/blob/98f12f4aeb9f6ec9f5c4de53f9faddc41043df59/solr/core/src/test/org/apache/solr/cloud/ReplicationFactorTest.java#L230
- https://github.com/apache/lucene-solr/blob/98f12f4aeb9f6ec9f5c4de53f9faddc41043df59/solr/core/src/test/org/apache/solr/cloud/ReplicationFactorTest.java#L427

I wasn't sure how to proceed with this as looking at the tests it seems that `expectedRf` is relied upon quite often.

solr/core/src/test/org/apache/solr/cloud/ReplicationFactorTest.java
```diff
@@ -461,38 +447,18 @@ protected void doDelete(UpdateRequest req, String msg, int expectedRf, int retri
 
   protected int sendDoc(int docId, int minRf) throws Exception {
     UpdateRequest up = new UpdateRequest();
-    boolean minRfExplicit = maybeAddMinRfExplicitly(minRf, up);
     SolrInputDocument doc = new SolrInputDocument();
     doc.addField(id, String.valueOf(docId));
     doc.addField("a_t", "hello" + docId);
     up.add(doc);
-    return runAndGetAchievedRf(up, minRfExplicit, minRf);
+    return runAndGetAchievedRf(up);
```
cpoerschke, line 454: Alright, so with `maybeAddMinRfExplicitly` gone and the `minRf` argument in `runAndGetAchievedRf` gone the `minRf` argument of `sendDoc` becomes unused, right? I wonder how the code would work out if `sendDoc` had its `minRf` argument taken away too? Bit like peeling an onion this here i.e. multiple layers (but it hopefully won't make anyone cry).

solr/core/src/test/org/apache/solr/cloud/HttpPartitionTest.java
cpoerschke: Thanks for the code pointers and detailed explanation!

How about we scope this pull request to remove `minRf` in almost all code paths i.e. `sendDoc` gets its unused argument removed in both the replication factor and http partition tests but `sendDocsWithRetry` remains unchanged for now since as you say it's used in other test classes too?

It appears that the `minRf` argument in `sendDocsWithRetry` is indeed unused already but it might be clearer to tidy that up in a separate follow-up pull request rather than mix it in here at this time.

solr/core/src/test/org/apache/solr/cloud/ReplicationFactorTest.java
trdillon: The tests for `ReplicationFactorTest` and `HttpPartitionTest` passed on my end after removing the `minRf` argument from `sendDoc`. I committed the `ReplicationFactorTest` changes separately as it meant removing the `expectedRf` argument from the `sendDoc` call in `addDocs` and I wasn't sure if this was OK, even though the test still passed.

trdillon:
> 
> 
> At https://github.com/apache/lucene-solr/blob/releases/lucene-solr/8.7.0/solr/core/src/java/org/apache/solr/update/processor/DistributedUpdateProcessor.java#L754-L756 there's also a `Implementing min_rf here was a bit tricky. ...` comment reference to `min_rf` -- any thoughts on either removing it or rewording it somehow?

The comment still seems to be relevant and makes sense without the first sentence that references `minRf`. From my limited perspective I think it would be OK to just remove the first sentence.

solr/core/src/test/org/apache/solr/cloud/HttpPartitionTest.java
tflobbe: > It appears that the minRf argument in sendDocsWithRetry is indeed unused already but it might be clearer to tidy that up in a separate follow-up pull request rather than mix it in here at this time.

I'm wondering if it would be better for this method to actually start using this parameter, and retry in the case of `minRf` not achieved? This is not something that was being done in this method, but may help improve some of the tests using it. I don't think such a change should be done in this PR, but maybe lets not remove the parameter yet then. WDYT?

tflobbe (approved):
LGTM

solr/core/src/test/org/apache/solr/cloud/ReplicationFactorTest.java
cpoerschke: I appreciate you committing the changes separately, thanks. The `ReplicationFactorTest` change looks good to me.

solr/core/src/test/org/apache/solr/cloud/HttpPartitionTest.java
cpoerschke: > I'm wondering if it would be better for this method to actually start using this parameter, ...

Ah, yes, I agree, if it's a case of a not-yet-used parameter then keeping it and starting to use it makes sense. Removal would only be for no-longer-used parameter cases. Good catch!

cpoerschke (approved):
LGTM. Will aim to proceed to merge the PR about this time tomorrow if there are no further comments etc. Thanks!

PR #2154, opened 2020-12-17
LUCENE-9642: Rotate triangle points before checking triangle orientation

iverase:
When encoding.a triangle we perform two actions: 

1) Make sure that the orientation of the triangle is CCW
2) Place the point with minimum X in the first position so we might rotate the points in the triangle.

It might happen if the triangle is degenerated that rotating the points might change the triangle orientation. In order to avoid that this change performs the rotation of the points first and then checks for the orientation.


madrob:
Can you add a unit test?

iverase:
good point @madrob, unit test added.

PR #2155, opened 2020-12-17
LUCENE-9641: Support for spatial relationships in LatLonPoint

iverase:
This PR adds the possibility of querying a LatLonPoint using the same spatial relationships supported by LatLonShape. In order to do that, I have rename ShapeQuery class to SpatialQuery and added  a new Interface (SpatialVisitor) that can be implemented by both points and shapes.

LatLonPointInGeometryQuery has been replaced by LatLonPointQuery.
LatLonDocValuesPointInGeometryQuery has been replaced by LatLonDocValuesQuery.

Performance test for LatLonPoint shows same performance as before:

```
||Approach||Shape||M hits/sec      ||QPS            ||Hit count      ||
|points|polyRussia|16.08|16.13|-0%|4.58|4.60|-0%|3508846|3508846| 0%|
|points|polyMedium|9.14|9.08| 1%|112.00|111.28| 1%|2693559|2693559| 0%|
|points|poly 10|87.29|87.03| 0%|55.20|55.03| 0%|355809475|355809475| 0%|
```

cc: @nknize


rmuir:
I think the JIRA doesn't give a good explanation for adding thousand lines of code. What is the use-case? To me these relationships don't make sense for points... since they are not shapes.

iverase:
I updated the jira ticket with the motivation for this change. What I disagree here is saying the points are not shapes, e.g you can index points using LatLonShape. Points are just the simplest shape you can have, and it is a very common use case to have data that only contains point so it make sense to have a specialise field to index just points.

Unfortunately LatLonPoint has some missing functionality in respect to LatLonShape, this change is just trying to close that gap.

lucene/core/src/java/org/apache/lucene/document/LatLonPointQuery.java
```diff
@@ -0,0 +1,174 @@
...
+
+      @Override
+      protected Predicate<byte[]> within() {
+        return packedValue -> component2DPredicate.test(NumericUtils.sortableBytesToInt(packedValue, 0),
+                NumericUtils.sortableBytesToInt(packedValue, Integer.BYTES));
+      }
+
+      @Override
+      protected Function<byte[], Component2D.WithinRelation> contains() {
+        return packedValue -> component2D.withinPoint(
```
nknize, line 115: It doesn't make sense for ```LatLonPointQuery``` to search for Points that contain any geometry other than points. I'd throw an exception if the incoming geometry type is not a point or array of points.

lucene/core/src/java/org/apache/lucene/document/LatLonPointQuery.java
```diff
@@ -0,0 +1,174 @@
...
+    this.component2D = LatLonGeometry.create(geometries);
+    this.geometries = geometries.clone();
+  }
+  
+  @Override
+  protected SpatialVisitor getSpatialVisitor() {
+    if (component2D.getMinY() > component2D.getMaxY()) {
+      // encodeLatitudeCeil may cause minY to be > maxY iff
+      // the delta between the longitude < the encoding resolution
+      return EMPTYVISITOR;
```
nknize, line 69: I feel like there should be a better way of handling this. This has the look and feel of silent ignorance; that the encoding allowed something that doesn't make sense. I think we should do a better job handling this upstream instead of creating a visitor that looks the other way on such oddities.

nknize (changes requested):
I agree with the motivation and use case (e.g., finding multipoint documents that are `WITHIN` a query geometry / geometry collection). I was initially concerned about the massive refactor and changes, though, that we were seemingly now blending Point and Shape query logic "just" for the ability to execute a `WITHIN` GeometryCollection query on a point index. But it looks like the critical path logic is pretty well isolated between Points and Shapes through the use of the `ShapeVisitor`. My primary concern is that a future change intended to improve Shape performance or introduce a new feature now affects logic shared by Points thus creating a "whackamole" situation. Whileit seems the `ShapeVisitor` maintains this "firewall" between the two, we need to at least make sure the javadocs are crystal clear about what's going on. 

I definitely don't like the idea of an `EmptyVisitor` and suggest this be rethought before landing these changes.

lucene/core/src/java/org/apache/lucene/document/LatLonPointQuery.java
iverase: The problem was in the way the component predicate is built. I have another thought and I think I push a better solution in 1cebeff

lucene/core/src/java/org/apache/lucene/document/LatLonPointQuery.java
iverase: Make sense and makes implementation cleaner. Still I return a MatchNoDocs query for the Query builders. Change is in 3e97bb8.

iverase:
if there are no more comments, I will push this change next week.

nknize:
With the holidays and folks out on vacation would you mind holding off until the end of next week? I plan to have a look at the other two changes by the end of this week and I think it would be fair to make sure @rmuir has enough time to have a look as well.

iverase:
Sure, just hope it can make it to the next release.

lucene/core/src/java/org/apache/lucene/document/LatLonPointQuery.java
nknize: :+1:  Much cleaner.

lucene/core/src/java/org/apache/lucene/document/LatLonPointQuery.java
nknize: This looks good!

nknize (approved):
I'm good with the changes. Wouldn't mind @rmuir having one more look (if he has time) to make sure there aren't any hidden dragons.

PR #2157, opened 2020-12-18
LUCENE-9644: diversity heuristic for HNSW graph neighbor selection

msokolov:
This replaces the simple nearest neighbor selection with a criterion that takes into account the distance of the neighbors from each other. It is seen to provide dramatically improved recall on at least two datasets, and is what is being used by our reference implementation, hnswlib. Also:
  * Split Neighbors into NeighborArray and NeighborQueue; use queue for gathering results; store graph arcs in arrays since diversity selection does not use a queue
  * Add InfoStream progress messages to HnswGraphBuilder
  * Add options to KnnGraphTester to support testing using ann-benchmarks (no warmup, write output to file)
  * Improve memory usage; eliminate more object allocations; replaced iterator objects with single-use iterator "views"


msokolov:
I'll push today if there are no comments.

PR #2159, opened 2020-12-21
SOLR-14923: Nested docs indexing performance.

dsmiley:
https://issues.apache.org/jira/browse/SOLR-14923

* UpdateLog.openRealtimeSearcher() was being called for every incoming document when the schema had _nest_path_.  It is now more limited to in-place-update of a child doc and in /get when a child doc is given, both of which are uncommon.
* Atomic/partial updates to nested documents should be faster. In-place updates of the same might be slower (needs to call openRealtimeSearcher).

Refactoring & minor bugs/improvements:
* Simplified AddUpdateCommand.getLuceneDoc & getLuceneDocsIfNested relationship, and DirectUpdateHandler2 which calls them.
* AddUpdateCommand.getIndexedId needed to be updated when _route_ was specified.
* NestedShardedAtomicUpdateTest no longer extends AbstractFullDistribZkTestBase because it wasn't really leveraging the "control client" checking, and it added too much complexity to debug failures.
* AtomicUpdateDocumentMerger: simplified merge; possibly now supports updates to anonymous children
* No longer need RTG.Resolution.DOC_WITH_CHILDREN

### DIFF
--- a/solr/core/src/java/org/apache/solr/handler/component/RealTimeGetComponent.java
+++ b/solr/core/src/java/org/apache/solr/handler/component/RealTimeGetComponent.java
@@ -29,16 +29,19 @@ import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicLong;
+import java.util.function.BiConsumer;
 import java.util.stream.Collectors;
 
 import com.google.common.collect.Lists;
-import com.google.common.collect.Sets;
 import org.apache.lucene.document.Document;
 import org.apache.lucene.document.Field;
 import org.apache.lucene.index.DocValuesType;
 import org.apache.lucene.index.IndexableField;
+import org.apache.lucene.index.LeafReader;
 import org.apache.lucene.index.LeafReaderContext;
 import org.apache.lucene.index.Term;
+import org.apache.lucene.index.Terms;
+import org.apache.lucene.index.TermsEnum;
 import org.apache.lucene.search.MatchAllDocsQuery;
 import org.apache.lucene.search.Query;
 import org.apache.lucene.search.ScoreMode;
@@ -49,7 +52,6 @@ import org.apache.solr.client.solrj.SolrResponse;
 import org.apache.solr.cloud.CloudDescriptor;
 import org.apache.solr.cloud.ZkController;
 import org.apache.solr.common.SolrDocument;
-import org.apache.solr.common.SolrDocumentBase;
 import org.apache.solr.common.SolrDocumentList;
 import org.apache.solr.common.SolrException;
 import org.apache.solr.common.SolrException.ErrorCode;
@@ -100,7 +102,6 @@ import static org.apache.solr.common.params.CommonParams.VERSION_FIELD;
 public class RealTimeGetComponent extends SearchComponent
 {
   private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
-  private static final Set<String> NESTED_META_FIELDS = Sets.newHashSet(IndexSchema.NEST_PATH_FIELD_NAME, IndexSchema.NEST_PARENT_FIELD_NAME);
   public static final String COMPONENT_NAME = "get";
 
   @Override
@@ -239,11 +240,16 @@ public class RealTimeGetComponent extends SearchComponent
 
    try {
 
-
+     boolean opennedRealtimeSearcher = false;
      BytesRefBuilder idBytes = new BytesRefBuilder();
      for (String idStr : reqIds.allIds) {
        fieldType.readableToIndexed(idStr, idBytes);
-       if (ulog != null) {
+       if (!opennedRealtimeSearcher && !params.get(ShardParams._ROUTE_, idStr).equals(idStr)) { // idStr is a child doc
+         searcherInfo.clear();
+         resultContext = null;
+         ulog.openRealtimeSearcher();  // force open a new realtime searcher
+         opennedRealtimeSearcher = true;
+       } else if (ulog != null) {
          Object o = ulog.lookup(idBytes.get());
          if (o != null) {
            // should currently be a List<Oper,Ver,Doc/Id>
@@ -257,9 +263,12 @@ public class RealTimeGetComponent extends SearchComponent
 
                if (mustUseRealtimeSearcher) {
                  // close handles to current searchers & result context
-                 searcherInfo.clear();
-                 resultContext = null;
-                 ulog.openRealtimeSearcher();  // force open a new realtime searcher
+                 if (!opennedRealtimeSearcher) {
+                   searcherInfo.clear();
+                   resultContext = null;
+                   ulog.openRealtimeSearcher();  // force open a new realtime searcher
+                   opennedRealtimeSearcher = true;
+                 }
                  o = null;  // pretend we never found this record and fall through to use the searcher
                  break;
                }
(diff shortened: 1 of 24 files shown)


dsmiley:
CC @thomaswoeckinger @moshebla 

Tests pass, precommit passes.

@markrmiller FYI I observed that NestedShardedAtomicUpdateTest on your ref-impl branch commented out an assertion that I wound up fixing in this PR.  I converted that test to use SolrCloudTestCase which made it way easier to debug some elusive bugs.

dsmiley:
I believe everything MUSE_DEV reported is a non-issue.  Only one thread touches any AddUpdateCommand and the SolrInputDoc on it.  Perhaps MUSE_DEV saw the `synchronized` keyword inside UpdateLog.add and thought otherwise.  But that thing is synchronizing on UpdateLog itself, not the command.

I included SOLR-15069 (parentFilter on [child] is now not needed) inside here, but that should get merged first, and thus it won't show up ultimately for this issue.  SOLR-15069 allows us to safely (no loss of data) do atomic/partial updates when there is no nest path field declared.

@ctargett perhaps you might look over the ref guide changes in here?  I did some editing to reduce the use of the word "block" in favor of the word "nest", and also did some other clarifications.

solr/core/src/java/org/apache/solr/response/transform/ChildDocTransformer.java
```diff
@@ -76,6 +87,37 @@ public String getName()  {
...
+
+  private int getPrevRootGivenId(LeafReaderContext leafReaderContext, int segRootId,
+                                 BytesRef idBytes) throws IOException {
+    final LeafReader reader = leafReaderContext.reader();
+    final Terms terms = reader.terms(IndexSchema.ROOT_FIELD_NAME); // never returns null here
+    final TermsEnum iterator = terms.iterator();
+    if (iterator.seekExact(idBytes)) {
+      PostingsEnum docs = iterator.postings(null, PostingsEnum.NONE);
+      docs = BitsFilteredPostingsEnum.wrap(docs, reader.getLiveDocs());
+      int id = docs.nextDoc();
```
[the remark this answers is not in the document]
dsmiley: Wrong; iterator.postings() never returns null -- so says its documentation.

solr/solr-ref-guide/src/indexing-nested-documents.adoc
```diff
@@ -218,31 +225,40 @@ Indexing nested documents _requires_ an indexed field named `\_root_`:
...
+** It may be infeasible to use `required` for any field names that aren't required for all types of
+documents.
 ** Even child documents need a _globally_ unique `id`.
-* `\_root_` must be configured to either be stored (`stored="true"`) or use doc values (`docValues="true"`) to enable <<updating-parts-of-documents#updating-child-documents,atomic updates of nested documents>>.
-** Also, beware of `uniqueBlock(\_root_)` <<json-facet-api#stat-facet-functions,field type limitation>>, if you plan to use one.
-* `\_nest_path_` is an optional field that (if definied) will be populated by Solr automatically with the ancestor path of each non-root document.
+* The `\_root_` must be indexed (`indexed="true"`) but doesn't need to
+ be either stored (`stored="true"`) or use doc values (`docValues="true"`), however you are free
+to do so if you find it useful.  If you want to use `uniqueBlock(\_root_)`
+<<json-facet-api#stat-facet-functions,field type limitation>>, then you should enable docValues.
```
ctargett, line 246: The removed section introduces the `_nest_path_` field which is shown in the example that has been retained. However the change here talks about the field requirements for the `_root_` field. That leaves the `_nest_path_` field unaddressed and makes the following example seem like it's coming out of nowhere. This gets a bit more confusing as the section goes on, with a later sub-bullet that talks about the `_nest_path_` field which hasn't been discussed really up to that point (the example is literally the first time in the doc that it's mentioned).

ctargett:
Other than the one section in docs that I commented on, the rest of the doc changes look good - thanks!

solr/solr-ref-guide/src/indexing-nested-documents.adoc
dsmiley: I don't think I broke the flow of introduction.  Perhaps the diff makes this unclear?  Any way, I just submitted an update that improves the whole matter (I hope).  I recommend building the HTML file looking at it in a browser.

ctargett: GitHub also allows one to view the whole file as of the commit, which is what I did so I feel like I did already look at your changes in their broader context. I think you've misunderstood the point of my comment, which is not only about that one change, but how it relates to the child bullets under it (which you didn't change, so maybe overlooked).

The bullet point in question talks about the `_route_` field.

Right under that, indented so it is clearly _part of the same bullet item_, is an example from a schema for the `_nest_path_` field. There is no `_route_` field shown in the example. This example used to have text which introduced the `_nest_path_` field, but that was removed to add the current paragraph about the `_route_` field.

This example is now the very first time the `_nest_path_` field is mentioned on the page. What is that field? How does it relate to the `_route_` field?

Right after this example, there is a child-level bullet that states: "In a nutshell, this fields makes child/nested documents easier to use, and so it’s recommended you add it." Which field? The typo makes it even more confusing - does it mean both fields?

The next child-level bullet again says "This field enables Solr to ..." - again, which field?

A later child-level bullet then says, "If you don't use `_nest_path_` ..." This is only the 2nd time this field is mentioned. What is this field? Why would I use it or not use it?

dsmiley: Thank you for entertaining my request to do a screen share.  We're cleared up now; you were looking at my earlier commit, it seems.

dsmiley:
I quashed the changes, rebased on top of master, added a commit of some minor things, and force pushed.

I plan to commit tonight.

solr/core/src/java/org/apache/solr/handler/component/RealTimeGetComponent.java
```diff
@@ -257,30 +264,37 @@ public void process(ResponseBuilder rb) throws IOException
...
                  assert entry.size() == 5;
                  // For in-place update case, we have obtained the partial document till now. We need to
                  // resolve it to a full document to be returned to the user.
-                 doc = resolveFullDocument(core, idBytes.get(), rsp.getReturnFields(), (SolrInputDocument)entry.get(entry.size()-1), entry, null);
+                 // resolveFullDocument applies the transformer, if present.
+                 doc = resolveFullDocument(core, idBytes.get(), rsp.getReturnFields(), (SolrInputDocument)entry.get(entry.size()-1), entry);
                  if (doc == null) {
                    break; // document has been deleted as the resolve was going on
                  }
+                 doc.visitSelfAndNestedDocs((label, d) -> removeCopyFieldTargets(d, req.getSchema()));
```
cpoerschke, line 293: Wondering about `req.getSchema()` vs. `core.getLatestSchema()` here - https://github.com/apache/solr/pull/980 opened to explore further.

solr/core/src/java/org/apache/solr/handler/component/RealTimeGetComponent.java
```diff
@@ -355,8 +370,9 @@ void processGetInputDocument(ResponseBuilder rb) throws IOException {
 
     String idStr = params.get("getInputDocument", null);
     if (idStr == null) return;
+    BytesRef idBytes = req.getSchema().indexableUniqueKey(idStr);
     AtomicLong version = new AtomicLong();
-    SolrInputDocument doc = getInputDocument(req.getCore(), new BytesRef(idStr), version, null, Resolution.DOC);
+    SolrInputDocument doc = getInputDocument(req.getCore(), idBytes, idBytes, version, null, Resolution.ROOT_WITH_CHILDREN);
```
cpoerschke, line 375: Wondering about `req.getSchema()` vs. `req.getCore().getLatestSchema()` here - https://github.com/apache/solr/pull/980 opened to explore further.

PR #2160, opened 2020-12-22
SOLR-15057: avoid unnecessary object retention in FacetRangeProcessor

cpoerschke:
https://issues.apache.org/jira/browse/SOLR-15057


HoustonPutman (approved):
Looks pretty straightforward!

PR #2162, opened 2020-12-22
SOLR-15051 Blob, DRAFT WIP

dsmiley:
https://issues.apache.org/jira/browse/SOLR-15051

Remember this is very WIP... just getting started here.

CC @bruno-roustant  @NazerkeBS @atris


bruno-roustant:
Replaced by same PR https://github.com/apache/solr/pull/20 on apache/solr.

PR #2163, opened 2020-12-22
LUCENE-9647 Add back github action for Ant

madrob:
https://issues.apache.org/jira/browse/LUCENE-9647

Tagging @tflobbe for review because you removed the file in the first place. If there's a better place to put it, let me know.

### DIFF
--- a/.github/workflows/ant.yml
+++ b/.github/workflows/ant.yml
@@ -1,0 +1,32 @@
+name: Ant precommit
+
+on: 
+  pull_request:
+    branches:
+    - 'branch_8x'
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Set up JDK 1.8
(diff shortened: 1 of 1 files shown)


tflobbe:
Does this go yo master or to 8.x branch? I don't think we had it running in 8.x before?

madrob:
It needs to be in the default branch. That we didn't have this for 8.x before isn't a reason not to add it now.

anshumg (approved):

tflobbe:
> That we didn't have this for 8.x before isn't a reason not to add it now.
+1! didn't mean we shouldn't, I was just referring to the fact that I didn't remove it for a particular reason, it just never existed for this branch.

dweiss:
Why merge it with master though? It does not apply to master - only to 8x?

dweiss:
bq. It needs to be in the default branch. 

Oh, ok. So it needs to be on the default branch even if it applies only to a filtered set of branches? Get it.

PR #2166, opened 2020-12-23
SOLR-15060: Introduce DelegatingDirectoryFactory.

bruno-roustant:
For the description, see https://issues.apache.org/jira/browse/SOLR-15060

This PR adds a new method get(String path, DirContext dirContext, String rawLockType, Function<Directory, Directory> wrappingFunction) in DirectoryFactory, with a new wrappingFunction parameter. The existing get(String path, DirContext dirContext, String rawLockType) calls the new one with Function.identity().

I wonder if this change should be 8x or 9x.

### DIFF
--- a/solr/core/src/java/org/apache/solr/core/CachingDirectoryFactory.java
+++ b/solr/core/src/java/org/apache/solr/core/CachingDirectoryFactory.java
@@ -30,6 +30,7 @@ import java.util.IdentityHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.function.Function;
 
 import org.apache.lucene.store.AlreadyClosedException;
 import org.apache.lucene.store.Directory;
@@ -97,14 +98,6 @@ public abstract class CachingDirectoryFactory extends DirectoryFactory {
   protected Map<Directory, List<CloseListener>> closeListeners = new HashMap<>();
 
   protected Set<CacheValue> removeEntries = new HashSet<>();
-
-  private Double maxWriteMBPerSecFlush;
-
-  private Double maxWriteMBPerSecMerge;
-
-  private Double maxWriteMBPerSecRead;
-
-  private Double maxWriteMBPerSecDefault;
 
   private boolean closed;
 
@@ -339,10 +332,10 @@ public abstract class CachingDirectoryFactory extends DirectoryFactory {
    * (non-Javadoc)
    *
    * @see org.apache.solr.core.DirectoryFactory#get(java.lang.String,
-   * java.lang.String, boolean)
+   * java.lang.String, boolean, java.util.function.Function)
    */
   @Override
-  public final Directory get(String path, DirContext dirContext, String rawLockType)
+  public final Directory get(String path, DirContext dirContext, String rawLockType, Function<Directory, Directory> wrappingFunction)
       throws IOException {
     String fullPath = normalize(path);
     synchronized (this) {
@@ -357,7 +350,7 @@ public abstract class CachingDirectoryFactory extends DirectoryFactory {
       }
 
       if (directory == null) {
-        directory = create(fullPath, createLockFactory(rawLockType), dirContext);
+        directory = wrappingFunction.apply(create(fullPath, createLockFactory(rawLockType), dirContext));
         assert ObjectReleaseTracker.track(directory);
         boolean success = false;
         try {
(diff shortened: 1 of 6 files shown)


solr/core/src/java/org/apache/solr/core/CachingDirectoryFactory.java
```diff
@@ -405,11 +398,6 @@ public void incRef(Directory directory) {
 
   @Override
   public void init(@SuppressWarnings("rawtypes") NamedList args) {
-    maxWriteMBPerSecFlush = (Double) args.get("maxWriteMBPerSecFlush");
-    maxWriteMBPerSecMerge = (Double) args.get("maxWriteMBPerSecMerge");
-    maxWriteMBPerSecRead = (Double) args.get("maxWriteMBPerSecRead");
-    maxWriteMBPerSecDefault = (Double) args.get("maxWriteMBPerSecDefault");
-
```
madrob, line 412: Why is this all deleted?

solr/core/src/java/org/apache/solr/core/DelegatingDirectoryFactory.java
```diff
@@ -0,0 +1,189 @@
...
+import java.io.IOException;
+import java.util.function.Function;
+
+/**
+ * Delegates to a {@link DirectoryFactory} defined by the {@code delegateFactory} parameter.
+ * <p>
+ * Configured with the {@code delegateFactory} parameter which defines the full class name of the delegate
+ * {@link DirectoryFactory}, plus any additional parameter for the delegate factory.</p>
+ */
+public class DelegatingDirectoryFactory extends DirectoryFactory {
```
madrob, line 35: should this be abstract?

solr/core/src/java/org/apache/solr/core/CachingDirectoryFactory.java
```diff
@@ -339,10 +332,10 @@ public boolean exists(String path) throws IOException {
    * (non-Javadoc)
    *
    * @see org.apache.solr.core.DirectoryFactory#get(java.lang.String,
-   * java.lang.String, boolean)
+   * java.lang.String, boolean, java.util.function.Function)
    */
   @Override
-  public final Directory get(String path, DirContext dirContext, String rawLockType)
+  public final Directory get(String path, DirContext dirContext, String rawLockType, Function<Directory, Directory> wrappingFunction)
```
madrob, line 338: is cachingdirectoryfactory supposed to extend delegatingDF now? I'm confused about the relationship here.

solr/core/src/java/org/apache/solr/core/CachingDirectoryFactory.java
bruno-roustant: I simply noticed that those fields were private and actually never used. I think they should be removed for clarity, but that's not necessary for this PR.

solr/core/src/java/org/apache/solr/core/CachingDirectoryFactory.java
bruno-roustant: I don't think CachingDF should extend DelegatingDF. Here the goal with this new param is to allow the caller to have a hook when the Directory is created internally. Otherwise there is no way a DelegatingDF "A" can delegate to any CachingDF "B" while leveraging the cache inside B and at the same time have control over the class of Directory cached and returned.
Another option could be to first extend the CachingDF with a B' class that overrides the createDirectory() method. But that means a DelegatingDF can only delegate to a specific class B' extending CachingDF. But this is too limiting.

solr/core/src/java/org/apache/solr/core/DelegatingDirectoryFactory.java
bruno-roustant: Good point, yes it misses abstract.

solr/core/src/java/org/apache/solr/core/CachingDirectoryFactory.java
bruno-roustant: A sample usage could be a BlobDF extending DelegatingDF to delegate to *any* configured DF, probably MMapDF which is a CachingDF. BlobDF would need to create BlobDir that delegates to a MMapDir. BlobDir cannot wrap the MMapDir outside of the CachingDF, because for the cache to actually work, it has to cache the BlobDir instance. That's why the idea is to provide a wrapper function to wrap a MMapDir with a BlobDir when it is created internally by the CachingDF.

dsmiley:
The PR is about introducing a low level utility that we expect will be useful for custom DirectoryFactory implementations _that we don't even have yet_.  It's the kind of thing that many of us wouldn't of even created a separate PR for -- it'd show up in a larger PR that introduces the first user of the utility.  It's not even worth a CHANGES.txt entry IMO; there's always the commit message.  As long as there is no user of this code yet, I'd prefer that we don't merge this yet.  Perhaps we may find that the abstraction here isn't quite right.

RE master (9) vs 8x... If we wait till BlobDirectory, it'd almost certainly be 9. If we don't, ehh... I'd just do 9.

solr/core/src/java/org/apache/solr/core/CachingDirectoryFactory.java
dsmiley: I looked at these a week ago.  It's lingering left-over stuff from a refactoring McCandless did years ago in where some thresholds are configured.

bruno-roustant:
Actually I'll need that to simplify the code in BlobDirectory. So based on your remark David, I should rather integrate this code in BlobDirectory code and close this Jira issue.

dsmiley:
An in-between is to leave this issue open until the BlobDirectory is much more ready.  It may change in the mean time; who knows.  I do like the notion of committing separate pieces instead of one feature!

PR #2169, opened 2020-12-27
SOLR-14723: Remove the class attribute for the caches in the _default/example configsets

trdillon:
# Description

In 8.x branch, the default and example configsets include LRUCache and FastLRUCache as the implementation for the various caches. Those implementations are deprecated.

# Solution

Remove the attribute completely and let the default take over (as it's in master), this should make upgrading configsets easier.

# Tests

_ant test_

# Checklist

### DIFF
--- a/solr/core/src/test-files/solr/collection1/conf/solrconfig-caching.xml
+++ b/solr/core/src/test-files/solr/collection1/conf/solrconfig-caching.xml
@@ -21,16 +21,19 @@
   <schemaFactory class="ClassicIndexSchemaFactory"/>
   <query>
     <cache name="lfuCacheDecayFalse"
+           class="solr.search.LFUCache"
            size="10"
            initialSize="9"
            timeDecay="false" />
 
     <cache name="lfuCacheDecayTrue"
+           class="solr.search.LFUCache"
            size="10"
            initialSize="9"
            timeDecay="true" />
 
     <cache name="lfuCacheDecayDefault"
+           class="solr.search.LFUCache"
            size="10"
            initialSize="9" />
   </query>
--- a/solr/core/src/test-files/solr/collection1/conf/solrconfig-caching.xml
+++ b/solr/core/src/test-files/solr/collection1/conf/solrconfig-caching.xml
@@ -21,19 +21,16 @@
   <schemaFactory class="ClassicIndexSchemaFactory"/>
   <query>
     <cache name="lfuCacheDecayFalse"
-           class="solr.search.LFUCache"
            size="10"
            initialSize="9"
            timeDecay="false" />
 
     <cache name="lfuCacheDecayTrue"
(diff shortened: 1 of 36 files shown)


solr/core/src/test-files/solr/collection1/conf/solrconfig-caching.xml
```diff
@@ -21,19 +21,16 @@
   <schemaFactory class="ClassicIndexSchemaFactory"/>
   <query>
     <cache name="lfuCacheDecayFalse"
```
tflobbe, line 23: I think we shouldn't be changing this file. It's specific for the LFUCache testing, so we should leave it as is.

solr/server/solr/configsets/_default/conf/solrconfig.xml
```diff
@@ -416,8 +416,7 @@
                       to occupy. Note that when this option is specified, the size
                       and initialSize parameters are ignored.
       -->
```
tflobbe, line 418: Could you fix this docs to state that the default cache class is `CaffeineCache`?

tflobbe (commented):
This looks great, just the two comments

tflobbe:
For `TestSolr4Spatial2`, I think you just need to add [this line](https://github.com/apache/lucene-solr/blob/master/solr/core/src/test/org/apache/solr/search/TestSolr4Spatial2.java#L323)

solr/core/src/test-files/solr/collection1/conf/solrconfig-caching.xml
trdillon: Ok, I reverted the changes

trdillon:
> 
> 
> For `TestSolr4Spatial2`, I think you just need to add [this line](https://github.com/apache/lucene-solr/blob/master/solr/core/src/test/org/apache/solr/search/TestSolr4Spatial2.java#L323)

Should I make changes to this class? I was working off `branch_8x` but seems those changes, along with some others, were made in `master` but not `branch_8x`.

tflobbe:
> Should I make changes to this class?

Looks to me that this change will be needed given the changes in this PR, right? this test is failing for me in your branch

tflobbe (approved):
Thanks, Changes LGTM. Could you add an entry in the solr/CHANGES.txt (in the 8.8 section)

tflobbe:
Merged to 8.x. Not needed in master

solr/core/src/test-files/solr/collection1/conf/solrconfig-caching.xml
tflobbe: @trdillon, I created https://github.com/apache/lucene-solr/pull/2174, to remove this file completely from master, it's not being used since the deprecated cache implementations and their tests have been removed.

PR #2170, opened 2020-12-27
LUCENE-9648: Add a numbers query in sandbox that takes advantage of index sorting

gf2121:
# Description

In [LUCENE-6539](https://issues.apache.org/jira/browse/LUCENE-6539), we introduced DocValuesNumbersQuery to benefit some special cases (many terms/numbers and fewish matching hits). Inspired by [LUCENE-7714](https://issues.apache.org/jira/browse/LUCENE-7714), maybe we can speed it up when meeting index sort since we can do a binary search to bound the docvalues before iter it. 

I'd like to raise another similary issue to add an `IndexSortDocValuesTermsQuery` if this can pass the review :)

# Benchmarks

I constructed some data by the following script:
```
Directory directory = FSDirectory.open(Paths.get("./test"));
IndexWriter indexWriter = new IndexWriter(directory,
        new IndexWriterConfig()
                .setOpenMode(IndexWriterConfig.OpenMode.CREATE_OR_APPEND)
                .setIndexSort(
                        new Sort(new SortedNumericSortField("a", SortField.Type.LONG))
                ));
for (int i = 0; i < TOTAL_DOC; i++) {
    int a = random.nextInt(1000000000);
    Document document = new Document();
    document.add(new LongPoint("a", a));
    document.add(new SortedNumericDocValuesField("a", a));
    indexWriter.addDocument(document);
}

indexWriter.flush();
indexWriter.commit();
indexWriter.close();
```
And search them like this pattern:
```
@Benchmark
@BenchmarkMode({Mode.AverageTime})
@Fork(1)
@Measurement(iterations = 30)
@Warmup(iterations = 10)
public static void baseline(Blackhole blackhole) throws IOException {
    for (int i = 0; i < termsLen; i++) {
        terms[i] = random.nextInt(1000000000);
    }
    Query query = createQuery("a", terms);
    AtomicInteger res = new AtomicInteger(0);
    indexSearcher.search(query, new SimpleCollector() {
        @Override
        public void collect(int doc) throws IOException {
            res.set(doc ^ res.get());
        }

        @Override
            return ScoreMode.COMPLETE_NO_SCORES;
        }
    });
    blackhole.consume(res.get());
}
```
**Note**: the random seed search used is the same with writing, so each long in the query can hit at least one doc.

## Results:

**Based on 50,000,000 doc:**
LongsCount | PointInSetQuery(s/op) | IndexSortDocValuesNumbersQuery(s/op) | DocValuesNumbersQuery(s/op)
-- | -- | -- | --
1000 | 0.144 ± 0.001 | 0.428 ± 0.008 | 1.159 ± 0.031
5000 | 0.404 ± 0.003 | 0.424 ± 0.008 | 1.343 ± 0.022
10000 | 0.661 ± 0.012 | 0.436 ± 0.008 | 1.372 ± 0.011
50000 | 1.319 ± 0.011 | 0.450 ± 0.005 | 1.173 ± 0.017
100000 | 1.328 ± 0.007 | 0.459 ± 0.003 | 1.160 ± 0.098

**Based on 100,000,000 doc:**
LongsCount | PointInSetQuery(s/op) | IndexSortDocValuesNumbersQuery(s/op) | DocValuesNumbersQuery(s/op)
-- | -- | -- | --
1000 | 0.189 ± 0.003 | 0.860 ± 0.028 | 2.408 ± 0.087
5000 | 0.715 ± 0.010 | 0.879 ± 0.010 | 2.642 ± 0.105
10000 | 1.075 ± 0.034 | 0.859 ± 0.010 | 2.629 ± 0.077
50000 | 2.652 ± 0.114 | 0.865 ± 0.009 | 2.737 ± 0.099
100000 | 2.905 ± 0.027 | 0.881 ± 0.005 | 2.820 ± 0.474


lucene/sandbox/src/java/org/apache/lucene/sandbox/search/IndexSortDocValuesNumbersQuery.java
```diff
@@ -0,0 +1,275 @@
...
+    long last = tmp.advanceExact(maxDoc - 1) ? tmp.longValue() : miss;
+    long min = sortField.getReverse() ? last : first;
+    long max = sortField.getReverse() ? first : last;
+    long lowerValue = values[0];
+    long upperValue = values[values.length - 1];
+    // different from range query, terms query's result docs can be thin
+    // terms: 1, 10, 11, 100
+    // doc values: [2, 99]
+    // the range can be [10, 11] instead of [1, 100] with following loops
+    for (int i = 0; i < values.length; i++) {
```
madrob, line 256: Does this optimization work in the reverse sort case? We check for it above, but then still do the same sided comparison here.

gf2121: Thanks for your review @madrob ! 
`values` is an array of the query numbers. I always sort it with `Arrays.sort(this.values)` so it can not get into a reverse sort case. We checked the `sortField.getReverse()` above because docvalues can be sorted reversely.

PR #2175, opened 2020-12-31
LUCENE-9652: DataInput.readFloats for use by Lucene90VectorReader

msokolov:
This adds `DataInput.readFloats`  and makes use of it in `Lucene90VectorReader`. The implementation and tests are essentially cloned from `readLELongs`. With these changes I observed definite speedups, although not as much as I expected from microbenchmarking. There's a fair amount of variability, but I see at least 20%, sometimes as much as 40% reduction in time for HnswGraph.search, as measured by KnnGraphTester.


uschindler:
IMHO, as this is little endian, we should name the new method like the long one.
After we have changed everything to little endian we may change nsming, but as of now little endian is different to standard behavior, so it should be explicit.

msokolov:
> IMHO, as this is little endian, we should name the new method like the long one.
Fair enough; I'll post an update soon.

uschindler:
FYI, I added some interface in oal.util.Unwrapable that can be implemented by our mocking layers in test-framework. The new interface allows us to unwrap the `Path` without knowing about test-framework internals. IMHO, we should use this also for other mocking layers, so we can easy unwrap them with some consistent API.

The remaining 2 issues in `MappedByteBuffer.mapFile()` were merged today and as soon as those are part of another preview build of JDK 16, we can remove the `mapFileBugfix()` method.

PR #2177, opened 2021-01-05
SOLR-15052: Per-replica states for reducing overseer bottlenecks (trunk)

noblepaul:
WIP


madrob:
Please note that there are still a lot of outstanding comments on the 8x version of this PR, I worry that you are going to end up doing double work trying to maintain two patch sets for a time.

noblepaul:
> Please note that there are still a lot of outstanding comments on the 8x version of this PR, I worry that you are going to end up doing double work trying to maintain two patch sets for a time.

Unfortunately, our 8x and trunk have diverged a lot and we cannot apply any patch across branches. We have to build them in parallel

PR #2179, opened 2021-01-05
SOLR-15055 Re-implement 'withCollection' and 'maxShardsPerNode'

sigram:
See Jira for details.


solr/solrj/src/java/org/apache/solr/client/solrj/request/CollectionAdminRequest.java
```diff
@@ -59,13 +59,8 @@
...
 import static org.apache.solr.common.cloud.ZkStateReader.REPLICATION_FACTOR;
 import static org.apache.solr.common.cloud.ZkStateReader.TLOG_REPLICAS;
-import static org.apache.solr.common.params.CollectionAdminParams.ALIAS;
-import static org.apache.solr.common.params.CollectionAdminParams.COLL_CONF;
-import static org.apache.solr.common.params.CollectionAdminParams.COUNT_PROP;
-import static org.apache.solr.common.params.CollectionAdminParams.CREATE_NODE_SET_PARAM;
-import static org.apache.solr.common.params.CollectionAdminParams.CREATE_NODE_SET_SHUFFLE_PARAM;
-import static org.apache.solr.common.params.CollectionAdminParams.ROUTER_PREFIX;
-import static org.apache.solr.common.params.CollectionAdminParams.SKIP_NODE_ASSIGNMENT;
+import static org.apache.solr.common.params.CollectionAdminParams.*;
```
madrob, line 62: please don't use wildcard imports

solr/solrj/src/java/org/apache/solr/common/cloud/ReplicaPosition.java
```diff
@@ -19,17 +19,20 @@
 
 
 public class ReplicaPosition implements Comparable<ReplicaPosition> {
```
murblanc, line 21: It's not from this PR, but the text `Note: this class has a natural ordering that is inconsistent with equals.` should have been added to the class Javadoc that should have been here.

solr/solrj/src/java/org/apache/solr/common/params/CollectionAdminParams.java
```diff
@@ -80,6 +80,17 @@
    */
   String COLL_CONF = "collection.configName";
 
+  /**
+   * The name of the collection with which a collection is to be co-located
```
murblanc, line 84: It would be helpful to explicit here what this really means and assumes.

solr/core/src/java/org/apache/solr/cloud/api/collections/CreateCollectionCmd.java
```diff
@@ -116,6 +121,25 @@ public void call(ClusterState clusterState, ZkNodeProps message, @SuppressWarnin
...
+    if (withCollection != null) {
+      String realWithCollection = aliases.resolveSimpleAlias(withCollection);
+      if (!clusterState.hasCollection(realWithCollection)) {
+        throw new SolrException(ErrorCode.BAD_REQUEST, "The 'withCollection' does not exist: " + realWithCollection);
+      } else  {
+        DocCollection collection = clusterState.getCollection(realWithCollection);
+        if (collection.getActiveSlices().size() > 1)  {
+          throw new SolrException(ErrorCode.BAD_REQUEST, "The `withCollection` must have only one shard, found: " + collection.getActiveSlices().size());
+        }
+        withCollectionShard = collection.getActiveSlices().iterator().next().getName();
```
murblanc, line 135: Assigned but never used?

solr/core/src/java/org/apache/solr/cloud/api/collections/CreateCollectionCmd.java
```diff
@@ -116,6 +121,25 @@ public void call(ClusterState clusterState, ZkNodeProps message, @SuppressWarnin
...
 
+    String withCollection = message.getStr(WITH_COLLECTION);
+    String withCollectionShard = null;
+    if (withCollection != null) {
+      String realWithCollection = aliases.resolveSimpleAlias(withCollection);
+      if (!clusterState.hasCollection(realWithCollection)) {
+        throw new SolrException(ErrorCode.BAD_REQUEST, "The 'withCollection' does not exist: " + realWithCollection);
+      } else  {
+        DocCollection collection = clusterState.getCollection(realWithCollection);
+        if (collection.getActiveSlices().size() > 1)  {
```
murblanc, line 132: Maybe test !=1 to make sure it really has at least one? Friendlier to fail with the `SolrException` here than below with `next()` throwing `NoSuchElementException`.

solr/core/src/java/org/apache/solr/cloud/api/collections/CreateCollectionCmd.java
```diff
@@ -116,6 +121,25 @@ public void call(ClusterState clusterState, ZkNodeProps message, @SuppressWarnin
       throw new SolrException(SolrException.ErrorCode.BAD_REQUEST, "collection alias already exists: " + collectionName);
     }
 
+    String withCollection = message.getStr(WITH_COLLECTION);
+    String withCollectionShard = null;
+    if (withCollection != null) {
+      String realWithCollection = aliases.resolveSimpleAlias(withCollection);
```
murblanc, line 127: Shouldn't this only be done when message says `followAliases` is `true`?

solr/core/src/java/org/apache/solr/cluster/placement/plugins/AffinityPlacementFactory.java
```diff
@@ -193,6 +195,22 @@ public PlacementPlan computePlacement(Cluster cluster, PlacementRequest request,
       Set<Node> nodes = request.getTargetNodes();
       SolrCollection solrCollection = request.getCollection();
 
+      final SolrCollection secondaryCollection;
+      String withCollection = solrCollection.getCustomProperty(CollectionAdminParams.WITH_COLLECTION);
```
murblanc, line 199: `withCollection` variable here will be the property configured on the collection, it will not resolve the alias if there's one. I believe we need to use params from the placement request to decide if we try to resolve?

murblanc: Taking this back, ignore previous comments. We resolved alias at collection creation time so expecting real collection name here.

solr/core/src/java/org/apache/solr/cluster/placement/plugins/AffinityPlacementFactory.java
```diff
@@ -242,6 +260,37 @@ public PlacementPlan computePlacement(Cluster cluster, PlacementRequest request,
         }
       }
 
+      if (secondaryCollection != null) {
```
murblanc, line 263: Can you extract the if bloc to a separate method?

solr/core/src/java/org/apache/solr/cluster/placement/plugins/AffinityPlacementFactory.java
```diff
@@ -242,6 +260,37 @@ public PlacementPlan computePlacement(Cluster cluster, PlacementRequest request,
...
+        // 2nd phase to allocate required secondary collection replicas
+        Set<Node> secondaryNodes = new HashSet<>();
+        Shard shard1 = secondaryCollection.iterator().next();
+        shard1.replicas().forEach(r -> {
+          secondaryNodes.add(r.getNode());
+        });
+        Set<ReplicaPlacement> secondaryPlacements = new HashSet<>();
+        Set<Node> alreadyAdded = new HashSet<>();
```
murblanc, line 271: Can add the nodes directly into `secondaryNodes` and save introducing `alreadyAdded`, given the actual placement decisions are captures elsewhere in `secondaryPlacements`.

solr/core/src/java/org/apache/solr/cluster/placement/plugins/AffinityPlacementFactory.java
murblanc: Also, comment on that method that the approach is a rather simplistic one, since the set of nodes chosen for the placement request does not take into account the existing `secondaryCollection` distribution, possibly leading to the creation of replicas for that collection that could have been avoided.
I assume clusters where such `withCollection` tricks are done can restrict placement of both collections to a set of nodes to better control what's going on...

solr/solrj/src/java/org/apache/solr/common/params/CollectionAdminParams.java
murblanc: Something along "Collection A with the attribute withCollection collection B implies two things: Collection B has only one shard, and on every node where collection A has a replica, there will also be a replica of collection B's shard".

solr/core/src/java/org/apache/solr/cloud/api/collections/CreateCollectionCmd.java
```diff
@@ -254,8 +280,27 @@ public void call(ClusterState clusterState, ZkNodeProps message, @SuppressWarnin
...
+            continue;
+          }
+          ZkNodeProps props = new ZkNodeProps(
+              Overseer.QUEUE_OPERATION, ADDREPLICA.toString(),
+              ZkStateReader.COLLECTION_PROP, replicaPosition.collection,
+              ZkStateReader.SHARD_ID_PROP, replicaPosition.shard,
+              "node", replicaPosition.node,
+              CommonAdminParams.WAIT_FOR_FINAL_STATE, Boolean.TRUE.toString()); // set to true because we want `withCollection` to be ready after this collection is created
+          new AddReplicaCmd(ocmh).call(clusterState, props, results);
+          clusterState = zkStateReader.getClusterState(); // refresh
```
murblanc, line 297: You know this call does absolutely nothing, right?

Good news is there's not need to refresh, not here and not in the other places where it's done.

When Collection API calls wait for something, they repeatedly read the in memory cluster state (maintained by ZkStateReader) until they observe the desired change. When they do it means that 1. the Overseer applied the change in ZK (directly as a cluster state change or as a side effect of a Collection API call itself doing a cluster state change as is the case here), and 2. the zookeeper watches that this node has on its cluster state did fire and the update from Zookeeper was applied to the cluster state.
It does happen that currently Collection API commands run on the Overseer node where the cluster state updater runs, but really the two are independent and could well be running on different nodes.

solr/solrj/src/java/org/apache/solr/common/params/CollectionAdminParams.java
murblanc: And likely need to specify (here or elsewhere) that this is done for collection creation and adding replicas for Collection A, but not moving them...

solr/core/src/java/org/apache/solr/cloud/api/collections/CreateCollectionCmd.java
murblanc: Or is the collection created with the `withCollection` property that is assumed to be an actual collection?
Wouldn't this somehow defeats the purpose of an alias? What prevents having the collection `withCollection` property follow aliases? Follow always? Or not always? The current proposal follows always given it doesn't check `followAliases`. In which case maybe resolving the alias each time replicas are added to the collection makes sense? Let the placement plugin do the failure, it's checking the `withCollection` collection is appropriate anyway.

murblanc:
**Blocking issue**: `CreateCollectionCmd` is not the only place where `withCollection` logic is going to get applied. All calls that end up calling the placement plugins now need to deal with returned placement for multiple collections: `SplitShardCmd`, `AddReplicaCmd`, `ReplaceNodeCmd`, `RestoreCmd`. Or at least ignore the additional replica assignments being returned.

I also see `CollectionsRepairEventListener`. We will also eventually have cluster rebalancers, will we expect them to support `withCollection`?

Dealing with replica placement for multiple collections returned from the placement plugins seems to deserve to be made a "first class" abstraction if we want to implement this (unreasonable?) feature in a reasonable way.

solr/solrj/src/java/org/apache/solr/common/params/CollectionAdminParams.java
sigram: At this stage we're not fully sure yet how far this feature will go in ensuring this constraint, and where it will be implemented... This flag here just for indicating the constraint, implementation notes should go to the plugins that support this flag.

solr/core/src/java/org/apache/solr/cloud/api/collections/CreateCollectionCmd.java
sigram: Frankly no idea :) that's the way it was done in 8x, there are probably edge cases either way. I'll add a note for now.

solr/core/src/java/org/apache/solr/cluster/placement/plugins/AffinityPlacementFactory.java
sigram: I think that we should probably completely disallow using aliases in this context... how would that work in practice? if you flip the alias that used to point to the secondary collection to some other collection then suddenly the co-location may be completely broken.

solr/core/src/java/org/apache/solr/cloud/api/collections/CreateCollectionCmd.java
sigram: It's a copy-pasta from 8x... I wasn't sure if it's required so it's good to know it doesn't make sense and can be removed.

solr/solrj/src/java/org/apache/solr/common/params/CollectionAdminParams.java
HoustonPutman: I think it's valid to specify and document what this flag does by default. Obviously plugins should add documentation on additional logic that they use for these flags as well.

solr/core/src/java/org/apache/solr/cloud/api/collections/CreateCollectionCmd.java
murblanc: We should not follow aliases at all and specify that the `withCollection` must be an existing real collection name.

sigram:
This approach has been abandoned in favor of PR 2199.

ljak:
Hi,

I know it's an old thread but I have a question.

As far as I can tell (after searching), the `maxShardsPerNode` function wasn't re-implemented right (in the new autoscaling framework) ?

Thanks!

PR #2181, opened 2021-01-05
SOLR-15069 child parent filter

dsmiley:
https://issues.apache.org/jira/browse/SOLR-15069
Repeated issue description:
The `parentFilter` param to `[child]` was once required, and then made only optional if you have a `_nest_path_` field, but I think it can be optional (perhaps obsolete) altogether.  All we have to do is grab the uniqueKey field from the document to be transformed, look this up in the `_root_` field, and you'll get the internal Lucene docId of the first child doc.  No problem, and quick.

CC @moshebla  @thomaswoeckinger

I came about doing this because I too-soon documented in #2159 that `_nest_path_` field was no longer required for doing atomic/partial updates in a nested schema.  Instead of removing the documentation I wrote, and then realizing I would then need to add safety prevention (otherwise child docs could vanish on updates!)... it occurred to me, lets just make it work :-)

### DIFF
--- a/solr/CHANGES.txt
+++ b/solr/CHANGES.txt
@@ -236,6 +236,9 @@ Improvements
 
 * SOLR-15062: /api/cluster/zk/ls should give the stat of the current node (noble)
 
+* SOLR-15069: [child]: the parentFilter parameter is now fully optional and perhaps obsolete.
+  (David Smiley)
+
 Optimizations
 ---------------------
 * SOLR-14975: Optimize CoreContainer.getAllCoreNames, getLoadedCoreNames and getCoreDescriptors. (Bruno Roustant)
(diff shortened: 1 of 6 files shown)


dsmiley:
CC @barrotsteindev   (anyone CC'ed or really anyone, I'd appreciate a review)

barrotsteindev:
Seems like a great idea.
I think `parentFilter` still has its place when you want to return a part of a hierarchy where the parent documents are not the root docs(top level parents).
I guess the `parentFilter` could still be computed using the \_nest_path\_ field, but I guess this is a specialized case and it might be better to let the user supply the `parentFilter` himself.

dsmiley:
I was thinking that maybe but I **suspect** (just a theory) the current code "just works" because it tracks the path information already and builds the tree.  It'd just be a sub-tree with docs that don't connect, and I think they would just be dropped/ignored.  It'd be nice to have a test for this to find out for sure.  Even if it does work, it could be better optimized for this case.

PR #2183, opened 2021-01-06
SOLR-15070: Remove HashMap usage in SuggestComponent rsp

gerlowskija:
# Description

The HashMap added to the response by SuggestComponent is handled differently by each output format/wt/ResponseParser that Solr supports.  This discrepancy causes SolrJ produce different NamedList response structures when different wt's are used.  This leads to ClassCastException's elsewhere in SolrJ, particularly when QueryResponse attempts to parse responses which include "suggest" info.

# Solution

This PR replaces the HashMap in SuggestComponent with a NamedList - ensuring that the NamedList SolrJ produces on the client side is the same regardless of the wire format in use.

# Tests

Manual testing via SolrJ snippets, as well as a revamping of TestSuggesterResponse to use more than just BinaryResponseParser.

# Checklist


madrob:
would using another SimpleOrderedMap (which extends NamedList) be better than a NamedList directly?

gerlowskija:
> would using another SimpleOrderedMap (which extends NamedList) be better than a NamedList directly?

Maybe?  I'll admit I'm not as familiar as I should be with the differences between the two and when each one is preferable.  I'll look into it!

gerlowskija:
It looks like SimpleOrderedMap is preferred in the sort of serialization/deserialization usecase involved here, so I've updated the PR to use it instead as Mike suggested.

This changes the object used on the server side - the client side already had SimpleOrderedMap objects by virtue of the deserialization implementations themselves (which this PR doesn't touch).

anshumg (approved):

madrob:
If the client side doesn't change, then does that mean this is safe to back port to 8x?

Also, looking at the code, CluteringInfo might be similarly affected, would you have any interest in validating that?

madrob (approved):

gerlowskija:
> If the client side doesn't change, then does that mean this is safe to back port to 8x?

Sorry, I think a comment I made above was misleading.  This PR as a whole will change the deserialized NamedList response structures that SolrJ users get back.  Which makes this approach tough to backport as-is.

My comment above that starts "This changes the object used on the server side [only] ..." was intended to address in advance a question I thought you might ask following the switch to SimpleOrderedMap.  ("Hey I still see the `NamedList` type in QueryResponse, why didn't the Suggest code change there as well?").  But in hindsight I see why that comment reads like I'm saying this is backwards-compatible.  Sorry for the confusion.

To your other question, I might have interest in looking into ClusterInfo.  But if so, I'll probably punt it to a separate PR so that I can wait for feedback on a [mailing list thread](http://mail-archives.apache.org/mod_mbox/lucene-dev/202101.mbox/%3CCAPCX2-%2BuGBgagBWPeD8wT%3Dud9p3_cbWxW77vH1RPZqaNVpZXGA%40mail.gmail.com%3E) that relates to all this (without holding up the Suggest code here).

PR #2189, opened 2021-01-08
SOLR-14413 fix unit test to use delayed handler

madrob:


solr/core/src/test/org/apache/solr/CursorPagingTest.java
```diff
@@ -521,6 +522,7 @@ public void testTimeAllowed() throws Exception {
         "fl", "id",
         "sort", "id asc",
         "rows", "50",
+        "qt", "/delayed",
```
slackhappy, line 525: ah! wasn't sure how to do this, `qt`!

slackhappy:
thanks for making these changes, certainly an improvement!

PR #2191, opened 2021-01-08
Add gradle settings for github actions.

HoustonPutman:
# Description

The github precommit action is failing because of OOMs. It should be fixed by setting local settings before running gradle commands.


madrob (approved):

anshumg (approved):

HoustonPutman:
Not sure why it cause docker tests to fail, but oh well. Works fine for the SolrJ and Gradle Precommit.

PR #2196, opened 2021-01-11
SOLR-15071: Fix ArrayIndexOutOfBoundsException in contrib/ltr SolrFeatureScorer

cpoerschke:
https://issues.apache.org/jira/browse/SOLR-15071


dsmiley:
Can you please include a simple test that exposes the problem?
It doesn't feel right for this Filter implementation to not delegate when the other methods do.  Note that a plausible alternative is to simply not specify it and thus inherit the null-returning implementation from the Scorer.  Still...
Admittedly I don't quite "get" the problem yet but once a failing test is in front of me, and with your analysis already done, I'm sure it won't take long.

cpoerschke:
> ... Note that a plausible alternative is to simply not specify it and thus inherit the null-returning implementation from the Scorer. ...

Good idea. Done.

> Can you please include a simple test that exposes the problem? ...

I've now included in the branch (and then reverted) the scripted version of the steps-to-reproduce with the .json files from the JIRA plus selective debug logging. It does not reliably reproduce with the script, about 2-of-10 or 3-of-11 runs hit the issue. Here's the debug logging output:

```
org.apache.lucene.search.MinShouldMatchSumScorer@5127aa0a cpoerschke debug: constructor freq = 6
org.apache.lucene.search.MinShouldMatchSumScorer@5127aa0a cpoerschke debug: constructor minShouldMatch = 5
org.apache.lucene.search.MinShouldMatchSumScorer@5127aa0a cpoerschke debug: setDocAndFreq caller = TwoPhaseIterator.advance
org.apache.lucene.search.MinShouldMatchSumScorer@5127aa0a cpoerschke debug: score
org.apache.lucene.search.MinShouldMatchSumScorer@5127aa0a cpoerschke debug: updateFreq freq = 1
org.apache.lucene.search.MinShouldMatchSumScorer@5127aa0a cpoerschke debug: updateFreq minShouldMatch = 5
org.apache.lucene.search.MinShouldMatchSumScorer@5127aa0a cpoerschke debug: updateFreq (freq >= minShouldMatch) = false
```

Note that `minShouldMatch = 5` matches the `mm=5` in the feature and that the `q` and `efi.term` is `"husa%20cu%20tastatura%20tableta%2010%20inch` which is `"husa cu tastatura tableta 10 inch"` i.e. 6 terms matching the `freq = 6`.

solr/contrib/ltr/src/java/org/apache/solr/ltr/feature/Feature.java
```diff
@@ -365,11 +364,6 @@ public DocIdSetIterator iterator() {
         return in.iterator();
       }
 
-      @Override
```
dsmiley, line 368: @cpoerschke  Let's add a comment saying that we intentionally don't delegate twoPhaseIterator because it doesn't work, and that we don't know why, with a reference to the JIRA issue?
Other than that, let's commit this to protect users as-is.

dsmiley (approved):

PR #2197, opened 2021-01-12
SOLR-15075: Solr docker gradle improvements

HoustonPutman:
https://issues.apache.org/jira/browse/SOLR-15075

(From JIRA)

Currently Solr uses the gradle docker plugin to manage building Solr docker images. When migrating the docker build into Solr, using the plugin was the path of least resistance and allowed us to migrate without having to think a lot about the gradle part.

Now that the docker image building works, it may be beneficial to migrate away from the docker plugin, so that we can have better control over build process. The steps are simple enough anyways, so we wouldn't be sacrificing much to have more flexibility.

Given the discussion in the community over release vs local docker image builds, there does not seem to be a desire to have separate builds for them. Therefore we no longer need the solr/docker/package module. Instead the solr/docker module will pull in the solr/packaging output and directly put it into the docker image.

Goals
- [x] Remove Gradle docker plugin
- [x] Remove docker-package intermediate image
- [x] Move docker gradle help
- [x] Make Gradle tasks respect changes to inputs


gradle/help.gradle
```diff
@@ -30,7 +30,7 @@ configure(rootProject) {
       ["Git", "help/git.txt", "Git assistance and guides."],
       ["ValidateLogCalls", "help/validateLogCalls.txt", "How to use logging calls efficiently."],
       ["IDEs", "help/IDEs.txt", "IDE support."],
-      ["Docker", "help/docker.txt", "Building Solr Docker images."],
```
dweiss, line 33: Do you think it makes sense to scatter those help files? I see some benefits of keeping them in one place.

solr/docker/build.gradle
```diff
@@ -18,106 +18,187 @@
...
+
+task testDocker(dependsOn: tasks.dockerBuild) {
+  group = 'Docker'
+  description = 'Test Solr docker image'
+
+  def inputDir = "tests/cases"
+  def outputDir = "$buildDir/tmp/tests"
+
+  // Ensure that the docker image is re-tested if the image ID changes or the test files change
+  inputs.properties([
```
dweiss, line 164: Does this work with complex types (if they're serializable)? I thought they need to be properties (string key-value pairs).

gradle/help.gradle
HoustonPutman: I was moving based off of a comment from @dsmiley. I could go either way. I do think it makes sense that the documentation for the docker module resides in the docker module. The other help files in that directory seem to be applicable project-wide, not for a specific module.

solr/docker/build.gradle
HoustonPutman: The type signature of the method takes a Map<String, Object>. But it'd be trivial to use a string. I'll switch just to be safe.

gradle/help.gradle
dweiss: Ok, fine.

solr/docker/build.gradle
dweiss: I think Gradle tries to serialize these for caches, hence the question. I don't know myself - that's why I asked.

HoustonPutman: Found the [relevant docs](https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/TaskInputs.html#property-java.lang.String-java.lang.Object-). I think having a set should be fine, since it is serializable and has a good `.equals()` method.

dweiss: Thanks. I thought they'd require a serializable type. It's a pity they don't just say ? extends Serializable in the signature.

dsmiley:
I'm glad you payed attention to task inputs/outputs.  I have shelved changes to this gradle build to improve this because changes to the Dockerfile weren't seen as changing the output! (it's sort of the most important input).  I can now remove my WIP because you've overhauled the whole thing, which I'm glad to see :-)

solr/docker/build.gradle
```diff
@@ -18,106 +18,203 @@
...
+  // Ensure that the docker image is rebuilt on build-arg changes or changes in the docker context
+  inputs.properties([
+          baseDockerImage: baseDockerImage,
+          githubUrlOrMirror: githubUrlOrMirror,
+          version: version
+  ])
+  inputs.dir(dockerBuildDistribution)
+
+  // Docker build must run after the docker context has been created & tarred
+  mustRunAfter tasks.dockerTar
```
madrob, line 85: is this redundant with the dependsOn declaration above? I'm surely missing some subtlety.

HoustonPutman: It's not neccessary right now. I mainly put it in because I'm contemplating removing the dependency at some point (I actually did add it, but then removed it without removing the `mustRunAfter` lines.

The issue now is that if you run:

```
gradlew dockerBuild -Psolr.docker.baseImage=something-else:latest
gradlew dockerTag -P solr.docker.imageName=custom-name:latest
```

In the above example the first command will build the dockerImage with the custom baseImage. However the second command doesn't have the `baseImage` specified, so in the dependent `dockerBuild` task it will build the docker image with the default baseImage, overwriting the image created in the first command.

```
gradlew dockerBuild dockerTag -Psolr.docker.baseImage=something-else:latest -P solr.docker.imageName=custom-name:latest
```

This is the command necessary to build & tag with a custom baseImage.

I think it might be nice to make `testDocker` and `dockerTag` not dependent on `dockerBuild`, but instead merely check to make sure that `dockerBuild` has already been run. (The resulting image-id should be stored in a file in the build directory). That way the two commands above work exactly the same.

I'm not convinced either way yet, just something I've been thinking about.

dweiss: It looks like a dependency to me, not a precondition - the behavior you described (dockerTag rebuilding the image in the second command) makes sense to me. If there is more than one option controlling something, you should just specify them (second command) and not rely on any previous build commands.

This is my gradle side viewpoint. I don't know that much about docker to be able to tell whether this makes sense from how these deployment processes are organized.

HoustonPutman: Thanks, I am not very familiar with gradle, so that's exactly the kind of information I need.

We will keep it as is then, and if people complain or it is causing mass confusion, then we can make the necessary changes in the future.

dweiss: That gradle code shows you're familiar enough! mustRunAfter is a rare case, really. 

I sporadically use it to order certain tasks if they're queued together in (possibly) parallel. An example: schedule integration tests to run after unit tests have passed. Then you can run both separately (no dependency on each other) but if you run them together (as part of check, which depends on both integration and unit tests), unit tests will be executed prior to integration tests.

HoustonPutman: Ahh ok, that's good to know.

I have one instance of it left, `dockerPush.mustRunAfter dockerTests` so that if the tests fail, the push doesn't happen. It's not a dependency because I don't want to enforce running the tests, in case a user doesn't want to.

solr/docker/build.gradle
```diff
@@ -18,106 +18,192 @@
...
+
+  dockerImage files(imageIdFile) {
+    builtBy 'dockerBuild'
+  }
 }
 
-tasks.docker {
-  // In order to create the solr docker image, the solr package image must be created first.
-  dependsOn(dockerPackage.tasks.docker)
+task dockerTar(type: Tar) {
```
dsmiley, line 53: I bet this task could be skipped entirely in lieu of the packaging tgz directly being provided as stdin to Docker.  If it doesn't contain these scripts, we could arrange for that easily enough.  I'm toying with the Dockerfile right now that could operate on this assumption.  It's eliminate a tar.gz generation and decompression, and consequently overall simpler.

dsmiley: Oops; I'm mistaken, as it must contain the Dockerfile at the root.

HoustonPutman: Yeah it'd be nice if the build via TGZ were a bit more flexible.

dsmiley: In our container centric world right now... I think it would not be a big deal if our tgz included a Dockerfile at the root.  I'm not sure I want to push for that at the moment (would others hate it?), but throwing it out there.  Re-building *any* official tgz our project produces would be marvelously simple then :-). (curl some official tgz then pipe right into docker build :-)

HoustonPutman: I don't think it's a bad idea. Possibly an effort to push for separately. Might be an issue if we start auto-loading first-party-packages in the docker image in the future. I imagine that those wouldn't be included in the solr tgz, but I might be wrong about that.

dsmiley: The plan for 1P packages does not conflict with this.  They remain where they are, and eventually might only ship metadata in the image instead of all the JARs which would be fetched on demand.  I don't think there's a Docker impact to that.

HoustonPutman: @dsmiley I still think that your idea is something that we should explore. I think that this PR (that I just merged) is just a step in the right direction. But we might as well have something that is working much better now, and add final touches later.

PR #2198, opened 2021-01-12
SOLR-15081: Metrics for core: isLeader, status

dsmiley:
https://issues.apache.org/jira/browse/SOLR-15081
Copying the description: 

> The core level metrics hold some interesting information, but I don't see information pertaining to the SolrCloud status of the core.  In particular, I'd like to see the leader status here, and also the replica state.  The use-case I have in mind is enabling the Prometheus Exporter to get the doc count (and maybe other basics) of only the leader replicas, thereby counting unique documents instead of a fully replicated figure.  This is an approximation to doing a match-all-docs query on all collections, but is a more sound approach when one has orders of magnitude more collections than nodes.

### DIFF
--- a/solr/core/src/java/org/apache/solr/core/SolrCore.java
+++ b/solr/core/src/java/org/apache/solr/core/SolrCore.java
@@ -1203,26 +1203,19 @@ public final class SolrCore implements SolrInfoBean, Closeable {
     parentContext.gauge(() -> isClosed() ? parentContext.nullString() : getIndexDir(), true, "indexDir", Category.CORE.toString());
     parentContext.gauge(() -> isClosed() ? parentContext.nullNumber() : getIndexSize(), true, "sizeInBytes", Category.INDEX.toString());
     parentContext.gauge(() -> isClosed() ? parentContext.nullString() : NumberUtils.readableSize(getIndexSize()), true, "size", Category.INDEX.toString());
-    if (coreContainer != null) {
-      final CloudDescriptor cd = getCoreDescriptor().getCloudDescriptor();
-      if (cd != null) {
-        parentContext.gauge(() -> {
-          if (cd.getCollectionName() != null) {
-            return cd.getCollectionName();
-          } else {
-            return parentContext.nullString();
-          }
-        }, true, "collection", Category.CORE.toString());
 
-        parentContext.gauge(() -> {
-          if (cd.getShardId() != null) {
-            return cd.getShardId();
-          } else {
-            return parentContext.nullString();
-          }
-        }, true, "shard", Category.CORE.toString());
-      }
+    final CloudDescriptor cd = getCoreDescriptor().getCloudDescriptor();
+    if (cd != null) {
+      parentContext.gauge(cd::getCollectionName, true, "collection", Category.CORE.toString());
+      parentContext.gauge(cd::getShardId, true, "shard", Category.CORE.toString());
+      parentContext.gauge(cd::isLeader, true, "isLeader", Category.CORE.toString());
+      parentContext.gauge(
+          () -> String.valueOf(cd.getLastPublished()),
+          true,
+          "replicaState",
+          Category.CORE.toString());
     }
+
     // initialize disk total / free metrics
     Path dataDirPath = Paths.get(dataDir);
     File dataDirFile = dataDirPath.toFile();
--- a/solr/CHANGES.txt
+++ b/solr/CHANGES.txt
@@ -201,13 +201,12 @@ New Features
 
 Improvements
 ---------------------
-
-* SOLR-15079: Block Collapse - Faster collapse code when groups are co-located via Block Join style nested doc indexing.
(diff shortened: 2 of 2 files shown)


solr/core/src/java/org/apache/solr/core/SolrCore.java
```diff
@@ -1202,26 +1203,26 @@ public void initializeMetrics(SolrMetricsContext parentContext, String scope) {
     parentContext.gauge(() -> isClosed() ? parentContext.nullString() : getIndexDir(), true, "indexDir", Category.CORE.toString());
     parentContext.gauge(() -> isClosed() ? parentContext.nullNumber() : getIndexSize(), true, "sizeInBytes", Category.INDEX.toString());
     parentContext.gauge(() -> isClosed() ? parentContext.nullString() : NumberUtils.readableSize(getIndexSize()), true, "size", Category.INDEX.toString());
-    if (coreContainer != null) {
```
megancarey, line 1205: Since we do reference coreContainer on like 1215 should we keep this check?

dsmiley: I removed the check for CoreContainer nullness as it seemed impossible for it to be null.  SolrCores hang off of a CoreContainer.

solr/core/src/java/org/apache/solr/core/SolrCore.java
```diff
@@ -1202,26 +1203,26 @@ public void initializeMetrics(SolrMetricsContext parentContext, String scope) {
...
-            return parentContext.nullString();
-          }
-        }, true, "shard", Category.CORE.toString());
-      }
+    final CloudDescriptor cd = getCoreDescriptor().getCloudDescriptor();
+    if (cd != null) {
+      // TODO
+      parentContext.gauge(cd::getCollectionName, true, "collection", Category.CORE.toString());
+      parentContext.gauge(() -> Objects.requireNonNullElse(cd.getShardId(), parentContext.nullString()), true, "shard", Category.CORE.toString());
+      //TODO should this instead be in a core status, or a metric?  When do we use which?
```
dsmiley, line 1212: I wrote these notes and code months ago, shelved it and nearly forgotten, and today I remembered it and submitted it.  The TODO here was kind of a note-to-self that can be removed.  I'd welcome anyone's thoughts on this though.  There appears, to me, to be overlap in scope between metrics and "status" type requests.  Years ago I thought Metrics was just numbers, but lately I've seen it can have all sorts of strings and basically completely compete with "status".

solr/core/src/java/org/apache/solr/core/SolrCore.java
```diff
@@ -1202,26 +1203,26 @@ public void initializeMetrics(SolrMetricsContext parentContext, String scope) {
...
+      //   SEE org.apache.solr.handler.admin.CoreAdminOperation.getCoreStatus
+      parentContext.gauge(() -> {
+        DocCollection docColl = coreContainer.getZkController().getZkStateReader().getClusterState().getCollectionOrNull(cd.getCollectionName(), true);
```
sigram, line 1215: Maybe use `cloudDescriptor.isLeader()`, it's simpler...

solr/core/src/java/org/apache/solr/core/SolrCore.java
```diff
@@ -1202,26 +1203,26 @@ public void initializeMetrics(SolrMetricsContext parentContext, String scope) {
...
+        Replica leaderReplica = docColl.getLeader(cd.getShardId());
+        return leaderReplica.getName().equals(cd.getCoreNodeName());
+      }, true, "isLeader", Category.CORE.toString());
+      parentContext.gauge(() -> {
+        DocCollection docColl = coreContainer.getZkController().getZkStateReader().getClusterState().getCollectionOrNull(cd.getCollectionName(), true);
+        final Replica myReplica = docColl.getReplica(cd.getCoreNodeName());
+        return Objects.requireNonNullElse(myReplica.getState().toString(), parentContext.nullString());
```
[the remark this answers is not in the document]
sigram: Similarly `cloudDescriptor.getLastPublished()`.

solr/core/src/java/org/apache/solr/core/SolrCore.java
sigram: Yeah, metrics today overlap a lot with "status" requests... something to clean up in 9x.

Initially the metrics API wasn't able to properly report complex values (esp. when reported via JMX) but this has been fixed around 7.0 or so - support for non-numeric values had to be added specifically to report things like paths, non-numeric state, etc. and for complex properties like eg. system properties, caches, etc. Now it can report basically anything you want.

HoustonPutman: I don't think it's inherently wrong to have overlap between status and metrics, as long as the overlapping information is coming from the same source.

PR #2199, opened 2021-01-13
SOLR-15055 (Take 2) Re-implement 'withCollection'

sigram:
See Jira for details.

This is a minimal approach that helps to set-up the initial co-location and then may veto collection layout changes (additions / removals or replicas) if it would violate the `withCollection` constraint.

This PR also refactors the placement API to add support for other types of "collection modification" requests, and extends the placement plugin API to add a method for vetoing such changes. `DeleteReplicaCmd` is also modified to make use of this functionality.


solr/core/src/java/org/apache/solr/cloud/ExclusiveSliceProperty.java
```diff
@@ -74,8 +74,8 @@
   ExclusiveSliceProperty(ClusterState clusterState, ZkNodeProps message) {
     this.clusterState = clusterState;
     String tmp = message.getStr(ZkStateReader.PROPERTY_PROP);
-    if (StringUtils.startsWith(tmp, OverseerCollectionMessageHandler.COLL_PROP_PREFIX) == false) {
-      tmp = OverseerCollectionMessageHandler.COLL_PROP_PREFIX + tmp;
+    if (StringUtils.startsWith(tmp, CollectionAdminParams.PROPERTY_PREFIX) == false) {
```
murblanc, line 77: Minor: use `!` rather than `== false` (I know it's old code but you touched it :)

solr/core/src/java/org/apache/solr/cloud/api/collections/DeleteCollectionCmd.java
```diff
@@ -92,6 +98,19 @@ public void call(ClusterState state, ZkNodeProps message, @SuppressWarnings({"ra
       collection = extCollection;
     }
 
+    PlacementPlugin placementPlugin = ocmh.overseer.getCoreContainer().getPlacementPluginFactory().createPluginInstance();
```
murblanc, line 101: Didn't dig into the details here, but when we delete a collection, we should just check if there's another collection that defines `withCollection` on it and refuse the delete based on that, no?
Just starting to look at the PR so maybe there's a reason for doing it this way (in which case maybe adding a comment?)

solr/core/src/java/org/apache/solr/cloud/api/collections/DeleteReplicaCmd.java
```diff
@@ -147,14 +170,27 @@ void deleteReplicaBasedOnCount(ClusterState clusterState,
       }
     }
 
+    if (placementPlugin != null) {
```
murblanc, line 173: By adding `if (placementPlugin != null)` logic in the `*Cmd` classes, we are breaking the encapsulation that placement logic is handled by `Assign.AssignStrategy`.
The only reason `*Cmd` code currently (before this PR) is even aware of the notion of `PlacementPlugin` is because `PlacementPluginAssignStrategy` configuration (rather than `LegacyAssignStrategy`) is dependent on a plugin being defined...
I suggest to move all `*Cmd` vetting logic to `Assign.AssignStrategy`, so that `*Cmd` only need to pass the instance of `PlacementPlugin` to that code (and later when we finally decide how clusters are configured this will go away, and we keep `*Cmd` clean of any specific assign strategy behavior).

In `Assign.AssignStrategy` the default vetting logic will be "accept all" and in `PlacementPluginAssignStrategy` we can implement the checks we like.

solr/core/src/java/org/apache/solr/cloud/overseer/ReplicaMutator.java
```diff
@@ -115,8 +116,8 @@ public ZkWriteCommand addReplicaProperty(ClusterState clusterState, ZkNodeProps
     String sliceName = message.getStr(ZkStateReader.SHARD_ID_PROP);
     String replicaName = message.getStr(ZkStateReader.REPLICA_PROP);
     String property = message.getStr(ZkStateReader.PROPERTY_PROP).toLowerCase(Locale.ROOT);
-    if (StringUtils.startsWith(property, OverseerCollectionMessageHandler.COLL_PROP_PREFIX) == false) {
-      property = OverseerCollectionMessageHandler.COLL_PROP_PREFIX + property;
+    if (StringUtils.startsWith(property, CollectionAdminParams.PROPERTY_PREFIX) == false) {
```
murblanc, line 119: `== false` -> `!`

solr/core/src/java/org/apache/solr/cluster/placement/DeleteReplicasRequest.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+package org.apache.solr.cluster.placement;
+
+import org.apache.solr.cluster.Replica;
+
+import java.util.Set;
+
+/**
+ *
```
murblanc, line 25: Javadoc needed

solr/core/src/java/org/apache/solr/cluster/placement/DeleteShardsRequest.java
```diff
@@ -0,0 +1,27 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.solr.cluster.placement;
+
+import java.util.Set;
+
+/**
+ *
```
murblanc, line 23: Javadoc needed.

murblanc: Also, this interface is implemented but the implementation is never used.
Unless we implement a use for it in this PR, I suggest we leave it out until we actually need it. I assume we don't need it for `withCollection` because the secondary collection has to be single shard so that shard will not be deleted.

solr/core/src/java/org/apache/solr/cloud/api/collections/DeleteReplicaCmd.java
murblanc: In `Assign.AssignStrategy` if we want to be exhaustive, we should for example reject shard splits for secondary that are targets of `withCollection` (given we refuse such targets to have more than one shard).
Not saying we should do it, but the vetting infra we put in place should allow logical extension to all these aspects (with minor impact on the commands).

Also, pushing all the logic to `Assign.AssignStrategy` and minimizing changes to `*Cmd` limits the impact of a regression (moving to `LegacyAssignStrategy` should be a workaround for most problems).

solr/core/src/java/org/apache/solr/cluster/placement/impl/ModificationRequestImpl.java
```diff
@@ -0,0 +1,106 @@
...
+   * @return
+   */
+  public static DeleteReplicasRequest deleteReplicasRequest(DocCollection docCollection, String shardName, Set<String> replicaNames) {
+    SolrCollection solrCollection = SimpleClusterAbstractionsImpl.SolrCollectionImpl.fromDocCollection(docCollection);
+    Shard shard = solrCollection.getShard(shardName);
+    Slice slice = docCollection.getSlice(shardName);
+    Set<Replica> solrReplicas = new HashSet<>();
+    replicaNames.forEach(name -> {
+      org.apache.solr.common.cloud.Replica replica = slice.getReplica(name);
+      Replica solrReplica = new SimpleClusterAbstractionsImpl.ReplicaImpl(replica.getName(), shard, replica);
```
murblanc, line 75: Why not just do `solrReplicas.add(shard.getReplica(name))` in the `forEach`?

Or an easier to read IMO (personal preference but a lambda here is fine):
`for (String name : replicaNames) { solrReplicas.add(shard.getReplica(name)); }`

murblanc: Or do it with streams...

solr/core/src/java/org/apache/solr/cluster/placement/PlacementContext.java
```diff
@@ -0,0 +1,44 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.solr.cluster.placement;
+
+import org.apache.solr.cluster.Cluster;
+
+/**
+ *
```
murblanc, line 23: Javadoc

solr/core/src/java/org/apache/solr/cluster/placement/impl/PlacementContextImpl.java
```diff
@@ -0,0 +1,39 @@
...
+import org.apache.solr.cluster.placement.AttributeFetcher;
+import org.apache.solr.cluster.placement.PlacementContext;
+import org.apache.solr.cluster.placement.PlacementPlanFactory;
+
+import java.io.IOException;
+
+/**
+ *
+ */
+public class PlacementContextImpl implements PlacementContext {
```
murblanc, line 14: Maybe have "Simple" somewhere in the name of this class given it's instantiating `SimpleClusterAbstractionsImpl`?

solr/core/src/java/org/apache/solr/cluster/placement/impl/SimpleClusterAbstractionsImpl.java
```diff
@@ -324,7 +324,7 @@ public int hashCode() {
       return new Pair<>(replicas, leader);
     }
 
-    private ReplicaImpl(String replicaName, Shard shard, org.apache.solr.common.cloud.Replica sliceReplica) {
+    ReplicaImpl(String replicaName, Shard shard, org.apache.solr.common.cloud.Replica sliceReplica) {
```
murblanc, line 327: Why are these no longer `private`?

solr/core/src/java/org/apache/solr/cluster/placement/plugins/AffinityPlacementConfig.java
```diff
@@ -43,14 +46,30 @@
...
+   * located on the same nodes as the secondary collections (values). The plugin will assume
+   * that the secondary collection replicas are already in place and ignore candidate nodes where
+   * they are not already present.
+   */
+  @JsonProperty
+  public Map<String, String> withCollections;
+
   // no-arg public constructor required for deserialization
   public AffinityPlacementConfig() {
     minimalFreeDiskGB = 20L;
```
murblanc, line 60: I prefer the no arg constructor here to call the appropriate constructor. That way logic is not replicated (might not be applicable here unless somebody replaces `withCollections = Map.of();` with `withCollections = null;` in a future commit), it looks cleaner and by tracing calls to the most complete constructor all callers are inventoried...

solr/core/src/java/org/apache/solr/cluster/placement/plugins/AffinityPlacementFactory.java
```diff
@@ -171,14 +174,17 @@ public AffinityPlacementConfig getConfig() {
 
     private final long prioritizedFreeDiskGB;
 
+    private final Map<String, String> withCollections;
```
murblanc, line 177: Q: a given collection can only be `withCollection` for a single secondary collection? Doesn't seem necessary...

Suggestion: maintain the inverse mapping as well (a multimap, but possibly this one should be a multimap as well) to save looping through map keys checking values...

solr/core/src/java/org/apache/solr/cluster/placement/plugins/AffinityPlacementFactory.java
```diff
@@ -238,11 +247,87 @@ public PlacementPlan computePlacement(Cluster cluster, PlacementRequest request,
...
+          shard.replicas().forEach(replica -> {
+            secondaryNodeShardReplicas.computeIfAbsent(replica.getNode(), n -> new HashMap<>())
+                .computeIfAbsent(replica.getShard().getShardName(), s -> new AtomicInteger())
+                .incrementAndGet();
+          }));
+
+      // find the colocated-with collections
+      Cluster cluster = placementContext.getCluster();
+      Set<SolrCollection> colocatedCollections = new HashSet<>();
+      AtomicReference<IOException> exc = new AtomicReference<>();
```
murblanc, line 280: This variable and how it's handled in the lambda below (and after the lambda) it too complex. If the `forEach` is replaced by a loop (a foreach loop...) the code is a lot simpler (and I believe shorter, although I didn't try to write it).

solr/core/src/java/org/apache/solr/cluster/placement/plugins/AffinityPlacementFactory.java
```diff
@@ -238,11 +247,87 @@ public PlacementPlan computePlacement(Cluster cluster, PlacementRequest request,
...
+              attrValues, replicaTypeToNodes, nodesWithReplicas, coresOnNodes, placementContext.getPlacementPlanFactory(), replicaPlacements);
         }
       }
 
-      return placementPlanFactory.createPlacementPlan(request, replicaPlacements);
+      return placementContext.getPlacementPlanFactory().createPlacementPlan(request, replicaPlacements);
+    }
+
+    @Override
+    public void verifyAllowedModification(ModificationRequest modificationRequest, PlacementContext placementContext) throws PlacementModificationException, InterruptedException {
```
murblanc, line 258: This method should be factored out and cut into a few pieces with meaningful names to make reading easier.

solr/core/src/java/org/apache/solr/cloud/ExclusiveSliceProperty.java
sigram: Heh, you touched it - you own it :)

solr/core/src/java/org/apache/solr/cloud/api/collections/DeleteCollectionCmd.java
sigram: The check can be simplified in the plugin if it only knew that we're deleting the whole collection. I guess we could create a `DeleteCollectionRequest` to let it know the type of modification we want to do.

solr/core/src/java/org/apache/solr/cluster/placement/impl/SimpleClusterAbstractionsImpl.java
sigram: Left-over from earlier refactoring, will revert.

solr/core/src/java/org/apache/solr/cluster/placement/plugins/AffinityPlacementFactory.java
sigram: There's no reason not to extend this to N:1 (N primary collections with 1 secondary), I just followed the model that was present in 8x.

However, the way this co-location is implemented now may degrade the placements if we allow N:1 or N:N models, because of the pre-filtering approach we could quickly run out of candidate nodes for placements. So I prefer to focus on implementing 1:1 now and making sure it works well.

solr/core/src/java/org/apache/solr/cluster/placement/plugins/AffinityPlacementFactory.java
sigram: Fixed.

chatman:
Even a "minimal approach" approach seems to touch 30 files, many of them pertaining to collection API commands. I'm really scared of the complexity here. Can you please consider dropping "withCollection" feature altogether? I've added a comment in JIRA to that effect.

sigram:
> Even a "minimal approach" approach seems to touch 30 files
Please see my response in Jira - some of the changed files are caused by moving a property, other are related to extending the plugin API to allow any kind of veto, not just `withCollection`.

solr/core/src/java/org/apache/solr/cloud/api/collections/Assign.java
```diff
@@ -381,7 +382,11 @@ public AssignmentException(String message, Throwable cause, boolean enableSuppre
 
   public interface AssignStrategy {
     List<ReplicaPosition> assign(SolrCloudManager solrCloudManager, AssignRequest assignRequest)
-        throws Assign.AssignmentException, IOException, InterruptedException;
+        throws AssignmentException, IOException, InterruptedException;
+    void verifyDeleteCollection(SolrCloudManager solrCloudManager, DocCollection collection)
```
murblanc, line 386: Maybe add default implementation here to not force empty implementation in legacy or other places?
I'd add javadoc at least for added methods...

solr/core/src/java/org/apache/solr/cluster/placement/plugins/AffinityPlacementFactory.java
```diff
@@ -238,11 +258,93 @@ public PlacementPlan computePlacement(Cluster cluster, PlacementRequest request,
...
+      if (modificationRequest instanceof DeleteShardsRequest) {
+        throw new UnsupportedOperationException("not implemented yet");
```
murblanc, line 271: If not implemented we should let it happen rather than have it fail.

solr/core/src/java/org/apache/solr/cluster/placement/plugins/AffinityPlacementFactory.java
```diff
@@ -238,11 +258,93 @@ public PlacementPlan computePlacement(Cluster cluster, PlacementRequest request,
...
+      } else if (modificationRequest instanceof DeleteCollectionRequest) {
+        verifyDeleteCollection((DeleteCollectionRequest) modificationRequest, placementContext);
+      } else if (modificationRequest instanceof DeleteReplicasRequest) {
+        verifyDeleteReplicas((DeleteReplicasRequest) modificationRequest, placementContext);
+      } else {
+        throw new UnsupportedOperationException("unsupported request type " + modificationRequest.getClass().getName());
```
murblanc, line 277: here too. Allow everything we don't plan to explicitly disallow.

solr/core/src/java/org/apache/solr/cluster/placement/plugins/AffinityPlacementFactory.java
```diff
@@ -238,11 +258,93 @@ public PlacementPlan computePlacement(Cluster cluster, PlacementRequest request,
...
+      }
+    }
+
+    private void verifyDeleteCollection(DeleteCollectionRequest deleteCollectionRequest, PlacementContext placementContext) throws PlacementModificationException, InterruptedException {
```
murblanc, line 281: Can we have cycles in the `withCollection` graph? Should we allow a way to override the vetting checks from the Collection API?

solr/core/src/java/org/apache/solr/cluster/placement/plugins/AffinityPlacementFactory.java
```diff
@@ -238,11 +258,93 @@ public PlacementPlan computePlacement(Cluster cluster, PlacementRequest request,
...
+      Cluster cluster = placementContext.getCluster();
+      Set<String> colocatedCollections = colocatedWith.getOrDefault(deleteCollectionRequest.getCollection().getName(), Set.of());
+      for (String primaryName : colocatedCollections) {
+        try {
+          if (cluster.getCollection(primaryName) != null) {
+            // still exists
+            throw new PlacementModificationException("colocated collection " + primaryName + " still present");
```
murblanc, line 288: The exception should also tell which collection we refuse to delete (unless that context will be added higher up the call stack?).

solr/core/src/java/org/apache/solr/cluster/placement/plugins/AffinityPlacementFactory.java
```diff
@@ -238,11 +258,93 @@ public PlacementPlan computePlacement(Cluster cluster, PlacementRequest request,
...
+        if (secondaryCount.get() > 1) {
+          // we can delete it - record the deletion
+          secondaryCount.decrementAndGet();
+          continue;
+        }
+        // fail - this replica cannot be removed
+        if (exception == null) {
+          exception = new PlacementModificationException("delete replica(s) rejected");
+        }
+        exception.addRejectedModification(replica.toString(), "co-located with replicas of " + colocatingNodes.get(replica.getNode()));
```
murblanc, line 343: Add the problematic node in the rejection reason so that client can delete the colocated collection replicas there to fix the issue.

solr/core/src/java/org/apache/solr/cluster/placement/DeleteShardsRequest.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+package org.apache.solr.cluster.placement;
+
+import java.util.Set;
+
+/**
+ * Delete shards request.
+ */
+public interface DeleteShardsRequest extends ModificationRequest {
```
murblanc, line 25: If we don't use this interface (i.e. the class that implements it) I suggest we do not include either in this PR. Or at least define and call the corresponding method in `AssignStrategy` from the appropriate `*Cmd` even if nothing does a real implementation and vetting based on it (but it would be ready to be consumed maybe by another plugin written by some user).

solr/core/src/java/org/apache/solr/cluster/placement/impl/ModificationRequestImpl.java
```diff
@@ -0,0 +1,107 @@
...
+    SolrCollection solrCollection = SimpleClusterAbstractionsImpl.SolrCollectionImpl.fromDocCollection(docCollection);
+    Shard shard = solrCollection.getShard(shardName);
+    Slice slice = docCollection.getSlice(shardName);
+    Set<Replica> solrReplicas = new HashSet<>();
+    replicaNames.forEach(name -> solrReplicas.add(shard.getReplica(name)));
+    return deleteReplicasRequest(solrCollection, solrReplicas);
+  }
+
+
+  public static DeleteShardsRequest deleteShardsRequest(SolrCollection collection, Set<String> shardNames) {
```
murblanc, line 83: This and next method should really be called `createDeleteShardsRequest` to convey what they really do, right?

solr/core/src/java/org/apache/solr/cluster/placement/impl/ModificationRequestImpl.java
```diff
@@ -0,0 +1,107 @@
...
+    };
+  }
+
+  /**
+   * Create a delete replicas request using the internal Solr API.
+   * @param docCollection Solr collection
+   * @param shardName shard name
+   * @param replicaNames replica names (aka. core-node names)
+   */
+  public static DeleteReplicasRequest deleteReplicasRequest(DocCollection docCollection, String shardName, Set<String> replicaNames) {
```
murblanc, line 73: The signature is not consistent with `deleteReplicasRequest` above that takes `Replica` instances rather than `String` names. I prefer the strongly typed signature.

solr/core/src/java/org/apache/solr/cluster/placement/impl/ModificationRequestImpl.java
```diff
@@ -0,0 +1,107 @@
...
+    SolrCollection solrCollection = SimpleClusterAbstractionsImpl.SolrCollectionImpl.fromDocCollection(docCollection);
+    return () -> solrCollection;
+  }
+
+  /**
+   * Create a delete replicas request.
+   * @param collection collection to delete replicas from
+   * @param replicas replicas to delete
+   */
+  public static DeleteReplicasRequest deleteReplicasRequest(SolrCollection collection, Set<Replica> replicas) {
```
murblanc, line 47: Method should be called `createDeleteReplicasRequest`?

solr/core/src/java/org/apache/solr/cluster/placement/plugins/AffinityPlacementConfig.java
```diff
@@ -43,14 +46,28 @@
...
+   * they are not already present.
+   */
+  @JsonProperty
+  public Map<String, String> withCollections;
+
   // no-arg public constructor required for deserialization
   public AffinityPlacementConfig() {
-    minimalFreeDiskGB = 20L;
-    prioritizedFreeDiskGB = 100L;
+    this (20L, 100L, Map.of());
```
murblanc, line 60: minor: remove space after `this`.

murblanc: Also, this constructor should call the constructor not requiring a `Map`, it will add an empty one.

BTW, the deserialization constructor should not pass "meaningful" default values since those will not be used.
I'd call `this(0L, 0L)` here instead, and switch the `AffinityPlacementConfig DEFAULT` static variable to use the constructor accepting values. That will also make seeing what are the defaults easier when reading the code.

solr/core/src/java/org/apache/solr/cluster/placement/plugins/AffinityPlacementFactory.java
murblanc: We do not support in Solr returning errors from Collection API calls in a way that is consumable by code, right? Would be a nice addition...

solr/core/src/java/org/apache/solr/cluster/placement/plugins/AffinityPlacementFactory.java
```diff
@@ -467,7 +569,7 @@ private void makePlacementDecisions(SolrCollection solrCollection, String shardN
         if (candidateAzEntries == null) {
           // This can happen because not enough nodes for the placement request or already too many nodes with replicas of
           // the shard that can't accept new replicas or not enough nodes with enough free disk space.
-          throw new PlacementException("Not enough nodes to place " + numReplicas + " replica(s) of type " + replicaType +
+          throw new PlacementException("Not enough eligible nodes to place " + numReplicas + " replica(s) of type " + replicaType +
```
murblanc, line 572: Can we add in the message if there were `withCollection` constraints used in the placement logic? No details, not even indicating that it's the failure reason, just mentioning it...

murblanc: Not sure if it's useful here or output elsewhere already though.

murblanc (approved):
LGTM. Minor comments here and there.

solr/core/src/java/org/apache/solr/cluster/placement/plugins/AffinityPlacementFactory.java
sigram: I don't think we should complicate the implementation to handle this. If the user created a cycle then deleting any collection in the cycle will be impossible - but this can be easily remedied by modifying the configuration of the plugin (which takes effect nearly immediately) to break the cycle.

murblanc: I agree. Or change the configuration of the collection (I assume `withCollection` can be removed).

solr/core/src/java/org/apache/solr/cluster/placement/plugins/AffinityPlacementFactory.java
sigram: it's already handled in `filterNodesWithCollection`

PR #2202, opened 2021-01-14
LUCENE-9664: Hunspell support: fix most IntelliJ warnings, cleanup

donnerpeter:
# Description

The code produces too many warnings, so it's hard to notice newly introduced ones when developing new functionality.

# Solution

Apply IDE quick fixes, sometimes cleanup manually.

# Tests

No semantic changes, just cleanup.

# Checklist

### DIFF
--- a/lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/Dictionary.java
+++ b/lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/Dictionary.java
@@ -16,7 +16,14 @@
  */
 package org.apache.lucene.analysis.hunspell;
 
-import java.io.*;
+import java.io.BufferedInputStream;
+import java.io.BufferedOutputStream;
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.io.LineNumberReader;
+import java.io.OutputStream;
 import java.nio.charset.Charset;
 import java.nio.charset.CharsetDecoder;
 import java.nio.charset.CodingErrorAction;
@@ -25,7 +32,16 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.text.ParseException;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.HashMap;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import java.util.TreeMap;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import org.apache.lucene.codecs.CodecUtil;
@@ -33,12 +49,25 @@ import org.apache.lucene.store.ByteArrayDataOutput;
 import org.apache.lucene.store.Directory;
 import org.apache.lucene.store.IOContext;
 import org.apache.lucene.store.IndexOutput;
-import org.apache.lucene.util.*;
+import org.apache.lucene.util.ArrayUtil;
+import org.apache.lucene.util.BytesRef;
+import org.apache.lucene.util.BytesRefBuilder;
+import org.apache.lucene.util.BytesRefHash;
+import org.apache.lucene.util.CharsRef;
+import org.apache.lucene.util.IOUtils;
+import org.apache.lucene.util.IntsRef;
+import org.apache.lucene.util.IntsRefBuilder;
+import org.apache.lucene.util.OfflineSorter;
 import org.apache.lucene.util.OfflineSorter.ByteSequencesReader;
 import org.apache.lucene.util.OfflineSorter.ByteSequencesWriter;
 import org.apache.lucene.util.automaton.CharacterRunAutomaton;
 import org.apache.lucene.util.automaton.RegExp;
-import org.apache.lucene.util.fst.*;
+import org.apache.lucene.util.fst.CharSequenceOutputs;
+import org.apache.lucene.util.fst.FST;
+import org.apache.lucene.util.fst.FSTCompiler;
+import org.apache.lucene.util.fst.IntSequenceOutputs;
+import org.apache.lucene.util.fst.Outputs;
+import org.apache.lucene.util.fst.Util;
 
 /** In-memory structure for the dictionary (.dic) and affix (.aff) data of a hunspell dictionary. */
 public class Dictionary {
@@ -221,21 +250,21 @@ public class Dictionary {
   }
 
   /** Looks up Hunspell word forms from the dictionary */
-  IntsRef lookupWord(char[] word, int length) {
-    return lookup(words, word, length);
+  IntsRef lookupWord(char[] word, int offset, int length) {
+    return lookup(words, word, offset, length);
   }
 
   // only for testing
   IntsRef lookupPrefix(char[] word) {
-    return lookup(prefixes, word, word.length);
+    return lookup(prefixes, word, 0, word.length);
   }
 
   // only for testing
   IntsRef lookupSuffix(char[] word) {
-    return lookup(suffixes, word, word.length);
+    return lookup(suffixes, word, 0, word.length);
   }
 
-  private IntsRef lookup(FST<IntsRef> fst, char[] word, int length) {
+  IntsRef lookup(FST<IntsRef> fst, char[] word, int offset, int length) {
     if (fst == null) {
       return null;
     }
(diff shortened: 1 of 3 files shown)


lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/Dictionary.java
```diff
@@ -251,34 +221,33 @@ public Dictionary(
...
     final FST.BytesReader bytesReader = fst.getBytesReader();
-    final FST.Arc<IntsRef> arc = fst.getFirstArc(new FST.Arc<IntsRef>());
+    final FST.Arc<IntsRef> arc = fst.getFirstArc(new FST.Arc<>());
     // Accumulate output as we go
     final IntsRef NO_OUTPUT = fst.outputs.getNoOutput();
     IntsRef output = NO_OUTPUT;
 
-    int l = offset + length;
     try {
-      for (int i = offset, cp = 0; i < l; i += Character.charCount(cp)) {
```
dweiss, line 280: So this offset thing was never used? I this not useful to have keep it though? Utility methods could still exist (they'd use offset = 0) but it seems like some flexibility with offset here is fine.

lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/Dictionary.java
```diff
@@ -16,14 +16,7 @@
...
 
-import java.io.BufferedInputStream;
-import java.io.BufferedOutputStream;
-import java.io.BufferedReader;
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.InputStreamReader;
-import java.io.LineNumberReader;
-import java.io.OutputStream;
+import java.io.*;
```
dweiss, line 19: Did you run gradlew tidy? Wildcard imports shouldn't be there, hence the question.

lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/Stemmer.java
```diff
@@ -654,7 +646,7 @@ private boolean checkCondition(
                   true,
                   circumfix,
                   caseVariant));
-        } else if (dictionary.complexPrefixes == false && dictionary.twoStageAffix) {
+        } else if (!dictionary.complexPrefixes && dictionary.twoStageAffix) {
```
dweiss, line 649: I don't care much but I know some other committers prefer the verbose version (== false).

lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/Dictionary.java
donnerpeter: The imports were collapsed by the IDE, but `gradle precheck` didn't complain of wildcards. I can expand them back if that matters.

lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/Dictionary.java
```diff
@@ -209,6 +178,7 @@ public Dictionary(
     this.needsOutputCleaning = false; // set if we have an OCONV
     flagLookup.add(new BytesRef()); // no flags -> ord 0
 
+    Path tempPath = getDefaultTempDir(); // TODO: make this configurable?
```
donnerpeter, line 181: BTW any idea why the default temp dir is used here for reading from aff file twice, but another temp dir is passed to the constructor and used for similarly-temp-looking task (external dictionary entry sorting)? Can these tasks use the same dir? And if yes, which one? Clients would be simpler if they hadn't to care about temp dirs, but maybe there are use cases for passing the temp dir explicitly.

lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/Dictionary.java
donnerpeter: No one uses it now, and it's easy to add the offset back if needed. So far it doesn't seem very likely to happen.

lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/Dictionary.java
dweiss: Darn. I see you're right. They should be expanded - enforcement of this has been lost somewhere during ant-gradle migration. gjf doesn't deal with this [1].

You can also leave it in if you wish - automatic import expansion/ ordering can be done later (please feel free to file a separate issue!).

https://github.com/google/google-java-format/issues/113

lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/Dictionary.java
dweiss: I didn't write this code, so I don't know for sure. I believe the externally-provided temp folder should be used (and reused) but it's of type Directory (it's a Lucene abstraction) so I'm not sure if you'll be able to reuse it for generic Paths.

lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/Stemmer.java
donnerpeter: I prefer shorter versions (especially given the short column limit of spotlessJava), and IntelliJ highlights this by default, but if you insist, I can disable the highlighting and live with it. What's the rule exactly, write `x == false` instead of `!x` everywhere?

dweiss: So do I. But @romseygeek has a different preference. I think you can follow your own gut feeling here - there are no explicit rules that we enforce.

lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/Dictionary.java
dweiss: I'd rather keep it in if it's already there. Seems like it's just more complete this way (you point at a range of characters at the source buffer rather than require zero-based indexes). I'd even add a test that verifies it works by randomizing the offset somewhere...

donnerpeter: Having it here complicates the code and its refactoring a bit. That said, adding a test would remove IntelliJ's warning, so it's also a way to go. But why randomized? Passing `2` or something looks simple enough, and it doesn't seem that randomization would catch new bugs in this context.

lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/Stemmer.java
donnerpeter: Then I'd shorten this :)

lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/Dictionary.java
dweiss: Passing a constant number gives you one test (which is fine). Passing a randomized number (0-X) gives you many tests over time. This seems stupid only until you find a bug deep down somewhere, hiding under a complex scenario of inputs. Maybe the offset is used to derive a different index, or an overflow occurs somewhere... I don't know but I've seen this too many times to call it redundant. ;)

donnerpeter: OK, having created a property-based test framework myself, I'm in a wrong position to argue about randomization usefulness. I'll try this, thanks!

dweiss: No pressure. :) I typically add known border cases and a randomized test on top of that. Fuzzifiers are great, really. Take a look at this bug which slipped through the cracks for so many years:
https://mail.openjdk.java.net/pipermail/javadoc-dev/2020-December/002255.html

lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/Stemmer.java
```diff
@@ -572,9 +566,7 @@ private boolean checkCondition(
     affixReader.setPosition(8 * affix);
     char flag = (char) (affixReader.readShort() & 0xffff);
     affixReader.skipBytes(2); // strip
```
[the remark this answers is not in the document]
donnerpeter: As promised in JIRA, I've inlined the `condition` variable to avoid potential confusion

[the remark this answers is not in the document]
dweiss: Ugh. Sorry. The code contains nearly similar fragments and I was looking at the wrong section. Maybe it'd be a good idea to keep that condition variable and assert it doesn't have any non-zero bits after the one we use?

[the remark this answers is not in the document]
donnerpeter: That assertion would be incorrect, as those bits are also used sometimes :)
In fact, I have another [refactoring](https://github.com/donnerpeter/lucene-solr/commit/9122208d04e0f5811c6d292ce40efa8423fc5aad) to be applied after this patch. It deduplicates the code around this place, and there `condition` and `crossProduct` are moved closer to where they're needed and become independent.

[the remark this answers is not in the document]
dweiss: Ah. Clear - thanks! I unfortunately look at it from the side. Please disregard then.

dweiss:
Thanks Peter.

PR #2204, opened 2021-01-14
LUCENE-9669: Add an expert API to allow opening indices created < N-1

s1monw:
Today we force indices that were created with N-2 and older versions of lucene
to fail on open. This check doesn't even check if the codecs are available. In order
to allow users to open older indices and for us to support N-2 versions this change
adds an API on DirectoryReader to specify a mininum index version on a per reader basis.
This doesn't apply for the IndexWriter which will fail on opening older indices.


s1monw:
I opened this prematurely....

PR #2205, opened 2021-01-15
LUCENE-9668: Deprecate MinShouldMatchSumScorer with WANDScorer

zacharymorn:
# Description

Deprecate MinShouldMatchSumScorer with WANDScorer since the two are very similar at this point. This is a follow-up PR to https://github.com/apache/lucene-solr/pull/2141 .

# Solution

As `WANDScorer` can already handle the cases where `scoreMode.needsScore() == true`, and `MinShouldMatchSumScorer` can handle the cases where `scoreMode.needsScore() == false` , this PR deprecates `MinShouldMatchSumScorer` with `WANDScorer` primarily by skipping block-max scoring update related logic when `scoreMode.needsScore() == false`.  

# Tests

The existing tests as well as 3 newly added tests are passing (currently the nocommit comment fails validation check). 

# Checklist


lucene/core/src/java/org/apache/lucene/search/DisjunctionMaxScorer.java
```diff
@@ -50,7 +50,7 @@
     if (tieBreakerMultiplier < 0 || tieBreakerMultiplier > 1) {
       throw new IllegalArgumentException("tieBreakerMultiplier must be in [0, 1]");
     }
-    if (scoreMode == ScoreMode.TOP_SCORES) {
+    if (scoreMode.needsScores()) {
```
jpountz, line 53: why do we need to change this?

lucene/core/src/java/org/apache/lucene/search/WANDScorer.java
```diff
@@ -149,23 +167,33 @@ private static long scaleMinScore(float minScore, int scalingFactor) {
...
-    OptionalInt scalingFactor = OptionalInt.empty();
-    for (Scorer scorer : scorers) {
-      scorer.advanceShallow(0);
-      float maxScore = scorer.getMaxScore(DocIdSetIterator.NO_MORE_DOCS);
-      if (maxScore != 0 && Float.isFinite(maxScore)) {
-        // 0 and +Infty should not impact the scale
-        scalingFactor =
-            OptionalInt.of(
-                Math.min(scalingFactor.orElse(Integer.MAX_VALUE), scalingFactor(maxScore)));
+    if (needsScore) {
```
jpountz, line 176: likewise here I would have expected to check `scoreMode == TOP_SCORES` rather than `scoreMode.needsScores()` since calling `advanceShallow` `getMaxScore` only really makes sense if you plan to skip documents based on scores?

lucene/core/src/java/org/apache/lucene/search/Boolean2ScorerSupplier.java
```diff
@@ -231,10 +231,27 @@ private Scorer opt(
...
+      // minShouldMatch > 1).
+      //
+      // Ultimately, WANDScorer should be able to handle the following conditions now:
+      // 1. Any ScoreMode (with scoring or not), although it might be a bit slower compared to
+      //   DisjunctionSumScorer due to data structures and algorithms usd
+      // 2. Any minCompetitiveScore ( >= 0 )
+      // 3. Any minShouldMatch ( >= 0 )
+      //
+      // So it seems we might need a different condition check to differentiate usage between
+      // WANDScorer and DisjunctionSumScorer ?
```
jpountz, line 252: The current condition looks right to me. Maybe the comment could say something like `WANDScorer can handle all cases but we specialize exhaustive pure disjunctions with DisjunctionSumScorer which is a bit faster`?

lucene/core/src/java/org/apache/lucene/search/WANDScorer.java
```diff
@@ -319,7 +356,10 @@ private void pushBackLeads(int target) throws IOException {
       }
     }
     lead = null;
-    leadMaxScore = 0;
+
+    if (needsScore) {
+      leadMaxScore = 0;
+    }
```
jpountz, line 362: I wonder if this condition could be removed in order to make this class easier to understand by minimizing differences in code paths between the scoring and the non-scoring cases? Are there other branches we could remove?

jpountz (commented):
Thank you for tackling this @zacharymorn. I left some questions but this looks like a great start.

lucene/core/src/java/org/apache/lucene/search/DisjunctionMaxScorer.java
zacharymorn: Sorry should have put in a nocommit here to explain this change. Basically without this change a NPE would be thrown for test `DisMaxRequestHandlerTest#testSomeStuff` with the following stack trace:

```
 2> 4261 ERROR (TEST-DisMaxRequestHandlerTest.testSomeStuff-seed#[66968D4353C5888]) [     ] o.a.s.h.RequestHandlerBase java.lang.NullPointerException
  2>    at org.apache.lucene.search.DisjunctionMaxScorer.advanceShallow(DisjunctionMaxScorer.java:78)
  2>    at org.apache.lucene.search.WANDScorer.<init>(WANDScorer.java:179)
  2>    at org.apache.lucene.search.Boolean2ScorerSupplier.opt(Boolean2ScorerSupplier.java:257)
  2>    at org.apache.lucene.search.Boolean2ScorerSupplier.getInternal(Boolean2ScorerSupplier.java:124)
  2>    at org.apache.lucene.search.Boolean2ScorerSupplier.get(Boolean2ScorerSupplier.java:96)
  2>    at org.apache.lucene.search.Boolean2ScorerSupplier.req(Boolean2ScorerSupplier.java:168)
  2>    at org.apache.lucene.search.Boolean2ScorerSupplier.getInternal(Boolean2ScorerSupplier.java:146)
  2>    at org.apache.lucene.search.Boolean2ScorerSupplier.get(Boolean2ScorerSupplier.java:96)
  2>    at org.apache.lucene.search.BooleanWeight.scorer(BooleanWeight.java:353)
  2>    at org.apache.lucene.search.Weight.bulkScorer(Weight.java:166)
  2>    at org.apache.lucene.search.BooleanWeight.bulkScorer(BooleanWeight.java:343)
  2>    at org.apache.lucene.search.IndexSearcher.search(IndexSearcher.java:743)
  2>    at org.apache.lucene.search.IndexSearcher.search(IndexSearcher.java:533)
  2>    at org.apache.solr.search.SolrIndexSearcher.buildAndRunCollectorChain(SolrIndexSearcher.java:210)
  2>    at org.apache.solr.search.SolrIndexSearcher.getDocListAndSetNC(SolrIndexSearcher.java:1709)
  2>    at org.apache.solr.search.SolrIndexSearcher.getDocListC(SolrIndexSearcher.java:1412)
  2>    at org.apache.solr.search.SolrIndexSearcher.search(SolrIndexSearcher.java:599)
  2>    at org.apache.solr.handler.component.QueryComponent.doProcessUngroupedSearch(QueryComponent.java:1508)
  2>    at org.apache.solr.handler.component.QueryComponent.process(QueryComponent.java:398)
  2>    at org.apache.solr.handler.component.SearchHandler.handleRequestBody(SearchHandler.java:355)
  2>    at org.apache.solr.handler.RequestHandlerBase.handleRequest(RequestHandlerBase.java:214)
  2>    at org.apache.solr.core.SolrCore.execute(SolrCore.java:2617)
  2>    at org.apache.solr.util.TestHarness.query(TestHarness.java:346)
  2>    at org.apache.solr.util.TestHarness.query(TestHarness.java:328)
  2>    at org.apache.solr.SolrTestCaseJ4.assertQ(SolrTestCaseJ4.java:982)
  2>    at org.apache.solr.DisMaxRequestHandlerTest.doTestSomeStuff(DisMaxRequestHandlerTest.java:141)
  2>    at org.apache.solr.DisMaxRequestHandlerTest.testSomeStuff(DisMaxRequestHandlerTest.java:72)
```

The scoreMode and minShouldMatch combination that caused this exception was `scoreMode == COMPLETE`, `minShouldMatch == 3`. 

I think what's happening here is, `MinShouldMatchSumScorer` was already used in test for other scoring ScoreMode in addition to `ScoreMode.TOP_SCORES` (and probably already used by clients in the wild as well), as the original condition before adding `minShouldMatch` support to `WANDScorer` is 

```
if (minShouldMatch > 1) {
  return MinShouldMatchSumScorer...
} else if (scoreMode == ScoreMode.TOP_SCORES) {
  return WANDScorer...
} else {
  return DisjunctionSumScorer...
}
```

so I was assuming we may need to preserve this behavior when we deprecate `MinShouldMatchSumScorer` with `WANDScorer`?  This was also the reason it was checking for `needsScore` instead of `scoreMode == TOP_SCORES` in `WANDScorer` constructor, and the long nocommit message in `Boolean2ScorerSupplier.java` to see if it may need a different conditional to differentiate between `WANDScorer` and `DisjunctionSumScorer`, as you noted in other review comments.

lucene/core/src/java/org/apache/lucene/search/WANDScorer.java
zacharymorn: Makes sense. Let me take a further look at this and try out.

lucene/core/src/java/org/apache/lucene/search/DisjunctionMaxScorer.java
jpountz: > so I was assuming we may need to preserve this behavior when we deprecate MinShouldMatchSumScorer with WANDScorer

Right I'm hoping we can use WANDScorer for minShouldMatch > 1 for all values of ScoreMode!

I believe that this exception you saw should disappear if you make sure that `WANDScorer` only calls `advanceShallow`/`getMaxScore` with `ScoreMode.TOP_SCORES`.

zacharymorn: Ahhhh I see now I had a wrong assumption / impression before. For some reason I mistakenly thought since `MinShouldMatchSumScorer` supports `ScoreMode.COMPLETE`, `WANDScorer` should also allow score computation for that mode after deprecation. Will push an update soon to check for `ScoreMode.TOP_SCORES` instead.

lucene/core/src/java/org/apache/lucene/search/WANDScorer.java
zacharymorn: Updated to check for `scoreMode == TOP_SCORES` instead.

lucene/core/src/java/org/apache/lucene/search/DisjunctionMaxScorer.java
zacharymorn: Updated to check for `scoreMode == TOP_SCORES` instead.

lucene/core/src/java/org/apache/lucene/search/Boolean2ScorerSupplier.java
zacharymorn: Updated the comment here.

lucene/core/src/java/org/apache/lucene/search/WANDScorer.java
zacharymorn: In the latest commit, I've done some refactoring and removed a few newly added branches that can be no-op for non-scoring mode, but not by a lot. Currently there are about 5 conditional branches left for handling non-scoring cases (checking `scoreMode  == ScoreMode.TOP_SCORES`), 1 is used in constructor, 2 used in assertions, and the last 2 were used in `advance` and `doNextCompetitiveCandidate`, so I felt that might be the minimal number of branches we could have now without major refactoring changes? What do you think?

lucene/core/src/java/org/apache/lucene/search/WANDScorer.java
```diff
@@ -149,23 +167,33 @@ private static long scaleMinScore(float minScore, int scalingFactor) {
...
       }
+
+      this.scalingFactor = scalingFactor.orElse(0);
+      this.maxScorePropagator = new MaxScoreSumPropagator(scorers);
+    } else {
+      this.scalingFactor = 0;
+      this.maxScorePropagator = null;
     }
+
     // Use a scaling factor of 0 if all max scores are either 0 or +Infty
```
jpountz, line 196: this comment looks misplaced now?

lucene/core/src/java/org/apache/lucene/search/WANDScorer.java
```diff
@@ -257,13 +288,17 @@ public int advance(int target) throws IOException {
...
 
-            // Pop the new 'lead' from 'head'
-            moveToNextCandidate(target);
+            if (scoreMode == ScoreMode.TOP_SCORES) {
+              // Update score bounds if necessary so
+              updateMaxScoresIfNecessary(target);
 
-            if (doc == DocIdSetIterator.NO_MORE_DOCS) {
-              return DocIdSetIterator.NO_MORE_DOCS;
+              if (doc == DocIdSetIterator.NO_MORE_DOCS) {
```
jpountz, line 295: we may have a problem, or maybe this line is not needed anymore

Before, `moveToNextCandidate` would advance `doc`. But now that you are calling `updateMaxScoresIfNecessary ` instead, which doesn't advance doc, it shouldn't be possible that doc == NO_MORE_DOCS on this line?

jpountz (commented):
It looks good in general, but I'd need to take a second look to make sure that the refactorings to WANDScorer are correct. I left a question on something that looks suspicious, maybe you can take a look?

lucene/core/src/java/org/apache/lucene/search/WANDScorer.java
zacharymorn: For this one I actually moved the doc advancement to `NO_MORE_DOCS` for condition `head.size() == 0` into `updateMaxScoresIfNecessary`, as initially I felt these two are closely related and only applicable for TOP_SCORES mode (I probably should have updated the method name though) https://github.com/zacharymorn/lucene-solr/blob/9243ef01de6f4a4154275919a32f5703e970f49d/lucene/core/src/java/org/apache/lucene/search/WANDScorer.java#L463-L468 . And thus the original `moveToNextCandidate` implementation was separated into `updateMaxScoresIfNecessary` and `setDocAndFreq`. 

However, in hindsight I see this is not a very good abstraction, and it's probably better to keep the original `moveToNextCandidate` and handle the non-scoring mode inside. Let me give that a try.

zacharymorn: I just pushed a commit to revert back to the original `moveToNextCandidate` abstraction, and handle the non TOP_SCORES mode inside that method. I think that really helps to localize / simplify the changes (and reduce 1 more branching) a lot!  Please let me know if this looks good to you.

jpountz (approved):

jpountz:
Thank you @zacharymorn !

zacharymorn:
> Thank you @zacharymorn !

No problem! Thanks Adrien for the review and guidance!

PR #2206, opened 2021-01-15
BlobDirectoryFactory correctly deletes directories in the blob store

bruno-roustant:
… and reuses BlobStore and BlobPusher.


dsmiley (approved):
LGTM.  I'm not super familiar with some of this but I found it interesting to explore some of the methods on DirectoryFactory like cleanupOldIndexDirectories to see when they are called.

I wouldn't add to the BlobStore abstraction unless it's totally equivalent to BackupRepository which I strongly think we're going to migrate to.

risdenk:
@bruno-roustant / @dsmiley is this still valid?

PR #2207, opened 2021-01-15
LUCENE-9667: Hunspell: add spellchecker API, support BREAK and FORBIDDENWORD affix rules

donnerpeter:
# Description

Create a `SpellChecker` class in the API with `spell` method that should mirror Hunspell in future. For start, support word splitting by BREAK directive.

# Solution

the new code is based on https://github.com/hunspell/hunspell/blob/master/src/hunspell/hunspell.cxx#L675

# Tests

Test data taken from hunspell C++

# Checklist


lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/SpellChecker.java
```diff
@@ -0,0 +1,103 @@
...
+
+import org.apache.lucene.util.BytesRef;
+
+/**
+ * A spell checker based on Hunspell dictionaries. Thread-safe. Not all Hunspell features are
+ * supported yet.
+ */
+public class SpellChecker {
+  private final Dictionary dictionary;
+  private final ThreadLocal<Stemmer> stemmer;
```
dweiss, line 27: A suggestion. If it's not absolutely necessary to make it thread-safe then I'd just use non-thread safe fields. If something is concurrent up the stack then it can be handled there (as a thread local SpellChecker). Or just store the dictionary and create an instance of SpellChecker on demand, it's a cheap operation?

donnerpeter: I did some performance testing and it seems that allocating Stemmer/BytesRef on each query seems fine indeed. Thanks!

dweiss: No problem at all. If we can postpone such optimizations, it's typically better solved in the particular context where it's used.

lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/SpellChecker.java
```diff
@@ -24,22 +24,24 @@
  */
 public class SpellChecker {
   private final Dictionary dictionary;
-  private final ThreadLocal<Stemmer> stemmer;
-  private final ThreadLocal<BytesRef> scratch = ThreadLocal.withInitial(BytesRef::new);
+  private final BytesRef scratch = new BytesRef();
 
-  public SpellChecker(Dictionary dictionary) {
+  private SpellChecker(Dictionary dictionary) {
```
dweiss, line 29: Hey Peter. Sorry, perhaps I wasn't clear - I'd leave the public SpellChecker class with a non-static spell(String word) method in. I just wouldn't try to make it thread-safe. Imagine you have a gazillion words to spell-check - then you'd create SpellChecker and just apply it in a loop to all the words. This seems intuitive to me. I just don't think it needs to be decorated with thread locals internally... this was the bit I was concerned with.

donnerpeter: Yes, I've just decided that the static method would be better, as it'd preserve the thread-safety of the API and avoid dirty thread-locals. If needed, the constructor can be made public later, but for now in my 3M-word test see almost no difference. If you insist I can make the constructor public, it just doesn't feel right to me when a supposedly pure function is thread-unsafe, and I've seen bugs caused by such assumptions.

dweiss: I think the default reasoning is that nothing is thread safe unless stated otherwise, period. I don't see why the SpellChecker would be different here. It is a facade over a raw dictionary providing some functionality, seems quite natural to me for it just to expose a method to stem words (once built). A similar pattern is used in other places - like here:

https://github.com/apache/lucene-solr/blob/master/lucene/analysis/morfologik/src/java/org/apache/lucene/analysis/morfologik/MorfologikFilter.java#L79

Neither MorfologikFilter nor the highlighted line (DictionaryLookup) are thread-safe. I imagine SpellChecker would correspond to what DictionaryLookup is in Morfologik's world.

https://github.com/morfologik/morfologik-stemming/blob/master/morfologik-stemming/src/main/java/morfologik/stemming/DictionaryLookup.java

donnerpeter: Yes, the threading bugs I've mentioned were related exactly to morfologik :) I'd prefer to prevent them from happening here.

For data structures, I agree that one shouldn't expect them to be thread-safe by default, but this isn't a data structure. It's a way to run queries (a facade) over a thread-safe Dictionary, and I expect such things to be thread-safe as well. It isn't, but that's an implementation detail, and I don't see much evil in hiding it for now, given that hiding is cheap and exposing can easily be done in future.

That said, I'm not totally committed to this approach. I'd estimate the strength of my desire to hide thread-unsafety as 7 out of 10. How strongly do you insist on your suggestion? :)

dweiss: Well, I'm sorry you had problems with morfologik (which I co-wrote), but this is consistent with what I said - if a class doesn't say it's thread safe then it's very likely not... :)

The way I see this is slightly different than data-structure vs. component - to me a dictionary is an immutable heavy data class (to load and store) and the rest is a mutable lightweight facade that allows convenient api over that immutable heavy dictionary... Even if the facade is lightweight, it's not really zero-cost. Creating objects is fairly cheap but when you create millions of them it will eventually accumulate; reusing the same buffers helps with caches too.

Also, as for thread-safe classes... I don't think many people do it anymore (perhaps with the notable exception of Doug Lea :). There are typically much more efficient (and simpler) ways of dealing with concurrency higher up the stack. If you're trying to solve concurrency at such a low level you end up having to debug things later - for example when your thread locals keep references to dictionaries that are no longer used anywhere.

This is all a bit philosophical and subjective, of course. There is no right and wrong and everyone has an unique angle of approaching things, depending of where they got burnt in the past. I just expressed my preference based on exactly that :)

So it's up to you, Peter, really.

donnerpeter: OK, I've convinced myself that clients re in a better position to manage threading and shouldn't pay unnecessarily. Thanks!

dweiss: Let's wait for the build to pass and I'll merge it in. Thank you!

PR #2212, opened 2021-01-16
LUCENE-9669: Add an expert API to allow opening indices created < N-1

s1monw:
Today we force indices that were created with N-2 and older versions of lucene
to fail on open. This check doesn't even check if the codecs are available. In order
to allow users to open older indices and for us to support N-2 versions this change
adds an API on DirectoryReader to specify a mininum index version on a per reader basis.
This doesn't apply for the IndexWriter which will fail on opening older indices.


lucene/backward-codecs/src/test/org/apache/lucene/backward_index/TestBackwardsCompatibility.java
```diff
@@ -1986,4 +1989,36 @@ static BytesRef toBytes(long value) {
...
     return bytes;
   }
+
+  public void testFailOpenOldIndex() throws IOException {
+    for (String name : oldNames) {
+      Directory directory = oldIndexDirs.get(name);
+      IndexCommit commit = DirectoryReader.listCommits(directory).get(0);
+      IndexFormatTooOldException ex = expectThrows(IndexFormatTooOldException.class, () ->
+              StandardDirectoryReader.open(commit, Version.LATEST.major));
+      ex.getMessage().contains("only supports reading from version " + Version.LATEST.major + " upwards." );
```
jpountz, line 1999: this should be within an `assertTrue`?

lucene/core/src/java/org/apache/lucene/index/DirectoryReader.java
```diff
@@ -104,6 +105,17 @@ public static DirectoryReader open(final IndexCommit commit) throws IOException
     return StandardDirectoryReader.open(commit.getDirectory(), commit);
   }
 
+  /**
+   * Expert: returns an IndexReader reading the index in the given {@link IndexCommit}.
+   *
+   * @param commit the commit point to open
+   * @param minIndexVersionCreated the minimum index version created to check before opening the index
```
jpountz, line 112: I like how `listCommits` calls it a "major" version which avoids some ambiguity, maybe do the same here?

Also could you improve javadocs to better describe how this version parameter affects the behavior of this method?

lucene/core/src/java/org/apache/lucene/index/StandardDirectoryReader.java
```diff
@@ -40,42 +41,49 @@
...
-      IndexWriter writer,
-      SegmentInfos sis,
-      boolean applyAllDeletes,
-      boolean writeAllDeletes)
+          Directory directory,
+          LeafReader[] readers,
+          IndexWriter writer,
+          SegmentInfos sis,
+          boolean applyAllDeletes,
+          boolean writeAllDeletes, int minVersionCreated)
```
jpountz, line 53: put on a new line like other parameters?

lucene/core/src/java/org/apache/lucene/index/LeafMetaData.java
```diff
@@ -31,15 +31,19 @@
...
+    }
     if (createdVersionMajor > Version.LATEST.major) {
       throw new IllegalArgumentException(
           "createdVersionMajor is in the future: " + createdVersionMajor);
     }
-    if (createdVersionMajor < 6) {
+    if (createdVersionMajor < minVersionSupported) {
       throw new IllegalArgumentException(
-          "createdVersionMajor must be >= 6, got: " + createdVersionMajor);
+          "createdVersionMajor must be >= " + minVersionSupported + ", got: " + createdVersionMajor);
```
jpountz, line 46: These checks were introduced in the first place because 7.0 was the version that started recording the major version that created the index, and any older index would have this constant set to 6, rather than because 6 was the minimum supported version. So I think we can keep the logic how it is today without adding a new parameter to the ctor?

jpountz (commented):
With this change, expert users would have to pass the minimum major version that is required to both `DirectoryReader#listCommits` and `DirectoryReader#open`. I wonder if we really need it in `listCommits`, maybe we could always return commits regardless of whether the index creation version is too old or not and only have the check in `DirectoryReader#open`? What do you think?

lucene/core/src/java/org/apache/lucene/index/LeafMetaData.java
s1monw: I rolled this back. I really misread the logic. thanks for catching it.

s1monw:
@jpountz thanks, great feedback. I pushed a new commit

lucene/backward-codecs/src/test/org/apache/lucene/backward_index/TestBackwardsCompatibility.java
```diff
@@ -1986,4 +1988,45 @@ static BytesRef toBytes(long value) {
...
+        Path oldIndexDir = createTempDir(name);
+        TestUtil.unzip(getDataInputStream("unsupported." + name + ".zip"), oldIndexDir);
+        try (BaseDirectoryWrapper dir = newFSDirectory(oldIndexDir)) {
+          // don't checkindex, we don't have the codecs yet
+          dir.setCheckIndexOnClose(false);
+          IllegalArgumentException iae =
+              expectThrows(IllegalArgumentException.class, () -> DirectoryReader.listCommits(dir));
+          // TODO fix this once we have the codec for 7.0 recreated
+          assertEquals(
+              "Could not load codec 'Lucene70'. Did you forget to add lucene-backward-codecs.jar?",
```
jpountz, line 2021: I'm surprised that this works, not all 7.x indices use Lucene70 as a codec, e.g. there must be some indices that use Lucene75 or something like that as a codec? What am I missing?

jpountz (approved):

lucene/backward-codecs/src/test/org/apache/lucene/backward_index/TestBackwardsCompatibility.java
s1monw: it seems like we only had `LUCENE70` in this release https://github.com/apache/lucene-solr/tree/branch_8x/lucene/backward-codecs/src/java/org/apache/lucene/codecs

lucene/core/src/java/org/apache/lucene/index/DirectoryReader.java
```diff
@@ -104,6 +104,23 @@ public static DirectoryReader open(final IndexCommit commit) throws IOException
     return StandardDirectoryReader.open(commit.getDirectory(), commit);
   }
 
+  /**
+   * Expert: returns an IndexReader reading the index in the given {@link IndexCommit}. This method
+   * allows to open indices that were created wih a Lucene version older than N-1 provided that all
+   * all codecs for this index are available in the classpath and the segment file format used was
```
msokolov, line 110: "all all" -> "all"

lucene/core/src/java/org/apache/lucene/index/DirectoryReader.java
```diff
@@ -104,6 +104,23 @@ public static DirectoryReader open(final IndexCommit commit) throws IOException
...
+   * created with Lucene 7 or older. Users of this API must be aware that Lucene doesn't guarantee
```
msokolov, line 111: Instead of "Lucene 7" should we reference the `minSupportedMajorVersion` parameter here? Or .. is it  `Version.MIN_SUPPORTED_MAJOR`?

msokolov (commented):
just a few documentation nits

lucene/core/src/java/org/apache/lucene/index/DirectoryReader.java
s1monw: this is due to the fact that the segments info format only supports 7.0 and upwards

lucene/core/src/java/org/apache/lucene/index/DirectoryReader.java
```diff
@@ -104,6 +104,23 @@ public static DirectoryReader open(final IndexCommit commit) throws IOException
     return StandardDirectoryReader.open(commit.getDirectory(), commit);
   }
 
+  /**
+   * Expert: returns an IndexReader reading the index in the given {@link IndexCommit}. This method
```
mikemccand, line 108: s/`in`/`on`

lucene/core/src/java/org/apache/lucene/index/DirectoryReader.java
```diff
@@ -104,6 +104,23 @@ public static DirectoryReader open(final IndexCommit commit) throws IOException
...
+   * allows to open indices that were created wih a Lucene version older than N-1 provided that all
+   * codecs for this index are available in the classpath and the segment file format used was
+   * created with Lucene 7 or older. Users of this API must be aware that Lucene doesn't guarantee
+   * semantic compatibility for indices created with versions older than N-1. All backwards
+   * compatibility aside of the file format is optional and applied on a best effort basis.
```
mikemccand, line 113: s/`of`/`from`

lucene/core/src/java/org/apache/lucene/index/SegmentInfos.java
```diff
@@ -499,7 +515,7 @@ private static Codec readCodec(DataInput input) throws IOException {
         throw new IllegalArgumentException(
             "Could not load codec '"
                 + name
-                + "'.  Did you forget to add lucene-backward-codecs.jar?",
+                + "'. Did you forget to add lucene-backward-codecs.jar?",
```
mikemccand, line 518: Heh

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
```diff
@@ -1009,6 +1009,14 @@ public IndexWriter(Directory d, IndexWriterConfig conf) throws IOException {
         changed();
 
       } else if (reader != null) {
+        if (reader.segmentInfos.getIndexCreatedVersionMajor() < Version.MIN_SUPPORTED_MAJOR) {
```
mikemccand, line 1012: Hmm does `addIndexes` try to verify version of the incoming index is not too old?  We will keep doing that, right?  I.e. the only added best effort here is when directly opening an `IndexReader` you can (with this change) now ask that older versions be allowed.

mikemccand (commented):
Looks great!  I left some minorish comments.

lucene/core/src/java/org/apache/lucene/index/DirectoryReader.java
romseygeek: The javadoc should read 'Lucene 7 or newer' I think?

s1monw: 👍

lucene/core/src/java/org/apache/lucene/index/CheckIndex.java
```diff
@@ -3900,6 +3907,12 @@ public static Options parseOptions(String[] args) {
         }
         i++;
         opts.dirImpl = args[i];
+      } else if ("-min_version_created".equals(args[i])) {
+        if (i == args.length - 1) {
+          throw new IllegalArgumentException("ERROR: missing value for -min_version_created");
```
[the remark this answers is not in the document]
s1monw: I am all for old indices and remove this option. WDOT?

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
s1monw: yes we verify that it's the same major as the index we are adding to.

s1monw:
@mikemccand pushed changes

mikemccand (approved):
Looks great, thanks @s1monw!

s1monw:
I plan to merge this during the next 24 hours thanks for the reviews

PR #2214, opened 2021-01-18
LUCENE-9674:Faster advance on Vector Values

codaitya:
Currently the advance() function in the class Lucene90VectorReader does a
linear search for the target document.
This will make retrieving vectors for a sparse set of documents efficient.

# Description

Currently the advance() function in the class Lucene90VectorReader does a
linear search for the target document. This can be an expensive operation if we are searching for
sparse documents having vector fields.

# Solution

Implement a binary search over the "ordToDoc" array which will make the advance operation
take logarithmic time to search.

# Tests

Added testAdvance() in class TestVectorValues. It creates an index with gaps for vector fields and randomly calls advance.

# Checklist


lucene/core/src/java/org/apache/lucene/codecs/lucene90/Lucene90VectorReader.java
```diff
@@ -387,8 +387,40 @@ public int nextDoc() {
 
     @Override
     public int advance(int target) throws IOException {
-      // We could do better by log-binary search in ordToDoc, but this is never used
-      return slowAdvance(target);
+      assert docID() < target;
+      int idx = binarySearch(target);
```
s1monw, line 391: I wonder if we can simply use `Arrays.binarySearch(fieldEntry.ordToDoc, ord + 1 ,fieldEntry.ordToDoc.length - 1, target);` here?

s1monw, line 391: I wonder if we can simply use `Arrays.binarySearch(fieldEntry.ordToDoc, ord + 1 ,fieldEntry.ordToDoc.length - 1, target);` here?

msokolov: I think we could, we would pay the very small cost of decoding the negative values it returns (and it must encode them), when we don't care about the distinction of present/not-present, but probably that's not significant and we avoid the cost of maintaining our own binary search here. +1

codaitya: Yes I like the idea of not maintaining our own binary search. Updated the PR.

lucene/core/src/java/org/apache/lucene/codecs/lucene90/Lucene90VectorReader.java
```diff
@@ -386,9 +387,18 @@ public int nextDoc() {
     }
 
     @Override
-    public int advance(int target) throws IOException {
-      // We could do better by log-binary search in ordToDoc, but this is never used
-      return slowAdvance(target);
+    public int advance(int target) {
+      assert docID() < target;
+      ord = Arrays.binarySearch(fieldEntry.ordToDoc, target);
```
codaitya, line 392: I should have used `Arrays.binarySearch(fieldEntry.ordToDoc,  ord + 1,  fieldEntry.ordToDoc.length - 1,  target)`. Will update

lucene/core/src/java/org/apache/lucene/codecs/lucene90/Lucene90VectorReader.java
```diff
@@ -386,9 +387,18 @@ public int nextDoc() {
...
-    public int advance(int target) throws IOException {
-      // We could do better by log-binary search in ordToDoc, but this is never used
-      return slowAdvance(target);
+    public int advance(int target) {
+      assert docID() < target;
+      ord = Arrays.binarySearch(fieldEntry.ordToDoc, ord + 1, fieldEntry.ordToDoc.length, target);
+      if (ord < 0) {
+        ord = -(ord + 1);
+      }
+      if (ord == fieldEntry.ordToDoc.length) {
```
s1monw, line 396: should this read `ord >= fieldEntry.ordToDoc.length` to make sure we never grow beyond?

lucene/core/src/test/org/apache/lucene/index/TestVectorValues.java
```diff
@@ -815,4 +816,47 @@ public void testSearchStrategyIdentifiers() {
...
+      try (IndexWriter w = new IndexWriter(dir, createIndexWriterConfig())) {
+        int numdocs = 2000;
+        String fieldName = "field";
+        for (int i = 0; i < numdocs; i++) {
+          Document doc = new Document();
+          // randomly add a vector field
+          if (random().nextInt(4) == 3) {
+            doc.add(
+                new VectorField(
+                    fieldName, new float[4], VectorValues.SearchStrategy.DOT_PRODUCT_HNSW));
```
msokolov, line 831: Maybe use SearchStrategy.NONE here so we don't bother building a graph for this test case, which works the same with or without

lucene/core/src/test/org/apache/lucene/index/TestVectorValues.java
```diff
@@ -815,4 +816,47 @@ public void testSearchStrategyIdentifiers() {
     assertEquals(2, VectorValues.SearchStrategy.DOT_PRODUCT_HNSW.ordinal());
     assertEquals(3, VectorValues.SearchStrategy.values().length);
   }
+
+  public void testAdvance() throws Exception {
+    try (Directory dir = newDirectory()) {
+      try (IndexWriter w = new IndexWriter(dir, createIndexWriterConfig())) {
+        int numdocs = 2000;
```
msokolov, line 823: `atLeast(1000)` will randomize the number of docs

lucene/core/src/test/org/apache/lucene/index/TestVectorValues.java
```diff
@@ -815,4 +816,47 @@ public void testSearchStrategyIdentifiers() {
...
+                new VectorField(
+                    fieldName, new float[4], VectorValues.SearchStrategy.DOT_PRODUCT_HNSW));
+          }
+          w.addDocument(doc);
+        }
+        w.forceMerge(1);
+        try (IndexReader reader = w.getReader()) {
+          LeafReader r = getOnlyLeafReader(reader);
+          VectorValues vectorValues = r.getVectorValues(fieldName);
+          TreeSet<Integer> vectorDocs = new TreeSet<>();
```
msokolov, line 839: could we store the values in an `int[]` here? They are guaranteed to be returned in increasing order by `nextDoc` (although asserting that is so would be good). Then below we can maintain an index into the array and advance by one to get the next expected docid.

lucene/core/src/test/org/apache/lucene/index/TestVectorValues.java
```diff
@@ -815,4 +816,47 @@ public void testSearchStrategyIdentifiers() {
...
+          int cur = -1;
+          while (cur != NO_MORE_DOCS) {
+            cur = vectorValues.nextDoc();
+            vectorDocs.add(cur);
+          }
+          vectorValues = r.getVectorValues(fieldName);
+          for (int i = 0; i != NO_MORE_DOCS && i < numdocs; i++) {
+            // randomly advance to i
+            if (random().nextInt(4) == 3) {
+              vectorValues.advance(i);
```
msokolov, line 849: let's assert that the return value of `advance` is what we expect it to be

lucene/core/src/test/org/apache/lucene/index/TestVectorValues.java
```diff
@@ -815,4 +816,47 @@ public void testSearchStrategyIdentifiers() {
...
+              assertEquals((int) vectorDocs.higher(i - 1), vectorValues.docID());
+              if (vectorValues.docID() == NO_MORE_DOCS) {
+                break;
+              }
+              // ensure i is always greater than vectorValues.docID()
+              i = vectorValues.docID();
```
msokolov, line 855: The comment above is confusing since i is manifestly *not* greater than vectorValues.docID() , it's equal!

lucene/core/src/java/org/apache/lucene/codecs/lucene90/Lucene90VectorReader.java
codaitya: I think the range of values for `ord` will be between `[0,fieldEntry.ordToDoc.length]`, so we dont need a `>` check ?

lucene/core/src/test/org/apache/lucene/index/TestVectorValues.java
codaitya: We might need to advance the index by more than 1 sometimes ? But I can also do that with an array since it's always increasing. I will update to use array instead of TreeSet.

lucene/core/src/test/org/apache/lucene/index/TestVectorValues.java
codaitya: sure

lucene/core/src/test/org/apache/lucene/index/TestVectorValues.java
codaitya: Will use it in next revision

lucene/core/src/test/org/apache/lucene/index/TestVectorValues.java
codaitya: Sure

lucene/core/src/test/org/apache/lucene/index/TestVectorValues.java
codaitya: The intention was that in the next loop iteration `i` wil be incremented,  but yes this confusing. I will update

lucene/core/src/java/org/apache/lucene/codecs/lucene90/Lucene90VectorReader.java
msokolov: I guess it could save us from some mistake. We could add an `assert ord <= fieldEntry.ordToDoc.length`

codaitya: sure, will add a check.

msokolov (approved):
Thanks, Anand, looks good - I'll commit soon. Just one note about github PRs. It works a bit better if you push a new commit for each revision rather than force-push. It makes it easier to follow along with changes here, and then when we merge to apache/master, github will squash all the commits together.

msokolov:
>  It works a bit better if you push a new commit for each revision rather than force-push.
Oh hmm, I see you did do that! I;m not sure why I thought you had force-pushed... never mind!

PR #2216, opened 2021-01-18
SOLR-15085 Prevent EmbeddedSolrServer calling shutdown on a CoreConta…

timatbw:
…iner that was passed to it

# Description

Prevent EmbeddedSolrServer calling shutdown on a CoreContainer that was passed to it.

# Solution

Now keeping track of whether the CoreContainer was provided or created internally and only calling shutdown for internally-created instances.

# Tests

Modified appropriate test to confirm behaviour, and removed overrides used in existing tests to workaround this issue.

# Checklist

### DIFF
--- a/solr/core/src/java/org/apache/solr/client/solrj/embedded/EmbeddedSolrServer.java
+++ b/solr/core/src/java/org/apache/solr/client/solrj/embedded/EmbeddedSolrServer.java
@@ -71,6 +71,7 @@ public class EmbeddedSolrServer extends SolrClient {
   protected final String coreName;
   private final SolrRequestParsers _parser;
   private final RequestWriterSupplier supplier;
+  private boolean containerIsLocal = false;
 
   public enum RequestWriterSupplier {
     JavaBin(() -> new BinaryRequestWriter()), XML(() -> new RequestWriter());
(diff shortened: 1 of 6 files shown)


solr/core/src/java/org/apache/solr/client/solrj/embedded/EmbeddedSolrServer.java
```diff
@@ -71,6 +71,7 @@
   protected final String coreName;
   private final SolrRequestParsers _parser;
   private final RequestWriterSupplier supplier;
+  private boolean containerIsLocal = false;
```
madrob, line 74: Can this be final?

timatbw: I tried to do that, but it gets awkward because there's 5 constructors and one calls another which calls another. I'd have to refactor all of them to call a private constructor instead, to avoid changing the external constructor parameters. Do you think I should I do that?

madrob: I think we can pass null from the other constructors and in the final constructor if the cc is still null we create one and set local to true?

timatbw: That would be good, but there's no context available to create a cc because the original parameters (e.g. NodeConfig or solrHome Path) aren't in scope then.

I'm not that worried about this boolean flag because it's private to the class so we know where it's set.

dsmiley (approved):
Notwithstanding any PR feedback, definitely a +1 from me on the whole idea!

madrob:
Added upgrade notes, pushed to branch_8x and master! Thanks!

PR #2221, opened 2021-01-19
LUCENE-9669: Restore Lucene70Codec

s1monw:
This change restores the Lucene70Codec for file format compatibility of
indices that are created within the Lucene 7 major version. These indices
can be opened via an expert API on DirectoryReader in read-only mode. Changes
to these indices are prohibited and will be rejected by the IndexWriter.
In fact, IndexWriter will not open an index that is created with a major version
less than N-1 to the current major version.


mikemccand (approved):
Back from the dead!

jimczi (approved):
LGTM too

PR #2228, opened 2021-01-21
LUCENE-9684: Hunspell: support COMPOUNDRULE

donnerpeter:
# Description

Hunspell uses COMPOUNDRULE e.g. for ordinal numbers in en-US dictionary (1st, 42nd, etc)

# Solution

COMPOUNDRULE is a regexp-like pattern over word parts flags. I've reimplemented Hunspell's logic, which breaks the word into parts in different ways and checks whether any COMPOUNDRULE matches them (commit 1). To support uppercase, I had also to repeat `Stemmer`'s case variations (commit 3). While doing this, I discovered a bug in all-caps treatment where HIDDEN flag didn't play well with non-single-character flag formats, which I fixed (commit 2).

# Tests

All `compoundrule*` tests taken from Hunspell repository, plus a randomized test for flag serialization and deserialization.

# Checklist


donnerpeter:
Reviewing and merging commits separately might be a good idea.

lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/CompoundRule.java
```diff
@@ -0,0 +1,105 @@
...
+ * limitations under the License.
+ */
+package org.apache.lucene.analysis.hunspell;
+
+import java.util.List;
+import org.apache.lucene.util.BytesRef;
+import org.apache.lucene.util.IntsRef;
+
+class CompoundRule {
+  private final char[] data;
```
dweiss, line 24: I wonder if keeping char[] instead of the string itself buys us anything here. Strings can use more efficient storage than on newer JVMs (compact strings) and using charAt instead of array accesses is pretty much the same?

dweiss (approved):
This is much more complex logic to review. Looks ok to me but I didn't verify it against what hunspell actually does (I trust you did it :). The tests look reasonable so if there are no regressions I'm +1 for committing it in.

lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/CompoundRule.java
donnerpeter: Memory shouldn't be of much concern, as AFAIK there are usually very few rules. Array access is a bit shorter in code, so I'd prefer to leave it.

donnerpeter:
> I didn't verify it against what hunspell actually does (I trust you did it :).

Sure, I did :)

lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/CompoundRule.java
dweiss: ok.

donnerpeter:
How do we proceed here? Should anyone else review this?

dweiss:
No, I'll commit it in. I have an unrelated day job so expect some delays.

PR #2229, opened 2021-01-21
LUCENE-9688: Hunspell: consider prefix's continuation flags when applying suffix

donnerpeter:
# Description

aff:
SFX s Y 1
SFX s 0 s .

PFX p Y 1
PFX p 0 0/s .

dic:
calorie/p

"calories" should be accepted

# Solution

When applying suffix "s" (which expects "s" flag to be set), check not only the word itself (which only has "p") flag, but the continuation flags of the just applied prefix "p" which has "s" as its continuation.

# Tests

`TestDependencies` updated

# Checklist

### DIFF
--- a/lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/Stemmer.java
+++ b/lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/Stemmer.java
@@ -249,8 +249,8 @@ final class Stemmer {
    * @param previous previous affix that was removed (so we dont remove same one twice)
    * @param prevFlag Flag from a previous stemming step that need to be cross-checked with any
    *     affixes in this recursive step
-   * @param prefixFlag flag of the most inner removed prefix, so that when removing a suffix, it's
-   *     also checked against the word
+   * @param prefixId ID of the most inner removed prefix, so that when removing a suffix, it's also
+   *     checked against the word
    * @param recursionDepth current recursiondepth
    * @param doPrefix true if we should remove prefixes
    * @param doSuffix true if we should remove suffixes
@@ -268,7 +268,7 @@ final class Stemmer {
       int length,
       int previous,
       int prevFlag,
-      int prefixFlag,
+      int prefixId,
       int recursionDepth,
       boolean doPrefix,
       boolean doSuffix,
@@ -396,7 +396,7 @@ final class Stemmer {
                     strippedWord,
                     strippedWord.length,
                     suffix,
-                    prefixFlag,
+                    prefixId,
                     recursionDepth,
                     false,
                     circumfix,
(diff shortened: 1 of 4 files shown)


lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/Stemmer.java
```diff
@@ -588,6 +585,12 @@ private boolean checkCondition(
     return stems;
   }
 
+  private boolean isFlagAppendedByAffix(int affixId, char flag) {
+    if (affixId < 0) return false;
```
dweiss, line 589: Add curly braces. I know, it's verbose but some folks on this list even went as far as to require it (the formatter doesn't care).

lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/Stemmer.java
```diff
@@ -472,9 +472,9 @@ private boolean checkCondition(
    * @param strippedWord Word the affix has been removed and the strip added
    * @param length valid length of stripped word
    * @param affix HunspellAffix representing the affix rule itself
-   * @param prefixFlag when we already stripped a prefix, we cant simply recurse and check the
-   *     suffix, unless both are compatible so we must check dictionary form against both to add it
-   *     as a stem!
+   * @param prefixId when we already stripped a prefix, we cant simply recurse and check the suffix,
```
dweiss, line 475: cant -> can't (I realize it's not you, but maybe correct while we're here).

dweiss (approved):
Otherwise LGTM.

lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/Stemmer.java
donnerpeter: Curly braces and 2 more lines? Oh pity, I love conciseness so much :(

dweiss: The discussion of whether braces in single-line statements in if/while/for loops are needed or not
is like religion - you can't end it, once started... I'm just relaying what has been said 
on the mailing list after the formatter folded code like this:
```
if (foo)
  doSomething()
else
  doSomethingElse()
```

Privately, I don't have a strong opinion. Depends on the context where it's at, I guess.

donnerpeter: In that case I'd also put braces (although I've seen people who don't). But my case is "then-branch is a simple `return` on the same line". Are braces also required in this case? And separate lines, like this?
```
if (affixId < 0) {
  return;
}
```

dweiss: Required is not the right word... Perhaps... advised. :) You are entitled to your own opinion (at least as long as this passes precommit checks!) - like I said, I relayed what's been mentioned on the mailing list. Some people just put braces everywhere and they are used to this consistency, even if it adds verbosity.

donnerpeter: Would it be acceptable to reformat the code when such a demand appears, or after I'm finished with these improvements? These "some people" might never need to look into Hunspell code.

dweiss: I think consistency in those braces (over verbosity) is a good rule in general, but ok.

PR #2230, opened 2021-01-21
SOLR-15011: /admin/logging handler is configured logs to all nodes

NazerkeBS:
# Description

# Solution

# Tests

# Checklist

### DIFF
--- a/solr/CHANGES.txt
+++ b/solr/CHANGES.txt
@@ -69,6 +69,8 @@ Improvements
 * SOLR-14949: Docker: Ability to customize the FROM image when building.
   (Houston Putman)
 
+* SOLR-15011: /admin/logging handler should be able to configure logs on all nodes (Nazerke Seidan, David Smiley)
+
 Other Changes
 ----------------------
 * SOLR-14656: Autoscaling framework removed (Ishan Chattopadhyaya, noble, Ilan Ginzburg)
--- a/solr/CHANGES.txt
+++ b/solr/CHANGES.txt
@@ -69,7 +69,8 @@ Improvements
 * SOLR-14949: Docker: Ability to customize the FROM image when building.
   (Houston Putman)
 
-* SOLR-15011: /admin/logging handler should be able to configure logs on all nodes (Nazerke Seidan, David Smiley)
+* SOLR-15011: /admin/logging handler will now propagate setLevel (log threshold) to all nodes
+when told to. The admin UI now tells it to. (Nazerke Seidan, David Smiley)
 
 Other Changes
 ----------------------
--- a/solr/core/src/java/org/apache/solr/handler/admin/LoggingHandler.java
+++ b/solr/core/src/java/org/apache/solr/handler/admin/LoggingHandler.java
@@ -49,7 +49,7 @@ public class LoggingHandler extends RequestHandlerBase implements SolrCoreAware 
 
   @SuppressWarnings({"rawtypes"})
   private LogWatcher watcher;
-  private CoreContainer cc;
+  private final CoreContainer cc;
   
   public LoggingHandler(CoreContainer cc) {
     this.cc = cc;
@@ -57,16 +57,13 @@ public class LoggingHandler extends RequestHandlerBase implements SolrCoreAware 
   }
   
   public LoggingHandler() {
-    
+    this.cc = null;
   }
   
   @Override
   public void inform(SolrCore core) {
     if (watcher == null) {
       watcher = core.getCoreContainer().getLogging();
-    }
-    if (cc == null) {
-      cc = core.getCoreContainer();
     }
   }
 
--- a/solr/core/src/java/org/apache/solr/handler/admin/LoggingHandler.java
+++ b/solr/core/src/java/org/apache/solr/handler/admin/LoggingHandler.java
@@ -49,8 +49,10 @@ public class LoggingHandler extends RequestHandlerBase implements SolrCoreAware 
 
   @SuppressWarnings({"rawtypes"})
   private LogWatcher watcher;
+  private CoreContainer cc;
   
   public LoggingHandler(CoreContainer cc) {
+    this.cc = cc;
     this.watcher = cc.getLogging();
   }
   
@@ -62,6 +64,9 @@ public class LoggingHandler extends RequestHandlerBase implements SolrCoreAware 
   public void inform(SolrCore core) {
     if (watcher == null) {
       watcher = core.getCoreContainer().getLogging();
+    }
+    if (cc == null) {
+      cc = core.getCoreContainer();
     }
   }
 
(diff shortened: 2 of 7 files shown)


solr/core/src/java/org/apache/solr/handler/admin/LoggingHandler.java
```diff
@@ -63,6 +65,9 @@ public void inform(SolrCore core) {
     if (watcher == null) {
       watcher = core.getCoreContainer().getLogging();
     }
+    if (cc == null) {
```
dsmiley, line 68: does this actually happen (how?).  Otherwise, let's not and furthermore declare cc as final.

solr/core/src/java/org/apache/solr/handler/admin/LoggingHandler.java
```diff
@@ -151,6 +156,9 @@ public void handleRequestBody(SolrQueryRequest req, SolrQueryResponse rsp) throw
       rsp.add("loggers", info);
     }
     rsp.setHttpCaching(false);
+    if (cc != null && AdminHandlersProxy.maybeProxyToNodes(req, rsp, cc)) {
```
dsmiley, line 159: Why do this at the end of this method instead of at the top (along with disabling http caching)?  By doing it at the end, if nodes=FOO yet the current node is BAR, it would do work it shouldn't be doing; right?  Notice where the other callers of AdminHandlersProxy.maybeProxyToNodes put their logic.

solr/webapp/web/js/angular/services.js
```diff
@@ -58,7 +58,7 @@ solrAdminServices.factory('System',
   }])
 .factory('Logging',
   ['$resource', function($resource) {
-    return $resource('admin/info/logging', {'wt':'json', '_':Date.now()}, {
+    return $resource('admin/info/logging', {'wt':'json', 'nodes': 'all', '_':Date.now()}, {
```
dsmiley, line 61: I'm very unfamiliar with the admin UI code structure.  Can you explain why nodes=all needed to be set in two places?

dsmiley (changes requested):
Could use a test.  See AdminHandlersProxyTest where the other users of this class test this functionality.

solr/core/src/java/org/apache/solr/handler/admin/LoggingHandler.java
NazerkeBS: SystemInfoHandler is doing similar to this logic;

dsmiley: SIH is doing what I suggest, which is different than what the PR is doing.

solr/core/src/test/org/apache/solr/handler/admin/AdminHandlersProxyTest.java
```diff
@@ -91,6 +98,32 @@ public void proxyMetricsHandlerAllNodes() throws IOException, SolrServerExceptio
...
+
+    Set<String> nodes = solrClient.getClusterStateProvider().getLiveNodes();
+    nodes.forEach(node -> {
+      mparams.clear();
+      mparams.set("nodes", node);
+      req = new GenericSolrRequest(SolrRequest.METHOD.GET, "/admin/logging", mparams);
+      try {
+        rsp = req.process(solrClient, "collection");
+      } catch (Exception e) {
+        fail("Exception while proxying request to node " + node);
```
dsmiley, line 120: If above you used a standard for loop, you wouldn't be forced to catch this exception, and it'd be more succinct.  Personally, I only use Collection.forEach lambda for a one-liner or something short and no exception propagation concern.

solr/core/src/test/org/apache/solr/handler/admin/AdminHandlersProxyTest.java
```diff
@@ -91,6 +98,32 @@ public void proxyMetricsHandlerAllNodes() throws IOException, SolrServerExceptio
     assertNotNull(((NamedList)nl.get(nl.getName(1))).get("metrics"));
   }
 
+  @Test
+  public void proxyLoggingHandlerAllNodes() throws IOException, SolrServerException {
+    CollectionAdminRequest.createCollection("collection", "conf", 2, 2).process(solrClient);
+
+    mparams = new ModifiableSolrParams();
```
dsmiley, line 105: BTW, Solr test infrastructure (SolrTestCaseJ4) defines a params(...) method that is more succinct.  This is fine though!

dsmiley (changes requested):
Real close now. Just some minor stuff.

solr/webapp/web/js/angular/services.js
dsmiley: @NazerkeBS ?

NazerkeBS: @dsmiley, I set nodes=all to only one place which is in services.js file

dsmiley: Okay; I see you improved that from an earlier commit.

@janhoy maybe you could take a look at this please. Should nodes=all should only be added when setLevel is submitted, not for all logging handler interactions.  Do you think this is done correctly here?

janhoy: I think nodes=all should be added only to the `setLevel` call. Should be easy to check in browser debugger that it sends correct request.

dsmiley: I verified this is a problem now!  Thanks for the tip Jan; I forgot how easy it is to verify :-)

NazerkeBS: Thank Jan and David! Just updated the PR.

solr/webapp/web/js/angular/services.js
```diff
@@ -58,10 +58,10 @@ solrAdminServices.factory('System',
   }])
 .factory('Logging',
   ['$resource', function($resource) {
-    return $resource('admin/info/logging', {'wt':'json', 'nodes': 'all', '_':Date.now()}, {
+    return $resource('admin/info/logging', {'wt':'json', '_':Date.now()}, {
       "events": {params: {since:'0'}},
       "levels": {},
-      "setLevel": {}
+      "setLevel": {'nodes': 'all'}
```
dsmiley, line 64: The JavaScript framework we use here is completely unknown to me.  Does this change make sense to you @janhoy ?

janhoy: It’s Angular 1. Looks right but I’m not sure if there should be a `params: ` there as well. Testing will tell..

NazerkeBS: I debugged it on Admin UI and it works without `params`. But I updated my PR to use `params` as other handlers also use `params` in this file .

solr/CHANGES.txt
```diff
@@ -69,6 +69,8 @@ Improvements
 * SOLR-14949: Docker: Ability to customize the FROM image when building.
   (Houston Putman)
 
+* SOLR-15011: /admin/logging handler should be able to configure logs on all nodes (Nazerke Seidan, David Smiley)
```
dsmiley, line 72:
```suggestion
* SOLR-15011: /admin/logging handler will now propagate setLevel (log threshold) to all nodes
when told to. The admin UI now tells it to. (Nazerke Seidan, David Smiley)
```

dsmiley (approved):
LGTM; I'll merge this tomorrow.  Please confirm with me that "gw check" passes.

janhoy (approved):

PR #2238, opened 2021-01-22
LUCENE-9693: Hunspell: check that all flags are > 0 and fit char range

donnerpeter:
# Description

Hunspell: check that all flags are > 0 and fit char range, as Hunspell does

# Solution

1. Add a corresponding assertion
2. Fix HIDDEN_FLAG serialization bug found by this assertion
3. Use char instead of int to store and pass flags around, because now we can use 0 for an unset flag instead of -1

# Tests

No new tests, but some tests failed after implementing step 1 and were fixed in step 2

# Checklist

### DIFF
--- a/lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/Dictionary.java
+++ b/lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/Dictionary.java
@@ -76,6 +76,7 @@ public class Dictionary {
 
   static final char[] NOFLAGS = new char[0];
 
+  static final int FLAG_UNSET = 0;
   private static final int DEFAULT_FLAGS = 65510;
   private static final char HIDDEN_FLAG = (char) 65511; // called 'ONLYUPCASEFLAG' in Hunspell
 
@@ -457,20 +458,6 @@ public class Dictionary {
     }
     assert currentIndex == seenStrips.size();
     stripOffsets[currentIndex] = currentOffset;
-  }
-
-  private String singleArgument(LineNumberReader reader, String line) throws ParseException {
-    return splitBySpace(reader, line, 2)[1];
-  }
-
-  private String[] splitBySpace(LineNumberReader reader, String line, int expectedParts)
-      throws ParseException {
-    String[] parts = line.split("\\s+");
-    if (parts.length < expectedParts
-        || parts.length > expectedParts && !parts[expectedParts].startsWith("#")) {
-      throw new ParseException("Invalid syntax", reader.getLineNumber());
-    }
-    return parts;
   }
 
   private List<CompoundRule> parseCompoundRules(LineNumberReader reader, int num)
@@ -1235,7 +1222,7 @@ public class Dictionary {
   }
 
   boolean isForbiddenWord(char[] word, int length, BytesRef scratch) {
-    if (forbiddenword != 0) {
+    if (forbiddenword != FLAG_UNSET) {
       IntsRef forms = lookupWord(word, 0, length);
       return forms != null && hasFlag(forms, forbiddenword, scratch);
     }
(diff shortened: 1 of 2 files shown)


lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/Dictionary.java
```diff
@@ -1299,7 +1314,12 @@ void appendFlag(char flag, StringBuilder to) {
         if (replacement.isEmpty()) {
           continue;
         }
-        flags[upto++] = (char) Integer.parseInt(replacement);
+        int flag = Integer.parseInt(replacement);
+        if (flag == 0 || flag >= Character.MAX_VALUE) { // read default flags as well
+          throw new IllegalArgumentException(
```
dweiss, line 1319: It'd be great to be consistent with exceptions when parsing input - sometimes it's ParsingException, here it's IAE.

lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/Dictionary.java
```diff
@@ -1310,10 +1330,8 @@ void appendFlag(char flag, StringBuilder to) {
 
     @Override
     void appendFlag(char flag, StringBuilder to) {
```
dweiss, line 1332: Is this intentional? Because it changes the logic of concatenation (always leaving the trailing comma). I liked the previous version better (always leaving the output neat).

lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/Stemmer.java
```diff
@@ -588,7 +577,7 @@ private boolean checkCondition(
   }
 
   private boolean isFlagAppendedByAffix(int affixId, char flag) {
-    if (affixId < 0) return false;
+    if (affixId < 0 || flag == 0) return false;
```
dweiss, line 580: Wouldn't it be cleaner to add a constant alias (static variable) FLAG_UNSET for 0 and replace it throughout the code where it compares to zero? You've changed it from -1 to 0 but it really doesn't make it any clearer that it's a "default" unset state. I think it would benefit from being more verbose here.

lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/Dictionary.java
donnerpeter: `ParseException` needs some `errorOffset` obligatorily (which is dubiously filled here with the current line number), and it's not available in this method, and not all callers have anything meaningful to pass there. For consistency, we could replace `ParseException` with something less choosy :)

lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/Dictionary.java
donnerpeter: Yes, that's `some tests failed after implementing step 1 and were fixed in step 2`. However nice it seemed, it was wrong, because other flags were appended after this one without any comma. Trailing commas are no problem, as empty flags are skipped in the previous method.

lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/Stemmer.java
donnerpeter: A good idea, thanks!

lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/Dictionary.java
dweiss: Eh, I was afraid of that. It'd be good to consolidate it at some point.

lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/Dictionary.java
dweiss: Oh, so something else (other than this method) appends to that stringbuilder? Maybe those places should be fixed instead? I don't have all of the code in front of me, so the question may be naive.

donnerpeter: Yes. Currently it's just one place which definitely appends no flags before this one, and may append some flags after this, so the implementation is tied to that.

PR #2239, opened 2021-01-23
LUCENE-9695: don't merge deleted vectors

msokolov:


lucene/core/src/java/org/apache/lucene/codecs/VectorWriter.java
```diff
@@ -194,6 +197,7 @@ public int nextDoc() throws IOException {
       current = docIdMerger.next();
       if (current == null) {
         docId = NO_MORE_DOCS;
+        size = ord;
```
mikemccand, line 200: Maybe also add a comment here so we understand the sneakiness in context?

lucene/core/src/test/org/apache/lucene/index/TestVectorValues.java
```diff
@@ -748,29 +751,107 @@ public void testRandom() throws Exception {
...
+      }
+    }
+  }
+
+  /**
+   * Index random vectors, sometimes skipping documents, sometimes deleting a document, sometimes
+   * merging, sometimes sorting the index, and verify that the expected values can be read back
+   * consistently.
+   */
+  public void testRandom2() throws Exception {
```
mikemccand, line 773: Yay!

lucene/core/src/test/org/apache/lucene/index/TestKnnGraph.java
```diff
@@ -124,21 +126,84 @@ public void testMerge() throws Exception {
     }
   }
 
-  private void dumpGraph(KnnGraphValues values, int size) throws IOException {
+  /**
+   * Verify that we get the *same* graph by indexing one segment as we do by indexing two segments
+   * and merging.
+   */
+  public void testMergeProducesSameGraph() throws Exception {
```
mikemccand, line 133: Nice test :)

mikemccand (commented):
Looks great -- I left some small comments.  I am really curious if this explains Anton's scary corruption looking exception from `luceneutil`.

lucene/core/src/test/org/apache/lucene/index/TestVectorValues.java
msokolov: I forgot why I had done this (there is already a testRandom), so I added some comments explaining how it differs.

lucene/core/src/java/org/apache/lucene/codecs/VectorWriter.java
msokolov: Thanks, yes I did.

lucene/core/src/java/org/apache/lucene/codecs/VectorWriter.java
```diff
@@ -153,13 +153,12 @@ public int nextDoc() throws IOException {
...
-    private final int size;
+    private int size;
 
     private int docId;
     private VectorValuesSub current;
-    // For each doc with a vector, record its ord in the segments being merged. This enables random
-    // access into the
-    // unmerged segments using the ords from the merged segment.
+    /* For each doc with a vector, record its ord in the segments being merged. This enables random access into the unmerged segments using the ords from the merged segment.
+     */
```
msokolov, line 161: Hmmm I thought spotless would wrap this line, but it doesn't seem to complain about it

lucene/core/src/test/org/apache/lucene/index/TestKnnGraph.java
msokolov: thanks, yes this flushed out the two problems I saw, so I'm pretty confident they are fixed now, after running this a few 100 times. I had also wanted to add a test asserting that KNN search precision remains above some threshold, but sadly with random vectors, it would always eventually fail, even though mostly it would succeed, so not a very useful unit test and I removed it. Probably we can add something to luceneutil

lucene/core/src/test/org/apache/lucene/util/hnsw/KnnGraphTester.java
```diff
@@ -578,6 +572,8 @@ private int createIndex(Path docsPath, Path indexPath) throws IOException {
     IndexWriterConfig iwc = new IndexWriterConfig().setOpenMode(IndexWriterConfig.OpenMode.CREATE);
     // iwc.setMergePolicy(NoMergePolicy.INSTANCE);
     iwc.setRAMBufferSizeMB(1994d);
+    iwc.setMaxBufferedDocs(10000);
```
msokolov, line 575: Oh I did not mean to include this change here. Probably we will want to have some command line parameter to control this, but for now having the default be to index a large segment is probably better

mikemccand (approved):

PR #2240, opened 2021-01-25
LUCENE-9575 Provide a producer for PatternTypingRule in TestRandomChains 

gus-asf:
to fix failure on seed 65EA739C95F40313


gus-asf:
need to remake to account for quickfix.

mikemccand:
> need to remake to account for quickfix.

What does that mean?

gus-asf:
While the build (github action) was running for this PR, @jimczi applied a quick fix directly. Though github would have happily merged this PR that would have left two producer implementations (at different ends of the code block where they were created), and it was easier/faster to just remake the PR than wait for the build to finish on this and then update it.

mikemccand:
Phew, OK thanks for the explanation @gus-asf!  And thanks for fixing the failing seed so quickly :)

PR #2243, opened 2021-01-25
LUCENE-9698: Hunspell: reuse char[] when possible when stripping affix

donnerpeter:
# Description

There's no need to allocate another char[] if we can analyze a sub-array of what we already have

# Solution

In addition to `char[]` and `int length`, pass `int offset` everywhere, and adjust offset/length instead of allocating a new array, when an affix is removed and nothing is added in its place.

# Tests

No behavior change, no new tests

# Checklist


lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/Stemmer.java
```diff
@@ -403,7 +408,12 @@ private CharsRef newStem(char[] buffer, int length, IntsRef forms, int formID) {
     return stems;
   }
 
-  private char[] stripAffix(char[] word, int length, int affixLen, int affix, boolean isPrefix) {
+  /**
+   * @return null if affix conditions isn't met; the same char[] if the affix has no strip data and
```
dweiss, line 412: I'd make it clearer that it's "the same *reference* to char[] if the affix....". I did wonder a bit why you're comparing array refs directly (above) for a while. Also, not sure "de-affixation" is an English word (although I'm not a native English speaker too); "[...] after affix stripping" would be more legible, maybe?

donnerpeter: "stripping" has a bit different meaning in Hunspell. Googling for "deaffixation" yields some results, some even seem to be by a native speaker (https://books.google.de/books?id=4A44AAAAIAAJ&pg=PA24&lpg=PA24&dq=deaffixation&source=bl&ots=bjtSTczny5&sig=ACfU3U1ujhDJpEFTR3BD_MiHYqgMJKbHMQ&hl=en&sa=X&ved=2ahUKEwjNw_uQurnuAhWvx4UKHXJoCQ0Q6AEwAnoECAUQAg#v=onepage&q=deaffixation&f=false), but rather few, indeed. How about "affix removal"?

PR #2245, opened 2021-01-26
LUCENE-9322: Move old field infos format to backwards-codecs.

jtibshirani:
We introduced a new `Lucene90FieldInfosFormat`, so the old
`Lucene60FieldInfosFormat` should live in backwards-codecs.


lucene/backward-codecs/src/java/org/apache/lucene/backward_codecs/lucene60/Lucene60FieldInfosFormat.java
```diff
@@ -311,6 +312,11 @@ private static IndexOptions getIndexOptions(IndexInput input, byte b) throws IOE
     }
   }
 
+  /**
+   * Note: although this format is only used on older versions, we need to keep the write logic
```
jtibshirani, line 316: I hope this assumption is accurate, would appreciate someone double-checking it.

jtibshirani:
Thanks @iverase for pointing this out.

lucene/backward-codecs/src/java/org/apache/lucene/backward_codecs/lucene60/Lucene60FieldInfosFormat.java
iverase: In theory it is possible to try to add a new vector field to one segment create with this format, I guess it would break?

lucene/backward-codecs/src/test/org/apache/lucene/backward_codecs/lucene60/TestLucene60FieldInfosFormat.java
```diff
@@ -14,17 +14,15 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.apache.lucene.codecs.lucene50;
```
mikemccand, line 17: Hmm, why did this previously live under `lucene50` sub-package if it was `lucene60` format?

lucene/backward-codecs/src/java/org/apache/lucene/backward_codecs/lucene60/Lucene60FieldInfosFormat.java
mikemccand: I don't see how a new vector field would ever be written with this format?  That should use the latest (90) field infos format.

Rather, I think the case that requires writing is when user opens old index/segment, that had used this `FieldInfosFormat`, and then user updates some doc values in the docs in this segment, and then `IndexWriter` needs to write that new doc values field, along with a new `.fnm` (field infos) file.

mikemccand (approved):
Thanks @jtibshirani!

I feel like we need some simple "best practices" documentation somewhere on how to bring new Codec/XXXFormats into existence, where/when to move the old ones etc.

lucene/backward-codecs/src/java/org/apache/lucene/backward_codecs/lucene60/Lucene60FieldInfosFormat.java
iverase: Thanks @mikemccand for the clarification.

I was thinking what would happen if a user opens an old index/segment and adds a new document that contains a vector field.  You are suggesting that in that case a totally new segment would be created. This method is only called in the case you mention above.

mikemccand: YW @iverase.

Right, that case should write a whole new segment, using the latest Codec, holding the just indexed document with vector field.

lucene/backward-codecs/src/test/org/apache/lucene/backward_codecs/lucene60/TestLucene60FieldInfosFormat.java
jtibshirani: This indeed seems like a mistake.

jtibshirani:
> I feel like we need some simple "best practices" documentation somewhere on how to bring new Codec/XXXFormats into existence, where/when to move the old ones etc.

This is a good idea, it took some digging to understand what the process/ strategy was. I'll plan to add some docs.

mikemccand:
> This is a good idea, it took some digging to understand what the process/ strategy was. I'll plan to add some docs.

Thanks @jtibshirani!

jtibshirani:
@mikemccand I opened https://github.com/apache/lucene-solr/pull/2395, it'd be great to get your feedback. I didn't include a section on adding a new format, because this felt like a once-in-a-few-years event? I'm also curious if the repo is the right place for such docs.

PR #2246, opened 2021-01-26
LUCENE-9694: New tool for creating a deterministic index

zhaih:
# Description

Create a new tool `IndexRearranger`, which could rearrange a built index concurrently to desired segment number and document distribution

# Solution

Essentially combines `IndexWriter.addIndexes` and `FilterCodecReader` to select only certain documents into 1 segment

# Tests

Added one unit test testing rearranger.

# Checklist

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -252,7 +252,7 @@ Other
 New Features
 ---------------------
 
-* LUCENE-9694: New tool for creating a deterministic index to enabling benchmarking changes
+* LUCENE-9694: New tool for creating a deterministic index to enable benchmarking changes
   on a consistent multi-segment index even when they require re-indexing. (Haoyu Zhai)
 
 ======================= Lucene 8.8.0 =======================
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -247,6 +247,13 @@ Other
 * LUCENE-9627: Remove unused Lucene50FieldInfosFormat codec and small refactor some codecs
   to separate reading header/footer from reading content of the file. (Ignacio Vera)
 
+======================= Lucene 8.9.0 =======================
+
+New Features
+---------------------
+
+* LUCENE-9694: New tool for creating a deterministic index (Haoyu Zhai)
+
 ======================= Lucene 8.8.0 =======================
 
 New Features
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -252,7 +252,8 @@ Other
 New Features
 ---------------------
 
-* LUCENE-9694: New tool for creating a deterministic index (Haoyu Zhai)
+* LUCENE-9694: New tool for creating a deterministic index to enabling benchmarking changes
+  on a consistent multi-segment index even when they require re-indexing. (Haoyu Zhai)
 
 ======================= Lucene 8.8.0 =======================
 
--- a/lucene/misc/src/java/org/apache/lucene/misc/index/BinaryDocValueSelector.java
+++ b/lucene/misc/src/java/org/apache/lucene/misc/index/BinaryDocValueSelector.java
@@ -1,0 +1,84 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.lucene.misc.index;
+
+import java.io.IOException;
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import org.apache.lucene.index.BinaryDocValues;
+import org.apache.lucene.index.CodecReader;
+import org.apache.lucene.index.DirectoryReader;
+import org.apache.lucene.index.IndexReader;
+import org.apache.lucene.index.LeafReaderContext;
+import org.apache.lucene.store.Directory;
+import org.apache.lucene.util.BitSet;
+import org.apache.lucene.util.Bits;
+import org.apache.lucene.util.FixedBitSet;
+
+/** Select documents using binary doc values */
+public class BinaryDocValueSelector implements IndexRearranger.DocumentSelector, Serializable {
+
+  private final String field;
+  private final HashSet<String> keySet;
+
+  public BinaryDocValueSelector(String field, HashSet<String> keySet) {
+    this.field = field;
+    this.keySet = keySet;
+  }
+
+  @Override
+  public BitSet getFilteredLiveDocs(CodecReader reader) throws IOException {
+    BinaryDocValues binaryDocValues = reader.getBinaryDocValues(field);
+    Bits oldLiveDocs = reader.getLiveDocs();
+    FixedBitSet bits = new FixedBitSet(reader.maxDoc());
+    for (int i = 0; i < reader.maxDoc(); i++) {
+      if (oldLiveDocs != null && oldLiveDocs.get(i) == false) {
+        continue;
+      }
+      if (binaryDocValues.advanceExact(i)
+          && keySet.contains(binaryDocValues.binaryValue().utf8ToString())) {
+        bits.set(i);
+      }
+    }
+    return bits;
+  }
+
+  public static List<IndexRearranger.DocumentSelector> createFromExistIndex(
+      String field, Directory directory) throws IOException {
+    List<IndexRearranger.DocumentSelector> selectors = new ArrayList<>();
+    try (IndexReader reader = DirectoryReader.open(directory)) {
+      for (LeafReaderContext context : reader.leaves()) {
+        HashSet<String> keySet = new HashSet<>();
+        Bits liveDocs = context.reader().getLiveDocs();
+        BinaryDocValues binaryDocValues = context.reader().getBinaryDocValues(field);
+        for (int i = 0; i < context.reader().maxDoc(); i++) {
+          if (liveDocs != null && liveDocs.get(i) == false) {
+            continue;
+          }
+          if (binaryDocValues.advanceExact(i)) {
+            keySet.add(binaryDocValues.binaryValue().utf8ToString());
+          }
+        }
+        selectors.add(new BinaryDocValueSelector(field, keySet));
+      }
+    }
+    return selectors;
+  }
+}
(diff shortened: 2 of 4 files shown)


lucene/CHANGES.txt
```diff
@@ -11,6 +11,8 @@ New Features
 
 * LUCENE-9004: Approximate nearest vector search via NSW graphs
 
+* LUCENE-9694: New tool for creating a deterministic index (Haoyu Zhai)
```
mikemccand, line 14: Probably we could also back-port this to Lucene 8.9, right?  We are not using any Lucene 9.x only APIs?

If so, can you move this CHANGES entry down to that section?  (Hmm, if it does not yet exist, you can create a skeletal one).

lucene/misc/src/java/org/apache/lucene/misc/index/IndexRearranger.java
```diff
@@ -0,0 +1,132 @@
...
+      ArrayList<Future<Void>> futures = new ArrayList<>();
+      for (DocumentSelector record : documentSelectors) {
+        Callable<Void> addSegment =
+            () -> {
+              addOneSegment(writer, reader, record);
+              return null;
+            };
+        futures.add(executor.submit(addSegment));
+      }
+      for (Future<Void> future : futures) {
```
mikemccand, line 78: Yay, concurrent!  (thread-per-segment).  And ironically if you "optimize" (force-merge to one segment) your index, you lose all concurrency :) That's fine ... just don't force-merge!

lucene/misc/src/java/org/apache/lucene/misc/index/IndexRearranger.java
```diff
@@ -0,0 +1,132 @@
...
+public class IndexRearranger {
+  protected final Directory input, output;
+  protected final IndexWriterConfig config;
+  protected final List<DocumentSelector> documentSelectors;
+
+  public IndexRearranger(
+      Directory input,
+      Directory output,
+      IndexWriterConfig config,
+      List<DocumentSelector> documentSelectors) {
```
mikemccand, line 53: Maybe we could (followon PR maybe) have some sort of utility class implementing `DocumentSelector` by opening an existing index, crawling its docids to segments based on some specified primary key field or so?

Then this command-line tool could take a pointer to source index (to extract `DocumentSelector` per segment), name of primary key field, a pointer to input index to rewrite, and a pointer to filesystem where the rearranged index should be written?

lucene/misc/src/test/org/apache/lucene/misc/index/TestIndexRearranger.java
```diff
@@ -0,0 +1,144 @@
...
+      } else {
+        assertEquals(0, numericDocValues.longValue() % 2);
+      }
+    }
+    reader.close();
+    inputDir.close();
+    outputDir.close();
+  }
+
+  private class OddDocSelector implements IndexRearranger.DocumentSelector {
```
mikemccand, line 107: This looks very similar to "shard splitting" implementations!

I think an interesting `DocumentSelector` might be one which assigns documents into a "perfect" logarithmic staircase according to some base and exponent, e.g. target 7 segments with 1M docs, 7 segments with 100K docs, 7 segments with 10K docs.  This would allow us to take any index and rearrange it into arbitrary multi-segment situations.  Fun!

lucene/misc/src/java/org/apache/lucene/misc/index/IndexRearranger.java
```diff
@@ -0,0 +1,132 @@
...
+        futures.add(executor.submit(addSegment));
+      }
+      for (Future<Void> future : futures) {
+        future.get();
+      }
+      executor.shutdown();
+    }
+  }
+
+  private static void addOneSegment(IndexWriter writer, IndexReader reader, DocumentSelector record)
```
mikemccand, line 85: Hmm maybe rename `record` to `selector`?

lucene/misc/src/java/org/apache/lucene/misc/index/IndexRearranger.java
```diff
@@ -0,0 +1,132 @@
...
+import org.apache.lucene.index.IndexWriterConfig;
+import org.apache.lucene.index.LeafReaderContext;
+import org.apache.lucene.index.NoMergePolicy;
+import org.apache.lucene.store.Directory;
+import org.apache.lucene.util.BitSet;
+import org.apache.lucene.util.Bits;
+import org.apache.lucene.util.NamedThreadFactory;
+
+/**
+ * Copy and rearrange index according to document selectors, from input dir to output dir Length of
```
mikemccand, line 41: Add period after `output dir` before `Length`?

lucene/misc/src/java/org/apache/lucene/misc/index/IndexRearranger.java
```diff
@@ -0,0 +1,132 @@
...
+ * documentSelectors determines how many segments there will be
+ */
+public class IndexRearranger {
```
mikemccand, line 44: Maybe add a `// TODO` here and roughly describe the alternative solution @s1monw had proposed on the dev list thread?  That solution sounded compelling but a bit tricky to implement by wrangling `IndexWriter` to flush/merge in exactly the right way, but, it would probably be faster than this two-phased approach.

mikemccand (approved):
This looks awesome; I left some small comments.

I am really curious how this performs, e.g. what is the time to rearrange versus original time to build the full initial index.

I think we need a followon PR that implements a `DocumentSeletor` by looking at an existing index and a primary key field in that index and copying that arrangement of documents to segments?  But that can come as another PR -- this is a great start.

Finally, could you open a followon issue in `luceneutil` to fold this into Lucene's benchmarking?  This is a much faster option than the way our benchmark tooling builds deterministic indices today (entirely single threaded indexing and merging).

lucene/CHANGES.txt
zhaih: Good point. I'll create an 8.9 entry instead

lucene/misc/src/java/org/apache/lucene/misc/index/IndexRearranger.java
zhaih: Good idea. I didn't include that part for keeping generality, but I think it is also worth implementing an example `DocumentSelector` as you suggested, I'll add one in this PR.

lucene/misc/src/test/org/apache/lucene/misc/index/TestIndexRearranger.java
zhaih: Yeah, in extreme case we might be able to optimize the search performance by changing document distribution

lucene/misc/src/java/org/apache/lucene/misc/index/IndexRearranger.java
zhaih: Ah good catch. Previously I named `DocumentSelector` as `SegmentRecord` and forgot to change this one after renaming.

lucene/misc/src/java/org/apache/lucene/misc/index/IndexRearranger.java
zhaih: Good idea!

lucene/CHANGES.txt
```diff
@@ -247,6 +247,13 @@ Other
 * LUCENE-9627: Remove unused Lucene50FieldInfosFormat codec and small refactor some codecs
   to separate reading header/footer from reading content of the file. (Ignacio Vera)
 
+======================= Lucene 8.9.0 =======================
+
+New Features
+---------------------
+
+* LUCENE-9694: New tool for creating a deterministic index (Haoyu Zhai)
```
mikemccand, line 255: Thanks for adding `CHANGES` entry!

Could we beef it up a bit, e.g.:

```
LUCENE-9694: New tool for creating a deterministic index to enabling benchmarking changes on a consistent multi-segment index even when they require re-indexing.
```

Or so?

lucene/misc/src/java/org/apache/lucene/misc/index/BinaryDocValueSelector.java
```diff
@@ -0,0 +1,84 @@
...
+import org.apache.lucene.util.Bits;
+import org.apache.lucene.util.FixedBitSet;
+
+/** Select documents using binary doc values */
+public class BinaryDocValueSelector implements IndexRearranger.DocumentSelector, Serializable {
+
+  private final String field;
+  private final HashSet<String> keySet;
+
+  public BinaryDocValueSelector(String field, HashSet<String> keySet) {
```
mikemccand, line 41: We could later (in follow-on issue) switch to Lucene's `CharArraySet` if heap usage of this possibly large set ever becomes a problem ... but it is not urgent since this tool need not be optimized.

lucene/misc/src/java/org/apache/lucene/misc/index/BinaryDocValueSelector.java
```diff
@@ -0,0 +1,84 @@
...
+      }
+      if (binaryDocValues.advanceExact(i)
+          && keySet.contains(binaryDocValues.binaryValue().utf8ToString())) {
+        bits.set(i);
+      }
+    }
+    return bits;
+  }
+
+  public static List<IndexRearranger.DocumentSelector> createFromExistIndex(
```
mikemccand, line 63: Maybe `createFromExistingIndex`?

lucene/misc/src/java/org/apache/lucene/misc/index/BinaryDocValueSelector.java
```diff
@@ -0,0 +1,84 @@
...
+        HashSet<String> keySet = new HashSet<>();
+        Bits liveDocs = context.reader().getLiveDocs();
+        BinaryDocValues binaryDocValues = context.reader().getBinaryDocValues(field);
+        for (int i = 0; i < context.reader().maxDoc(); i++) {
+          if (liveDocs != null && liveDocs.get(i) == false) {
+            continue;
+          }
+          if (binaryDocValues.advanceExact(i)) {
+            keySet.add(binaryDocValues.binaryValue().utf8ToString());
+          }
```
mikemccand, line 77: Maybe add `else throw exception`?

lucene/misc/src/java/org/apache/lucene/misc/index/IndexRearranger.java
mikemccand: Awesome, thanks for adding this!

mikemccand (commented):
This looks great @zhaih!

I left a few minor comments...

lucene/CHANGES.txt
zhaih: Sure, I'll add that, thanks!

lucene/misc/src/java/org/apache/lucene/misc/index/BinaryDocValueSelector.java
zhaih: ++

lucene/misc/src/java/org/apache/lucene/misc/index/BinaryDocValueSelector.java
zhaih: Good idea

lucene/misc/src/java/org/apache/lucene/misc/index/BinaryDocValueSelector.java
zhaih: Good idea!

lucene/CHANGES.txt
```diff
@@ -252,7 +252,8 @@ Other
 New Features
 ---------------------
 
-* LUCENE-9694: New tool for creating a deterministic index (Haoyu Zhai)
+* LUCENE-9694: New tool for creating a deterministic index to enabling benchmarking changes
```
mikemccand, line 255: s/`enabling`/`enable`

mikemccand (approved):
One small typo.

mikemccand:
Thank you @zhaih!  This will be a very useful tool for realistic benchmarking!

PR #2247, opened 2021-01-26
WIP: LUCENE-9476 Add basic functionality, basic tests

gautamworah96:
# Description

In [LUCENE-9450](https://issues.apache.org/jira/browse/LUCENE-9450) we switched the Taxonomy index from Stored Fields to `BinaryDocValues.` In the resulting implementation of the `getPath` code, we create a new `BinaryDocValues`'s values instance for each ordinal. 
It may happen that we may traverse over the same nodes over and over again if the `getPath` API is called multiple times for ordinals in the same segment/with the same `readerIndex`.

This PR takes advantage of that fact by sorting ordinals and then trying to find out if some of the ordinals are present in the same segment/have the same `readerIndex` (by trying to `advanceExact` to the correct position and not failing) thereby allowing us to reuse the previous `BinaryDocValues` object. 

# Solution

Steps:
1. Sort all ordinals and remember their position so as to store the path in the correct position
2. Try to `advanceExact` to the correct position with the previously calculated `readerIndex`. If the operation fails, try to find the correct segment for the ordinal and then `advanceExact` to the desired position.
3. Store this position for future ordinals.

# Tests

Added a new test for the API that compares the individual `getPath` results from ordinals with the bulk FacetLabels returned by the `getBulkPath` API

# Checklist

### DIFF
--- a/lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/DirectoryTaxonomyReader.java
+++ b/lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/DirectoryTaxonomyReader.java
@@ -18,8 +18,10 @@ package org.apache.lucene.facet.taxonomy.directory;
 
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.logging.Level;
@@ -314,7 +316,6 @@ public class DirectoryTaxonomyReader extends TaxonomyReader implements Accountab
 
   @Override
   public FacetLabel getPath(int ordinal) throws IOException {
-    ensureOpen();
 
     // Since the cache is shared with DTR instances allocated from
     // doOpenIfChanged, we need to ensure that the ordinal is one that this DTR
@@ -324,14 +325,9 @@ public class DirectoryTaxonomyReader extends TaxonomyReader implements Accountab
       return null;
     }
 
-    // TODO: can we use an int-based hash impl, such as IntToObjectMap,
-    // wrapped as LRU?
-    Integer catIDInteger = Integer.valueOf(ordinal);
-    synchronized (categoryCache) {
-      FacetLabel res = categoryCache.get(catIDInteger);
-      if (res != null) {
-        return res;
-      }
+    FacetLabel ordinalPath = getPathFromCache(ordinal);
+    if (ordinalPath != null) {
+      return ordinalPath;
     }
 
     int readerIndex = ReaderUtil.subIndex(ordinal, indexReader.leaves());
@@ -353,10 +349,61 @@ public class DirectoryTaxonomyReader extends TaxonomyReader implements Accountab
     }
 
     synchronized (categoryCache) {
-      categoryCache.put(catIDInteger, ret);
+      categoryCache.put(ordinal, ret);
     }
 
     return ret;
+  }
+
+  private FacetLabel getPathFromCache(int ordinal) {
+    ensureOpen();
+
+    // TODO: can we use an int-based hash impl, such as IntToObjectMap,
+    // wrapped as LRU?
+    synchronized (categoryCache) {
+      FacetLabel res = categoryCache.get(ordinal);
+      if (res != null) {
+        return res;
+      }
+    }
+    return null;
+  }
+
+  /* This API is only supported for indexes created with Lucene 8.7+ codec **/
+  public FacetLabel[] getBulkPath(int[] ordinal) throws IOException {
+    FacetLabel[] bulkPath = new FacetLabel[ordinal.length];
+    Map<Integer,Integer> originalPosition = new HashMap<>();
+    for (int i = 0 ; i < ordinal.length ; i++) {
+      if (ordinal[i] < 0 || ordinal[i] >= indexReader.maxDoc()) {
+        return null;
+      }
+      FacetLabel ordinalPath = getPathFromCache(ordinal[i]);
+      if (ordinalPath != null) {
+        bulkPath[i] = ordinalPath;
+      }
+      originalPosition.put(ordinal[i], i);
+    }
+
+    Arrays.sort(ordinal);
+    int readerIndex = 0;
+    BinaryDocValues values = null;
+
+    for (int ord : ordinal) {
+      if (bulkPath[originalPosition.get(ord)] == null) {
+        if (values == null || values.advanceExact(ord - indexReader.leaves().get(readerIndex).docBase) == false) {
+          readerIndex = ReaderUtil.subIndex(ord, indexReader.leaves());
+          LeafReader leafReader = indexReader.leaves().get(readerIndex).reader();
+          values = leafReader.getBinaryDocValues(Consts.FULL);
+          assert values.advanceExact(ord - indexReader.leaves().get(readerIndex).docBase);
+        }
+        bulkPath[originalPosition.get(ord)] = new FacetLabel(FacetsConfig.stringToPath(values.binaryValue().utf8ToString()));
+        synchronized (categoryCache) {
+          categoryCache.put(ord, bulkPath[originalPosition.get(ord)]);
+        }
+      }
+    }
+
+    return bulkPath;
   }
 
   @Override
--- a/lucene/facet/src/test/org/apache/lucene/facet/taxonomy/directory/TestDirectoryTaxonomyReader.java
+++ b/lucene/facet/src/test/org/apache/lucene/facet/taxonomy/directory/TestDirectoryTaxonomyReader.java
@@ -16,11 +16,6 @@
  */
 package org.apache.lucene.facet.taxonomy.directory;
 
-import java.io.IOException;
-import java.util.Arrays;
-import java.util.HashSet;
-import java.util.Random;
-import java.util.Set;
 import org.apache.lucene.analysis.MockAnalyzer;
 import org.apache.lucene.facet.FacetTestCase;
 import org.apache.lucene.facet.taxonomy.FacetLabel;
@@ -37,6 +32,12 @@ import org.apache.lucene.store.ByteBuffersDirectory;
 import org.apache.lucene.store.Directory;
 import org.apache.lucene.util.IOUtils;
 import org.junit.Test;
+
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Random;
+import java.util.Set;
 
 public class TestDirectoryTaxonomyReader extends FacetTestCase {
 
@@ -568,5 +569,32 @@ public class TestDirectoryTaxonomyReader extends FacetTestCase {
     assertTrue(taxoReader.getChildResources().size() > 0);
     taxoReader.close();
     dir.close();
+  }
+
+  public void testBulkPath() throws Exception {
+    Directory src = newDirectory();
+    DirectoryTaxonomyWriter w = new DirectoryTaxonomyWriter(src);
+    String arr[] = new String[]{"a", "b", "c", "d", "e"};
+
+    FacetLabel allpaths[] = new FacetLabel[arr.length];
+    int allords[] = new int[arr.length];
+
+    for (int i=0;i<arr.length;i++) {
+      allpaths[i] = new FacetLabel(arr[i]);
+      w.addCategory(allpaths[i]);
+    }
+    w.commit();
+    w.close();
+
+    DirectoryTaxonomyReader r1 = new DirectoryTaxonomyReader(src);
+
+    for (int i=0;i<allpaths.length;i++) {
+      allords[i] = r1.getOrdinal(allpaths[i]);
+    }
+
+    FacetLabel allBulkpath[] = r1.getBulkPath(allords);
+    assertArrayEquals(allpaths, allBulkpath);
+    r1.close();
+    src.close();
   }
 }


lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/DirectoryTaxonomyReader.java
```diff
@@ -314,7 +316,6 @@ public int getOrdinal(FacetLabel cp) throws IOException {
 
   @Override
   public FacetLabel getPath(int ordinal) throws IOException {
-    ensureOpen();
```
mikemccand, line 317: Hmm why is this removed?

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/DirectoryTaxonomyReader.java
```diff
@@ -353,12 +349,65 @@ public FacetLabel getPath(int ordinal) throws IOException {
...
+    synchronized (categoryCache) {
+      FacetLabel res = categoryCache.get(ordinal);
+      if (res != null) {
+        return res;
+      }
+    }
+    return null;
+  }
+
+  /* This API is only supported for indexes created with Lucene 8.7+ codec **/
```
mikemccand, line 372: Whoa, why is that!?  Oh, this is because it must be `BinaryDocValues` taxonomy index, OK.  If the user messes up and tries to use this API on stored fields, what happens?

Also, can we enhance the javadoc to state what the method does, link to the non-bulk API, etc.?

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/DirectoryTaxonomyReader.java
```diff
@@ -353,12 +349,65 @@ public FacetLabel getPath(int ordinal) throws IOException {
...
+  public FacetLabel[] getBulkPath(int[] ordinal) throws IOException {
+    FacetLabel[] bulkPath = new FacetLabel[ordinal.length];
+    Map<Integer, Integer> originalPosition = new HashMap<>();
+    for (int i = 0; i < ordinal.length; i++) {
+      if (ordinal[i] < 0 || ordinal[i] >= indexReader.maxDoc()) {
+        return null;
```
mikemccand, line 378: Hmm can we instead throw exception w/ meaningful message?

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/DirectoryTaxonomyReader.java
```diff
@@ -353,12 +349,65 @@ public FacetLabel getPath(int ordinal) throws IOException {
...
+        return res;
+      }
+    }
+    return null;
+  }
+
+  /* This API is only supported for indexes created with Lucene 8.7+ codec **/
+  public FacetLabel[] getBulkPath(int[] ordinal) throws IOException {
+    FacetLabel[] bulkPath = new FacetLabel[ordinal.length];
+    Map<Integer, Integer> originalPosition = new HashMap<>();
```
mikemccand, line 375: Maybe use `IntIntScatterMap` from HPPC instead, to create less garbage ... `facet` module already depends on HPPC.

But this optimization can come later, and is likely negligible for most search use cases anyways.

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/DirectoryTaxonomyReader.java
```diff
@@ -353,12 +349,65 @@ public FacetLabel getPath(int ordinal) throws IOException {
...
+      FacetLabel res = categoryCache.get(ordinal);
+      if (res != null) {
+        return res;
+      }
+    }
+    return null;
+  }
+
+  /* This API is only supported for indexes created with Lucene 8.7+ codec **/
+  public FacetLabel[] getBulkPath(int[] ordinal) throws IOException {
```
mikemccand, line 373: Maybe rename the parameter to `ordinals`?

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/DirectoryTaxonomyReader.java
```diff
@@ -353,12 +349,65 @@ public FacetLabel getPath(int ordinal) throws IOException {
...
+    Map<Integer, Integer> originalPosition = new HashMap<>();
+    for (int i = 0; i < ordinal.length; i++) {
+      if (ordinal[i] < 0 || ordinal[i] >= indexReader.maxDoc()) {
+        return null;
+      }
+      FacetLabel ordinalPath = getPathFromCache(ordinal[i]);
+      if (ordinalPath != null) {
+        bulkPath[i] = ordinalPath;
+      }
+      originalPosition.put(ordinal[i], i);
```
mikemccand, line 384: Instead of a `Map<Integer,Integer>` you could make two parallel arrays, and sort both arrays in a "matched" manner.  Lucene's sort APIs in `oal.util` make this simple, and there are example in Lucene's sources of "sorting parallel arrays".

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/DirectoryTaxonomyReader.java
```diff
@@ -353,12 +349,65 @@ public FacetLabel getPath(int ordinal) throws IOException {
...
+    }
+
+    Arrays.sort(ordinal);
+    int readerIndex = 0;
+    BinaryDocValues values = null;
+
+    for (int ord : ordinal) {
+      if (bulkPath[originalPosition.get(ord)] == null) {
+        if (values == null
+            || values.advanceExact(ord - indexReader.leaves().get(readerIndex).docBase) == false) {
```
mikemccand, line 394: Hmm, instead of relying on `advanceExact` returning `false` to know when to advance to the next reader, could we instead use `.maxDoc` of the reader we are on?

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/DirectoryTaxonomyReader.java
```diff
@@ -353,12 +349,65 @@ public FacetLabel getPath(int ordinal) throws IOException {
...
+          readerIndex = ReaderUtil.subIndex(ord, indexReader.leaves());
+          LeafReader leafReader = indexReader.leaves().get(readerIndex).reader();
+          values = leafReader.getBinaryDocValues(Consts.FULL);
+          assert values.advanceExact(ord - indexReader.leaves().get(readerIndex).docBase);
```
mikemccand, line 398: Hmm, this is dangerous: we are requiring that this `.advanceExact` be called, since we below use `.binaryValue()`?  But if user disables assertions, this won't happen, and then `.binaryValue()` might be wrong or throw an exception?

I think you should do it in two steps, e.g.:

```
  boolean success = values.advanceExact(ord - indexReader.leaves().get(readerIndex).docBase);
  assert success;
```
?

Also, why didn't the unit test uncover this failure?  Lucene's `test-framework` intentionally, randomly some percentage of the time, disables assertions precisely because of this situation :)  Long ago a Lucene release had a bug that accidentally "relied" on assertions being enabled in order to run correctly.

Edit: OK, I see -- the unit test only creates a single segment index, so this line is not exercised I think (yet).

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/DirectoryTaxonomyReader.java
```diff
@@ -353,12 +349,65 @@ public FacetLabel getPath(int ordinal) throws IOException {
...
+      }
+      originalPosition.put(ordinal[i], i);
+    }
+
+    Arrays.sort(ordinal);
+    int readerIndex = 0;
+    BinaryDocValues values = null;
+
+    for (int ord : ordinal) {
+      if (bulkPath[originalPosition.get(ord)] == null) {
```
mikemccand, line 392: Hmm, why wouldn't this entry be `null`?  Shouldn't it always be `null`?  I.e. each requested ordinal goes to a unique position?

lucene/facet/src/test/org/apache/lucene/facet/taxonomy/directory/TestDirectoryTaxonomyReader.java
```diff
@@ -569,4 +569,31 @@ public void testAccountable() throws Exception {
...
+    w.close();
+
+    DirectoryTaxonomyReader r1 = new DirectoryTaxonomyReader(src);
+
+    for (int i = 0; i < allpaths.length; i++) {
+      allords[i] = r1.getOrdinal(allpaths[i]);
+    }
+
+    FacetLabel allBulkpath[] = r1.getBulkPath(allords);
+    assertArrayEquals(allpaths, allBulkpath);
```
mikemccand, line 595: This is a good initial basic test case, thanks!

Could you also add a randomized test, indexing random number of facet labels, hopefully exercising multiple segments randomly, and then doing bulk lookup of random collections of ordinals in random order?  It could help uncover exciting bugs :)

mikemccand (commented):
Thanks @gautamworah96 

Once we iterate to a solid PR I am very curious how this helps facets performance -- we can switch `luceneutil` over to this bulk API to test.

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/DirectoryTaxonomyReader.java
```diff
@@ -353,12 +349,65 @@ public FacetLabel getPath(int ordinal) throws IOException {
...
     synchronized (categoryCache) {
-      categoryCache.put(catIDInteger, ret);
+      categoryCache.put(ordinal, ret);
     }
 
     return ret;
   }
 
+  private FacetLabel getPathFromCache(int ordinal) {
+    ensureOpen();
```
shaie, line 359: I think Mike asked about it above -- it's better that we call this method in the outer *public* API so it's visible and clear we perform this check rather than rely on calling it in a private method.

Also, potentially a `private` method may be called many times in the course of one `public` API invocation, and we don't need to perform this check over and over.

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/DirectoryTaxonomyReader.java
```diff
@@ -353,12 +349,65 @@ public FacetLabel getPath(int ordinal) throws IOException {
...
+
+    // TODO: can we use an int-based hash impl, such as IntToObjectMap,
+    // wrapped as LRU?
+    synchronized (categoryCache) {
+      FacetLabel res = categoryCache.get(ordinal);
+      if (res != null) {
+        return res;
```
shaie, line 366: It's a style comment, can we do this:

```
FacetLabel res = null;
sync (catCache) {
  res = catCache.get(ordinal);
}
return res;
```

It's just shorter and I believe achieves the same outcome.

In fact, the entire code can be changed to:

```
sync(catCache) {
  return catCache.get(ordinal);
}
```
??

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/DirectoryTaxonomyReader.java
shaie: I'd also change the method to take `int... ordinal` instead? More convenient if you want to only get the path for one ordinal (even though the method is called `getBulkXXX`?

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/DirectoryTaxonomyReader.java
shaie: I don't want to do premature optimization here, but `indexReader.maxDoc()` is called for each ordinal, but it cannot change between ordinals, so we could extract its value outside the loop?

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/DirectoryTaxonomyReader.java
```diff
@@ -353,12 +349,65 @@ public FacetLabel getPath(int ordinal) throws IOException {
...
+
+  /* This API is only supported for indexes created with Lucene 8.7+ codec **/
+  public FacetLabel[] getBulkPath(int[] ordinal) throws IOException {
+    FacetLabel[] bulkPath = new FacetLabel[ordinal.length];
+    Map<Integer, Integer> originalPosition = new HashMap<>();
+    for (int i = 0; i < ordinal.length; i++) {
+      if (ordinal[i] < 0 || ordinal[i] >= indexReader.maxDoc()) {
+        return null;
+      }
+      FacetLabel ordinalPath = getPathFromCache(ordinal[i]);
```
shaie, line 380: There you go - you call the new private method in the context of a single public API invocation, therefore no need to perform the `ensureOpen` check for each ordinal 😄

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/DirectoryTaxonomyReader.java
shaie: Separately, and it's been a while since I looked at this code so take treat this comment cautiously -- it confuses me that the method is called `getBulkPath` where we return `FacetLabel` ... perhaps we should rename it to `getFacetLabels`?

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/DirectoryTaxonomyReader.java
shaie: Again, don't want to prematurely optimize, but we could extract `docBase` outside the loop and update it after we update `readerIndex`?

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/DirectoryTaxonomyReader.java
shaie: @mikemccand it could also be that the test runs just didn't trip yet on the seed which disabled assertions for this code, right?

lucene/facet/src/test/org/apache/lucene/facet/taxonomy/directory/TestDirectoryTaxonomyReader.java
```diff
@@ -569,4 +569,31 @@ public void testAccountable() throws Exception {
     taxoReader.close();
     dir.close();
   }
+
+  public void testBulkPath() throws Exception {
```
shaie, line 573: I know Lucene tests are written with that naming convention, but over the last few years I've started to name tests that describe what's being tested. e.g. `testCallingBulkPathReturnsCorrectResult` vs `testBulkPathFailsIfReaderIsClosed` ...

Would be nice if we gradually started to name our tests with more descriptive names :)

shaie (commented):
I am slowly getting back on the horse here 😄 , so this review focuses mainly on style ..

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/DirectoryTaxonomyReader.java
gautamworah96: I was also confused as to why the method above this one is named `getPath` and not `getLabel` or `getFacetLabel`. I chose `getBulkPath` to maintain consistency.

lucene/facet/src/test/org/apache/lucene/facet/taxonomy/directory/TestDirectoryTaxonomyReader.java
gautamworah96: Done. This is definitely better

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/DirectoryTaxonomyReader.java
gautamworah96: This seems much better. Updated. 
My initial impression was that `ensureOpen()` needs to be called before looking up each ordinal.

lucene/facet/src/test/org/apache/lucene/facet/taxonomy/directory/TestDirectoryTaxonomyReader.java
gautamworah96: Added

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/DirectoryTaxonomyReader.java
gautamworah96: It will not be `null` if we already find it in the cache in the initial step i.e the `FacetLabel` for this `ordinal` was found in the cache itself.

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/DirectoryTaxonomyReader.java
gautamworah96: In the earlier iteration it was shifted to the helper method that looked up values in the cache. I have kept it in the outer method in the most recent commit

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/DirectoryTaxonomyReader.java
gautamworah96: Done in the recent commit

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/DirectoryTaxonomyReader.java
gautamworah96: Hmmm. I think using a map to remember initial positions is still simpler to understand? or I could still change it in the next revision

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/DirectoryTaxonomyReader.java
gautamworah96: Thanks for catching this. Fixed it in the new commit

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/DirectoryTaxonomyReader.java
```diff
@@ -353,12 +349,104 @@ public FacetLabel getPath(int ordinal) throws IOException {
...
+    int indexReaderMaxDoc = indexReader.maxDoc();
+    for (int i = 0; i < ordinals.length; i++) {
+      // check whether the ordinal is valid before accessing the cache
+      isOrdinalInIndexReaderRange(ordinals[i], indexReaderMaxDoc);
+      // check the cache before trying to find it in the index
+      FacetLabel ordinalPath = getPathFromCache(ordinals[i]);
+      if (ordinalPath != null) {
+        bulkPath[i] = ordinalPath;
+      }
+      originalPosition.put(ordinals[i], i);
```
mikemccand, line 408: Hmm, one reason for doing the parallel sort instead of the hash map is: what if the incoming ordinals have duplicates?  It is silly to do, but user might do it out of laziness or something, and as the code now stands, that would not work right, I think?  I.e. one of the returned `FacetLabel`s would be `null`?

Could you add a test case for that (admittedly, strange) usage?

And then I think switch to the parallel sort?  [Here is an example](https://github.com/apache/lucene-solr/blob/master/lucene/core/src/java/org/apache/lucene/util/automaton/Automaton.java#L423-L476) of how to do this using Lucene's `Sorter` utility APIs.

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/DirectoryTaxonomyReader.java
mikemccand: Ahhh, I see!

But then, shouldn't we just not enroll that ordinal/index into the map, up front?

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/DirectoryTaxonomyReader.java
```diff
@@ -353,12 +349,104 @@ public FacetLabel getPath(int ordinal) throws IOException {
...
+          leafReaderContext = indexReader.leaves().get(readerIndex);
+          leafReader = leafReaderContext.reader();
+          leafReaderMaxDoc = leafReader.maxDoc();
+          leafReaderDocBase = leafReaderContext.docBase;
+          values = leafReader.getBinaryDocValues(Consts.FULL);
+
+          // this check is only needed once to confirm that the index uses BinaryDocValues
+          boolean success = values.advanceExact(ord - leafReaderDocBase);
+          if (success == false) {
+            throw new IOException(
```
mikemccand, line 433: Hmm, instead, could we just forward to the slow `getPath` API for this case?  And then update the javadocs to state that if you call this on a pre-8.7 written taxonomy index, it will "emulate" bulk by calling the single lookup API multiple times, i.e. no performance gains, but on newer indices, yes performance gains?

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/DirectoryTaxonomyReader.java
```diff
@@ -353,12 +349,104 @@ public FacetLabel getPath(int ordinal) throws IOException {
...
+                "the taxonomy index is created using the older StoredFields format which uses a Lucene "
+                    + "codec older than 8.7. Use the getPath(int ordinal) API iteratively instead.");
+          }
+        }
+        values.advanceExact(ord - leafReaderDocBase);
```
mikemccand, line 438: Hmm can we `assert` that this indeed returned `true`?

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/DirectoryTaxonomyReader.java
```diff
@@ -353,12 +349,104 @@ public FacetLabel getPath(int ordinal) throws IOException {
...
+        bulkPath[originalPosition.get(ord)] =
+            new FacetLabel(FacetsConfig.stringToPath(values.binaryValue().utf8ToString()));
+        synchronized (categoryCache) {
```
mikemccand, line 441: We could move this `synchrtonized` out to a separate pass in the end, so we don't sync per ordinal but rather sync once, and then add all ordinals we looked up?

lucene/facet/src/test/org/apache/lucene/facet/taxonomy/directory/TestDirectoryTaxonomyReader.java
```diff
@@ -569,4 +569,32 @@ public void testAccountable() throws Exception {
...
+    DirectoryTaxonomyWriter w = new DirectoryTaxonomyWriter(src);
+    String randomArray[] = new String[random().nextInt(1000)];
+    Arrays.setAll(randomArray, i -> Integer.toString(random().nextInt()));
+
+    FacetLabel allPaths[] = new FacetLabel[randomArray.length];
+    int allOrdinals[] = new int[randomArray.length];
+
+    for (int i = 0; i < randomArray.length; i++) {
+      allPaths[i] = new FacetLabel(randomArray[i]);
+      w.addCategory(allPaths[i]);
```
mikemccand, line 584: Could we randomly `w.commit()` in here with low but non-zero chance?  This way we will sometimes exercise a multi-segment taxo index.

And, could you temporarily put the `assert` back from the first revision and confirm this test case eventually would catch that bug?

mikemccand (commented):
Thanks @gautamworah96 -- looking closer!

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/DirectoryTaxonomyReader.java
gautamworah96: Ooof. Missed this bug. I have added a small `Sorter` implementation for the parallel arrays (is this reusable code somewhere else as well?)

Modified the new test case so that it is forced to use duplicate ordinals with a probability of 0.5.

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/DirectoryTaxonomyReader.java
```diff
@@ -320,18 +322,12 @@ public FacetLabel getPath(int ordinal) throws IOException {
     // doOpenIfChanged, we need to ensure that the ordinal is one that this DTR
     // instance recognizes. Therefore we do this check up front, before we hit
     // the cache.
-    if (ordinal < 0 || ordinal >= indexReader.maxDoc()) {
-      return null;
-    }
+    int indexReaderMaxDoc = indexReader.maxDoc();
+    isOrdinalInIndexReaderRange(ordinal, indexReaderMaxDoc);
```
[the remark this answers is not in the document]
gautamworah96: Yep. In fact, we had also created a spinoff [issue](https://issues.apache.org/jira/browse/LUCENE-9460) for addressing this expected behavior change in JIRA last year.

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/DirectoryTaxonomyReader.java
gautamworah96: Done.

lucene/facet/src/test/org/apache/lucene/facet/taxonomy/directory/TestDirectoryTaxonomyReader.java
gautamworah96: Done. 

Thanks for the suggestion. I verified that this code trips when `values.advanceExact` is removed, **but** I was not able to verify this by just keeping the `assert values.advanceExact(ordinals[i] - leafReaderDocBase)` line. Perhaps assert was always being executed?

I also manually verified (with print statements) that multi segments were created and were working as expected.

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/DirectoryTaxonomyReader.java
gautamworah96: It is being added to the map in the line that reads 
```
if (ordinalPath != null) {
        bulkPath[i] = ordinalPath;
}
```

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/DirectoryTaxonomyReader.java
```diff
@@ -353,12 +349,137 @@ public FacetLabel getPath(int ordinal) throws IOException {
...
+      // check the cache before trying to find it in the index
+      FacetLabel ordinalPath = getPathFromCache(ordinals[i]);
+      if (ordinalPath != null) {
+        bulkPath[i] = ordinalPath;
+      }
+    }
+
+    // parallel sort the ordinals and originalPosition array based on the values in the ordinals
+    // array
+    new InPlaceMergeSorter() {
```
mikemccand, line 410: Yay, thanks!

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/DirectoryTaxonomyReader.java
```diff
@@ -353,12 +349,137 @@ public FacetLabel getPath(int ordinal) throws IOException {
...
+      protected void swap(int i, int j) {
+        int x = ordinals[i];
+        ordinals[i] = ordinals[j];
+        ordinals[j] = x;
+
+        x = originalPosition[i];
+        originalPosition[i] = originalPosition[j];
+        originalPosition[j] = x;
+      }
+      ;
```
mikemccand, line 421: Hmm, remove this lonely semi-colon?  Did `gradle precommit` pass?

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/DirectoryTaxonomyReader.java
```diff
@@ -353,12 +349,137 @@ public FacetLabel getPath(int ordinal) throws IOException {
...
+          leafReaderContext = indexReader.leaves().get(readerIndex);
+          leafReader = leafReaderContext.reader();
+          leafReaderMaxDoc = leafReader.maxDoc();
+          leafReaderDocBase = leafReaderContext.docBase;
+          values = leafReader.getBinaryDocValues(Consts.FULL);
+
+          // this check is only needed once to confirm that the index uses BinaryDocValues
+          boolean success = values.advanceExact(ordinals[i] - leafReaderDocBase);
+          if (success == false) {
+            return getBulkPathForOlderIndexes(ordinals);
```
mikemccand, line 450: Hmm, I'm confused -- wouldn't an older index have no `BinaryDocValues` field?  So, `values` would be null, and we should fallback then?

This code should hit `NullPointerException` on an old index I think?  How come our backwards compatibility test didn't expose this?

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/DirectoryTaxonomyReader.java
```diff
@@ -353,12 +349,137 @@ public FacetLabel getPath(int ordinal) throws IOException {
...
+        boolean success = values.advanceExact(ordinals[i] - leafReaderDocBase);
+        assert success;
+        bulkPath[originalPosition[i]] =
+            new FacetLabel(FacetsConfig.stringToPath(values.binaryValue().utf8ToString()));
+      }
+    }
+
+    for (int i = 0; i < ordinalsLength; i++) {
+      synchronized (categoryCache) {
+        categoryCache.put(ordinals[i], bulkPath[originalPosition[i]]);
```
mikemccand, line 462: We will sometimes put ordinals back into the cache that were already there at the start of this method right?  I guess that's harmless.  Or, maybe we should move this up above?  Then we can do it only for those ordinals that were not already cached?

lucene/facet/src/test/org/apache/lucene/facet/taxonomy/directory/TestDirectoryTaxonomyReader.java
mikemccand: OK, thanks for confirming.

Hmm, that is spooky if `assert` is always enabled for Lucene tests!  We may have lost this functionality in the gradle migration?  Maybe, separately confirm this with a tiny test case, and if you cannot get that tiny test case to fail, send an email to `dev@` list asking about it?

mikemccand (commented):
Thanks @gautamworah96, looks close!

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/DirectoryTaxonomyReader.java
gautamworah96: `gradle check` (equivalent to precommit) passes. This semicolon is put automatically on a new line because of `./gradlew :lucene:facet:spotlessApply` (used for auto fixing style violations)

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/DirectoryTaxonomyReader.java
gautamworah96: I think intuitively adding the ordinals back into the cache would not be a problem. This should also (theoretically) be faster than trying to get the lock again and again in a loop?

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/DirectoryTaxonomyReader.java
mikemccand: Whoa, OK, I see!  Wild, it is supposed to be there, and on its own line, OK :)

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/DirectoryTaxonomyReader.java
mikemccand: > This should also (theoretically) be faster than trying to get the lock again and again in a loop?

Hmm, I'm confused: this code is already getting the lock inside a `for` loop?  I guess we could move the `synchronized` outside of the `for` loop?  Or, maybe `javac` is doing this for us already?  But let's make it explicit, or, let's just merge this `for` loop with the one before (and keep acquiring the lock inside the `for` loop)?  One big benefit of the latter approach is that if all of the ordinals were already cached (hopefully typically a common case), we do not need any locking, but with this approach, we still do.

lucene/facet/src/test/org/apache/lucene/facet/taxonomy/directory/TestDirectoryTaxonomyReader.java
gautamworah96: I got a small test case working that modified a global variable 1000 times (via an `assert`) and checked the final value. It seems that `assert` is always enabled. 

I'll send an email about the same

gautamworah96:
> Once we iterate to a solid PR I am very curious how this helps facets performance -- we can switch luceneutil over to this bulk API to test.

Today both `IntTaxonomyFacets` and `FloatTaxonomyFacets` iteratively call `getPath` on all the top ordinals and then return the `topChildren` `FacetLabels` that the user wanted. With this API change, we could switch over `IntTaxonomyFacets` and `FloatTaxonomyFacets` to use this bulk API. All downstream children `Facets` such as `FastTaxonomyFacetCount` use this base `getTopChildren` function so all users will be able to benefit from this change by default.

Surprisingly,
Our `luceneutil` benchmark **only** tests the `getTopChildren` API so we should be able to see the performance change with the stock `luceneutil` package.

PR #2249, opened 2021-01-26
LUCENE-9700: Hunspell: support words with trailing dots

donnerpeter:
# Description

Hunspell/C++ has some special treatment for abbreviations

# Solution

Mimic the same behavior: remove all trailing dots and stem the word without any dots or with just one dot.

# Tests

From Hunspell repo: `base` taken as is, `keepcase` merged into the existing same-named test.

# Checklist


lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/SpellChecker.java
```diff
@@ -67,6 +75,14 @@ public boolean spell(String word) {
     return false;
   }
 
+  private boolean removeTrailingDots(String word) {
```
dweiss, line 78: Should it be called spellWithTrailingDots() or something? Right now it reads as if it had side effects... Maybe I'm paranoid though.

lucene/analysis/common/src/test/org/apache/lucene/analysis/hunspell/base.good
```diff
@@ -0,0 +1,27 @@
...
+HELLO
+NASA
+etc.
+etc
+HELLO
+lip.
+text.
+NASA.
+Text.
+TEXT.
```
dweiss, line 25: What happens with a word ending in ellipsis (...)? Should we add a test check for this here too? I also wonder if hunspell handles the ellipsis glyph (unicode character) somehow or it's treated as a character (just curiosity)?

donnerpeter: I'll add a test with `...`. AFAIU Hunspell's code has no special treatment for ellipsis glyph, but it shouldn't be an issue as it should be ignored by tokenizer. And dots in some languages are considered to be word parts.

dweiss: Yeah, I was just wondering how English-centric this is.

> And dots in some languages are considered to be word parts.

Can you give me an example? Interesting!

donnerpeter: French, see https://github.com/wooorm/dictionaries/blob/main/dictionaries/fr/index.aff#L13. I'm not sure why that is, but it is :)

For now, I'm leaving `WORDCHARS` treatment out of Lucene's Hunspell as it only affects tokenizer, and my understanding is that Lucene tokenizes itself somehow. I wonder if this is indeed right.

dweiss: Let's ask folks who speak French - @jpountz would you know of a word with dots inside? Bag.uette? :)

Lucene tokenizers are indeed completely independent. A normalization filter can handle odd glyphs and normalize them into something more palatable.

PR #2252, opened 2021-01-27
SOLR-15111: Use JDK8 Base64 instead of own implementation

asalamon74:
# Description

JDK8 has a builtin Base64 encoder and decoder, there is no need to use own implementaion for this.

# Solution

Eliminate own implementation.

# Tests

Unit tests.

# Checklist


solr/core/src/java/org/apache/solr/handler/admin/LukeRequestHandler.java
```diff
@@ -298,7 +299,7 @@ private static String getFieldFlags( SchemaField f )
 
       BytesRef bytes = field.binaryValue();
       if (bytes != null) {
-        f.add( "binary", Base64.byteArrayToBase64(bytes.bytes, bytes.offset, bytes.length));
+        f.add( "binary", StandardCharsets.ISO_8859_1.decode(Base64.getEncoder().encode(bytes.wrapToByteBuffer())).toString());
```
madrob, line 302: I don't think this is correct, why are we decoding something that we just encoded?

madrob (commented):
I'm a little surprised to see StandardCharsets.ISO_8859_1 in use, most of the bytes we work with are UTF-8. Can you clarify?

asalamon74:
@madrob 

**ISO_8859_1**

In the first version I used UTF_8 but later I checked the source code of java.util.Base64 ( http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/687fd7c7986d/src/share/classes/java/util/Base64.java ) and they are using ISO-8859-1 internally, so I thought to use the same.
```
public byte[] decode(String src) {
    return decode(src.getBytes(StandardCharsets.ISO_8859_1));
}
```
Probably we could use both character encodings, since base64 encoded strings only use a small subset of the characters.

**encode + decode**

```
StandardCharsets.ISO_8859_1.decode(Base64.getEncoder().encode(bytes.wrapToByteBuffer())).toString());
```

First I base64 encode the `ByteBuffer` which gives me a `ByteBuffer` then I convert this `ByteBuffer` to `String` using Charsets decode ( https://stackoverflow.com/a/39845152 ). So it's an encode + a decode but it's a different type of encoding/decoding.

madrob:
Unfortunately, we can’t use the solution provided from stack overflow -http://apache.org/legal/resolved.html#stackoverflow

Can you contact the original author and ask for permission to use this? Otherwise we will need somebody who hasn’t looked this code to create a clean room implementation.

asalamon74:
We also used a different way for String conversion, I modified the lines.

PR #2256, opened 2021-01-27
LUCENE-9507 Custom order for leaves in IndexReader and IndexWriter

mayya-sharipova:
1. Add an option to supply a custom leaf sorter for IndexWriter.
A DirectoryReader opened from this IndexWriter will have its leaf
readers sorted with the provided leaf sorter. This is useful for
indices on which it is expected to run many queries with particular
sort criteria (e.g. for time-based indices this is usually a
descending sort on timestamp). Providing leafSorter allows
to speed up early termination for this particular type of
sort queries.

2. Add an option to supply a custom leaf sorter for
StandardDirectoryReader. The leaf readers of this
StandardDirectoryReader will be sorted according the the provided leaf
sorter.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/DirectoryReader.java
+++ b/lucene/core/src/java/org/apache/lucene/index/DirectoryReader.java
@@ -21,6 +21,7 @@ import java.io.IOException;
 import java.nio.file.NoSuchFileException;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.List;
 import org.apache.lucene.search.SearcherManager; // javadocs
 import org.apache.lucene.store.Directory;
@@ -56,7 +57,24 @@ public abstract class DirectoryReader extends BaseCompositeReader<LeafReader> {
    * @throws IOException if there is a low-level IO error
    */
   public static DirectoryReader open(final Directory directory) throws IOException {
-    return StandardDirectoryReader.open(directory, null);
+    return StandardDirectoryReader.open(directory, null, null);
+  }
+
+  /**
+   * Returns a IndexReader for the the index in the given Directory
+   *
+   * @param directory the index directory
+   * @param leafSorter a comparator for sorting leaf readers. Providing leafSorter is useful for
+   *     indices on which it is expected to run many queries with particular sort criteria (e.g. for
+   *     time-based indices this is usually a descending sort on timestamp). In this case {@code
+   *     leafSorter} should sort leaves according to this sort criteria. Providing leafSorter allows
+   *     to speed up this particular type of sort queries by early terminating while iterating
+   *     though segments and segments' documents.
+   * @throws IOException if there is a low-level IO error
+   */
+  public static DirectoryReader open(final Directory directory, Comparator<LeafReader> leafSorter)
+      throws IOException {
+    return StandardDirectoryReader.open(directory, null, leafSorter);
   }
 
   /**
@@ -101,7 +119,7 @@ public abstract class DirectoryReader extends BaseCompositeReader<LeafReader> {
    * @throws IOException if there is a low-level IO error
    */
   public static DirectoryReader open(final IndexCommit commit) throws IOException {
-    return StandardDirectoryReader.open(commit.getDirectory(), commit);
+    return StandardDirectoryReader.open(commit.getDirectory(), commit, null);
   }
 
   /**
@@ -118,7 +136,8 @@ public abstract class DirectoryReader extends BaseCompositeReader<LeafReader> {
    */
   public static DirectoryReader open(final IndexCommit commit, int minSupportedMajorVersion)
       throws IOException {
-    return StandardDirectoryReader.open(commit.getDirectory(), minSupportedMajorVersion, commit);
+    return StandardDirectoryReader.open(
+        commit.getDirectory(), minSupportedMajorVersion, commit, null);
   }
 
   /**
--- a/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
+++ b/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
@@ -26,6 +26,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.Date;
 import java.util.Deque;
 import java.util.HashMap;
@@ -214,6 +215,8 @@ public class IndexWriter
   static int getActualMaxDocs() {
     return IndexWriter.actualMaxDocs;
   }
+
+  private final Comparator<LeafReader> leafSorter;
 
   /** Used only for testing. */
   private final boolean enableTestPoints;
@@ -933,6 +936,31 @@ public class IndexWriter
    *     low-level IO error
    */
   public IndexWriter(Directory d, IndexWriterConfig conf) throws IOException {
+    this(d, conf, null);
+  }
+
+  /**
+   * Constructs a new IndexWriter per the settings given in <code>conf</code>. If you want to make
+   * "live" changes to this writer instance, use {@link #getConfig()}.
+   *
+   * <p><b>NOTE:</b> after ths writer is created, the given configuration instance cannot be passed
+   * to another writer.
+   *
+   * @param d the index directory. The index is either created or appended according <code>
+   *     conf.getOpenMode()</code>.
+   * @param conf the configuration settings according to which IndexWriter should be initialized.
+   * @param leafSorter a comparator for sorting leaf readers. Providing leafSorter is useful for
+   *     indices on which it is expected to run many queries with particular sort criteria (e.g. for
+   *     time-based indices this is usually a descending sort on timestamp). In this case {@code
+   *     leafSorter} should sort leaves according to this sort criteria. Providing leafSorter allows
+   *     to speed up this particular type of sort queries by early terminating while iterating
+   *     though segments and segments' documents.
+   * @throws IOException if the directory cannot be read/written to, or if it does not exist and
+   *     <code>conf.getOpenMode()</code> is <code>OpenMode.APPEND</code> or if there is any other
+   *     low-level IO error
+   */
+  public IndexWriter(Directory d, IndexWriterConfig conf, Comparator<LeafReader> leafSorter)
+      throws IOException {
     enableTestPoints = isEnableTestPoints();
     conf.setIndexWriter(this); // prevent reuse by other instances
     config = conf;
@@ -941,6 +969,11 @@ public class IndexWriter
     // obtain the write.lock. If the user configured a timeout,
     // we wrap with a sleeper and this might take some time.
     writeLock = d.obtainLock(WRITE_LOCK_NAME);
+    if (config.getIndexSort() != null && leafSorter != null) {
+      throw new IllegalArgumentException(
+          "[IndexWriter] can't use index sort and leaf sorter at the same time!");
+    }
+    this.leafSorter = leafSorter;
 
     boolean success = false;
     try {
@@ -1408,6 +1441,16 @@ public class IndexWriter
   public Analyzer getAnalyzer() {
     ensureOpen();
     return config.getAnalyzer();
+  }
+
+  /**
+   * Returns a comparator for sorting leaf readers. If not {@code null}, this comparator is used to
+   * sort leaf readers within {@code DirectoryReader} opened from this {@code IndexWriter}.
+   *
+   * @return a comparator for sorting leaf readers
+   */
+  public Comparator<LeafReader> getLeafSorter() {
+    return leafSorter;
   }
 
   /**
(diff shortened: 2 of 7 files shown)


mikemccand (commented):
This is an exciting change, especially for non-concurrent early termination cases.

I left a couple questions.

Thanks @mayya-sharipova !

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
```diff
@@ -941,6 +969,11 @@ public IndexWriter(Directory d, IndexWriterConfig conf) throws IOException {
     // obtain the write.lock. If the user configured a timeout,
     // we wrap with a sleeper and this might take some time.
     writeLock = d.obtainLock(WRITE_LOCK_NAME);
+    if (config.getIndexSort() != null && leafSorter != null) {
+      throw new IllegalArgumentException(
+          "[IndexWriter] can't use index sort and leaf sorter at the same time!");
```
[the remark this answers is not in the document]
mayya-sharipova: @mikemccand Thank you for your review! From the discussion on the Jira ticket, we also wanted to use writer's `leafSorter` during merging for arranging docs in  a merged segment (by putting first docs from a segment with highest sort values according to `leafSorter`).

This will be in conflict with `indexSorter`, as if provided [it will arrange merged docs according to its sort](https://github.com/apache/lucene-solr/blob/master/lucene/core/src/java/org/apache/lucene/index/MergeState.java#L211).

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
```diff
@@ -933,6 +936,31 @@ protected final void ensureOpen() throws AlreadyClosedException {
...
+   *     indices on which it is expected to run many queries with particular sort criteria (e.g. for
+   *     time-based indices this is usually a descending sort on timestamp). In this case {@code
+   *     leafSorter} should sort leaves according to this sort criteria. Providing leafSorter allows
+   *     to speed up this particular type of sort queries by early terminating while iterating
+   *     though segments and segments' documents.
+   * @throws IOException if the directory cannot be read/written to, or if it does not exist and
+   *     <code>conf.getOpenMode()</code> is <code>OpenMode.APPEND</code> or if there is any other
+   *     low-level IO error
+   */
+  public IndexWriter(Directory d, IndexWriterConfig conf, Comparator<LeafReader> leafSorter)
```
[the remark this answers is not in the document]
mayya-sharipova: Thank you for your suggestion, I will explore this.

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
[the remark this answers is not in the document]
msokolov: I think it's OK - that MultiSorter is sorting (by index sort) documents in several segments being merged; it will control the order of the documents within the new segment, but shouldn't have any influence on or conflict with the order of the segments being merged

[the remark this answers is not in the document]
mayya-sharipova: @msokolov Thank you for your feedback and explanation. Sorry, I am still not super clear about this point. It seems to me as  `indexSorter` maps each leaf's documents into the merged segment according to its sort,  the same way `leafSorter` will map each leaf's documents into the merged segment according to its sort (given several merging segments, in the merged segment the first docs should be docs from a segment with highest sort values..)

But I am ok to remove this check in this PR, as this PR is not concerned with merging, and follow up on this point in the following PR. 

Addressed in 7ddff6775c8

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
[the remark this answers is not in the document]
mayya-sharipova: Addressed in 7ddff67

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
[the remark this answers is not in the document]
msokolov: OK, maybe I misunderstood the intent. Perhaps an example would clarify. Say that we have three segments, A, B , C containing documents `A={0, 3, 6}; B={1, 4, 7}; C={2, 5, 8}`, where the documents are understood to have a single field with the value shown, and the index sort is ordered in the natural way.  Without this change, if we merged A and B, we'd get a new segment `A+B={0, 1, 3, 4, 6, 7}`. Now suppose there is no index sort (and the documents just "happen" to be in the index in the order given above, for the sake of the example), and we apply a `leafSorter` that sorts by the minimum value of any document in the segment (I guess it could be any sort of aggregate over the segment?), then we would get `A+B={0, 3, 6, 1, 4, 7}`. Now if we apply both sorts, we would get the same result as in the first case, right? I'm still unclear how the conflict arises.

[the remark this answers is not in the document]
msokolov: Hmm I do see where you said we want to use `leafSorter` to sort documents. I guess I might challenge that design in favor of building on the `indexSort` we already have? But perhaps this is better sorted out in the context of a later PR, as you suggested.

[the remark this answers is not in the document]
mayya-sharipova: @msokolov Thank you for  the clarification.  Indeed, it is much clear with an example you provided. Looks like we need to think  and discuss more about merging scenario, may be in the next PR or Jira ticket.

lucene/core/src/java/org/apache/lucene/index/StandardDirectoryReader.java
```diff
@@ -143,12 +158,17 @@ static StandardDirectoryReader open(
 
       writer.incRefDeleter(segmentInfos);
 
+      if (writer.getConfig().getLeafSorter() != null) {
+        readers.sort(writer.getConfig().getLeafSorter());
```
mikemccand, line 162: Is IWC.leafSorter allowed to be live-updated?  If so, maybe we should pull out local variable here to hold onto the `leafSorter` and use that variable here and below, to ensure it is the same?

lucene/core/src/java/org/apache/lucene/index/StandardDirectoryReader.java
```diff
@@ -169,7 +189,10 @@ static StandardDirectoryReader open(
    * @lucene.internal
    */
   public static DirectoryReader open(
-      Directory directory, SegmentInfos infos, List<? extends LeafReader> oldReaders)
+      Directory directory,
```
mikemccand, line 192: Are we handling the near-real-time reader (`IndexWriter.getReader(...)`) path here too?

lucene/core/src/java/org/apache/lucene/index/StandardDirectoryReader.java
```diff
@@ -39,33 +40,44 @@
...
   final IndexWriter writer;
   final SegmentInfos segmentInfos;
+  private final Comparator<LeafReader> leafSorter;
   private final boolean applyAllDeletes;
   private final boolean writeAllDeletes;
 
-  /** called only from static open() methods */
+  /**
+   * package private constructor, called only from static open() methods. If parameter {@code
+   * leafSorter} is not {@code null}, the provided {@code readers} are expected to be already sorted
```
mikemccand, line 49: Could we add an `assertLeavesSorted` method to confirm they really were properly sorted by the caller?  There are so many paths for creating `StandardDirectoryReader` through are code base that it would not surprise me if one of the misses to sort properly.

mikemccand (commented):
Thanks @mayya-sharipova -- I left a few comments.  This is a neat new capability ...

lucene/core/src/java/org/apache/lucene/index/StandardDirectoryReader.java
mayya-sharipova: Great suggestion! I moved sorting of leaves to the constructor. 
Addressed in 8155f2e2abd4ed3a6280aae940d018dfe3b3dad1.

lucene/core/src/java/org/apache/lucene/index/StandardDirectoryReader.java
mayya-sharipova: @mikemccand  Thank you for your feedback.
The intent was to keep `leafSorter` constant and non-updatable, as  `indexWriter.getConfig` returns `LiveIndexWriterConfig` that doesn't allow setting of a new `leafSorter`. I understand that `leafSorter` in config can still be changed, and your comment is very valid. 

But In 8155f2e2abd4ed3a6280aae940d018dfe3b3dad1  I've moved sorting of leaves to the constructor, so hopefully this should resolve this concern.

lucene/core/src/java/org/apache/lucene/index/StandardDirectoryReader.java
mayya-sharipova: It looks like  `IndexWriter.getReader(...)` path is covered by a previous `open` method on line 123.

lucene/core/src/java/org/apache/lucene/index/DirectoryReader.java
```diff
@@ -56,7 +57,24 @@
...
+  /**
+   * Returns a IndexReader for the the index in the given Directory
+   *
+   * @param directory the index directory
+   * @param leafSorter a comparator for sorting leaf readers. Providing leafSorter is useful for
+   *     indices on which it is expected to run many queries with particular sort criteria (e.g. for
+   *     time-based indices this is usually a descending sort on timestamp). In this case {@code
+   *     leafSorter} should sort leaves according to this sort criteria. Providing leafSorter allows
+   *     to speed up this particular type of sort queries by early terminating while iterating
+   *     though segments and segments' documents.
```
jimczi, line 72: nit: s/though/through/

lucene/core/src/java/org/apache/lucene/index/StandardDirectoryReader.java
```diff
@@ -39,33 +40,47 @@
...
       Directory directory,
       LeafReader[] readers,
       IndexWriter writer,
       SegmentInfos sis,
+      Comparator<LeafReader> leafSorter,
       boolean applyAllDeletes,
       boolean writeAllDeletes)
       throws IOException {
-    super(directory, readers);
+    super(directory, sortLeaves(readers, leafSorter));
```
jimczi, line 57: I wonder if the leafSorter should be declared and executed in the base class `BaseCompositeReader` ?
That would expose the feature explicitly to `MultiReader` and friends.

lucene/core/src/test/org/apache/lucene/index/TestIndexWriterReader.java
```diff
@@ -169,7 +176,7 @@ public void testUpdateDocument() throws Exception {
     // writer.close wrote a new commit
     assertFalse(r2.isCurrent());
 
-    DirectoryReader r3 = DirectoryReader.open(dir1);
+    DirectoryReader r3 = open(dir1);
```
jimczi, line 179: Can you keep the explicit version ?

lucene/core/src/java/org/apache/lucene/index/IndexWriterConfig.java
```diff
@@ -478,6 +479,18 @@ public IndexWriterConfig setIndexSort(Sort sort) {
...
   }
 
+  /**
+   * Set the comparator for sorting leaf readers. A DirectoryReader opened from a IndexWriter with
+   * this configuration will have its leaf readers sorted with the provided leaf sorter.
+   *
+   * @param leafSorter – a comparator for sorting leaf readers
+   * @return IndexWriterConfig with leafSorter set.
+   */
+  public IndexWriterConfig setLeafSorter(Comparator<LeafReader> leafSorter) {
```
jimczi, line 489: You added a specific unit test for this feature but we could also set a random value in `LuceneTestCase#newIndexWriterConfig` to improve the coverage.

jimczi (commented):
The change looks good. I left some additional comments and questions.

mayya-sharipova:
I've moved this PR to a [new PR](https://github.com/apache/lucene/pull/32) in  the new Lucene github repo.  Closing this PR.

lucene/core/src/test/org/apache/lucene/index/TestIndexWriterReader.java
mayya-sharipova: Addressed in https://github.com/apache/lucene/pull/32/commits/f3fbba10d5ff391bc66bbd04032a873c91680493

lucene/core/src/java/org/apache/lucene/index/DirectoryReader.java
mayya-sharipova: Addressed in https://github.com/apache/lucene/pull/32/commits/f3fbba10d5ff391bc66bbd04032a873c91680493

lucene/core/src/java/org/apache/lucene/index/IndexWriterConfig.java
mayya-sharipova: How will leafSorter sort leaves in this case? By what criteria?  If a sorting criteria is anything different than the existing order of leaves,  this will break some of our tests that rely on specific docIds. For example tests in the class `TestBasics` that check for specific docIDs.

PR #2258, opened 2021-01-28
LUCENE-9686: Fix read past EOF handling in DirectIODirectory

zacharymorn:
# Description

Explicitly handle read past EOF case in DirectIODirectory#readByte to throw EOFException

# Solution

In DirectIODirectory.readByte, checks for `buffer.limit() == 0” after DirectIODirectory#refill was called.

# Tests

* Add a new test to for read past EOF scenario
* Passed reported randomized test with `./gradlew test —tests TestDirectIODirectory.testFloatsUnderflow -Dtests.seed=73B56EAB13269C91 -Dtests.slow=true -Dtests.badapples=true -Dtests.locale=haw-US -Dtests.timezone=America/Inuvik -Dtests.asserts=true -Dtests.file.encoding=UTF-8`
* Passed `./gradlew check`

# Checklist

### DIFF
--- a/lucene/misc/src/java/org/apache/lucene/misc/store/DirectIODirectory.java
+++ b/lucene/misc/src/java/org/apache/lucene/misc/store/DirectIODirectory.java
@@ -359,12 +359,8 @@ public class DirectIODirectory extends FilterDirectory {
         filePos = alignedPos - buffer.capacity();
 
         final int delta = (int) (pos - alignedPos);
-        refill();
-        try {
-          buffer.position(delta);
-        } catch (IllegalArgumentException iae) {
-          throw new EOFException("read past EOF: " + this);
-        }
+        refill(delta);
+        buffer.position(delta);
       }
       assert pos == getFilePointer();
     }
@@ -381,23 +377,18 @@ public class DirectIODirectory extends FilterDirectory {
     @Override
     public byte readByte() throws IOException {
       if (!buffer.hasRemaining()) {
-        refill();
-
-        // refill may not actually fill buffer if it reaches EOF
-        if (buffer.limit() == 0) {
-          throw new EOFException("read past EOF: " + this);
-        }
+        refill(1);
       }
 
       return buffer.get();
     }
 
-    private void refill() throws IOException {
+    private void refill(int byteToRead) throws IOException {
       filePos += buffer.capacity();
 
       // BaseDirectoryTestCase#testSeekPastEOF test for consecutive read past EOF,
       // hence throwing EOFException early to maintain buffer state (position in particular)
-      if (filePos > channel.size()) {
+      if (filePos > channel.size() || (channel.size() - filePos < byteToRead)) {
         throw new EOFException("read past EOF: " + this);
       }
 
@@ -423,9 +414,7 @@ public class DirectIODirectory extends FilterDirectory {
           buffer.get(dst, offset, left);
           toRead -= left;
           offset += left;
-          // when reading past EOF, this method will keep increasing filePos in the loop to
-          // eventually throw EOFException
-          refill();
+          refill(toRead);
         } else {
           buffer.get(dst, offset, toRead);
           break;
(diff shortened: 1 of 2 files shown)


lucene/misc/src/java/org/apache/lucene/misc/store/DirectIODirectory.java
```diff
@@ -381,17 +377,18 @@ public long length() {
...
       if (!buffer.hasRemaining()) {
-        refill();
+        refill(1);
       }
+
       return buffer.get();
     }
 
-    private void refill() throws IOException {
+    private void refill(int byteToRead) throws IOException {
```
dweiss, line 386: Should it be plural (bytesToRead)?

lucene/misc/src/java/org/apache/lucene/misc/store/DirectIODirectory.java
```diff
@@ -381,17 +377,18 @@ public long length() {
...
       filePos += buffer.capacity();
 
       // BaseDirectoryTestCase#testSeekPastEOF test for consecutive read past EOF,
       // hence throwing EOFException early to maintain buffer state (position in particular)
-      if (filePos > channel.size()) {
+      if (filePos > channel.size() || (channel.size() - filePos < byteToRead)) {
```
dweiss, line 391: I wonder if we should move the channel's position to actually point after the last byte, then throw EOFException... so that not only we indicate an EOF but also leave the channel pointing at the end. I have a scenario in my mind when somebody tries to read a bulk of bytes, hits an eof but then a single-byte read() succeeds. That would be awkward, wouldn't it? 

A refill should try to read as many bytes as it can (min(channel.size() -filePos, bytesToRead)), then potentially fail if bytesToRead is still >0 and channel is at EOF. Or is my thinking flawed somewhere?

dweiss (commented):
Hi Zach. Sorry for belated reply. Please take a look at my comments attached to the context. I have some doubts whether EOF should leave the channel undrained. Maybe I'm paranoid here though.

lucene/misc/src/java/org/apache/lucene/misc/store/DirectIODirectory.java
zacharymorn: Updated.

lucene/misc/src/java/org/apache/lucene/misc/store/DirectIODirectory.java
zacharymorn: If I understand your comment correctly, your concern is about the consistency of directory's internal state after EOF is raised right? I think DirectIODirectory already handles that actually (by manipulating `filePos`, but not `channel.position` per se), and I have added some more tests to confirm that to be the case in the latest commit. 

Please note that for the additional tests, I was originally adding them into `BaseDirectoryTestCase#testSeekPastEOF`, but that would fail some existing tests for other directory implementations, as read immediately after seek past EOF doesn't raise EOFException for them:

* TestHardLinkCopyDirectoryWrapper
* TestMmapDirectory
* TestByteBuffersDirectory
* TestMultiMMap

However, according to java doc here https://github.com/apache/lucene-solr/blob/15aaec60d9bfa96f2837c38b7ca83e2c87c66d8d/lucene/core/src/java/org/apache/lucene/store/IndexInput.java#L66-L73, this seems to be an unspecified state in general.

zacharymorn:
> Hi Zach. Sorry for belated reply. Please take a look at my comments attached to the context. I have some doubts whether EOF should leave the channel undrained. Maybe I'm paranoid here though.

Hi Dawid, no worry and thanks for the review! I've replied to the comment and added some tests to verify, please let me know if they look good to you.

lucene/misc/src/java/org/apache/lucene/misc/store/DirectIODirectory.java
dweiss: Ok. Thanks for explaining!

PR #2266, opened 2021-01-28
SOLR-15120 Reduce duplicated core creation work

madrob:
Use j.u.c collections instead of sync block
Rework how we load implicit handlers
Additional debug and trace logging for zookeeper comms

### DIFF
--- a/solr/core/src/java/org/apache/solr/core/SolrCore.java
+++ b/solr/core/src/java/org/apache/solr/core/SolrCore.java
@@ -176,7 +176,6 @@ import org.apache.zookeeper.data.Stat;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import static org.apache.solr.common.params.CommonParams.NAME;
 import static org.apache.solr.common.params.CommonParams.PATH;
 
 /**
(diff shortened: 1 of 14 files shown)


solr/core/src/java/org/apache/solr/core/SolrCore.java
```diff
@@ -3187,20 +3186,29 @@ public void cleanupOldIndexDirectories(boolean reload) {
     }
   }
 
-  @SuppressWarnings({"rawtypes"})
-  private static final Map implicitPluginsInfo = (Map) Utils.fromJSONResource("ImplicitPlugins.json");
+  private static final class ImplicitHolder {
```
dsmiley, line 3189: It appears this is here to lazy-load?  But why lazy-load (added complexity of ImplicitHolder wrapper) -- this cost is eventually inescapable since these are needed to load any core, not some obscure/optional feature.
@madrob

madrob: It was always eager reading from disk and lazy copying the plugins and a messy combination of the two. Now it’s lazy everything the first time, but yea probably could be eager. I don’t think we’re putting in too much complexity this way tho

madrob: I have no problem with this being eager loaded

PR #2268, opened 2021-01-29
LUCENE-9705: Move Lucene50CompoundFormat to Lucene90CompoundFormat

iverase:
see https://issues.apache.org/jira/browse/LUCENE-9705


lucene/core/src/java/org/apache/lucene/codecs/lucene90/Lucene90CompoundFormat.java
```diff
@@ -0,0 +1,140 @@
...
+ *       files length, and a String with that file's name.
+ * </ul>
+ */
+public final class Lucene90CompoundFormat extends CompoundFormat {
+
+  /** Extension of compound file */
+  static final String DATA_EXTENSION = "cfs";
+  /** Extension of compound file entries */
+  static final String ENTRIES_EXTENSION = "cfe";
+
```
iverase, line 72: Shall we keep the extensions or use new ones?

lucene/backward-codecs/src/test/org/apache/lucene/backward_codecs/lucene50/TestLucene50CompoundFormat.java
```diff
@@ -0,0 +1,30 @@
...
+import org.apache.lucene.backward_codecs.Lucene87.Lucene87RWCodec;
+import org.apache.lucene.codecs.Codec;
+import org.apache.lucene.index.BaseCompoundFormatTestCase;
+
+public class TestLucene50CompoundFormat extends BaseCompoundFormatTestCase {
+  ;
+
+  @Override
+  protected Codec getCodec() {
+    return new Lucene87RWCodec();
```
iverase, line 28: We could use other codecs as well?

lucene/backward-codecs/src/java/org/apache/lucene/backward_codecs/lucene50/Lucene50CompoundFormat.java
```diff
@@ -0,0 +1,88 @@
...
+public final class Lucene50CompoundFormat extends CompoundFormat {
+
+  /** Extension of compound file */
+  static final String DATA_EXTENSION = "cfs";
+  /** Extension of compound file entries */
+  static final String ENTRIES_EXTENSION = "cfe";
+
+  static final String DATA_CODEC = "Lucene50CompoundData";
+  static final String ENTRY_CODEC = "Lucene50CompoundEntries";
+  static final int VERSION_START = 0;
```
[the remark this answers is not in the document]
jpountz: I'd like to keep it for now, even if the version is always 0. My gut feeling is that we should fork file formats more aggressively than we do today but I still don't have full confidence that we will never use the internal versioning again.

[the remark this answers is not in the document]
jpountz: FWIW we have a tests that checks this structure: `TestAllFilesHaveCodecHeader`.

lucene/backward-codecs/src/test/org/apache/lucene/backward_codecs/lucene50/TestLucene50CompoundFormat.java
jpountz: For simplicity I'd rather only test the most recent codec that uses the Lucene50 compound format like here.

jpountz (approved):

lucene/backward-codecs/src/java/org/apache/lucene/backward_codecs/lucene50/Lucene50CompoundFormat.java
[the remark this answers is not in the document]
jtibshirani: Thanks, this helps clarify our current thinking. It's also nice that all file formats have the same header structure.

lucene/core/src/java/org/apache/lucene/codecs/lucene90/Lucene90CompoundFormat.java
jtibshirani: What would be the motivation? Do you have an idea for improved names?

jtibshirani (approved):
Looks good to me. I caused merge conflicts by sneaking in https://github.com/apache/lucene-solr/pull/2274 first, sorry about that!

lucene/core/src/java/org/apache/lucene/codecs/lucene90/Lucene90CompoundFormat.java
iverase: I guess the only benefit is that you can distinguish the files from the new codec vs previous codec by name.

jpountz: We would like to avoid sharing file extensions across file formats, e.g. given that `.liv` is used by live docs today, it could be confusing to use it for stored fields in the future.

However I don't feel the need to change extensions as we iterate on a given file format, even if we changed completely the way it works. This is something we did in Lucene 5.x for postings, moving from `.frq`/`.prox` to `.doc`/`.pos` but this is not a requirement at all, by the way we changed stored fields in Lucene 4.x and kept the same `.fdt` extension even though the new file format had very little in common with the previous one.

I think we can keep the same file extensions here. It's easy enough to read the first bytes of a `cfs`/`cfe` file in case of doubt to check out the name in the codec header.

iverase: Sounds good to me

PR #2269, opened 2021-01-29
LUCENE-9322:  Add TestLucene90FieldInfosFormat

iverase:
When introducing the new Vector format, we introduced as well a new FieldInfos format. This PR add the test for this format, similar to what it was done for the previous one.

@msokolov I think we should introduce one of this test for the Vector format (test that extends `BaseIndexFileFormatTestCase`). wdyt?


jtibshirani (commented):
Thanks @iverase. It looks like we're missing checks for vector values in `BaseFieldInfoFormatTestCase` (numeric vectors, not term vectors), we could add some to make sure there's good coverage. I also noticed we don't test points at all.

jtibshirani:
I opened https://issues.apache.org/jira/browse/LUCENE-9756 to add tests for points and vectors.

msokolov:
Sorry, just catching up on my emails now -- I agree @iverase , we should expand test coverage for these formats to the level we have for the others. Thanks for opening the issue @jtibshirani

PR #2273, opened 2021-01-29
SOLR-8319: Add method for term field search in FieldType

HoustonPutman:
Addresses NPE found in https://issues.apache.org/jira/browse/SOLR-8319

Will do some performance comparisons before merging as well.


HoustonPutman:
Fixed via 20799e737f2ac56e2ff5976e48a8c199a6b42220

PR #2274, opened 2021-01-29
LUCENE-9705: Create Lucene90LiveDocsFormat

jtibshirani:
For now this is just a copy of Lucene50LiveDocsFormat. The existing
Lucene50LiveDocsFormat was moved to backwards-codecs.


lucene/backward-codecs/src/java/org/apache/lucene/backward_codecs/lucene50/Lucene50LiveDocsFormat.java
```diff
@@ -0,0 +1,170 @@
...
+      data[i] = input.readLong();
+    }
+    return new FixedBitSet(data, length);
+  }
+
+  /**
+   * Note: although this format is only used on older versions, we need to keep the write logic in
+   * addition to the read logic. When we delete documents that live in an older segment, we write to
+   * the live docs for that segment.
+   */
```
jpountz, line 114: Thanks for documenting this.

jpountz (approved):

PR #2275, opened 2021-01-29
SOLR-15123: Make all Tool option descriptions follow the same general pattern.

epugh:
# Description
different people have written different SolrCLI tools over time, and it shows in the user help.

# Solution
Copy edit all the tool options.

# Tests

Just docs.   In fact, it appears tools don't have tests.  ;-(


solr/core/src/java/org/apache/solr/util/ExportTool.java
```diff
@@ -216,32 +216,32 @@ void end() throws IOException {
       Option.builder("url")
           .hasArg()
           .required()
-          .desc("Address of the collection, example http://localhost:8983/solr/gettingstarted")
+          .desc("Address of the collection, example http://localhost:8983/solr/gettingstarted.")
```
madrob, line 219: I think this will be confusing to people if the period is part of the URL or not (if they are copy pasting from a tutorial perhaps.

solr/core/src/java/org/apache/solr/util/PackageTool.java
```diff
@@ -261,44 +261,44 @@ protected void runImpl(CommandLine cli) throws Exception {
...
         .hasArg()
         .required(false)
-        .desc("List of collections. Run './solr package help' for more details.")
+        .desc("List of collections.")
         .build(),
 
         Option.builder("cluster")
         .required(false)
-        .desc("Needed to install cluster level plugins in a package. Run './solr package help' for more details.")
+        .desc("Needed to install cluster level plugins in a package.")
```
madrob, line 276: I'm not sure what this description means. There is no verb?

solr/core/src/java/org/apache/solr/util/SolrCLI.java
```diff
@@ -1795,25 +1795,25 @@ public ConfigSetUploadTool(PrintStream stdout) {
               .argName("confname") // Comes out in help message
               .hasArg() // Has one sub-argument
               .required(true) // confname argument must be present
-              .desc("Configset name on Zookeeper")
+              .desc("Configset name on Zookeeper.")
```
madrob, line 1798: s/on/in maybe?

solr/core/src/java/org/apache/solr/util/SolrCLI.java
```diff
@@ -1942,19 +1942,19 @@ public ZkRmTool(PrintStream stdout) {
...
               .required(true)
-              .desc("Path to remove")
+              .desc("Path to remove.")
               .build(),
           Option.builder("recurse")
               .argName("recurse")
               .hasArg()
               .required(false)
-              .desc("Recurse (true|false, default is false)")
+              .desc("Recurse (true|false), default is false.")
```
madrob, line 1951: These parentheses are inconsistent with the other tools

solr/core/src/java/org/apache/solr/util/SolrCLI.java
```diff
@@ -236,13 +236,13 @@ protected abstract void runCloudTool(CloudSolrClient cloudSolrClient, CommandLin
           .argName("HOST")
           .hasArg()
           .required(false)
-          .desc("Address of the Zookeeper ensemble; defaults to: "+ZK_HOST)
+          .desc("Address of the Zookeeper ensemble; defaults to: "+ ZK_HOST + '.')
```
ctargett, line 239: ZK is properly spelled 'ZooKeeper' (capital 'K').

solr/core/src/java/org/apache/solr/util/SolrCLI.java
```diff
@@ -1311,13 +1311,13 @@ protected void runCloudTool(CloudSolrClient cloudSolrClient, CommandLine cli) th
           .argName("HOST")
           .hasArg()
           .required(false)
-          .desc("Address of the Zookeeper ensemble; defaults to: " + ZK_HOST)
+          .desc("Address of the Zookeeper ensemble; defaults to: " + ZK_HOST + '.')
```
ctargett, line 1314: Should be 'ZooKeeper'

solr/core/src/java/org/apache/solr/util/SolrCLI.java
ctargett: Also, 'ZooKeeper'.

solr/core/src/java/org/apache/solr/util/ExportTool.java
epugh: what would you say to wrapping it in single quotes?  Some other places that appears to be the pattern.   

Address of the collection, example 'http://localhost:8983/solr/gettingstarted'.

Or maybe, I'm just being pedantic about my periods and should skip it!

solr/core/src/java/org/apache/solr/util/PackageTool.java
epugh: Take a look at what I put in now.    Also, now I see we sometimes say "package" and other times "plugin", I wonder wha tthe right term is?  I *think* it's "package".

solr/core/src/java/org/apache/solr/util/SolrCLI.java
epugh: "several" is right.

solr/core/src/java/org/apache/solr/util/ExportTool.java
epugh: also, we do this a lot for the ZooKeeper host:  

```
```

epugh: Hey @madrob what do you think of my comments?  I can go either way, or not change it at all!  Love to get these changes in soon.

solr/core/src/java/org/apache/solr/util/PackageTool.java
```diff
@@ -261,44 +261,44 @@ protected void runImpl(CommandLine cli) throws Exception {
...
-        .desc("Address of the Solr Web application, defaults to: " + SolrCLI.DEFAULT_SOLR_URL)
+        .desc("Address of the Solr Web application, defaults to: " + SolrCLI.DEFAULT_SOLR_URL + '.')
         .build(),
 
         Option.builder("collections")
         .argName("COLLECTIONS")
         .hasArg()
         .required(false)
-        .desc("List of collections. Run './solr package help' for more details.")
+        .desc("Specifies that this is to be a collection(s) level action.")
```
HoustonPutman, line 271: I don't know if this is accurate, but maybe something like:

"Specifies that this action should affect packages for the given collections only, excluding cluster level packages."

And for the cluster one:

"Specifies that this action should affect cluster-level packages only."

epugh: Love it!

PR #2277, opened 2021-02-01
LUCENE-9716: Hunspell: support flag usage before its format is even specified

donnerpeter:
# Description

e.g. `nl` dictionaries first have `KEEPCASE Kc` and only then `FLAG long`

# Solution

Like Hunspell/C++, read affix twice: first extract important info like encoding and flag format, next — the rest.

# Tests

Added

# Checklist

### DIFF
--- a/lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/Dictionary.java
+++ b/lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/Dictionary.java
@@ -22,6 +22,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.LineNumberReader;
+import java.io.OutputStream;
 import java.nio.charset.Charset;
 import java.nio.charset.CharsetDecoder;
 import java.nio.charset.CodingErrorAction;
@@ -216,8 +217,10 @@ public class Dictionary {
     boolean success = false;
     try {
       // Copy contents of the affix stream to a temp file.
-      Files.copy(affix, aff, StandardCopyOption.REPLACE_EXISTING);
-      
+      try (OutputStream os = Files.newOutputStream(aff)) {
+        affix.transferTo(os);
+      }
+
       // pass 1: get encoding & flag
       aff1 = new BufferedInputStream(Files.newInputStream(aff));
       readConfig(aff1);
--- a/lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/Dictionary.java
+++ b/lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/Dictionary.java
@@ -24,6 +24,7 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.LineNumberReader;
 import java.io.OutputStream;
+import java.io.PushbackInputStream;
 import java.nio.charset.Charset;
 import java.nio.charset.CharsetDecoder;
 import java.nio.charset.CodingErrorAction;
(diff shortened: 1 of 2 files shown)


lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/Dictionary.java
```diff
@@ -696,45 +690,25 @@ char affixData(int affixIndex, int offset) {
...
-          ||
-          // this test only at the end as ineffective but would allow lines only containing spaces:
-          encoding.toString().trim().length() == 0) {
-        if (ch < 0) {
-          return DEFAULT_CHARSET.name();
-        }
-        continue;
+      line = line.trim();
+
+      while (line.startsWith("\u00EF") || line.startsWith("\u00BB") || line.startsWith("\u00BF")) {
```
dweiss, line 702: Can the bom really be present on any line? Wouldn't a more elegant solution be to use a buffered input stream (or a pushback input stream) and just consume the bom if it's leading the file?

It is a bit awkward that those files are parsed as ascii (well, iso8859-1) and at the same time have utf bom (not to mention that bit where you convert to utf8 from usi8859-1)... Is this encoding situation really so messed up in hunspell?

donnerpeter: Most likely it's just on the first line, handling it this way was just easier. Pushback might indeed be more elegant, I'll try that, thanks!

The situation with encoding is complicated indeed. AFAIU the encodings are either ASCII-based 8-bit, or UTF-8, so for the first time we read the file we can safely check Latin letters. At least that's what Hunspell appears to do as well.

dweiss: Ok, so it's essentially an unknown byte stream with dynamic charset detection. Not fun. If it's restricted to a reasonable subset (like you said) then a preflight of the content could determine the actual encoding (at least until an explicit encoding declaration is found). Then things would be less messy down the road as you'd just have a Reader to read from... 

Pushback is fine too. Either this or a BufferedInputStream and use mark/reset to adjust stream position after you detect the BOM (or not). As much as I like PushbackInputStream, it predates dinosaurs. :)

dweiss:
Hi Peter. I pushed a commit which replaces bom consumption with a small function so that either it's read atomically or nothing is consumed at all. Looking at the code also made me wonder whether a sufficiently large leading buffer could be used to just parse the needed stuff from the input stream (bypassing the need to create a temp file)... can be left for a later improvement though.

donnerpeter:
Thank you, LGTM!

donnerpeter:
As for this:

```
    // TODO: can these flags change throughout the file? If not then we can abort sooner. And
    // then we wouldn't even need to create a temp file for the affix stream - a large enough
    // leading buffer (BufferedInputStream) would be sufficient?
```

Flags aren't supposed to be redefined, although Hunspell currently silently (and I believe erroneously) allows that. Aborting sooner is possible, but would need a bit more code.

I'd very much like to avoid a temp file, but I don't understand how. I don't think we can fix the size of the temp buffer in advance and expect all aff files to obey that, some of them include large comments in the beginning. And unfortunately, the files themselves can be quite large: for Kinyarwanda it's 38MB.

dweiss:
Hi Peter. The file size itself doesn't matter if we can assume some kind of leader buffer in which these flags have to occur and which we can rewind. Implementing this is technically easy - for example via BufferedInputStream with a reasonably large internal buffer, then a mark on the zero-eth byte. Once you reach your flags, you reset the buffer. 

The only problem I see here is a minor potential to have a the buffer limit fall on an UTF8 surrogate, for example, which could potentially trigger some kind of exception... but this can be worked around.

I'll try to do this, time permitting. It's not much of a problem - it can be done later too. I have to limit my time for Lucene to reasonable chunks though. :)

> for Kinyarwanda it's 38MB.

Is this one of the openoffice dictionaries? Once we have them all parse successfully it'd be a good baseline test.

donnerpeter:
Will this buffer help even if these flags occur at the 37th MB? :)

The mentioned dictionary is in the wooorm collection: https://github.com/wooorm/dictionaries/tree/main/dictionaries/rw

dweiss:
No, obviously not. But I think we can empirically check on existing data where those required flags actually are (how much buffer is realistically needed). If there is a file where a flag occurs at 37mb then I think it could just thrown an IOException with an appropriate message. If somebody *really* wants to, they can pass an InputStream supporting mark/reset which would support such a large scan window... I don't have all the details worked out. This is fine as it is now too, but... you know - why not try to make it better. ;)

donnerpeter:
Checking against available dictionaries looks like a good idea!

dweiss:
I'll commit this patch in and create a separate issue for improving it later.

donnerpeter:
It's already merged :)

dweiss:
Yup, yup. Saw that after I posted the comment.

PR #2281, opened 2021-02-01
SOLR-15118: Convert /v2/collections APIs to annotations

gerlowskija:
# Description

Solr supports two different ways to write v2 APIs: a JSON spec based
approach, and one based on annotated POJOs.  The POJO method is now
preferred.

This commit switches the /v2/collections APIs over to the
annotation-based approach.

# Solution

This commit introduces a relatively rote migration of the JSON spec over to annotated POJOs.  There's little of note here, except for cases where the v2 JSON spec was missing support for params supported by the v1 APIs.  Here I had to make a few decisions as to whether property names should be retained, etc.  A good example of this is the v2 backup API, which had several params not covered by the existing v2 spec.

# Tests

Adds a new test: `V2CollectionsAPIMappingTest.java` to test the v2->v1 conversion.

# Checklist


solr/core/src/java/org/apache/solr/handler/CollectionsAPI.java
```diff
@@ -50,11 +81,144 @@ public void getCollections(SolrQueryRequest req, SolrQueryResponse rsp) throws E
...
+            }
+        }
+
+        private void flattenMapWithPrefix(Map<String, Object> toFlatten, Map<String, Object> destination,
+                                          String additionalPrefix) {
+            if (toFlatten == null || toFlatten.isEmpty() || destination == null) {
+                return;
+            }
+
+            toFlatten.forEach((k, v) -> destination.put(additionalPrefix + k, v));
```
sigram, line 213: Maybe we should make this recursive to support potentially nested v2 payloads?

solr/solrj/src/java/org/apache/solr/client/solrj/request/beans/BackupCollectionBody.java
```diff
@@ -0,0 +1,51 @@
...
+
+import org.apache.solr.common.annotation.JsonProperty;
+import org.apache.solr.common.util.ReflectMapWriter;
+
+/**
+ * V2 API POJO for the /v2/collections 'backup-collection' command.
+ *
+ * Analogous to the request parameters for v1 /admin/collections?action=BACKUP API.
+ */
+public class BackupCollectionBody implements ReflectMapWriter {
```
sigram, line 27: What do you think about renaming all of these to use `*Payload` instead of `*Body`?

solr/solrj/src/java/org/apache/solr/client/solrj/request/beans/CreateBody.java
```diff
@@ -0,0 +1,77 @@
...
+    @JsonProperty
+    public String async;
+
+    @JsonProperty
+    public Boolean waitForFinalState;
+
+    @JsonProperty
+    public Boolean perReplicaState;
+
+    public static class RouterInfo implements ReflectMapWriter {
```
sigram, line 71: Router spec may in fact contain arbitrary properties in addition to these two - see `DocRouter.getRouterSpec`. Currently only `name` and `field` are used in the routers that Solr provides but other users may use some other properties in their implementations.

noblepaul:
Thanks for picking this up @gerlowskija 

Please let me know if you need any help

solr/core/src/java/org/apache/solr/handler/CollectionsAPI.java
```diff
@@ -36,7 +58,16 @@
  */
 public class CollectionsAPI {
 
-  private final CollectionsHandler collectionsHandler;
+    public static final String V2_CREATE_COLLECTION_CMD = "create";
+    public static final String V2_BACKUP_CMD = "backup-collection";
+    public static final String V2_RESTORE_CMD = "restore-collection";
+    public static final String V2_CREATE_ALIAS_CMD = "create-alias";
+    public static final String V2_SET_ALIAS_PROP_CMD = "set-alias-property";
+    public static final String V2_DELETE_ALIAS_CMD = "delete-alias";
```
sigram, line 66: Should we also add other collection commands here, such as: COLLECTIONPROP, CREATE / DELETESHARD, ADD / MOVE / DELETEREPLICA, MODIFYCOLLECTION, REINDEXCOLLECTION, RENAME, SPLITSHARD ?

noblepaul: Yes, please

solr/solrj/src/java/org/apache/solr/client/solrj/request/beans/CreateBody.java
gerlowskija: Good catch.  There's a Jackson annotation made for this case (`JsonAnyGetter`) - named properties will be mapped as usual but "extra" ones are put in a Map or whatever.  But it looks like we're not actually using Jackson here to avoid SolrJ needing to depend on Jackson.

I could create clone JsonAnyGetter the way we've done with JsonProperty, but idt it's worth it.  I'll just remove this RouterInfo class and handle all the properties as a Map for the translation.

Might be worth considering whether a Jackson-dep is worth it after all going forward though.

gerlowskija:
Hey, thanks for offering @noblepaul  .  I do have one thing I was puzzling over:

How is SolrJ's `getV2Request`/`V2RequestSupport` feature intended to work for annotation-based v2 APIs?  It looks like `V1ToV2ApiMapper` is really dependent on `CollectionApiMapping.Meta`, which the annotation-based APIs can't be found in obviously.

There are some annotation-based v2 APIs whose SolrJ request classes implement V2RequestSupport (e.g. CollectionAdminRequest.ClusterProp), so there should be logic somewhere to handle that I imagine?  Unless getV2Request will blow up now if called on certain CollectionAdminRequest objects?

More broadly I wonder whether there's any value in getV2Request going forward.  It never really "took off" to cover APIs other than CollectionAdmin, and there might be better ways to let SolrJ users dip their toes into the v2 waters (e.g. dedicated request classes that better align with the v2 paths/concepts? idk).

gerlowskija:
Going to merge this with the strategy for dealing with V2RequestSupport that I mentioned in my last comment.

If you eventually get a chance to reply to that comment above @noblepaul , lmk if you prefer some other approach and I can pick it back up then.

PR #2282, opened 2021-02-01
LUCENE-9615: Expose HnswGraphBuilder index-time hyperparameters as FieldType attributes

sbeniwal12:
Expose HnswGraphBuilder index-time hyperparameters as FieldType attributes

# Description

HnswGraphBuilder has a few tunables: maxConnections, beamWidth, and we may add a few more, such as whether to use a diversity heuristic when choosing neighbors to link in the graph. Currently these are locked to defaults set by global variables. Instead we should provide some interface for configuring them.

# Solution

Here with this change we allow to pass down these parameters as FieldType attributes of VectorFields as well provide a convenience method in VectorFields.java to  add these parameters in FieldType attributes. The value passed in the attributes would be used in `Lucene90VectorWriter ` while creating a HNSW graph, in case no value is found we switch back to default parameter values present in `HnswGraphBuilder`. 

# Tests

Updated way of setting these parameters  in `KnnGraphTester.java` for testing indexing and search performance of a knn-graph and added unit tests for new constructor and public method in `VectorField.java` in `TestVectorValues.java`.

# Checklist

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/codecs/lucene90/Lucene90VectorWriter.java
+++ b/lucene/core/src/java/org/apache/lucene/codecs/lucene90/Lucene90VectorWriter.java
@@ -123,7 +123,9 @@ public final class Lucene90VectorWriter extends VectorWriter {
             (RandomAccessVectorValuesProducer) vectors,
             vectorIndexOffset,
             offsets,
-            count);
+            count,
+            fieldInfo.getAttribute(HnswGraphBuilder.HNSW_MAX_CONN_ATTRIBUTE_KEY),
+            fieldInfo.getAttribute(HnswGraphBuilder.HNSW_BEAM_WIDTH_ATTRIBUTE_KEY));
       } else {
         throw new IllegalArgumentException(
             "Indexing an HNSW graph requires a random access vector values, got " + vectors);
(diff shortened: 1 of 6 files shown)


lucene/core/src/java/org/apache/lucene/codecs/lucene90/Lucene90VectorWriter.java
```diff
@@ -188,9 +190,29 @@ private void writeGraph(
...
-      int count)
+      int count,
+      String maxConnStr,
+      String beamWidthStr)
       throws IOException {
-    HnswGraphBuilder hnswGraphBuilder = new HnswGraphBuilder(vectorValues);
+    int maxConn, beamWidth;
+    if (maxConnStr == null) {
+      maxConn = HnswGraphBuilder.DEFAULT_MAX_CONN;
+    } else if (!maxConnStr.matches("[0-9]+")) {
```
msokolov, line 200: I don't think we need this - we can allow parseInt to throw an exception. Let's catch `NumberFormatException` and rethrow with more context (which attribute caused the exception). Also HnswGraphBuilder tests for `<= 0`, so we don't need to check that here.

lucene/core/src/java/org/apache/lucene/document/VectorField.java
```diff
@@ -53,6 +54,44 @@ private static FieldType getType(float[] v, VectorValues.SearchStrategy searchSt
...
+      int dimension, VectorValues.SearchStrategy searchStrategy, int maxConn, int beamWidth) {
+    if (dimension == 0) {
+      throw new IllegalArgumentException("cannot index an empty vector");
+    }
+    if (dimension > VectorValues.MAX_DIMENSIONS) {
+      throw new IllegalArgumentException(
+          "cannot index vectors with dimension greater than " + VectorValues.MAX_DIMENSIONS);
+    }
+    if (searchStrategy == null) {
+      throw new IllegalArgumentException("search strategy must not be null");
```
msokolov, line 77: Let's also assert `searchStrategy.isHnsw()` to catch attempts to use `NONE` or some other unsupported future strategy.

lucene/core/src/test/org/apache/lucene/util/hnsw/KnnGraphTester.java
```diff
@@ -132,13 +135,13 @@ private void run(String... args) throws Exception {
           if (iarg == args.length - 1) {
             throw new IllegalArgumentException("-beamWidthIndex requires a following number");
           }
-          HnswGraphBuilder.DEFAULT_BEAM_WIDTH = Integer.parseInt(args[++iarg]);
```
msokolov, line 135: With this change, we no longer have any need to make these static variables writable - let's change them to `final` in `HnswGraphBuilder`

lucene/core/src/java/org/apache/lucene/codecs/lucene90/Lucene90VectorWriter.java
sbeniwal12: Thanks for pointing out, threw `NumberFormatException` with message describing which attribute caused the exception.

lucene/core/src/java/org/apache/lucene/document/VectorField.java
sbeniwal12: Added this check and also added a unit test for this check.

lucene/core/src/test/org/apache/lucene/util/hnsw/KnnGraphTester.java
sbeniwal12: Made them final and also made changes to `TestKnnGraph.java` to accommodate this change.

msokolov (approved):
Looks good, thanks @sbeniwal12 !

PR #2283, opened 2021-02-01
LUCENE-9719: Resource files aren't deleted from build directory after…

donnerpeter:
… being deleted in source

# Description

e.g. *.sug files used in Hunspell

# Solution

Make Gradle `Sync` them instead of `Copy`

# Tests

No

# Checklist

### DIFF
--- a/gradle/ant-compat/folder-layout.gradle
+++ b/gradle/ant-compat/folder-layout.gradle
@@ -25,11 +25,15 @@ allprojects {
       test.resources.srcDirs = ['src/test-files']
     }
 
-    task copyTestResources(type: Copy) {
+    task copyTestResources(type: Sync) {
       from('src/test') {
         exclude '**/*.java'
       }
       into sourceSets.test.java.outputDir
+      preserve {
+        include '**/*.class'
+        include 'META-INF/services/*'
+      }
     }
     processTestResources.dependsOn copyTestResources
   }


gradle/ant-compat/folder-layout.gradle
```diff
@@ -25,11 +25,15 @@ allprojects {
...
 
-    task copyTestResources(type: Copy) {
+    task copyTestResources(type: Sync) {
       from('src/test') {
         exclude '**/*.java'
       }
       into sourceSets.test.java.outputDir
+      preserve {
+        include '**/*.class'
+        include 'META-INF/services/*'
```
donnerpeter, line 35: `Sync` deletes everything in the target which isn't in the source, and the target is a normal compiler output root, which contains class files etc. I've tried to enumerate things that shouldn't be deleted here, but I'm not terribly sure this list is exhaustive. I haven't seen issues with this settings so far, but I'm only working on Hunspell subset. What's worse, the issues I've seen before adding `preserve` appeared not immediately and somewhat randomly and weren't easy to decipher. So while the issue is there and should be addressed, I'm not sure this is the full fix.

dweiss:
This is for backward compatibility with ant and should be removed eventually. Don't make it a sync, it won't help. If you need to wipe out build, run gradlew clean. Sorry, this is ant mess and we shouldn't try to fix it - we should just leave it behind.

dweiss:
The way this should be cleaned up is to remove any resources from src/test (move them to the resources folder). But this will have to wait until Solr/Lucene repos are split.

donnerpeter:
Thanks for the explanation!

dweiss:
So, to summarize - moving those resource files to {{src/test-files}} should do the job. I don't know whether gradle handles syncing resources upon compilation (it should!). Even the plain sync task has its issues in gradle itself -
https://github.com/gradle/gradle/issues/1349

PR #2284, opened 2021-02-01
SOLR-11233: Add optional JAVA8_GC_LOG_FILE_OPTS for bin/solr.

cpoerschke:
https://issues.apache.org/jira/browse/SOLR-11233

Since `master` branch does not support Java8 this change is best tested on `branch_8x` but could then potentially still be cherry-picked to `master` branch if we want to keep the `bin/solr*` code on both branches similar.


solr/bin/solr
```diff
@@ -2026,7 +2026,11 @@ if [ "$GC_LOG_OPTS" != "" ]; then
     if [ "$JAVA_VENDOR" == "IBM J9" ]; then
       gc_log_flag="-Xverbosegclog"
     fi
-    GC_LOG_OPTS+=("$gc_log_flag:$SOLR_LOGS_DIR/solr_gc.log" '-XX:+UseGCLogFileRotation' '-XX:NumberOfGCLogFiles=9' '-XX:GCLogFileSize=20M')
+    if [ -z ${JAVA8_GC_LOG_FILE_OPTS+x} ]; then
```
madrob, line 2029: should be `:+`

cpoerschke: This variant is consistent with the existing variant at line 2010 but yes it confused me too when I first read it, looks there's a subtle difference between "unset" and "set to empty string" behaviour. Illustration:

```
$ unset GC_LOG_OPTS
$ if [ -z ${GC_LOG_OPTS+x} ]; then echo unset; else echo not unset; fi
unset
$ if [ -z ${GC_LOG_OPTS:+x} ]; then echo empty; else echo not empty; fi
empty
$ if [ -z ${GC_LOG_OPTS} ]; then echo empty; else echo not empty; fi
empty
$ 
$ GC_LOG_OPTS=
$ if [ -z ${GC_LOG_OPTS+x} ]; then echo unset; else echo not unset; fi
not unset
$ if [ -z ${GC_LOG_OPTS:+x} ]; then echo empty; else echo not empty; fi
empty
$ if [ -z ${GC_LOG_OPTS} ]; then echo empty; else echo not empty; fi
empty
$
$ GC_LOG_OPTS=foobar
$ if [ -z ${GC_LOG_OPTS+x} ]; then echo unset; else echo not unset; fi
not unset
$ if [ -z ${GC_LOG_OPTS:+x} ]; then echo empty; else echo not empty; fi
not empty
$ if [ -z ${GC_LOG_OPTS} ]; then echo empty; else echo not empty; fi
not empty
```

cpoerschke:
Returning to this today: pull request branch just updated with latest origin base.

If there are no further comments/questions/concerns/etc. from @madrob and/or others I will aim to proceed to merge these changes around this time next week. Thank you.

PR #2285, opened 2021-02-01
SOLR-14928: introduce distributed cluster state updates

murblanc:
SOLR-14928: introduce distributed cluster state updates (i.e. not having to go through a ZK queue and Overseer)

The motivation behind this PR is to simplify SolrCloud and be the first step to (eventually) a significant scale increase: handle orders of magnitude more collections than currently and manage cluster state caching in a more flexible way.

Changes introduced by this PR:

- Support of distributed state updates for collections through Compare and Swap for updating state.json Zookeeper files.
- Cluster wide configuration in `solr.xml` to pick the way cluster state updates are handled (default remains Overseer based updates)
- Randomization based on test seed of the cluster update strategy (Distributed vs Overseer) in tests using `MiniSolrCloudCluster`, so both execution paths get continuously tested.

A good entry point to explore this PR is class `DistributedClusterChangeUpdater`.

Performance wise, small (low replica count) collection creation is faster with the distributed strategy than with Overseer, and significantly faster when multiple collections are created in parallel (10 parallel threads continuously creating 4 replicas collections show a 30% reduction in collection creation time on my laptop). High replica count (~40 and more) collection creation is slower. This will be addressed by using Per Replica States that limit contention on access to a single state.json file and/or a future splitting of state.json into finer grain elements.

Next step after this one is distributing the execution of the Collection API commands to all nodes.

### DIFF
--- a/solr/core/src/java/org/apache/solr/cloud/DistributedClusterChangeUpdater.java
+++ b/solr/core/src/java/org/apache/solr/cloud/DistributedClusterChangeUpdater.java
@@ -1,0 +1,822 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.solr.cloud;
+
+import org.apache.solr.client.solrj.cloud.SolrCloudManager;
+import org.apache.solr.cloud.overseer.*;
+import org.apache.solr.common.SolrException;
+import org.apache.solr.common.cloud.*;
+import org.apache.solr.common.params.CollectionParams;
+import org.apache.solr.common.params.CommonParams;
+import org.apache.solr.common.util.Pair;
+import org.apache.solr.common.util.Utils;
+import org.apache.zookeeper.CreateMode;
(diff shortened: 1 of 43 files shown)


murblanc:
@noblepaul @chatman @madrob @sigram @markrmiller @thelabdude @tflobbe @gezapeti @gus-asf @HoustonPutman (people working in the area or who expressed interest in the past) let me know if you plan to have a look at this PR.
Barring need for rework, I'd like to merge this by the end of this week.
FYI the new distributed cluster updates strategy is off by default.

madrob:
I am interested in reviewing and testing this but thought we were still waiting on a rebase before this was more ready? I've started looking, but it's 2000 lines of code changed and not something I can easily digest in a single sitting. :)

murblanc:
Thanks @madrob. I did rebase on master before submitting the PR so this is totally ready for review (and merge, although I see there’s a conflict now with some recent change, likely simple to resolve).
The big rebase work was related to Per Replica States.

noblepaul:
Is not possible to make this a per collection feature? like the per-replica state feauture?

murblanc:
> Is not possible to make this a per collection feature? like the per-replica state feauture?

Given Zookeeper content is identical for Overseer based and distributed cluster state update strategies (as opposed to the Per Replica States that use different znodes), a SolrCloud cluster can be switched between the two strategies by a single change of configuration. This makes it easy to test and adopt distributed updates, or to quickly rollback to Overseer based updates in case of issues.
Doing this per collection would force modifying existing collections one by one.

The following phase is distributing the Collection API execution. This distribution will rely on the version and conditional updates of `state.json` to replace Collection API locking. Dynamically changing how Collection API commands are executed for a given collection (Overseer or distributed) would be subject to tricky race conditions so would likely not be supported. Per Replica States collection configuration changes on the other hand are safe, guarded by the Collection API locking.

In general, having all collections in the cluster follow the same execution and cluster state update strategies simplifies code and flow. For example enabling placement plugins to issue replica assignments for multiple collections or processing multiple collections for NodeDown.
Also, changes to the way cluster state is cached on nodes will be simpler (and their impact on node and Zookeeper memory and CPU utilization observable) if the node maintains a single cache strategy. Otherwise, if any new cluster state cache is maintained in addition to the existing one (to serve both Overseer based and distributed Collection API calls), things will always look worse...

murblanc:
Moved to [https://github.com/apache/lucene-solr/pull/2364](https://github.com/apache/lucene-solr/pull/2364)

PR #2291, opened 2021-02-02
SOLR-15122 Replace volatile+sleep with wait/notify

madrob:


solr/core/src/java/org/apache/solr/cluster/events/impl/DelegatingClusterEventProducer.java
```diff
@@ -90,7 +95,10 @@ public void setDelegate(ClusterEventProducer newDelegate) {
         log.debug("--- delegate {} already in state {}", delegate, delegate.getState());
       }
     }
-    this.version++;
+    synchronized (version) {
```
sigram, line 98: I don't think we need AtomicInteger if all sections that access `version` are synchronized?

solr/core/src/java/org/apache/solr/cluster/events/impl/DelegatingClusterEventProducer.java
```diff
@@ -144,7 +152,25 @@ public synchronized void stop() {
   }
 
   @VisibleForTesting
-  public int getVersion() {
-    return version;
+  public int waitForVersionChange(int currentVersion, int timeoutSec) throws InterruptedException, TimeoutException {
```
sigram, line 155: I debated whether to add this to the wrappers... it's only needed in tests. OTOH putting it here makes the test code much simpler.

solr/core/src/java/org/apache/solr/cluster/events/impl/DelegatingClusterEventProducer.java
```diff
@@ -90,7 +96,9 @@ public void setDelegate(ClusterEventProducer newDelegate) {
         log.debug("--- delegate {} already in state {}", delegate, delegate.getState());
       }
     }
-    this.version++;
+    if (versionTracker != null) {
```
murblanc, line 99: We have a synchronization issue (memory barrier type, not concurrent access type). The thread calling `setDelegate()` is accessing `versionTracker` set by another thread without synchronization.
Can be fixed by making `versionTracker` volatile.

solr/core/src/java/org/apache/solr/cluster/VersionTracker.java
```diff
@@ -0,0 +1,42 @@
...
+package org.apache.solr.cluster;
+
+import java.util.concurrent.TimeoutException;
+
+/**
+ * Allows for tracking state change from test classes. Typical use will be to set a version tracker on a stateful
+ * object, which will call {@link #increment()} every time state changes. Test clients observing the state will call
+ * {@link #waitForVersionChange(int, int)} to be notified of the next increment call.
+ */
+public interface VersionTracker {
```
murblanc, line 27: Tracking versions by incrementing is one possible implementation of this interface, but maybe the interface doesn't have to hint that this is the implementation?
Renaming `increment` into `notifyEvent` or something similar and `VersionTracker` into `NotificationCallback` would make it more generic (not suggesting these actual names, but you get the idea).

`waitForVersionChange` doesn't have to be part of the interface. Test code can use the implementation it instantiates directly. Some tests might not even want to wait but rather do other things with the notification (like count how many notifications are received, or make sure none are received etc).

solr/core/src/java/org/apache/solr/cluster/placement/impl/DelegatingPlacementPluginFactory.java
```diff
@@ -39,18 +39,20 @@ public PlacementPlugin createPluginInstance() {
...
 
+  @VisibleForTesting
+  public void setVersionTracker(VersionTracker tracker) {
+    versionTracker = tracker;
+  }
+
   public void setDelegate(PlacementPluginFactory<? extends PlacementPluginConfig> delegate) {
     this.delegate = delegate;
-    this.version++;
+    if (versionTracker != null) {
```
murblanc, line 49: If we want to be strict about correctness, we should copy `versionTracker` into another variable then test and call, otherwise it can be non `null` when evaluating the if condition then `null` on the next line when dereferencing it.

solr/core/src/test/org/apache/solr/cluster/VersionTrackerImpl.java
```diff
@@ -0,0 +1,57 @@
...
+    public synchronized void increment() {
+        version++;
+        this.notifyAll();
+    }
+
+    @Override
+    public int waitForVersionChange(int currentVersion, int timeoutSec) throws InterruptedException, TimeoutException {
+        TimeOut timeout = new TimeOut(timeoutSec, TimeUnit.SECONDS, TimeSource.NANO_TIME);
+        int newVersion = currentVersion;
+        while (! timeout.hasTimedOut()) {
```
murblanc, line 39: formatting: remove space after `!`

solr/core/src/test/org/apache/solr/cluster/VersionTrackerImpl.java
```diff
@@ -0,0 +1,57 @@
...
+            synchronized (this) {
+                if ((newVersion = version) != currentVersion) {
+                    break;
+                }
+                this.wait(timeout.timeLeft(TimeUnit.MILLISECONDS));
+            }
+        }
+        if (newVersion < currentVersion) {
+            // ArithmeticException? This means we overflowed
```
murblanc, line 48: Likely bad call from the test...

solr/core/src/test/org/apache/solr/cluster/VersionTrackerImpl.java
```diff
@@ -0,0 +1,57 @@
...
+            throw new RuntimeException("Invalid version - went back! currentVersion=" + currentVersion +
+                    " newVersion=" + newVersion);
+        } else if (newVersion == currentVersion) {
+            throw new TimeoutException("Timed out waiting for version change.");
```
murblanc, line 52: Add the version value to the exception, might help debug tests.

solr/core/src/test/org/apache/solr/cluster/VersionTrackerImpl.java
murblanc: I also realize I would have put the while loop inside the synchronized block, and now I wonder which way is better... Re-acquiring it over and over again or not holding it while seeing if a timeout occurred.

solr/core/src/test/org/apache/solr/cluster/events/ClusterEventProducerTest.java
```diff
@@ -102,7 +105,7 @@ public void teardown() throws Exception {
 
   @Test
   public void testEvents() throws Exception {
-    int version = waitForVersionChange(-1, 10);
+    int version = versionTracker.waitForVersionChange(-1, 10);
```
murblanc, line 108: Version starts at 0 in the implementation of the version tracker, so what does it mean to wait for it to be -1?

murblanc: Maybe we need a `getCurrentVersion` in `VersionTrackerImpl` so a given instance can be reused more easily in a given test?

solr/core/src/test/org/apache/solr/cluster/VersionTrackerImpl.java
madrob: It's hard to do profiling on this since it generally only loops once on my machine, but I'll switch it to the loop inside of the block. It's part of the condition that we are implicitly checking with the wait, so it makes sense this way too.

solr/core/src/test/org/apache/solr/cluster/events/ClusterEventProducerTest.java
madrob: -1 is the last value we've "seen", I'll add some docs around this. Effectively this is a getVersion in this implementation.

solr/core/src/test/org/apache/solr/cluster/CountingStateChangeListener.java
```diff
@@ -0,0 +1,68 @@
...
+     * occurred beyond the known value, this method will wait for additional changes to come in.
+     * If the current number of change events is unknown to the caller, then this method can be called with <tt>-1</tt>
+     * to return immediately with the number of events up to this point.
+     * @param lastVersion the previous number of changes seen
+     * @param timeoutSec how long to wait for additional changes to occur
+     * @return the number of changes seen since initialization
+     */
+    public int waitForVersionChange(int lastVersion, int timeoutSec) throws InterruptedException, TimeoutException {
+        TimeOut timeout = new TimeOut(timeoutSec, TimeUnit.SECONDS, TimeSource.NANO_TIME);
+        int newVersion = lastVersion;
```
murblanc, line 51: This assignment is not needed. I assume the compiler is unhappy if it is just removed, but maybe doing the assignment to `newVersion` first before the timeout check in the while condition makes the compiler happy?

solr/core/src/test/org/apache/solr/cluster/CountingStateChangeListener.java
```diff
@@ -0,0 +1,68 @@
...
+        synchronized (this) {
+            while (!timeout.hasTimedOut() && (newVersion = version) != lastVersion) {
```
murblanc, line 53: I don't get the condition here. Shouldn't we loop while `version == lastVersion` (so we exit the loop when it changes or when we time out) rather than loop while they're different?

I suspect that the test without this improved lockstep synchronization was passing always on your machine and it continues to pass for the same reason.

madrob: It's a copy-paste error

sigram:
+1, this looks neat! I wonder how many other places could use such refactoring...

solr/core/src/java/org/apache/solr/cluster/events/impl/DelegatingClusterEventProducer.java
```diff
@@ -90,7 +101,11 @@ public void setDelegate(ClusterEventProducer newDelegate) {
         log.debug("--- delegate {} already in state {}", delegate, delegate.getState());
       }
     }
-    this.version++;
+    Phaser localPhaser = phaser; // volatile read
+    if (localPhaser != null) {
+      assert localPhaser.getRegisteredParties() == 1;
+      localPhaser.arrive(); // we should be the only ones registered, so this will advance phase each time
```
murblanc, line 107: Why make the assumption this class is the only one used by the test? Maybe the test uses the `Phaser` to synchronize multiple classes? That just means remove the assert and the comment.

But here's the issue I alluded to in the Jira: by exposing directly `Phaser` and not a test provided implementation of an interface method to call, we restrict the usage of such synchronization to only allow the test code to know when something took place in the production code.
If we expose an interface, it allows the test to stall the production code at this stage, maybe to make it wait for another condition to happen (for example let the test verify the current state produced by the production code) and then the test can decide to resume the production code. This is useful when verifying state machines for example, as otherwise the state transitions can be so fast that it's impossible to verify reliably (by the time the test is notified of the first transition and tries to verify it, the state machine has made 5 more transitions...).

I'd really prefer to expose an interface as in the previous version of the code. The have the test implement it by basically delegating to `Phaser`.

madrob: While I agree that all of those things are useful sounding in theory, we don't ostensibly need them right now, and I'd prefer to keep the changes smaller. I don't like the idea of adding an interface with a single test implementation on the hunch that someday somebody else might want to do something else with it. If that's the case, we can refactor it in the future.

madrob: In the JIRA issue you specifically suggested "maybe the test can provide a latch" which would be a specific synchronization implementation, but a latch happened to be insufficient because we need something that can block for us multiple times. I don't see this as too divergent from that idea, and have seen other code available for injecting latches specifically. It feels the same.

murblanc: Fair enough.
Looks good to me!

PR #2293, opened 2021-02-03
LUCENE-9725: Allow BM25FQuery to use other similarities.

jtibshirani:
From a high level, BM25FQuery (1) computes statistic that represent the combined
field content and (2) passes these to a score function. This model makes sense
for many similarities besides BM25.

This PR unhardcodes BM25Similarity in BM25FQuery and instead uses the one
configured on IndexSearcher. It also renames BM25FQuery since it's no longer
specific to BM25.

### DIFF
--- a/lucene/sandbox/src/java/org/apache/lucene/sandbox/search/CombinedFieldQuery.java
+++ b/lucene/sandbox/src/java/org/apache/lucene/sandbox/search/CombinedFieldQuery.java
@@ -53,9 +53,7 @@ import org.apache.lucene.search.TermQuery;
 import org.apache.lucene.search.TermScorer;
 import org.apache.lucene.search.TermStatistics;
 import org.apache.lucene.search.Weight;
-import org.apache.lucene.search.similarities.BM25Similarity;
 import org.apache.lucene.search.similarities.Similarity;
-import org.apache.lucene.search.similarities.SimilarityBase;
 import org.apache.lucene.util.Accountable;
 import org.apache.lucene.util.BytesRef;
 import org.apache.lucene.util.RamUsageEstimator;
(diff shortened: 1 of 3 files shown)


lucene/sandbox/src/java/org/apache/lucene/sandbox/search/CombinedFieldQuery.java
```diff
@@ -64,39 +62,35 @@
...
+ * <ol>
+ *   <li>Given a list of fields and weights, it pretends there is a synthetic combined field where
+ *       all terms have been indexed. It computes new term and collection statistics for this
+ *       combined field.
+ *   <li>It uses a disjunction iterator and {@link IndexSearcher#getSimilarity} to score documents.
+ * </ol>
+ *
+ * <p>In order for a similarity to be compatible, {@link Similarity#computeNorm} must be additive:
+ * the norm of the combined field is the sum of norms for each individual field. This is usually
+ * true, since norms often represent the field length. Per-field similarities are not supported.
```
jpountz, line 76: The requirement is actually stronger, we need a similarity that uses an additive normalization factor AND that encodes it using `SmallFloat#intToByte4` in the index since the decoding of norms is hardcoded as `SmallFloat#byte4ToInt` in `MultiFieldNormValues#advanceExact`.

Also maybe mention explicitly that e.g. `BM25Similarity` and `DFRSimilarity` meet this requirement?

jpountz (approved):

jimczi (approved):
LGTM too, I started with a fixed BM25Similarity but the goal was to be able to use different `k1` or `b` on a per-field basis. I prefer your approach which is to make the query works with other similarity first. We can make `Similarity` aware of multi-field scoring in a follow up.

lucene/sandbox/src/java/org/apache/lucene/sandbox/search/CombinedFieldQuery.java
jtibshirani: Good point! I mentioned those similarities and restored the reference to `SimilarityBase#computeNorm`, which gives a clear example of what's acceptable.

jtibshirani:
Thanks @jpountz and @jimczi for reviewing.

PR #2301, opened 2021-02-04
LUCENE-9732: Hunspell: support dictionary entries starting with slash

donnerpeter:
# Description

Slash shouldn't be escaped at the very beginning of words

# Solution

Support that :)

# Tests

`TestEscaped` enhanced based in `slash` from Hunspell/C++

# Checklist


lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/WordCase.java
```diff
@@ -53,14 +54,19 @@ static WordCase caseOf(CharSequence word, int length) {
...
     }
 
-    return get(startsWithLower, seenUpper, seenLower);
+    return get(startCase, seenUpper, seenLower);
   }
 
-  private static WordCase get(boolean startsWithLower, boolean seenUpper, boolean seenLower) {
-    if (!startsWithLower) {
-      return !seenLower ? UPPER : !seenUpper ? TITLE : MIXED;
+  private static WordCase get(CharCase startCase, boolean seenUpper, boolean seenLower) {
```
dweiss, line 60: Darn this method is case-crazy. :)
Can you add a javadoc here and just show an example of what each of these WordCases are (or attach such an example to WordCase's enum values)?

I'd also remove the negation and swap arguments in case UPPER - one thing less to reason about.

donnerpeter: I'll add examples, a good idea.
As for UPPER, and given that I hate negations, I still find this easier to read. If "not seen lower" then UPPER, makes sense. If "seen lower" then TITLE — makes less sense to me. How about you?

dweiss: I tried to follow that routine but fell into infinite recursion somewhere half way... :) The docs explain what it does much better!

PR #2303, opened 2021-02-04
LUCENE-9715: fix int overflow in Lucene90VectorReader

msokolov:


mikemccand (approved):
Yay Lucene randomized tests!

msokolov:
I like Lucene's randomized tests too! But they didn't find this one. Since it requires a 2GB index to detect, it only turned up in the context of Anton's version of "nightly" benchmarks - I wonder if they are crashing for you?

Separately I think it would be cool if we did have a way of surfacing such bugs with randomized tests.

Anyway I'll push this now.

mikemccand:
> I like Lucene's randomized tests too! But they didn't find this one. Since it requires a 2GB index to detect, it only turned up in the context of Anton's version of "nightly" benchmarks - I wonder if they are crashing for you?

Ahh, good point.  It is also exciting when our benchmarks catch bugs that our unit tests miss!

Oddly/scarily, the nightly benchmarks are  NOT failing for me -- I just checked last night's.  I wonder why not?  Hmm, could it be because we accidentally stopped retrieving vectors in search tasks?

PR #2309, opened 2021-02-05
SOLR-14301: Remove external commons-codec usage in gradle validateJarChecksums

asalamon74:
…Checksums

# Description

Right now gradle calculates SHA-1 checksums using an external commons-codec library. We can calculate SHA-1 using Java 8 classes, no need for commons-codec here.

# Solution

Eliminate commons-codec usage in checksum validation.

# Tests

./gradlew validateJarChecksums

# Checklist

### DIFF
--- a/gradle/validation/jar-checks.gradle
+++ b/gradle/validation/jar-checks.gradle
@@ -20,21 +20,11 @@
 // 2) notice file
 // 3) checksum validation/ generation.
 
-import org.apache.commons.codec.digest.DigestUtils
+import java.nio.file.Files
+import java.security.MessageDigest
 
 // This should be false only for debugging.
 def failOnError = true
-
-// We're using commons-codec for computing checksums.
-buildscript {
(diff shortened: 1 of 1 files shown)


madrob:
I _believe_ this was done with an external dependency because there's an issue with how checksums were handled on Windows, but I don't remember all the details and can't find it now. Maybe @dweiss can remember.

We also do the same thing in `releasing.gradle`

dweiss:
I think it's just much shorter invocation in the end - when you look at the diff you'll see it. I also don't know if the checksum format is required to be supported by the JDK.

asalamon74:
Yes, the invocation is longer with this patch, on the other hand we avoid using an external library so the .gradle file is shorter.

We only require that SHA-1 is supported by the JDK, which is a requirement for all JDK implementations: https://docs.oracle.com/javase/7/docs/api/java/security/MessageDigest.html

PR #2310, opened 2021-02-05
LUCENE-9705: Create Lucene90PostingsFormat

jtibshirani:
For now this is just a copy of Lucene90PostingsFormat. The existing
Lucene84PostingsFormat was moved to backwards-codecs, along with its utility
classes.


jtibshirani:
I also plan to copy over the blocktree dependencies like `BlockTreeTermsWriter`. Any thoughts on naming or package structure for these classes? In [LUCENE-9616](https://issues.apache.org/jira/browse/LUCENE-9616) @jpountz mentioned adding version numbers to the names.

iverase:
I have a look and I have two thoughts:

1) I wonder if we should not version PFUtil classes, instead move then to a package under Util and change the visibility of the methods. Those classes seems more like a utility to me.

2) Under the codecs module, there are plenty of specialise posting readers, they all have hardcodes the posting format version. I wonder if we can expose through the posting format the creation of `PostingsReaderBase`? From the SegmentReadState we should be able to get the current codec.

jpountz:
> I wonder if we should not version PFUtil classes, instead move then to a package under Util and change the visibility of the methods. Those classes seems more like a utility to me.

I've become a bit wary of having shared utility classes for codecs given how it makes the code harder to evolve (e.g. I have the FST and PackedInts classes in mind). I'd rather like to copy this utility class wherever it's needed so that every file format that uses bit packing can more easily update the logic to fits its own needs.

jpountz:
> Any thoughts on naming or package structure for these classes?

What about putting the current blocktree classes into the `org.apache.lucene.backward_codecs.lucene40` (since they were introduced in Lucene 4.0) package and renaming the reader/writer classes to include the version too, ie. `BlockTree40TermsWriter` and `BlockTree40TermsReader`? And the new classes would be called `BlockTree90TermsWriter` and `BlockTree90TermsReader` and be under the `org.apache.lucene.codecs.lucene90` package?

jtibshirani:
Moving these classes under versioned packages like `org.apache.lucene.codecs.lucene90` makes sense to me. I slightly prefer the name `Lucene90BlockTreeTermsReader` because it isn't always clear where to stick the version number. But no strong opinion, and I see we already have some classes like `Completion84PostingsFormat`.

jpountz:
These names work for me too. Agreed we don't have to have numbers in small helper classes.

jpountz (approved):

jtibshirani:
@jpountz @iverase I pushed some changes, would you be able to take one last look?

jpountz (approved):
The new `FieldReader` still has this line:

```java
    if (metaIn == indexIn) { // Only true before Lucene 8.6
```

Should we remove this conditional now?

Otherwise LGTM.

PR #2317, opened 2021-02-07
LUCENE-9741: Add sequential optimization for stored fields

dnhatn:
If we are reading the stored-fields of document ids (25, 27, 28, 26, 99), and doc-25 triggers the stored-fields reader to decompress a block containing document ids [10-50], then we can tell the reader to read not only 25, but 26, 27, and 28 to avoid decompressing that block multiple times.

This PR proposes adding a new optimized instance of stored-fields reader that allows users to select the preferred fetching range.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/codecs/StoredFieldsReader.java
+++ b/lucene/core/src/java/org/apache/lucene/codecs/StoredFieldsReader.java
@@ -58,4 +58,95 @@ public abstract class StoredFieldsReader implements Cloneable, Closeable, Accoun
   public StoredFieldsReader getMergeInstance() {
     return this;
   }
+
+  /**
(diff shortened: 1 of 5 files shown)


jimczi (commented):
I wonder if prefetch is the best option here. If the documents to retrieve are close it should be the responsibility of the API to decompress less blocks. So something like a bulk API:
```
void visitDocument(int[] docIDs, DocumentStoredFieldVisitor visitor) throws IOException;
```
? 
I also don't think we should allow complex optimization if documents are not retrieved in order. Currently the choice is between a reader that aggressively decompress for sequential access but we could also have a middle ground where only the last block is kept ? That would be easier than a bulk API.

dnhatn:
@jimczi Thank you for reviewing. I am closing this PR because the proposed API is too complex.

PR #2327, opened 2021-02-08
LUCENE-9740: scan affix stream once.

dweiss:


lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/Dictionary.java
```diff
@@ -70,6 +70,9 @@
 
 /** In-memory structure for the dictionary (.dic) and affix (.aff) data of a hunspell dictionary. */
 public class Dictionary {
+  // Derived from woorm/ openoffice dictionaries.
```
donnerpeter, line 73: LibreOffice?

lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/Dictionary.java
```diff
@@ -224,26 +227,37 @@ public Dictionary(
...
-    boolean success = false;
-    try {
-      // Copy contents of the affix stream to a temp file.
-      try (OutputStream os = Files.newOutputStream(aff)) {
-        affix.transferTo(os);
+    try (BufferedInputStream affixStream =
+        new BufferedInputStream(affix, MAX_PROLOGUE_SCAN_WINDOW) {
+          @Override
+          public void close() throws IOException {
+            // TODO: maybe we should consume and close it? Why does it need to stay open?
```
donnerpeter, line 234: Probably so that the callers who opened the streams can close them safely using try-with-resources.

lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/Dictionary.java
```diff
@@ -346,6 +352,7 @@ private void readAffixFile(InputStream affixStream, CharsetDecoder decoder, Flag
       if (line.isEmpty()) continue;
 
       String firstWord = line.split("\\s")[0];
+      // TODO: convert to a switch?
```
donnerpeter, line 355: I thought about that. Maybe switch expression, when the language level allows that. Switch with break statements would be too verbose for my taste

lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/Dictionary.java
```diff
@@ -778,31 +791,36 @@ char affixData(int affixIndex, int offset) {
...
+      } else {
+        continue;
       }
+
+      if (charsetFound && flagFound) {
+        break;
+      }
+    }
+
+    if (flagFound) {
```
donnerpeter, line 822: flagLine != null?

dweiss: If flagLine is true then line had to be != null, otherwise you'd get an NPE earlier on line.split?

lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/Dictionary.java
dweiss: Ouch. Can you correct it for my and piggyback on any other subsequent patch? I overlooked this one.

lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/Dictionary.java
dweiss: Closeable.close() can be invoked any number of times without side-effects - this is a contract stated in the javadoc.

lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/Dictionary.java
dweiss: Ok. My taste tells me it'd be cleaner than that multi-level-if, especially that not all statements are identical there (one compares to the entire line if I believe). It may be worth considering the switch for performance reasons too but don't know if you'd see the difference.

lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/Dictionary.java
donnerpeter: True. It feels a bit more right to me when the ones creating the streams close them. But in fact I don't like the whole idea of passing streams to constructor. I believe most clients would be happier with passing paths (except the rare(?) cases when the content is created in memory or loaded from classpath).

lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/Dictionary.java
donnerpeter: Yes. I mean that `flagFound` is excessive since we already have `flagLine`

lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/Dictionary.java
donnerpeter: A bit cleaner, yes, but verbosity and the risk of forgetting a `break` outweigh that for me.
I also considered creating a map from first word into parsing lambdas, but decided it'd be quite verbose as well.

Performance should be negligible here. Last time I checked, it was dominated by writing/sorting/reading dic entries and building FSTs.

lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/Dictionary.java
dweiss: Leave it a stream, it's hard to beat its flexibility.

lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/Dictionary.java
dweiss: ok.

lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/Dictionary.java
dweiss: Yeah, I left that intentionally for parity with flagFound...

donnerpeter: It could be paired by using something nullable encoding-related :) Anyway it's very minor.

PR #2331, opened 2021-02-09
LUCENE-9322:  Lucene90VectorWriter can leak open files

iverase:
While trying to add a test class for Vectors based on `BaseIndexFileFormatTestCase`, a bug surface on the Lucene90VectorWriter constructor. If there is an exception during construction, it might happen that files are not properly closed and therefore it might leak.

Here is the proposal, move the current `TestVectorValues` to a `BaseVectorFormatTestCase` which extends `BaseIndexFileFormatTestCase`.

Fix the constructor so it handle closing files on error properly.


msokolov (approved):
Thanks for cleaning this up and fixing the atomicity of file operations, @iverase !

zacharymorn:
Hi, 

I was looking into https://issues.apache.org/jira/browse/LUCENE-9639 earlier and had the same thought of `BaseVectorFormatTestCase` seems to be missing, which led me to do some search and found this PR. I'm wondering if unit test for `SimpleTextVectorFormat` could be added as part of this PR as well since it appears to be a small change? If it's better to keep this PR focused, I'm willing to pick that up and submit a PR once this one has merged.

iverase:
@zacharymorn I think we can open another PR for the change you are proposing. Thanks!

PR #2334, opened 2021-02-09
LUCENE-9705: Create Lucene90TermVectorsFormat

iverase:
For now this is just a copy of Lucene90TermVectorsFormat. The existing
Lucene50TermVectorsFormat was moved to backwards-codecs, along with its utility
classes.


lucene/backward-codecs/src/java/org/apache/lucene/backward_codecs/compressing/Lucene50CompressingTermVectorsFormat.java
```diff
@@ -0,0 +1,113 @@
...
+import org.apache.lucene.store.Directory;
+import org.apache.lucene.store.IOContext;
+
+/**
+ * A {@link TermVectorsFormat} that compresses chunks of documents together in order to improve the
+ * compression ratio.
+ *
+ * @lucene.experimental
+ */
+public class Lucene50CompressingTermVectorsFormat extends TermVectorsFormat {
```
jtibshirani, line 37: This seems like a good name change to me. Thinking about the general pattern (so our other PRs are consistent):
* For substantial codec classes like 'format', 'reader', 'writer', we add `LuceneXX` to the name if it's not already there. This makes it clear the class follows our backwards compatibility scheme.
* For smaller helper classes like `FieldsReader` we move them but keep the name as-is.

Does this make sense to you? A couple other questions:
* For consistency, we could also rename the current `CompressingTermVectorsFormat` to `Lucene90CompressingTermVectorsFormat`.
* Should these move to version-specific packages like `lucene50` since they're versioned?

jtibshirani: We discussed in #2310 and agreed on this package/ naming pattern.

iverase: @jtibshirani I hope I addressed the suggestions here. I think this is ready for another review. Thanks!

lucene/core/src/java/org/apache/lucene/codecs/compressing/Lucene90CompressingTermVectorsFormat.java
```diff
@@ -33,7 +33,7 @@
  *
  * @lucene.experimental
  */
-public class CompressingTermVectorsFormat extends TermVectorsFormat {
+public class Lucene90CompressingTermVectorsFormat extends TermVectorsFormat {
```
jtibshirani, line 36: I think all these classes should be moved under the package `lucene90.compressing`. That way all of our "versioned" format files (plus their helper classes) will be in a package with the version name, ready to move to backwards-codecs when needed. This will be consistent with #2310.

jtibshirani (approved):
I left a couple comments, once those are addressed it looks ready to merge!

lucene/core/src/java/org/apache/lucene/codecs/compressing/Lucene90CompressingTermVectorsFormat.java
iverase: I will do that once we move Lucene87StoredFieldFormat

PR #2339, opened 2021-02-10
LUCENE-9705: Reset internal version in Lucene90FieldInfosFormat.

jtibshirani:
Since this is a fresh format, we can remove older version logic and reset the
internal version to 0.


iverase (approved):
lgtm, good catch

PR #2342, opened 2021-02-10
LUCENE-9406: Add IndexWriterEventListener to track events in IndexWriter

zacharymorn:
# Description / Solution

Add IndexWriterEventListener interface to track events in IndexWriter

# Tests
* Passed existing tests
* Added an additional test

Note: precommit is currently failing due to nocommit comment

# Checklist

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
+++ b/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
@@ -3518,21 +3518,11 @@ public class IndexWriter
       }
 
       if (pointInTimeMerges != null) {
-        MergePolicy.OneMerge nextMerge = null;
-
-        if (pendingMerges.size() > 0) {
-          // nocommit getting OneMerge instance here via mergeSource.getNextMerge() will
-          // change the state of pendingMerges, hanging tests.
-          // Is the current solution the right approach to get the merge snapshot for event
-          // callback?
-          nextMerge = pendingMerges.getFirst();
-        }
-
         if (infoStream.isEnabled("IW")) {
           infoStream.message(
               "IW", "now run merges during commit: " + pointInTimeMerges.segString(directory));
         }
-        config.getIndexWriterEventListener().beginMergeOnFullFlush(nextMerge);
+        config.getIndexWriterEventListener().beginMergeOnFullFlush(pointInTimeMerges);
 
         mergeScheduler.merge(mergeSource, MergeTrigger.COMMIT);
         pointInTimeMerges.await(maxCommitMergeWaitMillis, TimeUnit.MILLISECONDS);
@@ -3540,7 +3530,7 @@ public class IndexWriter
         if (infoStream.isEnabled("IW")) {
           infoStream.message("IW", "done waiting for merges during commit");
         }
-        config.getIndexWriterEventListener().endMergeOnFullFlush(nextMerge);
+        config.getIndexWriterEventListener().endMergeOnFullFlush(pointInTimeMerges);
 
         synchronized (this) {
           // we need to call this under lock since mergeFinished above is also called under the IW
--- a/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
+++ b/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
@@ -3518,15 +3518,30 @@ public class IndexWriter
       }
 
       if (pointInTimeMerges != null) {
+        MergePolicy.OneMerge nextMerge = null;
+
+        if (pendingMerges.size() > 0) {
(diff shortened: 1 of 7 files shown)


lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
```diff
@@ -3518,15 +3518,30 @@ private long prepareCommitInternal() throws IOException {
       }
 
       if (pointInTimeMerges != null) {
+        MergePolicy.OneMerge nextMerge = null;
+
+        if (pendingMerges.size() > 0) {
+          // nocommit getting OneMerge instance here via mergeSource.getNextMerge() will
```
mikemccand, line 3524: How about passing the `pendingMerges` to the event listener instead?  (Sorry if I asked for `OneMerge` on the issue!  `MergeSpecification` is better since it can hold multiple merges, allows event listener to log things like how many merges were requested during `commit`, etc.).

lucene/core/src/test/org/apache/lucene/index/TestIndexWriterMergePolicy.java
```diff
@@ -388,6 +388,69 @@ public void testMergeOnCommit() throws IOException {
...
+      endMergeCalled = true;
+    }
+
+    public boolean isEventsRecorded() {
+      return beginMergeCalled && endMergeCalled;
+    }
+  }
+
+  // Test basic semantics of merge on commit and events recording invocation
+  public void testMergeOnCommitWithEventListener() throws IOException {
```
mikemccand, line 411: You might also edit `LuceneTestCase.newIndexWriterConfig` to randomly swap in a `MockIndexWriterEventListener` just to exercise this listener in any tests using that API, which is quite a few.  It could uncover times when we accidentally break something when this listener is invoked ...

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
zacharymorn: No problem! From the context I assume you are actually meaning to pass `pointInTimeMerges` into event listener, since `pendingMerges` is of type `Deque<MergePolicy.OneMerge>`? I've pushed a commit to use `pointInTimeMerges` for now.

lucene/core/src/test/org/apache/lucene/index/TestIndexWriterMergePolicy.java
zacharymorn: Makes sense. Added.

dweiss:
I think it's good overall but I'm wondering whether it makes sense to make that field volatile... do we want to allow changing listeners over index writer lifecycle? I think it should be a regular field and IW should just read it once (and set forever).

zacharymorn:
> I think it's good overall but I'm wondering whether it makes sense to make that field volatile... do we want to allow changing listeners over index writer lifecycle? I think it should be a regular field and IW should just read it once (and set forever).

Thanks for the feedback Dawid! I thought about this a bit as well when I noticed most of the other fields are volatile, but I can't decide for sure which direction to go as I don't have enough context information about the different use cases IW may need to support in the wild, so I ended up following the existing pattern here. However, I do feel that it may not hurt for application to have the freedom to switch event listener in the middle if situation or need dictates (i.e. a service application  containing Lucene may need to switch event listener if it exposes an API for its client to choose how the IW event stream to be sent and stored) ?

dweiss:
If such a situation arises folks will have the ability to write a custom proxy listener with the ability to switch, add and/or remove delegates. This keeps the freedom to do anything you said and at the same time simplifies implementation (and concurrency-issues) on IW side?

zacharymorn:
> If such a situation arises folks will have the ability to write a custom proxy listener with the ability to switch, add and/or remove delegates. This keeps the freedom to do anything you said and at the same time simplifies implementation (and concurrency-issues) on IW side?

Ah that's very true! I've pushed a commit to remove volatile then. Please let me know if it looks good.

dweiss (commented):
Yep, looks good to me!

zacharymorn:
Circling back to this PR as I just realized it's not closed yet.. :D 

Thanks @dweiss for the review!

@mikemccand do you think this PR is ready now? Anything else you would like me to improve on?

dweiss:
Looks good to me. I've merged this in, sorry for the delay.

zacharymorn:
No worry! Thanks for the review and merge!

mikemccand:
Woops, sorry for the belated response, and thank you @zacharymorn for creating this and @dweiss for merging -- it looks great!  We can now add other events to track incrementally over time ...

zacharymorn:
No problem, and thanks for the review feedback as well Michael!

PR #2343, opened 2021-02-10
SOLR-15100: make ConfigSetService configurable in solr.xml

baisui1981:
# Description

modify the static method ConfigSetService.createConfigSetService in that I can define the customized class name of ConfigSetService in solr.xml and create it in the method,  [https://issues.apache.org/jira/browse/SOLR-15100](https://issues.apache.org/jira/browse/SOLR-15100)

# Solution

modify the static method ConfigSetService.createConfigSetService , get property 'configSetService' which is name of customized ConfigSetService class name config in solr.xml and initialize it by java refect

# Tests

add a test case in TestCoreContainer, test method is testCustomConfigSetService()

# Checklist


cffyh:
good job! I need this  feature customized ConfigSetService

solr/core/src/java/org/apache/solr/core/ConfigSetService.java
```diff
@@ -56,11 +58,24 @@
...
 
   public static ConfigSetService createConfigSetService(NodeConfig nodeConfig, SolrResourceLoader loader, ZkController zkController) {
-    if (zkController == null) {
+
+    String configSetServiceClass = nodeConfig.getConfigSetServiceClass();
+
+    if(configSetServiceClass != null){
+      try {
+        Class<? extends ConfigSetService> clazz = loader.findClass(configSetServiceClass, ConfigSetService.class);
+        Constructor<? extends ConfigSetService> constructor
```
dsmiley, line 67: I would rather we pass the NodeConfig instead of the boolean and let the implementation examine whatever it may want.  I think that's easier to maintain.

solr/core/src/java/org/apache/solr/core/NodeConfig.java
```diff
@@ -89,6 +89,8 @@
   private final boolean fromZookeeper;
   private final String defaultZkHost;
 
+  private final String configSetServiceClass;
```
dsmiley, line 92: Move up to the other classes to be more organized

solr/core/src/java/org/apache/solr/core/CoreContainer.java
```diff
@@ -2027,6 +2027,10 @@ public ConfigSetsHandler getConfigSetsHandler() {
     return configSetsHandler;
   }
 
+  public ConfigSetService getCoreConfigService() {
```
dsmiley, line 2030: Lets call this getConfigSetService and maybe rename the field now (or later)

dsmiley (changes requested):
Looks nice!  I just have some minor feedback.  Thank you for contributing!

solr/core/src/java/org/apache/solr/core/ConfigSetService.java
baisui1981: do you mean that make the examining logic whether should use custome ConfigSetService creation  "configSetServiceClass != null"  move into NodeConfig?  And the creation of custome ConfigSetService instance should be move to other class?

solr/core/src/java/org/apache/solr/core/NodeConfig.java
baisui1981: as i see ,there is no ready-made class to store the property 'configSetServiceClass'. it seem that make a new class to store  'configSetServiceClass' and  responsible for creating  custome ConfigSetService instance. is that ok?

solr/core/src/java/org/apache/solr/core/ConfigSetService.java
dsmiley: No.  I'm referring to the constructor signature of ConfigSetService.  Replace the Boolean (for schema caching) with NodeConfig.  If the ConfigSetService cares to honor that option (or others), it's up to it.

solr/core/src/java/org/apache/solr/core/NodeConfig.java
dsmiley: Sorry; I wasn't clear.  I simply mean move the field higher to be amongst the other fields that end with "Class" :-)

baisui1981: Ok, I got it

solr/core/src/java/org/apache/solr/core/ConfigSetService.java
baisui1981: OK，I got it.  that's easy

baisui1981:
> Looks nice! I just have some minor feedback. Thank you for contributing!

all the fix have done

solr/core/src/java/org/apache/solr/core/ConfigSetService.java
```diff
@@ -65,8 +65,8 @@ public static ConfigSetService createConfigSetService(NodeConfig nodeConfig, Sol
       try {
         Class<? extends ConfigSetService> clazz = loader.findClass(configSetServiceClass, ConfigSetService.class);
         Constructor<? extends ConfigSetService> constructor
-                = clazz.getConstructor(SolrResourceLoader.class, Boolean.TYPE, ZkController.class);
-        return constructor.newInstance(loader, nodeConfig.hasSchemaCache(), zkController);
+                = clazz.getConstructor(SolrResourceLoader.class, NodeConfig.class, ZkController.class);
```
dsmiley, line 68: Looks like we can remove SolrResourceLoader as well, which is on NodeConfig.

dsmiley (commented):
Can you please add a few lines to `solr/solr-ref-guide/src/format-of-solr-xml.adoc` about this?

Can you please also add a CHANGES.txt entry for 8.9 under Improvements?:
````
* SOLR-15100: Make the ConfigSetService pluggable/configurable via <string name="xxxxxx" /> in solr.xml
(your name here)
````

dsmiley:
Come to think of it, if the constructor was simply CoreContainer, it could get everything it might need?  That would be very clean.

baisui1981:
> Come to think of it, if the constructor was simply CoreContainer, it could get everything it might need? That would be very clean.


that's a good idea , Since that is so，shall i modify the method ConfigSetService.createConfigSetService() and simply the params list to one 'CoreContainer', that will be clean

dsmiley:
> shall i modify the method ConfigSetService.createConfigSetService() and simply the params list to one 'CoreContainer', that will be clean

Great idea!

PR #2344, opened 2021-02-10
SOLR-15131: Use collection properties instead of plugin configurations for per-collection placement plugin configs.

sigram:
See Jira for details.


sigram:
Abandoned in favor of PR 2421.

PR #2345, opened 2021-02-10
Benchmark custom

balmukundblr:
# Description

Lucene Benchmark Scaling Problem with Reuters Corpus

While Indexing 1 million documents with reuters21578 (plain text Document derived from reuters21578 corpus), we observed that with higher number of Index threads, the Index throughput does not scale and degrades. Existing implementation with synchronization block allows only one thread to pick up a document/file from list, at any given time – this code is part of getNextDocData() in ReutersContentSource.java. With multiple index threads, this becomes a thread contention bottleneck and does not allow the system CPU resource to be used efficiently.

# Solution
We developed a strategy to distribute total number of files across multiple number of Indexing threads, so that these threads work independently and parallelly.

# Tests

We mainly modified existing getNextDocData(), which is not altering  functionality, hence not added any new test cases.

- Passed existing tests

# Checklist

   [ ] I have created a Jira issue and added the issue ID to my pull request title.
  [ ] I have added tests for my changes.
  [ ] I have added documentation for the [Ref Guide](https://github.com/apache/lucene-solr/tree/master/solr/solr-ref-guide) (for Solr changes only).

### DIFF
--- a/lucene/benchmark/src/java/org/apache/lucene/benchmark/byTask/feeds/ReutersContentSource.java
+++ b/lucene/benchmark/src/java/org/apache/lucene/benchmark/byTask/feeds/ReutersContentSource.java
@@ -52,6 +52,8 @@ public class ReutersContentSource extends ContentSource {
   private ArrayList<Path> inputFiles = new ArrayList<>();
   private int nextFile = 0;
   private int iteration = 0;
+  private int[] threadIndex;
+  private volatile boolean threadIndexCreated;
 
   @Override
   public void setConfig(Config config) {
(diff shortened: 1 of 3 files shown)


lucene/benchmark/src/java/org/apache/lucene/benchmark/byTask/feeds/ReutersContentSource.java
```diff
@@ -102,19 +104,43 @@ public void close() throws IOException {
...
+     * name = f.toRealPath() + "_" +iteration;
+     * }
+     */
+    if (!threadIndexCreated) {
+      createThreadIndex();
+    }
+
+    int index = (int) Thread.currentThread().getId() % threadIndex.length;
+    int fIndex = index + threadIndex[index] * threadIndex.length;
+    threadIndex[index]++;
```
mikemccand, line 128: I'm confused how this approach ensures that we will indeed index every document in the `inputFiles`?

`Thread.currentThread().getId() % threadIndex.length` is not guaranteed to reach every possible int from `0 .. threadIndex.length`?

lucene/benchmark/src/java/org/apache/lucene/benchmark/byTask/feeds/ReutersContentSource.java
```diff
@@ -146,4 +172,11 @@ public synchronized void resetInputs() throws IOException {
     nextFile = 0;
     iteration = 0;
   }
+
+  private synchronized void createThreadIndex() {
+    if (!threadIndexCreated) {
```
mikemccand, line 177: `== false` instead?  Or maybe change to `assert threadIndexCreated == false` since you also check this up above with a real `if` already?

mikemccand (commented):
In general, I love this idea, to improve the concurrency of `ReutersContentSource`!

This source is very inefficient since it opens/closes a new file per document.  This inefficiency is why we added the `LineFileDocs` representation instead, which uses a single file and one `readLine()` per document, and (in Lucene's nightly benchmarks) the chunked binary form of `LineFileDocs` to even further reduce single-thread contention in reading/creating documents to index.

In general when testing Lucene's indexing performance it is vital to make the source of the documents as absolutely trivial as possible, to create as pure an indexing performance test as we can.

I left some small code style feedback, and also I did not understand how this concurrency optimization is functionally correct since you cannot ensure `.getId() % N` always distributes across all values `0 .. N`?

lucene/benchmark/src/java/org/apache/lucene/benchmark/byTask/feeds/ReutersContentSource.java
```diff
@@ -102,19 +104,43 @@ public void close() throws IOException {
...
+     *    throw new NoMoreDataException();
+     * }
+     * nextFile = 0;
+     * iteration++;
+     * }
+     * f = inputFiles.get(nextFile++);
+     * name = f.toRealPath() + "_" +iteration;
+     * }
+     */
+    if (!threadIndexCreated) {
```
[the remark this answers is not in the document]
balmukundblr: Sure, will do the required changes.

lucene/benchmark/src/java/org/apache/lucene/benchmark/byTask/feeds/ReutersContentSource.java
```diff
@@ -102,19 +104,43 @@ public void close() throws IOException {
...
-        }
-        nextFile = 0;
-        iteration++;
-      }
-      f = inputFiles.get(nextFile++);
-      name = f.toRealPath() + "_" + iteration;
+    int inputFilesSize = inputFiles.size();
+
+    /*
+     * synchronized (this) {
```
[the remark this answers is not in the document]
balmukundblr: Sure, will delete the commented codes.

lucene/benchmark/src/java/org/apache/lucene/benchmark/byTask/feeds/ReutersContentSource.java
balmukundblr: Sure, will do the required changes.

lucene/benchmark/src/java/org/apache/lucene/benchmark/byTask/feeds/ReutersContentSource.java
balmukundblr: Although, getId() is controlled by JVM but in our case, all threadIndex are getting initialized at once. Hence, there is high chance of getting guaranteed sequence of thread id, as we also observed. However, we understand your concern and  tweaked our code in such a way that it  guaranteed to reach every possible int from 0 .. threadIndex.length. We achieved it by setting a unique thread name and parsing the same for calculating the index value.

PR #2346, opened 2021-02-10
SOLR-15145: System property to control whether base_url is stored in state.json to enable back-compat with older SolrJ versions

thelabdude:
backport of https://issues.apache.org/jira/browse/SOLR-15145 to 8.8 branch

### DIFF
--- a/solr/CHANGES.txt
+++ b/solr/CHANGES.txt
@@ -4,6 +4,14 @@ This file lists Solr's raw release notes with details of every change to Solr.
 Most people will find the solr-upgrade-notes.adoc file more approachable.
 https://github.com/apache/lucene-solr/blob/master/solr/solr-ref-guide/src/solr-upgrade-notes.adoc
 
+==================  8.8.1 ==================
+
+Bug Fixes
+---------------------
+
+* SOLR-15145: System property to control whether base_url is stored in state.json to enable back-compat with older SolrJ versions.
+  (Timothy Potter)
+
 ==================  8.8.0 ==================
 
 Consult the LUCENE_CHANGES.txt file for additional, low level, changes in this release.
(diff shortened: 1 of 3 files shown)


solr/CHANGES.txt
```diff
@@ -4,6 +4,14 @@ This file lists Solr's raw release notes with details of every change to Solr.
 Most people will find the solr-upgrade-notes.adoc file more approachable.
 https://github.com/apache/lucene-solr/blob/master/solr/solr-ref-guide/src/solr-upgrade-notes.adoc
 
+==================  8.8.1 ==================
```
anshumg, line 7: Does it make sense to add upgrade notes and mention the property that'll allow them to upgrade the server without upgrading the client?

HoustonPutman: Yes, I very much think so. Similar to the fix for the default autoscaling policy in 8.6.1

thelabdude: The property already defaults to true, which means the `base_url` will be stored again, so really there's nothing people need to do with the sys prop to get an older SolrJ to work with an 8.8.1 server again. In reality, the sys prop is only there for people that are running the latest SolrJ and want to prepare their clusters for 9.x by explicitly setting to `false`. Doubt anyone will actually do that so the sys prop is probably not all that useful.

I still need to test to make sure an upgrade of server nodes from 8.8.0 to 8.8.1 does restore the `base_url` in ZK.

thelabdude: Added a note there, let me know if that makes sense?

anshumg (approved):
Thanks @thelabdude ! 

The upgrade note looks fine to me. We also need this in branch_8x if/when 8.9 release happens.

PR #2358, opened 2021-02-12
LUCENE-9762: FunctionScoreQuery must guard score() called twice

dsmiley:
https://issues.apache.org/jira/browse/LUCENE-9762

The score() may be called multiple times. It should take care to call DoubleValues.advanceExact only the first time, or risk faulty behavior including exceptions.

There isn't an 8.8.1 section in the CHANGES.txt on master branch but I can add it into branch_8x and branch_8_8 where eventually it will show up when the RM does a release?  Or I should just add it?


romseygeek (approved):
LGTM, thanks @dsmiley

dsmiley:
As explained in the JIRA issue, I reverted and filed a new PR: #2365

PR #2361, opened 2021-02-12
LUCENE-9768: Add source sets for src/tools, clean up forbidden API and formatting errors

dweiss:


dweiss:
@rmuir Can you check if this works with Eclipse? Tools should be included in classpath now.

dweiss:
Never mind, I see it in the generated .classpath. Should be fine.

rmuir (commented):
Thanks for bringing these tools up to the last decade here! it is good to avoid the old java.io stuff, I like good error messages.

dweiss:
I didn't really do much - I just corrected what the forbiddenapis were complaining about. I'm sure you could make them nicer. It's good they're part of the build now though!

PR #2362, opened 2021-02-12
LUCENE-9767: infrastructure for icu regeneration in place.

dweiss:


rmuir:
@dweiss This is now doing everything the ant build was doing, thanks! I tested everything works: nuked all the generated files locally, regenerated them, and git status is clean.

It still has the hairiness of having to deal with the c package, but that's nothing new. I added comments for now about how to deal with this locally.

gradle/generation/icu.gradle
```diff
@@ -0,0 +1,104 @@
...
+        args = [
+            "src/data/uax29",
+            "src/resources/org/apache/lucene/analysis/icu/segmentation"
+        ]
+      }
+    }
+  }
+
+  task regenerate() {
+    dependsOn genUtr30DataFiles
```
dweiss, line 101: these two dependsOn statements don't give ordering guarantees in gradle (genRbbi can run before genUtr). Is this correct? If there are dependencies between them then these should be declared on tasks (for example genRbbi.dependsOn genUtr...).

I'll wait for your decision whether it makes sense to automate binary downloads - if so, I'll try to do this and we can commit it cleaned up and shiny.

rmuir: thats right. the RBBI one is pure java. we can do it in parallel alongside the other one.

dweiss:
Hi Robert. Added an experimental task foo with initial steps downloading the matching ICU source package. Have to go again - family. But it should be doable to compile it from within gradle, it looks relatively easy from this point on.

PR #2365, opened 2021-02-14
LUCENE-9762: DoubleValuesSource.fromQuery bug

dsmiley:
When same doc visited twice and when the query is TPI based, an exception can be thrown.  We need to ensure the internal TPI.matches() is never called repeatedly.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/DoubleValuesSource.java
+++ b/lucene/core/src/java/org/apache/lucene/search/DoubleValuesSource.java
@@ -607,6 +607,7 @@ public abstract class DoubleValuesSource implements SegmentCacheable {
         private final TwoPhaseIterator tpi = scorer.twoPhaseIterator();
         private final DocIdSetIterator disi =
             (tpi == null) ? scorer.iterator() : tpi.approximation();
+        private Boolean tpiMatch = null; // cache tpi.matches()
 
         @Override
         public double doubleValue() throws IOException {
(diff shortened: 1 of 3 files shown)


lucene/core/src/java/org/apache/lucene/search/DoubleValuesSource.java
```diff
@@ -617,8 +618,17 @@ public double doubleValue() throws IOException {
         public boolean advanceExact(int doc) throws IOException {
           if (disi.docID() < doc) {
             disi.advance(doc);
+            tpiMatch = null;
           }
-          return disi.docID() == doc && (tpi == null || tpi.matches());
+          if (disi.docID() == doc) {
```
romseygeek, line 623: Rather than a `Boolean`, how about this?
```
private boolean matched = false; // avoid calling tpi.matches() multiple times on a single doc
...
public boolean advanceExact(int doc) throws IOException {
  if (disi.docId() == doc) {
    return matched;
  }
  if (disi.docId() < doc) {
    disi.advance(doc);
  }
  return matched = disi.docId() == doc && (tpi == null || tpi.matches());
}
```

dsmiley: At first glance, this looked simpler and correct, but on further reflection, I think this contains a bug.  Lets say advanceExact doc==1 and we call disi.advance(1) and it arrives at 2 because the disi doesn't match 1.  We'd then set matched=false and return (fine; it doesn't match doc==1).  Then say the caller calls advanceExact doc==2.  We'd quick return "matched" which is false.  But that's a mistake because the disi in fact matches doc==2 (ignoring tpi; say it's null).

I like the Boolean tpiMatch because it caches the very thing that needs to be cached, and it holds null to say we don't know or it doesn't matter.

romseygeek: Good catch.  Let's go with the Boolean then.

romseygeek (approved):

PR #2367, opened 2021-02-15
SOLR-15156: [child childFilter='...:...'] no longer escapes

dsmiley:
The query escaping it did was inconsistent with all other places in Solr where a Lucene query may be provided.

https://issues.apache.org/jira/browse/SOLR-15156


epugh (approved):
LGTM.

PR #2371, opened 2021-02-15
LUCENE-9322: Lucene90VectorReader can leak open files

iverase:
Related to #2331, this time change the logic when creating a `Lucene90VectorReader` so it closes properly files on error.


msokolov (approved):
Thanks @iverase ! I think this must have gotten rearranged during a refactor at some point

PR #2372, opened 2021-02-15
LUCENE-9773: upgrade icu to 68.2

rmuir:
Upgrade from icu 62.2 to 68.2, with Unicode 13 support.

Modify GenerateUTR30DataFiles to take the release tag as a program argument. Gradle populates this automatically, removing a manual step from regeneration process.

@dweiss made a minor tweak to regeneration: previously one of the java tools had a version number in java constant, I changed it to take this tag name as parameter, set by gradle instead.


dweiss (approved):
LGTM. I also checked and generation works on Windows with this new version, yay.

image

PR #2373, opened 2021-02-15
SOLR-15130: Allow per-collection replica placement node sets

sigram:
Details in Jira.

This branch is based on `jira/solr-15131` so that it uses collection properties for node type settings.


solr/core/src/java/org/apache/solr/cluster/placement/impl/SimpleClusterAbstractionsImpl.java
```diff
@@ -292,6 +303,17 @@ public boolean equals(Object obj) {
     public int hashCode() {
       return Objects.hash(shardName, collection, shardState);
     }
+
+    @Override
+    public String toString() {
+      return "ShardImpl{" +
+          "shardName='" + shardName + '\'' +
```
murblanc, line 310: Why quote the shard name and not the collection name?

solr/core/src/java/org/apache/solr/cluster/placement/impl/SimpleClusterAbstractionsImpl.java
```diff
@@ -432,5 +454,17 @@ public boolean equals(Object obj) {
     public int hashCode() {
       return Objects.hash(replicaName, coreName, shard, replicaType, replicaState, node);
     }
+
+    @Override
+    public String toString() {
+      return "ReplicaImpl{" +
+          "replicaName='" + replicaName + '\'' +
+          ", coreName='" + coreName + '\'' +
```
murblanc, line 462: quoting strategy on returned string unclear

solr/core/src/java/org/apache/solr/cluster/placement/plugins/AffinityPlacementFactory.java
```diff
@@ -280,25 +250,31 @@ public void verifyAllowedModification(ModificationRequest modificationRequest, P
...
-        try {
-          if (cluster.getCollection(primaryName) != null) {
-            // still exists
-            throw new PlacementModificationException("colocated collection " + primaryName +
-                " of " + deleteCollectionRequest.getCollection().getName() + " still present");
-          }
-        } catch (IOException e) {
-          throw new PlacementModificationException("failed to retrieve colocated collection information", e);
+      Set<String> colocatedWith = new HashSet<>();
+      for (SolrCollection collection : cluster.collections()) {
```
murblanc, line 254: This loop is frightening... Looping over all collection will be expensive. Can't we instead update collection properties with the "reverse" of the `withCollection` relation?
Or that means that for high scale SolrCloud clusters (WIP), `withCollection` is strongly discouraged.

murblanc (commented):
I didn't realize the previous implementation was defining `withCollection` in the plugin config.
I very (very) much prefer the way it's done here. A lot more flexible.

solr/core/src/java/org/apache/solr/cluster/placement/plugins/AffinityPlacementFactory.java
sigram: That's the most serious downside of keeping this config in collection properties - there's no easy way to keep the reverse mapping up to date. We could try updating the secondary collection... although this would require adding special case code to `OCMH.modifyCollection()`, which breaks the isolation of the plugin from the core.

murblanc: We could document to the user that in order to support safe deletes, the secondary collection must reference the primary one. Then up to the user to decide if the two collections are updated or not...
At least this leaves a scalable option available.
The other "cleaner" solution is to make collection references a first class citizen managed correctly (so we can maintain mapping and reverse mapping), but then the scope increases greatly and changes quite a few things in how collection are manages (and `withCollection` would then be just one specific usage of this collection reference mechanism, but maybe the only one...).

No perfect solution here, but once we get to very large numbers of collections (with collection lists not stored in node memory) the iteration will be prohibitively expensive (that's why I didn't want having such an iterator in the first place BTW).

sigram: Hmm, then at this point it makes more sense to keep this configuration in the plugin config :(

murblanc: I disagree, but your call.

sigram:
I'm not going to pursue this implementation - instead see PR 2421.

PR #2378, opened 2021-02-16
LUCENE-9779: Hunspell: add an API to interrupt long computations

donnerpeter:
# Description

For example, to ensure responsiveness, IntelliJ-based IDEs periodically call a special method which can throw a dedicated exception. This allows the currently executed computation to be interrupted and release all the locks it holds, allowing the UI thread to grab them as quickly as possible.

# Solution

Pass a `Runnable` to `Hunspell` constructor that may throw whatever exceptions the environment expects when needed

# Tests

`HunspellTest.testCheckCanceled`

# Checklist

### DIFF
--- a/lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/GeneratingSuggester.java
+++ b/lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/GeneratingSuggester.java
@@ -91,6 +91,7 @@ class GeneratingSuggester {
       IntsRefFSTEnum<IntsRef> fstEnum = new IntsRefFSTEnum<>(fst);
       IntsRefFSTEnum.InputOutput<IntsRef> mapping;
       while ((mapping = fstEnum.next()) != null) {
+        speller.checkCanceled.run();
         keyValueConsumer.accept(mapping.input, mapping.output);
       }
     } catch (IOException e) {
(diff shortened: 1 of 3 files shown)


dweiss:
Control-flow by exception is something I'm not particularly fond of, to be honest... I'd rather see a predicate (and graceful return) or leverage of existing Java mechanisms (Thread.checkInterrupted) to implement this. I realise a generic Runnable throwing an unchecked exception works in this case too (and is convenient for catching up the stack) but would you consider just resorting to checking the thread's interrupted status? The IDE can catch interrupted exception or wrap and rethrow as needed.

donnerpeter:
> Control-flow by exception is something I'm not particularly fond of, to be honest... I'd rather see a predicate (and graceful return) or leverage of existing Java mechanisms (Thread.checkInterrupted) to implement this. I realise a generic Runnable throwing an unchecked exception works in this case too (and is convenient for catching up the stack) but would you consider just resorting to checking the thread's interrupted status? The IDE can catch interrupted exception or wrap and rethrow as needed.

I've considered a `BooleanSupplier` and I couldn't decide which results to produce in `spell`/`suggest` during graceful return, and how to prevent callers from caching them (incorrectly). Exception solves both problems :)

There's no `Thread.checkInterrupted` in my JDK 11, did you mean something else? Anyway, `InterruptedException` is a checked one, and I'd prefer innocent clients not to care about that, so then one would have to decide which unchecked exception to use instead. Besides, the IDE that I have in mind has complicated relations with interrupted threads, which have a history of corrupting caches there. It's not unsolvable, but requires some care that not everybody has :(

dweiss:
Sorry - I wrote it off the top of my head. It should be Thread.currentThread().isInterrupted() which checks for the interrupted flag on the current thread. It doesn't throw anything, it merely checks the flag. How to return from an aborted computation is another question...

Interrupting things prematurely is always messy - I don't know if there's a clean way to do it, especially if you have sub-threads or I/O involved... 

Ok, let's leave it your way then.

PR #2380, opened 2021-02-16
SOLR-15087: Allow restoration to existing collections

gerlowskija:
# Description

As originally implemented, SolrCloud's `restore` operation only supported restoring to collections that didn't yet exist.  This was primarily done out of a fear of race conditions involving writes to the collection.  Since then the support for toggling a per-collection "readonly" property has made restoration to existing collections possible, but unimplemented.

# Solution

This PR adds a codepath to restore to an existing collection, assuming it is compatible with the stored backup files. (A backup with 2 shards can't be restored to a collection with 3, etc.).  If the collection doesn't exist, the restoration process continues to create it as it has done done in the past.

# Tests

Manual testing, and automated tests added to AbstractIncrementalBackupTest.

# Checklist

### DIFF
--- a/solr/core/src/java/org/apache/solr/cloud/api/collections/RestoreCmd.java
+++ b/solr/core/src/java/org/apache/solr/cloud/api/collections/RestoreCmd.java
@@ -28,6 +28,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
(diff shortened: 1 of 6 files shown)


solr/core/src/java/org/apache/solr/cloud/api/collections/RestoreCmd.java
```diff
@@ -67,6 +49,26 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.io.Closeable;
```
murblanc, line 52: Why move the `java` imports to after `org.apache`? Usually the fundamental imports are first, then the rest.

gerlowskija: Definitely no intentional reason - I suspect I just have some variation in my IDE's code-editor settings.

I've bootstrapped my IntelliJ with what comes out of `ant idea` though, so not sure how that could've arisen?  But I'll take a look to make sure I'm not out of line.

murblanc: I later realized my IDEA was doing the same thing and I've checked in files too with the same issue.
I since changed manually the config (settings - editor - java and then at the bottom you can define the ordering and spacing).

PR #2383, opened 2021-02-16
SOLR-15135: Use DocCollection to generate state.json format expected by UI to work with perReplicaState collections.

thelabdude:
Quick fix to render collections with `perReplicaState=true` by using DocCollection instead of reading ZK data directly.

Btw ... there is no unit test for the `ZookeeperInfoHandler` so testing for this fix was manual. Created 12 collections 6 with perReplicaState=true and 6 with perReplicaState=false and exercised the graph panel in the cloud UI. See SOLR-15098 where a unit test should be created when refactoring `ZookeeperInfoHandler`

### DIFF
--- a/solr/core/src/java/org/apache/solr/handler/admin/ZookeeperInfoHandler.java
+++ b/solr/core/src/java/org/apache/solr/handler/admin/ZookeeperInfoHandler.java
@@ -42,6 +42,8 @@ import org.apache.lucene.util.BytesRef;
 import org.apache.solr.cloud.ZkController;
 import org.apache.solr.common.SolrException;
 import org.apache.solr.common.SolrException.ErrorCode;
+import org.apache.solr.common.cloud.ClusterState;
+import org.apache.solr.common.cloud.DocCollection;
 import org.apache.solr.common.cloud.OnReconnect;
 import org.apache.solr.common.cloud.Replica;
 import org.apache.solr.common.cloud.SolrZkClient;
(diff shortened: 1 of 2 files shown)


solr/core/src/java/org/apache/solr/handler/admin/ZookeeperInfoHandler.java
```diff
@@ -516,31 +518,17 @@ void printPaginatedCollections() throws IOException {
         pagingSupport.fetchPage(page, zkClient);
         // keep track of how many collections match the filter
         boolean applyStatusFilter = (page.filterType == FilterType.status && page.filter != null);
-        List<String> matchesStatusFilter = applyStatusFilter ? new ArrayList<String>() : null;
-        Set<String> liveNodes = applyStatusFilter ?
-            zkController.getZkStateReader().getClusterState().getLiveNodes() : null;
+        List<String> matchesStatusFilter = applyStatusFilter ? new ArrayList<>() : null;
```
anshumg, line 521: should we just move this all into an `if(applyStatusFilter) {...}` block for readability? I understand it's not something you changed but while we are at it, might make sense?

solr/core/src/java/org/apache/solr/handler/admin/ZookeeperInfoHandler.java
```diff
@@ -516,31 +518,17 @@ void printPaginatedCollections() throws IOException {
...
-          } catch (Exception childErr) {
-            log.error("Failed to get {} due to", collStatePath, childErr);
-          }
-
-          if (childDataStr != null) {
-            Map<String, Object> extColl = (Map<String, Object>) Utils.fromJSONString(childDataStr);
-            // add the base_url to replica props
+          DocCollection dc = cs.getCollectionOrNull(collection);
+          if (dc != null) {
+            // TODO: for collections with perReplicaState, a ser/deser to JSON was needed to get the state to render correctly for the UI?
```
anshumg, line 529: I can't understand what the TODO here is. Code cleanup in the future?

anshumg (approved):
Thanks Tim. 

Haven't tested this but this seems good. Left a few minor comments but feel free to address them or not, nothing critical.

solr/core/src/java/org/apache/solr/handler/admin/ZookeeperInfoHandler.java
thelabdude: The TODO is a question for those that understand the `perReplicaState` code better to comment on ;-)

@chatman or @noblepaul I tried just using `dc.getProperties()` but the state of the collection did not render correctly in the UI for `perReplicaState=true` collections. But, as you can see ... a ser to JSON and back to a Map ends up having the correct state in the UI for test collections. Any thoughts on that?

PR #2392, opened 2021-02-17
LUCENE-9705: Create Lucene90DocValuesFormat and Lucene90NormsFormat

jtibshirani:
For now these are just copies of Lucene80DocValuesFormat and
Lucene80NormsFormat. The existing formats were moved to backwards-codecs.


lucene/backward-codecs/src/java/org/apache/lucene/backward_codecs/lucene80/Lucene80DocValuesFormat.java
```diff
@@ -163,6 +163,11 @@ public Lucene80DocValuesFormat(Mode mode) {
     this.mode = Objects.requireNonNull(mode);
   }
 
+  /**
+   * Note: although this format is only used on older versions, we need to keep the write logic in
+   * addition to the read logic. It's possible for doc values on older segments to be written to
+   * through in-place doc values updates.
```
jpountz, line 169: `in-place` might be a bit misleading since doc-value updates are not in-place but create a new generation of a segment?
```suggestion
   * through doc values updates.
```

lucene/core/src/java/org/apache/lucene/codecs/lucene90/Lucene90DocValuesFormat.java
```diff
@@ -0,0 +1,216 @@
...
+import org.apache.lucene.index.DocValuesType;
+import org.apache.lucene.index.IndexWriterConfig;
+import org.apache.lucene.index.SegmentReadState;
+import org.apache.lucene.index.SegmentWriteState;
+import org.apache.lucene.store.DataOutput;
+import org.apache.lucene.util.SmallFloat;
+import org.apache.lucene.util.packed.DirectWriter;
+
+/**
+ * Lucene 8.0 DocValues format.
```
jpountz, line 33:
```suggestion
 * Lucene 9.0 DocValues format.
```

jpountz (approved):

lucene/backward-codecs/src/java/org/apache/lucene/backward_codecs/lucene80/Lucene80DocValuesFormat.java
jtibshirani: Agreed, in-place can mean too many things...

